PR #1059, opened 2014-10-11
TRUNK-344 PersonName.getFullName() should use the configured nameLayout

sandeepraparthi:


djazayeri:
The approach here is incorrect. See my comments on the ticket.

PR #1138, opened 2014-11-06
TRUNK-3871 - Create UserService.saveUser(User user) method

jkondrat:


api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -719,4 +722,25 @@ public User saveUserProperties(Map<String, String> properties) {
 		return dao.saveUser(user, null);
 	}
 	
+	/**
+	 * @see UserService#changePassword(User, String, String)
+	 */
+	@Override
+	@Authorized(PrivilegeConstants.EDIT_USER_PASSWORDS)
+	@Logging(ignoredArgumentIndexes = 1)
```
wluyima: You need to ignore logging of both passwords, i see that you are only ignoring logging of the old password

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1196,4 +1198,108 @@ public void saveUserProperties_shouldRemoveAllExistingPropertiesAndAssignNewProp
...
+	
+	/**
+	 * @see UserService#changePassword(User,String,String)
+	 * @verifies throw exception if oldPassword is null and changing user have not privileges
+	 */
+	@Test(expected = APIException.class)
+	public void changePassword_shouldThrowExceptionIfOldPasswordIsNullAndChangingUserHaveNotPrivileges() throws Exception {
+		executeDataSet(XML_FILENAME_WITH_DATA_FOR_CHANGE_PASSWORD_ACTION);
+		final UserService userService = Context.getUserService();
+		//user 6001 has password userServiceTest
```
wluyima: Would be nice to add an assertion that the user actually doesn't have the privilege

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1196,4 +1198,108 @@ public void saveUserProperties_shouldRemoveAllExistingPropertiesAndAssignNewProp
...
+		String oldPassword = "userServiceTest";
+		String weakPassword = "weak";
+		userService.changePassword(user6001, oldPassword, weakPassword);
+	}
+	
+	/**
+	 * @see UserService#changePassword(User,String,String)
+	 * @verifies throw APIException if given user does not exist
+	 */
+	@Test(expected = APIException.class)
```
wluyima: We prefer to use Junit's ExpectedException approach since it also lets you check the actual error message because the exception could be from somewhere else especially for APIException which is throw from very many places. Can you do the same for the other tests where you expect and an APIException

PR #1151, opened 2014-11-13
Solved getName(Locale locale, boolean exact) simply calls getName() if exact = false,added JUnit Test 

sandeepraparthi:
...

### DIFF
--- a/api/src/main/java/org/openmrs/Concept.java
+++ b/api/src/main/java/org/openmrs/Concept.java
@@ -710,12 +710,16 @@ public class Concept extends BaseOpenmrsObject implements Auditable, Retireable,
 		
 		ConceptName exactName = getNameInLocale(locale);
 		
-		if (exactName != null || exact) {
+		if (exactName != null) {
 			return exactName;
 		}
 		
-		//just get any name
-		return getName();
+		if (exact == false) {
+			Locale broaderLocale = new Locale(locale.getLanguage());
+			ConceptName name = getNameInLocale(broaderLocale);
+			return name;
+		}
+		return null;
 	}
 	
 	/**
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/Concept.java
```diff
@@ -710,12 +710,16 @@ public ConceptName getName(Locale locale, boolean exact) {
...
 		ConceptName exactName = getNameInLocale(locale);
 		
-		if (exactName != null || exact) {
+		if (exactName != null) {
 			return exactName;
 		}
 		
-		//just get any name
-		return getName();
+		if (exact == false) {
```
djazayeri: By the way, this should be "if (!exact)" -- it's bad style test == true or == false this way

djazayeri:
For this ticket, we really should have a test that actually looks in a different locale (with the same language). E.g. we should be explicitly testing our preferred behavior when you look for a concept name in en_US but the concept only has a name in en, or the concept only has a name in en_UK.

api/src/main/java/org/openmrs/Concept.java
SandeepRaparthi1: @djazayeri Sorry for that Style of coding.However the change you advised is already updated in master https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/Concept.java#L714

SandeepRaparthi1:
@djazayeri  I Have updated the PR with test case  :)

djazayeri:
@SandeepRaparthi1 my comment was based on looking at this code now, way after the fact. Can you raise another PR with that change (since @dkayiwa merged the original one months ago)?

SandeepRaparthi1:
@djazayeri I am sorry I meant i created a PR https://github.com/openmrs/openmrs-core/pull/1472 which i posted as a comment on the  ticket :)

PR #1175, opened 2014-12-03
Use static method Integer.valueOf instead of constructor: new Integer

azuro:
Ticket: GCI-9
All tests passed.


k-joseph:
thanks @azuro

k-joseph:
See: https://github.com/openmrs/openmrs-core/pull/1335 created to resolve the conflict of removing api/src/test/java/org/openmrs/report/ReportSchemaXmlTest which happened after @azuro's PR

PR #1180, opened 2014-12-03
Changed String constructor to valueOf

chaityabshah:
http://issues.openmrs.org/browse/GCI-9
Objective of this commit is to replace new String() with String.valueOf() at several instances as valueOf is more efficient.


k-joseph:
Am merging this, it's alright to me as well

PR #1181, opened 2014-12-04
Using Random nextInt() instead of Math.random()

chaityabshah:
http://issues.openmrs.org/browse/GCI-9
Objective of this commit is to replace Math.random() with a Random object at several instances.


k-joseph:
Looks ok to me as well after a look into it, thanks @chaityabshah

PR #1192, opened 2014-12-08
GCI-9 #4

azuro:
Fixed:
-    Method may fail to close stream
-    Write to static field from instance method

### DIFF
--- a/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
+++ b/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
@@ -949,7 +949,9 @@ public class InitializationFilter extends StartupFilter {
 			// verify connection
 			//Set Database Driver using driver String
 			Class.forName(loadedDriverString).newInstance();
-			DriverManager.getConnection(databaseConnectionFinalUrl, connectionUsername, connectionPassword);
+			Connection tempConnection = DriverManager.getConnection(databaseConnectionFinalUrl, connectionUsername,
+			    connectionPassword);
+			tempConnection.close();
 			return true;
 			
 		}
@@ -1132,7 +1134,10 @@ public class InitializationFilter extends StartupFilter {
 			
 			// run the sql statement
 			Statement statement = connection.createStatement();
-			return statement.executeUpdate(replacedSql);
+			
+			int updateDelta = statement.executeUpdate(replacedSql);
+			statement.close();
+			return updateDelta;
 			
 		}
 		catch (SQLException sqlex) {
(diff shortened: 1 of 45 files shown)


web/src/main/java/org/openmrs/web/filter/update/UpdateFilter.java
```diff
@@ -350,9 +350,13 @@ protected boolean authenticateAsSuperUser(String usernameOrSystemId, String pass
...
 					String salt = results.getString(3);
 					String passwordToHash = password + salt;
-					return Security.hashMatches(storedPassword, passwordToHash) && isSuperUser(connection, userId);
+					boolean result = Security.hashMatches(storedPassword, passwordToHash) && isSuperUser(connection, userId);
+					statement.close();
+					return result;
 				}
 			}
+			// Close statement
+			statement.close();
```
harsha89: we can move this to finally block

web/src/main/java/org/openmrs/web/filter/util/FilterUtil.java
```diff
@@ -71,6 +71,9 @@ public static String restoreLocale(String username) {
 							currentLocale = results.getString(1);
 						}
 					}
+					
+					//close statement
+					statement.close();
```
harsha89: same here we should move this to finally block

web/src/main/java/org/openmrs/web/filter/util/FilterUtil.java
```diff
@@ -180,6 +183,8 @@ public static boolean storeLocale(String locale) {
 				if (statement.executeUpdate() != 1) {
 					log.warn("Unable to set system default locale property.");
 				}
+				// close statement to release resources
+				statement.close();
```
harsha89: same here we should move this to finally block

web/src/main/java/org/openmrs/web/filter/update/UpdateFilter.java
```diff
@@ -350,9 +350,13 @@ protected boolean authenticateAsSuperUser(String usernameOrSystemId, String pass
 					String storedPassword = results.getString(2);
 					String salt = results.getString(3);
 					String passwordToHash = password + salt;
-					return Security.hashMatches(storedPassword, passwordToHash) && isSuperUser(connection, userId);
+					boolean result = Security.hashMatches(storedPassword, passwordToHash) && isSuperUser(connection, userId);
+					statement.close();
```
harsha89: same here we should move this to finally block

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1132,7 +1134,10 @@ private int executeStatement(boolean silent, String user, String pw, String sql,
 			
 			// run the sql statement
 			Statement statement = connection.createStatement();
-			return statement.executeUpdate(replacedSql);
+			
+			int updateDelta = statement.executeUpdate(replacedSql);
+			statement.close();
```
harsha89: same here we should move this to finally block

PR #1207, opened 2014-12-11
Trunk 4479: EncounterService.getEncounters should use Person not User for providers

tmdugan:
Fixed TRUNK-4479: EncounterService.getEncounters should use Person not User for providers. The line endings should be correct now. I couldn't get the rebasing to work correctly through eclipse so there are three commits instead of 1.


dkayiwa:
@tmdugan do you still need this? :)

tmdugan:
No. We just added to our custom build of 1.7.x until we have a chance to
upgrade.

PR #1231, opened 2014-12-31
TRUNK-2280: Save button will be disabled and a user friendly error message will be displayed when the number of characters exceed 1024

Tharunya:


webapp/src/main/webapp/openmrs.js
```diff
@@ -626,11 +626,17 @@ function hideError(errorName) {
  * @param object(Required) to be limited.
  * @param maxLength(Required) the length of the limit.
  */
-function forceMaxLength(object, maxLength) {
-    if( object.value.length >= maxLength) {
-       object.value = object.value.substring(0, maxLength); 
-    }
-}
+ 
+function forceMaxLength(object, errorFieldId, maxLength) {
```
Tharunya: Agreed that the save button from elsewhere will not have the same id, but I assumed it to be right this way because forceMaxLength function is called only from the above changed files.. 
However, if I am wrong in my assumptions. Can you please give me a little detailed idea of how to capture the boolean value and disable the button without the ID? :)

wluyima: From the ticket summary, is it mentioned that this should be a general purpose function that will be called from anywhere, that is why you can't leave code like that behind

dkayiwa:
Closing this because the author has left the ticket for others to do.

PR #1237, opened 2015-01-07
TRUNK-4523 - Person attributes are not being displayed when viewing a vo...

jkondrat:
...ided person

### DIFF
--- a/api/src/main/java/org/openmrs/Person.java
+++ b/api/src/main/java/org/openmrs/Person.java
@@ -585,7 +585,7 @@ public class Person extends BaseOpenmrsData implements java.io.Serializable {
 		}
 		
 		attributeMap = new HashMap<String, PersonAttribute>();
-		for (PersonAttribute attribute : getActiveAttributes()) {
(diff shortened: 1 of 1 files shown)


dkayiwa:
@jkondrat reopen this after you have attached, on the ticket, a screenshot with the effects of your changes.

dkayiwa:
@jkondrat the contract for Person.getAttributeMap() is not to include voided ones. So that contract should not be broken as you have done in this commit.

jkondrat:
@dkayiwa In that case, should I introduce a new method for getting pa map including voided attributes?

dkayiwa:
@jkondrat doesn't the existing Person.getAttributes() already do so?

jkondrat:
@dkayiwa It does, but patient form iterates through all attribute types by name, so it is convenient to have them in a map form. See editPersonInfo.jsp:107

dkayiwa:
@jkondrat then in that case you can add the new convenient method you were proposing.

PR #1244, opened 2015-01-08
TRUNK-4571 - Get encounters by creation/modification date

jkondrat:


dkayiwa:
@jkondrat can you do this as a new service method which does not deprecate the old one?
Can you also make it more useful as per Nick's comments?

dkayiwa:
Am closing this pull request because we are going to use another approach different from this.

PR #1252, opened 2015-01-10
TRUNK-4507: Translate more messages for the Hindi Locale

Tharunya:


Tharunya:
@dkayiwa I have added some "message.properties" which previously didnt exist in the file "message_hi.properties". Also, I did not manipulate the already existing contents of the "message_hi.properties" file.
Is this how it was to be done ? :)

PR #1257, opened 2015-01-12
TRUNK-4500: Use of APIException not localized (mostly)

rpuzdrowski:


dkayiwa:
@rpuzdrowski you have very many places where you just pass null for the second argument of the api exception constructor. Why don't you simply overload the constructor with one that does not require the second argument?

rpuzdrowski:
@dkayiwa because the constructor already exists.
public APIException(String message) {
		super(message);
	}

PR #1273, opened 2015-01-14
TRUNK-4466: Schema design 'Find Field Elements' box should always be visible

Tharunya:


dkayiwa:
@Tharunya when the search results are many, the search box gets mixed up with the other widgets and links at the top. Just type "blo" to see this in action. You need to think harder for a better solution. :)

Tharunya:
@dkayiwa: Please refer to: https://github.com/openmrs/openmrs-core/pull/1283 :)

PR #1278, opened 2015-01-15
TRUNK-4523 - Person attributes are not being displayed when viewing a vo...

jkondrat:
...ided person


dkayiwa:
@jkondrat there are a few places where you need to set allAttributeMap to null, as is done for attributeMap

jkondrat:
@dkayiwa Oops, I didn't notice that. Updated the commit.

PR #1282, opened 2015-01-16
TRUNK-2717:Display a meaningful message when a scheduled task with no re...

SharonVarghese:
...peat interval fails to start


dkayiwa:
@SharonVarghese can you create a unit test for this?

PR #1293, opened 2015-01-21
TRUNK-3709 - Secret answer should be encrypted

jkondrat:


api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -344,7 +344,8 @@ public void changeQuestionAnswer(User u, String question, String answer) throws
 		
 		LoginCredential credentials = getLoginCredential(u);
 		credentials.setSecretQuestion(question);
-		credentials.setSecretAnswer(answer);
+		String hashedAnswer = Security.encodeString(answer + credentials.getSalt());
```
bmamlin: I would suggest `answer.toLowerCase()` to make secret answer validation case insensitive.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -360,8 +361,10 @@ public boolean isSecretAnswer(User u, String answer) throws DAOException {
 			return false;
 		}
 		
-		String answerOnRecord = getLoginCredential(u).getSecretAnswer();
-		return (answer.equals(answerOnRecord));
+		LoginCredential credentials = getLoginCredential(u);
+		String answerOnRecord = credentials.getSecretAnswer();
+		String hashedAnswer = Security.encodeString(answer + credentials.getSalt());
```
bmamlin: I would suggest `answer.toLowerCase()` to make secret answer validation case insensitive.

dkayiwa:
@jkondrat have you dealt with? Ideally, as this change is introduced, all secret answers in the database would be hashed in this manner.

PR #1302, opened 2015-01-23
TRUNK-257: Add Feature to enable moving Encounter Data to a different pa...

pmuchowski:
...tient


dkayiwa:
@pmuchowski i tried to move an encounter within a visit for a certain patient to another patient and got the error "failed to validate with reason: visit: The encounter and the visit it belongs to must both belong to the same patient". I have put the detailed stack trace on the ticket.

dkayiwa:
@pmuchowski for the message confirmation dialog box, i clicked cancel. Then i immediately tried again to click Save Encounter the second time. This time it just saved without showing the confirmation dialog box. Can you look into this?

PR #1306, opened 2015-01-26
TRUNK-4505 Display one editing concept message on edit concept page and ...

jbuczynski:
...handle when and how to validate Fully Specified Name* field


dkayiwa:
Closing this pull request because the bug it was supposed to fix cannot be reproduced.

PR #1308, opened 2015-01-27
TRUNK-4263if all the fields of option form are empty it takes it to a pa...

SandeepRaparthi1:
...ge which have an internal error


dkayiwa:
@SandeepRaparthi1 your unit tests pass even before applying your changes. Unit tests should start by failing. Then you make changes to the classes you are testing, to make the tests pass. Does this make sense?

PR #1311, opened 2015-01-29
Work in Progress - TRUNK-9 *DOES NOT RUN* (For dkayiwa)

geoff-wasilwa:
https://issues.openmrs.org/browse/TRUNK-9

While merging patients related patients data was being switched to point
to the preffered patient.

*Add method to create copies of encounters, obs and orders and void the original
*Add test to verify that encounters are being copied.


dkayiwa:
@geoff-wasilwa when you copy an encounter and hence its orders, you later on try to saving this encounter as a new one, hence the orders also as new ones. But since some of the orders you copied are no longer active, they cannot be saved as new. Does this make sense?

geoff-wasilwa:
@dkayiwa, that makes sense. But say I only migrate active orders and obs, when reviewing previous orders (not sure if this functionality exists :)) won't we be missing a certain chunk of the history (previously prescribed orders that have been deactivated)?

dkayiwa:
@geoff-wasilwa did you say that you only migrate active orders? Your code does not seem to be doing so. :)

geoff-wasilwa:
@dkayiwa I meant what if I changed my code to do that, what happens to the case where I'd like to review previous orders? My code migrates all orders and that is why it fails, and then we have the obs with a start date before encounter (but that is for another time).

dkayiwa:
@geoff-wasilwa you still need to migrate all orders, including the inactive ones. But just do not save them as new.

geoff-wasilwa:
@dkayiwa sorry for my poor choice of word (migrate). From the requirements, the orders and obs should be cloned (two copies saved; one referencing the non-preferred patient and the other referencing the preferred patient). And doing this without saving new obs or orders, doesn't seem possible to me.

dkayiwa:
@geoff-wasilwa can you create a non public method that just saves the orders while preserving their state, without doing the default processing that happens when a new order is being saved?

geoff-wasilwa:
@dkayiwa while implementing the change I noticed something interesting. It seems orders should not be copied over to the preferred patient (or should they...), since orders cannot/should not be [updated/edited](https://github.com/geoff-wasilwa/openmrs-core/blob/TRUNK-9/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java#L93-L95) and during patient merging only those with voided orders can be merged, see [here](https://github.com/geoff-wasilwa/openmrs-core/blob/TRUNK-9/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java#L807-L811). I think this should be documented, what do you say?

dkayiwa:
Can you bring this to the developers mailing list?

On Tue, Feb 17, 2015 at 12:36 PM, Geoffrey W Wasilwa <
notifications@github.com> wrote:

> @dkayiwa <https://github.com/dkayiwa> while implementing the change I
> noticed something interesting. It seems orders should not be copied over to
> the preferred patient (or should they...), since orders cannot/should not
> be updated/edited
> <https://github.com/geoff-wasilwa/openmrs-core/blob/TRUNK-9/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java#L93-L95>
> and during patient merging only those with voided orders can be merged, see
> here
> <https://github.com/geoff-wasilwa/openmrs-core/blob/TRUNK-9/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java#L807-L811>.
> I think this should be documented, what do you say?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openmrs/openmrs-core/pull/1311#issuecomment-74640299>.
>



-- 
If we keep uppermost in our minds the unkind and unjust acts of others, we
shall find it impossible to love them as Christ has loved us; but if our
thoughts dwell upon the wondrous love and pity of Christ for us, the same
spirit will flow out to others.

dkayiwa:
@geoff-wasilwa according to this email thread: https://groups.google.com/a/openmrs.org/forum/#!topic/dev/m0Lfk0Uz3Vc we are going to make changes that will make merging this pull request painful. Do you think you can finish this before Friday next week (27 Feb 2015), to avoid the unnecessary extra work?

geoff-wasilwa:
@dkayiwa did you see my comment on [TRUNK-257](https://issues.openmrs.org/browse/TRUNK-257)

geoff-wasilwa:
@dkayiwa I'd like to leverage the copyAndAssignToAnotherPatient method, but as I pointed out in [TRUNK-257](https://issues.openmrs.org/browse/TRUNK-257) copying orders leads to test failures and also violates some business rules as Darius pointed out [here](https://groups.google.com/a/openmrs.org/d/msg/dev/jMtz-y2qOUY/78695etNNuwJ)

dkayiwa:
Closing because we decided to take a different path.

PR #1317, opened 2015-02-01
TRUNK-3267: Delete any related objects here before deleting the obs

Tharunya:


dkayiwa:
@Tharunya this unit test passes even before applying your changes. Can you create one which fails and passes only after your changes?

dkayiwa:
@Tharunya according to this email thread: https://groups.google.com/a/openmrs.org/forum/#!topic/dev/m0Lfk0Uz3Vc we are going to make changes that will make merging this pull request painful. Do you think you can finish this before Friday next week (27 Feb 2015), to avoid the unnecessary extra work?

PR #1326, opened 2015-02-04
TRUNK-4480

bobbyToor:
corresponding to trunk-4480, if user wants to search for a patient 'prem singh' and types in 'prem s' , it doesn't return any matching record, while its there in the database.


dkayiwa:
@bobbyToor did you take a look at? http://go.openmrs.org/pull-request-tips
You would also need some unit tests?

dkayiwa:
@bobbyToor this commit also has compiler errors.

PR #1328, opened 2015-02-05
Rest webservices module bundled to 1.11.0 and clear format issues

k-joseph:
bundled the rest webservices module and auto formatted some files with maven


rkorytkowski:
It's a very bad practise to put binaries such as omod in the git repo. They will stay there forever and increase the size of the repo to clone. We should configure maven to download it instead and put in the war while building.

rkorytkowski:
Since we cannot do anything right now to fix it and the omod will stay there for eternity you can leave it as it is. I will create an issue in JIRA to bundle the next version properly using maven build. We definitely don't want to add 0.5 MB to the git repo with each release.

rkorytkowski:
Actually I just discovered https://github.com/openmrs/openmrs-distro-platform which promises to bundle things correctly. For the distro-platform to work you must revert the commit you just merged here.

rkorytkowski:
Could you please update https://wiki.openmrs.org/display/docs/Release+Process and mention you need to use openmrs-distro-platform to release the platform?

k-joseph:
Thanks @rkorytkowski, Surely keeping out the omod from our repo is a good idea I didn't think about before tagging. I will be updating the release process page of this which is what I used, I think we may use the distro-platform next time since I already tagged. For now we may bare this and hope to use this tool next time, probably just making another commit on 1.11.1-SNAPSHOT to remove the omod and use the tool

k-joseph:
I have updated the Release Process page and also removed the module, we shall follow that for 1.11.1 etc, 

thanks

PR #1343, opened 2015-02-17
TRUNK-309:Add Better Javadoc Comments To Services

SharonVarghese:


dkayiwa:
@SharonVarghese i have not gone through all of them because they have the same exact problem.
The method description needs to tell exactly what the method is supposed to do.
Each parameter description should be clear enough such that the caller understands what it is for?

dkayiwa:
@SharonVarghese according to this email thread: https://groups.google.com/a/openmrs.org/forum/#!topic/dev/m0Lfk0Uz3Vc we are going to make changes that will make merging this pull request painful. Do you think you can finish this before Friday next week (27 Feb 2015), to avoid the unnecessary extra work?

PR #1349, opened 2015-02-21
TRUNK-4617:patient_identifier.patient_id is missing foreign key to patie...

SandeepRaparthi1:
...nt


SandeepRaparthi1:
Closing so as to create a new one with required changes :) 
Squash was not possible because local branch was deleted apologies

PR #1352, opened 2015-02-23
changed the String.equals() method to String.equalsIgnoreCase() in User....

akshika47:
...containsRole() method and the unit test is written for the changed method


dkayiwa:
@akshika47 which ticket is this pull request for? Can you always include the ticket number in your commit messages? And in the pull request message?

PR #1355, opened 2015-02-24
Trunk-4457: User.containsRole()'s method String.equals() is changed to Str...

akshika47:
...ing.equalsIgnoreCase() and Test case is written to test the alteration.


akshika47:
@dkayiwa : I will do the following changes
1. remove is from the method name
2. use asserTrue() instead of assertEquals()
3.Add @should annotations.

Is there more that you suggest? 

Sorry for the inconvenience.

PR #1375, opened 2015-02-27
TRUNK-4497:Hide or disable (option of selecting) a child role instead of...

SharonVarghese:
... throwing an error

### DIFF
--- a/api/src/main/java/org/openmrs/Role.java
+++ b/api/src/main/java/org/openmrs/Role.java
@@ -45,6 +45,8 @@ public class Role extends BaseOpenmrsMetadata implements java.io.Serializable {
 	
 	private Set<Role> childRoles;
 	
+	private Set<Role> allChildRoles;
+	
 	// Constructors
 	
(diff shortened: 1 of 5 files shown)


dkayiwa:
@SharonVarghese does this actually fix the problem where someone has added a child role to a role and then tries to go and make the child role a parent, creating chaos?

SharonVarghese:
@dkayiwa  Yes because when we create a role called parent, then we create a role called child. This child inherits from the parent. Before applying the changes, the parent could also inherit from the child. After applying the changes, the parent cannot inherit from the child, because the button is disabled .

dkayiwa:
@SharonVarghese on the ticket, can you attach a screenshot showing the effects of your changes to confirm that what you are saying is so?

SharonVarghese:
@dkayiwa  I have attached screenshots of the changes , and while reproducing the error, if in the parent's role edit page, the child option was ticked, it would give an error, but now, the child option checkbox is disabled

PR #1378, opened 2015-02-27
removed equals methods from classes which added from GCI merge

harsha89:


k-joseph:
Thanks @harsha89 for this PR, actually we reasoned the need of this before the merge but didn't reach a consensus of removing it from gci commits, does this fix it @rkorytkowski ?

rkorytkowski:
Yes, looks good.

PR #1382, opened 2015-02-28
TRUNK_4612 correcting Call to equals() comparing different types

sashrika:
https://issues.openmrs.org/browse/TRUNK-4612 

Please refer the comments I put in the ticket to get a understand on what I have changed here.

### DIFF
--- a/api/src/main/java/org/openmrs/hl7/handler/ORUR01Handler.java
+++ b/api/src/main/java/org/openmrs/hl7/handler/ORUR01Handler.java
@@ -1293,7 +1293,7 @@ public class ORUR01Handler implements Application {
 			
 			PersonAttribute currentHealthCenter = patient.getAttribute("Health Center");
 			
-			if (currentHealthCenter == null || !currentHealthCenter.equals(newLocationId.toString())) {
+			if (currentHealthCenter == null || !currentHealthCenter.getValue().equals(newLocationId.toString())) {
 				PersonAttribute newHealthCenter = new PersonAttribute(healthCenterAttrType, newLocationId.toString());
 				
 				log.debug("Updating patient's location from " + currentHealthCenter + " to " + newLocationId);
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/openmrs/hl7/handler/ORUR01Handler.java
```diff
@@ -1293,7 +1293,7 @@ private void updateHealthCenter(Patient patient, PV1 pv1) {
 			
 			PersonAttribute currentHealthCenter = patient.getAttribute("Health Center");
 			
-			if (currentHealthCenter == null || !currentHealthCenter.equals(newLocationId.toString())) {
+			if (currentHealthCenter == null || !newLocationId.toString().equals(currentHealthCenter.getValue())) {
 				PersonAttribute newHealthCenter = new PersonAttribute(healthCenterAttrType, newLocationId.toString());
 				
 				log.debug("Updating patient's location from " + currentHealthCenter + " to " + newLocationId);
```
sashrika: Here "newLocationId" cannot be null or length equals to 0. If it so this expression doesn't execute.

dkayiwa: @sashrika by the way, do you remember how you got this? http://pastebin.com/0VaXVUDw

sashrika: @dkayiwa , I was building the Performance and Utilization Module. But i don't know a reason why i am getting this. Do you have any idea :)

dkayiwa: @sashrika i was able to reproduce and eventually fixed it. Just change it to reference openmrs 1.11.1-snapshot instead of 1.11.0-snapshot.

sashrika: @dkayiwa thanks. I tried with the above change. Still build fails. But now the exceptions are different. Earlier they were about FileNotFoundException. But now they are about mostly 
"Module unit test classes should extend BaseModuleContextSensitiveTest, not just BaseContextSensitiveTest"

full log : http://pastebin.com/pFAxdYWc
Do you have any ideas about them?

dkayiwa:
Closing because author opened a new one.

PR #1384, opened 2015-03-01
 replaced @NotTransactional with @Transactional(propagation=Propagation....

akshika47:
...NOT_SUPPORTED) in OpenmrsServiceTest.java


rkorytkowski:
It's not the right fix. See my comment on TRUNK-4606.

PR #1394, opened 2015-03-05
OpenMRSFormatter stacks enum elements one per line

teleivo:
instead of putting all elements in one line

* improves readability
* leads to good diffs, where added/deleted elements can be spotted right away


downeym:
:clap:

sunbiz:
Thank you @teleivo for your contribution

PR #1397, opened 2015-03-07
TRUNK-4627: Use StringUtils.isNotEmpty(stringObject) instead of if (stringObject == null || stringObject.equals(""))

tanayabh:
https://issues.openmrs.org/browse/TRUNK-4627


k-joseph:
Please confirm also that all results returned at a search of eqauls("") are replaced

tanayabh:
I also confirm that `object == null || object.equals("")` have also been replaced. There were very few such cases when compared to `"".equals(object)`

PR #1399, opened 2015-03-09
TRUNK-4632 Dodgy - Redundant nullcheck of value known to be non-null

tomaszmueller:
https://issues.openmrs.org/browse/TRUNK-4632


api/src/main/java/org/openmrs/PatientIdentifierType.java
```diff
@@ -44,12 +44,12 @@
 		 * Indicates that identifiers should be globally unique
 		 */
 		UNIQUE,
-
```
rkorytkowski: Why is this line removed?

tomaszmueller: those are build stylistical changes

PR #1400, opened 2015-03-09
Trunk 4633

tomaszmueller:
@rkorytkowski  https://issues.openmrs.org/browse/TRUNK-4633

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Context.java
+++ b/api/src/main/java/org/openmrs/api/context/Context.java
@@ -18,7 +18,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.WeakHashMap;
-
+import java.util.Arrays;
 import javax.mail.Authenticator;
 import javax.mail.PasswordAuthentication;
 import javax.mail.Session;
@@ -232,7 +232,7 @@ public class Context {
 		Object[] arr = userContextHolder.get();
 		
 		if (log.isTraceEnabled()) {
-			log.trace("Getting user context " + arr + " from userContextHolder " + userContextHolder);
+			log.trace("Getting user context " + Arrays.toString(arr) + " from userContextHolder " + userContextHolder);
 		}
 		
 		if (arr == null) {
@@ -1276,7 +1276,7 @@ public class Context {
 	
 	/**
 	 * @since 1.9
-	 * @see ServiceCotext#getDatatypeService()
+	 * @see ServiceContext#getDatatypeService()
 	 */
 	public static DatatypeService getDatatypeService() {
 		return getServiceContext().getDatatypeService();
--- a/web/src/main/java/org/openmrs/web/controller/patient/MergePatientsFormController.java
+++ b/web/src/main/java/org/openmrs/web/controller/patient/MergePatientsFormController.java
@@ -15,6 +15,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Vector;
+import java.util.Arrays;
 
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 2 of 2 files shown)


rkorytkowski:
You haven't used pull --rebase nor the up alias we configured yesterday, thus we see the extra merge commit. Please avoid that in the future.

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -232,7 +225,7 @@ public static UserContext getUserContext() {
 		Object[] arr = userContextHolder.get();
 		
 		if (log.isTraceEnabled()) {
-			log.trace("Getting user context " + arr + " from userContextHolder " + userContextHolder);
+			log.trace("Getting user context " + Arrays.toString(arr) + " from userContextHolder " + userContextHolder);
```
rkorytkowski: Is Arrays.toString accepting nulls? arr may be null and we do not want to see NullPointerException so please check that.

web/src/main/java/org/openmrs/web/controller/patient/MergePatientsFormController.java
```diff
@@ -55,7 +50,7 @@ protected ModelAndView processFormSubmission(HttpServletRequest request, HttpSer
 			log.debug("Error name: " + e.getObjectName());
 			log.debug("Error code: " + e.getCode());
 			log.debug("Error message: " + e.getDefaultMessage());
-			log.debug("Error args: " + e.getArguments());
+			log.debug("Error args: " + Arrays.toString(e.getArguments()));
```
rkorytkowski: The same as above.

tomaszmueller: Arrays.ToString() does handle null as shown below:

public static String toString(Object[] a) {
        if (a == null)
            return "null";

        int iMax = a.length - 1;
        if (iMax == -1)
            return "[]";

        StringBuilder b = new StringBuilder();
        b.append('[');
        for (int i = 0; ; i++) {
            b.append(String.valueOf(a[i]));
            if (i == iMax)
                return b.append(']').toString();
            b.append(", ");
        }
    }

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -10,14 +10,7 @@
...
 import java.text.SimpleDateFormat;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Properties;
-import java.util.Set;
-import java.util.WeakHashMap;
+import java.util.*;
```
SandeepRaparthi1: @tomaszmueller OpenMRS as per my Knowledge would not prefer wildcard Imports :)

tomaszmueller:
@rkorytkowski 
@SandeepRaparthi1 
fixed import issue

rkorytkowski:
There's something smelly about the commits in this pull request. There are 6 commits, but should be just 2 or 3 at most. They repeat twice. Do you remember by chance what git commands you used?

web/src/main/java/org/openmrs/web/controller/patient/MergePatientsFormController.java
rkorytkowski: Perfect! Thanks!

tomaszmueller:
When I was fixing the import issue I used "git pull origin master" command which caused a conflict that I had to merge manually and add extra commit. Could this be the reason of repeating commits?

rkorytkowski:
Probably conflict resolution didn't go that well. Unfortunately I can't say what precisely you did wrong. If you ever have to resolve conflict again please ping me before so we can pair and do it the right way. Also do not use "git pull origin master". It should be "git pull --rebase upstream master" or "git up upstream master".

rkorytkowski:
Please reach out to me tomorrow so we can share screen and pair on fixing this pull request. You'll learn some cherry-picking :)

tomaszmueller:
@rkorytkowski Hello, if you wish to share screen, I'm ready

PR #1402, opened 2015-03-10
Trunk 3465

approce:
https://issues.openmrs.org/browse/TRUNK-3465
Use static methods from StringUtils for check String empty or not empty.
Set invalidAgeFormat as static variable.
Refactor typos and redundancy.


djazayeri:
We don't generally use static imports in /src/main code, since it makes it less clear where functions are coming from. (We use static imports in /src/test code.)

Please remove the static imports, and re-issue this PR.

PR #1409, opened 2015-03-13
TRUNK-4639 ""equals(Object obj)" should be overridden along with the "co...

tomaszmueller:
...mpareTo(T obj)" method"

### DIFF
--- a/api/src/main/java/org/openmrs/PersonName.java
+++ b/api/src/main/java/org/openmrs/PersonName.java
@@ -10,6 +10,7 @@
 package org.openmrs;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 import org.apache.commons.lang.builder.EqualsBuilder;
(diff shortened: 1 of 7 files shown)


tomaszmueller:
@rkorytkowski https://issues.openmrs.org/browse/TRUNK-4639

tomaszmueller:
@rkorytkowski changed accordingly to comment in https://issues.openmrs.org/browse/TRUNK-4639

api/src/main/java/org/openmrs/PersonName.java
```diff
@@ -530,42 +531,12 @@ public String toString() {
 	 * @should return negative if other familynamePrefix is greater
 	 * @should return negative if other familyNameSuffix is greater
 	 * @should return negative if other dateCreated is greater
+	 * @Depracated since 1.12. Use DefaultCompartor instead.
```
rkorytkowski: misspelled DefaultComparator :)

web/src/main/java/org/openmrs/web/controller/encounter/EncounterDisplayController.java
```diff
@@ -387,6 +387,42 @@ public int compareTo(FieldHolder other) {
...
+					if (other.getGroupMemberConcepts() != null) {
+						return false;
+					}
+				} else if (!groupMemberConcepts.equals(other.getGroupMemberConcepts())) {
+					return false;
+				}
+			} // end else
+			
+			return true;
+		} //
```
rkorytkowski: Whenever you implement equals you must implement hashCode. Also read JavaDoc for compareTo. If equals returns true then compareTo must return 0. If equals returns false then compareTo must return != 0. You need to correct equals to behave accordingly. In other words it's enough to implement equals as return compareTo() == 0.

api/src/main/java/org/openmrs/PersonName.java
rkorytkowski: Could we please add in addition to @Deprecated the following:
Note: this comparator imposes orderings that are inconsistent with equals.

rkorytkowski: Please do the same for all other compareTo methods.

tomaszmueller:
@rkorytkowski done

web/src/main/java/org/openmrs/web/controller/encounter/EncounterDisplayController.java
```diff
@@ -387,6 +387,25 @@ public int compareTo(FieldHolder other) {
...
+		public boolean equals(FieldHolder other) {
+			return compareTo(other) == 0;
+		} //
+		
+		/**
+		 * Returns a hash code based on the formField fields used in comparing.
+		 * @see java.lang.Object#hashCode()
+		 */
+		public int hashCode() {
+			return formField.getPageNumber() * formField.getFieldNumber()
```
rkorytkowski: The rule is that hashCodes must be the same if equals returns true. It's not the case here if pageNumber is equal and some other field is different. The easiest fix here is to include just the first field that is checked in compareTo i.e. pageNumber. Please use HashCodeBuilder to generate the value.

tomaszmueller:
Fixed hashCode

web/pom.xml
```diff
@@ -136,6 +136,11 @@
 		 <artifactId>mysql-connector-mxj</artifactId>
 		 <scope>provided</scope>
 	  </dependency>
+       <dependency>
```
rkorytkowski: Please remove this dependency.

web/src/main/java/org/openmrs/web/controller/encounter/EncounterDisplayController.java
```diff
@@ -387,6 +388,23 @@ public int compareTo(FieldHolder other) {
...
+		public boolean equals(FieldHolder other) {
+			return compareTo(other) == 0;
+		} //
+		
+		/**
+		 * Returns a hash code based on the formField.getPageNumber field used in comparing.
+		 * @see java.lang.Object#hashCode()
+		 */
+		public int hashCode() {
+			return HashCodeBuilder.reflectionHashCode(formField.getPageNumber());
```
rkorytkowski: This should call .append(formField.getPageNumber()).build();

dkayiwa:
@tomaszmueller can you squash these commits into one?

tomaszmueller:
@dkayiwa  squashed

rkorytkowski:
Travis reports a failing build. The following test fails: saveShortPatient_shouldAddANewPersonAttributeWithANonEmptyValue(org.openmrs.web.controller.patient.ShortPatientFormControllerTest)
@tomaszmueller, does it pass for you locally?

tomaszmueller:
@rkorytkowski It didn't pass
I fixed it and updated the pull request

web/pom.xml
```diff
@@ -131,11 +131,6 @@
 		 <groupId>org.apache.velocity</groupId>
 		 <artifactId>velocity-tools</artifactId>
 	  </dependency>
-	  <dependency>
```
rkorytkowski: Why did you remove this?

tomaszmueller: @rkorytkowski Didn't you asked to remove this dependency in one of your comments on outdated diff 8 days ago? Or did I mistunderstood which dependency should I remove?

rkorytkowski: It was some other dependency that you included by mistake.

tomaszmueller: @rkorytkowski added the dependency again

rkorytkowski:
Well done!

PR #1417, opened 2015-03-16
TRUNK-4645 "Thread.run() and Runnable.run() should not be called directl...

tomaszmueller:
...y"

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Daemon.java
+++ b/api/src/main/java/org/openmrs/api/context/Daemon.java
@@ -164,6 +164,8 @@ public class Daemon {
 	 * @should throw error if called from a non daemon thread
 	 * @should not throw error if called from a daemon thread
 	 */
+	@SuppressWarnings("squid:3AS1217")
+	//We intentionally do not start a new thread yet, rather wrap the run call in a session.
 	public static Thread runInNewDaemonThread(final Runnable runnable) {
 		// make sure we're already in a daemon thread
 		if (!isDaemonThread()) {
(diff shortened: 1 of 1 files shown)


tomaszmueller:
@rkorytkowski https://issues.openmrs.org/browse/TRUNK-4645

api/src/main/java/org/openmrs/api/context/Daemon.java
```diff
@@ -178,7 +178,7 @@ public void run() {
 				isDaemonThread.set(true);
 				try {
 					Context.openSession();
-					runnable.run();
+                    new Thread(runnable).start();
```
rkorytkowski: This is actually intended as you can see the start method is called below in line 189: thread.start(). We need to annotate the run method (line 177) with: 
{code}
@SuppressWarnings("squid:3AS1217") //We intentionally do not start a new thread yet, rather wrap the run call in a session.
{code}

api/src/main/java/org/openmrs/api/context/Daemon.java
```diff
@@ -271,7 +271,7 @@ public void run() {
...
```
rkorytkowski: This is actually intended as you can see the start method is called below. Please revert it. We need to annotate the run method (line 270) with:

@SuppressWarnings("squid:3AS1217") //We intentionally do not start a new thread yet, rather wrap the run call in a session.

tomaszmueller:
@rkorytkowski done

PR #1420, opened 2015-03-17
TRUNK-4634 "Dodgy - Write to static field from instance method" - extrafix

tomaszmueller:


api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1423,6 +1423,15 @@ public Concept getUnknownConcept() {
 	}
 	
 	/**
+	 * Sets unknownConcept using static method
+	 *
+	 * @param currentUnknownConcept
+	 */
+	public static void setStaticUnknownConcept(Concept currentUnknownConcept) {
```
rkorytkowski: It should be private.

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -865,6 +865,15 @@ public void setHandlers(Map<String, ComplexObsHandler> newHandlers) throws APIEx
 	}
 	
 	/**
+	 * Sets handlers using static method
+	 *
+	 * @param currentHandlers
+	 */
+	public static void setStaticHandlers(Map<String, ComplexObsHandler> currentHandlers) {
```
rkorytkowski: It should be private.

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -1497,6 +1497,16 @@ public void setIdentifierValidators(Map<Class<? extends IdentifierValidator>, Id
 	}
 	
 	/**
+	 * Sets identifierValidators using static method
+	 *
+	 * @param currentIdentifierValidators
+	 */
+	public static void setStaticIdentifierValidators(
```
rkorytkowski: It should be private.

tomaszmueller:
fixed and merged into one commit

rkorytkowski:
Merged in a12f6a443bb6b750d49a4ae9304475bc26a9a037

PR #1423, opened 2015-03-19
TRUNK-4473 : Form Resource (Long-Text) causes error with binary file upload

karan173:
Fixes [TRUNK-4473](https://issues.openmrs.org/browse/TRUNK-4473)

### DIFF
--- a/web/src/main/java/org/openmrs/web/controller/form/FormResourceController.java
+++ b/web/src/main/java/org/openmrs/web/controller/form/FormResourceController.java
@@ -10,12 +10,15 @@
 package org.openmrs.web.controller.form;
 
 import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpSession;
 
 import org.apache.commons.lang.StringUtils;
+import org.jfree.util.Log;
 import org.openmrs.Form;
 import org.openmrs.FormResource;
 import org.openmrs.api.context.Context;
 import org.openmrs.customdatatype.CustomDatatypeUtil;
+import org.openmrs.web.WebConstants;
 import org.openmrs.web.attribute.WebAttributeUtil;
 import org.springframework.stereotype.Controller;
 import org.springframework.ui.Model;
@@ -81,7 +84,14 @@ public class FormResourceController {
 		if (errors.hasErrors()) {
 			throw new RuntimeException("Error handling not yet implemented");
 		} else {
-			Context.getFormService().saveFormResource(resource);
+			try {
+				Context.getFormService().saveFormResource(resource);
+			}
+			catch (Exception ex) {
+				HttpSession httpSession = request.getSession();
+				Log.error("Error while adding form resource", ex);
+				httpSession.setAttribute(WebConstants.OPENMRS_ERROR_ATTR, "form.resource.add.failure");
+			}
 			return "redirect:formResources.form?formId=" + resource.getForm().getId();
 		}
 	}


web/src/main/java/org/openmrs/web/controller/form/FormResourceController.java
```diff
@@ -10,12 +10,15 @@
 package org.openmrs.web.controller.form;
 
 import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpSession;
 
 import org.apache.commons.lang.StringUtils;
+import org.jfree.util.Log;
```
djazayeri: This is wrong. Please see other OpenMRS code to see the standard way we log, by convention (e.g. it uses log4j, not jfree.util).

web/src/main/java/org/openmrs/web/controller/form/FormResourceController.java
```diff
@@ -81,7 +84,14 @@ public String handleAddFormResource(@ModelAttribute("resource") FormResource res
...
 			throw new RuntimeException("Error handling not yet implemented");
 		} else {
-			Context.getFormService().saveFormResource(resource);
+			try {
+				Context.getFormService().saveFormResource(resource);
+			}
+			catch (Exception ex) {
+				HttpSession httpSession = request.getSession();
+				Log.error("Error while adding form resource", ex);
+				httpSession.setAttribute(WebConstants.OPENMRS_ERROR_ATTR, "form.resource.add.failure");
```
djazayeri: Is there a way that we can give a more specific error message? (E.g. in the specific case reported in this ticket, someone tried to upload a binary file when only text is accepted. Is there any way we can tell that's the error and report it?)

djazayeri:
Required: please log consistently with standard OpenMRS practice
Optional: is there a way we can give a more specific error message in some cases?

(You can reopen this PR after addressing those issues.)

karan173:
@djazayeri Can you please refer me to some code where such error handling is done. The errors thrown are all related to the database layer. There are various kind of resources present and a no of errors are being thrown, for eg, when one or both fields are empty in the form. Hence, I am not able to extract meaningful info. from the exceptions thrown.

djazayeri:
I suggest you look into whether it's possible to have the underlying code (in LongFreeTextDatatype or DatatypeServiceImpl) throw a meaningful exception in this specific case, or else report a more meaningful error message as a string message in the InvalidCustomValueException. It's probably better to display the message from there, rather than displaying _nothing_ to the user.

karan173:
But the error is not localized to the LongFreeTextDatatype. There is no error handling and input validation for adding resources to a form. So even if we leave the resource name and/or value empty, uncaught exceptions are thrown. So should I just focus on the case described solely in this issue?

I was thinking of first adding validation for the presence of resource name and resource value form fields(since they can't be null). According to me we should only then think about catching exceptions specific to a datatype.

karan173:
Another thing I have found is that, the handlers(say DateFieldGenDatatypeHandler) are not throwing InvalidCustomValueException for incompatible resource values because in the getValue method of WebAttributeUtils([link](https://github.com/openmrs/openmrs-core/blob/master/web/src/main/java/org/openmrs/web/attribute/WebAttributeUtil.java#L56)), a classCastException is being thrown, while typecasting the handler to its subclasses.

web/src/main/java/org/openmrs/web/controller/form/FormResourceController.java
karan173: @djazayeri  Do we wish to upload only files of the content type text/plain?

karan173: @djazayeri should we only allow text/plain files or files like text/html as well?

PR #1424, opened 2015-03-19
adding hibernate 4 support to openmrs

wluyima:


api/src/main/java/org/openmrs/api/db/hibernate/NativeIfNotAssignedIdentityGenerator.java
```diff
@@ -19,7 +19,7 @@
 import org.hibernate.HibernateException;
 import org.hibernate.MappingException;
 import org.hibernate.dialect.Dialect;
-import org.hibernate.engine.SessionImplementor;
+import org.hibernate.engine.spi.SessionImplementor;
 import org.hibernate.id.Configurable;
 import org.hibernate.id.IdentityGenerator;
 import org.hibernate.persister.entity.EntityPersister;
```
wluyima: @harsha89  @djazayeri  isn't this the custom id generator that will no longer be needed since we are using the standard hibernate jar file, i see it is being used in the mapping files for concepts and active list items

api/src/main/java/org/openmrs/util/HibernateEnumType.java
```diff
@@ -105,6 +106,16 @@ public int hashCode(Object x) throws HibernateException {
 		return x.hashCode();
 	}
 	
+	public Object nullSafeGet(ResultSet resultSet, String[] strings, SessionImplementor sessionImplementor, Object o)
+	        throws HibernateException, SQLException {
+		return null;
```
wluyima: Do you have to implement this?

api/src/main/java/org/openmrs/util/HibernateEnumType.java
```diff
@@ -105,6 +106,16 @@ public int hashCode(Object x) throws HibernateException {
...
 	}
 	
+	public Object nullSafeGet(ResultSet resultSet, String[] strings, SessionImplementor sessionImplementor, Object o)
+	        throws HibernateException, SQLException {
+		return null;
+	}
+	
+	public void nullSafeSet(PreparedStatement preparedStatement, Object o, int i, SessionImplementor sessionImplementor)
+	        throws HibernateException, SQLException {
+
```
wluyima: Do you have to implement this too?

api/src/main/java/org/openmrs/util/MemoryLeakUtil.java
```diff
@@ -99,19 +96,19 @@ public static void shutdownKeepAliveTimer() {
 	 * Clears the hibernate session factories cached in the SessionFactoryObjectFactory
 	 */
 	public static void clearHibernateSessionFactories() {
-		try {
```
wluyima: Did you test that this is no longer necessary? If yes, then we can get rid of the code instead of commenting it out

api/src/main/java/org/openmrs/api/db/hibernate/NativeIfNotAssignedIdentityGenerator.java
djazayeri: Correct, we should email the dev list and start publicizing that we'll likely remove this functionality in OpenMRS 1.12.

rkorytkowski:
Please send a pull request for the hibernate4 branch after addressing issues. I'm closing this one. @harsha89, please do not forget to reply to Wyclif's questions.

PR #1426, opened 2015-03-20
Prevent SNAPSHOT versions from being uploaded

joshim5:
In response to an introductory ticket posted on OpenMRS Issues.

https://issues.openmrs.org/browse/MOD-8


dkayiwa:
@joshim5 did you notice that there is a unit test which fails in ModuleUtilTest?

joshim5:
@dkayiwa Yes -- This is because the OpenMRS unit tests test the uploading of SNAPSHOT versions. Seeing as we no longer allow SNAPSHOT versions, should these unit tests be changed?

dkayiwa:
@joshim5 the changes should be in https://github.com/openmrs/openmrs-contrib-modulus

maywuy:
if we were to fix this issue, should we make changes to the unit test?

PR #1431, opened 2015-03-24
TRUNK-4654 "Performance - Method invokes inefficient Number constructor;...

tomaszmueller:
... use static valueOf instead"


tomaszmueller:
@rkorytkowski https://issues.openmrs.org/browse/TRUNK-4654

PR #1435, opened 2015-03-25
TRUNK-4642: Refactoring location class

yousefhamza:
Refactoring Locaiton class address attirbute

change hardcoded addresses attributes with one
array containing all of them with keeping all
the exisitng APIs as it is, and adding new
method to return all the address.

Signed-off-by: Yousef Hamza <jo.adam.93@gmail.com>


dkayiwa:
@rkorytkowski and @wluyima what do you generally think of this proposed refactoring?

PR #1446, opened 2015-03-27
TRUNK-4686 Warning during Compilation - KeepAliveCache

Ch3ck:
Fixed Warning during compilation


Ch3ck:
So what do you recommend here? @dkayiwa?

PR #1450, opened 2015-03-30
TRUNK-3382 Modify provider.getName() method so that it only looks at provider.person.name

alexei-grigoriev:
https://issues.openmrs.org/browse/TRUNK-3382

I believe it is better now.


alexei-grigoriev:
Closing it. Now I can see there are some more work in scope of the issue and more files should be changed.
dkayiwa, thank You a lot. It is much more clear now.

PR #1451, opened 2015-03-30
TRUNK 4385 "Annotated BaseOpenmrs... classes"

tomaszmueller:


djazayeri:
I'm surprised that doing this isn't going to break classes that are mapped via XML. But as long as you all test it...

api/src/main/java/org/openmrs/api/impl/BaseOpenmrsService.java
```diff
@@ -20,6 +22,7 @@
  * 
  * @see OpenmrsService
  */
+@MappedSuperclass
```
rkorytkowski: This should be removed.

tomaszmueller: removed

api/src/main/java/org/openmrs/BaseOpenmrsMetadata.java
```diff
@@ -22,29 +25,40 @@
...
  */
+@MappedSuperclass
 public abstract class BaseOpenmrsMetadata extends BaseOpenmrsObject implements OpenmrsMetadata {
 	
 	//***** Properties *****
+	@Column(name = "name", nullable = false, length = 255)
 	@Field
 	private String name;
 	
+	@Column(name = "description", length = 255)
```
wluyima: Shouldn't this be longer than 255?

tomaszmueller: 255 is the most common length value for description property in hibernate mapping files for classes derived from BaseOpenmrsMetadata. In some of them however, larger values are used, like: 500, 1024, 65535 or 2147483847. Should it be changed for one of these values?

wluyima: Then i guess we can leave it as 255 here and override it in the concrete classes

tomaszmueller: @wluyima I did the mapping in concrete classes in extra commit
@rkorytkowski  I'll squash the commits after I get the confirmation that it's OK.

rkorytkowski: This is wrong. You should not have overridden any existing classes. In fact annotations are ignored for entities mapped in xml. Reading Hibernate docs: "To ease the migration process from hbm files to annotations, the configuration mechanism detects the mapping duplication between annotations and hbm files. HBM files are then prioritized over annotated metadata on a class to class basis. You can change the priority using hibernate.mapping.precedence property. The default is hbm, class, changing it to class, hbm will prioritize the annotated classes over hbm files when a conflict occurs." https://docs.jboss.org/hibernate/stable/annotations/reference/en/html/ch01.html
So please revert changes that you made after Wyclif's comment.

The correct way to override any field is by using @AttributeOverwrite, but you must not use it for this issue. The issue is only about introducing annotations to Base.. classes. http://stackoverflow.com/questions/5257921/hibernate-how-override-an-attribute-from-mapped-super-class

tomaszmueller: @rkorytkowski reverted

tomaszmueller: @rkorytkowski
@wluyima
I created an issue for overriding the description value in concrete classes
https://issues.openmrs.org/browse/TRUNK-4674

PR #1457, opened 2015-04-01
TRUNK-4386 "Fix loading messages in tests for modules"

tomaszmueller:


tomaszmueller:
@rkorytkowski should I add additional classpath to surefireraports plugin in pom for the tests to look in the correct WEB-INF folder like in the pull request? 
What assertions you had on mind in your email as in openmrs-core only assertions connected to MutableResourceBundleMessageSource i could find were in AdministrationServiceTest in tests concerning locales?
Or should I create a separate test for checking MutableResourceBundleMessageSource.findPropertiesFiles and include the assertions there?

tomaszmueller:
That's how it was intended, but i'm not sure how should I test it and what more is needed to do. I need some more explanations from @rkorytkowski

rkorytkowski:
This is not correct. I replied on the issue.

PR #1461, opened 2015-04-03
TRUNK-3382 Modify provider.getName() method so that it only looks at provider.person.name

alexei-grigoriev:
Creating a new pull request. Please review
https://issues.openmrs.org/browse/TRUNK-3382


alexei-grigoriev:
bad rebase

PR #1463, opened 2015-04-06
TRUNK-4603 Test fails under JVM 1.8

approce:
Exceptions in TaskHelperTest and SchedullerFormControllerTest when run under JVM 1.8. 
Resolves: 
https://issues.openmrs.org/browse/TRUNK-4598
https://issues.openmrs.org/browse/TRUNK-4599
https://issues.openmrs.org/browse/TRUNK-4601
https://issues.openmrs.org/browse/TRUNK-4602
https://issues.openmrs.org/browse/TRUNK-4603 (here I have posted explanation)
https://issues.openmrs.org/browse/TRUNK-4604

Also, after merge this PR, we can resolve https://issues.openmrs.org/browse/TRUNK-3948 just by removing "@Ignore" annotation and it will works.

### DIFF
--- a/web/src/test/java/org/openmrs/scheduler/web/controller/SchedulerFormControllerTest.java
+++ b/web/src/test/java/org/openmrs/scheduler/web/controller/SchedulerFormControllerTest.java
@@ -151,6 +151,7 @@ public class SchedulerFormControllerTest extends BaseWebContextSensitiveTest {
 		assertTrue(mav.getModel().isEmpty());
 		
 		Assert.assertSame(oldTaskInstance, task.getTaskInstance());
+		deleteAllData();
 	}
 	
 	/**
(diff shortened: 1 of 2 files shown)


web/src/test/java/org/openmrs/scheduler/web/controller/SchedulerFormControllerTest.java
```diff
@@ -151,6 +151,7 @@ public void onSubmit_shouldNotRescheduleAnExecutingTask() throws Exception {
 		assertTrue(mav.getModel().isEmpty());
 		
 		Assert.assertSame(oldTaskInstance, task.getTaskInstance());
+		super.deleteAllData();
```
wluyima: I think calling deleteAllData() should be enough without the super since it is not overidden in this class

approce: @wluyima sure it is. Should I rewtite without super?

approce:
@wluyima have removed super.

wluyima:
Thanks!

PR #1464, opened 2015-04-07
Resolve JIRA #3933

justincohler:
https://issues.openmrs.org/browse/TRUNK-3933

Updated the ForgotPasswordFormController to select from a list of dummy questions in the case where a user enters an invalid username or a username with no existing security questions.

For a user's IP address, they will have 5 attempts on the fake question before being locked out. The dummy question selected is selected based on IP address. If for some reason the IP address cannot be read, a random question from the dummy list will be selected.


dkayiwa:
@justincohler  can you squash these commits into one as advised at? http://go.openmrs.org/pull-request-tips

dkayiwa:
@justincohler this pull request has lots more than it should. Just create a new one.

PR #1465, opened 2015-04-08
Add Spring Data Jpa Config, add EncounterDao jpa interface

StasPereverziev:
I get stuck with issue https://issues.openmrs.org/browse/TRUNK-4667
When spring trying to initializemy dao bean I am getting an exception
"Caused by: java.lang.IllegalStateException: No supertype found".

### DIFF
--- a/api/src/main/java/org/openmrs/config/DatabaseConfig.java
+++ b/api/src/main/java/org/openmrs/config/DatabaseConfig.java
@@ -1,0 +1,109 @@
+package org.openmrs.config;
+
+import org.springframework.context.annotation.*;
+import org.springframework.core.env.Environment;
+import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
+import org.springframework.jdbc.datasource.DriverManagerDataSource;
+import org.springframework.orm.jpa.JpaTransactionManager;
+import org.springframework.orm.jpa.JpaVendorAdapter;
+import org.springframework.orm.jpa.LocalContainerEntityManagerFactoryBean;
+import org.springframework.orm.jpa.vendor.HibernateJpaVendorAdapter;
+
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/openmrs/config/DatabaseConfig.java
```diff
@@ -0,0 +1,109 @@
...
+import javax.sql.DataSource;
+import java.util.Properties;
+
+@Configuration
+@EnableJpaRepositories("org.openmrs.api.db")
+@PropertySource("classpath:hibernate.default.properties")
+public class DatabaseConfig {
+	
+	@Bean
+	LocalContainerEntityManagerFactoryBean entityManagerFactory(DataSource dataSource, Environment env) {
```
rkorytkowski: My understanding is that this will result in us having both entityManagerFactory and hibernateSessionFactory, thus using up extra resources and adding complexity. Please see http://stackoverflow.com/a/21673855 for the alternative.

rkorytkowski: See also http://docs.spring.io/spring-data/jpa/docs/1.8.0.RELEASE/reference/html/#_infrastructure

rkorytkowski: Another thing to consider is that we use custom https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java, the functionality of which must be preserved.

StasPereverziev:
@lluismf it's just an initial PR, and of course it would be hardcoded in merged candidate.

rkorytkowski:
I will close this PR for now, until we find a way out.

PR #1466, opened 2015-04-08
TRUNK-4672 Correctness - int value cast to float and then passed to Math...

tomaszmueller:
....round


tomaszmueller:
https://issues.openmrs.org/browse/TRUNK-4672
"Correctness - int value cast to float and then passed to Math.round"

PR #1469, opened 2015-04-10
TRUNK-4639 "equals(Object obj)" should be overridden along with the "com...

tomaszmueller:
...pareTo(T obj)" method"


tomaszmueller:
@rkorytkowski https://issues.openmrs.org/browse/TRUNK-4639

PR #1482, opened 2015-04-22
Add method executeSQLWithSession that works with Session

StasPereverziev:
TRUNK-4676

### DIFF
--- a/api/src/main/java/org/openmrs/util/DatabaseUtil.java
+++ b/api/src/main/java/org/openmrs/util/DatabaseUtil.java
@@ -14,13 +14,12 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.Vector;
+import java.util.*;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.hibernate.Session;
+import org.hibernate.jdbc.Work;
 import org.openmrs.api.db.DAOException;
 import org.springframework.util.StringUtils;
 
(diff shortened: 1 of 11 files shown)


wluyima:
@StasPereverziev  can you please include the ticket number in the title or description of this pull request?

api/src/main/java/org/openmrs/util/DatabaseUtil.java
```diff
@@ -94,9 +93,80 @@ public static String loadDatabaseDriver(String connectionUrl, String connectionD
 	}
 	
 	/**
-	 * Executes the passed SQL query, enforcing select only if that parameter is set
+	 * Executes the passed SQL query, enforcing select only if that parameter is set for given Session
 	 */
-	public static List<List<Object>> executeSQL(Connection conn, String sql, boolean selectOnly) throws DAOException {
+	public static List<List<Object>> executeSQLWithSession(Session session, String sql, boolean selectOnly)
```
rkorytkowski: I would recommend renaming it back to executeSQL.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
```diff
@@ -94,9 +93,80 @@ public static String loadDatabaseDriver(String connectionUrl, String connectionD
...
+		final List<List<Object>> result = new ArrayList<List<Object>>();
+		final String query = sql;
+		final boolean sessionDataManipulation = dataManipulation;
+		
+		//todo replace with lambdas after moving on to Java 8
+		session.doWork(new Work() {
+			
+			@Override
+			public void execute(Connection conn) {
+				PreparedStatement ps = null;
```
rkorytkowski: Instead of duplicating code you should call "executeSQL(connection, sql, selectOnly)". Use session.doReturningWork to free yourself from using the final result object.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
```diff
@@ -94,9 +93,80 @@ public static String loadDatabaseDriver(String connectionUrl, String connectionD
...
+			}
+		});
+		
+		return result;
+	}
+	
+	/**
+	 * Executes the passed SQL query, enforcing select only if that parameter is set for given Connection
+	 */
+	public static List<List<Object>> executeSQLWithConnection(Connection conn, String sql, boolean selectOnly)
```
rkorytkowski: I would recommend naming it executeSQL (as it was before). No need to introduce backwards incompatibility here. It would be nice to include in a JavaDoc that "since OpenMRS 1.12 it is preferred to pass Session managed by Hibernate instead of Connection if possible".

api/src/main/java/org/openmrs/util/DatabaseUtil.java
rkorytkowski: The javadoc should state that "It is preferred to use session.createSQLQuery with correct synchronization (sqlQuery.addSynchronizedEntityClass) to invalidate session cache if necessary."

api/src/main/java/org/openmrs/util/DatabaseUtil.java
rkorytkowski: I've just noticed that doReturningWork is not available in the old version of Hibernate, which this branch is still using. For now I do approve of using doWork and copying results from executeSQL(connection, sql, selectOnly) to the final result object.

PR #1491, opened 2015-05-04
TRUNK-4497: Hide or disable (option of selecting) a child role instead o...

midnighteuler:
...f throwing an error

This is a continuation of jbuczynski/sharonvarghese abandoned pull requests.
- Updated UserServiceTest
- Has localization for English, German, Spanish
- Design Note: hides the items in the UI if they are descendant rather than showing disabled and with a tooltip. I think this is more intuitive--but I'm willing to go with the tooltip approach if preferred.

trunk4497_hidden_child_role


dkayiwa:
@midnighteuler when i click on the Child Role link, it takes me to a page with an empty role name. Can you look into that?

midnighteuler:
I'll take a look.

PR #1497, opened 2015-05-08
TRUNK-4703 fix jetty-maven-plugin configuration

sintjuri:
https://issues.openmrs.org/browse/TRUNK-4703


sintjuri:
The new one use extraClasspath which is documented here http://download.eclipse.org/jetty/8.1.16.v20140903/apidocs/org/eclipse/jetty/webapp/WebAppContext.html#setExtraClasspath%28java.lang.String%29

PR #1500, opened 2015-05-12
TRUNK-4701 throw an AmbiguousOrderException while trying to discontinue

sintjuri:
https://issues.openmrs.org/browse/TRUNK-4701

### DIFF
--- a/api/src/main/java/org/openmrs/api/AmbiguousOrderException.java
+++ b/api/src/main/java/org/openmrs/api/AmbiguousOrderException.java
@@ -15,7 +15,7 @@ package org.openmrs.api;
  * @since 1.12
  */
 public class AmbiguousOrderException extends APIException {
-
+	
 	private static final long serialVersionUID = -2946935560419378572L;
 	
 	public AmbiguousOrderException(String message) {
--- a/api/src/main/java/org/openmrs/api/AmbiguousOrderException.java
+++ b/api/src/main/java/org/openmrs/api/AmbiguousOrderException.java
@@ -1,0 +1,31 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api;
+
+/**
+ * This exception is thrown on attempt to do some action with order, and there 
+ * are multiple active orders for the given concept so the action is ambiguous
+ */
+public class AmbiguousOrderException extends APIException {
+	
+	/**
+	 * 
+	 */
+	private static final long serialVersionUID = -2946935560419378572L;
+	
+	public AmbiguousOrderException() {
+		super();
+	}
+	
+	public AmbiguousOrderException(String message) {
+		super(message);
+	}
+	
+}
--- a/api/src/main/java/org/openmrs/api/AmbiguousOrderException.java
+++ b/api/src/main/java/org/openmrs/api/AmbiguousOrderException.java
@@ -12,17 +12,11 @@ package org.openmrs.api;
 /**
  * This exception is thrown on attempt to do some action with order, and there 
  * are multiple active orders for the given concept so the action is ambiguous
+ * @since 1.12
  */
 public class AmbiguousOrderException extends APIException {
-	
-	/**
-	 * 
-	 */
+
 	private static final long serialVersionUID = -2946935560419378572L;
-	
-	public AmbiguousOrderException() {
-		super();
-	}
 	
 	public AmbiguousOrderException(String message) {
 		super(message);
--- a/api/src/main/java/org/openmrs/api/OrderService.java
+++ b/api/src/main/java/org/openmrs/api/OrderService.java
@@ -91,6 +91,8 @@ public interface OrderService extends OpenmrsService {
 	 * @should pass for revision order if an active test order for the same concept and care settings exists
 	 * @should roll the autoExpireDate to the end of the day if it has no time component
 	 * @should not change the autoExpireDate if it has a time component
+	 * @should discontinue first Drug Order with matched Drug
+	 * @should throw AmbiguousOrderException if disconnecting multiple active orders for the given concept
 	 */
 	@Authorized( { PrivilegeConstants.EDIT_ORDERS, PrivilegeConstants.ADD_ORDERS })
 	public Order saveOrder(Order order, OrderContext orderContext) throws APIException;
--- a/api/src/main/java/org/openmrs/api/OrderService.java
+++ b/api/src/main/java/org/openmrs/api/OrderService.java
@@ -92,6 +92,7 @@ public interface OrderService extends OpenmrsService {
 	 * @should roll the autoExpireDate to the end of the day if it has no time component
 	 * @should not change the autoExpireDate if it has a time component
 	 * @should throw AmbiguousOrderException if disconnecting multiple active orders for the given concept
+	 * @should throw AmbiguousOrderException if disconnecting multiple active drug orders with the same drug
 	 */
 	@Authorized( { PrivilegeConstants.EDIT_ORDERS, PrivilegeConstants.ADD_ORDERS })
 	public Order saveOrder(Order order, OrderContext orderContext) throws APIException;
--- a/api/src/main/java/org/openmrs/api/OrderService.java
+++ b/api/src/main/java/org/openmrs/api/OrderService.java
@@ -91,7 +91,6 @@ public interface OrderService extends OpenmrsService {
 	 * @should pass for revision order if an active test order for the same concept and care settings exists
 	 * @should roll the autoExpireDate to the end of the day if it has no time component
 	 * @should not change the autoExpireDate if it has a time component
-	 * @should discontinue first Drug Order with matched Drug
 	 * @should throw AmbiguousOrderException if disconnecting multiple active orders for the given concept
 	 */
 	@Authorized( { PrivilegeConstants.EDIT_ORDERS, PrivilegeConstants.ADD_ORDERS })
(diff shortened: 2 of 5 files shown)


api/src/main/java/org/openmrs/api/AmbiguousOrderException.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api;
+
+/**
+ * This exception is thrown on attempt to do some action with order, and there 
+ * are multiple active orders for the given concept so the action is ambiguous
+ */
```
wluyima: missing @since

api/src/main/java/org/openmrs/api/AmbiguousOrderException.java
```diff
@@ -0,0 +1,31 @@
...
+public class AmbiguousOrderException extends APIException {
+	
+	/**
```
wluyima: You might as well get rid of these since they are blank

api/src/main/java/org/openmrs/api/AmbiguousOrderException.java
```diff
@@ -0,0 +1,31 @@
...
+	 * 
+	 */
+	private static final long serialVersionUID = -2946935560419378572L;
+	
+	public AmbiguousOrderException() {
+		super();
```
wluyima: To avoid an exception with no helpful error message, i would get rid of this default constructor

api/src/main/java/org/openmrs/api/OrderService.java
```diff
@@ -88,6 +88,8 @@
 	 * @should pass if an active order for the same concept exists in a different care setting
 	 * @should roll the autoExpireDate to the end of the day if it has no time component
 	 * @should not change the autoExpireDate if it has a time component
+	 * @should discontinue first Drug Order with matched Drug
```
wluyima: as per the ticket description i would expect the opposite of this

wluyima: So you probably don't need this @should

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -320,28 +321,37 @@ private void discontinueExistingOrdersIfNecessary(Order order) {
 		    null);
 		boolean isDrugOrderAndHasADrug = DrugOrder.class.isAssignableFrom(getActualType(order))
 		        && ((DrugOrder) order).getDrug() != null;
+		Order orderToBeDiscontinued = null;
 		for (Order activeOrder : orders) {
```
wluyima: Shouldn't this entire for loop go away?

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -320,28 +321,37 @@ private void discontinueExistingOrdersIfNecessary(Order order) {
...
-				stopOrder(activeOrder, aMomentBefore(order.getDateActivated()));
-				break;
+				if(orderToBeDiscontinued==null){
+					orderToBeDiscontinued = activeOrder;
+				}else{
+					throw new AmbiguousOrderException();
+				}
 			}
 		}
+		if(orderToBeDiscontinued!=null){
```
wluyima: And this should go away i.e the if clause

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
wluyima: In fact, i think you need to get rid of this discontinueExistingOrdersIfNecessary() method since this is no longer necessary after all it is called in one palce

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -2814,4 +2814,40 @@ public void saveOrder_shouldSetAutoExpireDateForReviseOrderWithSimpleDosingInstr
 		
 		assertNotNull(revisedOrder.getAutoExpireDate());
 	}
```
wluyima: I believe there was a test that ensured the existing order gets discontinued, and i would assume after your change it would fail and then it passes once you specify the expected exception class and message

api/src/test/resources/org/openmrs/api/include/OrderServiceTest-discontinueAmbiguousOrderByConcept.xml
```diff
@@ -0,0 +1,20 @@
...
+    This Source Code Form is subject to the terms of the Mozilla Public License,
+    v. 2.0. If a copy of the MPL was not distributed with this file, You can
+    obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+    the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+
+    Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+    graphic logo is a trademark of OpenMRS Inc.
+
+-->
+<dataset>
```
wluyima: I don't think you need this test data, you should just place an order for with a concept or drug that matches that of an existing one

sintjuri:
1. >> In fact, i think you need to get rid of this discontinueExistingOrdersIfNecessary() method since this is no longer necessary after all it is called in one palce

I think you mean method org.openmrs.api.impl.OrderServiceImpl.checkDrugOrdersForDiscontinuing(DrugOrder, DrugOrder)
I made it just to avoid 3 level of ifs. Should I remove this method?
2. >> Order orderToBeDiscontinued = null;
this variable is used to store order to be discontinued. In case of drag orders, when there are two drug orders with same drug we will discontinue the last founded. In case of orders with same concept we will throw AmbiguousOrderException on attempt to store second order to this variable. 
Should this logic be changed?
3. >> I believe there was a test that ensured the existing order gets discontinued, and i would assume after your change it would fail and then it passes once you specify the expected exception class and message
I found org.openmrs.api.OrderServiceTest.saveOrder_shouldPassIfTheExistingDrugOrderMatchesTheConceptAndThereIsNoDrugOnThePreviousOrder(), which I think you mean. But in this order Drug Orders are tested. And Drug Orders have to be discontinued in without AmbiguousOrderException. 
May be you can show me Test, where there are two Orders (not Drug Orders) with same concept and attempt to discontinue one of them.
4. >> I don't think you need this test data, you should just place an order for with a concept or drug that matches that of an existing one
I prepared this test data for situation described in 3. Should I remove this and in the test store a new Order with same concept and them try to discontinue this order?

wluyima:
Wait to make changes, have you read the recent ticket comments?

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -320,28 +321,36 @@ private void discontinueExistingOrdersIfNecessary(Order order) {
...
-					shouldMarkAsDiscontinued = true;
-				}
-			} else if (activeOrder.getConcept().equals(order.getConcept())) {
-				shouldMarkAsDiscontinued = true;
-			}
-			
-			if (shouldMarkAsDiscontinued) {
-				order.setPreviousOrder(activeOrder);
-				stopOrder(activeOrder, aMomentBefore(order.getDateActivated()));
+				orderToBeDiscontinued = checkDrugOrdersForDiscontinuing((DrugOrder) order, (DrugOrder) activeOrder);
```
wluyima: The same applies for drug orders, if someone calls discontinue a given drug and there is multiple orders with a matching drug and lets say different strengths, then the call is ambiguous too

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -2814,4 +2814,22 @@ public void saveOrder_shouldSetAutoExpireDateForReviseOrderWithSimpleDosingInstr
...
+		DrugOrder order = new DrugOrder();
+		order.setAction(Order.Action.DISCONTINUE);
+		order.setOrderReasonNonCoded("Discontinue this");
+		order.setConcept(conceptService.getConcept(88));
+		order.setEncounter(encounterService.getEncounter(7));
+		order.setPatient(patientService.getPatient(9));
+		order.setOrderer(providerService.getProvider(1));
+		order.setCareSetting(orderService.getCareSetting(1));
+		order = (DrugOrder) orderService.saveOrder(order, null);
+	}
```
wluyima: Can you also add a test for drug orders too?

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -2814,4 +2814,22 @@ public void saveOrder_shouldSetAutoExpireDateForReviseOrderWithSimpleDosingInstr
 		
 		assertNotNull(revisedOrder.getAutoExpireDate());
 	}
+	
+	/**
+	 * @see OrderServiceImpl#discontinueExistingOrdersIfNecessary()
+	 * @verifies throw AmbiguousOrderException if disconnecting multiple active orders for the given concepts
+	 */
```
wluyima: Can you please add the @should to the method in OrderService?

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -325,28 +326,39 @@ private void discontinueExistingOrdersIfNecessary(Order order) {
...
+			order.setPreviousOrder(orderToBeDiscontinued);
+			stopOrder(orderToBeDiscontinued, aMomentBefore(order.getDateActivated()));
+		}
+	}
+	
+	
+	private DrugOrder checkDrugOrdersForDiscontinuing(DrugOrder drugOrder1, DrugOrder drugOrder2){
+		if (OpenmrsUtil.nullSafeEquals(drugOrder1.getDrug(), drugOrder2.getDrug())) {
+			return drugOrder2;
+		}else{
```
wluyima: You don't need the else clause, it is redundant, just return null

wluyima:
Had one little comment, otherwise it looks good, can you look into why the travis build above is failing?

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -325,28 +326,38 @@ private void discontinueExistingOrdersIfNecessary(Order order) {
...
 			//For drug orders, the drug must match if the order has a drug
 			if (isDrugOrderAndHasADrug) {
-				DrugOrder drugOrder1 = (DrugOrder) order;
-				DrugOrder drugOrder2 = (DrugOrder) activeOrder;
-				if (OpenmrsUtil.nullSafeEquals(drugOrder1.getDrug(), drugOrder2.getDrug())) {
-					shouldMarkAsDiscontinued = true;
-				}
+				if(orderToBeDiscontinued==null){
+					orderToBeDiscontinued = checkDrugOrdersForDiscontinuing((DrugOrder) order, (DrugOrder) activeOrder);
+				}else{
```
wluyima: This logic is actually not exactly worng because checkDrugOrdersForDiscontinuing could return null, the else clause on line 338 above assumes checkDrugOrdersForDiscontinuing won't return null, i will fix it

alexei-grigoriev:
Java files are with wrong formatting. 'mvn package' getting them changed by formatting. Could You please check it?

PR #1510, opened 2015-06-17
Add @annotations mapping to Visit and VisitType entities

spereverziev:
I've added a sample mapping as an example to work with TRUNK-4666 - https://issues.openmrs.org/browse/TRUNK-4666


dkayiwa:
Closing this because it is simply for demonstration purposes.

PR #1512, opened 2015-06-23
TRUNK-4702. OrderService should allow you to save overlapping DrugOrders

sintjuri:
if the caller acknowledges that they are aware of the overlap, otherwise
AmbiguousOrderException have to be thrown

https://issues.openmrs.org/browse/TRUNK-4702

### DIFF
--- a/api/src/main/java/org/openmrs/api/OrderService.java
+++ b/api/src/main/java/org/openmrs/api/OrderService.java
@@ -32,6 +32,8 @@ import org.openmrs.util.PrivilegeConstants;
  */
 public interface OrderService extends OpenmrsService {
 	
+	public static final String PARALLEL_ORDERS = "PARALLEL_ORDERS";
+	
 	/**
 	 * Setter for the Order data access object. The dao is used for saving and getting orders
 	 * to/from the database
@@ -85,7 +87,7 @@ public interface OrderService extends OpenmrsService {
 	 * @should pass if an active order for the same concept exists in a different care setting
 	 * @should set Order type of Drug Order to drug order if not set and concept not mapped
 	 * @should set Order type of Test Order to test order if not set and concept not mapped
-	 * @should fail if an active drug order for the same drug formulation exists
+	 * @should throw AmbiguousOrderException if an active drug order for the same drug formulation exists
 	 * @should pass if an active order for the same concept exists in a different care setting
 	 * @should fail for revision order if an active drug order for the same concept and care settings exists
 	 * @should pass for revision order if an active test order for the same concept and care settings exists
--- a/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
@@ -14,6 +14,7 @@ import static org.openmrs.Order.Action.REVISE;
 
 import java.lang.reflect.Field;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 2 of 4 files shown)


api/src/main/java/org/openmrs/api/OrderService.java
```diff
@@ -85,7 +85,7 @@
 	 * @should pass if an active order for the same concept exists in a different care setting
 	 * @should set Order type of Drug Order to drug order if not set and concept not mapped
 	 * @should set Order type of Test Order to test order if not set and concept not mapped
-	 * @should fail if an active drug order for the same drug formulation exists
+	 * @should throw AmbiguousOrderException if an active drug order for the same drug formulation exists
 	 * @should pass if an active order for the same concept exists in a different care setting
```
wluyima: You also need a test to ensure that it passes for overlapping orders in case the parallel orders are specified

api/src/main/java/org/openmrs/api/OrderService.java
```diff
@@ -98,6 +98,20 @@
...
+	* exists on the orderContext, then it will be set to the one associated to the ConceptClass of
+	* the ordered concept otherwise the save fails. If the CareSetting field of the order is not
+	* specified then it will default to the one set on the passed in OrderContext if any otherwise
+	* the save fails. This method is different from saveOrder(Order order, OrderContext orderContext) above: 
+	* param parallelOrders contains known overlapping orders to be bypassed in overlapping check, otherwise 
+	* AmbiguousOrderException will be thrown
+	* @should pass if an known drug order for the same drug formulation specified
+	*/
+	@Authorized( { PrivilegeConstants.EDIT_ORDERS, PrivilegeConstants.ADD_ORDERS })
+	public Order saveOrder(Order order, OrderContext orderContext, Order[]  parallelOrders) throws APIException;
```
wluyima: I think you should use the orderContext for now instead and not overload this method

wluyima: I think the parallel orders should be uuids instead of order objects

sintjuri:
Hi Wyclif,

I changed array of Orders to array of UIDs. It is clear. I also removed overridden method 
saveOrder(Order order, OrderContext orderContext, Order[] parallelOrders)
despite of requirements in https://issues.openmrs.org/browse/TRUNK-4702. What about test is not clear. I have already test org.openmrs.api.OrderServiceTest.saveOrder_shouldPassIfAnKnownDrugOrderForTheSameDrugFormulationSpecified(). Should I write another one?

dkayiwa:
@wluyima what do you think of the above?

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -201,18 +202,22 @@ public synchronized Order saveOrder(Order order, OrderContext orderContext) thro
...
+ 			List<Order> activeOrders = getActiveOrders(order.getPatient(), null, order.getCareSetting(), null);
+			List<String> parallelOrders = Collections.emptyList();
+			if(orderContext!=null && orderContext.getAttribute(PARALLEL_ORDERS)!=null){
+				parallelOrders= Arrays.asList((String[])orderContext.getAttribute(PARALLEL_ORDERS));
 			}
-		}
+ 			for (Order activeOrder : activeOrders) {
+ 				//Reject if there is an active drug order for the same orderable with overlapping schedule
+				if (!parallelOrders.contains(activeOrder.getUuid()) && areDrugOrdersOfSameOrderableAndOverlappingSchedule(order, activeOrder)) {
+					throw new AmbiguousOrderException("Order.cannot.have.more.than.one", (Object[]) null);
```
wluyima: It is pointless to pass in a null value here when there is a constructor that takes no argument, i will remove it

PR #1517, opened 2015-07-02
TRUNK-4711: Fix javadoc compilation errors

suniala:
Fix all the errors that cause the javadoc generation to fail. Leave the
problems that only cause a warning to be reported. Remove the previously
added javadoc error suppression now that all errors are fixed.

Note that javadocs in test classes were not fixed as javadoc generation
does not cover test resources.

Summary of the types of problems fixed:

 * @link usage in @see tags

 * @return descriptions missing - The new @return descriptions not just
   fix the validation issues byt try to be meaningful (to the extent that
   the method contract could be easily figured out from the implementation
   or method description).

 * invalid @link or @see references - Some of the references were not
   meaningful any more and were just removed. Others were fixed to point at
   a most likely intended target.

 * invalid @param parameter names

 * javadoc tags in invalid case

 * unknown javadoc tags

 * invalid html markup

 * exceptions that are not actually thrown


suniala:
Now then, I do not have any idea as to why the CI build is failing. If I do "mvn clean install" I get:

```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] OpenMRS ........................................... SUCCESS [2.045s]
[INFO] openmrs-tools ..................................... SUCCESS [1.296s]
[INFO] openmrs-test ...................................... SUCCESS [0.152s]
[INFO] openmrs-api ....................................... SUCCESS [2:17.351s]
[INFO] openmrs-web ....................................... SUCCESS [58.216s]
[INFO] openmrs-webapp .................................... SUCCESS [8.305s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
```

If I do "mvn clean package javadoc:aggregate-jar" I get:

```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] OpenMRS ........................................... SUCCESS [24.676s]
[INFO] openmrs-tools ..................................... SUCCESS [1.345s]
[INFO] openmrs-test ...................................... SUCCESS [0.084s]
[INFO] openmrs-api ....................................... SUCCESS [2:09.815s]
[INFO] openmrs-web ....................................... SUCCESS [53.343s]
[INFO] openmrs-webapp .................................... SUCCESS [8.331s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
```

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -665,7 +665,7 @@ public void saveGlobalProperty_shouldNotAllowDifferentPropertiesToHaveTheSameStr
 	}
 	
 	/**
-	 * @see {@link AdministrationService#saveGlobalProperties(List<QGlobalProperty;>)}
+	 * @see AdministrationService#saveGlobalProperties(List<QGlobalProperty;>)
```
wluyima: I see a couple of these 'Q's after the opening angel brackets, can you please remove them?

suniala:
Looks like the "Q...;" type parameters are only used in test classes. As I described in the commit comment, I decided that javadocs of test classes are out of scope. The reasons being that the test class javadocs do not get validated (or even compiled?) and that there was enough work even without the test class javadocs. However, some of the regex stuff I did did also fix problems in test classes. Hence the modified line of code you mentioned.

As it stands, fixing the "Q...;" type parameters is out of scope (of the unit of work I have committed) and I suggest we leave them as they are for now. Or was there some specific reason why these particular problems in test class javadocs should be fixed?

Of course we can and should also discuss about what to do with the test class javadocs.

wluyima:
Ok, that's fine

PR #1519, opened 2015-07-07
TRUNK-4571: Get encounters by creation/modification date

suniala:
Deprecate the old api method and add a new one that uses a parameter
object (EncounterSearchCriteria) instead of individual parameters.
Provide a convenience builder EncounterSearchCriteriaBuilder for
creating unmodifiable instances of the search criteria.


suniala:
> Did you just forget to include "sorted" or "ordered" in the above should?

I agree that the phrasing in that should tag is a bit unclear. Actually it refers to the filtering and not ordering. Just like in the deprecated getEncounters(...) method there are "@should get encounters by location" and "@should get encounters on or after date".

PR #1526, opened 2015-07-15
Trunk 4408

macorrales:
I have some commits due to syncing with upstream that I did not expect, however only one file change. I am not sure how to create a clean pull request only with the single commit that matters. Only idea that occurs to me is start over again. Let me know whether that is needed.


dkayiwa:
@macorrales did you start by pulling from upstream master?

PR #1530, opened 2015-07-26
TRUNK-4614: Multithreaded correctness - Call to static DateFormat

macorrales:


dkayiwa:
@macorrales what do you think is the performance implication of this change, in reference to the original?

macorrales:
I think it is going to perform  worse as new instances of SimpleDateFormat will be created every time those methods are used. I don't think it will perform so bad as to be noticed though. Do you have ideas about how to measure that?  Also,  sonar bug is about avoiding static use of DateFormat, not sure what other alternatives are there. Any ideas are welcome.

dkayiwa:
@macorrales thinking about this again, since these static fields are assigned values once and never changed, do you think the ticket is valid about the multithread safety?

macorrales:
I guess not, if code is kept that way. I understand that Sonar rule is about potential risks, the reference could be assigned a different value by another developer later on and there you could have the issue.

dkayiwa:
@macorrales since the value assignment is on declaration, how about if we just declare them as constants to prevent being assigned a different value by another developer?

macorrales:
@dkayiwa only the reference to the object will be constant. The object itself can change, ie following code will compile:

    private static final DateFormat dateFormatter = new SimpleDateFormat(DATE_TIME_FORMAT);

...

    dateFormatter.setCalendar(Calendar.getInstance());

dkayiwa:
@macorrales you are right! :)

PR #1536, opened 2015-08-03
TRUNK-3624: Created Program data type support

foolchan2556:
@dkayiwa Squashed here


api/src/main/java/org/openmrs/customdatatype/datatype/ProgramDatatype.java
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * @see org.openmrs.customdatatype.SerializingCustomDatatype#deserialize(java.lang.String)
+     * @should return the object by given uuid string
+     * @override
+     */
+    public Program deserialize(String uuid){
+        if (null == uuid)
+            return null;
+        return new ProgramWorkflowServiceImpl().getProgramByUuid(uuid);
```
wluyima: This won't work, you need to use Context.getProgramWorkflowService() and not new ProgramWorkflowServiceImpl()

PR #1537, opened 2015-08-03
TRUNK-3624: Created Program data type support

foolchan2556:
Hey thanks, how about it then? @wluyima


api/src/main/java/org/openmrs/customdatatype/datatype/ProgramDatatype.java
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * @see org.openmrs.customdatatype.SerializingCustomDatatype#deserialize(java.lang.String)
+     * @should return the object by given uuid string
+     * @override
+     */
+    public Program deserialize(String uuid){
+        if (null == uuid)
+            return null;
+        return Context.getProgramWorkflowService().getProgramByUuid(uuid);
```
wluyima: I already committed the fix and closed the ticket, thanks though!

PR #1539, opened 2015-08-06
TRUNK-4473: Form Resource (Long-Text) causes error with binary file u…

ivange94:
Caught exception and provided an appropriate error message. This is a screen cast of what the page would look like when user tries to upload binary files.
openmrs_form_resource


ivange94:
created a new one

PR #1546, opened 2015-08-09
TRUNK-4691 - Force isSet when concept has set members

michaelhofer:
Since my last pull request did not match the guidelines, I have now created a new one. 
This is for TRUNK-4691, ConceptService.saveConcept now forces isSet==true, if the concept has set members.


michaelhofer:
@dkayiwa I have now added the unit test as requested and squashed the commits into one.

michaelhofer:
@dkayiwa thanks for your comment, I have now added the @should annotations. the test passes on my system.

dkayiwa:
@michaelhofer i still do not see the @should annotation on the service.

michaelhofer:
@dkayiwa sorry for that, seems like my latest commit was not included earlier. I hope everything is now ok, sorry for the trouble.

PR #1553, opened 2015-08-10
TRUNK-4473: Handled exception thrown when user tries to upload binary…

ivange94:
… files with form resource


robbyoconnor:
Why close this -- you can just commit and squash and then push up to your fork and it will appear here...

PR #1555, opened 2015-08-11
Fixing bug that two clicks were necessary to force  the year menu to …

smv-recommind:
I decided to patch Jquery DatePicker. Created jquery-ui-timepicker-month-year-patch.js patch and included it in headerFull.js
Tested. Worked as expected on Google Chrome and IE.


dkayiwa:
@smv-recommind does it still work in Firefox? :)

smv-recommind:
Yes, of course it works in Firefox. Checked.

PR #1561, opened 2015-08-12
TRUNK-4473: Fixed error occuring when user tries to upload binary fil…

ivange94:
…es with form resource

### DIFF
--- a/web/src/test/java/org/openmrs/web/controller/form/FormResourceControllerTest.java
+++ b/web/src/test/java/org/openmrs/web/controller/form/FormResourceControllerTest.java
@@ -1,0 +1,65 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.web.controller.form;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import org.hibernate.exception.ConstraintViolationException;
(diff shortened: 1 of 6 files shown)


dkayiwa:
@ivange94 when done with the above, remember to add some unit tests.

ivange94:
Yes, i though the same thing.

ivange94:
@dkayiwa i already had a custom exception on standby, all that is left is the unit test. will make my final commit soon

dkayiwa:
Good thinking! :)

dkayiwa:
Test both the service layer change and the controller.

ivange94:
ok

ivange94:
For the FormServiceImpTest, i know i have to pass an image file to the form resource, so that i can i can simulate what happens when an invalid file is passed to the resource. But the documentation of the org.openmrs.FormResource does not tell which field holds files. Also Since this error is thrown when saving in the database, am i suppose to create a test database?

dkayiwa:
Before committing, did you do an mvn clean install to ensure that the formatting is as per our convention? http://go.openmrs.org/pull-request-tips

web/src/test/java/org/openmrs/web/controller/form/FormResourceControllerTest.java
```diff
@@ -0,0 +1,92 @@
...
+    	byte[] fileContent = "Hello World".getBytes();
+    	String fileName = "/resources/testfile.ico";
+    	MockMultipartHttpServletRequest request = new MockMultipartHttpServletRequest();
+    	MockMultipartFile file = new MockMultipartFile("resourceValue", fileName, "text/plain", fileContent);
+    	
+    	request.setMethod("POST");
+    	request.setParameter("datatype", (new LongFreeTextDatatype()).getClass().getName());
+    	request.setParameter("handler", (new LongFreeTextFileUploadHandler()).getClass().getName());
+    	request.addFile(file);
+
```
ivange94: @dkayiwa no matter what value i pass to the request and the form resource, can't get it to throw the expected exceptions in test. What am saying is, on the page where the form resource is been saved, if you submit no or invalid file, you get an error immediately due to the exceptions thrown. But when writing the unit tests, no matter what file i pass to the multipart request, no exception is thrown.  Also in the FormServiceImpl.saveFormResource() method, no matter what values i contstruct the resource with, it always seem to complete successfully. I don't know why.

PR #1568, opened 2015-09-18
TRUNK-4752: User password should be validated before changing it

ivange94:
Added password validation using OpenmrsUtil.validatePassword() to UserServiceImpl changePassword, methods and also created test cases for every change i made

https://issues.openmrs.org/browse/TRUNK-4752

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
@@ -350,6 +350,7 @@ public class UserServiceImpl extends BaseOpenmrsService implements UserService {
 	 */
 	@Deprecated
 	public void changePassword(User u, String pw) throws APIException {
+		OpenmrsUtil.validatePassword(u.getUsername(), pw, u.getSystemId());
 		dao.changePassword(u, pw);
 	}
 	
@@ -357,7 +358,9 @@ public class UserServiceImpl extends BaseOpenmrsService implements UserService {
 	 * @see org.openmrs.api.UserService#changePassword(java.lang.String, java.lang.String)
 	 */
 	public void changePassword(String pw, String pw2) throws APIException {
-		dao.changePassword(pw, pw2);
+		User user = Context.getAuthenticatedUser();
+		
+		changePassword(user, pw, pw2);
 	}
 	
 	/**
@@ -734,6 +737,7 @@ public class UserServiceImpl extends BaseOpenmrsService implements UserService {
 		} else if (!dao.getLoginCredential(user).checkPassword(oldPassword)) {
 			throw new APIException("old.password.not.correct", (Object[]) null);
 		}
+	
 		OpenmrsUtil.validatePassword(user.getUsername(), newPassword, user.getSystemId());
 		dao.changePassword(user, newPassword);
 	}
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -364,6 +369,8 @@ public void changePassword(String pw, String pw2) throws APIException {
 	 * @see org.openmrs.api.UserService#changeHashedPassword(User, String, String)
 	 */
 	public void changeHashedPassword(User user, String hashedPassword, String salt) throws APIException {
+		OpenmrsUtil.validatePassword(user.getUsername(), hashedPassword, user.getSystemId());
```
wluyima: Does this really work? That validate method i think expects non hashed passwords, probably you shouldn't validate this, the ticket says changePassword methods. I also think these should delegate to the changePassword method that validates the password to centralize the code

api/src/test/java/org/openmrs/api/impl/UserServiceImplTest.java
```diff
@@ -0,0 +1,47 @@
...
+	
+	@Before
+	public void before() {
+		userService = Context.getUserService();
+		user = new User();
+		user.setUsername("john");
+		user.setSystemId("1");
+	}
+	
+	@Test(expected = PasswordException.class)
```
wluyima: Use junit's @ExpectedException instead of this because it allows to check the actual error message since there is several sub classes of this with different error messages

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -350,6 +350,8 @@ public Role saveRole(Role role) throws APIException {
 	 */
 	@Deprecated
 	public void changePassword(User u, String pw) throws APIException {
+		OpenmrsUtil.validatePassword(u.getUsername(), pw, u.getSystemId());
```
wluyima: As i mentioned earlier, these should delegate to the changePassword method that validates the password since it does some useful checks that these need too.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -350,14 +350,16 @@ public Role saveRole(Role role) throws APIException {
 	 */
 	@Deprecated
 	public void changePassword(User u, String pw) throws APIException {
-		dao.changePassword(u, pw);
+		changePassword(u, "test", pw);
```
wluyima: Is this stray? Or do you actually set it to test?

ivange94: i did but it was temporally. I was intending to ask you if it was possible to get a user's password given that i already have a User object. I used test because in my Unit test, i used the default autenticated user who's password is test but i new this would not be okay because if in feature this password changes, my unit test will break also, if that changePassword method is called with a user who's password is not test will cause the code to fail. So my question now is, given that i have a user object, how can i retrieve his password as a string. Thanks

wluyima: This method was a mistake in the first place, i guess you have to validate the new password manually and keep the old code for now, it is going to be removed soon since it is deprecated. Sorry about that!

ivange94:
I have done manual validation for new password using OpenmrsUtil.validatePassword() as you requested for that particular deprecated method. Is it ok that way?

PR #1570, opened 2015-09-24
TRUNK-4637 : delete temp jar files related to modules

jvena1:
ref: https://issues.openmrs.org/browse/TRUNK-4637


jvena1:
1. the '.' is not automatically generated between the file name and the suffix
2. existing code expects a '.' (see for example line 1075)
3. allows tools to recognize the file type

dkayiwa:
Then how come it was working well before your change?
I want to make sure the new changes do not break things.
The method you mentioned is not related to this in any way. :)

PR #1581, opened 2015-09-29
TRUNK-4760, Implement API method for retrieving only the active encou…

cioan:
…nter providers


cioan:
Sorry @mogoodrich for all those identical repeatable requests. Thanks!

PR #1585, opened 2015-10-04
TRUNK-4767 - Fix env variable use

sunbiz:


sunbiz:
@dkayiwa not accidentally. The documentation says system property (including env variable) take precedence over runtime property. So the system property (-Dvar passed) is first, env var second and then runtime property third and fallback lastly.

dkayiwa:
After having looked at the code again, the only use of the environmental variable is in getting the location of the runtime properties file "OpenmrsUtil.getRuntimePropertiesFilePathName". And the name of this environmental variable is not fixed. It depends on the webapp name. As for "OpenmrsUtil.getApplicationDataDirectory", it uses the "OPENMRS_APPLICATION_DATA_DIRECTORY" system property and "application_data_directory" runtime property". That is as far as the code and javadoc comments are.
So could we be looking at different documentation? :)

PR #1588, opened 2015-10-05
TRUNK-4761, Ability to add free text drug orders

sravanthi17:
Please review the code..
https://issues.openmrs.org/browse/TRUNK-4761

### DIFF
--- a/api/src/main/java/org/openmrs/DrugOrder.java
+++ b/api/src/main/java/org/openmrs/DrugOrder.java
@@ -526,7 +526,7 @@ public class DrugOrder extends Order implements java.io.Serializable {
         }
         DrugOrder otherDrugOrder = (DrugOrder) otherOrder;
 
-        if (isNonCodedDrug() && otherDrugOrder.isNonCodedDrug()) {
+        if (isNonCodedDrug() || otherDrugOrder.isNonCodedDrug()) {
             return OpenmrsUtil.nullSafeEqualsIgnoreCase(this.getDrugNonCoded(), otherDrugOrder.getDrugNonCoded());
         }
         return OpenmrsUtil.nullSafeEquals(this.getDrug(), otherDrugOrder.getDrug());
--- a/api/src/main/java/org/openmrs/DrugOrder.java
+++ b/api/src/main/java/org/openmrs/DrugOrder.java
@@ -526,7 +526,7 @@ public class DrugOrder extends Order implements java.io.Serializable {
         }
         DrugOrder otherDrugOrder = (DrugOrder) otherOrder;
 
-        if (isNonCodedDrug() || otherDrugOrder.isNonCodedDrug()) {
+        if (isNonCodedDrug() && otherDrugOrder.isNonCodedDrug()) {
             return OpenmrsUtil.nullSafeEqualsIgnoreCase(this.getDrugNonCoded(), otherDrugOrder.getDrugNonCoded());
         }
         return OpenmrsUtil.nullSafeEquals(this.getDrug(), otherDrugOrder.getDrug());
--- a/api/src/main/java/org/openmrs/DrugOrder.java
+++ b/api/src/main/java/org/openmrs/DrugOrder.java
@@ -9,6 +9,7 @@
  */
 package org.openmrs;
 
+import org.apache.commons.lang.StringUtils;
 import static org.openmrs.Order.Action.DISCONTINUE;
 import org.openmrs.util.OpenmrsUtil;
 
@@ -486,7 +487,7 @@ public class DrugOrder extends Order implements java.io.Serializable {
 	public String toString() {
 		String prefix = DISCONTINUE == getAction() ? "DC " : "";
 		return prefix + "DrugOrder(" + getDose() + getDoseUnits() + " of "
-		        + (getDrugNonCoded() != null ? getDrugNonCoded() : (getDrug() != null ? getDrug().getName() : "[no drug]")) + " from " + getDateActivated() + " to "
+		        + (isNonCodedDrug() ? getDrugNonCoded() : (getDrug() != null ? getDrug().getName() : "[no drug]")) + " from " + getDateActivated() + " to "
 		        + (isDiscontinuedRightNow() ? getDateStopped() : getAutoExpireDate()) + ")";
 	}
 
@@ -544,7 +545,7 @@ public class DrugOrder extends Order implements java.io.Serializable {
 	 * sets drugNonCoded
 	 */
 	public void setDrugNonCoded(String drugNonCoded) {
-		this.drugNonCoded = drugNonCoded != null ? drugNonCoded.trim() : drugNonCoded;
+		this.drugNonCoded = StringUtils.isNotBlank(drugNonCoded) ? drugNonCoded.trim() : drugNonCoded;
 	}
 
 	/**
@@ -552,6 +553,6 @@ public class DrugOrder extends Order implements java.io.Serializable {
 	 * return true if a drug is non coded
 	 */
 	public boolean isNonCodedDrug() {
-		return getDrugNonCoded() != null;
+		return StringUtils.isNotBlank(this.drugNonCoded);
 	}
 }
--- a/api/src/main/java/org/openmrs/DrugOrder.java
+++ b/api/src/main/java/org/openmrs/DrugOrder.java
@@ -14,58 +14,60 @@ import org.openmrs.util.OpenmrsUtil;
 
 /**
  * DrugOrder
- * 
+ *
  * @version 1.0
  */
 public class DrugOrder extends Order implements java.io.Serializable {
-	
+
 	public static final long serialVersionUID = 72232L;
-	
+
 	// Fields
-	
+
 	private Double dose;
-	
+
 	private Concept doseUnits;
-	
+
 	private OrderFrequency frequency;
-	
+
 	private Boolean asNeeded = false;
-	
+
 	private Double quantity;
-	
+
 	private Concept quantityUnits;
-	
+
 	private Drug drug;
-	
+
 	private String asNeededCondition;
-	
+
 	private Class<? extends DosingInstructions> dosingType = SimpleDosingInstructions.class;
-	
+
 	private Integer numRefills;
-	
+
 	private String dosingInstructions;
-	
+
 	private Integer duration;
-	
+
 	private Concept durationUnits;
-	
+
 	private Concept route;
-	
+
 	private String brandName;
-	
+
 	private Boolean dispenseAsWritten = Boolean.FALSE;
-	
+
+	private String drugNonCoded;
+
 	// Constructors
-	
+
 	/** default constructor */
 	public DrugOrder() {
 	}
-	
+
 	/** constructor with id */
 	public DrugOrder(Integer orderId) {
 		this.setOrderId(orderId);
 	}
-	
+
 	/**
 	 * @see org.openmrs.Order#copy()
 	 * @should copy all drug order fields
@@ -73,7 +75,7 @@ public class DrugOrder extends Order implements java.io.Serializable {
 	public DrugOrder copy() {
 		return copyHelper(new DrugOrder());
 	}
-	
+
 	/**
 	 * @see org.openmrs.Order#copyHelper(Order)
 	 */
@@ -95,63 +97,64 @@ public class DrugOrder extends Order implements java.io.Serializable {
 		target.setRoute(getRoute());
 		target.setBrandName(getBrandName());
 		target.setDispenseAsWritten(getDispenseAsWritten());
+		target.setDrugNonCoded(getDrugNonCoded());
 		return target;
 	}
-	
+
 	public boolean isDrugOrder() {
 		return true;
 	}
-	
+
 	// Property accessors
-	
+
 	/**
 	 * Gets the doseUnits of this drug order
-	 * 
+	 *
 	 * @return doseUnits
 	 */
 	public Concept getDoseUnits() {
 		return this.doseUnits;
 	}
-	
+
 	/**
 	 * Sets the doseUnits of this drug order
-	 * 
+	 *
 	 * @param doseUnits
 	 */
 	public void setDoseUnits(Concept doseUnits) {
 		this.doseUnits = doseUnits;
 	}
-	
+
 	/**
 	 * Gets the frequency
-	 * 
+	 *
 	 * @return frequency
 	 * @since 1.10 (signature changed)
 	 */
 	public OrderFrequency getFrequency() {
 		return this.frequency;
 	}
-	
+
 	/**
 	 * Sets the frequency
-	 * 
+	 *
 	 * @param frequency
 	 * @since 1.10 (signature changed)
 	 */
 	public void setFrequency(OrderFrequency frequency) {
 		this.frequency = frequency;
 	}
-	
+
 	/**
 	 * Returns true/false whether the drug is a "pro re nata" drug
-	 * 
+	 *
 	 * @return Boolean
 	 * @since 1.10
 	 */
 	public Boolean getAsNeeded() {
 		return asNeeded;
 	}
-	
+
 	/**
 	 * @param asNeeded the value to set
 	 * @since 1.10
@@ -159,25 +162,25 @@ public class DrugOrder extends Order implements java.io.Serializable {
 	public void setAsNeeded(Boolean asNeeded) {
 		this.asNeeded = asNeeded;
 	}
-	
+
 	/**
 	 * Gets the quantity
-	 * 
+	 *
 	 * @return quantity
 	 */
 	public Double getQuantity() {
 		return this.quantity;
 	}
-	
+
 	/**
 	 * Sets the quantity
-	 * 
+	 *
 	 * @param quantity
 	 */
 	public void setQuantity(Double quantity) {
 		this.quantity = quantity;
 	}
-	
+
 	/**
 	 * @since 1.10
 	 * @return concept
@@ -185,7 +188,7 @@ public class DrugOrder extends Order implements java.io.Serializable {
 	public Concept getQuantityUnits() {
 		return quantityUnits;
 	}
-	
+
 	/**
 	 * @since 1.10
 	 * @param quantityUnits
@@ -193,19 +196,19 @@ public class DrugOrder extends Order implements java.io.Serializable {
 	public void setQuantityUnits(Concept quantityUnits) {
 		this.quantityUnits = quantityUnits;
 	}
-	
+
 	/**
 	 * Gets the drug
-	 * 
+	 *
 	 * @return drug
 	 */
 	public Drug getDrug() {
 		return this.drug;
 	}
-	
+
 	/**
 	 * Sets the drug
-	 * 
+	 *
 	 * @param drug
 	 */
 	public void setDrug(Drug drug) {
@@ -214,7 +217,7 @@ public class DrugOrder extends Order implements java.io.Serializable {
 			setConcept(drug.getConcept());
 		}
 	}
-	
+
 	/**
 	 * @return the asNeededCondition
 	 * @since 1.10
@@ -222,7 +225,7 @@ public class DrugOrder extends Order implements java.io.Serializable {
 	public String getAsNeededCondition() {
 		return asNeededCondition;
 	}
-	
+
 	/**
 	 * @param asNeededCondition the asNeededCondition to set
 	 * @since 1.10
@@ -230,56 +233,56 @@ public class DrugOrder extends Order implements java.io.Serializable {
 	public void setAsNeededCondition(String asNeededCondition) {
 		this.asNeededCondition = asNeededCondition;
 	}
-	
+
 	/**
 	 * Gets the route
-	 * 
+	 *
 	 * @since 1.10
 	 */
 	public Concept getRoute() {
 		return route;
 	}
-	
+
 	/**
 	 * Sets the route
-	 * 
+	 *
 	 * @param route
 	 * @since 1.10
 	 */
 	public void setRoute(Concept route) {
 		this.route = route;
 	}
-	
+
 	public void setDose(Double dose) {
 		this.dose = dose;
 	}
-	
+
 	public Double getDose() {
 		return dose;
 	}
-	
+
 	/**
 	 * Gets the dosingType
-	 * 
+	 *
 	 * @since 1.10
 	 */
 	public Class<? extends DosingInstructions> getDosingType() {
 		return dosingType;
 	}
-	
+
 	/**
 	 * Sets the dosingType
-	 * 
+	 *
 	 * @param dosingType the dosingType to set
 	 * @since 1.10
 	 */
 	public void setDosingType(Class<? extends DosingInstructions> dosingType) {
 		this.dosingType = dosingType;
 	}
-	
+
 	/**
 	 * Gets the dosingInstructions instance
-	 * 
+	 *
 	 * @since 1.10
 	 */
 	public DosingInstructions getDosingInstructionsInstance() {
@@ -294,103 +297,103 @@ public class DrugOrder extends Order implements java.io.Serializable {
 			throw new IllegalStateException(e);
 		}
 	}
-	
+
 	/**
 	 * Gets numRefills
-	 * 
+	 *
 	 * @since 1.10
 	 */
 	public Integer getNumRefills() {
 		return numRefills;
 	}
-	
+
 	/**
 	 * Sets numRefills
-	 * 
+	 *
 	 * @param numRefills the numRefills to set
 	 * @since 1.10
 	 */
 	public void setNumRefills(Integer numRefills) {
 		this.numRefills = numRefills;
 	}
-	
+
 	/**
 	 * Sets the dosingInstructions
-	 * 
+	 *
 	 * @param dosingInstructions to set
 	 * @since 1.10
 	 */
 	public void setDosingInstructions(String dosingInstructions) {
 		this.dosingInstructions = dosingInstructions;
 	}
-	
+
 	/**
 	 * Gets the dosingInstructions
-	 * 
+	 *
 	 * @since 1.10
 	 */
 	public String getDosingInstructions() {
 		return this.dosingInstructions;
 	}
-	
+
 	/**
 	 * Gets the duration of a Drug Order
-	 * 
+	 *
 	 * @since 1.10
 	 */
 	public Integer getDuration() {
 		return duration;
 	}
-	
+
 	/**
 	 * Sets the duration of a Drug Order
-	 * 
+	 *
 	 * @param duration to set
 	 * @since 1.10
 	 */
 	public void setDuration(Integer duration) {
 		this.duration = duration;
 	}
-	
+
 	/**
 	 * Gets durationUnits of a Drug Order
-	 * 
+	 *
 	 * @since 1.10
 	 */
 	public Concept getDurationUnits() {
 		return durationUnits;
 	}
-	
+
 	/**
 	 * Sets the durationUnits of a Drug Order
-	 * 
+	 *
 	 * @param durationUnits
 	 * @since 1.10
 	 */
 	public void setDurationUnits(Concept durationUnits) {
 		this.durationUnits = durationUnits;
 	}
-	
+
 	/**
 	 * Gets the brandName
-	 * 
+	 *
 	 * @return brandName
 	 * @since 1.10
 	 */
 	public String getBrandName() {
 		return brandName;
 	}
-	
+
 	/**
 	 * Sets the brandName
-	 * 
+	 *
 	 * @since 1.10
 	 * @param brandName the brandName to set to
 	 */
 	public void setBrandName(String brandName) {
 		this.brandName = brandName;
 	}
-	
+
 	/**
 	 * @return true or false
 	 * @since 1.10
@@ -398,7 +401,7 @@ public class DrugOrder extends Order implements java.io.Serializable {
 	public Boolean getDispenseAsWritten() {
 		return dispenseAsWritten;
 	}
-	
+
 	/**
 	 * @param dispenseAsWritten
 	 * @since 1.10
@@ -406,7 +409,7 @@ public class DrugOrder extends Order implements java.io.Serializable {
 	public void setDispenseAsWritten(Boolean dispenseAsWritten) {
 		this.dispenseAsWritten = dispenseAsWritten;
 	}
-	
+
 	/**
 	 * @see org.openmrs.Order#cloneForDiscontinuing()
 	 * @should set all the relevant fields
@@ -422,13 +425,14 @@ public class DrugOrder extends Order implements java.io.Serializable {
 		newOrder.setPatient(getPatient());
 		newOrder.setDrug(getDrug());
 		newOrder.setOrderType(getOrderType());
+		newOrder.setDrugNonCoded(getDrugNonCoded());
 		return newOrder;
 	}
-	
+
 	/**
 	 * Creates a DrugOrder for revision from this order, sets the previousOrder, action field and
 	 * other drug order fields.
-	 * 
+	 *
 	 * @return the newly created order
 	 * @since 1.10
 	 * @should set all the relevant fields
@@ -438,7 +442,7 @@ public class DrugOrder extends Order implements java.io.Serializable {
 	public DrugOrder cloneForRevision() {
 		return cloneForRevisionHelper(new DrugOrder());
 	}
-	
+
 	/**
 	 * @see Order#cloneForRevisionHelper(Order)
 	 */
@@ -460,10 +464,11 @@ public class DrugOrder extends Order implements java.io.Serializable {
 		target.setNumRefills(getNumRefills());
 		target.setBrandName(getBrandName());
 		target.setDispenseAsWritten(getDispenseAsWritten());
-		
+		target.setDrugNonCoded(getDrugNonCoded());
+
 		return target;
 	}
-	
+
 	/**
 	 * Sets autoExpireDate based on duration.
 	 *
@@ -477,27 +482,27 @@ public class DrugOrder extends Order implements java.io.Serializable {
 			setAutoExpireDate(getDosingInstructionsInstance().getAutoExpireDate(this));
 		}
 	}
-	
+
 	public String toString() {
 		String prefix = DISCONTINUE == getAction() ? "DC " : "";
 		return prefix + "DrugOrder(" + getDose() + getDoseUnits() + " of "
-		        + (getDrug() != null ? getDrug().getName() : "[no drug]") + " from " + getDateActivated() + " to "
+		        + (getDrugNonCoded() != null ? getDrugNonCoded() : (getDrug() != null ? getDrug().getName() : "[no drug]")) + " from " + getDateActivated() + " to "
 		        + (isDiscontinuedRightNow() ? getDateStopped() : getAutoExpireDate()) + ")";
 	}
-	
+
 	/**
 	 * Set dosing instructions to drug order
-	 * 
+	 *
 	 * @param di dosing instruction object to fetch data
 	 * @since 1.10
 	 */
 	public void setDosing(DosingInstructions di) {
 		di.setDosingInstructions(this);
 	}
-	
+
 	/**
 	 * Checks whether orderable of this drug order is same as other order
-	 * 
+	 *
 	 * @since 1.10
 	 * @param otherOrder the other order to match on
 	 * @return true if the drugs match
@@ -512,13 +517,41 @@ public class DrugOrder extends Order implements java.io.Serializable {
 	 */
 	@Override
 	public boolean hasSameOrderableAs(Order otherOrder) {
-		if (!super.hasSameOrderableAs(otherOrder)) {
-			return false;
-		}
-		if (!(otherOrder instanceof DrugOrder)) {
-			return false;
-		}
-		DrugOrder otherDrugOrder = (DrugOrder) otherOrder;
-		return OpenmrsUtil.nullSafeEquals(this.getDrug(), otherDrugOrder.getDrug());
+        if (!super.hasSameOrderableAs(otherOrder)) {
+            return false;
+        }
+        if (!(otherOrder instanceof DrugOrder)) {
+            return false;
+        }
+        DrugOrder otherDrugOrder = (DrugOrder) otherOrder;
+
+        if (isNonCodedDrug() || otherDrugOrder.isNonCodedDrug()) {
+            return OpenmrsUtil.nullSafeEqualsIgnoreCase(this.getDrugNonCoded(), otherDrugOrder.getDrugNonCoded());
+        }
+        return OpenmrsUtil.nullSafeEquals(this.getDrug(), otherDrugOrder.getDrug());
+    }
+
+	/**
+	 * @since 1.12
+	 * @return drugNonCoded
+	 */
+	public String getDrugNonCoded() {
+		return drugNonCoded;
+	}
+
+	/**
+	 * @since 1.12
+	 * sets drugNonCoded
+	 */
+	public void setDrugNonCoded(String drugNonCoded) {
+		this.drugNonCoded = drugNonCoded != null ? drugNonCoded.trim() : drugNonCoded;
+	}
+
+	/**
+	 * @since 1.12
+	 * return true if a drug is non coded
+	 */
+	public boolean isNonCodedDrug() {
+		return getDrugNonCoded() != null;
 	}
 }
--- a/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
@@ -184,13 +184,13 @@ public class OrderServiceImpl extends BaseOpenmrsService implements OrderService
 			}
 			
 			//concept should be the same as on previous order, same applies to drug for drug orders
-			boolean isDrugOrderAndHasADrug = isDrugOrder && ((DrugOrder) order).getDrug() != null;
+			boolean isDrugOrderAndHasADrug = isDrugOrder && (((DrugOrder) order).getDrug() != null || ((DrugOrder) order).isNonCodedDrug()) ;
 			if (!OpenmrsUtil.nullSafeEquals(order.getConcept(), previousOrder.getConcept())) {
 				throw new APIException("The concept of the previous order and the new one order don't match");
 			} else if (isDrugOrderAndHasADrug) {
 				DrugOrder drugOrder1 = (DrugOrder) order;
 				DrugOrder drugOrder2 = (DrugOrder) previousOrder;
-				if (!OpenmrsUtil.nullSafeEquals(drugOrder1.getDrug(), drugOrder2.getDrug())) {
+				if (!drugOrder1.hasSameOrderableAs(drugOrder2)) {
 					throw new APIException("The drug of the previous order and the new one order don't match");
 				}
 			} else if (!order.getOrderType().equals(previousOrder.getOrderType())) {
@@ -365,7 +365,7 @@ public class OrderServiceImpl extends BaseOpenmrsService implements OrderService
 	}
 	
 	private DrugOrder checkDrugOrdersForDiscontinuing(DrugOrder drugOrder1, DrugOrder drugOrder2) {
-		if (OpenmrsUtil.nullSafeEquals(drugOrder1.getDrug(), drugOrder2.getDrug())) {
+		if (drugOrder1.hasSameOrderableAs(drugOrder2)) {
 			return drugOrder2;
 		}
 		return null;
@@ -945,7 +945,13 @@ public class OrderServiceImpl extends BaseOpenmrsService implements OrderService
 	public List<Concept> getTestSpecimenSources() {
 		return getSetMembersOfConceptSetFromGP(OpenmrsConstants.GP_TEST_SPECIMEN_SOURCES_CONCEPT_UUID);
 	}
-	
+
+	@Override
+	public Concept getNonCodedDrugConcept() {
+		String conceptUuid = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_DRUG_NON_CODED_CONCEPT_UUID);
+		return Context.getConceptService().getConceptByUuid(conceptUuid);
+	}
+
 	private List<Concept> getSetMembersOfConceptSetFromGP(String globalProperty) {
 		String conceptUuid = Context.getAdministrationService().getGlobalProperty(globalProperty);
 		Concept concept = Context.getConceptService().getConceptByUuid(conceptUuid);
@@ -954,5 +960,4 @@ public class OrderServiceImpl extends BaseOpenmrsService implements OrderService
 		}
 		return Collections.emptyList();
 	}
-	
 }
--- a/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
@@ -936,7 +936,7 @@ public class OrderServiceImpl extends BaseOpenmrsService implements OrderService
 	@Override
 	public Concept getNonCodedDrugConcept() {
 		String conceptUuid = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_DRUG_ORDER_DRUG_OTHER);
-		if(!StringUtils.isEmpty(conceptUuid)){
+		if(org.apache.commons.lang.StringUtils.isNotBlank(conceptUuid)){
 			return Context.getConceptService().getConceptByUuid(conceptUuid);
 		}
 		return null;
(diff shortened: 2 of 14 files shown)


api/src/main/java/org/openmrs/DrugOrder.java
```diff
@@ -14,58 +14,60 @@
 
 /**
  * DrugOrder
- * 
+ *
```
wluyima: My assumption is that if you build the project before issuing the committing then these format changes should get weeded out

api/src/main/java/org/openmrs/DrugOrder.java
```diff
@@ -477,27 +482,27 @@ public void setAutoExpireDateBasedOnDuration() {
 			setAutoExpireDate(getDosingInstructionsInstance().getAutoExpireDate(this));
 		}
 	}
-	
+
 	public String toString() {
 		String prefix = DISCONTINUE == getAction() ? "DC " : "";
 		return prefix + "DrugOrder(" + getDose() + getDoseUnits() + " of "
-		        + (getDrug() != null ? getDrug().getName() : "[no drug]") + " from " + getDateActivated() + " to "
+		        + (getDrugNonCoded() != null ? getDrugNonCoded() : (getDrug() != null ? getDrug().getName() : "[no drug]")) + " from " + getDateActivated() + " to "
```
wluyima: I think this check should be if the drug is the drug other concept

api/src/main/java/org/openmrs/DrugOrder.java
```diff
@@ -512,13 +517,41 @@ public void setDosing(DosingInstructions di) {
...
+	public void setDrugNonCoded(String drugNonCoded) {
+		this.drugNonCoded = drugNonCoded != null ? drugNonCoded.trim() : drugNonCoded;
+	}
+
+	/**
+	 * @since 1.12
+	 * return true if a drug is non coded
+	 */
+	public boolean isNonCodedDrug() {
+		return getDrugNonCoded() != null;
```
wluyima: Same here, this should check if the drug is the drug other concept

wluyima: And in fact you should delegate to this method in the if clause in the toString() method above

api/src/main/java/org/openmrs/DrugOrder.java
```diff
@@ -512,13 +517,41 @@ public void setDosing(DosingInstructions di) {
...
-		return OpenmrsUtil.nullSafeEquals(this.getDrug(), otherDrugOrder.getDrug());
+        if (!super.hasSameOrderableAs(otherOrder)) {
+            return false;
+        }
+        if (!(otherOrder instanceof DrugOrder)) {
+            return false;
+        }
+        DrugOrder otherDrugOrder = (DrugOrder) otherOrder;
+
+        if (isNonCodedDrug() || otherDrugOrder.isNonCodedDrug()) {
```
wluyima: I would think this should check if both orders have their drug set to drug concept, if one of them isn't then they can't be deemed to be the same

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -184,13 +184,13 @@ public synchronized Order saveOrder(Order order, OrderContext orderContext) thro
 			}
 			
 			//concept should be the same as on previous order, same applies to drug for drug orders
-			boolean isDrugOrderAndHasADrug = isDrugOrder && ((DrugOrder) order).getDrug() != null;
+			boolean isDrugOrderAndHasADrug = isDrugOrder && (((DrugOrder) order).getDrug() != null || ((DrugOrder) order).isNonCodedDrug()) ;
```
wluyima: This change is not required, an order with a non coded drug value but that has a blank drug field  should be invalid

api/src/main/java/org/openmrs/DrugOrder.java
sravanthi17: How would we do that.. ? Is it okay to use OrderService here, since the getNonCodedDrugConcept is in OrderService ?

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
sravanthi17: How come nonCoded drug with drug field empty invalid..?

api/src/main/java/org/openmrs/DrugOrder.java
wluyima: I thought you added a method to OrderService to get the nonCodedDrug, use that here

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
wluyima: DrugOrder.isNonCodedDrug() will never return true if getDrug order returns null if you implement isNonCodedDrug well where it checks if the drug's concept is equals to DRUG OTHER concept, so that check would be redundant, makes sense?

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -1100,6 +1100,12 @@ public static String getOpenmrsProperty(String property) {
 	 * @since 1.11
 	 */
 	public static final Integer SEARCH_INDEX_VERSION = 3;
+
+	/**
+	 * @since 1.12
+	 * Specifies the uuid of concept which represents drug non coded
+	 */
+	public static final String GP_DRUG_NON_CODED_CONCEPT_UUID = "order.drugNonCodedConceptUuid";
```
wluyima: According to the ticket description, I thought the GP name is supposed to be drugOrder.drugNonCoded, i actually think drugOrder.drugNonCodedConceptUuid

wluyima: Or even drugOrder.NonConceptUuid

api/src/main/java/org/openmrs/validator/DrugOrderValidator.java
```diff
@@ -108,16 +109,30 @@ public void validate(Object obj, Errors errors) {
...
+	private void validateForRequireDrug(Errors errors, DrugOrder order) {
 		//Reject if global property is set to specify a formulation for drug order
 		boolean requireDrug = Context.getAdministrationService().getGlobalPropertyValue(
-		    OpenmrsConstants.GLOBAL_PROPERTY_DRUG_ORDER_REQUIRE_DRUG, false);
-		if (requireDrug) {
-			ValidationUtils.rejectIfEmpty(errors, "drug", "DrugOrder.error.drugIsRequired");
+				OpenmrsConstants.GLOBAL_PROPERTY_DRUG_ORDER_REQUIRE_DRUG, false);
+		OrderService orderService = Context.getOrderService();
+
+		if (order.isNonCodedDrug()) {
```
wluyima: You will need to change this code when you edit the code in DrugOrder.isNonCodedDrug

api/src/main/java/org/openmrs/validator/DrugOrderValidator.java
```diff
@@ -108,16 +109,30 @@ public void validate(Object obj, Errors errors) {
...
+			if (OpenmrsUtil.nullSafeEquals(orderService.getNonCodedDrugConcept(), order.getConcept())) {
+				ValidationUtils.rejectIfEmpty(errors, "drugNonCoded", "DrugOrder.error.drugNonCodedIsRequired");
+			}
+			if (order.getDrug() != null) {
+				errors.rejectValue("drug", "DrugOrder.error.onlyOneOfDrugOrNonCodedShouldBeSet");
```
wluyima: This doesn't seem correct, you need to fetch all the active drug order and of the patient, loop through them to find all drug other orders and compare their none coded value to this drug's none coded value

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,100 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
...
+	 * @see DrugOrder#hasSameOrderableAs(Order)
+	 */
+	@Test
+	public void hasSameOrderableAs_shouldReturnTrueIfTheConceptsMatchAndHaveSameDrugNonCoded() throws Exception {
+		DrugOrder order = new DrugOrder();
+		Concept concept = new Concept();
+		Drug drug1 = new Drug();
+		drug1.setConcept(concept);
+		order.setDrug(drug1);
+		order.setDrugNonCoded("Chrocine");
```
wluyima: This clarifies what i was saying above, do you realize that you always have to set drug for a non coded order? Meaning isNoneCodedDrug technically can never return true if itself drug is null

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,100 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
...
+		assertTrue(order.hasSameOrderableAs(otherOrder));
+	}
+
+	/**
+	 * @verifies return false if the concepts match and either of drug non coded is null
+	 * @see DrugOrder#hasSameOrderableAs(Order)
+	 */
+	@Test
+	public void hasSameOrderableAs_shouldReturnFalseIfTheConceptsMatchAndEitherOfDrugNonCodedIsNull() throws Exception {
+		DrugOrder order = new DrugOrder();
```
wluyima: This test would be redundant if your validator is doing its job correctly, If both orders have the same concept, and one of them only has the none coded value, this can can imply 2 things:
- Their concept is the drug other concept and one of them is missing the none coded value which already makes it invalid

OR
- The semantics are a bit off, because their concept is not the drug other concept and one of them is has the none coded value which already makes it also invalid because the validator should have already rejected it anyways

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,100 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
...
+	/**
+	 * @verifies return false if the concepts match and both of drug non coded is null
+	 * @see DrugOrder#hasSameOrderableAs(Order)
+	 */
+	@Test
+	public void hasSameOrderableAs_shouldReturnFalseIfTheConceptsMatchAndBothDrugNonCodedIsNull() throws Exception {
+		DrugOrder order = new DrugOrder();
+		Concept concept = new Concept();
+		Drug drug1 = new Drug();
+		drug1.setConcept(concept);
```
wluyima: Do you realize you have not set the drug on the order

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,100 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
...
+
+		assertFalse(order.hasSameOrderableAs(otherOrder));
+	}
+
+	/**
+	 * @verifies return false if the concepts match and both of drug non coded is null
+	 * @see DrugOrder#hasSameOrderableAs(Order)
+	 */
+	@Test
+	public void hasSameOrderableAs_shouldReturnFalseIfTheConceptsMatchAndBothDrugNonCodedIsNull() throws Exception {
```
wluyima: I believe this test already exist since this is what the old code was checking

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -461,7 +461,7 @@ public void saveGlobalProperties_shouldAssignUuidToAllNewProperties() throws Exc
 	@Verifies(value = "should return all global properties in the database", method = "getAllGlobalProperties()")
 	public void getAllGlobalProperties_shouldReturnAllGlobalPropertiesInTheDatabase() throws Exception {
 		executeDataSet(ADMIN_INITIAL_DATA_XML);
-		Assert.assertEquals(19, Context.getAdministrationService().getAllGlobalProperties().size());
+		Assert.assertEquals(20, Context.getAdministrationService().getAllGlobalProperties().size());
```
wluyima: It is good to add such test data to new test data file instead of the existing that you avoid touching these existing tests

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -2969,4 +2969,101 @@ public void saveOrder_shouldPassIfAnKnownDrugOrderForTheSameDrugFormulationSpeci
...
+		assertEquals(nonCodedDrugConcept.getName().getName(), "DRUG OTHER");
+		
+	}
+	
+	/**
+	 * @verifies pass if an active drug order for the same coded concept, care setting and different drug non coded exists
+	 * @see OrderService#saveOrder(Order, OrderContext)
+	 */
+	@Test
+	public void saveOrder_shouldPassForANewNonCodedDrugOrderIfAnotherActiveNonCodedDrugOrderAlreadyExists()
```
wluyima: I think these tests are unnecessary because you already tested the same code in the tests for isSameOrderable

api/src/test/resources/org/openmrs/include/standardTestDataset.xml
```diff
@@ -81,6 +81,7 @@
   <concept concept_id="792" retired="false" datatype_id="4" class_id="3" is_set="false" creator="1" date_created="2004-04-08 00:00:00.0" version="" changed_by="1" date_changed="2008-08-15 14:05:56.0" uuid="d144d24f-6913-4b63-9660-a9108c2bebef"/>
   <concept concept_id="5089" retired="false" datatype_id="1" class_id="1" is_set="false" creator="1" date_created="2004-08-12 00:00:00.0" version="" changed_by="1" date_changed="2005-02-25 11:43:43.0" uuid="c607c80f-1ea9-4da3-bb88-6276ce8868dd"/>
   <concept concept_id="5497" retired="false" datatype_id="1" class_id="1" is_set="false" creator="1" date_created="2004-08-12 00:00:00.0" version="0.1" changed_by="1" date_changed="2006-06-09 14:57:33.0" uuid="a09ab2c5-878e-4905-b25d-5784167d0216"/>
+  <concept concept_id="5584" retired="false" datatype_id="4" class_id="3" is_set="false" creator="1" date_created="2004-08-12 00:00:00.0" version="0.1" changed_by="1" date_changed="2006-08-12 00:00:00.0" uuid="6584d845-7584-11e4-945f-0858400c9s89"/>
```
wluyima: If you added this concept and the GO to OrderServiceTest-nonCodedDrugs.xml above instead of here, then you wouldn't need to touch existing tests in ConceptServiceTest and AdministrationServiceTest

api/src/main/java/org/openmrs/DrugOrder.java
sravanthi17: Is it a good idea to use service in the DrugOrder model..? The reason we proceeded with just nonCoded value is because the concept validations are already in DrugOrderValidator.

lluismf: In my opinion domain objects should not access service or DAO classes, but it's just from an architectural point of view.

sravanthi17: @lluismf.. I agree with you.. 
In this case, can you think of any other way to solve this validation problem ?

lluismf: It depends, from where is the method being called? 
Also toString should be used for logging or debugging purposes, not business logic. It's not even localized.

wluyima:
@sravanthi17 the ticket description was updated, do you mind making sure your code addresses everything

api/src/main/java/org/openmrs/DrugOrder.java
wluyima: I agree it is not good to call the service layer from domain objects but sometimes we have to e.g we already do this for Obs for some good reasons. Since we updated the design of this and said DrugOrder.drug and DrugOrder.drugNonCoded should be mutually exclusive then i guess you can ignore my comment

wluyima: But it would still be nice to delegate to isNonCodedDrug() below when making the check.

api/src/main/java/org/openmrs/DrugOrder.java
wluyima: You can ignore this comment

wluyima: In fact theu should check if both orders have their concepts are non coded, if one of them isn't then they can't be deemed to be the same

shruthidipali:
@wluyima I think @sravanthi17 has made all the required changes. Please could you review and give us your feedback?

api/src/main/java/org/openmrs/DrugOrder.java
```diff
@@ -14,58 +14,60 @@
 
 /**
  * DrugOrder
- * 
+ *
  * @version 1.0
  */
 public class DrugOrder extends Order implements java.io.Serializable {
-
```
wluyima: You need build the project before committing then these format changes should get weeded out

api/src/main/java/org/openmrs/DrugOrder.java
wluyima: Delegate to isNonCodedDrug() when making the check.

api/src/main/java/org/openmrs/DrugOrder.java
wluyima: This should be checking if both orders have a none coded drug

api/src/main/java/org/openmrs/DrugOrder.java
wluyima: I think this should check that also the non coded value is not blank, we normally user StringUtils.isNotBlank

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -184,13 +184,13 @@ public synchronized Order saveOrder(Order order, OrderContext orderContext) thro
...
 			if (!OpenmrsUtil.nullSafeEquals(order.getConcept(), previousOrder.getConcept())) {
 				throw new APIException("The concept of the previous order and the new one order don't match");
 			} else if (isDrugOrderAndHasADrug) {
 				DrugOrder drugOrder1 = (DrugOrder) order;
 				DrugOrder drugOrder2 = (DrugOrder) previousOrder;
-				if (!OpenmrsUtil.nullSafeEquals(drugOrder1.getDrug(), drugOrder2.getDrug())) {
+				if (!drugOrder1.hasSameOrderableAs(drugOrder2)) {
```
wluyima: You should actually replace all this logic, lines 187 to 195 should be replace with using hasSameOrderable, and exception message should say "The orderable of the previous order and the new one order don't match"

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -365,7 +365,7 @@ private void discontinueExistingOrdersIfNecessary(Order order) {
 	}
 	
 	private DrugOrder checkDrugOrdersForDiscontinuing(DrugOrder drugOrder1, DrugOrder drugOrder2) {
```
wluyima: Get rid of this method instead and where it is being used change the code to call hasSameOrderableAs()

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -945,7 +945,13 @@ public OrderType getOrderTypeByConcept(Concept concept) {
 	public List<Concept> getTestSpecimenSources() {
 		return getSetMembersOfConceptSetFromGP(OpenmrsConstants.GP_TEST_SPECIMEN_SOURCES_CONCEPT_UUID);
 	}
-	
+
+	@Override
+	public Concept getNonCodedDrugConcept() {
+		String conceptUuid = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_DRUG_NON_CODED_CONCEPT_UUID);
```
wluyima: You need to check that conceptUuid is not null and blank

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -1100,6 +1100,12 @@ public static String getOpenmrsProperty(String property) {
 	 * @since 1.11
 	 */
 	public static final Integer SEARCH_INDEX_VERSION = 3;
+
+	/**
+	 * @since 1.12
+	 * Specifies the uuid of concept which represents drug non coded
```
wluyima: uuid of concept -> uuid of the concept

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -1604,7 +1610,9 @@ public static String getOpenmrsProperty(String property) {
 		                GLOBAL_PROPERTY_PERSON_ATTRIBUTE_SEARCH_MATCH_MODE,
 		                GLOBAL_PROPERTY_PERSON_ATTRIBUTE_SEARCH_MATCH_EXACT,
 		                "Specifies how person attributes are matched while searching person. Valid values are 'ANYWHERE' or 'EXACT'. Defaults to exact if missing or invalid value is present."));
-		
+
+		props
+				.add(new GlobalProperty(GP_DRUG_ORDER_DRUG_OTHER, "", "Specifies the uuid of concept which represents drug other non coded"));
```
wluyima: Same here should be 'the concept'

api/src/main/java/org/openmrs/validator/DrugOrderValidator.java
```diff
@@ -108,16 +109,30 @@ public void validate(Object obj, Errors errors) {
...
+
+	private void validateForRequireDrug(Errors errors, DrugOrder order) {
 		//Reject if global property is set to specify a formulation for drug order
 		boolean requireDrug = Context.getAdministrationService().getGlobalPropertyValue(
-		    OpenmrsConstants.GLOBAL_PROPERTY_DRUG_ORDER_REQUIRE_DRUG, false);
-		if (requireDrug) {
-			ValidationUtils.rejectIfEmpty(errors, "drug", "DrugOrder.error.drugIsRequired");
+				OpenmrsConstants.GLOBAL_PROPERTY_DRUG_ORDER_REQUIRE_DRUG, false);
+		OrderService orderService = Context.getOrderService();
+
```
wluyima: You need to first check if drug is required, it is then drug or drugNoneCoded is required. Also if a drugNoneCoded is specified, you need to check that the concept is the DRUG OTHER concept

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,100 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
...
+		DrugOrder order = new DrugOrder();
+		Concept concept = new Concept();
+		Drug drug1 = new Drug();
```
wluyima: Get rid of drug1 since it never gets used anywhere

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,100 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
...
+		drug1.setConcept(concept);
+
+		DrugOrder otherOrder = new DrugOrder();
+		otherOrder.setConcept(concept);
+
+		assertFalse(order.hasSameOrderableAs(otherOrder));
```
wluyima: If both drugNonCoded fields are null, then hasSameOrderableAs should return true if the concepts match

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
wluyima: Did you see this comment?

api/src/test/resources/org/openmrs/include/standardTestDataset.xml
wluyima: Did you see this comment?

api/src/test/java/org/openmrs/validator/DrugOrderValidatorTest.java
```diff
@@ -649,4 +650,118 @@ public void validate_shouldFailIfDurationUnitsHasNoMappingToSNOMEDCTSource() thr
...
+		order.setOrderer(Context.getProviderService().getProvider(1));
+		order.setDateActivated(encounter.getEncounterDatetime());
+		order.setOrderType(orderType);
+		order.setDosingType(FreeTextDosingInstructions.class);
+		order.setInstructions("None");
+		order.setDosingInstructions("Test Instruction");
+		Errors errors = new BindException(order, "order");
+		new DrugOrderValidator().validate(order, errors);
+		Assert.assertTrue(errors.hasFieldErrors());
+		assertEquals("DrugOrder.error.cannotSetBothDrugAndNonCodedDrug", errors.getFieldError("concept").getCode());
```
wluyima: The test method says should fail if drug none coded is not set but the error message says "DrugOrder.error.cannotSetBothDrugAndNonCodedDrug", they seem contradictory

api/src/test/java/org/openmrs/validator/DrugOrderValidatorTest.java
```diff
@@ -649,4 +650,118 @@ public void validate_shouldFailIfDurationUnitsHasNoMappingToSNOMEDCTSource() thr
...
+	}
+
+	@Test
+	@Verifies(value = "should pass validation if drug non coded is set for non coded drug order", method = "validate(Object,Errors)")
+	public void saveOrder_shouldPassDrugOrderWithADrugNonCodedWhenDrugOrderIsNonCoded() throws Exception {
+		Patient patient = Context.getPatientService().getPatient(7);
+		CareSetting careSetting = Context.getOrderService().getCareSetting(2);
+		OrderType orderType = Context.getOrderService().getOrderTypeByName("Drug order");
+		assertNull(Context.getAdministrationService().getGlobalPropertyObject(
```
wluyima: Should not be null

api/src/test/java/org/openmrs/validator/DrugOrderValidatorTest.java
```diff
@@ -649,4 +650,118 @@ public void validate_shouldFailIfDurationUnitsHasNoMappingToSNOMEDCTSource() thr
...
 	}
+	
+	@Test
+	@Verifies(value = "should fail validation if drug non coded is not set for non coded drug order", method = "validate(Object,Errors)")
+	public void saveOrder_shouldFailDrugOrderWithoutADrugNonCodedWhenDrugOrderIsNonCoded() throws Exception {
+		Patient patient = Context.getPatientService().getPatient(7);
+		CareSetting careSetting = Context.getOrderService().getCareSetting(2);
+		OrderType orderType = Context.getOrderService().getOrderTypeByName("Drug order");
+
+		assertNull(Context.getAdministrationService().getGlobalPropertyObject(
```
wluyima: Should not be null

wluyima: Is this required really?

api/src/test/java/org/openmrs/validator/DrugOrderValidatorTest.java
```diff
@@ -649,4 +650,118 @@ public void validate_shouldFailIfDurationUnitsHasNoMappingToSNOMEDCTSource() thr
...
+		order.setInstructions("None");
+		order.setDosingInstructions("Test Instruction");
+		Errors errors = new BindException(order, "order");
+		new DrugOrderValidator().validate(order, errors);
+		Assert.assertTrue(errors.hasFieldErrors());
+		assertEquals("DrugOrder.error.onlyOneOfDrugOrNonCodedShouldBeSet", errors.getFieldError("drug").getCode());
+	}
+
+	@Test
+	@Verifies(value = "should fail validation if drug non coded is set for coded drug order", method = "validate(Object,Errors)")
```
wluyima: This tests duplicates the one before it, i think you only need one of them

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -2969,4 +2969,101 @@ public void saveOrder_shouldPassIfAnKnownDrugOrderForTheSameDrugFormulationSpeci
 		orderService.saveOrder(order, orderContext);
 		assertNotNull(orderService.getOrder(order.getOrderId()));
 	}
+
```
wluyima: I think you are missing some tests, see the acceptance criteria for the ticket and make sure all those scenario have been or already exist

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
sravanthi17: since order need not be drug order always line 188 should still remain the same..
(or) we should call order.hasSameOrderableAs() and drugOrder.hasSameOrderableAs()

api/src/test/java/org/openmrs/validator/DrugOrderValidatorTest.java
sravanthi17: If concept is drugOther, String drugNonCoded is compulsory.. this test is for that validation

api/src/test/java/org/openmrs/validator/DrugOrderValidatorTest.java
sravanthi17: what should not be null ?

api/src/main/java/org/openmrs/validator/DrugOrderValidator.java
sravanthi17: Are you saying if drug required is false, drug non coded is compulsory or when drug is required  drug other cannot be non coded ? Also on the second point, if a drugNonCoded is specified the concept can be anything and vice versa is not possible ie..,  if a concept is drugOther then drugNonCoded is compulsory

api/src/test/resources/org/openmrs/include/standardTestDataset.xml
sravanthi17: Is it okay to executeDataSet in DrugOrderValidatorTest ?

api/src/main/java/org/openmrs/DrugOrder.java
sravanthi17: This condition works fine for both non coded drugs as well.. could you please with an example scenario you are referring to

sravanthi17:
@wluyima .. Made the changes you suggested.. Please review it and give us your feedback

api/src/main/java/org/openmrs/DrugOrder.java
sravanthi17: Sure.. We have built the project and pushed.

api/src/main/java/org/openmrs/validator/DrugOrderValidator.java
sravanthi17: We are first checking if drug is required.. 
Based on concept DRUG OTHER and drug further validations will be done..

api/src/test/java/org/openmrs/DrugOrderTest.java
sravanthi17: Agreed.. Made corresponding changes

api/src/test/java/org/openmrs/api/OrderServiceTest.java
sravanthi17: As per acceptance criteria in OrderService.. Added tests  for following scenarios.
1) Adding new non coded drug - pass condition
2) Revising existing non coded drug - pass condition
3) discontinuing existing non coded drug - pass condition
4) Adding same non coded drug as existing - fail condition
5) Revising non existing non coded drug - fail condition
6) discontinuing non existing non coded drug - fail condition

Let us know if we miss out any acceptance criteria.

api/src/main/java/org/openmrs/validator/DrugOrderValidator.java
sravanthi17: As per Acceptance criteria 6, validation for drug in line 96 and 97 will reject if drug is empty.. Should we modify this ?

wluyima:
I couldn't find your comments on the pull request to find the exact pieces of code where you added them, so i can't really respond to them

sravanthi17:
@wluyima .. Added comments .. please review the commit and let us know your feedback...
https://github.com/sravanthi17/openmrs-core/commit/8d527a25012a5832411088536d957b8a5b6e2e10
https://github.com/sravanthi17/openmrs-core/commit/43c5c43aed77fab1aefa7ddff3bc4992bff04bf3

api/src/main/java/org/openmrs/DrugOrder.java
wluyima: For example, DrugOrderA is nonCoded but with a blank nonCodedValue, DrugOrderB is a regular drug order meaning it has a blank nonCodedValue too, it means we will get into this if clause and the comparison will return true when actually these are not the same based on our acceptance criteria on the ticket but if you use AND instead of OR then the if clause would fail and the logic stays correct. Of course the argument would be that DrugOrderA is technically invalid and should have already been caught by the drug order validator but the validators can be compromised in some cases depending on where you call the service method from which i won't elaborate how, in summary to be on a safe side, i would use AND instead of OR, makes sense?

wluyima: In theory, they should both be nonCoded drug orders

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -945,7 +932,16 @@ public OrderType getOrderTypeByConcept(Concept concept) {
 	public List<Concept> getTestSpecimenSources() {
 		return getSetMembersOfConceptSetFromGP(OpenmrsConstants.GP_TEST_SPECIMEN_SOURCES_CONCEPT_UUID);
 	}
-	
+
+	@Override
+	public Concept getNonCodedDrugConcept() {
+		String conceptUuid = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_DRUG_ORDER_DRUG_OTHER);
+		if(!StringUtils.isEmpty(conceptUuid)){
```
wluyima: isEmpty returns true for white space characters

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,95 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
 		
 		assertEquals(expectedAutoExpireDate, drugOrder.getAutoExpireDate());
 	}
+
+	/**
+	 * @verifies return true if the drugs match and drug non coded match
+	 * @see DrugOrder#hasSameOrderableAs(Order)
+	 */
+	@Test
+	public void hasSameOrderableAs_shouldReturnTrueIfTheConceptsMatchAndHaveSameDrugNonCoded() throws Exception {
```
wluyima: Actually this should be if concepts are both the DRUG OTHER  and and not just any concept

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,95 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
...
+
+		assertTrue(order.hasSameOrderableAs(otherOrder));
+	}
+
+	/**
+	 * @verifies return false if the concepts match and drug non coded do not match
+	 * @see DrugOrder#hasSameOrderableAs(Order)
+	 */
+	@Test
+	public void hasSameOrderableAs_shouldReturnFalseIfTheConceptsMatchAndHaveDifferentDrugNonCoded() throws Exception {
```
wluyima: Same here, should say if concept is DRUG OTHER and not any concept

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,95 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
...
+		DrugOrder order = new DrugOrder();
+		Concept concept = new Concept();
+		order.setDrugNonCoded("Chrocine");
```
wluyima: You forgot to set concept on this first order

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,95 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
...
+
+		assertFalse(order.hasSameOrderableAs(otherOrder));
+	}
+
+	/**
+	 * @verifies return true if the concepts match and drug non coded match trimming spaces
+	 * @see DrugOrder#hasSameOrderableAs(Order)
+	 */
+	@Test
+	public void hasSameOrderableAs_shouldReturnTrueIfTheConceptsMatchAndHaveSameDrugNonCodedTrimmingSpaces() throws Exception {
```
wluyima: Same here, should say if concept is DRUG OTHER and not any concept

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,95 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
...
+	 * @verifies return true if the drugs match and drug non coded match
+	 * @see DrugOrder#hasSameOrderableAs(Order)
+	 */
+	@Test
+	public void hasSameOrderableAs_shouldReturnTrueIfTheConceptsMatchAndHaveSameDrugNonCoded() throws Exception {
+		DrugOrder order = new DrugOrder();
+		Concept concept = new Concept();
+		Drug drug1 = new Drug();
+		drug1.setConcept(concept);
+		order.setDrug(drug1);
```
wluyima: I thought drug should be null for a nonCoded order, but i see you are setting it here

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,95 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
...
+	 * @verifies return true if the concepts match and drug non coded match trimming spaces
+	 * @see DrugOrder#hasSameOrderableAs(Order)
+	 */
+	@Test
+	public void hasSameOrderableAs_shouldReturnTrueIfTheConceptsMatchAndHaveSameDrugNonCodedTrimmingSpaces() throws Exception {
+		DrugOrder order = new DrugOrder();
+		Concept concept = new Concept();
+		Drug drug1 = new Drug();
+		drug1.setConcept(concept);
+		order.setDrug(drug1);
```
wluyima: Same here, I thought drug should be null for a nonCoded order, but i see you are setting it here

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,95 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
...
+
+		assertTrue(order.hasSameOrderableAs(otherOrder));
+	}
+
+	/**
+	 * @verifies return false if the concepts match and either of drug non coded is null
+	 * @see DrugOrder#hasSameOrderableAs(Order)
+	 */
+	@Test
+	public void hasSameOrderableAs_shouldReturnFalseIfTheConceptsMatchAndEitherOfDrugNonCodedIsNull() throws Exception {
```
wluyima: Same here, should say if concept is DRUG OTHER and not any concept

api/src/test/java/org/openmrs/DrugOrderTest.java
```diff
@@ -314,4 +314,95 @@ public void setAutoExpireDateBasedOnDuration_shouldNotCalculateIfAutoExpireDateA
...
+		DrugOrder order = new DrugOrder();
```
wluyima: Forgot to set set concept on this order

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -2969,4 +2969,178 @@ public void saveOrder_shouldPassIfAnKnownDrugOrderForTheSameDrugFormulationSpeci
...
+	/**
+	 * @verifies pass if an active drug order for the same coded concept, care setting and different drug non coded exists
+	 * @see OrderService#saveOrder(Order, OrderContext)
+	 */
+	@Test
+	public void saveOrder_shouldPassForANewNonCodedDrugOrderIfAnotherActiveNonCodedDrugOrderAlreadyExists()
+	        throws Exception {
+		executeDataSet("org/openmrs/api/include/OrderServiceTest-nonCodedDrugs.xml");
+		final Patient patient = patientService.getPatient(2);
+		final Concept nonCodedConcept = conceptService.getConcept(5584);
```
wluyima: It is better to call getNonCodedConcept instead of getConcept

api/src/test/java/org/openmrs/api/OrderServiceTest.java
wluyima: I think this should say should fail, since this is a duplicate

wluyima: Unless you meant to say different care setting

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -2969,4 +2969,178 @@ public void saveOrder_shouldPassIfAnKnownDrugOrderForTheSameDrugFormulationSpeci
...
+		//sanity check that we have an active order for the same concept
+		DrugOrder duplicateOrder = (DrugOrder) orderService.getOrder(584);
+		assertTrue(duplicateOrder.isActive());
+		assertEquals(nonCodedConcept, duplicateOrder.getConcept());
+		
+		DrugOrder drugOrder = new DrugOrder();
+		drugOrder.setPatient(patient);
```
wluyima: I think the reason this test passes could be because the new order is in a different care setting, the best way to write these tests is to copy most of the field values from the existing to ensure they are always the same

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -2969,4 +2969,178 @@ public void saveOrder_shouldPassIfAnKnownDrugOrderForTheSameDrugFormulationSpeci
...
+		final Patient patient = patientService.getPatient(2);
+		final Concept nonCodedConcept = conceptService.getConcept(5584);
+		//sanity check that we have an active order for the same concept
+		DrugOrder duplicateOrder = (DrugOrder) orderService.getOrder(584);
+		assertTrue(duplicateOrder.isActive());
+		assertEquals(nonCodedConcept, duplicateOrder.getConcept());
+
+		DrugOrder drugOrder = new DrugOrder();
+		drugOrder.setPatient(patient);
+		drugOrder.setCareSetting(orderService.getCareSetting(1));
```
wluyima: You should be copying care setting and other fields values from the existing to guarantee they are always the same

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -2969,4 +2969,178 @@ public void saveOrder_shouldPassIfAnKnownDrugOrderForTheSameDrugFormulationSpeci
...
+		assertNotNull(orderService.getOrder(saveOrder.getOrderId()));
+	}
+
+	@Test
+	@Verifies(value = "should fail discontinue previousNonCodedDrugOrder if the orderable of the previous order and the new one order don't match ", method = "saveOrder(Order)")
+	public void saveOrder_shouldFailDiscontinueNonCodedDrugOrderIfOrderableOfPreviousAndNewOrderDontMatch() throws Exception {
+		executeDataSet("org/openmrs/api/include/OrderServiceTest-nonCodedDrugs.xml");
+		DrugOrder previousOrder = (DrugOrder) orderService.getOrder(584);
+		DrugOrder drugOrder = previousOrder.cloneForDiscontinuing();
+		drugOrder.setDrugNonCoded("non coded drug paracetemol");
```
wluyima: Copy this value from the previous and ensure the previous' value isn't null

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -2969,4 +2969,178 @@ public void saveOrder_shouldPassIfAnKnownDrugOrderForTheSameDrugFormulationSpeci
...
+	}
+
+
+	@Test
+	@Verifies(value = "should fail revising previousNonCodedDrugOrder if the orderable of the previous order and the new one order don't match ", method = "saveOrder(Order)")
+	public void saveOrder_shouldFailIfDrugNonCodedInPreviousDrugOrderDoesNotMatchThatOfTheRevisedDrugOrder() throws Exception {
+		executeDataSet("org/openmrs/api/include/OrderServiceTest-nonCodedDrugs.xml");
+		DrugOrder previousOrder = (DrugOrder)orderService.getOrder(584);
+		DrugOrder order = previousOrder.cloneForRevision();
+		String drugNonCodedParacetemol = "non coded paracetemol";
```
wluyima: Would be nice to have an asssertion that this isn't equal to the previous' non coded value

api/src/test/java/org/openmrs/DrugOrderTest.java
sravanthi17: The reason for not depending on "DRUG OTHER" concept is because concept can be anything for non coded drug as per the updated ticket..

api/src/test/java/org/openmrs/api/OrderServiceTest.java
sravanthi17: what should be copied?

api/src/test/java/org/openmrs/DrugOrderTest.java
wluyima: I think bullet 3 in the acceptance criteria is what you are referring to, it depends on how you read it, may be it should have been phrased in a better way. In theory, a drug order where the concept is not DRUG OTHER and drugNonCoded is specified is invalid. One should ONLY set drugNonCoded for a drug order where concept is DRUG OTHER

api/src/main/java/org/openmrs/validator/DrugOrderValidator.java
```diff
@@ -108,16 +109,33 @@ public void validate(Object obj, Errors errors) {
...
+
```
wluyima: There is one missing check in this validator, if DrugOrder.drugNonCoded is specified and concept is not the same as DRUG OTHER, that is invalid

api/src/test/java/org/openmrs/DrugOrderTest.java
sravanthi17: But here in this https://talk.openmrs.org/t/can-we-implement-freetext-drugorders/3045/30 discussion, it was mentioned by Darius that "Basically we should also allow the case where DrugOrder.concept = Amoxicillin and DrugOrder.drugNonCoded = "Amoxicillin 123 mg special tablet" (or "Foocillin brand 500mg tablet")."   so I was referring to this scenario...

wluyima: That's right, but the acceptance criteria is more correct, the concept should be DRUG OTHER or of class drug

api/src/main/java/org/openmrs/validator/DrugOrderValidator.java
sravanthi17: The reason for not adding that check is because, for a non coded drug the concept need not be DRUG OTHER. This is what is mentioned in the ticket comment right ? 

----> Client is allowed to create a DrugOrder with concept=Ampicillin (e.g. a known concept) and drugNonCoded="Ampicillin 17mg tablet"
the ticket was originally written to say that you can only specify drugNonCoded if concept=DRUG OTHER, but we changed our minds about this

wluyima: What is in the ticket's acceptance criteria is what i think you should implement, the concept needs to be DRUG OTHER or of class drug, we have a design call in about 35min, i think we can confirm this with the others

api/src/test/java/org/openmrs/api/OrderServiceTest.java
wluyima: Did you see this comment?

api/src/test/java/org/openmrs/DrugOrderTest.java
wluyima: Did you see this comment?

api/src/test/java/org/openmrs/DrugOrderTest.java
wluyima: This test is still incorrect, the test method says if the concept match but you are not setting the concept on the first drug order

api/src/main/java/org/openmrs/DrugOrder.java
sravanthi17: Agreed to your point. But it will be caught in validator.
Lets say, we changed to use AND instead of OR. Now below scenario would fail..

Drug Orders with drugs null and one drug non coded value is null.. 
Since the if condition fails, it executes the drug check where it returns true and in reality they both are different orders.

wluyima: But other code will fail before the validator is called.
If both drug orders have null drug and null non coded values, i think it is fair to think they are the same and it is okay for the condition to fail.

wluyima: Anyways, it everything is passing we can leave it this way hoping the validator to catch it

wluyima:
Merged

PR #1589, opened 2015-10-06
TRUNK-4769 : Add updateSearchIndexAsync method to Context

aj-jaswanth:
Added new method called updateSearchIndexAsync to Contex to be able to rebuild the search index asynchronously.


dkayiwa:
@rkorytkowski what do you think about this?

PR #1594, opened 2015-10-08
Changeset Preconditions to enable migration from 1.6.3 to 1.11.x

ssmusoke:
Added pre-conditions to changesets with ID 20111222-1659 and 201306141103-TRUNK-3884 which make changes that are available in a migration from 1.6.3

### DIFF
--- a/api/src/main/resources/liquibase-update-to-latest.xml
+++ b/api/src/main/resources/liquibase-update-to-latest.xml
@@ -6337,6 +6337,11 @@
 	</changeSet>
 	
 	<changeSet id="20111222-1659" author="djazayeri">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<tableExists tableName="clob_datatype_storage"/>
+			</not>
+		</preConditions>
 		<comment>Create clob_datatype_storage table</comment>
 		<createTable tableName="clob_datatype_storage">
 			<column name="id" type="int" autoIncrement="true">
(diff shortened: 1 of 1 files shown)


dkayiwa:
I have not yet deeply looked into this. But i suspect that this could be hiding the actual cause of failure. :)
That being put aside, have you taken a look at? https://wiki.openmrs.org/display/docs/How+to+Modify+a+Bad+Changeset

ssmusoke:
Actually the changesets work fine. The problem is that the changes being implemented by the changesets have already occurred (not sure how or why since its a migration) so running the changesets fails.

dkayiwa:
Do you already have a ticket for this pull request where this discussion can happen?
http://go.openmrs.org/pull-request-tips

ssmusoke:
Apologies was not sure of the process have created a ticket here https://issues.openmrs.org/browse/TRUNK-4774.

However I have only made the change on 1.11.x branch and not on master - do I still have to go ahead and move the change to the master branch which will mean deleting this PR?

api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -6337,6 +6337,11 @@
 	</changeSet>
 	
 	<changeSet id="20111222-1659" author="djazayeri">
+		<preConditions onFail="MARK_RAN">
```
wluyima: It is good to have preconditions, but we try to avoid editing changesets, especially one like this where we don't expect one to have the table since it is new, have you tried to investigate why it already exists because that's what you might want to fix instead of this.

dkayiwa:
Closing because of the above comments.

PR #1596, opened 2015-10-13
TRUNK-4754 | Updating lucene index while saving drug

vinkesh:


api/src/test/java/org/openmrs/api/ConceptServiceTest.java
```diff
@@ -399,7 +399,26 @@ public void saveConceptSource_shouldNotSetDateCreatedIfOneIsSuppliedAlready() th
...
+		drug.setName("ASPIRIN BRANDED");
+		drug.setConcept(conceptService.getConceptByName("ASPIRIN"));
+		drug.setStrength("500mg");
+
+		Drug saveDrug = conceptService.saveDrug(drug);
+		assertNotNull(saveDrug);
+
+		List<Drug> drugs = conceptService.getDrugs("ASPIRIN BRANDED");
+		assertEquals(1, drugs.size());
+		assertEquals("ASPIRIN BRANDED", drugs.get(0).getName());
```
rkorytkowski: This test does not prove that updating the index automatically does not work, because the update only takes place when a transaction is committed, which never happens here (the transaction is rolled back by the end of the test). The test would have to be @NotTransactional to prove a bug. Can you try annotating the test with @NotTransactional and running it without your fix in line 301 in saveConcept?

vinkesh: @rkorytkowski : Thanks for the pointer. I did not knew search index is updated after every transaction. Closing this pull request as not required.

PR #1597, opened 2015-10-16
Removing deprecated methods - Trunk 4772

dkayiwa:
Removing deprecated methods - Trunk 4772


api/src/main/java/org/openmrs/PersonAttribute.java
```diff
@@ -296,14 +296,26 @@ public void voidAttribute(String reason) {
...
 	 * @should return negative if other attribute has lower value
 	 * @should return negative if this attribute has lower attribute id than argument
-	 * @should not throw exception if attribute type is null
-	 * @deprecated since 1.12. Use DefaultComparator instead.
-	 * Note: this comparator imposes orderings that are inconsistent with equals.
 	 */
-	@SuppressWarnings("squid:S1210")
 	public int compareTo(PersonAttribute other) {
-		DefaultComparator paDComparator = new DefaultComparator();
-		return paDComparator.compare(this, other);
```
wluyima: Aren't you just supposed to remove the deprecated annotation and not change the method code? Because it was already delegating to the the DefaultComparator anyways, i think this deprecation was invalid

api/src/main/java/org/openmrs/PersonName.java
```diff
@@ -531,21 +539,33 @@ public String toString() {
 	 * @should return negative if other familynamePrefix is greater
 	 * @should return negative if other familyNameSuffix is greater
 	 * @should return negative if other dateCreated is greater
-	 * @deprecated since 1.12. Use DefaultComparator instead.
-	 * Note: this comparator imposes orderings that are inconsistent with equals.
 	 */
-	@SuppressWarnings("squid:S1210")
 	public int compareTo(PersonName other) {
```
wluyima: Same here, you don't have to change the code, just remove the deprecate annotation

api/src/main/java/org/openmrs/PersonName.java
```diff
@@ -531,21 +539,33 @@ public String toString() {
...
-		return pnDefaultComparator.compare(this, other);
-	}
-	
-	/**
-	 * @since 1.5
-	 * @see org.openmrs.OpenmrsObject#getId()
-	 */
-	public Integer getId() {
-		return getPersonNameId();
+		int ret = isVoided().compareTo(other.isVoided());
```
wluyima: And get rid of this

dkayiwa:
Nice catch @wluyima. Committed at https://github.com/openmrs/openmrs-core/commit/d1d193f2b4a7d8d628c94dc7abc57fde993ae8a0

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -142,67 +122,18 @@
...
-	 * @return the given user unvoided
+	 * @return the saved user
 	 * @throws APIException
-	 * @should unvoid and unmark all attributes
-	 * @deprecated use {@link #unretireUser(User)}
 	 */
-	@Deprecated
 	@Authorized( { PrivilegeConstants.EDIT_USERS })
-	public User unvoidUser(User user) throws APIException;
+	public User saveUser(User user) throws APIException;
```
wluyima: Why are you renaming this to saveUser? Shouldn't you just be removing it?

dkayiwa: Since i had to have saveUser that takes a user object, it is just the same. :)

wluyima: I thought we already have saveUser() with 2 arguments

wluyima:
Github wasn't able to display all the entire PR's diff, so i was unable to review all the changes

api/src/main/java/org/openmrs/api/UserService.java
dkayiwa: We agreed to have saveUser(User) taking only one argument. Then createUser(User, password) takes two.

dkayiwa:
I noticed that too. Are you aware of any way of making it display more?

api/src/main/java/org/openmrs/api/UserService.java
wluyima: I thought createUser was deprecated and removed, saveUser should take an optional password so that we don't need a new saveUser, did i miss the design call where this was decided?

dkayiwa: Here is the ticket: https://issues.openmrs.org/browse/TRUNK-3871

wluyima: I see! But i would imagine that the change should be effective in the commit for that ticket and not this PR

dkayiwa: I noticed that we closed the ticket before all that was needed.

api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
```diff
@@ -254,51 +258,12 @@ public Encounter getEncounter(Integer encounterId) throws APIException {
 		}
 		
 		List<Encounter> encs = new Vector<Encounter>();
-		for (Patient p : Context.getPatientService().getPatients(null, identifier, null, false)) {
+		for (Patient p : Context.getPatientService().getPatients(identifier, null, null, false)) {
```
wluyima: Does this mean, this was a bug? Why did you have to change it?

dkayiwa: No it was not. If you look at the DAO, these methods were deprecated in favor of PatientDAO#getPatients(String, Integer, Integer) where the first parameter string can either be a name or identifier.

wluyima: I guess if it was working, then you might as well just leave it the way it was to avoid any issues coming up

dkayiwa: It actually failed after the deprecated methods removal. So this change was the fix.

dkayiwa:
Closing because i created a new one with one commit at https://github.com/openmrs/openmrs-core/pull/1600

PR #1598, opened 2015-10-17
- Fixed TRUNK-4778

seekme94:
-  The location Id, when greated than 1000 was beingreceived as grouped Id, like 1,000 causing NumberFormatException. Fixed by handling the Id in string using DecimalFormat with grouping flag
disabled.
- Added test case to reproduce bug reported in TRUNK-4778


dkayiwa:
@seekme94 the changes should be in the jsp or somewhere other than in the location editor because the convention of our editors is to accept numbers that are not formatted.

seekme94:
Interesting. But the stack trace doesn't say much on this. I'll dig into the JSP files now.

lluismf:
Could be some javascript also ...

PR #1602, opened 2015-10-22
Sonar:Unused private method should be removed

lluismf:
A few private methods removed according to Sonar rules


dkayiwa:
Thanks @lluismf :)

PR #1606, opened 2015-10-29
TRUNK-3871: Added back changePassword(User u, String pw)

geoff-wasilwa:
Added method to change user's password if one has correct privilege.

Also modified method changePassword(String pw1, String pw2) to
take either current password or answer to secret question when changing password.

PR for: https://issues.openmrs.org/browse/TRUNK-3871

### DIFF
--- a/api/src/main/java/org/openmrs/api/UserService.java
+++ b/api/src/main/java/org/openmrs/api/UserService.java
@@ -52,7 +52,7 @@ public interface UserService extends OpenmrsService {
 	 * Change user password.
 	 *
 	 * @param user the user to update password
-	 * @param oldPassword the user password to update
+	 * @param oldPassword the user password  to update
 	 * @param newPassword the new user password
 	 * @throws APIException for not existing user and if old password is weak
 	 * @since 1.12
@@ -311,6 +311,17 @@ public interface UserService extends OpenmrsService {
 	 */
 	@Logging(ignoredArgumentIndexes = { 0, 1 })
 	public void changePassword(String pw, String pw2) throws APIException;
+
+	/**
+	 * Changes password of {@link User} passed in
+	 * @param user user whose password is to be changed
+	 * @param newPassword new password to set
+	 * @throws APIException
+	 * @should update password of given user when logged in user has edit users password privilege
+	 * @should not update password of given user when logged in user does not have edit users password privilege
+	 */
+	@Authorized({PrivilegeConstants.EDIT_USER_PASSWORDS})
+	public void changePassword(User user, String newPassword) throws APIException;
 	
 	/**
 	 * Changes the current user's password directly. This is most useful if migrating users from
@@ -520,4 +531,15 @@ public interface UserService extends OpenmrsService {
 	 */
 	@Authorized
 	public User saveUserProperties(Map<String, String> properties);
+	
+	/**
+	 * Change user password given the answer to the secret question
+	 * @param secretAnswer the answer to secret question
+	 * @param pw the new password
+	 * @should update password if secret is correct
+	 * @should not update password if secret is not correct
+	 */
+	@Authorized
+	public void changePasswordUsingSecretAnswer(String secretAnswer, String pw) throws APIException;
+
 }
--- a/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
@@ -628,8 +628,26 @@ public class UserServiceImpl extends BaseOpenmrsService implements UserService {
 			throw new APIException("old.password.not.correct", (Object[]) null);
 		}
 	
+		updatePassword(user, newPassword);
+	}
+
+	private void updatePassword(User user, String newPassword) {
 		OpenmrsUtil.validatePassword(user.getUsername(), newPassword, user.getSystemId());
 		dao.changePassword(user, newPassword);
+	}
+
+	@Override
+	public void changePassword(User user, String newPassword) throws APIException {
+		dao.changePassword(user, newPassword);
+	}
+
+	@Override
+	public void changePasswordUsingSecretAnswer(String secretAnswer, String pw) throws APIException {
+		User user = Context.getAuthenticatedUser();
+		if(!isSecretAnswer(user, secretAnswer)) {
+			throw new APIException("secret.answer.not.correct", (Object[]) null);
+		}
+		updatePassword(user, pw);
 	}
 	
 }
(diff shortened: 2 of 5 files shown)


api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -65,7 +65,7 @@
 	 */
 	@Authorized( { PrivilegeConstants.EDIT_USER_PASSWORDS })
 	@Logging(ignoredArgumentIndexes = { 1, 2 })
-	public void changePassword(User user, String oldPassword, String newPassword) throws APIException;
+	public void changePassword(User user, String secret, String newPassword) throws APIException;
```
wluyima: Don't change this method because they should also be able to change their password by providing the old password

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -616,20 +616,26 @@ public User saveUserProperties(Map<String, String> properties) {
...
+			throw new APIException("old.password.or.secret.not.correct", (Object[]) null);
 		}
 	
 		OpenmrsUtil.validatePassword(user.getUsername(), newPassword, user.getSystemId());
 		dao.changePassword(user, newPassword);
 	}
+
+	@Override
+	public void changePassword(User user, String newPassword) throws APIException {
+		changePassword(user, null, newPassword);
```
wluyima: This should call the DAO directly as it did in 1.11

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -301,16 +301,29 @@
...
 	 * @should match on correctly hashed sha1 stored password
 	 * @should match on incorrectly hashed sha1 stored password
 	 * @should match on sha512 hashed password
 	 * @should be able to update password multiple times
+	 * @should update password if secret is correct
+	 * @should not update password if secret is not correct
 	 */
 	@Logging(ignoredArgumentIndexes = { 0, 1 })
-	public void changePassword(String pw, String pw2) throws APIException;
+	public void changePassword(String secret, String pw2) throws APIException;
```
wluyima: You need to add a separate new method for this and keep the old one that takes the old and new password

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1325,4 +1325,58 @@ public void changePassword_shouldThrowShortPasswordExceptionWithShortPassword()
...
     }
+    
+	@Test
+	@Verifies(value = "should update password of given user when logged in user has edit users password privilege", method = "changePassword(User,String)")
+	public void changePassword_shouldUpdatePasswordOfGivenUserWhenLoggedInUserHasEditUsersPasswordPrivilege() throws Exception {
+		UserService userService = Context.getUserService();
+		User user = userService.getUserByUsername("admin");
+		assertNotNull("There needs to be a user with username 'admin' in the database", user);
+		
+		userService.changePassword(user, "testTest123");
```
wluyima: You need to try and authenticate the user with the new password after this

api/src/main/java/org/openmrs/api/UserService.java
geoff-wasilwa: @wluyima I just changed the variable name because it accepts either old password or answer to secret question, `secret` sounded like a better name than `oldPassword`.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
geoff-wasilwa: What about the business rule in `changePassword(User, String, String)` that checks if user has required privilege. See [here](https://github.com/geoff-wasilwa/openmrs-core/blob/TRUNK-3871/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java#L623)

api/src/main/java/org/openmrs/api/UserService.java
wluyima: No, that is supposed to be oldPassword, you need to add a new method that takes secret question and newPassword only

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
wluyima: Because you changed it, that other method should only accept old and newPassword and not secret question

api/src/main/java/org/openmrs/api/UserService.java
geoff-wasilwa: The new method will have the same signature as `public void changePassword(String pw, String pw2)` so I just thought and have that method take either `secret answer` or `old password` and work either way.

lluismf: Then its signature will be confusing (parameters with mutable semantics is not a good idea in general).

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -520,4 +531,15 @@
 	 */
 	@Authorized
 	public User saveUserProperties(Map<String, String> properties);
+	
+	/**
+	 * Change user password by providing answer to secret question
```
wluyima: I think this should be "Change user password given the answer to the secret question"

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -631,5 +631,19 @@ public void changePassword(User user, String oldPassword, String newPassword) th
...
 	}
+
+	@Override
+	public void changePassword(User user, String newPassword) throws APIException {
+		dao.changePassword(user, newPassword);
+	}
+
+	@Override
+	public void changePasswordUsingSecretAnswer(String secretAnswer, String pw) throws APIException {
+		User user = Context.getAuthenticatedUser();
```
wluyima: The code here should pretty much be the same, only different should be that this checks that the secret answer is correct. You could factor the other code in the other changePassword method into a utility method that they both call after checking that the old or secret password is correct

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -520,4 +531,15 @@
...
+	 * @param secretAnswer the answer to secret question
+	 * @param pw the new password
+	 * @should update password if secret is correct
```
wluyima: I actually can't see these unit tests

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -628,8 +628,26 @@ public void changePassword(User user, String oldPassword, String newPassword) th
 			throw new APIException("old.password.not.correct", (Object[]) null);
 		}
 	
+		updatePassword(user, newPassword);
+	}
+
+	private void updatePassword(User user, String newPassword) {
```
wluyima: I thought you were going to update the other changePassword method too to delegate to this new utility method

PR #1609, opened 2015-11-08
Added mapping to standard test dataset to demonstrate …

WolfSchlegel:
… that this breaks an existing test. The mapping is commented and needs to be uncommented to see the test fail. 

Step 1: Uncomment line 141 in standardTestDataSet.xml
Step 2: Run org.openmrs.validator.DrugOrderValidatorTest#validate_shouldFailIfDurationUnitsHasNoMappingToSNOMEDCTSource(). 

The purpose of this test is to ensure that missing mappings are handled correctly. The tests expects an error that no longer occurs after the mapping has been added to the test data set.


dkayiwa:
Closing because this was only for demonstration purposes.

PR #1611, opened 2015-11-10
Update changePassword(User user, String newPassword)

geoff-wasilwa:
Update the changePassword method to delegate to the new updatePassword
utility method

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
@@ -638,7 +638,7 @@ public class UserServiceImpl extends BaseOpenmrsService implements UserService {
 
 	@Override
 	public void changePassword(User user, String newPassword) throws APIException {
-		dao.changePassword(user, newPassword);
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -638,7 +638,7 @@ private void updatePassword(User user, String newPassword) {
 
 	@Override
 	public void changePassword(User user, String newPassword) throws APIException {
-		dao.changePassword(user, newPassword);
+		updatePassword(user, newPassword);
```
wluyima: There is also another changePassword method that takes the old and new password

geoff-wasilwa: The changePassword method that takes old and new password delegates to `changePassword(String user, String oldPassword, String newPassword)` and this already uses the utility method.

wluyima: Ok, thanks

PR #1614, opened 2015-11-18
Some types shouldn't be serialized/deserialized

wluyima:


api/src/test/java/org/openmrs/serialization/SimpleXStreamSerializerTest.java
```diff
@@ -91,4 +90,18 @@ public void deserialize_shouldDeserializeStringToClassInstance() throws Serializ
...
 		
 	}
+	
+	/**
+	 * @see SimpleXStreamSerializer#deserialize(String,Class)
+	 * @verifies not deserialize dynamic proxies
+	 */
+	@Test
+	public void deserialize_shouldNotDeserializeDynamicProxies() throws Exception {
+		String serialized = "<dynamic-proxy>" + "<interface>org.openmrs.OpenmrsObject</interface>"
```
djazayeri: Could you please change the content here to be something that is _not_ a blueprint for an attack string that could work on older OpenMRS versions?

api/src/main/java/org/openmrs/serialization/SimpleXStreamSerializer.java
```diff
@@ -100,4 +107,28 @@ public String serialize(Object o) throws SerializationException {
...
+	 * attacks.
+	 */
+	private class OpenmrsDynamicProxyConverter extends DynamicProxyConverter {
+		
+		OpenmrsDynamicProxyConverter() {
+			super(null);
+		}
+		
+		public void marshal(Object value, HierarchicalStreamWriter writer, MarshallingContext context) {
+            //Should we serialize proxies?
```
djazayeri: This comment does not make sense to me.

api/src/main/java/org/openmrs/serialization/SimpleXStreamSerializer.java
```diff
@@ -100,4 +107,28 @@ public String serialize(Object o) throws SerializationException {
 			throw new SerializationException("Unable to deserialize class: " + clazz.getName(), e);
 		}
 	}
+	
+	/**
+	 * An instance of this converter needs to be registered with a higher priority than the rest so
+	 * that it's called early in the converter chain. This way, we can make sure we never get to
+	 * xstream's DynamicProxyConverter that can deserialize proxies which can be a channel for
+	 * attacks.
```
djazayeri: could you provide a link to something that documents/describes why we are doing this?

PR #1616, opened 2015-11-24
TRUNK-4796 | Drug name search should accept special characters

sravanthi17:
https://issues.openmrs.org/browse/TRUNK-4796


djazayeri:
Peeking at the code, it seems good. This needs to also be forward-ported to the 1.12.x and master branches.

PR #1617, opened 2015-11-24
Update Vagrant config and Puppet manifests for local dev

chenjr95:
This is most closely related to TRUNK-4028.

The previous Vagrant setup does not appear to work at the
moment, failing due to a missing .bashrc.erb template.  In
addition, that setup only aims to provide a MySQL instance,
rather than providing a full development box suitable for
running the application.

This borrows heavily from OpenMRS-Australia/openmrs-vagrant
to provide a box that makes it easier for new developers to
spin up an instance of openmrs-core.


dkayiwa:
Thanks @chenjr95 :)

PR #1621, opened 2015-11-26
RA-954 Completed ehCache configuration

jdegraft:


dkayiwa:
@jdegraft my atlas module compiles fine without this change to the platform. So which other modules should i test with? And what are the steps?

jdegraft:
@dkayiwa 
http://stackoverflow.com/questions/10013288/another-unnamed-cachemanager-already-exists-in-the-same-vm-ehcache-2-5

you only see this error when you upgrade spring and hibernate to the versions used in platform 2.0, because the spring/hibernate implementations have changed semantics allowing multiple unnamed caches.

I guess we have to wait for the discussions as to how to build modules for platform 2.0 in a backwards compatible manner.
If we are going for the same java vm 1.8, spring and hibernate versions then this change is needed for the tests to run.

Note, as far as my effort are concerned, java 1.6 and java 1.8 are build incompatible, and java 1.7 and java 1.8 are reflection incompatible. In other words, you can't build a module with java 1.6 against platform 2.0, and you can't build a module with java 1.7 with platform 2.0 and get reflection code to build.

PR #1628, opened 2015-12-07

Vincentes:
Fixed three critical issues on api/src/main/java/org/openmrs/module/ModuleFactory.java.
Critical issues:
org.openmrs.module.ModuleFactory.loadModules(List) makes inefficient use of keySet iterator instead of entrySet iterator.
org.openmrs.module.ModuleFactory.getLoadedModulesMapPackage() makes inefficient use of keySet iterator instead of entrySet iterator.
Redundant nullcheck of tmpExtensions, which is known to be not-null. org.openmrs.module.ModuleFactory.stopModule(Module, boolean, boolean)


chaityabshah:
Please squash these all into 1 commit.

PR #1629, opened 2015-12-08

Vincentes:
Selected critical issues
org.openmrs.module.ModuleFactory.loadModules(List) makes inefficient use of keySet iterator instead of entrySet iterator.
org.openmrs.module.ModuleFactory.getLoadedModulesMapPackage() makes inefficient use of keySet iterator instead of entrySet iterator.
Redundant nullcheck of tmpExtensions, which is known to be not-null. org.openmrs.module.ModuleFactory.stopModule(Module, boolean, boolean)

### DIFF
--- a/api/src/main/java/org/openmrs/module/ModuleFactory.java
+++ b/api/src/main/java/org/openmrs/module/ModuleFactory.java
@@ -193,8 +193,8 @@ public class ModuleFactory {
 		
 		//inform modules, that they can't start before other modules
 		Map<String, Module> loadedModulesMap = getLoadedModulesMapPackage();
-		for (String key : loadedModules.keySet()) {
-			Module m = loadedModules.get(key);
+		for (Entry<String, Module> entry : loadedModules.entrySet()) {
+			Module m = entry.getValue();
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -1169,10 +1169,7 @@ public static void stopModule(Module mod, boolean isShuttingDown) {
 						String extId = ext.getExtensionId();
 						try {
 							List<Extension> tmpExtensions = getExtensions(extId);
-							if (tmpExtensions == null) {
-								tmpExtensions = new Vector<Extension>();
-							}
-							
+							tmpExtensions = new Vector<Extension>();
```
Undo1: Are you sure about this line? Seems like it would change the behavior - `tmpExtensions` would always end up as a new Vector. 

With the old code, it's only initialized to a new Vector if `getExtensions()` fails.

Vincentes: Oh, I just realized that. The method never returns null though, it does the null checking itself.

PR #1632, opened 2015-12-13
Adding the Drug Ingredients to service layer

k-joseph:


djazayeri:
This is incorrect. DrugIngredient is already exposed as a property of Drug, so this is not needed.

(And if we were adding it anywhere it would be in ConceptService, alongside the getDrug methods, not in OrderService.)

PR #1634, opened 2015-12-14
TRUNK-248 changing TODO junit tests to @should

chenjr95:
@dkayiwa


dkayiwa:
@chenjr95 i thought the todos are to remind us about getting these test created and hence should be removed only when this is done. If we change to shoulds, then we cannot easily tell the missing tests. Did i just get this wrong? :)

dkayiwa:
@chenjr95 ping.

PR #1635, opened 2015-12-17
TRUNK-4782 : Retrospective edit of Orders

aj-jaswanth:
@djazayeri @angshu Created saveRetrospectiveOrder method which will delegate the responsibility to save() method.


aj-jaswanth:
@dkayiwa I first thought that, it is already OrderService so, save() would be sufficient. But, yeah saveOrder is more consistent with other method names.

PR #1636, opened 2015-12-19
TRUNK-4803: Fixed and simplified email regex

Vincentes:
It would not accept emails like `me@t-mobile.de`. Now it does.

### DIFF
--- a/api/src/main/java/org/openmrs/validator/UserValidator.java
+++ b/api/src/main/java/org/openmrs/validator/UserValidator.java
@@ -40,7 +40,7 @@ public class UserValidator implements Validator {
 	protected final Log log = LogFactory.getLog(getClass());
 	
 	private static final Pattern EMAIL_PATTERN = Pattern
-	        .compile("^[_A-Za-z0-9-]+(\\.[_A-Za-z0-9-]+)*@[A-Za-z0-9]+(\\.[A-Za-z0-9]+)*(\\.[A-Za-z]{2,})$");
+	        .compile("^.+@.+\\..+$");
 	
 	@Autowired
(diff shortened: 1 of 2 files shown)


robbyoconnor:
LGTM despite the CI failure -- it doesn't seem to be touching hide code -- I can see no reason why this change would cause any tests to fail. 

@dkayiwa, @bmamlin  :grinning:

lluismf:
It does:

isUserNameAsEmailValid_shouldReturnFalseIfEmailInvalid(org.openmrs.validator.UserValidatorTest)  Time elapsed: 0.007 sec  <<< FAILURE!
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertFalse(Assert.java:64)
	at org.junit.Assert.assertFalse(Assert.java:74)
	at org.openmrs.validator.UserValidatorTest.isUserNameAsEmailValid_shouldReturnFalseIfEmailInvalid(UserValidatorTest.java:203)

Vincentes:
The point of this was to make the email regex simpler, to later add email verification. This causes some checks to be valid. I'll refactor the tests.

robbyoconnor:
I was about to suggest one but it gets interesting to write a regexp, there's a saying that regex is almost never the answer(it introduces lil bugs that we never looked at before, I think e-mail confirmations are the answer here -- barring that maybe using Apache Commons, where they did the hard work for us!) -- n regex pattern to validate an email address...is insane -- consider that I own a `.ninja` domain name and some validators reject that TLD in email addresses!

PR #1639, opened 2015-12-20
TRUNK-4798: Resolving error in HibernateContextDAO.authenticate

Breta01:
This set default value in case that GP_ALLOWED_FAILED_LOGINS_BEFORE_LOCKOUT is null which caused the error.


robbyoconnor:
@Breta01, in the future use topic branches -- because now any changes you make to this branch will be also made here.

Breta01:
@dkayiwa It solves error when you run command "mvn package" log: http://pastebin.com/sgcxDusc And I already tested my solution.

dkayiwa:
@Breta01 that does not look like the correct solution. Do you see any other line in that validator which is throwing an exception when validation fails?

PR #1652, opened 2015-12-28
GCI-90: Fix three critical issues on OpenMRS sonarqube.

collinsgeek:
Removed inefficient use of keySet iterator instead of entrySet
Replaced inefficient new Long(long) constructor with Long.valueOf(long)

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
@@ -332,7 +332,7 @@ public class HibernateLocationDAO implements LocationDAO {
 		    Projections.rowCount()).add(Restrictions.eqProperty("alias.locationId", "outer.locationId"));
 		
 		return sessionFactory.getCurrentSession().createCriteria(Location.class, "outer").add(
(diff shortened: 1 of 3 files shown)


collinsgeek:
Hi Sir @Ch3ck , I tried that but it doesn't work since localized.values() returns a collection of PresentationMessageMap() while allMessage.addAll() is expecting a collection of PresentationMessage. I tried making the former to work for over 3 hours and still no avail. This was the best solution in this case. At least I discard the entrSet use case here.

Ch3ck:
Ok @collinsgeek  I see, but that's odd though since PresentationMessageMap implements PresentationMessage. It's supposed to return a PresentationMessage collection. Any ways good work with this.

@dkayiwa  @rkorytkowski  Any thoughts on this?

LGTM

robbyoconnor:
I'm not familiar with the codebase

Ch3ck:
Alright, any ways thanks for sticking with this @collinsgeek. But  If @dkayiwa  or @rkorytkowski  have any issues with this; you will have to fix it.

collinsgeek:
Hi Sir, @Ch3ck @dkayiwa 

I will make the changes and submit.

Ch3ck:
Nice work @collinsgeek  @dkayiwa I think you could merge this now. Let me go and approve this!

PR #1656, opened 2016-01-04
TRUNK-4782 : Retrospective edit of Orders

aj-jaswanth:
This is already merged in 1.12.x branch. This pull request is an upward merge.


wluyima:
I think you should have waited to merge this, i think all work going into 1.12.x needs to be merged at once into master. And the workflow is such that you don't have to create a pull request for forward and back ports, the dev who merges the original PR is responsible for cherry-picking the commit(s) and apply them to the necessary branches

PR #1658, opened 2016-01-05
Fix for TRUNK-3920

ThisuraThejith:
Stopped automatically preprending moduleId to the start of message properties


ThisuraThejith:
@dkayiwa Added the test copyModuleMessagesIntoWebapp_shouldNotPrefixMessagesWithModuleId.

ChamaraPhilipsuom:
That is put in the "buildModuleForMessageTest" class. So it should be there as it is created in the "buildModuleForMessageTest" method. Therefore it should still exist. Please correct if I am wrong.

PR #1663, opened 2016-01-11
RA-968 Fixed BaseContextSensitiveTest to handle @DirtiesContext

jdegraft:


dkayiwa:
@jdegraft are you still convinced that we need this? :)

jdegraft:
Well, without it, any module which uses DirtiesContext will fail tests in OpenMRS 2.0. Howevre it is not the best fix.

We need a method called reloadDatabase() really which modules can call if they use DirtiesContext.
As it is, only modules which use DirtiesContext and call authenticate() will be ok.

So we need reloadDatabase() which we can call in authenticate() as a final solution.
Also we need a test, which attempts to authenticate as a way of knowing that the database has been cleared before reloadDatabase() is called as it is an expensive operation and slows down tests, eg testCanAuthenticate(), needDatabaseReload().

The code speaks for it self and can be validated logically.
Also this code is only for testing and does not affect core platform functionality so is quite a safe change.

However the changes for appframework RA-968 has to have a new approach based on Darius's comments.

wluyima:
@jdegraft should't this be closed?

PR #1665, opened 2016-01-12
TRUNK-4805 | Retrospective stop of orders

aj-jaswanth:
Issue is reopened as a scenario was not considered in the previous commit.


api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -337,9 +337,13 @@ private void discontinueExistingOrdersIfNecessary(Order order, Boolean isRetrosp
 			return;
 		}
 
+		Date asOfDate = null;
+		if (isRetrospective)
```
wluyima: Our convention is to wrap if clauses in curly braces

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -337,9 +337,13 @@ private void discontinueExistingOrdersIfNecessary(Order order, Boolean isRetrosp
...
+			asOfDate = order.getDateActivated();
+
 		//Mark first order found corresponding to this DC order as discontinued.
 		List<? extends Order> orders = getActiveOrders(order.getPatient(), order.getOrderType(), order.getCareSetting(),
-		    null);
+		    asOfDate);
```
wluyima: I think i would disagree with doing this, discontinuing existing  orders is any only makes when doing real time order entry

vinayvenu: @wluyima If a user wants to record the fact that a discontinue order was issued on a certain day, how would he/she do it?

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -3109,5 +3109,52 @@ public void saveOrder_shouldRevisePreviousNonCodedOrderIfItIsAlreadyExisting() t
...
+
+		Order discontinueationOrder = originalOrder.cloneForDiscontinuing();
+		discontinueationOrder.setPreviousOrder(originalOrder);
+		discontinueationOrder.setEncounter(encounterService.getEncounter(17));
+		discontinueationOrder.setOrderer(providerService.getProvider(1));
+		discontinueationOrder.setDateActivated(discontinuationOrderDate);
+		orderService.saveRetrospectiveOrder(discontinueationOrder, null);
+
+		assertNotNull(originalOrder.getDateStopped());
+		assertEquals(discontinueationOrder.getAutoExpireDate(), discontinueationOrder.getDateActivated());
```
wluyima: Correct me if am wrong, shouldn't you instead be testing that the originalOrder.dateStopped is equal to discontinuationOrderDate.getDateActivated() but less by a moment?

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -3109,5 +3109,52 @@ public void saveOrder_shouldRevisePreviousNonCodedOrderIfItIsAlreadyExisting() t
...
+	}
+
+	@Test
+	public void saveRetrospectiveOrder_shouldDiscontinueAndStopActiveOrderInRetrospectiveEntry() throws Exception {
```
wluyima: I see no difference between this test and the one above except the test method name

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
wluyima: On the design call, we agreed to what you're saying so you can ignore my comment, there is more details on the ticket, see Darius' last comment or the updated ticket description under dev notes

api/src/test/java/org/openmrs/api/OrderServiceTest.java
aj-jaswanth: First one has a previous order. The second doesn't, so it tries to stop an active order by looking up the active orders present at that time.

wluyima: Ok, thanks!

api/src/test/java/org/openmrs/api/OrderServiceTest.java
wluyima: Did you see this comment?

dkayiwa:
@aj-jaswanth can you pull the latest upstream changes and resolve the conflicts?

PR #1676, opened 2016-01-20
TRUNK-2205: Jackson JSON cannot handle BaseOpenmrsObject

ivange94:
https://issues.openmrs.org/browse/TRUNK-2205

After deprecating all isXyz() having corresponding getXyz() methods, i realized that there were now calls to the deprecated isXyz() inside the code, even getXyz() only calls isXyz() internally. I will create a jira ticket to remove calls to the deprecated isXyz() methods.

### DIFF
--- a/api/src/main/java/org/openmrs/ConceptNumeric.java
+++ b/api/src/main/java/org/openmrs/ConceptNumeric.java
@@ -12,6 +12,7 @@ package org.openmrs;
 import java.util.HashSet;
 import java.util.TreeSet;
 
+import org.codehaus.jackson.annotate.JsonIgnore;
 import org.hibernate.search.annotations.Indexed;
 
 /**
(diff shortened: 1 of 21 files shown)


robbyoconnor:
Is convention when you have a boolean to always have `isFoo()` for getters? `getDead()` doesn't sound right to me for example.

robbyoconnor:
Disregard my previous comment.

dkayiwa:
@ivange94 within this pull request, can you change the deprecated methods to call the expected methods? For instance, replace "return dead" with "return getDead()"

ivange94:
@dkayiwa i have addressed your comment and updated the pull request

ivange94:
Boolean.valueOf(false) will actually return Boolean.FALSE.

return preferred == null ? Boolean.valueOf(false) : Boolean.valueOf(preferred);

return preferred == null ? Boolean.FALSE : preferred;

comparing the two statements i thought, it would be more obvious that i am returning a boolean value preferred when preferred is not equal to null and false when it is equal to null. I was aiming for readability. I was actually avoiding using the constructor Boolean(preferred), that i saw in the code.

dkayiwa:
The constructor is obviously bad. :)
preferred is already know to be a boolean.

ivange94:
is it ok now?

api/src/main/java/org/openmrs/ConceptNumeric.java
```diff
@@ -210,14 +211,19 @@ public void setDisplayPrecision(Integer displayPrecision) {
 	}
 	
 	public Boolean getAllowDecimal() {
-		return isAllowDecimal();
+		return allowDecimal == null ? false : allowDecimal;
 	}
```
robbyoconnor: ?! Why can't re rely on autoboxing here?

lluismf: Actually this prevents NPE when unboxing (Boolean to boolean). A null cannot be converted to a primitive boolean.

robbyoconnor: lame...just be like every other language and treat `null` as `false`...

PR #1679, opened 2016-01-20
Module architecture fails to populate aware_of modules in ModuleClassLoader if they start in the wrong order - TRUNK-4487

dkayiwa:
Module architecture fails to populate aware_of modules in ModuleClassLoader if they start in the wrong order - TRUNK-4487


mogoodrich:
I confirmed that this fixes my issue, thanks @dkayiwa

PR #1683, opened 2016-01-25
TRUNK-4805 | Retrospective discontinuation of orders.

nehashri:
creating a new pull request for the story.


api/src/main/java/org/openmrs/api/OrderService.java
```diff
@@ -105,7 +105,55 @@
...
 	 * specified then it will default to the one set on the passed in OrderContext if any otherwise
-	 * the save fails.
+	 * the save fails. Retrospective entry of orders can affect downstream systems that acts on orders created.
+     * Orders cannot be stopped if they are already stopped in retrospective entry.
+     *
+     * @param order the Order to save
+     * @param orderContext the OrderContext object
+     * @return the Order that was saved
+     * @throws APIException
+     * @should not save order if order doesnt validate
```
wluyima: Did you really add these @should with matching tests? Or you just copied them from saveOrder? I would assume @See saveOrder should be enough

nehashri: yes see would be enough, but I did add a few more and changed a few according to context.

wluyima: Each @should is added because you're adding a unit test which you're not doing particularly, so these @should are actually redundant, you can keep the javadocs though that says 'Retrospective entry of orders can affect downstream systems that acts on orders created. Orders cannot be stopped if they are already stopped in retrospective entry.' since it's specific to this method

dkayiwa:
@nehashri are you looking into these merge conflicts?

PR #1688, opened 2016-01-27
Redirecting Authenticated User to home if user tries to access 'login.htm' page [TRUNK-4817]

bholagabbar:
Issue: [[TRUNK-4817] Backport LUI-60 where user is redirected to home page if he/she tries to access login page while logged in.](https://issues.openmrs.org/browse/TRUNK-4817)<br>


wluyima:
Same here, i believe the issue is the same why the travis build is failing

bholagabbar:
What was the reason behind the branch closing? Am I missing something? Let's patch this issue up asap since LUI-60 is done and all that remains is backporting it. :)

wluyima:
See my comments on the ticket about the 2 PRs?

PR #1690, opened 2016-01-29
[TRUNK-4818] : Changed `presentationLocale` to LinkedHashSet to Preserve Insertion Order of Locales

bholagabbar:
**Issue**:[TRUNK-4818](https://issues.openmrs.org/browse/TRUNK-4818): 
In AdministrationServiceImpl.getPresentationLocales() the set we return is a HashSet, but this causes us to lose the ordering of locales (that ultimately comes from an admin-configured global property)

Instead we should use a LinkedHashSet, in order to preserve the order of locales we get from getAllowedLocales() (and ultimately the admin configured via a global property).<br>
**Fix**: Changed `presentationLocales` from `HashSet` to `LinkedHashSet` to preserve insertion order as discussed in ticket. Also added a Unit Test

### DIFF
--- a/api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
@@ -927,4 +927,41 @@ public class AdministrationServiceTest extends BaseContextSensitiveTest {
 		Assert.assertTrue("en", presentationLocales.contains(new Locale("en")));
 		Assert.assertTrue("es", presentationLocales.contains(new Locale("es")));
 	}
+	
+	/**
+	 * @see AdministrationService#getPresentationLocales()
+	 */
+	@Test
+	@Verifies(value = "should preserve insertion order in Set returned by method", method = "getPresentationLocales()")
+	public void getPresentationLocales_shouldPreserveInsertionOrderInSetReturnedByMethod()
+			throws Exception {
+		String globalPropertyLocaleListAllowedData = "en_GB, es, ja_JP, it_IT, pl_PL";
+		//The order of languages and locales is described above and should be followed bt `presentationLocales` Set
+		Context.getAdministrationService().saveGlobalProperty(
+				new GlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_LOCALE_ALLOWED_LIST, globalPropertyLocaleListAllowedData));
+		
+		List<Locale> locales = new ArrayList<Locale>();
+		//Add data in random order and verify that order is maintained in the end by checking against order in global property
+		locales.add(new Locale("pl", "PL"));
+		locales.add(new Locale("es"));
+		locales.add(new Locale("en"));
+		locales.add(new Locale("it", "IT"));
+		
+		MutableResourceBundleMessageSource mutableResourceBundleMessageSource = Mockito
+				.mock(MutableResourceBundleMessageSource.class);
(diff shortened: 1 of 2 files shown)


bholagabbar:
@djazayeri I have tried to make the unit test as verbose as possible. Put in a line comment if you need some clarification/explanation from my side :)

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -927,4 +928,59 @@ public void getPresentationLocales_shouldReturnLanguageLocaleIfItIsSpecifiedInAl
...
+		List<Locale> locales = new ArrayList<Locale>();
+		//Add data in random order and verify that order is maintained in the end by checking against order in global property
+		locales.add(new Locale("pl", "PL"));
+		locales.add(new Locale("es"));
+		locales.add(new Locale("en"));
+		locales.add(new Locale("it", "IT"));
+		
+		MutableResourceBundleMessageSource mutableResourceBundleMessageSource = Mockito
+				.mock(MutableResourceBundleMessageSource.class);
+		Mockito.when(mutableResourceBundleMessageSource.getLocales()).thenReturn(locales);
```
djazayeri: My preference would be to inline the definition of locales, as just `Arrays.asList(new Locale(...), ...)`

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -927,4 +928,59 @@ public void getPresentationLocales_shouldReturnLanguageLocaleIfItIsSpecifiedInAl
...
+		
+		MutableMessageSource mutableMessageSource = Context.getMessageSourceService().getActiveMessageSource();
+		Context.getMessageSourceService().setActiveMessageSource(mutableResourceBundleMessageSource);
+		
+		Set<Locale> presentationLocales = Context.getAdministrationService().getPresentationLocales();
```
djazayeri: From here onwards there is way too much unnecessary logic for the test. Just do something like:

List<Locale> presentationLocales = new ArrayList<Locale>(Context.getAdministrationService().getPresentationLocales());
assertThat(presentationLocales.get(0), is(...));
assertThat(presentationLocales.get(1), is(...));
// etc

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
bholagabbar: Ok but the above template is followed in all the other test methods. Wouldn't you prefer I keep it that way? I tried to emulate the pattern of the other tests as much as I could since I figured they would be they way someone wanted to for a reason. Should I let it be then?

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
bholagabbar: Yeah I thought I would make it generalized (sort of). I'll just manually assert the hard-coded data then as you said.

bholagabbar:
Updated. Seems fine?

djazayeri:
Looks great!

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
djazayeri: That's fine. Personally I tend to start writing with better more modern style.

I've merged this already, so no need to change it here. But going forwards I'd go with a style like this in tests like this:

List<String> expected = Arrays.asList(
        "one",
        "two",
        "three");

PR #1691, opened 2016-01-29
MessageSourceService.getMessage methods that take a MessageSourceResolvable should return the proper default message

ShekharReddy4:


ShekharReddy4:
@dkayiwa  the file changed doesn't have a Test class file in the core , So i did not write the test cases for the changes i have made. :)

dkayiwa:
@ShekharReddy4 did you run mvn clean install before committing this? The formatting does not look correct.

ShekharReddy4:
@dkayiwa  Updated the PR as per your comments. :)

ShekharReddy4:
@dkayiwa  updated as per your comments :)

ShekharReddy4:
@dkayiwa  written the required testcases. :)

api/src/main/java/org/openmrs/messagesource/impl/MessageSourceServiceImpl.java
```diff
@@ -96,7 +96,13 @@ public void setActiveMessageSource(MutableMessageSource activeMessageSource) {
 	 *      java.util.Locale)
 	 */
 	public String getMessage(MessageSourceResolvable resolvable, Locale locale) {
-		return activeMessageSource.getMessage(resolvable, locale);
+		if((resolvable.getCodes()[0]) == (activeMessageSource.getMessage(resolvable, locale))){
```
wluyima: I thought you are comparing strings, right? So i would expect use of equals

api/src/test/java/org/openmrs/messagesource/impl/MessageSourceServiceImplTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.openmrs.api.context.Context;
+import org.openmrs.test.BaseContextSensitiveTest;
+import org.springframework.context.MessageSourceResolvable;
+import org.springframework.validation.MapBindingResult;
+
+import java.util.HashMap;
+import java.util.Locale;
+
+
+public class MessageSourceServiceImplTest extends BaseContextSensitiveTest {
```
wluyima: This should be MessageSourceServiceTest

ShekharReddy4:
@wluyima  Updated the PR as per your comments.Renamed the Test class file and moved to other corresponding directory.

There are some problems  : 
1)Build  was Unsuccessful, giving a warning " Missing header in: openmrs-core\api\src\test\java\org\openmrs\messagesource\MessageSourceServiceTest.java" .  which i think because of the new test class file I have written.
2) Travis CiI build is failing.

@wluyima  Please let me know where I went wrong . :))

api/src/test/java/org/openmrs/messagesource/MessageSourceServiceTest.java
```diff
@@ -0,0 +1,47 @@
...
+    @Verifies(value = "getMessage_shouldReturnTheLastErrorCodeIfnoLocalizationIsFound", method = "getMessage(MessageSourceResolvable resolvable, Locale locale)")
+    public void getMessage_shouldReturnTheLastErrorCodeIfnoLocalizationIsFound() throws Exception {
+        MapBindingResult errors = new MapBindingResult(new HashMap<String, Object>(), "request");
+        errors.rejectValue("myField", "myErrorCode");
+        MessageSourceResolvable fieldError = errors.getFieldError("myField");
+        Assert.assertEquals(3, fieldError.getCodes().length);
+        Assert.assertEquals("myErrorCode.request.myField", fieldError.getCodes()[0]);
+        Assert.assertEquals("myErrorCode.myField", fieldError.getCodes()[1]);
+        Assert.assertEquals("myErrorCode", fieldError.getCodes()[2]);
+        Assert.assertNotEquals("myErrorCode.request.myField", Context.getMessageSourceService().getMessage(fieldError, Context.getLocale()));
```
djazayeri: This is unnecessary. The next line asserts what this value _should_ be. You don't need to also assert what it should _not_ be.

(In fact, if this test were ever to fail, it would be confusing to whoever looks at it because they'd be seeing the failure on this line, i.e. should _not_ be xyz, instead of the next line saying what the value _should_ be.)

api/src/main/java/org/openmrs/messagesource/impl/MessageSourceServiceImpl.java
```diff
@@ -96,7 +96,12 @@ public void setActiveMessageSource(MutableMessageSource activeMessageSource) {
 	 *      java.util.Locale)
 	 */
 	public String getMessage(MessageSourceResolvable resolvable, Locale locale) {
-		return activeMessageSource.getMessage(resolvable, locale);
+		if((resolvable.getCodes()[0]).equals((activeMessageSource.getMessage(resolvable, locale)))){
```
djazayeri: I'm quite surprised that this is the correct fix. (I would expect that activeMessageSource.getMessageSource(MessageSourceResolvable, Locale) would be part of a Spring class that should already be behaving correctly to Spring's specifications....

wluyima: Apparently it is, I've always guessed that this is a spring bug because we never do anything with resolving the code in our message source implementation, you can clearly see [here](http://grepcode.com/file/repository.springsource.com/org.springframework/org.springframework.context/3.0.3/org/springframework/context/support/AbstractMessageSource.java#155) that spring is returning the first message code which is contrary to what they preach in their javadocs

PR #1694, opened 2016-01-29
Redirecting Authenticated User to home if user tries to access 'login.htm' page [TRUNK-4817]

bholagabbar:
**Issue**: [[TRUNK-4817] A user should be redirected to home page if logged in instead of the login page](https://issues.openmrs.org/browse/TRUNK-4817)<br>

**Fix**: Redirect authenticated user with no privilege problems back to 'index.htm'


bholagabbar:
@wluyima : Error log says 'buffer overflow detected, Exiting. It fails on open-jdk-6 but passes on oracle-jdk-7. Check the link [here](https://travis-ci.org/openmrs/openmrs-core/jobs/105800629)

PR #1699, opened 2016-01-31
Redirecting Authenticated User to home if user tries to access 'login .htm' page [TRUNK-4817]

bholagabbar:
**Issue**: [[TRUNK-4817] Backport LUI-60 where user is redirected to home page if he/she tries to access login page while logged in.](https://issues.openmrs.org/browse/TRUNK-4817)<br>

**Note**: There is a travis build failure due to a buffer overflow which is not caused due to this commit. The current code as already been reviewed and verified for merging in LUI-60 which is fixed


bholagabbar:
Great! Finally done. I'll try and reproduce this on 1.10.x and 1.9.x. I'm currently working on a sprint ticket so hope this can wait. Dropping the DB and setting it up again is a real pain :P

PR #1702, opened 2016-02-02
Deleted Dependencies related to 'Template' class in notifications folder'

bholagabbar:
Try 1: Delete stuff related to 'Template'. Looks alright? The tests were failing before I even committed. Probably some commit last week caused it. Build passes otherwise


api/src/main/resources/applicationContext-service.xml
```diff
@@ -162,9 +162,6 @@
 	<bean id="programWorkflowDAO" class="org.openmrs.api.db.hibernate.HibernateProgramWorkflowDAO">
 		<property name="sessionFactory"><ref bean="sessionFactory"/></property>	
 	</bean>
-	<bean id="templateDAO" class="org.openmrs.api.db.hibernate.HibernateTemplateDAO">
-		<property name="sessionFactory"><ref bean="sessionFactory"/></property>	
-	</bean>
 	<bean id="patientSetDAO" class="org.openmrs.api.db.hibernate.HibernatePatientSetDAO">
 		<property name="sessionFactory"><ref bean="sessionFactory"/></property>
```
bholagabbar: Line 324 in this file as to go too right? Sorry I couldn't add a line comment there and github wouldn't allow me to do that

bholagabbar: I'm not sure how well I've accomplished the job. Quite a task this! :)

api/src/main/java/org/openmrs/notification/MessagePreparator.java
```diff
@@ -20,6 +20,6 @@
 	 * @param template
 	 * @return the prepared <code>Message</code>
 	 */
-	public Message prepare(Template template) throws MessageException;
+	//public Message prepare(Template template) throws MessageException;
```
djazayeri: Don't just comment this out, delete the whole file. (Note that this is going to generate a cascade of different things to delete or fix, but that is the point.)

api/src/main/resources/applicationContext-service.xml
djazayeri: yes, you should also be removing lines somewhere else in this file that were wiring templateDAO into some service.

api/src/main/java/org/openmrs/notification/MessagePreparator.java
bholagabbar: Yeah I actually went ahead and fixed a whole lot of stuff and it got wayyy to complex. So I'm doing it in parts. I'm currently dealing with this class now. It's a lot of files interlinked, espeically to this `MessagePreparator` class

dkayiwa:
Closing because the ticket was closed as a wont fix.

PR #1704, opened 2016-02-03
TRUNK-2296 PersonNameValidator gives the wrong field names

ShekharReddy4:


api/src/test/java/org/openmrs/validator/PersonNameValidatorTest.java
```diff
@@ -735,22 +736,17 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() throws Ex
...
-		Object obj = personNameValidatorClass.newInstance();
-		Method method = personNameValidatorClass.getDeclaredMethod("getFieldKey", new Class[]{String.class, boolean.class, boolean.class});
-		method.setAccessible(true);
-		String string1 = (String) method.invoke(obj,"givenName", false, true);
-		String string2 = (String) method.invoke(obj,"givenName", true, false);
-		Assert.assertEquals("givenName", string1);
-		Assert.assertEquals("names[0].givenName", string2);
+		MapBindingResult errors = new MapBindingResult(new HashMap<String, Object>(), "personName");
+		PersonNameValidator personNameValidator = new PersonNameValidator();
+		personNameValidator.validatePersonName(personName, errors, true, false);
```
djazayeri: This is wrong.

1. The point of this ticket is to test the standard Spring Validator.validate(Object, Errors) method, and you have changed this to instead test the validatePersonName method.
2. You are testing two different things in the same test.

Basically you should have one test like "validate should report errors on correct field names" and _another test_ like "validatePersonName should report errors with nonstandard prefix when called in historic way"

PR #1708, opened 2016-02-04
TRUNK-2004:Ignore blank person addresses on the patient forms

vishnuraom:


api/src/main/java/org/openmrs/api/handler/PersonSaveHandler.java
```diff
@@ -42,6 +42,9 @@ public void handle(Person person, User creator, Date dateCreated, String other)
 		// address collection
 		if (person.getAddresses() != null && person.getAddresses().size() > 0) {
 			for (PersonAddress pAddress : person.getAddresses()) {
+				if (pAddress.isBlank()){
```
wluyima: I would imagine that removeAddress and the call to PersonAddress.setPerson are technically mutually exclusive, so this should probably be an if-else clause or you should call continue after removing it. I know the ConceptSaveHandler code for descriptions has this same issue, i will fix it

vishnuraom: @wluyima Thanks for pointing it out,I have added a continue statement.

PR #1709, opened 2016-02-05
TRUNK-4539 : Object validation should enforce maximum length condition.

puevigreven:
changing validate method and test case

indentation and addition in test

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
@@ -27,6 +27,7 @@ import org.hibernate.cfg.Configuration;
 import org.hibernate.criterion.MatchMode;
 import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Restrictions;
+import org.hibernate.engine.spi.SessionImplementor;
 import org.hibernate.mapping.Column;
 import org.hibernate.mapping.PersistentClass;
 import org.openmrs.GlobalProperty;
@@ -215,7 +216,7 @@ public class HibernateAdministrationDAO implements AdministrationDAO, Applicatio
 	 * @should Pass validation for location class if field lengths are correct
 	 */
 
-	@SuppressWarnings({ "deprecation", "unchecked", "rawtypes" })
+	//@SuppressWarnings({ "deprecation", "unchecked", "rawtypes" })
 	@Override
 	public void validate(Object object, Errors errors) throws DAOException {
 		Class entityClass = object.getClass();
@@ -226,10 +227,11 @@ public class HibernateAdministrationDAO implements AdministrationDAO, Applicatio
 			String identifierName = metadata.getIdentifierPropertyName();
 			if (IdentifierTpye instanceof StringType || IdentifierTpye instanceof TextType) {
 				int maxLength = getMaximumPropertyLength(entityClass, identifierName);
-				String identifierValue = (String) metadata.getIdentifier(object);
+				String identifierValue = (String) metadata.getIdentifier(object,(SessionImplementor) null);
 				if (identifierValue != null) {
 					int identifierLength = identifierValue.length();
 					if (identifierLength > maxLength) {
+						
 						errors.rejectValue(identifierName, "error.exceededMaxLengthOfField", new Object[] { maxLength }, null);
 					}
 				}
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
@@ -14,6 +14,7 @@ import java.util.Vector;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.hibernate.metadata.ClassMetadata;
 import org.hibernate.Criteria;
 import org.hibernate.FlushMode;
 import org.hibernate.SessionFactory;
@@ -203,18 +204,37 @@ public class HibernateAdministrationDAO implements AdministrationDAO, Applicatio
 	 */
 	@Override
 	public void validate(Object object, Errors errors) throws DAOException {
+		// String clz = String.valueOf(object.getClass());
+		Class entityClass = object.getClass();
+		ClassMetadata metadata = sessionFactory.getClassMetadata(entityClass);
+		String[] propNames = metadata.getPropertyNames();
+		for (int i = 0; i < propNames.length; i++) {
+			String propType = metadata.getPropertyType(propNames[i]).getName();
+			if (propType.equalsIgnoreCase("String") || propType.equalsIgnoreCase("java.lang.String")) {
+				int p1 = getMaximumPropertyLength(entityClass, propNames[i]);
+				String str = (String) metadata.getPropertyValue(object, propNames[i]);
+				int p2 = str.length();
+				if (p2 > p1) {
+					errors.rejectValue(propNames[i], " can be at max [" + p1 + "] characters in length.");
+				}
+			}
+
+		}
 		FlushMode previousFlushMode = sessionFactory.getCurrentSession().getFlushMode();
 		sessionFactory.getCurrentSession().setFlushMode(FlushMode.MANUAL);
 		try {
 			for (Validator validator : getValidators(object)) {
 				validator.validate(object, errors);
 			}
+
 		}
+
 		finally {
 			sessionFactory.getCurrentSession().setFlushMode(previousFlushMode);
 		}
+
 	}
-	
+
 	/**
 	 * Fetches all validators that are registered
 	 *
@@ -237,7 +257,8 @@ public class HibernateAdministrationDAO implements AdministrationDAO, Applicatio
 	
 	@Override
 	public boolean isDatabaseStringComparisonCaseSensitive() {
-		GlobalProperty gp = (GlobalProperty) sessionFactory.getCurrentSession().get(GlobalProperty.class, OpenmrsConstants.GP_CASE_SENSITIVE_DATABASE_STRING_COMPARISON);
+		GlobalProperty gp = (GlobalProperty) sessionFactory.getCurrentSession().get(GlobalProperty.class,
+				OpenmrsConstants.GP_CASE_SENSITIVE_DATABASE_STRING_COMPARISON);
 		if (gp != null) {
 			return Boolean.valueOf(gp.getPropertyValue());
 		} else {
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
@@ -18,6 +18,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.hibernate.metadata.ClassMetadata;
 import org.hibernate.type.StringType;
+import org.hibernate.type.TextType;
 import org.hibernate.type.Type;
 import org.hibernate.Criteria;
 import org.hibernate.FlushMode;
@@ -208,37 +209,46 @@ public class HibernateAdministrationDAO implements AdministrationDAO, Applicatio
 	
 	/**
 	 * @see org.openmrs.api.db.AdministrationDAO#validate(java.lang.Object, Errors)
+	 * @should Pass validation if field lengths are correct
+	 * @should Fail validation if field lengths are not correct
+	 * @should Fail validation for location class if field lengths are not correct
+	 * @should Pass validation for location class if field lengths are correct
 	 */
+
+	@SuppressWarnings({ "deprecation", "unchecked", "rawtypes" })
 	@Override
 	public void validate(Object object, Errors errors) throws DAOException {
-		
 		Class entityClass = object.getClass();
-
 		ClassMetadata metadata = sessionFactory.getClassMetadata(entityClass);
 		if (metadata != null) {
 			String[] propNames = metadata.getPropertyNames();
 			Object IdentifierTpye = metadata.getIdentifierType();
 			String identifierName = metadata.getIdentifierPropertyName();
-			if (IdentifierTpye instanceof StringType && identifierName != null) {
-				ValidateUtil.validateFieldLengths(errors, object.getClass(), identifierName);
+			if (IdentifierTpye instanceof StringType || IdentifierTpye instanceof TextType) {
+				int maxLength = getMaximumPropertyLength(entityClass, identifierName);
+				String identifierValue = (String) metadata.getIdentifier(object);
+				if (identifierValue != null) {
+					int identifierLength = identifierValue.length();
+					if (identifierLength > maxLength) {
+						errors.rejectValue(identifierName, "error.exceededMaxLengthOfField", new Object[] { maxLength }, null);
+					}
+				}
 			}
 			for (int i = 0; i < propNames.length; i++) {
-				String propType = metadata.getPropertyType(propNames[i]).getName();
-				if (propType.equalsIgnoreCase("String")) {
-					Field fieldTerm = null;
-					try {
-						ValidateUtil.validateFieldLengths(errors, object.getClass(), propNames[i]);
-					} catch (NotReadablePropertyException e1) {//NotReadablePropertyException
-						validateUsingReflection(object, errors,propNames[i]);
-					}
-					 catch (InvalidPropertyException e1) {//InvalidPropertyException
-							validateUsingReflection(object, errors,propNames[i]);
+				Type propType = metadata.getPropertyType(propNames[i]);
+				if (propType instanceof StringType || propType instanceof TextType) {
+					String propertyValue = (String) metadata.getPropertyValue(object, propNames[i]);
+					if (propertyValue != null) {
+						int maxLength = getMaximumPropertyLength(entityClass, propNames[i]);
+						int propertyValueLength = propertyValue.length();
+						if (propertyValueLength > maxLength) {
+							errors.rejectValue(propNames[i], "error.exceededMaxLengthOfField", new Object[] { maxLength },
+									null);
 						}
+					}
 				}
-
 			}
 		}
-
 		FlushMode previousFlushMode = sessionFactory.getCurrentSession().getFlushMode();
 		sessionFactory.getCurrentSession().setFlushMode(FlushMode.MANUAL);
 		try {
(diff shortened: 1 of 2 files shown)


puevigreven:
@wluyima  please look into this. :)

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -200,9 +202,52 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
+		Class entityClass = object.getClass();
+		for (Field field : object.getClass().getDeclaredFields()) {
+			field.setAccessible(true);
+			if (field.getName().equalsIgnoreCase(name)) {
+				String prala = field.getName();
+				Object value = null;
+				try {
+					value = field.get(object);
+				} catch (IllegalArgumentException e) {
+					log.warn("Unable to get the value of " + field.getName());
```
puevigreven: @dkayiwa please have a look at this.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -200,9 +202,52 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
+	 * @should pass validation if all required fields have proper values
 	 */
 	@Override
 	public void validate(Object object, Errors errors) throws DAOException {
+		String name = object.getClass().getSimpleName();
+		Class entityClass = object.getClass();
+		for (Field field : object.getClass().getDeclaredFields()) {
+			field.setAccessible(true);
+			if (field.getName().equalsIgnoreCase(name)) {
+				String prala = field.getName();
```
lluismf: What's this variable for?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -200,9 +202,52 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
+				} catch (IllegalArgumentException e) {
+					log.warn("Unable to get the value of " + field.getName());
+				} catch (IllegalAccessException e) {
+					log.warn("Unable to get the value of " + field.getName());
+				}
+				if (value != null) {
+					int l1 = getMaximumPropertyLength(entityClass, field.getName());
+					int l2 = value.toString().length();
+					if (l2 > l1) {
+						errors.rejectValue(field.getName(), " can be at max [" + l1 + "] characters in length.");
```
lluismf: Do not harcode english messages, use properties.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -200,9 +202,51 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
+				} catch (IllegalArgumentException e) {
+					log.warn("Unable to get the value of " + field.getName());
+				} catch (IllegalAccessException e) {
+					log.warn("Unable to get the value of " + field.getName());
+				}
+				if (value != null) {
+					int l1 = getMaximumPropertyLength(entityClass, field.getName());
+					int l2 = value.toString().length();
+					if (l2 > l1) {
+						errors.rejectValue(field.getName(),  "error.exceededMaxLengthOfField",new Object[] { l1 }, null);
```
puevigreven: @lluismf please look into this, i have tried to use properties instead of hard coded message. 
thank you. :)

lluismf: Yes that's it. I assume error.exceededMaxLengthOfField property already exists, right?

puevigreven: yes @lluismf it exists. Its been used inValidateUtils.validateFieldLengths().

lluismf: I see, but then why don't you use ValidateUtils.validateFieldLengths ?

puevigreven:
@lluismf  i got the motivation to solve the problem in this way from @wluyima's comment https://issues.openmrs.org/browse/TRUNK-4539 at 2016-01-13 22:22:00 GMT+0550

lluismf:
I agree with Wyclif, but I still don't see why the validators use ValidateUtil.validateFieldLengths and we are providing a new method to do the same (apparently).

puevigreven:
@lluismf after going through the method ,  i managed to solve the problem using ValidateUtils.validateFieldLengths(). 
i have code ready for the same, please give me a nod to create a pr for that. 
Thank You :)

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -200,9 +202,51 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
 	
 	/**
 	 * @see org.openmrs.api.db.AdministrationDAO#validate(java.lang.Object, Errors)
+	 * @should pass validation if all required fields have proper values
 	 */
 	@Override
 	public void validate(Object object, Errors errors) throws DAOException {
+		String name = object.getClass().getSimpleName();
+		Class entityClass = object.getClass();
+		for (Field field : object.getClass().getDeclaredFields()) {
```
wluyima: Why don't you use hibernate's ClassMetadata, otherwise getDeclaredFields wouldn't return inherited fields and also not all fields defined on a class are persistent.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -200,9 +202,51 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
+			}
+		}
+
+		ClassMetadata metadata = sessionFactory.getClassMetadata(entityClass);
+		if(metadata != null){
+			String[] propNames = metadata.getPropertyNames();
+			for (int i = 0; i < propNames.length; i++) {
+				String propType = metadata.getPropertyType(propNames[i]).getName();
+				if (propType.equalsIgnoreCase("String")) {
+					int p1 = getMaximumPropertyLength(entityClass, propNames[i]);
```
wluyima: The way getMaximumPropertyLength is implemented, the argument value is supposed to be the database column and not the java property name which can be different e.g Location.cityVillage is named city_village in the DB

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -200,9 +202,51 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
+					if(str != null){
+						int p2 = str.length();
+						if (p2 > p1) {
+						errors.rejectValue(propNames[i], "error.exceededMaxLengthOfField",new Object[] { p1 }, null);
+						}
+					}
+				}
+
+			}
+		}
```
wluyima: Why are you using both reflection and hibernate class metadata to do the same thing? ClassMetadata should be all you need

api/src/test/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAOTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.springframework.validation.BindException;
+import org.springframework.validation.Errors;
+
+public class HibernateAdministrationDAOTest extends BaseContextSensitiveTest {
+
+	@Autowired
+	private HibernateAdministrationDAO dao;
+
+	private SessionFactory sessionFactory;
+
```
wluyima: Can you please have a test with a field where the column name differs from its mapped java property name? E.g Location.cityVillage

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
wluyima: I think we need to fix AdministrationService.getMaximumPropertyLength to be clear, i think it was intended to take the java property name and not the column name in the DB

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
lluismf: Exactly, just get the properties of string type and invoke ValidateUtil.validateFieldLengths. There are 52 references to this method that will be redundant after this change.

wluyima: @lluismf is right, just use hibernate class metadata to get the mapped fields of the persistent type and delegate to ValidateUtil.validateFieldLengths. The only problem is that there is a bud with this methods, it actually takes column names and not property names

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
lluismf: Its name is clear, otherwise it will be getMaximumColumnLength. 
And its implementation expects a property, not a column:
fieldLength = ((Column) persistentClass.getProperty(fieldName).getColumnIterator().next()).getLength();

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -237,7 +281,8 @@ public void validate(Object object, Errors errors) throws DAOException {
 	
 	@Override
 	public boolean isDatabaseStringComparisonCaseSensitive() {
-		GlobalProperty gp = (GlobalProperty) sessionFactory.getCurrentSession().get(GlobalProperty.class, OpenmrsConstants.GP_CASE_SENSITIVE_DATABASE_STRING_COMPARISON);
+		GlobalProperty gp = (GlobalProperty) sessionFactory.getCurrentSession().get(GlobalProperty.class,
```
lluismf: Why the line wrapping ?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
wluyima: I was on the 1.12.x branch, i see this was fixed in 2.0.x and master at 41207c6c8a065a5af97be14807a7fe2cb965fe0b. Previously the method actually used column name but it has been fixed ever since.

puevigreven:
@lluismf @wluyima i have tried using `ValidateUtils.validateFieldLengths() ` but we have certain test which fails. 
For example  `ConceptServiceTest.purgeConcept_shouldFailIfAnyOfTheConceptNamesOfTheConceptIsBeingUsedByAnObs` test fails.
I have investigated the problem, the problem is in ValidateUtils.validateFieldLengths()'s `Object value = errors.getFieldValue(field);
`
it gives an error when it take the parameter like `Obs.formNamespaceAndPath`
I guess that ValidateUtils.validateFieldLengths() was written to check fields which are taken as input from user. 
But since we are validating every item that is being saved to the DB through the service layer, we have to take extra care.
I propose the previous approach which i followed to solve this issue, here is the commit : https://github.com/puevigreven/openmrs-core/commit/b463f5f85e40ae59d0337344d6c690ac9409166b
I would also love to discuss this issue on wednesday's design call.

lluismf:
Maybe it fails because formNamespaceAndPath has no setter/getter ?

2016-02-16 13:46 GMT+01:00 pralay <notifications@github.com>:

puevigreven:
@lluismf yes i think the same, there are few other fields like this. 
 formNamespaceAndPath getter and setter won't be used in any place other than this.

lluismf:
Then the problem lies in Obs, not in the test or the validation algorithm.
If it's a property it needs a getter and setter.

2016-02-16 17:56 GMT+01:00 pralay <notifications@github.com>:

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -200,9 +204,56 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
+	 * @should Pass validation if field lengths are correct
+	 * @should Fail validation if field lengths are not correct
+	 * @should Fail validation for location class if field lengths are not correct
+	 * @should Pass validation for location class if field lengths are correct
 	 */
 	@Override
 	public void validate(Object object, Errors errors) throws DAOException {
+		Class entityClass = object.getClass();
+		String name = object.getClass().getSimpleName();
+		for (Field field : object.getClass().getDeclaredFields()) {
```
wluyima: You don't need to use reflection here, you need to delegate to ValidationUtil when looping over the class metadata below. The ClassMetadata Object has three methods to get information about identifiers, i.e getIndentifierPropertyName(), getIdentifier() and getIndentifierType, so you don't need to use reflection

wluyima:
The reason you're having issues with Obs.formNamespaceAndPath is because it has no getters, you might want to update the code in ValidateUtil.validateFieldLengths to use use reflection, i.e if it's throwing a sepcific exception catch it when it is thrown in case no getter exists, i then default to reflection just for that field

lluismf:
And why not just add the getter for Obs.formNamespaceAndPath? Is there a
good reason to not have it?

2016-02-17 4:03 GMT+01:00 Wyclif Luyima <notifications@github.com>:

wluyima:
From the an API consumer's point of view formNamespaceAndPath is meaningless that is why we don't expose property accessors for it, there is other methods on Obs that return meaningful values that use the field though. In any case this isn't only the only domain object where we do this, we won't keep adding the accessors for all because even them there are reasons behind the decisions, that is why we need to solve this in a generic way for fields with no accessors

lluismf:
ok

2016-02-17 16:42 GMT+01:00 Wyclif Luyima <notifications@github.com>:

puevigreven:
@wluyima @lluismf please look at the new commit.I tried to address all your comments. Please look into this.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -203,18 +211,82 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
+		if (metadata != null) {
+			String[] propNames = metadata.getPropertyNames();
+			Object IdentifierTpye = metadata.getIdentifierType();
+			String identifierName = metadata.getIdentifierPropertyName();
+			if (IdentifierTpye instanceof StringType && identifierName != null) {
+				ValidateUtil.validateFieldLengths(errors, object.getClass(), identifierName);
+			}
+			for (int i = 0; i < propNames.length; i++) {
+				String propType = metadata.getPropertyType(propNames[i]).getName();
+				if (propType.equalsIgnoreCase("String")) {
```
wluyima: you don't need to check for java.lang.String or does hibernate automatically convert it to string.
You also need to check for text

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -203,18 +211,82 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
 	public void validate(Object object, Errors errors) throws DAOException {
+		
+		Class entityClass = object.getClass();
+
+		ClassMetadata metadata = sessionFactory.getClassMetadata(entityClass);
+		if (metadata != null) {
+			String[] propNames = metadata.getPropertyNames();
+			Object IdentifierTpye = metadata.getIdentifierType();
+			String identifierName = metadata.getIdentifierPropertyName();
+			if (IdentifierTpye instanceof StringType && identifierName != null) {
```
wluyima: I would make this consistent with the one below

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -203,18 +211,82 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
+				ValidateUtil.validateFieldLengths(errors, object.getClass(), identifierName);
+			}
+			for (int i = 0; i < propNames.length; i++) {
+				String propType = metadata.getPropertyType(propNames[i]).getName();
+				if (propType.equalsIgnoreCase("String")) {
+					Field fieldTerm = null;
+					try {
+						ValidateUtil.validateFieldLengths(errors, object.getClass(), propNames[i]);
```
wluyima: I starting to think you should get the value from the class metadata and you don't delegate to ValidateUtil so that you don't need the reflective code in case there are fields with no getters

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -203,18 +211,82 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
+			String[] propNames = metadata.getPropertyNames();
+			Object IdentifierTpye = metadata.getIdentifierType();
+			String identifierName = metadata.getIdentifierPropertyName();
+			if (IdentifierTpye instanceof StringType && identifierName != null) {
+				ValidateUtil.validateFieldLengths(errors, object.getClass(), identifierName);
+			}
+			for (int i = 0; i < propNames.length; i++) {
+				String propType = metadata.getPropertyType(propNames[i]).getName();
+				if (propType.equalsIgnoreCase("String")) {
+					Field fieldTerm = null;
```
wluyima: Where do you use this?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -203,18 +211,82 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
 	 */
 	@Override
 	public void validate(Object object, Errors errors) throws DAOException {
+
```
wluyima: The code in you had in this commit 71fc81ddb092939e33f9e1f6b2e54fb8623a9b11 seems more correct to me as long you address the comments on it

puevigreven:
@wluyima please look into this commit, i have refined my previous approach and didn't use ValidateUtil
Please look into this.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -200,21 +209,61 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
 	public void validate(Object object, Errors errors) throws DAOException {
+		Class entityClass = object.getClass();
+		ClassMetadata metadata = sessionFactory.getClassMetadata(entityClass);
+		if (metadata != null) {
+			String[] propNames = metadata.getPropertyNames();
+			Object IdentifierTpye = metadata.getIdentifierType();
+			String identifierName = metadata.getIdentifierPropertyName();
+			if (IdentifierTpye instanceof StringType || IdentifierTpye instanceof TextType) {
+				int maxLength = getMaximumPropertyLength(entityClass, identifierName);
+				String identifierValue = (String) metadata.getIdentifier(object);
```
wluyima: this getIdentifier(Object) method is deprecated, can you please use the other one that takes a SessionImplementer instance?

api/src/test/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAOTest.java
```diff
@@ -0,0 +1,168 @@
...
+		location.setPostalCode(longString);
+		location.setLatitude(longString);
+		location.setLongitude(longString);
+		location.setCountyDistrict(longString);
+		location.setRetireReason(longString);
+		
+		Errors errors = new BindException(location, "location");
+		dao.validate(location, errors);
+		
+		Assert.assertTrue(errors.hasFieldErrors("name"));
```
wluyima: I think it would be nice to check the actual error message code because the field might be invalid but because of something else, you can loop over all the fields since the error message is expected to be the same

api/src/test/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAOTest.java
```diff
@@ -0,0 +1,168 @@
...
+	@Test
+	public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() throws Exception {
+		Role role = new Role();
+		role.setRole(
+				"too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text");
+		role.setDescription(
+				"too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text too long text");
+		Errors errors = new BindException(role, "type");
+		dao.validate(role, errors);
+		Assert.assertTrue(errors.hasFieldErrors("role"));
```
wluyima: check the message code

puevigreven:
@wluyima please look into latest commit https://github.com/puevigreven/openmrs-core/commit/ebbeae4e8c26cf1c6d58d517833e932fce31d422
i have made an effort to rectify the mistakes

wluyima:
Merged at 698c690825488bc29e97f12dad73fe26e8cbfdeb

PR #1716, opened 2016-02-11
TRUNK-4093 FIX- return null if uuid AND id is null

puevigreven:
In the description of the ticket, it was explicitly mention that return value of form should be null if uuid OR id is null.
While fixing the bug i realised that this will not solve the issue, since if either of uuid or id is present we get the form using that value (we prefer form id if uuid is not found or prefer form uuid over id if both are present), thus if either of the value is present the return value should not be null.
Incase both uuid and id is null then, the return value of should be null.


puevigreven:
@dkayiwa  the tests fail without my conditions

dkayiwa:
@puevigreven i have just run that test without your changes and it passes.

puevigreven:
@dkayiwa  i have changed the code just before commenting, please get the recent changes and try.

PR #1719, opened 2016-02-16
Trunk 3915

GautamSidtharth:
Adding a changeSet that adds uuids where it doesn't exist, before adding not null constraint.
This pull request is raised against the following issue https://issues.openmrs.org/browse/TRUNK-3915.


rkorytkowski:
Wrong PR (250+ commits).

PR #1720, opened 2016-02-16
TRUNK-3915: Adding a changeSet that adds uuids where it doesn't exist, before adding not null constraint.

GautamSidtharth:
Adding a changeSet that adds uuids where it doesn't exist, before adding not null constraint.
https://issues.openmrs.org/browse/TRUNK-3915.

### DIFF
--- a/api/src/main/resources/liquibase-update-to-latest.xml
+++ b/api/src/main/resources/liquibase-update-to-latest.xml
@@ -8650,7 +8650,32 @@
 			<column name="drug_non_coded" type="varchar(255)" />
 		</addColumn>
 	</changeSet>
-	
+
+	<changeSet id="20160216-1700" author="bahmni">
+		<comment>
+			Set uuid for columns in all tables which has uuid as null. This is required for successful run of next changeSet.
+		</comment>
+		<customChange class="org.openmrs.util.databasechange.GenerateUuid">
+			<param name="tableNames">
+				cohort concept concept_answer concept_class concept_datatype concept_description
+				concept_name concept_name_tag concept_proposal concept_reference_map concept_reference_source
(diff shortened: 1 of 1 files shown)


api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -8650,7 +8650,38 @@
...
 	</changeSet>
-	
+
+	<changeSet id="20160216-1700" author="bahmni">
+		<preConditions onFail="MARK_RAN">
+			<or>
+				<dbms type="mysql" />
+				<dbms type="oracle" />
+			</or>
+		</preConditions>
```
rkorytkowski: Please remove the precondition. GenerateUuid can be run on any database as it uses generic SQL.

api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -8650,7 +8650,32 @@
 			<column name="drug_non_coded" type="varchar(255)" />
 		</addColumn>
 	</changeSet>
-	
+
+	<changeSet id="20160216-1700" author="bahmni">
+		<comment>
+			Set uuid for columns in all tables which has uuid as null. This is required for successful run of next changeSet.
+		</comment>
+		<customChange class="org.openmrs.util.databasechange.GenerateUuid">
```
wluyima: The javadocs of this custom changesets class actually suggest that it should be used for only databases that don't have inbuilt uuid generator functions, so i think it that we should have the preconditions and only use this for those DBs, otherwise you should have 2 changesets with the second using the in built functions for the appropriate functions, we do have a couple changests that do this that you can look at. The primary reason i', saying this is because this that the idExceptions feature i think is flaky in that it assumes you have exhausted all tables that don't follow the naming convention of id columns which might not be the case, so i would avoid it as much as possible

PR #1721, opened 2016-02-17
TRUNK-4538 AbstractHandler needs to delete the previous file when a complex obs is updated with a new value

ShekharReddy4:


ShekharReddy4:
@wluyima Test cases included :)

PR #1724, opened 2016-02-25
TRUNK-4668: when a module is started, all its dependencies should be …

ivange94:
https://issues.openmrs.org/browse/TRUNK-4668

### DIFF
--- a/api/src/main/java/org/openmrs/module/ModuleFactory.java
+++ b/api/src/main/java/org/openmrs/module/ModuleFactory.java
@@ -234,16 +234,7 @@ public class ModuleFactory {
 					if (mod.isStarted()) {
 						continue; // skip over modules that are already started
 					}
-					
-					// Skip module if required ones are not started
-					if (!requiredModulesStarted(mod)) {
-						String message = getFailedToStartModuleMessage(mod);
-						log.error(message);
-						mod.setStartupErrorMessage(message);
-						notifySuperUsersAboutModuleFailure(mod);
-						continue;
-					}
-					
+										
 					try {
 						if (log.isDebugEnabled()) {
 							log.debug("starting module: " + mod.getModuleId());
@@ -579,7 +570,7 @@ public class ModuleFactory {
 	 * @see Daemon#startModule(Module)
 	 */
 	public static Module startModule(Module module) throws ModuleException {
-		return Daemon.startModule(module);
+		return startModule(module, false, null);
 	}
 	
 	/**
@@ -598,6 +589,31 @@ public class ModuleFactory {
 	 */
 	public static Module startModule(Module module, boolean isOpenmrsStartup,
 	        AbstractRefreshableApplicationContext applicationContext) throws ModuleException {
+		
+		if (!requiredModulesStarted(module)) {
+			int missingModules = 0;
+			
+			for (String packageName : module.getRequiredModulesMap().keySet()) {
+				Module mod = getModuleByPackage(packageName);
+				
+				// mod not installed
+				if (mod == null) {
+					missingModules++;
+					continue;
+				}
+				
+				if (!mod.isStarted())
+					startModule(mod);
+			}
+			
+			if (missingModules > 0) {
+				String message = getFailedToStartModuleMessage(module);
+				log.error(message);
+				module.setStartupErrorMessage(message);
+				notifySuperUsersAboutModuleFailure(module);
+				return module; // instead of return null, i realized that Daemon.startModule() always returns a Module object,irrespective of whether the startup succeeded  
+			}
+		}
 		return Daemon.startModule(module, isOpenmrsStartup, applicationContext);
 	}
 	
@@ -636,9 +652,10 @@ public class ModuleFactory {
 	public static Module startModuleInternal(Module module, boolean isOpenmrsStartup,
 	        AbstractRefreshableApplicationContext applicationContext) throws ModuleException {
 		
+		
 		if (module != null) {
-			
 			String moduleId = module.getModuleId();
+			
 			try {
 				
 				// check to be sure this module can run with our current version
@@ -794,7 +811,6 @@ public class ModuleFactory {
 				log.warn("Error while trying to start module: " + moduleId, e);
 				module.setStartupErrorMessage("Error while trying to start module", e);
 				notifySuperUsersAboutModuleFailure(module);
-				
 				// undo all of the actions in startup
 				try {
 					boolean skipOverStartedProperty = false;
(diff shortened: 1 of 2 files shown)


dkayiwa:
@ivange94 did you actually test this out by running the webapp? If yes, with what specific modules?

ivange94:
@dkayiwa i have not actually been able to test this on webapp. I tried to test from webapp. But the only module i could figure out with dependencies was the reporting module. But when i install the reporting module and its dependencies, the legacyui module stops working. The user interface disappears. The dependencies for reporting could not work well on 2.0.

dkayiwa:
@ivange94 if you are trying this out on platform 2.0, then you need to compile the latest snapshot versions of these modules. Can you try that again? It is important to manually test this out in the running webapp because it touches a critical part of the system.

ivange94:
@dkayiwa i was actually testing with latest version of these modules. But since it has been a long time since i ran the tests, i will try it again with the latest changes.

dkayiwa:
@ivange94 yes you need to try again because of recent i have made a number of random commits to make reference application modules run on platform 2.0

dkayiwa:
@ivange94 did you get a chance to try this out again?

ivange94:
Not yet. I have been a little occupied. I will look at it friday. If this is a blocker, i should be unassigned.

ivange94:
@dkayiwa i finally found some time to test these. I used the latest version of reporting from master. Well i can't still get all its dependencies even the updated versions to start in 2.0. But calculation has worked(it is also one of the dependencies for reporting). When i stop calculation then try to start reporting, calculation will be start automatically even though reporting will still fail to start because its other dependencies are not yet started. Am still working on the other dependencies, just wanted to let you know my progress so far.

ivange94:
@dkayiwa I made a small 2min video demonstrating that, this actually works. https://www.youtube.com/watch?v=07vS8qVzxDE&feature=youtu.be

ivange94:
@dkayiwa i uploaded a new video. The previous video had some parts strip out during the upload. https://www.youtube.com/watch?v=wv4mMgKTb2M&feature=youtu.be

api/src/test/java/org/openmrs/module/ModuleFactoryTest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.Assert;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.openmrs.test.BaseContextSensitiveTest;
+
+public class ModuleFactoryTest extends BaseContextSensitiveTest {
+
+	
+	@Test
+	@Ignore // TODO : this should be removed and make sure test passes
```
dkayiwa: So when will you ensure that this test passes? :)

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -574,7 +565,7 @@ public static Module getModuleByPackage(String modulePackage) {
 	 * @see Daemon#startModule(Module)
 	 */
 	public static Module startModule(Module module) throws ModuleException {
-		return Daemon.startModule(module);
```
dkayiwa: What is the reason for changing this?

api/src/test/java/org/openmrs/module/ModuleFactoryTest.java
dkayiwa: You can use the same modules as those in ModuleActivatorTest

api/src/main/java/org/openmrs/module/ModuleFactory.java
ivange94: I wanted to make sure a modules dependencies are always started before the module is starting. And i thought it would make sense to do that only in one method, that is why i delegated Daemon.startModule(Module) to the one that has my changes. Looking at the implementation of Daemon.startModule(Module) shows that my change makes no difference in the logic because Daemon.startModule(Module) will call Daemon.startModule(Module, false, null) which is essentially the same thing i have done but just at higher level. @dkayiwa does that make sense to you?

dkayiwa: Thanks is exactly why i asked :)

ivange94: @dkayiwa oh, i understand now, everything will still boil down to the method that has my changes. :)

ivange94: @dkayiwa i have done some checking again in the code and i realised that if i don't do that change as i did, when someone tries to start a module using ModuleFactory.startModule(module), following the method calls shows that it will end up in [ModuleFactory.startModuleInternal(Module, Boolean, ApplicationContext)](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/context/Daemon.java#L64), therefore completely bypassing my changes in the other method. I think this is what i was trying to avoid when i wrote the code, please check again and see if am right.

dkayiwa: I do not get you because Daemon.startModule(module) automatically calls startModule(module, false, null)

ivange94: yes, but its calling Daemon.startModule(module, false, null) not ModuleFactory.startModule(module, false, null) and my changes are in the latter method.

ivange94: @dkayiwa does the above comment explain my point?

dkayiwa: Yes. But the problem with your change is that the module no longer starts in a daemon thread.

ivange94: i thought this [line](https://github.com/ivange94/openmrs-core/blob/TRUNK-4668/api/src/main/java/org/openmrs/module/ModuleFactory.java#L612) is enough to start in a demon thread?

dkayiwa: Correct. Have you looked at ModuleFactory.startModules() in reference to your changes? See where is the comment: // Skip module if required ones are not started

ivange94: I presume you are talking about the section i deleted that was taking care of missing dependencies. I did that because i moved the functionality to [here](https://github.com/ivange94/openmrs-core/blob/TRUNK-4668/api/src/main/java/org/openmrs/module/ModuleFactory.java#L604) and since ModuleFactory.startModules() makes a call to ModuleFactory.startModule(module) [here](https://github.com/ivange94/openmrs-core/blob/TRUNK-4668/api/src/main/java/org/openmrs/module/ModuleFactory.java#L237) which ends up making a call to ModuleFactory.startModule(module, false, null), i didn't think it was necessary to leave that there again.

dkayiwa: Oh i see! Ok finish the tests. :)

ivange94: I will look at the examples you proposed in ModuleActivatorTest and see if can get the tests working before the end of tomorrow :).

dkayiwa:
@ivange94 did you notice that ModuleFactoryTest fails?

PR #1727, opened 2016-02-26
TRUNK-4678 | Fix ModuleActivatorTest

aj-jaswanth:
The main problem was with the order of plugins in pom.xml. While starting a module, it checks for required openmrs version. Since, the build-helper was out of order, the openmrs.properties file wasn't updated with any value. It was throwing start up error for modules.

### DIFF
--- a/api/src/test/java/org/openmrs/module/ModuleActivatorTest.java
+++ b/api/src/test/java/org/openmrs/module/ModuleActivatorTest.java
@@ -13,14 +13,12 @@ import static org.hamcrest.Matchers.lessThanOrEqualTo;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 
-import org.junit.Ignore;
 import org.junit.Test;
 
 /**
  * Tests methods of the module activator that do not require refreshing of the spring application
  * context. For those that require refreshing, see WebModuleActivatorTest
  */
-@Ignore("Unignore after investigating and fixing - TRUNK-4678")
 public class ModuleActivatorTest extends BaseModuleActivatorTest {
 	
 	@Test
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/openmrs/module/ModuleActivatorTest.java
```diff
@@ -93,15 +90,15 @@ public void shouldStopDependantModulesOnStopModule() throws Exception {
 		assertTrue(moduleTestData.getStoppedCallCount(MODULE1_ID) == 1);
 		assertTrue(moduleTestData.getStoppedCallCount(MODULE2_ID) == 1);
 		assertTrue(moduleTestData.getStoppedCallCount(MODULE3_ID) == 1);
-		
-		//willStop() and stopped() should have been called in the right order
-		//which is the reverse of the startup. that is module3, module2, module1
-		assertThat(moduleTestData.getWillStopCallTime(MODULE3_ID), lessThanOrEqualTo(moduleTestData
+
+        //willStop() should have been called in the order module1, module2, module3
```
aj-jaswanth: @dkayiwa I changed the comments. The comment in line 80 still stands true.

aj-jaswanth:
@dkayiwa I was wrong. Thanks for correcting. I thought the plugin run sequence was also affecting the parseVersion variables. But, it is fine.

aj-jaswanth:
The confusion came when I was running the test in IDE.

dkayiwa:
@aj-jaswanth i did you make any changes other than simply capitalizing L and moving some few lines around? :)

aj-jaswanth:
@dkayiwa The main change is to change the asserts. Capitalizing L is for readability :)

dkayiwa:
Shouldn't the order of calling willStop be module3, module2, module1?

aj-jaswanth:
No. Actually it first calls willStop() on module1 and recursively calls it on its dependent modules. So, the order will be module1, module2, module3.
https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/module/ModuleFactory.java#L1091

aj-jaswanth:
@dkayiwa Fixed formatting. The problem was that, by default the code uses 'tabs' for spacing. My IDE converted 'tabs' to equivalent number of spaces.

PR #1670, opened 2016-01-14
Trunk-3422(new privileges which are not added by default)

puevigreven:
Adding a liquibase changeset that inserts the "View Encounter Provider" and "View Visits" privileges for roles that have "View Encounter".

PR #1729, opened 2016-03-01
Trunk-3422(new privileges which are not added by default) #1670

puevigreven:


puevigreven:
@dkayiwa i have removed the merge conflicts, 
please look into this :)

dkayiwa:
@puevigreven did we remove "Get Encounter Provider"? I do not seem to see it anywhere.

dkayiwa:
@puevigreven did you see the comment above?

api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -9056,4 +9056,53 @@
...
+                <sqlCheck expectedResult="0">
+                SELECT count(*) FROM role_privilege  WHERE privilege='View Encounters';
+                </sqlCheck>
+            </not>
+        </preConditions>
+        <comment>
+            Add "Get Encounter Provider" privilege to the roles having "View Encounter"
+        </comment>
+            <sql>
+                INSERT INTO privilege (privilege,description)
```
wluyima: Do you really need to insert the privilege? I thought the API does it automatically unless if you saying that they're not yet inserted by the time this gets run. If it's the case, I would think this should be split into 2 changesets to be more readable where one adds the the privilege and the other associates to the appropriate roles

downeym:
This PR has been open 75 days. What is its current status? 📅 🕚

PR #1733, opened 2016-03-05
TRUNK-4837: removed redundant declaration of validator beans

micgn:
the ticket is to be approved, this is the solution


dkayiwa:
@micgn do you mind pointing me to the exact place where these are redundantly declared?

micgn:
only short distance below in the commited file you can see the component-scan, going for all Validators

PR #1734, opened 2016-03-05
Redirect user when logged in away from login page if already authenticated

bholagabbar:
Attempt Fix for [TRUNK-4817](https://issues.openmrs.org/browse/TRUNK-4817)

### DIFF
--- a/web/src/main/java/org/openmrs/web/WebConstants.java
+++ b/web/src/main/java/org/openmrs/web/WebConstants.java
@@ -74,7 +74,7 @@ public class WebConstants {
 	/**
 	 * Global property name for the number of times one IP can fail at logging in before being
 	 * locked out. A value of 0 for this property means no IP lockout checks.
-	 * 
+	 *
 	 * @see org.openmrs.web.servlet.LoginServlet
 	 */
 	public static String GP_ALLOWED_LOGIN_ATTEMPTS_PER_IP = "security.loginAttemptsAllowedPerIP";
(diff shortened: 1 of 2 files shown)


bholagabbar:
@wluyima have a look

web/src/main/java/org/openmrs/web/servlet/LoginServlet.java
```diff
@@ -59,6 +59,17 @@
 	private Map<String, Date> lockoutDateByIP = new HashMap<String, Date>();
 	
 	/**
+	 * @see javax.servlet.http.HttpServlet#doGet(javax.servlet.http.HttpServletRequest,
+	 *      javax.servlet.http.HttpServletResponse)
+	 */
+	@Override
+	protected void doGet(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
+		if(Context.getAuthenticatedUser() != null) {
```
wluyima: Don't you need to also check if there was a privilege check that failed like you did in the other commit? Otherwise you don't want a user to be taken to the login page if they are already logged in when they failed a privilege check

bholagabbar: Yes right. I'll do that. But that was being done in a controller and we added model attributes there when privi. Should we add a LoginFormController here then?

bholagabbar:
@wluyima I've added a controller for this page but control is not being transferred to it. I suspect I need to mention that this is the controller for the login page in a file somewhere but not sure where. Could you help me out?

wluyima:
I didn't actually mean that you add the controller though it might be okay. The @Controller annotation is all you need to map it to the URL

wluyima:
What i asked was that in the goGet method you would check if the privilege fail session attribute exists

bholagabbar:
I've already added the annotation at the top. Still doesn't work..do I add it elsewhere? I believe @RequestMapping takes care of the stuff when it lands on the page. That's done too.

bholagabbar:
@wluyima : Functional code ready :)

PR #1736, opened 2016-03-06
Trunk 4813

mnsowmya:


mnsowmya:
@dkayiwa In the link they mentioned to add both new check sum as well the previous old one

dkayiwa:
@mnsowmya shouldn't  you be adding only for the previous one?

mnsowmya:
@dkayiwa will there be any problem if i mention both can you tell me what would be the problem.i am not sure about this

dkayiwa:
@mnsowmya we generally do not include those that are not yet released. Does this make sense?

mnsowmya:
@dkayiwa ok

PR #1746, opened 2016-03-17
TRUNK-4838 Remove org.openmrs.arden package from core

ShekharReddy4:


sjmckee:
Is the arden package being moved to a separate module???  We're currently using this in Production.

ShekharReddy4:
No, it is not moved to a module yet but, I'll try doing that

PR #1747, opened 2016-03-17
TRUNK-4841: Using @JoinColumn instead of @Column with @ManyToOne in B…

ivange94:
…aseOpenmrsData


bholagabbar:
Ha! Easy ticket to fix :1234: :)

PR #1749, opened 2016-03-23
Obs not voided when updated via saving an Encounter - Trunk 50

wluyima:


dkayiwa:
Could you squash these into one commit? :)

wluyima:
I will, i think i got the fixes to the failures

wluyima:
I will once i push the fix to the failing tests, worked on it on the plane and just got internet in UG

PR #1754, opened 2016-04-03
TRUNK-4845:Update drug_order.dosing_type to use fully-qualified Java …

vishnuraom:
…class names

### DIFF
--- a/api/src/main/resources/liquibase-update-to-latest.xml
+++ b/api/src/main/resources/liquibase-update-to-latest.xml
@@ -9280,5 +9280,23 @@
 			<column name="sort_weight" type="double"></column>
 		</addColumn>
 	</changeSet>
+
+
+	<changeSet id="201604030003" author="vishnuraom">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<sqlCheck expectedResult="0">SELECT count(*) FROM drug_order WHERE dosing_type = 'SIMPLE' OR dosing_type = 'FREE_TEXT';</sqlCheck>
+			</not>
+		</preConditions>
+		<comment>Converting values in drug_order.dosing_type column(TRUNK-4845)</comment>
+		<update tableName="drug_order">
+			<column name="dosing_type" value="org.openmrs.SimpleDosingInstructions"/>
+			<where>dosing_type = 'SIMPLE'</where>
+		</update>
+		<update tableName="drug_order">
+			<column name="dosing_type" value="org.openmrs.FreeTextDosingInstructions"/>
+			<where>dosing_type = 'FREE_TEXT'</where>
+		</update>
+	</changeSet>
 	
 </databaseChangeLog>


api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -9280,5 +9280,23 @@
...
+	<changeSet id="201604030003" author="vishnuraom">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<sqlCheck expectedResult="0">SELECT count(*) FROM drug_order</sqlCheck>
+			</not>
+		</preConditions>
+		<comment>Converting values in drug_order.dosing_type column(TRUNK-4845)</comment>
+		<update tableName="drug_order">
+			<column name="dosing_type" value="org.openmrs.SimpleDosingInstructions"/>
+			<where>dosing_type = '0'</where>
```
kaweesi: Dosing type is no-longer either of these value, please change the where clause to use the right values as mentioned on the ticket

api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -9280,5 +9280,23 @@
 			<column name="sort_weight" type="double"></column>
 		</addColumn>
 	</changeSet>
+
+
+	<changeSet id="201604030003" author="vishnuraom">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<sqlCheck expectedResult="0">SELECT count(*) FROM drug_order</sqlCheck>
```
kaweesi: You could actually zero down this check to the real column you are interested in

k-joseph:
This looks fine, thanks @vishnuraom

api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -9280,5 +9280,23 @@
...
+
+
+	<changeSet id="201604030003" author="vishnuraom">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<sqlCheck expectedResult="0">SELECT count(*) FROM drug_order WHERE dosing_type = 'SIMPLE' OR dosing_type = 'FREE_TEXT';</sqlCheck>
+			</not>
+		</preConditions>
+		<comment>Converting values in drug_order.dosing_type column(TRUNK-4845)</comment>
+		<update tableName="drug_order">
```
wluyima: These changesets needs to be two, one for converting SIMPLE and the other FREE_TEXT

kaweesi: I think since it is an update query and the precondition is an either or check, then the whole changeset is good enough instead of  being split into two

wluyima: You should never  combine changesets like this because:
1. The precondition tests 2 scenarios where some installations can meet just one of them
2. If one of the updates fails, it should not affect the other.

api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -9280,5 +9280,23 @@
...
+			<column name="dosing_type" value="org.openmrs.SimpleDosingInstructions"/>
+			<where>dosing_type = 'SIMPLE'</where>
+		</update>
+		<update tableName="drug_order">
+			<column name="dosing_type" value="org.openmrs.FreeTextDosingInstructions"/>
```
lluismf: Excuse my ignorance, but is liquibase able to perform this update? What will be the actual value when executed?

lluismf: I've read https://issues.openmrs.org/browse/TRUNK-4845 to understand why the classname is stored here. It's kind of weird. What happens if the class is refactored? Is there any good reason to do this ?

wluyima: Initially we added dosing_type as an enum and was required, we went a head to add a changeset that set the default values for existing rows. Later we changed it from being an enum to being fully qualified java class names for reasons i won't go through and were legit, but we forgot to change the changeset that sets the default values for existing rows from being enum values to class names and this is meant to fix that.

PR #1756, opened 2016-04-06
TRUNK-208: Fix Transactions Across Service Methods

maany:
https://issues.openmrs.org/browse/TRUNK-208


api/src/main/java/org/openmrs/logic/LogicService.java
```diff
@@ -18,7 +18,6 @@
 import org.openmrs.logic.result.Result;
 import org.openmrs.logic.result.Result.Datatype;
 import org.openmrs.logic.rule.RuleParameterInfo;
-import org.springframework.transaction.annotation.Transactional;
```
dkayiwa: By the way, do we use this service anywhere?

maany: Nope, no implementations in core.

PR #1759, opened 2016-04-10
Added tests to org.openmrs.web.WebUtilTest. The tests are for org.ope…

winglam:
…nmrs.web.WebUtil.normalizeLocale to validate the arguments accepted by this method. The test normalizeLocale_shouldNotFailWithUnderline fails when testing the normalizeLocale method with the argument "_". However according to the comments of normalizeLocale, "_" is a valid argument.

A bug report related to this commit has been reported here:
https://issues.openmrs.org/browse/TRUNK-4848

### DIFF
--- a/web/src/main/java/org/openmrs/web/WebUtil.java
+++ b/web/src/main/java/org/openmrs/web/WebUtil.java
@@ -290,13 +290,14 @@ public class WebUtil implements GlobalPropertyListener {
 	 * @should not accept invalid locales
 	 * @should not fail with empty strings
 	 * @should not fail with whitespace only
+	 * @should not fail with "_" character only
 	 */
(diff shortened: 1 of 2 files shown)


dkayiwa:
@winglam according to the ticket, you said that the normalize method throws an exception. But your unit tests show that it does not throw this exception even without any fix.

winglam:
@dkayiwa The normalize method throws an exception before we applied our fix. This fix was missing before and has now been added. The unit tests should now pass.

dkayiwa:
@winglam why would anyone pass "_" for locale?

winglam:
The same argument can be made for why anyone would pass in empty strings or whitespaces only. But more importantly, this method was never meant to throw a NullPointerException for any input and yet "_" was one that would.

PR #1760, opened 2016-04-10
Adding tests for org.openmrs.util.Format in org.openmrs.util.FormatTe…

winglam:
…st. The added tests calls org.openmrs.util.Format.format with nulls and valid arguments. The following tests fail because it is expecting org.openmrs.util.Format.format to return null. format_shouldAllowAllNullParameters, format_shouldAllowLocaleTypeNullParameters, format_shouldAllowLocaleNullParameters, and format_shouldAllowDateLocaleNullParameters gets a NullPointerException instead and format_shouldReturnDateWithNullType and format_shouldAllowDateNullParameters receives an empty string as the return.

A bug report related to this pull request can be found on: 
https://issues.openmrs.org/browse/TRUNK-4849

### DIFF
--- a/api/src/main/java/org/openmrs/util/Format.java
+++ b/api/src/main/java/org/openmrs/util/Format.java
@@ -62,8 +62,30 @@ public class Format {
 	public static String format(Date date, FORMAT_TYPE type) {
 		return format(date, Context.getLocale(), type);
 	}
-	
+
+	/**
+	 * This method formats a date object according to a particular locale and returns the date as a string.
+	 * The string can contain only the date (month, day and year), only the time (hours, minutes, seconds) or as a
+	 * timestamp (both date and time).
+	 *
+	 * @param date input date to format as a string
+	 * @param locale input locale to determine how to format the date
+	 * @param type input type to determine how much information from the date is returned
+	 * @return empty string if one of the parameters is null. Otherwise a string object for the date such that it is
+	 * formatted according to locale and the amount of information it contains is determined by type.
+	 * @should not fail when only date is null
+	 * @should not fail when only locale is null
+	 * @should not fail when only type is null
+	 * @should not fail when date and locale is null
+	 * @should not fail when date and type is null
+	 * @should not fail when locale and type is null
+	 * @should not fail when all parameters are null
+	 * @should not fail when none of the parameters are null
+	 */
 	public static String format(Date date, Locale locale, FORMAT_TYPE type) {
+		if (date == null || locale == null || type == null) {
+			return "";
+		}
 		log.debug("Formatting date: " + date + " with locale " + locale);
 		
 		DateFormat dateFormat = null;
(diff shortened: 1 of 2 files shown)


dkayiwa:
@winglam did you take a look at this? https://wiki.openmrs.org/display/docs/Unit+Testing+With+at-should+Annotation

winglam:
@dkayiwa Thanks for pointing out the @should annotation. I have added that, other documentation, null checks to org.openmrs.util.Format and changed the tests in org.openmrs.util.FormatTest to respect the documentation.

dkayiwa:
Can you squash these commits into one? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

winglam:
I have squashed the commits.

api/src/main/java/org/openmrs/util/Format.java
```diff
@@ -62,8 +62,25 @@ public static String format(Date date) {
...
+	 * The string can contain only the date (month, day and year), only the time (hours, minutes, seconds) or as a
+	 * timestamp (both date and time).
+	 *
+	 * @param date input date to format as a string
+	 * @param locale input locale to determine how to format the date
+	 * @param type input type to determine how much information from the date is returned
+	 * @return empty string if one of the parameters is null. Otherwise a string object for the date such that it is
+	 * formatted according to locale and the amount of information it contains is determined by type.
+	 * @should not fail with null date
+	 * @should not fail with null locale
```
dkayiwa: Did you add a should annotation for each and every test you added?

winglam: No I did not. However I believe these three annotations are sufficient for the 8 tests that I have added. 3 of the 8 tests test the case when one of these parameters are null. Another 3 of the 8 tests test the case when two of the parameters are null. The remaining two tests test the case when all of them are null or none of them are null. The should annotations I have added implies that regardless of whether any of the parameters are null or not, the method should not fail.

dkayiwa: If you want to break the convention, you will need to start a discussion on talk to defend your argument.
We even have a tool that generates a test for each should annotation. https://wiki.openmrs.org/display/docs/Generate+Test+Case+Plugin

winglam: I did not realize that each should annotation needs to correspond to a test. I have added/changed the should annotations so that each one corresponds to a test case now. Thanks.

dkayiwa:
@winglam if you check again on the above wiki page, you will notice that the should annotation matches the method name.

winglam:
Sorry, I seem to have overlooked that from before. I have now changed the method names to match the should annotation.

dkayiwa:
@winglam hopefully you are aware of this plugin which should automatically generate test methods from the should annotations: https://wiki.openmrs.org/display/docs/Generate+Test+Case+Plugin

PR #1763, opened 2016-04-12
Changed getExtensions() so extensionNames expands when contents do not match

ern2:
This pull request accompanies JIRA ticket TRUNK-4850 (https://issues.openmrs.org/browse/TRUNK-4850)

In Module.java, getExtensions() was changed so that extensionNames are expanded when its contents do not match extensions.  Originally, getExtensions() would not call expandExtensionNames() if extensions and extensionNames were the same size, even if the contents of the two variables did not match.  The JIRA ticket goes into more detail on this issue.  

The Unit tests added in ModuleTest.java test these changes when extensionNames is null, is empty, matches extensions, and is the same size as extensions but does not match it.

### DIFF
--- a/api/src/test/java/org/openmrs/module/ModuleTest.java
+++ b/api/src/test/java/org/openmrs/module/ModuleTest.java
@@ -1,0 +1,124 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module;
+
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+import static org.powermock.api.mockito.PowerMockito.verifyPrivate;
+import static org.powermock.api.mockito.PowerMockito.spy;
+import static org.mockito.Mockito.never;
+
+import org.junit.Test;
+import org.junit.Before;
+import org.junit.runner.RunWith;
+
+import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/openmrs/module/ModuleTest.java
```diff
@@ -0,0 +1,127 @@
...
+		Extension mockExtension = new MockExtension();
+		mockExtension.setPointId("1");
+		extensions.add(mockExtension);
+		extensionNames.put("2", mockExtension.getClass().getName());
+
+		mockModule.setExtensions(extensions);
+		ArrayList<Extension> ret = new ArrayList<Extension>(mockModule.getExtensions());
+
+		//getExtensionNames will return an empty extensions list since MockExtension won't
+		//have a valid class loader
```
dkayiwa: According to this comment, this assertion has nothing to do with what this test. In other words, the assertion would pass if MockExtension had a valid class loader.

dkayiwa:
@ern2 can you just spy on expandExtensionNames() being called?
Here are some examples:
https://github.com/openmrs/openmrs-module-owa/commit/fdf1d06e2f19e591e2377650feb388cb1eb4d2f7
https://github.com/openmrs/openmrs-module-webservices.rest/commit/21d868e9f435f649cb2681661fa08b68790a574e

ern2:
Thanks for pointing out those examples.  I will get back to you with those changes.

ern2:
I updated this branch with master before this most recent commit and it now shows the file changes from the update.  Was that the wrong thing to do?  

Otherwise, I have added Powermock to my tests to spy on expandExtensionNames.

dkayiwa:
@ern2 this does not look correct because it includes more than your changes. Which commands did you use? By the way, have you ever seen this? https://wiki.openmrs.org/display/docs/Using+Git

ern2:
@dkayiwa, I was fetching the upstream then merging instead of using rebase as is done in the wiki page you just linked.  I have re-done it now and have run mvn clean package.  I am about to push again.  Hopefully this will fix the issue.

ern2:
@dkayiwa I have only made it worse.  Would it be easier at this point to delete this branch, re-branch from master, and then add my changes back in?

ern2:
@dkayiwa, deleting the branch and starting over is what finally worked.  Sorry for making such a mess.  Only my changes are showing now.

PR #1768, opened 2016-04-15
TRUNK-4854: error in uni test of WebModuleUtilTest

astorga2:
class: openmrs-core\web\src\test\java\org\openmrs\web\WebModuleUtilTest
the unit test:  startModule_shouldCreateDwrModulesXmlIfNotExists
attempts to test the startModule method in WebModuleUtil.java
 
public static boolean startModule(Module mod, ServletContext servletContext, boolean delayContextRefresh)

however this unit test will not always pass all the time.
In the case where variable realPath(a directory) in startModule contains whitespace(meaning that some clients directory name contains a space), for example dir/subdir/subdir 1/, the unit test will throw a ModuleException class. This is because in startModule the FileNotFoundException is thrown and it is caught at line 276 in WebModuleUtil.java

### DIFF
--- a/web/src/test/java/org/openmrs/module/web/WebModuleUtilTest.java
+++ b/web/src/test/java/org/openmrs/module/web/WebModuleUtilTest.java
@@ -128,6 +128,24 @@ public class WebModuleUtilTest {
 		
 		ModuleFactory.getStartedModulesMap().clear();
 	}
+
+	@Test(expected = ModuleException.class)
+	public void startModule_shouldnotThrowModuleExceptionWhenSpacesInDirectonryName() throws Exception {
+		// create dummy module and start it
+		Module mod = buildModuleForMessageTest();
+		ModuleFactory.getStartedModulesMap().put(mod.getModuleId(), mod);
+		
+		ServletContext servletContext = mock(ServletContext.class);
+		String realPath = servletContext.getRealPath("");
+		if (realPath == null)
+			realPath = System.getProperty("user.dir");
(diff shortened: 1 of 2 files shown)


dkayiwa:
@astorga2 can you have not more than one should in your test method name and also include a should annotation on the method being tested, as per https://wiki.openmrs.org/display/docs/Unit+Testing+With+at-should+Annotation

astorga2:
I have added should annotation, updated the name of unit test and squashed commits.

web/src/test/java/org/openmrs/module/web/WebModuleUtilTest.java
```diff
@@ -128,6 +128,24 @@ public void startModule_shouldCreateDwrModulesXmlIfNotExists() throws Exception
...
 		ModuleFactory.getStartedModulesMap().clear();
 	}
+
+	@Test(expected = ModuleException.class)
+	public void startModule_shouldnotThrowModuleExceptionWhenSpacesInDirectonryName() throws Exception {
+		// create dummy module and start it
+		Module mod = buildModuleForMessageTest();
+		ModuleFactory.getStartedModulesMap().put(mod.getModuleId(), mod);
+		
+		ServletContext servletContext = mock(ServletContext.class);
```
dkayiwa: Your method name says should NOT throw module exception. But your test expects it. Do you notice this contradiction?

dkayiwa:
Am closing because i tried a path with spaces and it worked well.

PR #1776, opened 2016-04-16
TRUNK-4858 startModule method should not throw ModuleExeception

astorga2:
Class: web/src/main/java/org/openmrs/module/web/WebModuleUtil.java
method: startModule(Module mod, ServletContext servletContext, boolean delayContextRefresh)
When String realPath = getrealPath(servletContext) return a directory name containing spaces, startModule method should not throw ModuleException
when a directory name contains a whitespace line 251 (inputStream = new FileInputStream(f) throws FileNotFoundException then a ModuleException is thrown.
startModule should not fail simply because there is a whitespace in the directory name.
see pull request with unit test that causes failure:

### DIFF
--- a/web/src/test/java/org/openmrs/module/web/WebModuleUtilTest.java
+++ b/web/src/test/java/org/openmrs/module/web/WebModuleUtilTest.java
@@ -256,5 +256,22 @@ public class WebModuleUtilTest {
 		
 		WebModuleUtil.setDispatcherServlet(dispatcherServlet);
 	}
+	/*
+	startModule should be fixed. ModuleException should not be thrown
+	*/
+	@Test(expected = ModuleException.class)
+	public void startModule_shouldnotThrowModuleExceptionWhenRealPathContainsWhitespace() throws Exception {
+		// create dummy module and start it
+		Module mod = buildModuleForMessageTest();
+		ModuleFactory.getStartedModulesMap().put(mod.getModuleId(), mod);
+		
+		ServletContext servletContext = mock(ServletContext.class);
+		String realPath = System.getProperty("user.dir");
+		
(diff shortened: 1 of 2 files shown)


web/src/test/java/org/openmrs/module/web/WebModuleUtilTest.java
```diff
@@ -256,5 +256,22 @@ private static void setupMocks(boolean includeTrailingSlash) {
 		
 		WebModuleUtil.setDispatcherServlet(dispatcherServlet);
 	}
+	/*
+	startModule should be fixed. ModuleException should not be thrown
+	*/
+	@Test(expected = ModuleException.class)
+	public void startModule_shouldCreateDwrModulesXmlIfNotExists_shouldnotThrowsModuleException() throws Exception {
```
dkayiwa: Our convention is for each test to have one should instead of two as you have done here. https://wiki.openmrs.org/display/docs/Unit+Testing+With+at-should+Annotation

astorga2:
I have updated the name of unit test to  contain one "should"

dkayiwa:
@astorga2 did you also add the corresponding should annotation on the method being tested?

astorga2:
Yes, I have added the corresponding should annotation.

dkayiwa:
@astorga2 am closing this because that failure was not caused by having spaces in the path. Just try remove space in your change and you will still get the same failure. It is simply caused by a non existing file. If the file exists, even if the path has spaces, it works

astorga2:
This is fine. 
In any case, if I try   home/astorga2 home/openmrs-core  and run mvn test from the openmrs-core directory, I still get the error (with or without the file).
I'll see if I can debug this further and update.

PR #1777, opened 2016-04-16
Added more tests to ModuleTest.java and moved the Module.getExtensions() tests

ern2:
The accompanying JIRA ticket is: https://issues.openmrs.org/browse/APPTEST-56

Since the Module.getExtensions() tests in ModuleTest.java used PowerMock, their test coverage is not tracked by jacoco.  These tests were moved into their own file, ModuleExtensionsTest.java, so that new tests could be added to ModuleTest.java and still be trackied with jacoco and Sonarqube.

### DIFF
--- a/api/src/main/java/org/openmrs/module/Module.java
+++ b/api/src/main/java/org/openmrs/module/Module.java
@@ -718,6 +718,10 @@ public final class Module {
 		return ModuleFactory.isModuleStarted(this);
 	}
 	
+	/**
+	 * @param e string to set as startup error message
+	 * @should throw exception when message is null
+	 */
 	public void setStartupErrorMessage(String e) {
 		if (e == null) {
 			throw new ModuleException("Startup error message cannot be null", this.getModuleId());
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/Module.java
```diff
@@ -718,6 +718,10 @@ public boolean isStarted() {
 		return ModuleFactory.isModuleStarted(this);
 	}
 	
+	/**
+	 * @param e string to set as startup error message
+	 * @should throw exception when message is null
```
dkayiwa: @ern2 my recommendation is that you spend your energy while prioritizing on real functionality tests instead of simple null argument validation tests. :)

ern2:
@dkayiwa, I will not concentrate on null value tests so much in the future.  Thank you for the advice.  Would you like me to take them out of these tests or leave them in since they are already written?

dkayiwa:
@ern2 just leave them since they are already written anyway! :)
By the way, have you looked at our community priority tickets?

ern2:
@dkayiwa, I haven't looked at the priority tickets very closely.  I wanted to focus on testing the Module package since I (and others I am working with) have experienced the most issues when trying to install and use modules with the core.  I will make sure to look at priority tickets from now on, too.

PR #1782, opened 2016-04-25
TRUNK-4861 adding unit test to ConceptTest

astorga2:
Adding unit test for previously un-tested methods (business logic).


dkayiwa:
@astorga2 when you create a ticket, can you wait for it to be accessed before you can work on it? We would like to prioritise tests on real business logic instead of simple setters, getters, remove, add item properties. Do you agree? :)

astorga2:
This is fair. I see the interesting logic in openmrs needs to be tested with an in-memory database. Please see here for what I would like to do before proceeding.
https://talk.openmrs.org/t/on-writing-mock-test-for-openmrs/5935

PR #1790, opened 2016-04-28
Add tests for TRUNK-4864 and fix NullPointerException issue

ldf92:
If we call method PatientSetService#getPatientsHavingDrugOrder(java.util.Collection, java.util.Collection, GroupMethod, Date, Date) with following parameters, it will throw NullPointerException.

Parameters:
patientIds = null
takingIds = a collection list with number of element larger than zero
groupMethod = GroupMethod.NONE
fromDate = null
toDate = null

Original code fails for added test: getPatientsHavingDrugOrder_shouldGetAllPatientsDoNotContainAnyDrugInTheDrugList

Ticket: https://issues.openmrs.org/browse/TRUNK-4864

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/PatientSetServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientSetServiceImpl.java
@@ -249,7 +249,9 @@ public class PatientSetServiceImpl extends BaseOpenmrsService implements Patient
 		} else {
 			if (groupMethod == GroupMethod.NONE) {
 				// Patients taking none of the specified drugs
-				
+				if (patientIds == null) {
+					patientIds = getAllPatients().getMemberIds();
+				}
 				// first get all patients taking no drugs at all
 				ret.addAll(patientIds);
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/api/impl/PatientSetServiceImpl.java
```diff
@@ -249,7 +249,9 @@ public Cohort getPatientsHavingDrugOrder(Collection<Integer> patientIds, Collect
 		} else {
 			if (groupMethod == GroupMethod.NONE) {
 				// Patients taking none of the specified drugs
-				
+				if (patientIds == null) {
```
wluyima: I wonder what you're coding against, I see this check [here](https://github.com/ldf92/openmrs-core/blob/PatientSetServiceTest/api/src/main/java/org/openmrs/api/impl/PatientSetServiceImpl.java#L239) already

wluyima: Never mind, I was looking at a different line

PR #1791, opened 2016-04-28
TRUNK-4845:Update drug_order.dosing_type to use fully-qualified Java …

vishnuraom:
…class names

### DIFF
--- a/api/src/main/resources/liquibase-update-to-latest.xml
+++ b/api/src/main/resources/liquibase-update-to-latest.xml
@@ -9298,5 +9298,33 @@
 			<where>dosing_type = 'FREE_TEXT'</where>
 		</update>
 	</changeSet>
-	
+
+
+	<changeSet id="201604281645" author="vishnuraom">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<sqlCheck expectedResult="0">SELECT count(*) FROM drug_order WHERE dosing_type = 'SIMPLE'</sqlCheck>
+			</not>
+		</preConditions>
+		<comment>Converting values in drug_order.dosing_type column from SIMPLE to org.openmrs.SimpleDosingInstructions (TRUNK-4845)</comment>
+		<update tableName="drug_order">
+			<column name="dosing_type" value="org.openmrs.SimpleDosingInstructions"/>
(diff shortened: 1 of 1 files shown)


api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -9298,5 +9298,33 @@
 			<where>dosing_type = 'FREE_TEXT'</where>
 		</update>
 	</changeSet>
-	
+
+
+	<changeSet id="201604281645" author="vishnuraom">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<sqlCheck expectedResult="0">SELECT count(*) FROM drug_order WHERE dosing_type = 'SIMPLE';</sqlCheck>
```
wluyima: I don't think you need the semi colon at the end

api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -9298,5 +9298,33 @@
...
+			</not>
+		</preConditions>
+		<comment>Converting values in drug_order.dosing_type column(TRUNK-4845)</comment>
```
wluyima: I think the comment needs to be specific that you're converting for simple dosing type so that it is different from that of the next changeset

api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -9298,5 +9298,33 @@
...
+	</changeSet>
+
+
+	<changeSet id="201604281646" author="vishnuraom">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<sqlCheck expectedResult="0">SELECT count(*) FROM drug_order WHERE dosing_type = 'FREE_TEXT';</sqlCheck>
+			</not>
+		</preConditions>
+		<comment>Converting values in drug_order.dosing_type column(TRUNK-4845)</comment>
```
wluyima: Same here

api/src/main/resources/liquibase-update-to-latest.xml
wluyima: Although this works too

vishnuraom:
@wluyima Sorry for the delay,I have updated the PR with required changes.

PR #1793, opened 2016-04-30
Fix issue LUI-63. Catch exception if there is error when refresh context

ldf92:
Catch exception if there is error when refreshing context.


api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -841,13 +842,21 @@ public static AbstractRefreshableApplicationContext refreshApplicationContext(Ab
 		try {
 			ctx.refresh();
 		}
+		catch (Exception e){
+			//Unload the module that cause the Exception and refresh again to restore
```
wluyima: I thought the ticket requires fixing why the exception is thrown and not handling it

dkayiwa:
Closing because the ticket was closed with a cannot reproduce resolution.

PR #1794, opened 2016-05-02
Added test as per TRUNK-4867

hirschf2:
Added test that throws exception if a deathdate is set for a patient before their birthdate.

### DIFF
--- a/api/src/test/java/org/openmrs/PersonTest.java
+++ b/api/src/test/java/org/openmrs/PersonTest.java
@@ -23,6 +23,7 @@ import java.util.Date;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


hirschf2:
This test should fail until the bug is patched, hence the API exception.

sunbiz:
@hirschf2 you should include the fix in this PR for validating death date here - https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/validator/PersonValidator.java

dkayiwa:
@hirschf2 did you see the comment from @sunbiz?

dkayiwa:
@hirschf2 i can reopen if anyone commits a fix that makes this test pass.

PR #1800, opened 2016-05-07
Added unit tests as per TRUNK-4880

Lemun:
Added unit tests for exception handling and two get methods. Upon testing, I have found that getDatatypeOrDefault() can return a NullPointerException is null is passed in, however the paramater dictates that a CustomValueDescriptor should be passed into it.
Similarly, getDatatype() and getHandler() will also throw a NullPointerException if null is passed in. This could be a bug or a feature, and in my tests I assume it is a feature.


dkayiwa:
@Lemun am closing this because of the comments i have put on the ticket.

PR #1802, opened 2016-05-19
TRUNK-4885 | Preethi, Chethan | Concept Attribute

chethandeshpande:


api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1780,7 +1785,102 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
+	@Override
+	@Transactional(readOnly = true)
+	public ConceptAttribute getConceptAttributeByUuid(String uuid) {
+		return dao.getConceptAttributeByUuid(uuid);
+	}
+
+	/**
+	 * @see org.openmrs.api.ConceptService#hasAnyConceptAttribute(ConceptAttributeType)
+	 */
+	@Override
```
dkayiwa: Shouldn't this also be transactional read only?

api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -9307,4 +9307,112 @@
 		</update>
 	</changeSet>
 
+	<changeSet id="20160427-0950-create-concept-attribute-type-table" author="bahmni">
+		<createTable tableName="concept_attribute_type">
```
dkayiwa: Don't you want to add the if not table exists precondition?

dkayiwa: You may also want to add a comment

PR #1808, opened 2016-06-23
TRUNK-4720 search will not match on id unless it contains digits

blavigne13:
## Description

Previously, the code made a decision on whether the query was a name or identifier by using a regex to determine if the query contained one or more digits, then created the appropriate PatientSearchCriteria. In cases of identifiers without digits, or names with them, it was not possible to obtain the correct result. Now, both names and identifiers are searched, without attempting to categorize the query beforehand.

Behavior has also been slightly changed: An exact match is now required for identifier to produce a result. Note: this only applies to identifier; name will still match on partial search terms. This is consistent with how other "name or identifier" queries are handled (e.g.: Find Patient, Find Provider) and is presumably for performance reasons.

Code changes:
Removed local vars "name" and "identifier" since "query" is used in both cases
Removed conditional branching based upon query type
Arguments to prepareCriteria have had the following changes applied:
  "query" is now passed as both name and identifier
  "true" is now passed for both matchIdentifierExactly and searchOnNamesOrIdentifiers

## Related Issue

## Checklist:

### DIFF
--- a/api/src/main/java/org/openmrs/api/EncounterService.java
+++ b/api/src/main/java/org/openmrs/api/EncounterService.java
@@ -381,6 +381,8 @@ public interface EncounterService extends OpenmrsService {
 	 * @should get all unvoided encounters for the given patient identifier
 	 * @should throw error if given null parameter
 	 * @should include voided encounters in the returned list if includedVoided is true
+	 * @should get all unvoided encounters for the given substring of patient names
+	 * @should not get encounters for the given substring of patient identifiers
 	 * @since 1.7
 	 */
 	@Authorized( { PrivilegeConstants.GET_ENCOUNTERS })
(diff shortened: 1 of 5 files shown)


dkayiwa:
@blavigne13 can you create some unit tests for this?

blavigne13:
@dkayiwa Will do. 

After giving it some thought, my plan is to put together a few tests to ensure that substring matches are only returned for names, and not identifiers. So far that's the only thing I've noticed that the existing tests don't cover. They could also be used for similar queries, e.g.: find patient, find provider, but I'm not sure if adding those same tests to methods unrelated to this ticket would be appropriate...

dkayiwa:
@blavigne13 the pull request requires squashing commits into one. Did you get a chance to look at this? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

dkayiwa:
@blavigne13 if you find it hard rectifying this, feel free to ask for help on the IRC channel. https://wiki.openmrs.org/display/IRC/Home

dkayiwa:
@blavigne13 am closing this so that you create a new one.

PR #1809, opened 2016-06-28
TRUNK-4898 Cannot resolve dependency azeckoski:reflectutils:0.9.14

teleivo:
## Description

org.azeckoski:reflectutils:jar: version 0.9.14 cannot be resolved which fails
the build, version: 0.9.20 is available on http://mavenrepo.openmrs.org/

* update pom.xml reflectutils version to 0.9.20

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4898

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -425,7 +425,7 @@
 			<dependency>
 				<groupId>org.azeckoski</groupId>
 				<artifactId>reflectutils</artifactId>
-				<version>0.9.14</version>
+				<version>0.9.20</version>
 			</dependency>
 			<dependency>
 				<groupId>org.springframework</groupId>


lluismf:
Do we really need this dependency?

teleivo:
@lluismf don't know, was thee before, broke the build because of the version that's why I updated it.

teleivo:
@lluismf it is used in https://github.com/openmrs/openmrs-core/blob/2.0.x/api/src/main/java/org/openmrs/util/Reflect.java

so its needed by the api project

dkayiwa:
@teleivo could this be related to? https://talk.openmrs.org/t/module-releases-missing-from-maven-repo/6980

teleivo:
@dkayiwa thanks, it was related! problem solved

PR #1810, opened 2016-06-29
TRUNK-4846: Concept search by ID displays double results 

geoff-wasilwa:
## Description

Changed the data structure that holds Concept Results from `List` to `Set` since `Set` only allows for unique results.

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4846

## Checklist:

- [x] My pull request is based on the latest 1.11.x branch
  `git pull --rebase upstream 1.11.x`.

### DIFF
--- a/web/src/main/java/org/openmrs/web/dwr/DWRConceptService.java
+++ b/web/src/main/java/org/openmrs/web/dwr/DWRConceptService.java
@@ -94,6 +94,7 @@ public class DWRConceptService {
 	 * @should not return concept by given id if datatype is not included
 	 * @should not return concept by given id if datatype is excluded
 	 * @should include
+	 * @should not return duplicates when searching by concept id
 	 * @since 1.8
 	 */
 	public List<Object> findBatchOfConcepts(String phrase, boolean includeRetired, List<String> includeClassNames,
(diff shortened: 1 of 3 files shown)


dkayiwa:
@geoff-wasilwa shouldn't the fix be in the data access layer?

geoff-wasilwa:
@dkayiwa the `DWRConceptService` actually calls the data layer twice (see line [144](https://github.com/geoff-wasilwa/openmrs-core/blob/feb47d3570f0e9a46420bda9d649419bf179ed6a/web/src/main/java/org/openmrs/web/dwr/DWRConceptService.java#L144) and [199](https://github.com/geoff-wasilwa/openmrs-core/blob/feb47d3570f0e9a46420bda9d649419bf179ed6a/web/src/main/java/org/openmrs/web/dwr/DWRConceptService.java#L199)) and the results of both calls are added to the collection.

dkayiwa:
@geoff-wasilwa so in that case, shouldn't the fix be to simply remove the extra redundant call?

geoff-wasilwa:
@dkayiwa I wouldn't say it's redundant, the first call tries to find the concept by id (if the search phrase contains digits), while the next call tries to find any concept name matching the phrase. So that, when searching for `1000` and you have a concept with id `1000` and a concept name with the string `1000` you get both as results.

PR #1815, opened 2016-07-09
TRUNK-4720 Encounter search will not match on id unless it contains digits

blavigne13:
## Description
Previously, the code made a decision on whether the query was a name or identifier by using a regex to determine if the query contained one or more digits, then created the appropriate PatientSearchCriteria. In cases of identifiers without digits, or names with them, it was not possible to obtain the correct result. Now, both names and identifiers are searched, without attempting to categorize the query beforehand.

Behavior has also been slightly changed: An exact match is now required for identifier to produce a result. Note: this only applies to identifier; name will still match on partial search terms. This is consistent with how other "name or identifier" queries are handled (e.g.: Find Patient, Find Provider) and is presumably for performance reasons.

Two test cases were added to ensure that results are returned on substring matches for patient name, and not patient identifier.

## Related Issue

## Checklist:

…igits

### DIFF
--- a/api/src/test/java/org/openmrs/api/EncounterServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/EncounterServiceTest.java
@@ -2959,4 +2959,30 @@ public class EncounterServiceTest extends BaseContextSensitiveTest {
 		
 		assertEquals("Two New Order Groups Get Saved", 2, orderGroups.size());
 	}
+
+	/**
+	 * @see EncounterService#getEncountersByPatient(String,boolean)
+	 */
+	@Test
+	@Verifies(value = "get all unvoided encounters for the given substring of patient names", method = "getEncountersByPatient(String,boolean)")
+	public void getEncountersByPatient_shouldGetAllUnvoidedEncountersForTheGivenSubstringOfPatientNames() throws Exception {
+		EncounterService encounterService = Context.getEncounterService();
+		
+		List<Encounter> encounters = encounterService.getEncountersByPatient("Joh", false);
+		assertEquals(3, encounters.size());
+	}
+	
+	/**
+	 * @see EncounterService#getEncountersByPatient(String,boolean)
+	 */
+	@Test
+	@Verifies(value = "not get encounters for the given substring of patient identifiers", method = "getEncountersByPatient(String,boolean)")
+	public void getEncountersByPatient_shouldNotGetEncountersForTheGivenSubstringOfPatientIdentifiers() throws Exception {
+		EncounterService encounterService = Context.getEncounterService();
+		
+		List<Encounter> encounters = encounterService.getEncountersByPatient("123", false);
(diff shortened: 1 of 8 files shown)


dkayiwa:
@blavigne13 the tests that you have added pass even without your changes. Generally, you need to start with tests that fail. Then make changes to make the tests pass.

dkayiwa:
@blavigne13 do you mind squashing the commits into one?

dkayiwa:
@blavigne13 how is this going? Do you have any blockers where you may need help?

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -2959,4 +2959,30 @@ public void shouldSaveMultipleOrderGroupsIfDifferentOrdersHaveDifferentOrderGrou
 		
 		assertEquals("Two New Order Groups Get Saved", 2, orderGroups.size());
 	}
+
+	/**
+	 * @see EncounterService#getEncountersByPatient(String,boolean)
+	 */
+	@Test
+	@Verifies(value = "get all unvoided encounters for the given substring of patient names", method = "getEncountersByPatient(String,boolean)")
+	public void getEncountersByPatient_shouldGetAllUnvoidedEncountersForTheGivenSubstringOfPatientNames() throws Exception {
```
wluyima: I believe there is already tests that test partial name searches

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -2959,4 +2959,30 @@ public void shouldSaveMultipleOrderGroupsIfDifferentOrdersHaveDifferentOrderGrou
...
+	}
+	
+	/**
+	 * @see EncounterService#getEncountersByPatient(String,boolean)
+	 */
+	@Test
+	@Verifies(value = "not get encounters for the given substring of patient identifiers", method = "getEncountersByPatient(String,boolean)")
+	public void getEncountersByPatient_shouldNotGetEncountersForTheGivenSubstringOfPatientIdentifiers() throws Exception {
+		EncounterService encounterService = Context.getEncounterService();
+
```
wluyima: Would be nice to have an assertion here that ensures there is actually at least one patient with one non voided identifier that matches the search text. Otherwise this could be passing just because there is no match in the test dataset

wluyima:
Why does this PR contain changes to do with BinaryDataHandler?

wluyima:
I also don't see a test that tests that a patient can be found where the search text matches their identifier when it contains no digit, technically this test should be failing before your fix

dkayiwa:
@blavigne13 do you still have interest in finishing this? :)

dkayiwa:
@blavigne13 this pull request has more commits than it should. Do you mind creating a new one as you take advantage of these tips? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

PR #1816, opened 2016-07-14
TRUNK-4901 Scheduled orders cannot be discontinued

teleivo:
## Description

* add Order.isActivated()
* add Order.isActivated(Date)
* fix issues introduced by switching isFuture with isStarted/!isStarted for
orders scheduled in the future
  * Order.isActive(Date) of stopped orders scheduled for the future were considered active
  * Order.isDiscontinued(Date) of stopped orders scheduled for the future which were not considered discontinued
* add OrderService tests to ensure we can revise orders scheduled in the
future and that the original orders are considered discontinued
* add OrderService tests to ensure we can discontinue orders scheduled in the
future

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4901

excuse the formatting noise, I did use the OpenMRS formatter in this repo, seems like it wasnt always used before

### DIFF
--- a/api/src/main/java/org/openmrs/Order.java
+++ b/api/src/main/java/org/openmrs/Order.java
@@ -363,6 +363,38 @@ public class Order extends BaseOpenmrsData implements java.io.Serializable {
 	}
 	
 	/**
+	 * Convenience method to determine if the order is activated as of the current date
+	 * 
+	 * @return boolean indicating whether the order was activated before or on the current date
+	 * @since 2.0
+	 * @see #isActivated(java.util.Date)
+	 */
+	public boolean isActivated() {
+		return isActivated(new Date());
+	}
+	
+	/**
+	 * Convenience method to determine if the order is activated as of the specified date
+	 * 
+	 * @param checkDate - the date on which to check order. if null, will use current date
+	 * @return boolean indicating whether the order was activated before or on the check date
+	 * @since 2.0
+	 * @should return true if an order was activated on the check date
+	 * @should return true if an order was activated before the check date
+	 * @should return false if dateActivated is null
+	 * @should return false for an order activated after the check date
+	 */
+	public boolean isActivated(Date checkDate) {
+		if (dateActivated == null) {
+			return false;
+		}
+		if (checkDate == null) {
+			checkDate = new Date();
+		}
+		return OpenmrsUtil.compare(dateActivated, checkDate) <= 0;
+	}
+	
+	/**
 	 * Convenience method to determine if the order was active as of the current date
 	 * 
 	 * @since 1.10.1
@@ -381,6 +413,7 @@ public class Order extends BaseOpenmrsData implements java.io.Serializable {
 	 * @should return true if an order expired on the check date
 	 * @should return true if an order was discontinued on the check date
 	 * @should return true if an order was activated on the check date
+	 * @should return true if an order was activated on the check date but scheduled for the future
 	 * @should return false for a voided order
 	 * @should return false for a discontinued order
 	 * @should return false for an expired order
@@ -395,7 +428,7 @@ public class Order extends BaseOpenmrsData implements java.io.Serializable {
 			checkDate = new Date();
 		}
 		
-		return isStarted(checkDate) && !isDiscontinued(checkDate) && !isExpired(checkDate);
+		return isActivated(checkDate) && !isDiscontinued(checkDate) && !isExpired(checkDate);
 	}
 	
 	/**
@@ -452,6 +485,8 @@ public class Order extends BaseOpenmrsData implements java.io.Serializable {
 	 * @should fail if date stopped is after auto expire date
 	 * @should return true if check date is after date stopped but before auto expire date
 	 * @should return true if check date is after both date stopped auto expire date
+	 * @should return true if the order is scheduled for the future and activated on check date but
+	 *         the check date is after date stopped
 	 */
 	public boolean isDiscontinued(Date checkDate) {
 		if (dateStopped != null && autoExpireDate != null && dateStopped.after(autoExpireDate)) {
@@ -463,7 +498,7 @@ public class Order extends BaseOpenmrsData implements java.io.Serializable {
 		if (checkDate == null) {
 			checkDate = new Date();
 		}
-		if (dateActivated == null || !isStarted(checkDate) || dateStopped == null) {
+		if (!isActivated(checkDate) || dateStopped == null) {
 			return false;
 		}
 		return checkDate.after(dateStopped);
@@ -505,7 +540,7 @@ public class Order extends BaseOpenmrsData implements java.io.Serializable {
 		if (checkDate == null) {
 			checkDate = new Date();
 		}
-		if (dateActivated == null || !isStarted(checkDate)) {
+		if (!isActivated(checkDate)) {
 			return false;
 		}
 		if (isDiscontinued(checkDate) || autoExpireDate == null) {
--- a/api/src/main/java/org/openmrs/Order.java
+++ b/api/src/main/java/org/openmrs/Order.java
@@ -34,6 +34,7 @@ import org.openmrs.util.OpenmrsUtil;
  */
 public class Order extends BaseOpenmrsData implements java.io.Serializable {
 	
+	
 	public static final long serialVersionUID = 4334343L;
 	
 	/**
@@ -94,9 +95,9 @@ public class Order extends BaseOpenmrsData implements java.io.Serializable {
 	private Date scheduledDate;
 	
 	/**
-	 * Allows the orders if ordered as an orderGroup,
-	 * to maintain a sequence of how members are added in the group
-	 * ex - for two orders of isoniazid and ampicillin, the sequence of 1 and 2 needed to be maintained
+	 * Allows the orders if ordered as an orderGroup, to maintain a sequence of how members are
+	 * added in the group ex - for two orders of isoniazid and ampicillin, the sequence of 1 and 2
+	 * needed to be maintained
 	 */
 	private Double sortWeight;
 	
@@ -775,8 +776,7 @@ public class Order extends BaseOpenmrsData implements java.io.Serializable {
 	}
 	
 	/**
-	 * @since 1.12
-	 * {@link org.openmrs.OrderGroup}
+	 * @since 1.12 {@link org.openmrs.OrderGroup}
 	 * @returns the OrderGroup
 	 */
 	public OrderGroup getOrderGroup() {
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/openmrs/Order.java
```diff
@@ -362,6 +363,41 @@ public void setCommentToFulfiller(String commentToFulfiller) {
 	}
 	
 	/**
+	 * Convenience method to determine if the order is activated as of the current date
+	 * 
+	 * @return boolean indicating whether the order was activated on the current date
```
wluyima: This doesn't sound exactly correct, I think this should say "indicating whether the order was activated before or on the current date"

api/src/main/java/org/openmrs/Order.java
```diff
@@ -362,6 +363,41 @@ public void setCommentToFulfiller(String commentToFulfiller) {
...
+	 * @since 2.0
+	 * @should return true if an order was activated on the current date
+	 * @should return true if an order was activated before the current date
+	 * @should return false for an order not activated
+	 * @should return false for an order activated after the current date
```
wluyima: How different is this from 'return false for an order not activated'

api/src/main/java/org/openmrs/Order.java
```diff
@@ -462,7 +501,7 @@ public boolean isDiscontinued(Date checkDate) {
 		if (checkDate == null) {
 			checkDate = new Date();
 		}
-		if (dateActivated == null || !isStarted(checkDate) || dateStopped == null) {
+		if (dateActivated == null || !isActivated(checkDate) || dateStopped == null) {
```
wluyima: isActivated() already checks if dateActivated  == null, so you don't need to check if it is null again here

api/src/main/java/org/openmrs/Order.java
```diff
@@ -504,7 +543,7 @@ public boolean isExpired(Date checkDate) {
 		if (checkDate == null) {
 			checkDate = new Date();
 		}
-		if (dateActivated == null || !isStarted(checkDate)) {
+		if (dateActivated == null || !isActivated(checkDate)) {
```
wluyima: Same here

wluyima:
The PR seems to have a lot of auto format changes, did you set the openmrs formatter in your IDE?

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -985,16 +980,16 @@ public OrderType getOrderTypeByConcept(Concept concept) {
 	public List<Concept> getTestSpecimenSources() {
 		return getSetMembersOfConceptSetFromGP(OpenmrsConstants.GP_TEST_SPECIMEN_SOURCES_CONCEPT_UUID);
 	}
-
+	
 	@Override
 	public Concept getNonCodedDrugConcept() {
 		String conceptUuid = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_DRUG_ORDER_DRUG_OTHER);
-		if(org.apache.commons.lang.StringUtils.isNotBlank(conceptUuid)){
+		if (org.apache.commons.lang.StringUtils.isNotBlank(conceptUuid)) {
```
wluyima: Can you update this to use the StringUtils from spring since it is already imported in to the class, I think the equivalent method to call is StringUtils.hasText()

api/src/test/java/org/openmrs/OrderTest.java
```diff
@@ -511,6 +514,32 @@ public void isDiscontinued_shouldReturnTrueIfCheckDateIsAfterBothDateStoppedAuto
...
+	@Test
+	public void isDiscontinued_shouldReturnTrueIfTheOrderIsScheduledForTheFutureAndActivatedOnCheckDateButTheCheckDateIsAfterDateStopped()
+	        throws Exception {
+		// tests the case when a scheduled order is revised:
+		// in that case its original order is stopped.
+		// the stopped date of the original order is set to a moment before the activated date of the revised order.
+		// the order here represents such an original order
+		Order order = new Order();
+		order.setUrgency(Urgency.ON_SCHEDULED_DATE);
+		Date scheduledDateInFuture = DateUtils.addMonths(new Date(), 2);
```
wluyima: Can you add a local variable and assign new Date() to it? So that you re-use it instead of creating multiple instances

api/src/test/java/org/openmrs/OrderTest.java
```diff
@@ -632,6 +662,112 @@ public void isExpired_shouldReturnTrueIfDateStoppedIsNullAndAutoExpireDateIsBefo
 	}
 	
 	/**
+	 * @see Order#isActivated()
+	 * @verifies return true if an order was activated on the current date
+	 */
+	@Test
+	public void isActivated_shouldReturnTrueIfAnOrderWasActivatedOnTheCurrentDate() throws Exception {
+		Order order = new Order();
+		order.setDateActivated(new Date());
```
wluyima: This test doesn't seem correct to me, to guarantee that the dates are the same, you have to create one instance of date and reuse it as dateActivated and checkDate. Otherwise the only reason this test is passing is because by the time isActivated() is called, a new instance of Date is created a couple of milliseconds later which puts checkDate always after dateActivated, feel free to print the dates when comparing them and you will realize this.

wluyima: In theory, you should not duplicate these tests for both isActivated() and isActivated(checkDate), all tests should be against isActivated(checkDate) except for one where you test that when isActivated() is called it defaults to now

wluyima:
If you can't set up the openmrs formatter, you can make 2 commits in your PR where one contains the auto format changes while the other only has your changes otherwise it is hard for a reviewer to know what your code changes are. You can always squash them before merging the code

api/src/main/java/org/openmrs/Order.java
teleivo: @wluyima :+1:

api/src/main/java/org/openmrs/Order.java
teleivo: :+1:

api/src/main/java/org/openmrs/Order.java
teleivo: :+1:

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
teleivo: @wluyima should I? This is not part of the issue but if you insist I will.

api/src/test/java/org/openmrs/OrderTest.java
teleivo: sure

api/src/main/java/org/openmrs/Order.java
teleivo: @wluyima tests case wher dateActivated is null

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
teleivo: @wluyima done

api/src/main/java/org/openmrs/Order.java
wluyima: Or you can say "indicating whether the order was activated as of the current date"

api/src/main/java/org/openmrs/Order.java
wluyima: Then I think the wording needs to be clearer, one should say when date activated is null and the other should say when date activated is after the check date

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
wluyima: Thanks!

wluyima: You don't have to but it would be nice to clean it up, the two method calls are equivalent but the with the second you won't have that 'ugly' fully qualified java class name

wluyima:
Are you going to address the auto format changes?

teleivo:
@wluyima please read "excuse the formatting noise, I did use the OpenMRS formatter in this repo, seems like it wasnt always used before"

teleivo:
@wluyima I am using the formatter of this repo in eclipse. and I just ran mvn clean package, formatter didnt change a thing. look at the OrderService on how it was formatted before, its such a pain in the neck!

wluyima:
Can you instead make 2 commits? This is what you would need to do, create a new branch off the 2.0.x branch, format the classes you've touched in this PR and commit, then merged this PR's commit into that the new branch and commit, the diff for the last commit will only contain your changes  excluding auto format changes in the first, then create a new pull request for the new branch

teleivo:
@wluyima done.

wluyima:
Looks much better! thanks!

api/src/main/java/org/openmrs/Order.java
```diff
@@ -363,6 +363,38 @@ public void setCommentToFulfiller(String commentToFulfiller) {
 	}
 	
 	/**
+	 * Convenience method to determine if the order is activated as of the current date
+	 * 
+	 * @return boolean indicating whether the order was activated before or on the current date
+	 * @since 2.0
+	 * @see #isActivated(java.util.Date)
```
wluyima: You should have should to test that this method defaults to now as the check date

teleivo: @wluyima as I already told you isActive(), isActive(null), isStarted() and isStarted(null) don't have the test you are asking me to write. Tell me how to test this and I will.

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -988,7 +984,7 @@ public OrderType getOrderTypeByConcept(Concept concept) {
 	@Override
 	public Concept getNonCodedDrugConcept() {
 		String conceptUuid = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_DRUG_ORDER_DRUG_OTHER);
-		if(org.apache.commons.lang.StringUtils.isNotBlank(conceptUuid)){
+		if (StringUtils.hasText(conceptUuid)) {
```
wluyima: 👍

api/src/main/java/org/openmrs/Order.java
teleivo: @djazayeri can I get your opinion on this topic. I think @wluyima and I disagree here. he wants a test for case `isActive(null)` which leads checkDate to default to `new Date()`. This hasnt been tested in any method in this class (isActive(), isActive(null), isStarted() and isStarted(null) are not tested). Not saying this is an argument for not doing it but I just dont know how. I would be happy to do so but need more instruction.

djazayeri: I wouldn't write this test myself, but I can see the argument for doing so. @wluyima can decide whether this is required or not.

teleivo: @wluyima if I should do so, can you please be specific in how I should do this? Extracting new Date() into an interface method so it can be mocked during testing? Yes, no? Is this worth the effort?

wluyima: As Darius mentioned, it is not much of a big requirement to have the test and If it's going to take much of your time to write it then I guess you can ignore it.

wluyima:
Merged at c35550b6dde0ecc2ad40be7b3e2d2b0fb4fd83dc and db99602df4e9b0e99ed426676186506467d2a423

teleivo:
@wluyima should I close it? since its merged

PR #1821, opened 2016-08-09
TRUNK-4903 Annotate Order classes

teleivo:
## Description

* added mappings for Order, TestOrder and DrugOrder via annotations
* deleted the Order.hbm.xml
* add dateChanged and changedBy to the ImmutableOrderInterceptor otherwise it
fails since these properties change when for ex. unvoiding an Order
* add date_changed and changed_by to orders table

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4903


api/src/main/java/org/openmrs/DrugOrder.java
```diff
@@ -10,52 +10,88 @@
...
+@Entity
+@Table(name = "drug_order")
+@PrimaryKeyJoinColumn(name="order_id")
 public class DrugOrder extends Order implements java.io.Serializable {
 
 	public static final long serialVersionUID = 72232L;
 
 	// Fields
 
+	// TODO - Order.hbm.xml says length is 22 but length is only used for strings. Shouldnt we use precision?
```
dkayiwa: I think length was an error resulting from a copy and paste of a string type element. So it can be removed.

dkayiwa:
If it was not for requiring to add new un used columns to the orders table, i would be merging this. :)

teleivo:
@dkayiwa I know, its sad :(
but maybe a discussion about this is needed. this issue + https://talk.openmrs.org/t/order-datechanged-changedby-causing-exception-in-immutableorderinterceptor/7559/5 feels fishy.
https://talk.openmrs.org/t/trunk-4730-complete-list-of-domain-objects-missing-changed-by-or-date-changed/3419/17

wluyima:
@dkayiwa that's not the only concern by the way, what happens to existing modules that extend the orders table via XML? The problematic annotations happen to be in a superclass, there is other inherent issues with having annotations on super classes in openmrs which is something against which I've had a bias and in general I don't want us ending up with non standard JPA annotations in the API, this is something that we should possibly discuss on the design call, it has several implications and I personally think to truly use JPA annotations in the API first we should break up the Auditable interface so that creator and createdBy setters/getters belong to their own interface while dateChanged and changedBy also belong to a separate one so that a domain object can implement just one of the interfaces without the baggage of unused columns from the other interface like in the case of Orders. I think this splitting up of the interface was best suited to be done in 2.0 when we had major changes, there is great value in doing it going forward.

dkayiwa:
Ok. Then in that regard, am closing this for now until when we have agreed on the way forward.

teleivo:
@wluyima I totally agree with you that the dateChanged, changedBy should go to another interface since obviously they are not need in Order.

on the other hand you are wrong in that this PR shows only imports of javax.persistence which I assume is JPA and not hibernate.

+ I tested this change deploying it and on top of it a module with a class (configured via xml mappings) which extends TestOrder, without issue.

so please also keep that in mind in your argumentation since I already have pointed those things out in lenght.

I do agree with you that the two members/columns if not wanted should not be inherited by Order. And I also agree that it seems that we need a design call for this :+1: 

Thank you guys!!

wluyima:
@teleivo I'm not saying that you used the non standard annotations, It was something I mentioned as a by the way as a potential pitfall when using annotations in general somewhere else. If modules extending the orders table still work fine, that's great!

PR #1825, opened 2016-08-30
TRUNK-4922 ConceptDAO filters concept classes only if results were found

Schwasrz94:
## Description

The skipSame-method in LuceneQuery generates a new instance of
fullTextQuery even if there are no search results. In this way all
concepts will be returned without filtering them.

* only call buildQuery() in skipSame when results were found
* add test to ConceptServiceTest

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4922

### DIFF
--- a/api/src/test/java/org/openmrs/api/ConceptServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/ConceptServiceTest.java
@@ -2756,6 +2756,20 @@ public class ConceptServiceTest extends BaseContextSensitiveTest {
 	 * @see ConceptService#getConcepts(String, List, boolean, List, List, List, List, Concept, Integer, Integer)
 	 */
 	@Test
+	@Verifies(value = "should return empty list if no concept within specified classes was found", method = "getConcepts(String,List<QLocale;>,boolean,List<QConceptClass;>,List<QConceptClass;>,List<QConceptDatatype;>,List<QConceptDatatype;>,Concept,Integer,Integer)")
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/openmrs/api/ConceptServiceTest.java
```diff
@@ -2756,6 +2756,20 @@ public void getConcepts_shouldReturnConceptsWithSpecifiedClasses() throws Except
 	 * @see ConceptService#getConcepts(String, List, boolean, List, List, List, List, Concept, Integer, Integer)
 	 */
 	@Test
+	@Verifies(value = "should return empty list if no concept within specified classes was found", method = "getConcepts(String,List<QLocale;>,boolean,List<QConceptClass;>,List<QConceptClass;>,List<QConceptDatatype;>,List<QConceptDatatype;>,Concept,Integer,Integer)")
+	public void getConcepts_shouldReturnEmptyListIfNoConceptWithinSpecifiedClassesWasFound() throws Exception {
```
wluyima: Does this test fail without your changes?

teleivo: @wluyima yes it does

wluyima: Ok, thanks!

teleivo:
@wluyima what can we do to get this merged?

PR #1835, opened 2016-09-16
Preethi | TRUNK-4944 | Fixed obs and group members getting dirty when…

preethi29:
TRUNK-4944
## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4944

## Checklist:

… saved without any change

### DIFF
--- a/api/src/test/java/org/openmrs/api/ObsServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/ObsServiceTest.java
@@ -1878,18 +1878,27 @@ public class ObsServiceTest extends BaseContextSensitiveTest {
 	 */
 	@Test
 	public void saveObs_shouldNotVoidAnObsWithNoChanges() throws Exception {
-		executeDataSet(INITIAL_OBS_XML);
+		executeDataSet(ENCOUNTER_OBS_XML);
 		ObsService os = Context.getObsService();
-		Obs obs = os.getObs(2);
-		assertFalse(obs.getGroupMembers(true).isEmpty());
+		Obs obs = os.getObs(14);
+		Set<Obs> groupMembersBeforeSave = obs.getGroupMembers(true);
+		assertFalse(groupMembersBeforeSave.isEmpty());
 		assertFalse(obs.getGroupMembers(false).isEmpty());
 		assertFalse(obs.isDirty());
 		Set<Obs> originalMembers = new HashSet<>(obs.getGroupMembers());
 		for (Obs o : originalMembers) {
 			assertFalse(o.isDirty());
 		}
+
 		Obs saveObs = os.saveObs(obs, "no change");
+
 		assertEquals(obs, saveObs);
 		assertFalse(saveObs.getVoided());
+		Set<Obs> savedMembers = new HashSet<>(saveObs.getGroupMembers());
+		assertFalse(saveObs.isDirty());
+		for (Obs o : savedMembers) {
+			assertFalse(o.isDirty());
+		}
+
 	}
 }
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -139,9 +139,11 @@ public Obs saveObs(Obs obs, String changeMessage) throws APIException {
 							Obs replacement = os.saveObs(o, changeMessage);
 							//The logic in saveObs evicts the old obs instance, so we need to update
 							//the collection with the newly loaded and voided instance
-							toRemove.add(o);
-							toAdd.add(os.getObs(o.getId()));
-							toAdd.add(replacement);
+							if(!replacement.equals(o)) {
```
wluyima: Nice catch!

wluyima: Do you mind adding a test for the situation where this condition would pass i.e when one of child Obs is edited and and check if the parent still comes back not dirty, or should it be marked as dirty too? This is a little interesting

preethi29: Ideally parent should not be made dirty in that situation ryt?

preethi29: If parent is made dirty, then during the next save, the entire hierarchy would be voided and created again!

preethi29: @wluyima So the question will be, whether parent obs has to be made dirty when a groupmember is added or removed?

wluyima: The current logic is such that if you add a new member or remove an existing one, the parent is marked as dirty, my guess is that if you edit one of the child obs the parent is gonna come back as dirty because of [these calls](https://github.com/openmrs/openmrs-core/blob/2.0.0/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java#L148-L153) and I'm not sure if this should be the case.

preethi29: should this be moved to talk then?

wluyima: In theory, my take is that if you add a new member or remove an existing one from a parent, the parent should  be marked as dirty. On the other hand, editing a child without changes to the parent should not make the parent dirty  even though technically we replace the voided obs with the new one. May be we should ask this on talk and see what others think

preethi29: So when the edited obs is removed from the group members of the parent, the obsGroup field is made null. Does that mean obs_group_id column is made null for the old obs? When is it saved to the db? I see no save operation of the edited obs after the obs group is made null.
https://github.com/openmrs/openmrs-core/blob/2.0.0/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java#L155

wluyima: When the session is flushed, hibernate is the one that takes care of unsetting the foreign key column for the child and when it's removed from its parent. What you're trying to ask I think was when does the parent get saved to flush changes to the collection since we don't call save, again I believe hibernate will auto flush the changes to the parent when the session is flushed

preethi29: okay

wluyima: But I guess it wouldn't hurt to explicitly dao.saveObs for the parent :)

PR #1836, opened 2016-09-17
TRUNK-4941 Add ModuleUtil.isOpenmrsVersionInVersions()

teleivo:
## Description

added ModuleUtil.isOpenmrsVersionInVersions() but need review and help in how to best write the tests that I annotated with '@Ignore'

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4941

### DIFF
--- a/api/src/main/java/org/openmrs/module/ModuleUtil.java
+++ b/api/src/main/java/org/openmrs/module/ModuleUtil.java
@@ -50,6 +50,7 @@ import org.openmrs.api.context.Context;
 import org.openmrs.api.context.ServiceContext;
 import org.openmrs.scheduler.SchedulerUtil;
 import org.openmrs.util.OpenmrsClassLoader;
+import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.OpenmrsUtil;
 import org.springframework.context.support.AbstractRefreshableApplicationContext;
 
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -211,6 +212,35 @@ public static File insertModuleFile(InputStream inputStream, String filename) {
...
 	}
+
+	/**
+	 * Checks if the current OpenMRS version is in an array of versions.
+	 * <p>
+	 * This method calls {@link ModuleUtil#matchRequiredVersions(String, String)} internally.
+	 * </p>
+	 *
+	 * @param versions the openmrs versions to be checked against the current openmrs version
+	 * @return true if the current openmrs version is in versions and false otherwise
```
wluyima: How phrasing this as "return true if the current openmrs version is in versions otherwise false"?

api/src/test/java/org/openmrs/module/ModuleUtilTest.java
```diff
@@ -79,7 +82,58 @@ public void getMandatoryModules_shouldReturnMandatoryModuleIds() throws Exceptio
...
+			throws Exception {
+
+		Field versionField = OpenmrsConstants.class.getDeclaredField("OPENMRS_VERSION_SHORT");
+		Field modifiersField = Field.class.getDeclaredField("modifiers");
+		modifiersField.setAccessible(true);
+		modifiersField.setInt(versionField, versionField.getModifiers() & ~Modifier.FINAL);
+		final String currentVersion = "1.9.8";
+		versionField.set(null, currentVersion);
+
+		Assert.assertTrue(ModuleUtil.isOpenmrsVersionInVersions( currentVersion, "1.10.*"));
```
wluyima: Shouldn't this be false? Because 1.9.8 isn't matched by the expression **1.10.***

api/src/main/java/org/openmrs/module/ModuleUtil.java
teleivo: @wluyima should I still rephrase? or leave it, since you already merged the PR.

teleivo:
@wluyima thanks!

PR #1840, opened 2016-09-21
TRUNK-4937 | Save visit should not cascade save encounter

preethi29:
## Related Issue

## Checklist:

### DIFF
--- a/api/src/test/java/org/openmrs/api/VisitServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/VisitServiceTest.java
@@ -11,12 +11,14 @@ package org.openmrs.api;
 
 import junit.framework.Assert;
 import org.apache.commons.lang.StringUtils;
+import org.hibernate.TransientObjectException;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.openmrs.Concept;
 import org.openmrs.Encounter;
 import org.openmrs.Location;
+import org.openmrs.Obs;
 import org.openmrs.Patient;
 import org.openmrs.Visit;
 import org.openmrs.VisitAttribute;
@@ -43,6 +45,7 @@ import java.util.Map;
 import static junit.framework.Assert.assertEquals;
 import static junit.framework.Assert.assertNotNull;
 import static junit.framework.Assert.assertTrue;
+import static junit.framework.Assert.assertNull;
 
 /**
  * Tests methods in the {@link VisitService}
@@ -962,33 +965,25 @@ public class VisitServiceTest extends BaseContextSensitiveTest {
 	/**
 	 * @see VisitService#saveVisit(Visit)
 	 */
-	@Test
-	@Verifies(value = "save visit should persist new encounter", method = "saveVisit(Visit)")
-	public void saveVisit_shouldPersistNewEncounter() throws Exception {
+	@Test(expected = TransientObjectException.class)
+	@Verifies(value = "save visit should not persist new encounter", method = "saveVisit(Visit)")
+	public void saveVisit_shouldNotPersistNewEncounter() throws Exception {
 		
 		VisitService vs = Context.getVisitService();
 		Visit visit = vs.getVisit(1);
 		
 		Encounter encounter = new Encounter();
 		encounter.setEncounterDatetime(new Date());
+		encounter.setEncounterType(Context.getEncounterService().getEncounterType(1));
 		encounter.setPatient(visit.getPatient());
 		encounter.setLocation(visit.getLocation());
 		visit.addEncounter(encounter);
 		
 		vs.saveVisit(visit);
-		
+
 		Context.flushSession();
-		Context.clearSession();
-		
-		Integer encounterId = encounter.getEncounterId();
-		
-		// reload the visit
-		visit = Context.getVisitService().getVisit(1);
-		
-		assertEquals(1, visit.getEncounters().size());
-		assertEquals(encounterId, ((Encounter) visit.getEncounters().toArray()[0]).getEncounterId());
 	}
-	
+
 	/**
 	 * @see VisitService#getAllVisitTypes(boolean)
 	 */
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/openmrs/api/VisitServiceTest.java
```diff
@@ -963,32 +965,74 @@ public void saveVisit_shouldAssociateEncounterWithVisitOnSaveEncounter() throws
...
+		assertEquals(0, encounter.getAllObs(true).size());
+		int obsBeforeCount = encounter.getAllObs(true).size();
+		Obs obs = new Obs();
+		obs.setConcept(Context.getConceptService().getConcept(19));
+		obs.setValueText("new obs");
+		encounter.addObs(obs);
+
+		vs.saveVisit(visit);
+
+		assertNull(encounter.getObs().iterator().next().getObsId()); // should not have saved the encounter
```
wluyima: Shouldn't this comment say 'should not have saved the encounter'?

wluyima: Why not change a simpler field on the encounter e.g location instead of the collection of Obs, flush and clear the session and then check that the old location is still set to the old one?

preethi29: sure will update it

preethi29: But when we set the location to encounter and flush the session, location will be saved to encounter since encounter is attached to the session

wluyima: What I'm saying is that if encounter.location is set to location A, change it to location B, save the visit and make sure that the encounter location is still set to location A and not B after the flush and clear

preethi29: I understand what you are saying but when I flush the session, even though encounter is not saved via VisitService.saveVisit, the location change will get flushed to the DB because encounter is attached to the session.

wluyima: May be we're trying to solve something that is unsolvable, hibernate will always flush everything in the current session when it gets flushed, in this case adding an encounter to a visit attaches it to the current session because it then belongs to a persistent collection which makes it vulnerable to automatic flushing.

preethi29: which means the same will be applied to obs as well!

wluyima: The cascade attribute seems almost useless for collections except may be for the delete operation

preethi29: It will also be useful on the create operation since object wont be attached to the session

dkayiwa:
@preethi29 did you notice the above travis failure? :)

preethi29:
@dkayiwa can I retrigger it? It was failing last week because openmrs mavne repo was slow

dkayiwa:
Oh i see! Then no problem. :)

api/src/test/java/org/openmrs/api/VisitServiceTest.java
preethi29: @wluyima Any update on this PR?

wluyima: I think we should have tests for only the situations where the cascade option makes e.g delete and not worry about the others

preethi29: But a visit cant be deleted if there are encounters associated to it. So we cant really test that scenario.
Also cascade option would be applicable for the case when a new encounter is added to visit.

wluyima: When you delete a parent with cascading disabled, my assumption is that hibernate will delete the parent and just drop the references to it from the oprhaned children meaning the children stick around standing on their own and we need a test to prove this behavior. Choosing to tell hibernate to delete vs keep 'orphan' children is purely based on real world semantics i.e can an encounter exist without a visit? Based on our data model this is true where as for certain objects like ConceptName that can't exist on their own based on our data model we explicitly tell hibernate to delete them when they become orphans.

As for saving a parent with a new child object, disabling the cascade would result in a hibernate exception complaining about a reference to a transient associated object so yes it might be worth adding a test that ensures this is happening.

wluyima: If the VisitService doesn't allow it at least that is our own custom logic and that's fine

api/src/test/java/org/openmrs/api/VisitServiceTest.java
```diff
@@ -963,32 +965,36 @@ public void saveVisit_shouldAssociateEncounterWithVisitOnSaveEncounter() throws
...
 		Visit visit = vs.getVisit(1);
 		
 		Encounter encounter = new Encounter();
 		encounter.setEncounterDatetime(new Date());
+		encounter.setEncounterType(Context.getEncounterService().getEncounterType(1));
 		encounter.setPatient(visit.getPatient());
 		encounter.setLocation(visit.getLocation());
 		visit.addEncounter(encounter);
 		
 		vs.saveVisit(visit);
```
wluyima: Are you sure this test doesn't pass before your changes? I think it passed since you're not flushing the the session before check the encounter id.

I think the test needs to ensure that when the visit is flushed an exception is thrown since the flushed visit would be referencing a transient encounter

preethi29: Makes sense. Will update the pull request.

wluyima: Thanks, I see you've already implemented it, are you done?

preethi29: Yes updated few minutes back.

wluyima: So in theory these changes imply that new encounters have to be saved prior to being added to a visit, right?

wluyima: As in an existing visit, it should be fine if the visit is also new

preethi29: Be it new visit or existing visit, encounters have to be saved first. In fact, the test is written using existing visit.

PR #1843, opened 2016-09-23
TRUNK-4939:VisitValidator should ensure that the visit start date can…

MichalSlawin:
## Description

Added two new validate methods and test for them. Also changed already existing tests so they could pass validation.

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4939

## Checklist:

…'t be before the birth date of the patient


api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -159,6 +167,50 @@ private void validateStopDatetime(Visit visit, Visit otherVisit, Errors errors)
...
 		}
-		
+
+	}
+
+	private void validateStartDateTimeToBirthdate(Visit visit, Errors errors) {
+
+		Patient patient = visit.getPatient();
+
+		if (patient != null) {
```
ssmusoke: @dkayiwa Are there any visits without patients? Should this not fail with an error?

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -159,6 +167,50 @@ private void validateStopDatetime(Visit visit, Visit otherVisit, Errors errors)
...
+						"This visit has a start date before birth date of the patient.");
+			}
+		}
+	}
+
+	private void validateStartDateTimeToEstimatedBirthdate(Visit visit, Errors errors) {
+
+		Patient patient = visit.getPatient();
+
+		if (patient != null) {
```
ssmusoke: @dkayiwa Are there visits without a patient? Should this not fail with an error?

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -98,6 +102,10 @@ public void validate(Object target, Errors errors) {
 		
 		// check attributes
 		super.validateAttributes(visit, errors, Context.getVisitService().getAllVisitAttributeTypes());
+
+		validateStartDateTimeToBirthdate(visit, errors);
+
+		validateStartDateTimeToEstimatedBirthdate(visit, errors);
```
wluyima: If birthdate is not estimated, I'd think that you don't have to validate it again here since you've already validated it above

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -159,6 +167,50 @@ private void validateStopDatetime(Visit visit, Visit otherVisit, Errors errors)
...
+
+		if (patient != null) {
+			if (visit.getStartDatetime() != null && patient.getBirthdate() != null && patient.getBirthdateEstimated()) {
+
+				Calendar calendar = Calendar.getInstance();
+				calendar.setTime(patient.getBirthdate());
+
+				long ageInMillis = Calendar.getInstance().getTimeInMillis() - calendar.getTimeInMillis();
+				long ageInYears = (ageInMillis / 86400000) / 365;
+				int ageInYearsInt = (int) ageInYears;
```
wluyima: Won't converting age to years throw off your next computations for the upper bound?

wluyima: Since you are discarding smaller age components when you cast to int

api/src/test/java/org/openmrs/api/PatientServiceTest.java
```diff
@@ -600,6 +600,11 @@ public void mergePatients_shouldChangeUserRecordsOfNonPreferredPersonToPreferred
 		executeDataSet(USERS_WHO_ARE_PATIENTS_XML);
 		Patient notPreferred = patientService.getPatient(2);
 		voidOrders(Collections.singleton(notPreferred));
+
+		Calendar calendar = Calendar.getInstance();
+		calendar.set(1950, Calendar.FEBRUARY, 2);
+		patientService.getPatient(6).setBirthdate(calendar.getTime());
```
wluyima: Can't you just set the birthdate in the test dataset?

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -469,4 +466,42 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() throws Ex
...
+		Visit visit = makeVisit(42);
+		visit.setStartDatetime(calendar.getTime());
+
+		visit.setPatient(Context.getPatientService().getPatient(7));
+
+		Errors errors = new BindException(visit, "visit");
+		new VisitValidator().validate(visit, errors);
+
+		assertTrue(errors.hasFieldErrors("startDatetime"));
+	}
```
wluyima: It would be nice to have more tests for estimated birthdate, i.e where the patient is less than 2yrs old and also one that ensures that the validation passes for a valid visit start date when birth date is estimated

api/src/main/java/org/openmrs/validator/VisitValidator.java
MichalSlawin: Should I remove "Patient patient = visit.getPatient();" from my methods and add it to VisitValidator.validate? I would be able to check if birthdate is estimated or not and call suitable function.

api/src/test/java/org/openmrs/api/PatientServiceTest.java
MichalSlawin: I don' t know how to find the patient with id=6. It' s not in PatientServiceTest-usersWhoArePatients.xml

api/src/main/java/org/openmrs/validator/VisitValidator.java
MichalSlawin: @wluyima Will it be precise enough if I do something like this:
calendar.add(Calendar.DAY_OF_YEAR, -ageDaysInt);
instead of substracting years? ageDayisInt is converted ageDays which is ageInMillis / 86400000

api/src/test/java/org/openmrs/api/PatientServiceTest.java
wluyima: Note that the you're adding extra in the file that matches the value of USERS_WHO_ARE_PATIENTS_XML, may be the patient is in there

api/src/main/java/org/openmrs/validator/VisitValidator.java
wluyima: Not sure if I understand what you're saying, all I'm saying is you can merge the 2 methods into 1 that way you don't have to validate the birthdate that's not estimated twice

MichalSlawin:
@wluyima pushed a commit with changes you asked for

api/src/main/java/org/openmrs/validator/VisitValidator.java
wluyima: That's true @ssmusoke, visit.patient is a required field and should be one of the first checks in the validation logic

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -159,6 +165,44 @@ private void validateStopDatetime(Visit visit, Visit otherVisit, Errors errors)
...
-		
+
+	}
+
+	private void validateStartDateTimeToBirthdate(Visit visit, Errors errors) {
+
+		Patient patient = visit.getPatient();
+
+		if (patient != null) {
+			if (visit.getStartDatetime() != null && patient.getBirthdate() != null) {
```
wluyima: Same here, startDatetime is required and was checked earlier, I understand why he checks why it's null here but it's a result of this validator not being well written, in theory a validator needs to check required field first at the top like this does and then only proceeds if they are all set, kind of like the snippet below:
```
ValidationUtils.rejectIfEmpty(errors, "patient", "Visit.error.patient.required");
ValidationUtils.rejectIfEmpty(errors, "visitType", "Visit.error.visitType.required");
ValidationUtils.rejectIfEmpty(errors, "startDatetime", "Visit.error.startDate.required");
if(errors.hasErrors()){
    return;
}
//At this point we don't need any null checks for the required fields
```

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -159,6 +165,44 @@ private void validateStopDatetime(Visit visit, Visit otherVisit, Errors errors)
...
+				if (!patient.getBirthdateEstimated() && visit.getStartDatetime().before(patient.getBirthdate())) {
+					errors.rejectValue("startDatetime", "Visit.startDateCannotBeBeforeBirthdate",
+							"This visit has a start date before birth date of the patient.");
```
wluyima: Can you please translate this error message? i.e. add a code and its translation to message.properties file and use the code here

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -159,6 +165,44 @@ private void validateStopDatetime(Visit visit, Visit otherVisit, Errors errors)
...
+	}
+
+	private void validateStartDateTimeToBirthdate(Visit visit, Errors errors) {
+
+		Patient patient = visit.getPatient();
+
+		if (patient != null) {
+			if (visit.getStartDatetime() != null && patient.getBirthdate() != null) {
+				if (!patient.getBirthdateEstimated() && visit.getStartDatetime().before(patient.getBirthdate())) {
+					errors.rejectValue("startDatetime", "Visit.startDateCannotBeBeforeBirthdate",
```
wluyima: You can avoid code duplication by creating a local variable for the birthdate, assign the estimated date or the actual date of course depending on if it's estimated vs not and then do the comparison and rejection of the value once

api/src/test/resources/org/openmrs/include/standardTestDataset.xml
```diff
@@ -237,7 +237,7 @@
...
   <person person_id="1" gender="M" birthdate="1948-01-01 00:00:00.0" birthdate_estimated="0" dead="false" creator="1" date_created="2005-01-01 00:00:00.0" voided="false" uuid="ba1b19c2-3ed6-4f63-b8c0-f762dc8d7562"/>
   <person person_id="2" gender="M" birthdate="1975-04-08 00:00:00.0" birthdate_estimated="0" dead="false" creator="1" date_created="2005-09-22 00:00:00.0" voided="false" uuid="da7f524f-27ce-4bb2-86d6-6d1d05312bd5"/>
-  <person person_id="6" gender="M" birthdate="2007-05-27 00:00:00.0" birthdate_estimated="0" dead="false" creator="1" date_created="2006-01-18 00:00:00.0" voided="false" uuid="a7e04421-525f-442f-8138-05b619d16def"/>
+  <person person_id="6" gender="M" birthdate="2000-05-27 00:00:00.0" birthdate_estimated="0" dead="false" creator="1" date_created="2006-01-18 00:00:00.0" voided="false" uuid="a7e04421-525f-442f-8138-05b619d16def"/>
```
wluyima: Why are you changing this? You should've created a new patient record may be in the VisitValidatorTest.xml file since that technically should only be used in the VisitValidatorTest that way you don't break existing tests that are not related

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -469,4 +466,82 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() throws Ex
...
+	/**
+	 *  @see VisitValidator#validate(Object,Errors)
+	 */
+	@Test
+	@Verifies(value= "should fail validation if the visit start date is before the birth date of the patient", method = "validate(Object, Errors)")
+	public void validate_shouldFailValidationIfVisitStartDateIsBeforeBirthdate() throws Exception {
+		Calendar calendar = Calendar.getInstance();
+		calendar.set(1970, Calendar.FEBRUARY, 2);
+		Visit visit = makeVisit(42);
+		visit.setStartDatetime(calendar.getTime());
```
wluyima: It's good practice actually to include a sanity check that ensures that actually the visit start date is before the birtdate just in case the test data is edited.

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -469,4 +466,82 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() throws Ex
...
+	 */
+	@Test
+	@Verifies(value= "should pass validation if the visit start date is after the estimated birth date of the patient", method= "validate(Object, Errors)")
+	public void validate_shouldPassValidationIfVisitStartDateIsAfterEstimatedBirthdate() throws Exception {
+		Calendar calendar = Calendar.getInstance();
+		calendar.set(2012, Calendar.FEBRUARY, 2);
+		Visit visit = makeVisit(42);
+		visit.setStartDatetime(calendar.getTime());
+
+		visit.setPatient(Context.getPatientService().getPatient(7));
```
wluyima: Would be nice to add an assertion as a sanity check that ensures the birthdate is estimated just in case test data is edited

wluyima: Wouldn't it be better to use a visit date that is in within the allowed range from the validator instead of 2012 which is way off?

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -469,4 +466,82 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() throws Ex
...
+		new VisitValidator().validate(visit, errors);
+
+		assertTrue(errors.hasFieldErrors("startDatetime"));
+	}
+
+	/**
+	 *  @see VisitValidator#validate(Object,Errors)
+	 */
+	@Test
+	@Verifies(value= "should pass validation if the visit start date is after the estimated birth date of the patient", method= "validate(Object, Errors)")
```
wluyima: I would think this should say something like `should pass validation if the visit start date is before but within range of the estimated birth date of the patient`

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -469,4 +466,82 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() throws Ex
...
+	}
+
+	/**
+	 *  @see VisitValidator#validate(Object,Errors)
+	 */
+	@Test
+	@Verifies(value= "should fail validation if the visit start date is before the estimated birth date of the patient less than 2 YO", method= "validate(Object, Errors)")
+	public void validate_shouldFailValidationIfVisitStartDateIsBeforeEstimatedBirthdateLessThanTwoYO() throws Exception {
+		Calendar calendar = Calendar.getInstance();
+		calendar.set(2012, Calendar.FEBRUARY, 2);
```
wluyima: I like this test better since the date is barely outside the allowed 2yr range

wluyima: But it also needs to have the visit startDate set to before the estimated birthdata for it to be more meaningful since we are rejecting only dates that come earlier

wluyima: But I think it can be even closer the test patient's birthdate is 2015 so I would expect a test visit start date that is a day before the 2yr range

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
wluyima: It would be nice to use dates that are just day off

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -159,6 +165,44 @@ private void validateStopDatetime(Visit visit, Visit otherVisit, Errors errors)
...
+							"This visit has a start date before birth date of the patient.");
+				}
+
+				if (patient.getBirthdateEstimated()) {
+
+					Calendar calendar = Calendar.getInstance();
+					calendar.setTime(patient.getBirthdate());
+
```
wluyima: I would say for estimated birthdate and in general birthdate precision when comparing it to other dates should be to day and ignoring the time

wluyima: So you should possibly use age in days

api/src/main/java/org/openmrs/validator/VisitValidator.java
MichalSlawin: @wluyima Do I have to add this manually to these 20 files or is there a faster way to do that?

wluyima: I mean the error message only

api/src/test/resources/org/openmrs/include/standardTestDataset.xml
MichalSlawin: After adding one more patient to dataset I get errors like this one:
org.openmrs.validator.VisitValidatorTest.validate_shouldAcceptAVisitIfStartDateTimeFallsIntoAnotherVoidedVisitOfTheSamePatient(org.openmrs.validator.VisitValidatorTest)
  Run 1: VisitValidatorTest.before:53->BaseContextSensitiveTest.executeDataSet:670->BaseContextSensitiveTest.executeDataSet:775 » JdbcSQL
  Run 2: VisitValidatorTest.tearDown:71 NullPointer

dkayiwa:
@MichalSlawin are you still responding to the review comments? :)

MichalSlawin:
@dkayiwa I don' t really know much about using datasets yet and I am keep making other tests fail. So I think it might be better if I unassign myself for now so someone else can take this ticket.

dkayiwa:
@MichalSlawin thanks for the feedback. In that case, let me close this pull request.

PR #1844, opened 2016-09-27
Refactored obsService#saveObs(org.openmrs.Obs,String)

bharatak:
TRUNK-4944

## Description
This PR refactors the existing org.openmrs.api.ObsService#saveObs(org.openmrs.Obs, String).  No additional functionality is added in this PR.  The subsequent PRs will add a fix for the saving the childObs if it is dirty and not the complete obs hierarchy as mentioned by @bmamlin [here](org.openmrs.api.ObsService#saveObs(org.openmrs.Obs, String))

## Checklist:

- [x ] My pull request only contains one single commit.

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
@@ -9,14 +9,6 @@
  */
 package org.openmrs.api.impl;
 
-import java.io.File;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Vector;
-
 import org.openmrs.Concept;
 import org.openmrs.ConceptName;
 import org.openmrs.Encounter;
@@ -39,6 +31,14 @@ import org.openmrs.util.OpenmrsConstants.PERSON_TYPE;
 import org.openmrs.util.OpenmrsUtil;
 import org.openmrs.util.PrivilegeConstants;
 import org.springframework.transaction.annotation.Transactional;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Vector;
 
 /**
  * Default implementation of the Observation Service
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -87,7 +87,148 @@ public void onShutdown() {
 	 * @see org.openmrs.api.ObsService#saveObs(org.openmrs.Obs, String)
 	 */
 	public Obs saveObs(Obs obs, String changeMessage) throws APIException {
-		if (null != obs && null != obs.getConcept() && obs.getConcept().isComplex()
+		if(obs == null){
+			throw new APIException("obs cannot be null!!", (Object[]) null); //TODO: change this to a message string!!
```
wluyima: Why not just pass in only the message?

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -87,7 +87,148 @@ public void onShutdown() {
...
+		handleExistingObsWithComplexConcept(obs);
+
+		ensureRequirePrivilege(obs);
+
+		//Should allow updating a voided Obs, it seems to be pointless to restrict it,
+		//otherwise operations like merge patients won't be possible when to moving voided obs
+		if (obs.getObsId() == null || obs.getVoided()) {
+			return saveNewAndVoidedObs(obs,changeMessage);
+		} else if(!obs.isDirty()){
+			setPersonFromEncounter(obs);
```
wluyima: I find it strange that if the Obs is not dirty, that you're setting the person, should this logic really be in Encounter.java when you add an Obs? And may be in Obs.java when a group member is added

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -87,7 +87,148 @@ public void onShutdown() {
...
+
+		RequiredDataAdvice.recursivelyHandle(SaveHandler.class, newObs, changeMessage);
+
+		// save the new row to the database with the changes that
+		// have been made to it
+		dao.saveObs(newObs);
+
+		saveObsGroup(newObs,null);
+
+		voidExistingObs(obs, changeMessage, newObs);
```
wluyima: voidExistingObs(), This method seems to be just saving rather than voiding an Obs and you've already marked the Obs as voided, unless you intended to move setting of voided fields to that method.

api/src/main/resources/messages.properties
```diff
@@ -3291,4 +3292,4 @@ uncaughtException.additionalInfo.modules=Names and versions of all installed mod
 OrderFrequency.concept.shouldBeClassFrequency=Concept should be of class Frequency
 OrderFrequency.concept.shouldNotBeShared=Only one OrderFrequency should be mapped to a particular concept
 
-Module.error.cyclicDependencies=Could not start some modules because they depend on others which are not started. Please first start them or use the "Start All" button on the "Manage Modules" page: {0}
```
wluyima: It this a line ending change? I would just discard it

api/src/test/java/org/openmrs/api/ObsServiceTest.java
```diff
@@ -75,6 +75,13 @@
 	protected static final String ENCOUNTER_OBS_XML = "org/openmrs/api/include/ObsServiceTest-EncounterOverwrite.xml";
 	
 	protected static final String COMPLEX_OBS_XML = "org/openmrs/api/include/ObsServiceTest-complex.xml";
+
+	@Test(expected = APIException.class)
+	@Verifies(value = "should throw APIException when obs is null", method = "saveObs(Obs,String)")
+	public void shouldReturnAPIExceptionWhenObsIsNull(){
+		ObsService os = Context.getObsService();
+		Obs o = os.saveObs(null,"Null Obs");
```
wluyima: How about you use a junit @Rule with ExpectedException? That way you check the message otherwise the exception could be thrown for something else since APIExceptions are thrown all over the code base.

wluyima:
I think it would have been great to do the other non-related refactorings in a separate commit because it was hard for me to figure out what changes were addressing the actual bug mentioned by the ticket.

It also be nice to add a unit test that demonstrates the fix.

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
bharatak: Done! Will update the PR

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -87,7 +87,148 @@ public void onShutdown() {
...
+		}
+
+		handleExistingObsWithComplexConcept(obs);
+
+		ensureRequirePrivilege(obs);
+
+		//Should allow updating a voided Obs, it seems to be pointless to restrict it,
+		//otherwise operations like merge patients won't be possible when to moving voided obs
+		if (obs.getObsId() == null || obs.getVoided()) {
+			return saveNewAndVoidedObs(obs,changeMessage);
```
bharatak: Makes sense! Will update the PR

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
bharatak: Agreed with you!  But this PR is as-is behavior of the [current code](https://github.com/openmrs/openmrs-core/blob/2.0.x/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java#L127) and I just did the refactored it.  The next PR will be on the basis of this.  I will clean it up there.   We are setting up the Person on the obs irrespective of if the obs is dirty.

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
bharatak: We are actually voiding the obs using obsService.voidObs() method [here](https://github.com/bharatak/openmrs-core/blob/854f1648099266f2ef3e60ca43d19343b49865fd/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java#L139)

api/src/main/resources/messages.properties
bharatak: Will discard the last line.

api/src/test/java/org/openmrs/api/ObsServiceTest.java
bharatak: Done! Fixed it. Will be in next PR

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
bharatak: I have added the message code instead of string.  Sorry, missed it here.

bharatak:
@wluyima This is the first PR which does the refactoring of obsService.saveObs.  The subsequent PR will provide a fix for this issue.  Can you please let me know if this PR can be merged as part of this issue? Or should I be creating a new issue, now that the issue is already reviewed.  From next time, I'll ensure that a ticket is created separately for refactoring.

wluyima:
I see, thanks!

wluyima:
In theory you don't have to create a ticket for refactoring, next time you can use the same PR but use different commits and wait to squash them

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
wluyima: Thanks!

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
wluyima: Makes sense

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
wluyima: You're right, I didn't realize that you are voiding the old Obs

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -87,7 +87,148 @@ public void onShutdown() {
...
+		handleExistingObsWithComplexConcept(obs);
+
+		ensureRequirePrivilege(obs);
+
+		//Should allow updating a voided Obs, it seems to be pointless to restrict it,
+		//otherwise operations like merge patients won't be possible when to moving voided obs
+		if (obs.getObsId() == null || obs.getVoided()) {
+			return saveNewOrVoidedObs(obs,changeMessage);
+		} else if(!obs.isDirty()){
+			setPersonFromEncounter(obs);
```
wluyima: I think you changed this, in the current code this happens outside this if clause on line 106 above

wluyima: Ah I see that you call it twice which I guess preserves the logic before, although I would have preferred to call it once just like before but it's no big deal.

PR #1845, opened 2016-09-28
Trunk 4949

karrtikiyer-tw:
TRUNK-4949 Patient Merge - HibernatePatientDAO - getDuplicatePatientsByAttributes - Not performing on large database 

## Description
Modified the getDuplicatePatientsByAttributes method to:

1. have a group by + having count greater than 1 clause to identify duplicate patient IDs using native sql 

2. Then using the patient IDs identified in step 1 fetching Patients using HQL instead of all the self joins which was was case earlier.

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4949

## Checklist:

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
@@ -12,10 +12,13 @@ package org.openmrs.api.db.hibernate;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.Vector;
 
@@ -23,11 +26,14 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.hibernate.Criteria;
+import org.hibernate.HibernateException;
 import org.hibernate.Query;
+import org.hibernate.SQLQuery;
 import org.hibernate.SessionFactory;
 import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Projections;
 import org.hibernate.criterion.Restrictions;
+import org.hibernate.persister.entity.AbstractEntityPersister;
 import org.openmrs.Allergies;
 import org.openmrs.Allergy;
 import org.openmrs.Location;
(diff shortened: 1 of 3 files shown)


dkayiwa:
@karrtikiyer-tw thank you so much for working on this! :)
Do you mind squashing these commits into one as advised at? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

karrtikiyer-tw:
@dkayiwa : Thanks for the inputs, have squashed the commits.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -425,85 +431,177 @@ public PatientIdentifierType getPatientIdentifierType(Integer patientIdentifierT
...
-		
+		List<Integer> patientIds = new Vector<Integer>();
+
 		if (attributes.size() > 0) {
-			String select = "select distinct p1 from Patient p1, Patient p2";
-			String where = " where p1 <> p2 ";
-			String orderBy = " order by ";
-			
-			Class patient = Patient.class;
-			Set<String> patientFieldNames = new HashSet<String>(patient.getDeclaredFields().length);
```
wluyima: I actually don't seem to understand why we had this logic, seems like it can cause issues

karrtikiyer-tw:
@dkayiwa and @wluyima : Wanted to check if there is any update on the code review initiated for this item?

dkayiwa:
@karrtikiyer-tw it is a shame that we did not have any unit tests for this method. I tried your new tests and found them to fail with the existing code (before your changes). Looking into why.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -425,85 +431,177 @@ public PatientIdentifierType getPatientIdentifierType(Integer patientIdentifierT
...
+				throw ex;
+
 			}
-			
-			Class identifier = PatientIdentifier.class;
-			Set<String> identifierFieldNames = new HashSet<String>(identifier.getDeclaredFields().length);
-			for (Field f : identifier.getDeclaredFields()) {
-				identifierFieldNames.add(f.getName());
-				log.debug(f.getName());
+			finally {}
```
dkayiwa: What is the use of the above empty finally block? :)

dkayiwa:
@karrtikiyer-tw do you also mind giving a summary regarding what you had to change in order to improve the performance?

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
karrtikiyer-tw: :-) agreed, i had initially added some code and then removed it, now it is not of much use, I shall remove this block.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -425,85 +431,177 @@ public PatientIdentifierType getPatientIdentifierType(Integer patientIdentifierT
...
+				}
 			}
-			
-			Class personName = PersonName.class;
-			Set<String> personNameFieldNames = new HashSet<String>(personName.getDeclaredFields().length);
-			for (Field f : personName.getDeclaredFields()) {
-				personNameFieldNames.add(f.getName());
-				log.debug(f.getName());
+			catch (HibernateException ex) {
+				log.warn("Hibernate exception: " + ex.getStackTrace().toString());
```
karrtikiyer-tw: Yes, valid, shall correct it.

karrtikiyer-tw:
@dkayiwa : I think I know why the new test cases that I have written are failing on the old code, this is because the SQL query formed requires the field present in order by clause to be present in select clause, and in the old code this is not happening, the select clause in old code only contains select * from patient, and not fields from identifier and person name table which are added to the order by clause. I agree that we should have ideally had some test cases for the earlier code, if so, these SQL query formation issues would have been caught by now leaving aside the performance issue.

karrtikiyer-tw:
@dkayiwa : I have tried providing the summary in the description of the pull request, can you go through it once and let me know if you need more details? Basically now the query is split in two parts instead of a single query: 1> First fetch the duplicate patient ID's using native SQL, 2> Fetch all the patients based on the patient ID's retrieved from step 1. So the self joins and <> checks which were present earlier are removed, which improves the performance.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -425,85 +431,177 @@ public PatientIdentifierType getPatientIdentifierType(Integer patientIdentifierT
...
+		String innerWhereCondition = "";
+		if (!attributes.contains("includeVoided")) {
+			innerWhereCondition = " where p1.voided = false ";
+		}
+		String innerQuery = "(Select " + innerFieldsJoined + innerSelect + innerWhereCondition + " group by "
+				+ innerFieldsJoined + " having count(*) > 1" + " order by " + innerFieldsJoined + ") t5";
+		return outerSelect + ", " + innerQuery + " where " + whereFieldsJoined + ";";
+	}
+
+	private void sortDuplicatePatients(List<Patient> patients, List<Integer> patientIds) {
```
dkayiwa: What happens if we do not sort?

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -425,85 +431,177 @@ public PatientIdentifierType getPatientIdentifierType(Integer patientIdentifierT
...
-					where += " and pi1." + s + " = pi2." + s;
-					orderBy += "pi1." + s + ", ";
-				} else {
-					log.warn("Unidentified attribute: " + s);
+
+			@Override
+			public int compare(Patient i1, Patient i2) {
+				Integer weekdayPos1 = sortOrder.get(i1.getPatientId());
+				if (weekdayPos1 == null) {
+					throw new IllegalArgumentException("Bad patient encountered: " + i1.getPatientId());
```
dkayiwa: What would lead to this? Patients without ids?

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
karrtikiyer-tw: @dkayiwa : Please refer the comment below to understand what happens if sorting isn't done.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
karrtikiyer-tw: the variable name Integer weekdayPos1 isn't correct, so I shall correct the variable name to indicate what is the variable used for.

dkayiwa: Would it be faster if we dropped the order by clause in the first query? Since we end up having to sort in memory?

karrtikiyer-tw: @dkayiwa : I am on vacation for rest of this week, I shall respond to your question by early next week. Apologies for the delay.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
teleivo: @dkayiwa is there much value to logging the exception and throwing it again? why not just not catch it at all?

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -425,85 +431,177 @@ public PatientIdentifierType getPatientIdentifierType(Integer patientIdentifierT
...
+			identifierFieldNames.add(f.getName());
+			log.debug(f.getName());
+		}
+		List<String> whereConditions = new ArrayList<String>();
+
+
+		List<String> innerFields = new ArrayList<String>();
+		String innerSelect = " from patient p1 ";
+
+		for (String s : attributes) {
```
teleivo: @karrtikiyer-tw can you please use more expressive variable names, like maybe `attribute` in this case. your method is very long so one has to mentally keep track of all the different variables. good names make it easier, thanks :smile:

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -425,85 +431,177 @@ public PatientIdentifierType getPatientIdentifierType(Integer patientIdentifierT
...
+			if (s != null)
+				s = s.trim();
+			if (patientFieldNames.contains(s)) {
+
+				AbstractEntityPersister aep = ((AbstractEntityPersister) sessionFactory.getClassMetadata(Patient.class));
+				String[] properties = aep.getPropertyColumnNames(s);
+				if (properties.length >= 1)
```
teleivo: @karrtikiyer-tw can you put the curly braces also on your single line if statements. reduces the chance of someone messing things up later.

api/src/test/java/org/openmrs/api/db/PatientDAOTest.java
```diff
@@ -2033,4 +2033,33 @@ public void getPatients_shouldGetNoVoidedPersonWhenVoidedFalseIsPassed() throws
 		List<Patient> patients = dao.getPatients("voided-bravo", false, 0, 11);
 		Assert.assertEquals(0, patients.size());
 	}
+	/**
```
teleivo: @karrtikiyer-tw what about adding a tests for when one adds an attribute that does not exist like "foobar" or one that is not persisted in the db. these are definitely cases that could happen and so if we have those tests we can make sure it works, when someone else later refactors your methods we make sure he doesnt break this behavior.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -425,85 +431,177 @@ public PatientIdentifierType getPatientIdentifierType(Integer patientIdentifierT
...
+					s = properties[0];
+				whereConditions.add(" t1." + s + " = t5." + s);
+				innerFields.add("p1." + s);
+			} else if (personFieldNames.contains(s)) {
+				if (!outerSelect.contains("person")) {
+					outerSelect += "inner join person t2 on t1.patient_id = t2.person_id ";
```
teleivo: @dkayiwa and @wluyima I would be in favor of using hibernates criteria, since there is a lot of "manual" SQL construction going on here that is hard to read overall. With criteria we prevent some errors already at compile time and it makes it more readable in my view. Whats your opinion on this?

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
dkayiwa: @teleivo  very good point. I completely agree! 👍

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
dkayiwa: @karrtikiyer-tw thanks for the update. Enjoy your vacation! :)

dkayiwa:
@teleivo thank you very much for doing this code review. Please keep up the great spirit!!! 👍 👍 👍

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
karrtikiyer-tw: @teleivo : Shall take care and update the code accordingly.

api/src/test/java/org/openmrs/api/db/PatientDAOTest.java
karrtikiyer-tw: @teleivo : Yes there are lot more test cases with permutation and combination of various attributes selected for duplicate matching as well. However since there were no cases initially when I picked this up, wanted to add the cases to cover the scenarios for which this fix was being done, we shall create more cases for all positive and negative scenarios.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
teleivo: @karrtikiyer-tw happy vacation! enjoy

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
dkayiwa: @karrtikiyer-tw am satisfied with your explanation. :)

api/src/test/java/org/openmrs/api/db/PatientDAOTest.java
karrtikiyer-tw: This has been taken care in the latest commit.

PR #1846, opened 2016-09-30
TRUNK-4936 Add ConceptSource.uniqueId

teleivo:
## Description

* add ConceptSource.uniqueId
* add property to hibernate mapping file ConceptSource.hbm.xml
* ensure ConceptSourceValidator validates max conceptId length
* add liquibase changeset adding the column concept_source.unique_id
with unique constraint (but defaults to null)
* add service method to ConceptService with implementation in
ConceptServiceImpl
* add ConceptDAO method with HibernateDAO implementation
* add tests for the service layer
* add the official registered OID of "SNOMED CT" into its unique_id
column in the standardTestDataset.xml so it can be used in tests

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4936


api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
```diff
@@ -1141,7 +1141,19 @@ public ConceptSource getConceptSourceByName(String conceptSourceName) throws DAO
...
 		return (ConceptSource) criteria.uniqueResult();
 	}
-	
+
+	/**
+	 * @see org.openmrs.api.db.ConceptDAO#getConceptSourceByUniqueId(java.lang.String)
+	 */
+	public ConceptSource getConceptSourceByUniqueId(String uniqueId) {
+		if (StringUtils.isBlank(uniqueId)) {
+			return null;
```
dkayiwa: How about throwing an exception instead of returning null?

api/src/main/java/org/openmrs/validator/ConceptSourceValidator.java
```diff
@@ -52,7 +52,8 @@ public void validate(Object obj, Errors errors) throws IllegalArgumentException
 			        + ConceptSource.class);
 		} else {
 			ValidationUtils.rejectIfEmptyOrWhitespace(errors, "name", "error.name");
-			ValidateUtil.validateFieldLengths(errors, obj.getClass(), "name", "hl7Code", "description", "retireReason");
+			ValidateUtil.validateFieldLengths(errors, obj.getClass(), "name", "hl7Code", "uniqueId", "description",
```
dkayiwa: Does this mean that we are no longer able to save any existing concept source that has not yet been assigned a value to this new field?

teleivo: Nope. Just means if you exceed it max length its going to fail. Like retire  reason. Not needed at creation. But if more than max not allowed

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
teleivo: No problem if that's better, I'll do it. Its just what's the difference in asking for the concept source of " " and "2.25.89.1" if they both don't exist

dkayiwa: My line of thinking was that, why would one ask for a concept source with a null or empty parameter? Wouldn't this be a mistake that they should get an exception about?

teleivo: Probably right. I think it should be like the other methods of this service. That's why I did it like this. I think some with null param throw an exception and with an empty string they return null. Generally we need more of a best practice/this is how we do it kind of style

api/src/test/java/org/openmrs/validator/ConceptSourceValidatorTest.java
```diff
@@ -9,10 +9,10 @@
  */
 package org.openmrs.validator;
 
+import org.apache.commons.lang3.StringUtils;
```
dkayiwa: How about using lang instead of lang3?

teleivo: Why? Isn't lang3 the future? ;) I mean "back to the future" was probably better than the sequel but

dkayiwa: Oh you are right about this! :)

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
dkayiwa: What do you think of this? http://www.javapractices.com/topic/TopicAction.do?Id=5

teleivo: I am totally for failing fast. But I think we might need to split it. Passing null leads to an exception and passing an empty string you get null because an empty string is more like passing a string that doesn't exist. Right?

dkayiwa: I agree!

teleivo:
@dkayiwa done, please review again :)
thank you!!

api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -9534,5 +9534,13 @@
             AND NOT EXISTS (SELECT role, privilege FROM role_privilege rp2 WHERE rp2.privilege='Add Visits'                     AND rp2.role=rp.role);
         </sql>
     </changeSet>
+	<changeSet id="TRUNK-4936-20160930-1000" author="teleivo">
+		<comment>Add unique_id column to concept_reference_source</comment>
```
dkayiwa: Can you add a precondition to do this only if the column does not already exist?

teleivo: added it at https://github.com/openmrs/openmrs-core/pull/1847

is it necessary to change the changeSet id  ? I left it as is in the current PR.

PR #1848, opened 2016-10-03
TRUNK-4953 ConceptSourceValidator doesnt fail on empty description

teleivo:
## Description

Fixes ConceptSourceValidator which does not fail if "description" is
empty altough the description column is required.

* add not-null="true" to name and description fields in ConceptSource.hbm.xml
* Validator now fails on empty "description"

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4953

### DIFF
--- a/api/src/main/java/org/openmrs/validator/ConceptSourceValidator.java
+++ b/api/src/main/java/org/openmrs/validator/ConceptSourceValidator.java
@@ -40,7 +40,7 @@ public class ConceptSourceValidator implements Validator {
 	 * 	 * @see org.springframework.validation.Validator#validate(java.lang.Object,
 	 *      org.springframework.validation.Errors)
 	 * @should fail validation if name is null or empty or whitespace
-	 * @should pass validation if description is null or empty or whitespace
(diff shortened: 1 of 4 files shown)


wluyima:
These changes are definitely going to break implementations

wluyima:
The build seems to be broken due to some failing tests

teleivo:
@wluyima how can these changes break implementations?

These are DB constraints!

teleivo:
@wluyima or do you mean that ConceptSource.description should be allowed empty strings?

thats the table on 2.0

```sql
concept_reference_source | CREATE TABLE `concept_reference_source` (
  `concept_source_id` int(11) NOT NULL AUTO_INCREMENT,
  `name` varchar(50) NOT NULL DEFAULT '',
  `description` text NOT NULL,
```

wluyima:
It appears like description is not nullable so I guess it's fine to add the constraint in the validator and hbm file

teleivo:
@wluyima build is fine now :+1:

PR #1850, opened 2016-10-04
Fix for editing a child obs makes the parent obs dirty and creates a new instance

bharatak:
TRUNK-4944

## Description
This PR fixes the cloning of parent obs when a child (leaf) obs is edited.   The following are the things that are done

- Removed setting the dirty flag when obs.setGroupMembers(), obs.addGroupMember() and obs.removeGroupMember() is called
- EncounterService.saveEncounter() will use encounter.getObsAtTopLevel() so that we just pick up the top level obs.  Saving a top level obs using obsService.saveObs() will take care of individually saving the leaf obs.
- Added integration tests for various obs hirearchies and saving of obs at different levels - leaf, middle level, top level and peer at top level

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4944

## Checklist:

### DIFF
--- a/api/src/test/java/org/openmrs/api/EncounterServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/EncounterServiceTest.java
@@ -9,28 +9,6 @@
  */
 package org.openmrs.api;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNotSame;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertSame;
-import static org.junit.Assert.assertTrue;
-
-import java.lang.reflect.Field;
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Calendar;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeSet;
-import java.util.Vector;
-
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Ignore;
@@ -57,7 +35,6 @@ import org.openmrs.TestOrder;
 import org.openmrs.User;
 import org.openmrs.Visit;
 import org.openmrs.VisitType;
-import org.openmrs.TestOrder;
 import org.openmrs.api.builder.OrderBuilder;
 import org.openmrs.api.context.Context;
 import org.openmrs.api.handler.EncounterVisitHandler;
@@ -72,6 +49,28 @@ import org.openmrs.util.DateUtil;
 import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.PrivilegeConstants;
 
+import java.lang.reflect.Field;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.Vector;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+
 /**
  * Tests all methods in the {@link EncounterService}
  */
@@ -82,9 +81,11 @@ public class EncounterServiceTest extends BaseContextSensitiveTest {
 	protected static final String UNIQUE_ENC_WITH_PAGING_XML = "org/openmrs/api/include/EncounterServiceTest-pagingWithUniqueEncounters.xml";
 	
 	protected static final String TRANSFER_ENC_DATA_XML = "org/openmrs/api/include/EncounterServiceTest-transferEncounter.xml";
+
+	protected static final String ENC_OBS_HIERARCHY_DATA_XML = "org/openmrs/api/include/EncounterServiceTest-saveObsHierarchyTests.xml";
+
 	protected static final String ORDER_SET = "org/openmrs/api/include/OrderSetServiceTest-general.xml";
-	
-	private EncounterService encounterService;
+
 
 	/**
 	 * This method is run before all of the tests in this class because it has the @Before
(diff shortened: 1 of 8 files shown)


wluyima:
Thanks for the PR @bharatak I think we all agreed that removing or adding a member from or to a obs group should make it dirty. I also think you're mixing up obs in an encounter with obs in an obs group. TRUNK-4944 is purely about obs in obs group and we should address that.

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -98,7 +97,235 @@
...
+		Obs oParent =createObs(cs.getConcept(23), null); //in the concept set table as a set
+
+		//add o and o2 to the parent obs
+		oParent.addGroupMember(o2);
+		oParent.addGroupMember(o);
+
+		Encounter enc = buildEncounter();
+		enc.addObs(oParent);
+
+		enc = es.saveEncounter(enc);
```
wluyima: Don't you have to flush the session for the test to be concrete, couldn't you actually use an existing encounter already in the DB?

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -98,7 +97,235 @@
...
+		//create an obs
+		Obs o = createObs(cs.getConcept(3),"original obs value text");
+		//create a second obs
+		Obs o2 = createObs(cs.getConcept(3),"second obs value text");
+		//create a parent obs
+		Obs oParent =createObs(cs.getConcept(23), null); //in the concept set table as a set
+
+		//add o and o2 to the parent obs
+		oParent.addGroupMember(o2);
+		oParent.addGroupMember(o);
```
wluyima: When you say leaf, I was assume one of the  obs should not be an immediate child but i see you're adding both as children of the same parent, i would expect a child-parent-grandparent tree

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -98,7 +97,235 @@
...
+		assertFalse(oParent.getVoided());
+		assertFalse(o2.getVoided());
+
+		assertNotNull(o3);
+		assertFalse(o3.getVoided());
+		assertEquals("Obs value updated", o3.getValueText());
+	}
+
+	@Test
+	@Verifies(value = "should save encounter correctly when a level 1 child obs updated without recreating an entire obs hierarchy", method = "saveEncounter(Encounter)")
```
wluyima: So this test seems to contain the tree of obs I expected in the other test, when you say level one i would expect editing a child and not a grand parent. I think use child and nest child to be clearer

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -98,7 +97,235 @@
...
+
+		//add o and o2 to the parent obs
+		oParent.addGroupMember(o2);
+		oParent.addGroupMember(o1);
+
+		Encounter enc = buildEncounter();
+		enc.addObs(oGrandParent);
+
+		enc = es.saveEncounter(enc);
+
```
wluyima: Again you would need to flush and clear the session

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -98,7 +97,235 @@
...
+		assertEquals(1,obsAtTopLevelUpdated.size());
+
+		assertEquals(2, obsAtTopLevelUpdated.iterator().next().getGroupMembers(true).size()); // should be oParent and o3
+
+		oParent = os.getObs(oParent.getObsId());
+		assertFalse(oParent.getVoided());
+
+		o4 = os.getObs(o4.getObsId());
+		assertTrue(oParent.getGroupMembers(true).contains(o4));
+
```
wluyima: I don't see the assertions for the grand parent that they are not voided and still contain the same unvoided children except for the edited one

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -98,7 +97,235 @@
...
+	}
+
+	@Test
+	@Verifies(value = "should save encounter correctly when a top level obs updated without recreating an entire obs hierarchy", method = "saveEncounter(Encounter)")
+	public void saveEncounter_shouldSaveEncounterWhenTopLevelObsIsUpdated(){
```
wluyima: How different is this test from the first? You're updating an immediate child in both cases

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -98,7 +97,235 @@
...
+		assertEquals(3, obsAtTopLevelUpdated.iterator().next().getGroupMembers(true).size()); // should be o1, o2 and o3
+
+		oParent = os.getObs(oParent.getObsId());
+		assertTrue(oParent.getGroupMembers(true).contains(os.getObs(o1.getObsId())));
+		assertTrue(oParent.getGroupMembers(true).contains(os.getObs(o2.getObsId())));
+		assertTrue(oParent.getGroupMembers(true).contains(os.getObs(o3.getObsId())));
+	}
+
+	@Test
+	@Verifies(value = "should save encounter correctly when a top level obs is updated by removing a child obs", method = "saveEncounter(Encounter)")
```
wluyima: 'is update by removing a child obs'? Or 'if an immediate child is removed'

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -98,7 +97,235 @@
...
+		o.setObsDatetime(new Date());
+		o.setPerson(new Patient(3));
+
+		if(value!=null){
+			o.setValueText(value);
+		}
+
+		return o;
+	}
+
```
wluyima: I really don't think you needed to have tests for top level vs nested child, I think you should have tests for removing, adding or editing a child regardless of the level

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
bharatak: Fixed this.. Moved the test data into dataset xml

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -98,7 +97,235 @@
 	public void runBeforeEachTest() throws Exception {
 		executeDataSet(ENC_INITIAL_DATA_XML);
 	}
-	
+
+	@Test
+	@Verifies(value = "should save encounter correctly when a leaf child obs updated without recreating an entire obs hierarchy", method = "saveEncounter(Encounter)")
+	public void saveEncounter_shouldNotCreateACopyOfParentObsButEditChildObs(){
```
bharatak: Fixed it.

bharatak:
@wluyima I have updated the PR with your comments.  Can I request you to please have a look once? Also, if possible, if you can give all the comments at once, it will work better.  We are nearing a deadline and we appreciate your efforts regarding this PR merge.  Thank you.

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -98,7 +99,127 @@
...
+		EncounterService es = Context.getEncounterService();
+		ObsService os = Context.getObsService();
+
+		Encounter enc = es.getEncounter(100);
+
+		Obs o = os.getObs(101);
+		o.setValueText("Obs value updated");
+
+		es.saveEncounter(enc);
+		Context.evictFromSession(enc);
```
wluyima: Not evict, you should flush and then clear the session

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -98,7 +99,127 @@
...
+
+		Encounter enc = es.getEncounter(100);
+		Obs oParent = os.getObs(100);
+		Obs oChild = os.getObs(101);
+
+		oParent.removeGroupMember(oChild);
+
+		es.saveEncounter(enc);
+
+		Context.evictFromSession(enc);
```
wluyima: For this particular test when I change to flushSession() and then clearSession() it fails, I personally think these are the correct calls in this order and not evicting the encounter and flushing after. Evicting means remove from the without persisting any changes, is this what you want to happen? I'd think what you want is to push the changes to the DB, clear the whole session so that all entities get reloaded with their new saved state.

bharatak: @wluyima I agree with you and I see this test case fails.  The reason it fails as follows: When the original oChild is [removed](https://github.com/bharatak/openmrs-core/blob/76ec6f3b4d6a6ed39de892cfda35e7f25efc144c/api/src/test/java/org/openmrs/api/EncounterServiceTest.java#L191) from oParent, the obsGroup is set to null in oChild.  But while we save the oChild, we evict the original obs to get the fresh one [here](https://github.com/bharatak/openmrs-core/blob/76ec6f3b4d6a6ed39de892cfda35e7f25efc144c/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java#L131) and there by resetting the obs_group to previous parent.

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -98,7 +99,127 @@
...
+		Context.flushSession();
+
+		enc = es.getEncounter(100);
+
+		Set<Obs> obsAtTopLevel = enc.getObsAtTopLevel(true);
+		assertEquals(2,obsAtTopLevel.size()); //oChild's new version is still associated to encounter but not part of the oParent anymore
```
bharatak: When we perform a oParent.removeGroupMember(oChild), the oChild's setObsGroup is set to null.  When we save oChild (which happens as part of es.saveEncounter), a clone of it is created, but it is still associated to the encounter with its parent set to null.  i.e. it will be a top level obs.

wluyima: That's correct thanks!

bharatak:
@wluyima Thank you.

PR #1851, opened 2016-10-07
TRUNK-4955 | Obs not made dirty when obsdatetime is set with same val…

preethi29:
## Description
Obs.setObsDateTime makes obs as dirty when comparing timestamp against date. So Changed OpenmrsUtil.nullSafeEquals to use compare(Date, Date) method when any of its argument is of Timestamp type.

## Related Issue
see https://issues.openmrs.org/browse/TRUNK-4955

## Checklist:

…ue but compared against timestamp instance

### DIFF
--- a/api/src/main/java/org/openmrs/util/OpenmrsUtil.java
+++ b/api/src/main/java/org/openmrs/util/OpenmrsUtil.java
@@ -617,15 +617,14 @@ public class OpenmrsUtil {
 		}
 		return ret;
 	}
-	
+
 	public static <Arg1, Arg2 extends Arg1> boolean nullSafeEquals(Arg1 d1, Arg2 d2) {
 		if (d1 == null) {
 			return d2 == null;
 		} else if (d2 == null) {
 			return false;
-		} else {
-			return d1.equals(d2);
 		}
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -617,15 +617,14 @@ public static void applyLogLevel(String logClass, String logLevel) {
...
+
 	public static <Arg1, Arg2 extends Arg1> boolean nullSafeEquals(Arg1 d1, Arg2 d2) {
 		if (d1 == null) {
 			return d2 == null;
 		} else if (d2 == null) {
 			return false;
-		} else {
-			return d1.equals(d2);
 		}
+		return (d1 instanceof Timestamp || d2 instanceof Timestamp) ? compare((Date) d1, (Date) d2) == 0 : d1.equals(d2);
```
wluyima: Even better!

wluyima: Wouldn't it actually be better to check if instance of Date and then delegate to compare(Date, Date)?

preethi29: will do that

PR #1854, opened 2016-10-10
TRUNK-2877 Don't set password values in the UI in the installation wizard

drbobrinkman:
## Description
Password values were pre-filled in the installation wizard UI because they were stored in the wizardModel. I interpreted "run the installation wizard" to mean any access where the hidden page variable was not set. In that case we set all password entries in the model back to an empty string.

I was not able to find working tests for the InstallationWizard. It appears that UI tests are non-working ... probably because most of the UI was moved to a module.

## Checklist:


dkayiwa:
@drbobrinkman thanks for working on this! 👍 
Did you actually test it out in the web interface to confirm that all is working as expected?

drbobrinkman:
Sort of. I thoroughly tested the "simple" route. With the "advanced" route
I tested it, and I *think* I was thorough, but there were some options that
I may not have understood.

With the "testing" route it is untested. I don't know proper values to put
in the web form for that version. Of course I don't think the previous test
suite tested that option either.

PR #1855, opened 2016-10-13
TRUNK-4963 | Reload parent obs from db once child is changed instead …

preethi29:
## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4963

## Checklist:

…updating the collection manually

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Context.java
+++ b/api/src/main/java/org/openmrs/api/context/Context.java
@@ -775,6 +775,16 @@ public class Context {
 	}
 
 	/**
+	 * Used to re-read the state of the given instance from the underlying database.
+	 * @since 2.0
+	 * @param obj The object to refresh from the database in the session
+	 */
+	public static void refreshEntity(Object obj) {
+		log.trace("refreshing object: "+obj);
+		getContextDAO().refreshEntity(obj);
+	}
+
+	/**
 	 * Used to clear a cached object out of a session in the middle of a unit of work. Future
 	 * updates to this object will not be saved. Future gets of this object will not fetch this
 	 * cached copy
--- a/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
@@ -33,7 +33,6 @@ import org.openmrs.util.PrivilegeConstants;
 import org.springframework.transaction.annotation.Transactional;
 
 import java.io.File;
-import java.util.ArrayList;
 import java.util.Date;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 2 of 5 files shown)


api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -180,28 +179,15 @@ private Obs saveObsNotDirty(Obs obs, String changeMessage) {
...
 
 		ObsService os = Context.getObsService();
-		List<Obs> toRemove = new ArrayList<>();
-		List<Obs> toAdd = new ArrayList<>();
 		for (Obs o : obs.getGroupMembers(true)) {
 			if (o.getId() == null) {
 				os.saveObs(o, null);
 			} else {
-				Obs replacement = os.saveObs(o, changeMessage);
-				if(!replacement.equals(o)){
```
wluyima: I'd imagine we need to keep this if clause that way you only refresh the parent when there is a replaced child

wluyima: I.e you need a flag that you set to true if we have a replacement and then and check it below to determine is a refresh is needed

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -775,6 +775,16 @@ public static boolean isSessionOpen() {
 	}
 
 	/**
+	 * Used to re-read the state of the given instance from the underlying database.
+	 * @since 2.0
+	 * @param obj The object to refresh from the database in the session
+	 */
+	public static void refreshEntity(Object obj) {
+		log.trace("clearing session");
```
wluyima: Should this say refreshing object and not clearing?

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -136,6 +136,21 @@ public void saveEncounter_shouldUpdateExistingEncounterWhenAChildObsIsEdited() t
...
+		executeDataSet(ENC_OBS_HIERARCHY_DATA_XML);
+		Encounter encounter = Context.getEncounterService().getEncounter(100);
+		assertEquals(1, encounter.getObsAtTopLevel(true).size());
+		Obs topLevelObs = encounter.getObsAtTopLevel(true).iterator().next();
+		topLevelObs.getGroupMembers().iterator().next().setValueText("editing first obs");
+		encounter.addObs(topLevelObs);
+
+		Encounter savedEncounter = Context.getEncounterService().saveEncounter(encounter);
+
+		assertEquals(1, savedEncounter.getObsAtTopLevel(true).size());
```
wluyima: There is already existing tests around this behavior, If you comment out the line you added that refreshes the parent you'll notice there are some existing tests that fail, in theory after your changes as long those tests are passing, the code is good

preethi29: The thing is, if i dont flush and clear the session and try to resave the enconunter, the obs which is now attached directly to the encounter will get saved. So I thought I'll add a test around it. Without the line that I added, whenever the leaf obs is edited, a duplicate obs is created which is attached only to encounter, since we save encounter more than once in our use cases.

preethi29: If the test is not needed, I can remove it.

api/src/main/java/org/openmrs/api/context/Context.java
preethi29: Will update this.

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
wluyima: Hmmm... So you're imagining a situation where say one calls saveEncounter() from within another transactional method and the session might not yet be flushed since the outer transaction is still going on, that might make sense I guess, okay we can keep the test

PR #1858, opened 2016-10-15
TRUNK-4883 Update lucene search indexes on startup

jdegraft:
## Description

Add call to update lucene indexes after openers is started after database updates are run in UpdateFilter.
## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4883
## Checklist:

### DIFF
--- a/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
+++ b/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
@@ -1707,6 +1707,10 @@ public class InitializationFilter extends StartupFilter {
 						// start openmrs
 						try {
 							UpdateFilter.setUpdatesRequired(false);
+							
+							// rebuild lucene search indexes
+							Context.updateSearchIndex();
+
 							WebDaemon.startOpenmrs(filterConfig.getServletContext());
 						}
 						catch (DatabaseUpdateException updateEx) {


dkayiwa:
@jdegraft would this work for the initial setup?

jdegraft:
@dkayiwa, this runs after the update filter which is part of the initial setup. The update filter runs the liquidate updates.

dkayiwa:
@jdegraft if am doing an initial setup (installation for the first time), would this run?

jdegraft:
Yes, doesn't the initial setup go through the update filter. Besides it is the update filter which starts up openmrs and the call to update the indexes happens after this call to start up openmrs.

dkayiwa:
@jdegraft did you actually test it out by trying to setup a brand new OpenMRS instance and confirm that this line is executed?

jdegraft:
@dkayiwa, I will do some more tests overnight.

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -242,6 +242,9 @@ public static void startOpenmrs(ServletContext servletContext) throws ServletExc
 			Listener.loadBundledModules(servletContext);
 
 			Context.startup(getRuntimeProperties());
+			
+			// rebuild lucene search indexes
```
dkayiwa: @jdegraft this will keep rebuilding the index every time openmrs starts. The ticket is about building it on the initial install.

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1707,6 +1707,10 @@ public void executing(ChangeSet changeSet, int numChangeSetsToRun) {
 						// start openmrs
 						try {
 							UpdateFilter.setUpdatesRequired(false);
+							
+							// rebuild lucene search indexes
+							Context.updateSearchIndex();
```
wluyima: You would still need this change if one isn't running the standalone

wluyima:
I guess this is

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
wluyima: I guess this is fine, the  GP gets cleared after each run so there will be no second run happening immediately after

dkayiwa: @jdegraft do you think this change is necessary? Note that the next call WebDaemon.startOpenmrs will eventually call Context. updateSearchIndex() via Context.startup()

dkayiwa:
Closing this because of failure to reproduce the reported problem.

PR #1860, opened 2016-10-23
TRUNK-4696: added specific exception classes for the Order Service

samshuster:
## Description
Modifying some Order Exception definitions to be more in line with the messages.properties. Fixing test to map to these new definitions.

## Related Issue

## Checklist:

- [ x] My pull request only contains one single commit.
- [ x] My pull request is based on the latest master branch
- [ x] My code follows the code style of this project.
- [x ] I have added tests to cover my changes.
- [x ] All new and existing tests passed.

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
@@ -49,6 +49,17 @@ import org.openmrs.api.OrderNumberGenerator;
 import org.openmrs.api.OrderService;
 import org.openmrs.api.context.Context;
 import org.openmrs.api.db.OrderDAO;
+import org.openmrs.api.order.exceptions.CannotDeleteOrderPropertyInUseException;
+import org.openmrs.api.order.exceptions.CannotDiscontinueAlreadyDiscontinuedOrderException;
+import org.openmrs.api.order.exceptions.CannotDiscontinueOrderWithActionException;
+import org.openmrs.api.order.exceptions.CannotEditAlreadyExistingOrderException;
+import org.openmrs.api.order.exceptions.CannotEditOrderPropertyInUseException;
+import org.openmrs.api.order.exceptions.CannotStopRetrospectiveDiscontinuedOrderException;
+import org.openmrs.api.order.exceptions.CannotUnvoidOrderException;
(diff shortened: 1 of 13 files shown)


api/src/main/java/org/openmrs/api/OrderExceptions.java
```diff
@@ -0,0 +1,170 @@
...
+package org.openmrs.api;
+
+import static org.openmrs.Order.Action.DISCONTINUE;
+
+/**
+ * Represents specific error conditions that are part of OrderService.
+ * The specific error classes define only a default constructor if they are meant to be used with only a single error message.
+ * In other cases, there should be static factory methods that can be used to create the error class with the specific error message needed.
+ */
+public class OrderExceptions {
```
dkayiwa: @samshuster do you mind throwing some more light as to why you prefer creating a new OrderExceptions class for grouping purposes instead of just putting them in the org.openmrs.api or any other package, as the rest of the existing custom exception classes? :)

samshuster: will obviously defer to you on whats best.
But here is why I made the decision not to make them there own classes.
1. There is a lot of exceptions being created and it would pollute the number of classes in that package.
2. These exceptions are all similar to each other and it might make it easier on user to see a list of all different Order Exception classes.
3. Unless static import used, when throwing them, the name is very expressive. new OrderExceptions.CannotEditExistingOrderException()

Downsides:
1. Not consistent with rest of project
2. OrderExceptions is essentially a package name. Making a separate package called orderexceptions would help alleviate concern 1 and 2.

samshuster: i will move them to custom exception package

api/src/main/java/org/openmrs/api/OrderExceptions.java
```diff
@@ -0,0 +1,170 @@
...
+
+	/**
+	 * Thrown with trying to discontinue an order with a specific action that cannot be discontinued.
+	 */
+	public static class CannotDiscontinueOrderWithActionException extends APIException {
+
+		public static final long serialVersionUID = 22121215L;
+		
+		public CannotDiscontinueOrderWithActionException() {
+			super("Order.action.cannot.discontinued", new Object[] { DISCONTINUE }, null);
```
dkayiwa: If just passing null for the third argument, why not just use the two argument constructor? :)

PR #1863, opened 2016-10-31
TRUNK-3675 Introduce caching to getSearchLocales()

PawelGutkowski:
# TRUNK-3675

## Description
This PR introduces Spring's caching functionality to OpenMRS. As proof of concept, results of `getSearchLocales` are cached. 

We couldn't use annotation driven cache configuration, because we are still using old `TransactionProxyFactoryBean` to handle transactions, so CacheInterceptor is added to preInterceptors of `TransactionProxyFactoryBean`, to handle cache. More info can be found [in Spring documentation](http://docs.spring.io/spring/docs/current/spring-framework-reference/html/cache.html).

~~In long run, we probably will want to replace SimpleCacheManager with EhCacheManager, but I couldn't properly inject it, I guess HIbernate's second level cache ehCache is colliding with it, because when it was disabled, I was able to register my API EhCacheManager~~

I managed to set up ehcache manager for API instead of SimpleCacheManager. Previously it didn't work because Hibernate used singleton factory, Right now we will have 2 caches  - `hibernateCache` and `apiCache`. I'm still working on making cache configuration extendable by modules.

## Related Issue
see https://issues.openmrs.org/browse/TRUNK-3675

## Checklist:

### DIFF
--- a/api/src/main/java/org/openmrs/api/AdministrationService.java
+++ b/api/src/main/java/org/openmrs/api/AdministrationService.java
@@ -18,12 +18,14 @@ import java.util.SortedMap;
 import org.openmrs.GlobalProperty;
 import org.openmrs.ImplementationId;
 import org.openmrs.OpenmrsObject;
+import org.openmrs.User;
 import org.openmrs.annotation.Authorized;
 import org.openmrs.api.db.AdministrationDAO;
 import org.openmrs.util.HttpClient;
 import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.PrivilegeConstants;
 import org.openmrs.validator.ValidateUtil;
+import org.springframework.cache.annotation.Cacheable;
 import org.springframework.validation.Errors;
 
 /**
@@ -213,6 +215,7 @@ public interface AdministrationService extends OpenmrsService {
 	 * @should overwrite global property if exists
 	 * @should not allow different properties to have the same string with different case
 	 * @should save a global property whose typed value is handled by a custom datatype
+	 * @should evict all entries of search locale cache
 	 */
 	@Authorized(PrivilegeConstants.MANAGE_GLOBAL_PROPERTIES)
 	public GlobalProperty saveGlobalProperty(GlobalProperty gp) throws APIException;
@@ -333,7 +336,17 @@ public interface AdministrationService extends OpenmrsService {
 	 * @should throw throw APIException if the input is null
 	 */
 	public void validate(Object object, Errors errors) throws APIException;
-	
+
+	/**
+	 * Returns a list of locales used by the user when searching.
+	 *
+	 * @param currentLocale currently selected locale
+	 * @param user authenticated user
+	 * @return
+	 * @throws APIException
+     */
+	public List<Locale> getSearchLocales(Locale currentLocale, User user) throws APIException;
+
 	/**
 	 * Returns a list of locales used by the user when searching.
 	 * <p>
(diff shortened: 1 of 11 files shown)


api/src/main/java/org/openmrs/api/AdministrationService.java
```diff
@@ -345,6 +358,7 @@
 	 * @should include currently selected full locale and langugage
 	 * @should include users proficient locales
 	 * @should exclude not allowed locales
+	 * @should cache results for an user
```
wluyima: typo here 'an user' -> 'a user'

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -964,4 +974,64 @@ public void getPresentationLocales_shouldPreserveInsertionOrderInSetReturnedByMe
 		Assert.assertEquals(new Locale("it", "IT"), presentationLocales.get(2));
 		Assert.assertEquals(new Locale("pl", "PL"), presentationLocales.get(3));
 	}
+
+	/**
+	 * @see AdministrationService#getSearchLocales()
+	 * @verifies cache results for an user
+	 */
+	@Test
+	public void getSearchLocales_shouldCacheResultsForAnUser() throws Exception {
```
wluyima: ForAnUser -> ForAUser

api/src/main/resources/api-ehcache.xml
```diff
@@ -0,0 +1,31 @@
...
+    <cache name="userSearchLocales"
+           maxElementsInMemory="5000"
+           eternal="true"
+           overflowToDisk="false"
+           timeToIdleSeconds="12000"
+           timeToLiveSeconds="12000"
+           diskPersistent="false"
+           diskExpiryThreadIntervalSeconds="120"
+           memoryStoreEvictionPolicy="LRU"
+    />
```
rkorytkowski: If I recall correctly, when you setup ehcache this way you don't have to declare caches for each name... If they are not declared the defaults are used... Am I wrong here?

rkorytkowski: If I'm right modules can use caches based on the default cache config without having to explicitly declare names here. Sys admins seeing too much memory consumption could edit the file and disable some of caches by overriding their settings. We could provide more granular adjustments of config for module developers in a separate issue if needed.

PawelGutkowski: Yes, you are right, default caches are created. 

> We could provide more granular adjustments of config for module developers in a separate issue if needed.

I think it is a good idea

api/src/main/resources/api-ehcache.xml
```diff
@@ -0,0 +1,31 @@
...
+<ehcache name="apiCache">
+
+    <diskStore path="user.home"/>
+
+    <defaultCache maxElementsInMemory="1000" eternal="false"
+                  timeToIdleSeconds="30" timeToLiveSeconds="30" overflowToDisk="false"
+                  diskPersistent="false" diskExpiryThreadIntervalSeconds="45"/>
+
+    <cache name="userSearchLocales"
+           maxElementsInMemory="5000"
```
wluyima: isn't this too high?

PawelGutkowski: Probably is, I've copied this configuration from commit which was reverted a couple months ago, I guess we can limit it to 500?

api/src/main/resources/ehcache.xml
```diff
@@ -9,7 +9,7 @@
     graphic logo is a trademark of OpenMRS Inc.
 
 -->
-<ehcache>
+<ehcache name="hibernateCache">
```
rkorytkowski: I just wondered why you added name="hibernateCache" here...

PawelGutkowski: I did it for debugging purposes, because default name is "_DEFAULT_". At first I had problem with those multiple CacheManager instances, so having those named was handy, and I think it can be useful in future if anybody will have to do something with those.

PR #1866, opened 2016-11-04
TRUNK-4870 Add Unit tests for ModuleUtil.expandJar(File, File, Stri…

mjanuchowski:
## Description
- created 5 unit tests for expandJar function
- added 1 simplified test jar file used by unit tests
- created 5 @should annotations in expandJar header
- one of the test cases was failing (name parameter is a file, keepFullPath is false), had to do change to expandJar function to fix a bug

## Related Issue
https://issues.openmrs.org/browse/TRUNK-4870

## Checklist:
- [X] My pull request only contains one single commit.
- [X] My pull request is based on the latest master branch
- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.


api/src/test/java/org/openmrs/module/ModuleUtilTest.java
```diff
@@ -651,4 +652,131 @@ public void getResourceFromApi_shouldReturnNullIfFileIsNotFoundInApi() throws Ex
...
+		// Act
+		ModuleUtil.expandJar(getJarFile(), destinationFolder, directoryPath, false);
+		// Assert
+		Object[] filesList = FileUtils.listFiles(destinationFolder, null, true).toArray();
+		Assert.assertEquals(numberOfFilesInSpecifiedDirectory, filesList.length);
+		String expectedPath = destinationFolder.getPath() + File.separator;
+		String ActualPath = ((File)filesList[0]).getPath().replace(((File)filesList[0]).getName(), "");
+		Assert.assertEquals(expectedPath, ActualPath);
+		// Clean up
+		FileUtils.deleteDirectory(destinationFolder);
```
teleivo: i think you can simplify your tests by leveraging the java `File` you dont need to use/add/replace file separators.

I think you can do something like

```java
	public void expandJar_shouldExpandDirectoryWithoutParentTreeIfNameIsDirectoryAndKeepFullPathIsFalse() throws IOException {

		final int numberOfFilesInSpecifiedDirectory = 2;
		String directoryPath = "META-INF/maven/org.openmrs.module/test1-api";
		File destinationFolder = this.getEmptyJarDestinationFolder();

		ModuleUtil.expandJar(getJarFile(), destinationFolder, directoryPath, false);

		List<File> actualExpandedFiles = (List<File>) FileUtils.listFiles(destinationFolder, null, true);
		Assert.assertEquals(numberOfFilesInSpecifiedDirectory, actualExpandedFiles.size());
		Assert.assertEquals(destinationFolder.toString(), actualExpandedFiles.get(0).getParent());

		FileUtils.deleteDirectory(destinationFolder);
	}
```

api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -571,7 +576,15 @@ public static void expandJar(File fileToExpand, File tmpModuleDir, String name,
 					String entryName = jarEntry.getName();
 					// trim out the name path from the name of the new file
 					if (!keepFullPath && name != null) {
-						entryName = entryName.replaceFirst(name, "");
+						if (name.equals(entryName)) {
+							int lastSlash = name.lastIndexOf('/');
```
teleivo: you use File.separator in your tests but hard code the separator here, why?

mjanuchowski: Since `JarEntry` uses slash as a separator I would expect that `name` parameter will also use them. This is in line with rest of the code in this function, which also uses hard coded slash.

teleivo:
@mjanuchowski thanks for the making the changes :+1:

api/src/main/java/org/openmrs/module/ModuleUtil.java
wluyima: Are you saying jarEntry uses forward slash on other systems like windows?

api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -571,7 +577,15 @@ public static void expandJar(File fileToExpand, File tmpModuleDir, String name,
...
+							if (lastSlash >= 0) {
+								entryName = entryName.replaceFirst(name.substring(0, lastSlash + 1), "");
```
wluyima: Can you please add a comment here to explain what you're trying to solve?

teleivo: @wluyima what about extracting it into a private method with a descriptive name instead of a comment? A lot of methods are already bloated and hard to read since they do so many things

wluyima: Which ever way he does it, as long as it's clear to the reader I'd be okay with it

api/src/main/java/org/openmrs/module/ModuleUtil.java
mjanuchowski: Yes, I ran unit tests also on Windows and they do not fail, which proves that paths in `JarEntry` do not use File.Separator but always forward slash.

api/src/main/java/org/openmrs/module/ModuleUtil.java
mjanuchowski: Actually `if` block in which this code is nested has already a pretty good description:
`// trim out the name path from the name of the new file`
I'm just making sure that we trim only a path, without the file name itself.

mjanuchowski: This comment is still valid - the whole block trims out the name path from the name of the new file. Code added in inner if statement makes sure that it happens no matter if `name` parameter is file or directory. But of course I can add comments to make that clear :)

wluyima: That's what I've trying to ask for :), i.e to add the comment to say why you have that inner clause and logic, thanks!

mjanuchowski:
Thank you for all the suggestions! Is there any extra work on this ticket?

api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -571,7 +577,16 @@ public static void expandJar(File fileToExpand, File tmpModuleDir, String name,
 					String entryName = jarEntry.getName();
 					// trim out the name path from the name of the new file
 					if (!keepFullPath && name != null) {
-						entryName = entryName.replaceFirst(name, "");
+						// if 'name' is a file then make sure that only path is trimmed out and not the file name itself 
+						if (name.equals(entryName)) {
+							int lastSlash = name.lastIndexOf('/');
+							if (lastSlash >= 0) {
+								entryName = entryName.replaceFirst(name.substring(0, lastSlash + 1), "");
```
mjanuchowski: The failing test was expanding file without full path: `expandJar_shouldExpandFileWithoutParentTreeIfNameIsFileAndKeepFullPathIsFalse()`. In this case `name` provides filename with path e.g. FOLDER1/FOLDER2/FILE. Original function was stripping this whole string from jar filename which was leaving it empty. As a result nothing was getting extracted. What I did is to make sure in that specific case that only path part of `name` gets trimmed out, so that jar filename still stays and gets extracted.

lluismf:
Check out Apache FilenameUtils, it contains a lot of handy methods to deal with file names (maybe you won't need any of it)

api/src/main/java/org/openmrs/module/ModuleUtil.java
wluyima: I understand it when you say that but do we actually ever call this method with a path and not a filename?

mjanuchowski: You're right, in openmrs-core this function is always called with `name` parameter as filename, but at the same time it is always called with `keepFullPath == true`, so the case causing an error never occured.

wluyima: What am saying is this code is intended to be called by the module framework in core only and if it always calls the method with keepFullPath set to true and name as a filename, then we don't need this change

mjanuchowski: Ok, I can revert the change and get rid of failing test. 
Still I would expect such function to work properly with all parameter combinations and not only with those that we can observe in a present code. :)

wluyima: I'm trying to go by this saying [don't try to fix it if ain't broke](http://www.phrases.org.uk/meanings/if-it-aint-broke-dont-fix-it.html), unless something is failing because this is missing then we will fix it at that point, as long as all reference to it in core set the parameters to true and the file name respectively it all good the way it is otherwise you risk introducing other obscure bugs, we don't expect external code to be calling this method.

api/src/test/java/org/openmrs/module/ModuleUtilTest.java
```diff
@@ -651,4 +653,120 @@ public void getResourceFromApi_shouldReturnNullIfFileIsNotFoundInApi() throws Ex
...
+		return new File(this.getClass().getResource("/org/openmrs/module/include/testJarExpand.omod").getFile());
+	}
+	
+	/**
+	 * Gets folder to which Jar should be extracted. 
+	 * 
+	 * @return <code>File</code> containing folder for Jar tests.
+	 */
+	protected File getEmptyJarDestinationFolder() throws IOException {
+		File destinationFolder = new File("/tmp/expandedJar");
```
dkayiwa: Does the path above work on windows? :)

mjanuchowski: Yes it does :) I've ran test on both Linux and Windows.

dkayiwa: I would need another windows user to run and confirm it. :)
Why don't you use System.getProperty("java.io.tmpdir")?

mjanuchowski: Good idea! I will use it. 
How can we get confirmation from another Windows user? :)

PR #1875, opened 2016-11-21
TRUNK-4958: Refactored OpenmrsUtil.copyFile method

lkukielka:
## Description

I replaced custom implementation of copy file in OpenmrsUtil.copyFile(...) with the apache commons IOUtils.copy(...). This is for reducing the code. I also added 3 tests for testing above functionality. I also decided to change method return type from void to boolean. It didnt change usage of method but helped with testing.

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4958

## Checklist:

### DIFF
--- a/api/src/main/java/org/openmrs/util/OpenmrsUtil.java
+++ b/api/src/main/java/org/openmrs/util/OpenmrsUtil.java
@@ -10,7 +10,6 @@
 package org.openmrs.util;
 
 import java.io.BufferedInputStream;
-import java.io.BufferedOutputStream;
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
@@ -66,6 +65,7 @@ import javax.xml.transform.TransformerFactory;
 import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.stream.StreamResult;
 
+import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -124,7 +124,7 @@ public class OpenmrsUtil {
 	private static Map<Locale, SimpleDateFormat> dateFormatCache = new HashMap<Locale, SimpleDateFormat>();
 	
 	private static Map<Locale, SimpleDateFormat> timeFormatCache = new HashMap<Locale, SimpleDateFormat>();
-		
+
 	/**
 	 * Compares origList to newList returning map of differences
 	 * 
@@ -270,47 +270,25 @@ public class OpenmrsUtil {
 	 * @param inputStream Stream to copy from
 	 * @param outputStream Stream/location to copy to
 	 * @throws IOException thrown if an error occurs during read/write
+	 * @should not copy the outputstream if outputstream is null
+	 * @should not copy the outputstream if inputstream is null
+	 * @should copy inputstream to outputstream and close the outputstream
 	 */
 	public static void copyFile(InputStream inputStream, OutputStream outputStream) throws IOException {
+
 		if (inputStream == null || outputStream == null) {
 			if (outputStream != null) {
-				try {
-					outputStream.close();
-				}
-				catch (Exception e) { /* pass */
-				}
+				IOUtils.closeQuietly(outputStream);
 			}
-			
 			return;
 		}
-		
-		InputStream in = null;
-		OutputStream out = null;
+
 		try {
-			in = new BufferedInputStream(inputStream);
-			out = new BufferedOutputStream(outputStream);
-			while (true) {
-				int data = in.read();
-				if (data == -1) {
-					break;
-				}
-				out.write(data);
-			}
+			IOUtils.copy(inputStream, outputStream);
+		} finally {
+			IOUtils.closeQuietly(outputStream);
 		}
-		finally {
-			if (in != null) {
-				in.close();
-			}
-			if (out != null) {
-				out.close();
-			}
-			try {
-				outputStream.close();
-			}
-			catch (Exception e) { /* pass */
-			}
-		}
-		
+
 	}
 	
 	/**
@@ -2210,4 +2188,5 @@ public class OpenmrsUtil {
 		return (c1.get(Calendar.ERA) == c2.get(Calendar.ERA) && c1.get(Calendar.YEAR) == c2.get(Calendar.YEAR) && c1
 		        .get(Calendar.DAY_OF_YEAR) == c2.get(Calendar.DAY_OF_YEAR));
 	}
+
 }
--- a/api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
+++ b/api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
@@ -10,14 +10,19 @@
 package org.openmrs.util;
 
 import static org.hamcrest.Matchers.is;
-import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 
+import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.net.URL;
 import java.nio.charset.Charset;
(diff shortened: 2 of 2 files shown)


api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -280,37 +280,19 @@ public static void copyFile(InputStream inputStream, OutputStream outputStream)
...
-			}
-			try {
-				outputStream.close();
-			}
-			catch (Exception e) { /* pass */
-			}
+
+		try (
+
+				AutoCloseable closeable = () -> IOUtils.copy(inputStream, outputStream)
```
teleivo: the javadoc says the "outputStream /is/ closed at completion of this method" the inputstream is not.

I dont think your implementation using the try-with-resource block closes the outputStream.

you could try:

```java
 try {
     IOUtils.copy(inputStream, outputStream);
 } finally {
     IOUtils.closeQuietly(outputStream);
 }
```

see https://commons.apache.org/proper/commons-io/javadocs/api-release/org/apache/commons/io/IOUtils.html

also please write test cases for all possible cases. for example the logic at the top where it closes the outputStream if the inputStream is null is not tested.

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -735,4 +735,20 @@ public void storeProperties_shouldEscapeSlashes() throws Exception {
 		
 		assertThat(actual.toByteArray(), is(expected.toByteArray()));
 	}
+
+	@Test
+	public void copyFile_shouldCopyFile() throws IOException {
```
teleivo: please refer to https://wiki.openmrs.org/display/docs/Unit+Tests
for the testing style used within OpenMRS. there is a plugin which can help you create the scaffolds for each test.

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -270,47 +270,26 @@ public static String getFileAsString(File file) throws IOException {
 	 * @param inputStream Stream to copy from
 	 * @param outputStream Stream/location to copy to
 	 * @throws IOException thrown if an error occurs during read/write
+	 * @return true if file copy was successful, false if was not
+	 * @should not copy the file when outputStream is null
+	 * @should not copy the file when inputStream is null
+	 * @should copy the file
 	 */
-	public static void copyFile(InputStream inputStream, OutputStream outputStream) throws IOException {
+	public static boolean copyFile(InputStream inputStream, OutputStream outputStream) throws IOException {
```
teleivo: dear @lkukielka changing the method signature was not part of the ticket. this would break compatibility with users of this method so please revert that change.

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -270,47 +270,26 @@ public static String getFileAsString(File file) throws IOException {
 	 * @param inputStream Stream to copy from
 	 * @param outputStream Stream/location to copy to
 	 * @throws IOException thrown if an error occurs during read/write
+	 * @return true if file copy was successful, false if was not
+	 * @should not copy the file when outputStream is null
+	 * @should not copy the file when inputStream is null
+	 * @should copy the file
```
teleivo: can you test that the outputstream was closed as well and you are not copying files although the method name says it (I know its misleading), but copying input to outputstream

I suggest tests like
"should close the outputstream if inputstream is null"
"should copy inputstream to outputstream and close the outputstream"

teleivo:
@lkukielka you should also squash your commits into one

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -270,47 +270,25 @@ public static String getFileAsString(File file) throws IOException {
...
-			}
-			try {
-				outputStream.close();
-			}
-			catch (Exception e) { /* pass */
-			}
+			IOUtils.copy(inputStream, outputStream);
+		} finally {
+			IOUtils.closeQuietly(outputStream);
+			return;
```
teleivo: no need for the return statement.

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -270,47 +270,25 @@ public static String getFileAsString(File file) throws IOException {
 	 * @param inputStream Stream to copy from
 	 * @param outputStream Stream/location to copy to
 	 * @throws IOException thrown if an error occurs during read/write
+	 * @should not copy the outputstream when outputstream is null
+	 * @should not copy the outputstream if inputstream is null
+	 * @should copy inputstream to outputstream
```
teleivo: you should also test that the outputstream is closed at completion.

so "should copy inputstream to outputstream and close the outputstream"

lkukielka: @teleivo Do you maybe have any idea how to check is the outputstream closed in this case?

teleivo: Since the test class is a `BaseContextSensitiveTest`  I dont think you can use Mockito to mock the `outputStream` which would make it easy to verify if it has been closed. A solution might be to override the close() of the  `outputStream` method so you can check that it was called see http://stackoverflow.com/a/13800353/3996784

@dkayiwa or @wluyima maybe you can add some more wisdom :smile:

dkayiwa: @teleivo i have no more wisdom than what you have already provided. :)

lkukielka:
@teleivo Is current solution good enough?

teleivo:
@lkukielka your implementation was good. My suggested changes were ment for the tests. So please revert your changes. The goal is: to test that the method closes the given output stream as it suggests in the javadocs. Since you control what output stream gets passed into the method in your tests you can pass in a mock or an output stream where you override the close method, so you notice that it was closed.

Just as a note: having global state as your current implementation has it is not a good solution. Imagine this method is used by different threads,  they would share the state but might mean different streams.

lkukielka:
@teleivo What do you think about current solution?

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -270,47 +270,25 @@ public static String getFileAsString(File file) throws IOException {
 	 * @param inputStream Stream to copy from
 	 * @param outputStream Stream/location to copy to
 	 * @throws IOException thrown if an error occurs during read/write
+	 * @should not copy the outputstream when outputstream is null
```
lkukielka: No, Its my mistake. Do you think is it necessary to fix it?

dkayiwa: @lkukielka what is the difference between the two? :)

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -735,4 +744,59 @@ public void storeProperties_shouldEscapeSlashes() throws Exception {
...
+	@Test
+	public void copyFile_shouldNotCopyTheOutputstreamWhenOutputstreamIsNull() throws Exception {
+		String exampleInputStreamString = "ExampleInputStream";
+		ByteArrayInputStream input = new ByteArrayInputStream(exampleInputStreamString.getBytes());
+
+		OutputStream output = null;
+
+		OpenmrsUtil.copyFile(input, output);
+
+		assertEquals(null, output);
```
dkayiwa: Why not use assertNull and assertNotNull?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -508,28 +517,28 @@ public void openmrsDateFormat_shouldNotAllowDatesWithInvalidDaysOrMonths() throw
 		try {
 			SimpleDateFormat sdf = OpenmrsUtil.getDateFormat(new Locale("en", "GB"));
 			sdf.parse("1/13/2001");
-			Assert.fail("Date with invalid month should throw exception.");
+			fail("Date with invalid month should throw exception.");
```
dkayiwa: Is switching from Assert.fail to fail part of the changes required for the ticket? Or did you just change it accidentally? :)

dkayiwa:
@lkukielka are you still working on this?

lkukielka:
@dkayiwa I will finish it tomorrow. Sorry for such a long delay.

lkukielka:
@dkayiwa Hi, I made above changes. Is it correct now?

dkayiwa:
@lkukielka i did not review because the ticket status says you are still doing some work on it.

lkukielka:
@dkayiwa I finished work on thursday and forgot to inform you. Sorry

dkayiwa:
@lkukielka you actually do not have to inform me. We have a process here that you need to follow: https://wiki.openmrs.org/display/docs/Pull+Request+Tips
Can you go through it and do the needful?

dkayiwa:
@lkukielka any response to the above? :)

lkukielka:
@dkayiwa Yes, using try-with-resource,in my opinion, was technically impossible in that case.

PR #1876, opened 2016-11-23
TRUNK-4971 Add caches programmatically - WORK IN PROGRESS

AdamGrzybkowski:
## Description

This PR introduce possibility of configuring cache in external modules.

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4971

## Checklist:


AdamGrzybkowski:
@rkorytkowski I've updated the PR. Tomorrow I will add some tests :)

api/src/main/java/org/openmrs/CachingConfig.java
```diff
@@ -0,0 +1,163 @@
...
+       return new EhCacheCacheManager(apiEhCacheManager());
+    }
+
+    @Bean(destroyMethod = "shutdown", name = "apiEhCacheManager")
+    public net.sf.ehcache.CacheManager apiEhCacheManager(){
+
+        net.sf.ehcache.config.Configuration config = ehCacheManagerFactoryBean().getObject().getConfiguration();
+
+        List<OpenmrsCacheConfiguration> cacheConfigurations = getCacheConfigurations();
+        cacheConfigurations.stream()
```
AdamGrzybkowski: @rkorytkowski Could you take a look here ? There is some infinite recurrency going on here and I can't find out why. The cacheManager() method is being called over and over...

rkorytkowski: Maybe it's http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/cache/annotation/CachingConfigurer.html#cacheManager--

Look cacheManger() has to be annotated with @Bean... In such a case we need to delete apiCacheManger method and annotate cacheManager with @Bean(name = "apiCacheManager")

rkorytkowski: Also it suggests to extend  CachingConfigurerSupport instead of implementing CachingConfigurer

api/src/main/java/org/openmrs/CachingConfig.java
```diff
@@ -0,0 +1,163 @@
...
+
+    @Override
+    public KeyGenerator keyGenerator() {
+        return new SimpleKeyGenerator();
+    }
+
+    @Override
+    public CacheErrorHandler errorHandler() {
+        return new SimpleCacheErrorHandler();
+    }
```
rkorytkowski: extend CachingConfigurerSupport and override only cacheManager()...

api/src/main/java/org/openmrs/CachingConfig.java
AdamGrzybkowski: Unfortunately this is not the problem... I'm almost sure it's being called in config.addCache()

AdamGrzybkowski: @rkorytkowski I did try to just set the cache config only from properties file so it looks almost like one to one with this http://stackoverflow.com/questions/21944202/using-ehcache-in-spring-4-without-xml and it also doesn't work with the same reason...

AdamGrzybkowski:
@rkorytkowski I've extracted the logic to a separate file to be able to unit test it but the error is still present. What's more with the current approach we are trying to modify cache config after creating one from ehcache-api.xml which we cannot do

AdamGrzybkowski:
@rkorytkowski Ok, I've modified the code so now we have two cache managers which prevents us from getting "you can't modify cache dynamically" but the problem with recurrency is still there.

rkorytkowski:
Thanks for the update. I reserved time tomorrow to look at it.

AdamGrzybkowski:
@rkorytkowski PR updated :)

PR #1901, opened 2016-12-15
TRUNK-4906 Adding unique constraint to cohort_member uuid

vshl:
TRUNK-4906 Adding unique constraint to cohort_member uuid
## Description

Adding unique constraint to the cohort_member column uuid
## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4906

## Checklist:

### DIFF
--- a/api/src/main/resources/liquibase-update-to-2.1.xml
+++ b/api/src/main/resources/liquibase-update-to-2.1.xml
@@ -266,9 +266,17 @@
 		<comment>Generating UUIDs for all rows in cohort_member table via built in uuid function.</comment>
 		<update tableName="cohort_member" >
 			<column name="uuid" valueNumeric="name-of-uuid-function" />
+			<where>uuid = '' OR uuid is null</where>
 		</update>
 		<modifySql dbms="mysql"><replace replace="name-of-uuid-function" with="UUID()"/></modifySql>
 		<modifySql dbms="oracle"><replace replace="name-of-uuid-function" with="SYS_GUID()"/></modifySql>
+	</changeSet>
+	<changeSet id="201610042145-2.2" author="vshankar">
+		<preConditions>
+			<columnExists tableName="cohort_member" columnName="uuid"/>
+		</preConditions>
+		<comment>Adding unique constraint to cohort_member.uuid</comment>
+		<addUniqueConstraint tableName="cohort_member" columnNames="uuid"/>
 	</changeSet>
 	<changeSet id="201610242135-1" author="vshankar">
 		<comment>Adding foreign key constraint to cohort_member.cohort_id</comment>


dkayiwa:
@vshl since this is not yet released, just modify the existing changeset by adding a unique=true attribute.

vshl:
@dkayiwa Doesn't seem like we can add NOT NULL and UNIQUE in the same ALTER query:

```
> ALTER TABLE `cohort_member` ADD `uuid` CHAR(38) NOT NULL UNIQUE;
ERROR 1062 (23000): Duplicate entry '' for key 'uuid'
```

I tried this in the command line as well.

vshl:
@dkayiwa These set of changesets work.

api/src/main/resources/liquibase-update-to-2.1.xml
```diff
@@ -257,7 +257,7 @@
 		<comment>Adding column cohort_member.uuid</comment>
 		<addColumn tableName="cohort_member">
 			<column name="uuid" type="char(38)">
-				<constraints nullable="false"/>
```
dkayiwa: Did you intentionally remove the nullable?

dkayiwa:
@vshl we have always just used unique=true and it works.
I do not expect it to fail for a brand new column like this, unless you are inserting duplicate uuid values. Can you investigate why you are having duplicate uuids?

vshl:
@dkayiwa `ALTER TABLE cohort_member ADD uuid CHAR(38) NOT NULL` creates a column with value **''** (empty string) for existing rows, so it fails while adding a unique constraint since it has duplicate values **''**. I tried this in the command-line replicating the steps multiple times.

Similar problem: http://stackoverflow.com/questions/17823322/1062-duplicate-entry-for-key-unique-id-when-trying-to-add-unique-key-my

vshl:
Here's a screenshot of adding the column on the command-line:

screen shot 2016-12-15 at 11 33 58 pm

dkayiwa:
@vshl can you run mvn clean install -Pintegration-test with these changes?

vshl:
Sorry, had to revert the change as my original commit. This works fine.

api/src/main/resources/liquibase-update-to-2.1.xml
```diff
@@ -266,10 +266,18 @@
 		<comment>Generating UUIDs for all rows in cohort_member table via built in uuid function.</comment>
 		<update tableName="cohort_member" >
 			<column name="uuid" valueNumeric="name-of-uuid-function" />
+			<where>uuid = ''</where>
```
vshl: In my tests, adding a column with nullable="false" adds an empty string to existing rows.

dkayiwa: How about databases which are not MySQL?

vshl: I really don't know :) Shouldn't it not have NULL values because nullable="false"?

vshl: Should I just remove the where clause?

dkayiwa: So when PostgreSQL or Oracle puts NULL, what will happen with your changes?

vshl: It will fail obviously, does appending an "OR uuid is null" work in this case?

dkayiwa: Oh ya! 👍

PR #1908, opened 2016-12-20
TRUNK-4925 Deep copy all fields of Person

mjanuchowski:
## Description
Changed `Person` copying constructor so that it makes deep copy of an object. This includes also sets of addresses, names and attributes. To achieve those `deepCopy()` functions were added to `PersonAddress`, `PersonName` and `PersonAttribute`. 

Fields of following data types are deep copied:
- String
- Integer
- Date
- Boolean
- Set and each element in it
- PersonAddress
- PersonName
- PersonAttribute

Fields of following data types are shallow copied:
- User
- Person
- PersonAttributeType
- Concept

Two new classes were added:
- `DeepCopyUtil` - provides helper functions to make deep copies of `String`, `Date`, `Integer`, and `Boolean` objects
- `DeepCopyUtilTestHelper` - via reflection tests if all private (non-static, non-final) fields of a class and its super classes are properly copied. This assures that if a new field is added to any of classes then unit test will fails. This will indicate that new field should be explicitly included in `deepCopy()` function.

## Related Issue

https://issues.openmrs.org/browse/TRUNK-4925

## Checklist:

### DIFF
--- a/api/src/main/java/org/openmrs/Person.java
+++ b/api/src/main/java/org/openmrs/Person.java
@@ -101,10 +101,9 @@ public class Person extends BaseOpenmrsData {
 	/**
 	 * This constructor is used to build a new Person object copy from another person object
 	 * (usually a patient or a user subobject). All attributes are copied over to the new object.
-	 * NOTE! All child collection objects are copied as pointers, each individual element is not
-	 * copied. <br>
 	 * 
 	 * @param person Person to create this person object from
+	 * @should deep copy all collections
 	 */
 	public Person(Person person) {
 		if (person == null) {
@@ -113,9 +112,10 @@ public class Person extends BaseOpenmrsData {
 		
 		personId = person.getPersonId();
 		setUuid(person.getUuid());
-		addresses = person.getAddresses();
-		names = person.getNames();
-		attributes = person.getAttributes();
+		
+		deepCopyAddresses(person.getAddresses());
+		deepCopyNames(person.getNames());
+		deepCopyAttributes(person.getAttributes());
 		
 		gender = person.getGender();
 		birthdate = person.getBirthdate();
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/util/DeepCopyUtil.java
```diff
@@ -0,0 +1,75 @@
...
+	 * @param dateSrc Date to be copied
+	 * @return deep copy of a given Date
+	 */
+	public static Date copy(Date dateSrc) {
+		if (dateSrc == null) {
+			return null;
+		}
+		
+		return (Date) dateSrc.clone();
+	}
```
dkayiwa: Do you really need to do this? The ticket is simply about copying child collection objects.

mjanuchowski: Ticket description says 
> if any of the original or copied object modifies the collection, the other's is not affected
that is why I'm making sure that every field of collection elements is a new instance. The same approach was already used in `PersonName.newInstance()`. If I leave out this copying then e.g. Date fields in elements of both collections will point to the same instances. And since Date is not immutable I thought I should handle it.

dkayiwa: That statement is referring to collection object references, not the collection members. That is, all those of type List or Set as the examples given on the ticket.

mjanuchowski: Ok :) Just to make it completely clear - I should make new instances of collections, but what about members of those collection (PersonName, PersonAddress, PersonAttribute)? Should I make new instances but without deep copy of members, or should I just add the same instances to new collections?

dkayiwa: @mjanuchowski you are right about deep copying the members like (PersonName, PersonAddress, PersonAttribute) 👍

mjanuchowski: Done, that made it muuuch simpler :)

dkayiwa:
@mjanuchowski before committing, did you run mvn clean install which automatically formats your changes according to our formatter?

api/src/test/java/org/openmrs/PersonTest.java
```diff
@@ -911,4 +911,39 @@ public PersonAddress build() {
 			return personAddress;
 		}
 	}
+
+	/**
+	 * @see Person#Person(Person)
+	 * @verifies deep copy all collections
+	 */
+	@Test
+	public void Person_shouldDeepCopyAllCollections() throws Exception {
```
dkayiwa: The convention is to start with lower case for method names. :)

mjanuchowski: I know :) but it was created like that by unit test plug-in. It's because it is testing copying consturctor, which starts with upper case. Maybe it should stay like that to match tested function name?

dkayiwa: OK

dkayiwa: How about the formatting?

mjanuchowski: Yes, I did. Even if I run it again, it does not change anything in those files. Do I miss some settings?

api/src/main/java/org/openmrs/Person.java
```diff
@@ -1055,4 +1055,48 @@ public void setId(Integer id) {
...
+	// Deep copy helpers
+	
+	/**
+	 * Creates deep copy of set of addresses.
+	 * 
+	 * @param personAddressesSrc Set of addresses to be copied.
+	 * @return Deep copy of given set of addresses.
+	 */
+	private void deepCopyAddresses(Set<PersonAddress> personAddressesSrc) {
+		for(PersonAddress personAddressSrc : personAddressesSrc) {
```
dkayiwa: I would have expected our formatter to put space after the above "for"

mjanuchowski: Well, it did not :/ I can fix it manually at push the change. Shall I?

dkayiwa: No. Its the formatter's job. Does this formatter make any difference? https://wiki.openmrs.org/display/docs/Coding+Conventions

mjanuchowski: Yes, now it got fixed. Thank you! Shall I push this formatting change? I ask because after your merge my commit will consist only of this additional formatting, it will not indicate my previous changes as I got them now from upstream.

dkayiwa:
@mjanuchowski just create a new pull request.

PR #1960, opened 2017-01-10
TRUNK-4868 added validation for dead and deathDate

piyush9620:
## Related Issue

## Checklist:

### DIFF
--- a/api/src/main/java/org/openmrs/validator/PersonValidator.java
+++ b/api/src/main/java/org/openmrs/validator/PersonValidator.java
@@ -94,7 +94,8 @@ public class PersonValidator implements Validator {
 				index++;
 			}
 		}
-		
+
+		validateDeathDateDead(errors, person.getDeathDate(), person.getDead());
 		validateBirthDate(errors, person.getBirthdate());
 		validateDeathDate(errors, person.getDeathDate());
 		
@@ -107,7 +108,36 @@ public class PersonValidator implements Validator {
 		
 		ValidateUtil.validateFieldLengths(errors, Person.class, "gender", "personVoidReason");
 	}
-	
+
+	/**
+	 * Checks if the death date is in accordance with dead status.
+	 *
+	 * @param errors Stores information about errors encountered during validation
+	 * @param deathDate the deathdate to validate
+	 * @param dead is the dead status
+	 */
+	private void validateDeathDateDead(Errors errors, Date deathDate, boolean dead) {
+		if(deathDate == null || dead == true) {
+			return;
+		}
+		rejectIfConflict(errors, deathDate, dead,"dead");
+	}
+
+	/**
+	 * Checks if the death date is in accordance with dead status.
+	 *
+	 * @param errors Stores information about errors encountered during validation
+	 * @param date the deathDate to validate
+	 * @param dead is the dead status
+	 * @param field the name of field
+	 */
+	private void rejectIfConflict(Errors errors, Date date, boolean dead, String field) {
+		if (date != null && dead == false) {
+			errors.rejectValue(field, "error.dead.false");
+		}
+	}
+
+
 	/**
 	 * Checks if the birth date specified is in the future or older than 120 years old..
 	 *
--- a/api/src/test/java/org/openmrs/validator/PersonValidatorTest.java
+++ b/api/src/test/java/org/openmrs/validator/PersonValidatorTest.java
@@ -76,7 +76,26 @@ public class PersonValidatorTest extends BaseContextSensitiveTest {
 		
 		Assert.assertTrue(errors.hasFieldErrors("deathDate"));
 	}
-	
+
+	/**
+	 * @see PersonValidator#validate(Object,Errors)
+	 * @verifies fail validation if deathDate is set if dead is set as false.
+	 */
+	@Test
+	public void validate_shouldFailValidationIfDeathDateSetAndDeadFalse() throws Exception {
+		Patient pa = new Patient(1);
+		Calendar date = Calendar.getInstance();
+		date.setTime(new Date());
+		date.add(Calendar.YEAR, -1);
+		pa.setDeathDate(date.getTime());
+		pa.setDead(false);
+		Errors errors = new BindException(pa, "patient");
+		validator.validate(pa, errors);
+
+		Assert.assertTrue(errors.hasFieldErrors("dead"));
+	}
+
+
 	/**
 	 * @see PersonValidator#validate(Object,Errors)
 	 * @verifies fail validation if birthdate makes patient older that 120 years old


dkayiwa:
@piyush9620 did you see the above travis failure?

piyush9620:
@dkayiwa 
i tried a lot but there is some or the other test failing or there are null pointer exceptions all around. can you please help me with this.

dkayiwa:
@piyush9620 can you first squash the commits into one?

piyush9620:
@dkayiwa  I squashed it as you asked, but I need your help with this. Can you please look if the approach i am taking is correct or not. Thanks.

api/src/main/java/org/openmrs/validator/PersonValidator.java
```diff
@@ -121,6 +122,19 @@ private void validateBirthDate(Errors errors, Date birthDate) {
...
+	/**
+	 * Checks if the death date is in accordance with dead status.
+	 *
+	 * @param errors Stores information about errors encountered during validation
+	 * @param deathDate the deathdate to validate
+	 * @param dead is the dead status
+	 */
+	private void validateDeathDateDead(Errors errors, Date deathDate, boolean dead, boolean deathDateEstimated) {
+		if(deathDate != null && dead == false && deathDateEstimated == false) {
+			errors.rejectValue("deathDate", "person.dead.false");
```
dkayiwa: What text does "person.dead.false" translate to?

api/src/main/java/org/openmrs/validator/PersonValidator.java
```diff
@@ -121,6 +122,19 @@ private void validateBirthDate(Errors errors, Date birthDate) {
...
+
+	/**
+	 * Checks if the death date is in accordance with dead status.
+	 *
+	 * @param errors Stores information about errors encountered during validation
+	 * @param deathDate the deathdate to validate
+	 * @param dead is the dead status
+	 */
+	private void validateDeathDateDead(Errors errors, Date deathDate, boolean dead, boolean deathDateEstimated) {
+		if(deathDate != null && dead == false && deathDateEstimated == false) {
```
dkayiwa: Can you put the above line in an english sentence? In other words, what is the if condition trying to test?

api/src/test/java/org/openmrs/api/PatientServiceTest.java
```diff
@@ -198,6 +198,7 @@ private Patient createBasicPatient() {
 		// patient.removeAddress(pAddress);
 		
 		patient.setDeathDate(new Date());
+		patient.setDead(true);
 		// patient.setCauseOfDeath("air");
 		patient.setBirthdate(new Date());
 		patient.setBirthdateEstimated(true);
```
dkayiwa: I would also expect you to add some new tests for the functionality you are putting in place.

api/src/main/java/org/openmrs/validator/PersonValidator.java
piyush9620: i was to specify that that the attribute dead of the person class was false and due to this the death date can not be set unless we set dead to true.

api/src/main/java/org/openmrs/validator/PersonValidator.java
piyush9620: the above statement checks if deathDate has some non-null value along with dead being set as false ie (the person is not dead but deathDate has been set) which is what we shouldn't allow according to the ticket. The third condition was added explicitly because of one test case in the PersonServeiceTest class. In that test, unlike other tests (where i manually set dead to true whenever deathdate was being set), setting dead to true was leading to a null pointer exception. so i added it. bcoz earlier what i tried was
private void validateDeathDateDead(Errors errors, Date deathDate, boolean dead, boolean deathDateEstimated) {
      if(dead == true || deathDate == null ) {
               return;
       }
       else if(dead == false && deathDate != null) {

       }

api/src/test/java/org/openmrs/api/PatientServiceTest.java
piyush9620: yes i completely understand, but without  ( patient.setDead(true);) this particular test shows validation exception as you see above that the deathdate has been set to current date (new Date())
which means we are setting the deathdate but not setting the person as dead which is a voilation of the issue in the ticket, isn't it?

piyush9620:
@dkayiwa 
i have added comments, please tell me if i am understanding thigs correctly. thanks :)

api/src/test/java/org/openmrs/api/PatientServiceTest.java
piyush9620: while this test runs the exception that my added code throws shows up.

api/src/main/java/org/openmrs/validator/PersonValidator.java
```diff
@@ -107,7 +108,36 @@ public void validate(Object target, Errors errors) {
...
+
+	/**
+	 * Checks if the death date is in accordance with dead status.
+	 *
+	 * @param errors Stores information about errors encountered during validation
+	 * @param deathDate the deathdate to validate
+	 * @param dead is the dead status
+	 */
+	private void validateDeathDateDead(Errors errors, Date deathDate, boolean dead) {
+		if(deathDate == null || dead == true) {
```
piyush9620: this line checks if deathDate is null or is dead true, if so, then we do not need to check further.
(the actual case we want to validate is dead == false and deathdate == null.)

dkayiwa: Would there be any problem if dead == false and deathdate == null?

piyush9620: @dkayiwa  there shouldn't be a problem , bcoz that way the person is not dead and no deathdate is being associated with it.
I noticed a thing in a lot of test cases that the deathdate has been initialized with new Date(), which return an instance of current date, which means that the deathdate is not null. also in these tests dead is not set to true. i think this may be a problem.

dkayiwa: I asked because of your response which was: (the actual case we want to validate is dead == false and deathdate == null.)

piyush9620: if the person is dead, we dont care if deathdate is set or not, if its set its fair, if not fair too.
if deathdate is null we dont care if person is dead or not.
but if deathdate is set the peron should surely be dead.

dkayiwa: And if the person is dead, the death date should not be null.

piyush9620: should a check for this be included too?
but i don't think even after including this the failing test case would pass.

dkayiwa: Without your changes, we have no failing tests. Which means that look at your changes and check for causes of failure to fix

piyush9620: yes i know that, but my point is some test cases have set deathDate without setting dead and also called the validate method in the person validator class. isn't this making them fail?

dkayiwa: Fix those tests by making them behave as expected. :)

piyush9620: ok, i will try and get back asap :)

dkayiwa: @piyush9620 are you still working on this?

dkayiwa:
@piyush9620 ping me to reopen this when you resume. :)

PR #1974, opened 2017-01-24
TRUNK-5010 Able to merge patients with discontinued orders

sidtharthanan:
## Related Issue

see https://issues.openmrs.org/browse/TRUNK-5010

## Checklist:

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -530,7 +530,7 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 		return rel.getRelationshipType().getRelationshipTypeId().toString() + (isA ? "A" : "B")
 		        + (isA ? rel.getPersonB().getPersonId().toString() : rel.getPersonA().getPersonId().toString());
 	}
-	
+
 	/**
 	 * 1) Moves object (encounters/obs) pointing to <code>nonPreferred</code> to
 	 * <code>preferred</code> 2) Copies data (gender/birthdate/names/ids/etc) from
@@ -549,12 +549,7 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 			log.debug("Merge operation cancelled: Cannot merge user" + preferred.getPatientId() + " to self");
 			throw new APIException("Patient.merge.cancelled", new Object[] { preferred.getPatientId() });
 		}
-		List<Order> orders = Context.getOrderService().getAllOrdersByPatient(notPreferred);
-		for (Order order : orders) {
-			if (!order.isVoided()) {
-				throw new APIException("Patient.cannot.merge", (Object[]) null);
-			}
-		}
+		requireNoActiveOrderOfSameType(preferred,notPreferred);
 		PersonMergeLogData mergedData = new PersonMergeLogData();
 		mergeVisits(preferred, notPreferred, mergedData);
 		mergeEncounters(preferred, notPreferred, mergedData);
@@ -593,6 +588,23 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 		Context.getPersonService().savePersonMergeLog(personMergeLog);
 	}
 	
+	private void requireNoActiveOrderOfSameType(Patient patient1, Patient patient2) {
+		String messageKey = "Patient.merge.cannotHaveSameTypeActiveOrders";
+		List<Order> ordersByPatient1 = Context.getOrderService().getAllOrdersByPatient(patient1);
+		List<Order> ordersByPatient2 = Context.getOrderService().getAllOrdersByPatient(patient2);
+		ordersByPatient1.forEach((Order order1) -> {
+			ordersByPatient2.forEach((Order order2) -> {
+				if (order1.isActive() && order2.isActive() && order1.getOrderType().equals(order2.getOrderType())) {
+					Object[] parameters = { patient1.getPatientId(), patient2.getPatientId(), order1.getOrderType() };
+					String message = Context.getMessageSourceService().getMessage(messageKey, parameters,
+							Context.getLocale());
+					log.debug(message);
+					throw new APIException(message);
+				}
+			});
+		});
+	}
+
 	private void mergeProgramEnrolments(Patient preferred, Patient notPreferred, PersonMergeLogData mergedData) {
 		// copy all program enrollments
 		ProgramWorkflowService programService = Context.getProgramWorkflowService();
--- a/api/src/main/resources/messages.properties
+++ b/api/src/main/resources/messages.properties
@@ -1667,6 +1667,7 @@ Patient.merge.notPreferred=Not Preferred
 Patient.merge.includeVoided=Include deleted
 Patient.merged=Patients merged successfully
 Patient.merge.patient.NoUnvoidedOrders=Not preferred patient(s) should not have any unvoided orders
+Patient.merge.cannotHaveSameTypeActiveOrders=Cannot merge patients. Both {0} and {1} patients have active [{2}] orders.
 Patient.id=Patient Id
 Patient.voidedMessage=This patient has been deleted.
 Patient.delete.warningMessage=<b><i>Are you sure you want to do this?</i></b><br/><br/>This will attempt to completely delete this patient from the system. This is usually the wrong thing to do: if you want to &quot;delete&quot; a patient then you should &quot;Delete&quot; them with the checkbox above.
--- a/api/src/test/java/org/openmrs/api/PatientServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/PatientServiceTest.java
@@ -52,6 +52,7 @@ import org.openmrs.GlobalProperty;
 import org.openmrs.Location;
 import org.openmrs.Obs;
 import org.openmrs.Order;
+import org.openmrs.OrderType;
 import org.openmrs.Patient;
 import org.openmrs.PatientIdentifier;
 import org.openmrs.PatientIdentifierType;
@@ -143,7 +144,17 @@ public class PatientServiceTest extends BaseContextSensitiveTest {
 			}
 		}
 	}
-	
+
+	private void voidOrdersForType(Collection<Patient> patients, OrderType ot) {
+		patients.forEach(patient -> {
+			Context.getOrderService().getAllOrdersByPatient(patient).forEach(order -> {
+				if(order.getOrderType().equals(ot)){
+					order.setVoided(true);
+				}
+			});
+		});
+	}
+
 	/**
 	 * @see PatientService#getAllIdentifierValidators()
 	 */
@@ -3304,15 +3315,43 @@ public class PatientServiceTest extends BaseContextSensitiveTest {
 	}
 	
 	/**
-	 * @verifies fail if not preferred patient has unvoided orders
+	 * @verifies fail if multiple patients has active order of same type
+	 * @see PatientService#mergePatients(org.openmrs.Patient, org.openmrs.Patient)
+	 * @throws APIException
+	 */
+	@Test
+	public void mergePatients_shouldFailIfMultiplePatientsHasActiveOrderOfSameType() throws Exception {
+		expectedException.expect(APIException.class);
+		String message = Context.getMessageSourceService().getMessage("Patient.merge.cannotHaveSameTypeActiveOrders",
+				new Object[] { "2", "7", "Drug order" }, Context.getLocale());
+		expectedException.expectMessage(Matchers.is(message));
+		Patient preferredPatient = patientService.getPatient(2);
+		Patient notPreferredPatient = patientService.getPatient(7);
+		patientService.mergePatients(preferredPatient, notPreferredPatient);
+	}
+
+	/**
+	 * @verifies not fail if one patient has active order
 	 * @see PatientService#mergePatients(org.openmrs.Patient, org.openmrs.Patient)
 	 */
 	@Test
-	public void mergePatients_shouldFailIfNotPreferredPatientHasUnvoidedOrders() throws Exception {
-		expectedException.expect(APIException.class);
-		expectedException.expectMessage(Matchers.is(Context.getMessageSourceService().getMessage("Patient.cannot.merge")));
-		Patient preferredPatient = patientService.getPatient(8);
+	public void mergePatients_shouldNotFailIfOnePatientHasActiveOrder() throws Exception {
+		Patient preferredPatient = patientService.getPatient(2);
 		Patient notPreferredPatient = patientService.getPatient(7);
+		voidOrders(Collections.singleton(notPreferredPatient));
+		patientService.mergePatients(preferredPatient, notPreferredPatient);
+	}
+
+	/**
+	 * @verifies not fail if multiple patients has active order of different types
+	 * @see PatientService#mergePatients(org.openmrs.Patient, org.openmrs.Patient)
+	 */
+	@Test
+	public void mergePatients_shouldNotFailIfMultiplePatientsHasActiveOrderOfDifferentTypes() throws Exception {
+		Patient preferredPatient = patientService.getPatient(2);
+		Patient notPreferredPatient = patientService.getPatient(7);
+		OrderType DrugOrder = Context.getOrderService().getOrderTypeByName("Drug order");
+		voidOrdersForType(Collections.singleton(preferredPatient), DrugOrder);
 		patientService.mergePatients(preferredPatient, notPreferredPatient);
 	}
 }
--- a/api/src/main/java/org/openmrs/api/db/hibernate/ImmutableOrderInterceptor.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/ImmutableOrderInterceptor.java
@@ -23,7 +23,7 @@ import org.springframework.stereotype.Component;
 public class ImmutableOrderInterceptor extends ImmutableEntityInterceptor {
 	
 	private static final String[] MUTABLE_PROPERTY_NAMES = new String[] { "dateStopped", "voided", "dateVoided", "voidedBy",
-	        "voidReason" };
+	        "voidReason", "patient" };
 	
 	/**
 	 * @see ImmutableEntityInterceptor#getSupportedType()


dkayiwa:
@sidtharthanan did you run mvn clean install before committing? The formatting does not look correct. :)

sidtharthanan:
I ran both the commands `mvn clean install` and `mvn clean package`.
I don't see any formatting changes after the `mvn clean...` commands.
If you could tell me the exact file, I can modify them.

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -592,7 +587,22 @@ public void mergePatients(Patient preferred, Patient notPreferred) throws APIExc
...
+	private void requireNoActiveOrderOfSameType(Patient patient1, Patient patient2) {
+		List<Order> ordersByPatient1 = Context.getOrderService().getAllOrdersByPatient(patient1);
+		List<Order> ordersByPatient2 = Context.getOrderService().getAllOrdersByPatient(patient2);
+		ordersByPatient1.forEach((Order order1)->{
+			ordersByPatient2.forEach((Order order2)->{
+				if(order1.isActive() && order2.isActive() && order1.getOrderType().equals(order2.getOrderType())){
+					log.debug("Merge operation cancelled: Cannot merge patient" + patient1.getPatientId()
+						+ " to patient" + patient2.getPatientId());
+					log.debug("Patients having active order of " + order1.getOrderType());
+					throw new APIException("Patient.cannot.merge", (Object[]) null);
```
dkayiwa: Does the exception message reflect the new business logic?

api/src/test/java/org/openmrs/api/PatientServiceTest.java
```diff
@@ -143,7 +144,17 @@ private void voidOrders(Collection<Patient> patientsWithOrders) {
 			}
 		}
 	}
-	
+
+	private void voidOrders(Collection<Patient> patients, OrderType ot) {
```
dkayiwa: Something like voidOrdersForType or anything else along those lines would make this method more meaningful.

dkayiwa:
@sidtharthanan did you take a look at the formatter here? https://wiki.openmrs.org/display/docs/Coding+Conventions

https://talk.openmrs.org/t/openmrs-code-formatter-styles/794

api/src/test/java/org/openmrs/api/PatientServiceTest.java
sidtharthanan: Okay. I will do that.

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
sidtharthanan: Yes, It means that patients are having active order of <certain type>. Hence, cannot merge.

sidtharthanan:
I read this: https://wiki.openmrs.org/display/docs/Coding+Conventions.
I think the code follows all of them.

I read this: https://talk.openmrs.org/t/openmrs-code-formatter-styles/794.
I don't get what to do.

I tried installing this following code formatter: https://github.com/suniala/openmrs-code-style-idea.
This formatter changes a lot of lines that I have not worked at all.
Could you tell me what formatter to use for intellij or command line?

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
dkayiwa: Did you get a chance to look at the actual text that this code represents in the messages.properties file? :)

sidtharthanan: Do I have to update the `"Patient.cannot.merge"` property just in the messages.properties file or all the *.properties files?
Some of those files have commented lines with `english text`.

dkayiwa: I would be in favour of a new message code that is descriptive enough to reflect the message. For message properties files, if you do not know the equivalent text in other languages, then you do not need to touch their message properties files.

sidtharthanan: `Patient.merge.bothHasActiveOrders=Cannot merge patients. Both {0} and {1} patients have active {2} orders.`

ex: `Cannot merge patients. Both 'COM238821' and 'BAH238591' patients have active 'Drug' orders.`

Something like this. Is it fine?

sidtharthanan:
@dkayiwa I have updated the PR as requested. Could check it?

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -593,6 +588,23 @@ public void mergePatients(Patient preferred, Patient notPreferred) throws APIExc
...
+		List<Order> ordersByPatient1 = Context.getOrderService().getAllOrdersByPatient(patient1);
+		List<Order> ordersByPatient2 = Context.getOrderService().getAllOrdersByPatient(patient2);
+		ordersByPatient1.forEach((Order order1) -> {
+			ordersByPatient2.forEach((Order order2) -> {
+				if (order1.isActive() && order2.isActive() && order1.getOrderType().equals(order2.getOrderType())) {
+					Object[] parameters = { patient1.getPatientId(), patient2.getPatientId(), order1.getOrderType() };
+					String message = Context.getMessageSourceService().getMessage(messageKey, parameters,
+							Context.getLocale());
+					log.debug(message);
+					throw new APIException(message);
```
dkayiwa: Why not just use this? new APIException(String messageKey, Object[] parameters)

api/src/main/resources/messages.properties
```diff
@@ -1667,6 +1667,7 @@ Patient.merge.notPreferred=Not Preferred
 Patient.merge.includeVoided=Include deleted
 Patient.merged=Patients merged successfully
 Patient.merge.patient.NoUnvoidedOrders=Not preferred patient(s) should not have any unvoided orders
+Patient.merge.bothHasActiveOrders=Cannot merge patients. Both {0} and {1} patients have active [{2}] orders.
```
dkayiwa: I would expect the error message to also reflect the second condition under which you are throwing the exception. The same applies to the message key.

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
sidtharthanan: Could you elaborate on `this`?

sidtharthanan: `String message = Context.getMessageSourceService().getMessage(messageKey, parameters,
							Context.getLocale());
					log.debug(message);
					throw new APIException(message);`
We're logging the message.

api/src/main/resources/messages.properties
sidtharthanan: Could you tell me what is the second condition?
The logic is that if `both the patients are having active orders of same type` the error is thrown.

dkayiwa: Let me ask it this way, would the merge fail if two patients simply have active orders?

sidtharthanan: If both the patients are having active orders of **same type** the error is thrown.

sidtharthanan: Lets say patient A has `1 Active Drug Order` and patient B has `1 Active Lab Order`.
We will allow to merge those patients.

dkayiwa: Does the messages.properties changes reflect that? :)

dkayiwa: I think i now see type reflected in the third parameter. :)

dkayiwa: How about? Patient.merge.sameTypeActiveOrders

sidtharthanan: An possible error message would be:
Cannot merge patients. Both BAH109245 and BAH239281 patients have active [Drug Order] orders.
I hope that message is clear.
It would be good, If you can tell a better message, If you have any.

dkayiwa: Yes that makes perfect sense! I have just noticed that the third parameter takes care of the order type. :)

sidtharthanan: Lets go with this: Patient.merge.bothHasSameTypeActiveOrders.
What do you think?

dkayiwa: How about cannotHaveSameTypeActiveOrders? :)

sidtharthanan: Okay. I'll go with that. `cannotHaveSameTypeActiveOrders`

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -593,6 +588,23 @@ public void mergePatients(Patient preferred, Patient notPreferred) throws APIExc
 		Context.getPersonService().savePersonMergeLog(personMergeLog);
 	}
 	
+	private void requireNoActiveOrderOfSameType(Patient patient1, Patient patient2) {
+		String messageKey = "Patient.merge.cannotHaveSameTypeActiveOrders";
+		List<Order> ordersByPatient1 = Context.getOrderService().getAllOrdersByPatient(patient1);
```
wluyima: There is a getActiveOrders() that you can call and simplify this code

api/src/test/java/org/openmrs/api/PatientServiceTest.java
```diff
@@ -3304,15 +3315,43 @@ public void purgePatientIdentifierType_shouldThrowErrorWhenTryingToDeleteAPatien
...
 	
 	/**
-	 * @verifies fail if not preferred patient has unvoided orders
+	 * @verifies fail if multiple patients has active order of same type
 	 * @see PatientService#mergePatients(org.openmrs.Patient, org.openmrs.Patient)
+	 * @throws APIException
 	 */
 	@Test
-	public void mergePatients_shouldFailIfNotPreferredPatientHasUnvoidedOrders() throws Exception {
+	public void mergePatients_shouldFailIfMultiplePatientsHasActiveOrderOfSameType() throws Exception {
```
wluyima: This method name is a little confusing, can you rephrase it please and actually it would be nice for it to mention drug order in the method name, how about `mergePatients_shouldFailIfBothPatientsHaveActiveDrugOrders`

api/src/main/resources/messages.properties
```diff
@@ -1667,6 +1667,7 @@ Patient.merge.notPreferred=Not Preferred
 Patient.merge.includeVoided=Include deleted
 Patient.merged=Patients merged successfully
 Patient.merge.patient.NoUnvoidedOrders=Not preferred patient(s) should not have any unvoided orders
+Patient.merge.cannotHaveSameTypeActiveOrders=Cannot merge patients. Both {0} and {1} patients have active [{2}] orders.
```
wluyima: You should have specific messages based on the error and not this generic one, you should have an error message when multiple patients have active drug orders

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
sidtharthanan: getActiveOrders() method **excludes** discontinued orders.
But, getAllOrdersByPatient **includes** discontinued orders.

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -593,6 +588,23 @@ public void mergePatients(Patient preferred, Patient notPreferred) throws APIExc
...
+		List<Order> ordersByPatient2 = Context.getOrderService().getAllOrdersByPatient(patient2);
+		ordersByPatient1.forEach((Order order1) -> {
+			ordersByPatient2.forEach((Order order2) -> {
+				if (order1.isActive() && order2.isActive() && order1.getOrderType().equals(order2.getOrderType())) {
```
sidtharthanan: Do you mean to say that:
Active 'Drug Order' of 'Same chemical' (for clarity) should not exist in both patients while merging?

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
wluyima: I thought that's the whole point of this ticket and logic, i.e to check if the patient has any active orders, discontinued orders are never active

wluyima: I guess the question here is do we also want to copy the medical history of the not preferred?

wluyima: Note that it's one thing to check if any of the patients has any active orders and it's another to copy over the orders, when checking if any of the patients has any active orders, getActiveOrders() is what you need to call, when moving over the orders to the preferred  patient, then you need to call getAllOrders, makes sense? In this method I believe you are checking if they have active orders and you should be calling getActiveOrder

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
sidtharthanan: [Could you have a look it](https://talk.openmrs.org/t/patient-merge-functionality/9310/16) and whether it clarifies some of your questions or not?
Specially this one: for other order types, you cannot merge patient records if both have active orders for the same concept
\*forgot the link

wluyima: Sorry1 I don't seem to understand your comment above

wluyima: Do you mind clarifying it please?

sidtharthanan: We have discussed some of the scenarios of patient merge in this [talk](https://talk.openmrs.org/t/patient-merge-functionality/9310).
I want to highlight this [part](https://talk.openmrs.org/t/patient-merge-functionality/9310/16).
We have decided the following:
 We will check only whether the patients have active orders of SAME ORDER TYPE or not.
 We will not check whether the patients have active orders of SAME CONCEPT or not, as of now.

\*I hope this is clear and forgive me for the previous incomprehensible comment.

wluyima: I'm also commenting based on the conclusion that was made from that discussion. As per the conclusion from that discussion, I don't think the intended behavior is well reflected by the unit tests, it could be because of the way they are named, and the code changes to me don't  seem explicitly address all 3 conditions. Can you try renaming the test methods so that they are more intuitive based on the 3 conditions that need to be met.

sidtharthanan: There may be a gap between our understandings, since there was further discussion on the openmrs talk after the design call on 11th of Jan. Could you have a look at it once more?
The second condition of the three conditions, you are referring to, we **didn't agree** to do.

wluyima: Well, based on the discussion it was a suggestion that was never confirmed, i have posted to the talk topic to get a green light from others. Once we get the green light, it would be nice to clean up the tests methods i.e rename them to be more intuitive to portray what they are testing more clearly and also to have more useful assertions so don't end passing simply because the test data is not well set up. E.g if a test says that it should fail if both patients have active drug orders, you need to have an assertion for each patient that ensures that they actually  have at least one drug order before proceeding with the unit test.

sidtharthanan: Great! Let us hear what people say and we will proceed.

api/src/test/java/org/openmrs/api/PatientServiceTest.java
wluyima: Can you fix this method name to say Patients**Have** and not Patient**Has**

api/src/test/java/org/openmrs/api/PatientServiceTest.java
```diff
@@ -3304,15 +3315,43 @@ public void purgePatientIdentifierType_shouldThrowErrorWhenTryingToDeleteAPatien
...
 		expectedException.expect(APIException.class);
-		expectedException.expectMessage(Matchers.is(Context.getMessageSourceService().getMessage("Patient.cannot.merge")));
-		Patient preferredPatient = patientService.getPatient(8);
+		String message = Context.getMessageSourceService().getMessage("Patient.merge.cannotHaveSameTypeActiveOrders",
+				new Object[] { "2", "7", "Drug order" }, Context.getLocale());
+		expectedException.expectMessage(Matchers.is(message));
+		Patient preferredPatient = patientService.getPatient(2);
+		Patient notPreferredPatient = patientService.getPatient(7);
```
wluyima: Before merging the patients, can you add an assertion here to act as a sanity check to ensure that actually both patients have active orders of the same order type, this is helpful in case someone alters the test data and the test  ends up passing simply because the test is not well setup.

api/src/test/java/org/openmrs/api/PatientServiceTest.java
```diff
@@ -3304,15 +3315,43 @@ public void purgePatientIdentifierType_shouldThrowErrorWhenTryingToDeleteAPatien
...
+
+	/**
+	 * @verifies not fail if one patient has active order
+	 * @see PatientService#mergePatients(org.openmrs.Patient, org.openmrs.Patient)
+	 */
+	@Test
+	public void mergePatients_shouldNotFailIfOnePatientHasActiveOrder() throws Exception {
+		Patient preferredPatient = patientService.getPatient(2);
+		Patient notPreferredPatient = patientService.getPatient(7);
+		voidOrders(Collections.singleton(notPreferredPatient));
```
wluyima: Same here add sanity checks which ensure that only one patient has active orders

api/src/test/java/org/openmrs/api/PatientServiceTest.java
```diff
@@ -3304,15 +3315,43 @@ public void purgePatientIdentifierType_shouldThrowErrorWhenTryingToDeleteAPatien
...
+	/**
+	 * @verifies not fail if multiple patients has active order of different types
+	 * @see PatientService#mergePatients(org.openmrs.Patient, org.openmrs.Patient)
+	 */
+	@Test
+	public void mergePatients_shouldNotFailIfMultiplePatientsHasActiveOrderOfDifferentTypes() throws Exception {
+		Patient preferredPatient = patientService.getPatient(2);
 		Patient notPreferredPatient = patientService.getPatient(7);
+		OrderType DrugOrder = Context.getOrderService().getOrderTypeByName("Drug order");
+		voidOrdersForType(Collections.singleton(preferredPatient), DrugOrder);
```
wluyima: Same here, add some sanity checks that ensure that actually both patients have active orders of different types

sidtharthanan:
@wluyima , I've made the updates you suggested. I raised a PR [here](https://github.com/openmrs/openmrs-core/pull/1988).

wluyima:
I've merged the new PR into 2.0.x and master, thanks @sidtharthanan

PR #1977, opened 2017-01-29
Add JUnit @should annotations according to code comments

hilz041:
## Description
I put a @should annotation describing what the unit test should confirm 

- @should return a criterion when identifierTypes is not empty
- @should return a list of patientIdentifiers when patientIdentifierTypes is not null
- @sholud return a list of patientIdentifiers when patient is not empty
- @should return criteria if containingAnyFormField any formField is not empty

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-248

## Checklist:

- [x ] My pull request only contains one single commit.
- [x ] My pull request is based on the latest master branch
- [x ] My code follows the code style of this project.
- [ x] All new and existing tests passed.

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
@@ -273,7 +273,8 @@ public class HibernatePatientDAO implements PatientDAO {
 
 		
 		if (!patients.isEmpty()) {
-
+			criteria.add(Restrictions.in("patient", patients));
+		}
 		
 		if (patients.size() > 0) {
 
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
@@ -239,6 +239,8 @@ public class HibernatePatientDAO implements PatientDAO {
 	
 	/**
 	 * @see org.openmrs.api.PatientService#getPatientIdentifiers(java.lang.String, java.util.List, java.util.List, java.util.List, java.lang.Boolean)
+	 * @should return a list of patientIdentifiers when patientIdentifierTypes is not null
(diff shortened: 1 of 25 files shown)


hilz041:
TRUNK-248 Add JUnit @should annotations according to code comments.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -268,8 +270,13 @@ public void deletePatientIdentifierType(PatientIdentifierType patientIdentifierT
 			criteria.add(Restrictions.in("location", locations));
 		}
 		
-		// TODO add junit test for getting by patients
+
+		
 		if (!patients.isEmpty()) {
```
ivange94: Compilation error. You opened a block but did not close it.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -268,8 +270,13 @@ public void deletePatientIdentifierType(PatientIdentifierType patientIdentifierT
...
+
+		
+		if (patients.size() > 0) {
```
ivange94: Why do you think this is better than calling isEmpty()? This can be actually very inefficient as you'll have to compute the size of the list which might be expensive.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
ivange94: You should definitely use an IDE. A good IDE will never let this escape. :)

hilz041: @ivange94 thanks alot let me look into it and clarify.

dkayiwa:
@hilz041 because squashing is giving you trouble, since the changes are small, just create a new branch and a brand new pull request. Am sure it will be easier for you that way. :)

PR #1985, opened 2017-02-06
TRUNK-5027 - Added MIME types support for Data handlers

tmarzeion:
## Description
This change makes every Data Handler able to guess Complex Data File's MIME type (it had fixed value before), so it will be possible to add support for MIME types when posting Complex Obs.
See this OpenMRS Talk topic for more details about this change:
https://talk.openmrs.org/t/support-for-mime-types-core-issue/9864
## Related Issues
see https://issues.openmrs.org/browse/TRUNK-5027
and https://issues.openmrs.org/browse/RESTWS-641


tmarzeion:
@dkayiwa Could You take a look on this PR? :)

dkayiwa:
@tmarzeion did you take a look at this? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

tmarzeion:
@dkayiwa Yup

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -67,37 +70,47 @@ public File getOutputFileToWrite(Obs obs) throws IOException {
...
+			String mimetype = "";
+			if (StringUtils.isNotBlank(obs.getComplexData().getMimeType())) {
+				mimetype = "_" + obs.getComplexData().getMimeType().replace("/", "-slash-");
+			}
+			
+			// If the Obs does not exist, but the File does, append a two-digit
+			// count number to the filename and save it.
+			String twoDigitNumber = "";
+			if (new File(tmp + mimetype + "." + extension).exists()) {
+				for (int i = 1; obs.getObsId() == null && i < 100; i++) {
```
dkayiwa: Is this change also about mime type?

tmarzeion: @dkayiwa This loop is responsible for finding two digit number (which is useful when there are multiple files with same name). It was solution that was implemented here before, I've only upgraded it to handle MIMEs in filename :)

dkayiwa: @tmarzeion am a bit confused by this change because it seems to do more than i would expect. For instance, instance, can you throw some more light as to why you removed the while loop?

tmarzeion: @dkayiwa Okay, so:
Generally, `getOutputFileToWrite` is about creating new blank file with proper filename that contain some information about data is would get.
Before it was like: `TITLE + NUMBER + EXTENSION` (`NUMBER` is appended when file with same `TITLE` already exists.
Now it is like: `TITLE + MIME + NUMBER + EXTENSION` (`MIME` is appended when it is defined in `Obs`)

You are asking about `while` loop that I replaced with `for` loop, explanation:
- `for` looks better, since we don't need to initialize iterator out of the loop.
- this loop is responsible for finding proper `NUMBER` that would be appended to file if it already exists.
- after number is found (no file named `TITLE + MIME + NUMBER + EXTENSION` exists) it goes out of the loop and create returns file with given `TITLE + MIME + NUMBER + EXTENSION`.

Isn't it more readable than it was before? :smiley:

dkayiwa: Am please very sorry to make you go back and forth regarding this. But am just careful not to accidentally break existing code. I would prefer is the refactoring is a separate pull request from this who purpose is to add mime type. My thinking is that the code would not have to change this much for just the mime type addition. I would expect the mime type generated and then this line modified to include it: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java#L96

dkayiwa: I also noticed that Files.probeContentType(file.toPath()) always returns null for the mime type.
Did you notice it too?

tmarzeion: @dkayiwa Sorry for long response,
I've minimized changes in `getOutputFileToWrite()` as You said.
I don't have any problems with `Files.probeContentType(file.toPath())`, for example when running `shouldGuessMimeTypeIfNotSpecified()` test it is guessing type just fine :)

tmarzeion:
@dkayiwa I've applied changes according to Your review :)
Please take a look on that when You will have some time :)

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -78,10 +78,14 @@ public File getOutputFileToWrite(Obs obs) throws IOException {
 			outputfile = new File(dir, title + "." + extension);
 			// outputfile = new File(dir, title);
 		}
-		
+		// Get MIME type
+		String mimetype = "";
+		if (StringUtils.isNotBlank(obs.getComplexData().getMimeType())) {
+			mimetype = "_" + obs.getComplexData().getMimeType().replace("/", "-slash-");
```
dkayiwa: Will windows clients have "/" or "\"?

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -119,6 +121,35 @@ public String getExtension(String filename) {
 		return extension;
 	}
 	
+	public String getMimeTypeFromFile(File file) {
+		if (file == null) {
+			return null;
+		}
```
dkayiwa: How do you compare this method with OpenmrsUtil.getFileMimeType(file)?

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
tmarzeion: It doesn't matter since this slash is for parsing MIME type and filename in both UNIX and Windows will look like "`somefile_image-slash-jpg`".

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
tmarzeion: `getMimeTypeFromFile` is more complex since it tries to parse MIME type from filename, then if there is no MIME type in file it tries to probe it and returns default MIME type (`application/octet-stream`) if it couldn't guess MIME by any way mentioned before.
`getFileMimeType` is just trying to guess MIME type by using `MimetypesFileTypeMap.getContentType()` which is not enough and it is less reliable than `Files.probeContentType()`

dkayiwa:
@tmarzeion are you still working on this? By the way, did you see the merge conflicts?

dkayiwa:
@tmarzeion will reopen when you get back to finish it up.

PR #1997, opened 2017-02-15
TRUNK-4916 Validators should be called when unretiring/unvoiding an object

jtatia:
made changes to unretireXXX() and unvoidXXX() methods such that they call save methods on  spring proxy object obtained from calling Context.getXXXService().
see https://issues.openmrs.org/browse/TRUNK-4916

- [x] All existing tests passed.

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
@@ -283,7 +283,7 @@ public class ObsServiceImpl extends BaseOpenmrsService implements ObsService {
 	 */
 	@Override
 	public Obs unvoidObs(Obs obs) throws APIException {
-		return dao.saveObs(obs);
+		return Context.getObsService().saveObs(obs,"unvoid obs");
 	}
 	
 	/**
--- a/api/src/test/java/org/openmrs/api/OrderServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/OrderServiceTest.java
@@ -1343,22 +1343,23 @@ public class OrderServiceTest extends BaseContextSensitiveTest {
 	
 	@Test
 	public void unretireOrderFrequency_shouldUnretireGivenOrderFrequency() {
-		OrderFrequency orderFrequency = orderService.getOrderFrequency(3);
+		executeDataSet("org/openmrs/api/include/OrderServiceTest-otherOrderFrequencies.xml");
+		OrderFrequency orderFrequency = orderService.getOrderFrequency(103);
 		assertNotNull(orderFrequency);
 		assertTrue(orderFrequency.getRetired());
-		assertEquals("Some Retire Reason", orderFrequency.getRetireReason());
+		assertEquals(null,orderFrequency.getRetireReason());
 		assertNotNull(orderFrequency.getDateRetired());
-		
+
 		orderService.unretireOrderFrequency(orderFrequency);
 		
-		orderFrequency = orderService.getOrderFrequency(3);
+		orderFrequency = orderService.getOrderFrequency(103);
 		assertNotNull(orderFrequency);
 		assertFalse(orderFrequency.getRetired());
 		assertNull(orderFrequency.getRetireReason());
 		assertNull(orderFrequency.getDateRetired());
-		
+
 		//Should not change the number of order frequencies.
-		assertEquals(3, orderService.getOrderFrequencies(true).size());
+		assertEquals(7, orderService.getOrderFrequencies(true).size());
 	}
 	
 	@Test
--- a/api/src/test/java/org/openmrs/api/ProviderServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/ProviderServiceTest.java
@@ -294,7 +294,7 @@ public class ProviderServiceTest extends BaseContextSensitiveTest {
 	 */
 	@Test
 	public void unretireProvider_shouldUnretireAProvider() {
-		Provider provider = service.getProvider(1);
+		Provider provider = service.getProvider(2);
 		service.unretireProvider(provider);
 		assertFalse(provider.getRetired());
 		assertNull(provider.getRetireReason());
--- a/api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
@@ -302,7 +302,7 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 	 */
 	@Override
 	public Drug unretireDrug(Drug drug) throws APIException {
-		return dao.saveDrug(drug);
+		return Context.getConceptService().saveDrug(drug);
 	}
 	
 	/**
@@ -1649,7 +1649,7 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 	 */
 	@Override
 	public ConceptMapType unretireConceptMapType(ConceptMapType conceptMapType) throws APIException {
-		return dao.saveConceptMapType(conceptMapType);
+		return Context.getConceptService().saveConceptMapType(conceptMapType);
 	}
 	
 	/**
@@ -1751,7 +1751,7 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 	 */
 	@Override
 	public ConceptReferenceTerm unretireConceptReferenceTerm(ConceptReferenceTerm conceptReferenceTerm) throws APIException {
-		return dao.saveConceptReferenceTerm(conceptReferenceTerm);
+		return Context.getConceptService().saveConceptReferenceTerm(conceptReferenceTerm);
 	}
 	
 	/**
@@ -1968,7 +1968,7 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 	 */
 	@Override
 	public ConceptAttributeType unretireConceptAttributeType(ConceptAttributeType conceptAttributeType) {
-		return dao.saveConceptAttributeType(conceptAttributeType);
+		return Context.getConceptService().saveConceptAttributeType(conceptAttributeType);
 	}
 
 	/**
--- a/api/src/main/java/org/openmrs/api/impl/LocationServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/LocationServiceImpl.java
@@ -441,7 +441,7 @@ public class LocationServiceImpl extends BaseOpenmrsService implements LocationS
 	 */
 	@Override
 	public LocationAttributeType unretireLocationAttributeType(LocationAttributeType locationAttributeType) {
-		return dao.saveLocationAttributeType(locationAttributeType);
+		return Context.getLocationService().saveLocationAttributeType(locationAttributeType);
 	}
 	
 	/**
--- a/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
@@ -812,7 +812,7 @@ public class OrderServiceImpl extends BaseOpenmrsService implements OrderService
 	 */
 	@Override
 	public OrderFrequency unretireOrderFrequency(OrderFrequency orderFrequency) {
-		return dao.saveOrderFrequency(orderFrequency);
+		return Context.getOrderService().saveOrderFrequency(orderFrequency);
 	}
 	
 	/**
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -345,7 +345,7 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 		
 		// patient and patientidentifier attributes taken care of by the BaseUnvoidHandler
 		
-		return dao.savePatient(patient);
+		return Context.getPatientService().savePatient(patient);
 	}
 	
 	/**
--- a/api/src/main/java/org/openmrs/api/impl/PersonServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PersonServiceImpl.java
@@ -228,8 +228,8 @@ public class PersonServiceImpl extends BaseOpenmrsService implements PersonServi
 		type.setDateRetired(null);
 		type.setRetiredBy(null);
 		type.setRetireReason(null);
-		dao.savePersonAttributeType(type);
-		
+		Context.getPersonService().savePersonAttributeType(type);
+
 	}
 	
 	/**
@@ -357,7 +357,7 @@ public class PersonServiceImpl extends BaseOpenmrsService implements PersonServi
 			return null;
 		}
 		
-		return dao.savePerson(person);
+		return Context.getPersonService().savePerson(person);
 	}
 	
 	/**
--- a/api/src/main/java/org/openmrs/api/impl/ProviderServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ProviderServiceImpl.java
@@ -79,7 +79,7 @@ public class ProviderServiceImpl extends BaseOpenmrsService implements ProviderS
 	 */
 	@Override
 	public Provider unretireProvider(Provider provider) {
-		return dao.saveProvider(provider);
+		return Context.getProviderService().saveProvider(provider);
 	}
 	
 	/**
--- a/api/src/main/java/org/openmrs/api/impl/VisitServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/VisitServiceImpl.java
@@ -206,7 +206,7 @@ public class VisitServiceImpl extends BaseOpenmrsService implements VisitService
 	 */
 	@Override
 	public Visit unvoidVisit(Visit visit) throws APIException {
-		return dao.saveVisit(visit);
+		return Context.getVisitService().saveVisit(visit);
 	}
 	
 	/**
@@ -326,7 +326,7 @@ public class VisitServiceImpl extends BaseOpenmrsService implements VisitService
 	 */
 	@Override
 	public VisitAttributeType unretireVisitAttributeType(VisitAttributeType visitAttributeType) {
-		return dao.saveVisitAttributeType(visitAttributeType);
+		return Context.getVisitService().saveVisitAttributeType(visitAttributeType);
 	}
 	
 	/**


dkayiwa:
Is this all that you found in all the services?

jtatia:
most of them have already been changed. 2 of them were giving validation exception.  others were calling a method from within the same class.

api/src/test/java/org/openmrs/api/ProviderServiceTest.java
```diff
@@ -317,7 +317,7 @@ public void saveProviderAttributeType_shouldSaveTheProviderAttributeType() throw
 	 */
 	@Test
 	public void unretireProvider_shouldUnretireAProvider() throws Exception {
-		Provider provider = service.getProvider(1);
+		Provider provider = service.getProvider(2);
```
dkayiwa: Do you mind explaining a bit why this changed necessitated a switch from 1 to 2 for provider id?

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -1414,22 +1414,20 @@ public void retireOrderFrequency_shouldRetireGivenOrderFrequency() throws Except
 	@Test
 	@Verifies(value = "should unretire given order frequency", method = "unretireOrderFrequency(OrderFrequency)")
 	public void unretireOrderFrequency_shouldUnretireGivenOrderFrequency() throws Exception {
-		OrderFrequency orderFrequency = orderService.getOrderFrequency(3);
+		executeDataSet("org/openmrs/api/include/OrderServiceTest-otherOrderFrequencies.xml");
+		OrderFrequency orderFrequency = orderService.getOrderFrequency(103);
 		assertNotNull(orderFrequency);
 		assertTrue(orderFrequency.isRetired());
```
dkayiwa: I though the ticket changes would not require modifications in this test. Not so?

api/src/test/java/org/openmrs/api/ProviderServiceTest.java
jtatia: Provider id 1 does not have a provider/person name because of which its giving a validation exception. 
log:
[Provider: providerId:1 providerName: ]
[Provider: providerId:2 providerName:[Mr. Horatio Test Hornblower, Mr. Horatio Peeter Hornblower2, Mr. John Peeter Hornblower3, Mr. adolf hitler] ]

org.openmrs.api.ValidationException: '[Provider: providerId:1 providerName: ]' failed to validate with reason: name: Provider Name or Person required, person: Provider Name or Person required

api/src/test/java/org/openmrs/api/OrderServiceTest.java
jtatia: before this ticket we were not validating on unretiring, and hence had skipped on some exceptions. On making changes to the method according to the ticket, these exceptions were caught (maybe due to some data missing from that particular set). Hence i switched to new id's

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -1414,22 +1414,20 @@ public void retireOrderFrequency_shouldRetireGivenOrderFrequency() throws Except
...
 		
-		orderFrequency = orderService.getOrderFrequency(3);
+		orderFrequency = orderService.getOrderFrequency(103);
 		assertNotNull(orderFrequency);
 		assertFalse(orderFrequency.isRetired());
 		assertNull(orderFrequency.getRetireReason());
 		assertNull(orderFrequency.getDateRetired());
-		
-		//Should not change the number of order frequencies.
-		assertEquals(3, orderService.getOrderFrequencies(true).size());
```
dkayiwa: Can you also share some light on why the above assertion was removed?

jtatia: sorry about that.  I have made the relevant changes and added the assert statement again

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -1414,22 +1414,23 @@ public void retireOrderFrequency_shouldRetireGivenOrderFrequency() throws Except
...
+		orderFrequency = orderService.getOrderFrequency(103);
 		assertNotNull(orderFrequency);
 		assertFalse(orderFrequency.isRetired());
 		assertNull(orderFrequency.getRetireReason());
 		assertNull(orderFrequency.getDateRetired());
-		
+
 		//Should not change the number of order frequencies.
-		assertEquals(3, orderService.getOrderFrequencies(true).size());
+    	assertEquals(7, orderService.getOrderFrequencies(true).size());
```
dkayiwa: The above formatting (TAB) does not look correct. Could you also explain why you changed from 3 to 7?

jtatia: adding an new dataset increased the overall size of order frequencies from 3 to 7.

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -276,7 +276,7 @@ public Obs voidObs(Obs obs, String reason) throws APIException {
 	 * @throws APIException
 	 */
 	public Obs unvoidObs(Obs obs) throws APIException {
-		return dao.saveObs(obs);
+		return Context.getObsService().saveObs(obs,"unretire obs");
```
dkayiwa: unretire or unvoid?

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -1414,22 +1414,23 @@ public void retireOrderFrequency_shouldRetireGivenOrderFrequency() throws Except
 	@Test
 	@Verifies(value = "should unretire given order frequency", method = "unretireOrderFrequency(OrderFrequency)")
 	public void unretireOrderFrequency_shouldUnretireGivenOrderFrequency() throws Exception {
-		OrderFrequency orderFrequency = orderService.getOrderFrequency(3);
+		executeDataSet("org/openmrs/api/include/OrderServiceTest-otherOrderFrequencies.xml");
+		OrderFrequency orderFrequency = orderService.getOrderFrequency(103);
 		assertNotNull(orderFrequency);
 		assertTrue(orderFrequency.isRetired());
-		assertEquals("Some Retire Reason", orderFrequency.getRetireReason());
```
dkayiwa: Did you just accidentally remove the above assertion?

jtatia: no. Since the id was changed so was the retire reason associated with it.

dkayiwa: This is problematic because you are changing the expected behaviour. The assertion ensured that the retire reason should of an expected text, which is different from just ensuring that it is not null.

jtatia: I have made a small change but the thing is that we are using an already existing dataset and this orderFrequency has been set retired without any reason. 
The orignal orderFrequency = 3 was failing with below exception on validation at unretiring

org.openmrs.api.ValidationException: 'CIVIL STATUS' failed to validate with reason: concept: Concept should be of class Frequency

dkayiwa: Can't you make the change to ensure that the existing test passes as it was?

jtatia: I had a look at https://github.com/openmrs/openmrs-core/blob/9b977a18fcc759af9ad782110b18efe27175dd76/api/src/test/resources/org/openmrs/include/standardTestDataset.xml  where the original dataset was located. The reason order frequency 3 is giving an error is because its class_id is not defined as 19 (line 19) i.e the the concept of order frequency 3 from the original dataset id not of class frequency.   On the other hand the one i have replaced it with (from file:///home/jthacker/openmrs-core/api/target/test-classes/org/openmrs/api/include/OrderServiceTest-otherOrderFrequencies.xml )  order frequency 103, its concept has class_id =19

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -278,7 +278,7 @@ public Obs voidObs(Obs obs, String reason) throws APIException {
 	 * @throws APIException
 	 */
 	public Obs unvoidObs(Obs obs) throws APIException {
-		return dao.saveObs(obs);
+		return Context.getObsService().saveObs(obs,"unvoid obs");
```
wluyima: We might want to have an exception for Obs, there is some complex logic around updating Obs and this may introduce some obscure behavior.

api/src/test/java/org/openmrs/api/OrderServiceTest.java
wluyima: That seems to make sense, now that validation is kicking in, you might need that extra text data

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
jtatia: should i revert the changes

api/src/test/java/org/openmrs/api/OrderServiceTest.java
wluyima: Though the question becomes, what is in that file that is required for validation? I thought the file contains just other frequencies along with their concepts. Does it fail because there is no backing concept?

api/src/test/java/org/openmrs/api/OrderServiceTest.java
wluyima: unretireOrderFrequency should unset the reason, so the assertion should instead check that the reason if blank.

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -1414,22 +1414,23 @@ public void retireOrderFrequency_shouldRetireGivenOrderFrequency() throws Except
...
+		assertEquals(null,orderFrequency.getRetireReason());
```
jtatia: @wluyima  yeah i have already changed the assertion

api/src/test/java/org/openmrs/api/OrderServiceTest.java
jtatia: @wluyima  yes. it seems that the original data was failing because the class_id of the backing concept was not set to class frequency

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
wluyima: I guess the question is what was the behavior after your change? 
I.e if you void and unvoid an Obs with no other changes in both operations, do you shouldn't get an extra 'replacment' obs for the old one, but if there in any other changes then it should get voided and replaced with a new one. The tricky thing is how to know if there were changes when it was initially getting voided. This is why one might argue that may be we should have your change, I guess we need to know what the behavior is.

api/src/test/java/org/openmrs/api/OrderServiceTest.java
wluyima: Does that mean the frequency is backed up by a concept that has the same concept_id from the standard test dataset and then the order frequency dataset is overwriting it (adding a concept with  same concept_id) with the corrrect class_id?

jtatia: in the standard data set concept_class_id="19" has been set for order Frequency (https://github.com/openmrs/openmrs-core/blob/9b977a18fcc759af9ad782110b18efe27175dd76/api/src/test/resources/org/openmrs/include/standardTestDataset.xml#L35)   and the backing concept (L172) for order frequency 3 (L342 of standardTestDataset.xml) has not been assigned this class_id because of which a validation exception was caused.
In the new dataset a new concept with new concept id has been created that has class_id ="19". This is the backing concept for orderFrequency 103 that i have switched to

api/src/test/java/org/openmrs/api/OrderServiceTest.java
wluyima: My apologies, I've just realized that you're now using a new frequency with id 103, so it makes sense to include that file. Too early here for me to be tired :)

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
jtatia: Its true that we can update voided obs and cannot track such a change. To track such a change maybe we could change isDirty() property of the obs to false on voiding it and change it to true if a voided obs is modified. Then we could keep a check that if an obs is voided and is marked dirty means than it means has been modified.  Anyway I do think we should validate an obs obs before unvoiding it. The fact whether an obs has been modified after voiding or not should not affect that.(the current test still seem to pass) This is what I understand. What do you think @wluyima @dkayiwa :)

wluyima: What is the current behavior with your changes?

wluyima: What I want to know is the behavior for the scenarios below after your changes;

- One voids and unvoid an obs with no changes.
- One voids an obs with changes and then unvoids it, there is a ticket about discarding any changes when voiding an item which would partly address this particular scenario.
- One voids an obs with no changes and unvoids it with changes.

I guess the question is what happens to the unvoided obs and is there any new Obs getting created to repalce it?

jtatia: behavior of change 
This is what I observed after writing a few tests.
Before the change:-
on voiding and unvoiding an obs without any change no new obs is created.
On voiding an obs with change and then unvoiding it, the change is not considered and thus yet again no new obs is created and the same obs gets unvoided. The same thing happens when an obs without any change is voided and then unvoided with changes.
After the change:- 
on voiding and unvoiding an obs without any change no new obs is created.
On voiding an obs with change and then unvoiding it, the change is now considered, and hence a new obs is created on unvoiding and this change is carried to the groupmembers of the obs. Same thing happens when an obs is voided and then unvoided with changes.
Hence as @wluyima said  there is obscure behavior. How should i proceed with the ticket ?
(Here are the test that i added to ObsServiceTest.java http://pastebin.com/J6FTY1rx, please check and let me know whether the test are wrong) :)

jtatia: In the test :- 
before the change:- all the new tests pass
after the test:- the last 2 new tests fail since new obs have been created to replace the old one

wluyima: Thanks @jtatia , that's interesting, when you void an Obs with changes and unvoid it later, I wonder how the code knows about the previous change that was made when voiding it, might be because you did this in a unit test and the same object is still in the the hibernate session, I think the way tests are setup can they can mislead sometimes. Do you mind making these checks through the application and not unit tests please?

jtatia: @wluyima  I did what u asked. I repeated the steps locally in the application and it turns out that even after the change no new Obs is created when an obs is voided with changes and then unvoided. Same with an obs being voided without changes and then being unvoided with changes. I have recorded the screenshots of what i performed here https://drive.google.com/drive/folders/0B54Y2dWNGuH1Y0VXbVgtZG5yVFE?usp=sharing.
I repeated the with_change procedure this time changing the actual parameters like height and weight when voiding and unvoiding and the changes didn't get saved :-) please let me know whether i followed the correct procedure

wluyima: I see, I guess it would be nice to discuss this on a design call about what the proper behavior should be, I would Imagine that if an Obs is voided and later unvoided with changes as it was getting voided or unvoided that would technically constitute a change but it would fairly be complex to track the change.

wluyima: If the behavior before and after your change is exactly the same for those scenarios, we can merge the PR the way it is and if we want any other behavior we implement it in a separate ticket

dkayiwa:
@jtatia did you see the merge conflicts?

jtatia:
merge conflict resolved. pr ready to be merged :)

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -1407,22 +1407,23 @@ public void retireOrderFrequency_shouldRetireGivenOrderFrequency() throws Except
 	@Test
 	@Verifies(value = "should unretire given order frequency", method = "unretireOrderFrequency(OrderFrequency)")
 	public void unretireOrderFrequency_shouldUnretireGivenOrderFrequency() throws Exception {
-		OrderFrequency orderFrequency = orderService.getOrderFrequency(3);
+		executeDataSet("org/openmrs/api/include/OrderServiceTest-otherOrderFrequencies.xml");
+		OrderFrequency orderFrequency = orderService.getOrderFrequency(103);
 		assertNotNull(orderFrequency);
 		assertTrue(orderFrequency.getRetired());
-		assertEquals("Some Retire Reason", orderFrequency.getRetireReason());
```
dkayiwa: Did you intentionally replace "Some Retire Reason" with null?

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -1407,22 +1407,23 @@ public void retireOrderFrequency_shouldRetireGivenOrderFrequency() throws Except
 	@Test
 	@Verifies(value = "should unretire given order frequency", method = "unretireOrderFrequency(OrderFrequency)")
 	public void unretireOrderFrequency_shouldUnretireGivenOrderFrequency() throws Exception {
-		OrderFrequency orderFrequency = orderService.getOrderFrequency(3);
```
dkayiwa: Could you explain why order frequency id 3 required replacing with 103?

api/src/test/java/org/openmrs/api/OrderServiceTest.java
dkayiwa: Did you forget to fix the tabbing?

api/src/test/java/org/openmrs/api/OrderServiceTest.java
jtatia: as i have already explained in previous comments the original data was failing because the class_id of the backing concept was not set to class frequency. The new data covers for that.

dkayiwa: So what made it pass before your changes?

jtatia: before this ticket it was not validated.

api/src/test/java/org/openmrs/api/OrderServiceTest.java
jtatia: @dkayiwa yes. the dataset has been replaced and this change is for the new dataset since its retire reason is null.

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -1343,22 +1343,23 @@ public void retireOrderFrequency_shouldRetireGivenOrderFrequency() {
 	
 	@Test
 	public void unretireOrderFrequency_shouldUnretireGivenOrderFrequency() {
-		OrderFrequency orderFrequency = orderService.getOrderFrequency(3);
+		executeDataSet("org/openmrs/api/include/OrderServiceTest-otherOrderFrequencies.xml");
+		OrderFrequency orderFrequency = orderService.getOrderFrequency(103);
 		assertNotNull(orderFrequency);
 		assertTrue(orderFrequency.getRetired());
-		assertEquals("Some Retire Reason", orderFrequency.getRetireReason());
+		assertEquals(null,orderFrequency.getRetireReason());
```
dkayiwa: @jtatia could you explain a bit why you introduced the OrderServiceTest-otherOrderFrequencies.xml dataset in this test? Wouldn't you make the original test pass without this?

PR #2020, opened 2017-02-27
TRUNK-5077: AbstractHandler doesn't fail to purge obs with missing file.

mks-d:
modified:   api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
This unit test was throwing an APIException prior to the change.
  modified:   api/src/test/java/org/openmrs/api/ObsServiceTest.java

### DIFF
--- a/api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
+++ b/api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
@@ -142,7 +142,10 @@ public class AbstractHandler {
 	 */
 	public boolean purgeComplexData(Obs obs) {
 		File file = getComplexDataFile(obs);
-		if (file.exists() && file.delete()) {
+		if (!file.exists()) {
+			return true;
+		}
+		else if (file.delete()) {
 			obs.setComplexData(null);
 			// obs.setValueComplex(null);
 			return true;
--- a/api/src/test/java/org/openmrs/api/ObsServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/ObsServiceTest.java
@@ -59,8 +59,8 @@ import org.openmrs.obs.handler.TextHandler;
 import org.openmrs.test.BaseContextSensitiveTest;
 import org.openmrs.test.Verifies;
 import org.openmrs.util.OpenmrsConstants;
-import org.openmrs.util.OpenmrsUtil;
 import org.openmrs.util.OpenmrsConstants.PERSON_TYPE;
+import org.openmrs.util.OpenmrsUtil;
 
 /**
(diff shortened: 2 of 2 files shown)


api/src/main/java/org/openmrs/api/ObsService.java
```diff
@@ -218,7 +218,26 @@
 	 * @param obs
 	 * @throws APIException
 	 * @see #purgeObs(Obs, boolean)
-	 * @should delete the given obs from the database
+	 * @see #purgeObs(Obs, boolean)
+	 * @see #purgeObsDao(Obs, boolean)
+	 * @should delete the given obs from the database and ignore its complex data when applicable
+	 */
+	@Authorized(PrivilegeConstants.DELETE_OBS)
+	public void purgeObsDao(Obs obs) throws APIException;
```
dkayiwa: A public service layer method mentioning dao does not sound very nice. 😊

mks-d: Ok ;-)

I'll rather provide another boolean argument to the `purgeObs(..)` then, it'll make everything simpler. I'll do that tomorrow.

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -224,6 +227,17 @@ public void purgeObs(Obs obs, boolean cascade) throws APIException {
 	}
 	
 	/**
+	 * @see org.openmrs.api.ObsService#purgeObs(org.openmrs.Obs, boolean)
+	 */
+	public void purgeObs(Obs obs, boolean cascade) throws APIException {
+		if (purgeComplexData(obs) == false) {
```
dkayiwa: How about just changing purgeComplexData() to return true if the complex obs is just not found. But always return false if something else went wrong (Failure to actually delete an existing complex obs, e.g due to permissions, etc)

mks-d: The possible implementations for complex obs handlers are too broad to just reduce the behaviour of `purgeComplexData(..)` like that, IMHO. In reality this method should return some form of success or status code that can be unequivocally linked to a clearly identified situation.

I will settle on @djazayeri suggestion to add a `forcePurge` parameter that permits ignoring non-DAO deletions. It will of course default to false.

mks-d:
@dkayiwa , are you ok with the latest iteration here (the change in `AbstractHandler`)?

dkayiwa:
@mks-d you still need to follow our process as tipped here https://wiki.openmrs.org/display/docs/Pull+Request+Tips 😊

PR #2026, opened 2017-03-01
TRUNK-5077: AbstractHandler doesn't fail to purge obs with missing file.

mks-d:
## Description
We ensure that `AbstractHandler`'s purge method returns true when the file is missing on disk. We assume that in that case the complex data purge is already successful. 
Until know, when purging a complex obs through `ObsService`, and whose complex data file is missing on disk, the above scenario would produce an `APIException`.

We also added a unit test that validates that purging a complex obs handled by `AbstractHandler` and whose complex data file is missing on disk is a valid scenario.
Note that the XML testing dataset that can be loaded in a unit test with
```java
executeDataSet(COMPLEX_OBS_XML);
```
does **not** generate test files on disk. This state of things permits to not bother about deleting them beforehand for the sake of our test.

## Related Issue

https://issues.openmrs.org/browse/TRUNK-5077

## Checklist:

### DIFF
--- a/api/src/test/java/org/openmrs/api/ObsServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/ObsServiceTest.java
@@ -9,8 +9,37 @@
  */
 package org.openmrs.api;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+import java.awt.image.BufferedImage;
+import java.awt.image.WritableRaster;
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.CharArrayReader;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.Reader;
+import java.io.Writer;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import javax.imageio.ImageIO;
+
 import org.junit.Assert;
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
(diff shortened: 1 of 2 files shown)


dkayiwa:
@mks-d did you take a look at this? https://wiki.openmrs.org/display/docs/Pull+Request+Tips
The same applies to #2020

api/src/test/java/org/openmrs/api/ObsServiceTest.java
```diff
@@ -540,6 +535,22 @@ public void getHandler_shouldHaveDefaultImageAndTextHandlersRegisteredBySpring()
 		Assert.assertNotNull(textHandler);
 	}
 	
+	@Test
+	@Verifies(value = "should purge complex obs when complex data file is missing from disk", method = "purgeObs(Obs)")
+	public void purgeComplexObs_shouldSucceedWhenCompleDataFileMissingFromDisk() throws Exception {
```
teleivo: dear @mks-d can you please update the test. it is written in the old style.
the https://wiki.openmrs.org/display/docs/Generate+Test+Case+Plugin can generate the test boilerplate according to the new style for you

mks-d: Hi @teleivo , thanks for the tip.
So I did all that and the boiletplate test had already been generated and implemented. However it wasn't handling my case about complex obs. So I added my case in it.
Repushing now...

api/src/test/java/org/openmrs/api/ObsServiceTest.java
```diff
@@ -1352,6 +1347,22 @@ public void purgeObs_shouldDeleteTheGivenObsFromTheDatabase() throws Exception {
...
 		Assert.assertNull(obsService.getObs(7));
+		
+		executeDataSet(COMPLEX_OBS_XML);
+		Obs complexObs = obsService.getComplexObs(44, ComplexObsHandler.RAW_VIEW);
+		// obs #44 is coded by the concept complex #8473 pointing to ImageHandler
+		// ImageHandler inherits AbstractHandler which handles complex data files on disk
+		assertNotNull(complexObs.getComplexData());
+		AdministrationService as = Context.getAdministrationService();
+		File complexObsDir = OpenmrsUtil.getDirectoryInApplicationDataDirectory(as
+		        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_COMPLEX_OBS_DIR));
```
dkayiwa: Did you intentionally make this test different from the one for master?

mks-d: Yes, I reacted here to @teleivo 's comments about the using the Generate Test Case Plugin for Eclipse. While I was doing so I also improved the unit test a little (= the deletion of complex data files).
Eventually I wanted to submit 2.0.x's commit on the master branch, but you have merged the former one on master already.

Also this one here would be the new style, but then I had to add my test case in an already existing test method, this was not done on master where my test case has its own method. Which one do you think is the better one?

PR #2046, opened 2017-03-07
TRUNK-4906 Expanded Cohort

vshl:
Signed-off-by: Vishal Ravi Shankar <rsh.vishal@gmail.com>

## Description

Addressed merge commit [comments](https://github.com/openmrs/openmrs-core/commit/0b662c88e8352da41bb3ac6525557b4801bf3e48).
## Related Issue

see https://issues.openmrs.org/browse/TRUNK-4906

## Checklist:

### DIFF
--- a/api/src/main/java/org/openmrs/Cohort.java
+++ b/api/src/main/java/org/openmrs/Cohort.java
@@ -31,13 +31,11 @@ public class Cohort extends BaseOpenmrsData  {
 	private String name;
 	
 	private String description;
-
-	private Set<Integer> memberIds;
-
-	private Collection<CohortMembership> members;
+	
+	private Collection<CohortMembership> memberships;
 
 	public Cohort() {
-		members = new TreeSet<CohortMembership>();
+		memberships = new TreeSet<CohortMembership>();
 	}
 	
 	/**
@@ -79,7 +77,7 @@ public class Cohort extends BaseOpenmrsData  {
 	public Cohort(String name, String description, Patient[] patients) {
 		this(name, description, (Integer[]) null);
 		if (patients != null) {
-			Arrays.stream(patients).forEach(p -> addMembership(new CohortMembership(p)));
+			Arrays.stream(patients).forEach(p -> addMembership(new CohortMembership(p.getPatientId())));
 		}
 	}
 	
@@ -108,9 +106,9 @@ public class Cohort extends BaseOpenmrsData  {
 		if (patientsOrIds != null) {
 			for (Object o : patientsOrIds) {
 				if (o instanceof Patient) {
-					addMembership(new CohortMembership((Patient) o));
+					addMembership(new CohortMembership(((Patient) o).getPatientId()));
 				} else if (o instanceof Integer) {
-					addMembership(new CohortMembership(new Patient((Integer) o)));
+					addMembership(new CohortMembership((Integer) o));
 				}
 			}
 		}
@@ -136,10 +134,13 @@ public class Cohort extends BaseOpenmrsData  {
 	}
 
 	public boolean contains(Integer patientId) {
-		return getMembers() != null && getMembers().stream()
-				.anyMatch(m -> m.getPatient().getPatientId().equals(patientId) && m.isMemberActive());
+		return getMemberships() != null && getMemberships().stream()
+				.anyMatch(m -> m.getPatientId().equals(patientId) && m.isActive());
 	}
 	
+	/**
+	 * @since 2.1.0
+	 */
 	public boolean contains(Patient patient) {
 		return contains(patient.getPatientId());
 	}
@@ -150,61 +151,119 @@ public class Cohort extends BaseOpenmrsData  {
 		if (getName() != null) {
 			sb.append(" name=").append(getName());
 		}
-		if (getMembers() != null) {
-			sb.append(" size=").append(getMembers().size());
+		if (getMemberships() != null) {
+			sb.append(" size=").append(getMemberships().size());
 		}
 		return sb.toString();
 	}
 
 	public void addMember(Integer memberId) {
-		this.addMembership(new CohortMembership(new Patient(memberId)));
+		this.addMembership(new CohortMembership(memberId));
 	}
 
 	public void removeMember(Integer memberId) {
 		removeMember(new Patient(memberId));
 	}
 	
-	public void removeMember(Patient patient) {
-		List<CohortMembership> memberToRemoveList = getMembers().stream()
-				.filter(m -> m.getPatient().getPatientId().equals(patient.getPatientId())).collect(Collectors.toList());
-		memberToRemoveList.forEach(m -> m.setEndDate(new Date()));
-	}
-
-	public void addMembership(CohortMembership cohortMembership) {
-		if (!this.contains(cohortMembership.getPatient().getPatientId())) {
-			cohortMembership.setCohort(this);
-			getMembers().add(cohortMembership);
+	/**
+	 * @since 2.1.0
+	 */
+	public boolean removeMember(Patient patient) {
+		CohortMembership memberToRemove = this.getActiveMembership(patient);
+		if (memberToRemove != null && memberToRemove.getPatientId().equals(patient.getPatientId())) {
+			memberToRemove.setEndDate(new Date());
+			return true;
 		}
+		return false;
 	}
 	
-	public void addMembership(Integer patientId) {
-		addMembership(new CohortMembership(new Patient(patientId)));
+	/**
+	 * @since 2.1.0
+	 */
+	public boolean addMembership(CohortMembership cohortMembership) {
+		if (cohortMembership != null && !this.contains(cohortMembership.getPatientId())) {
+			cohortMembership.setCohort(this);
+			return getMemberships().add(cohortMembership);
+		}
+		return false;
+	}
+	
+	public boolean addMembership(Integer patientId) {
+		return addMembership(new CohortMembership(patientId));
 	}
 
+	/**
+	 * @since 2.1.0
+	 */
 	public void removeMembership(CohortMembership cohortMembership) {
-		List<CohortMembership> memberToRemoveList = getMembers().stream()
-				.filter(m -> m.equals(cohortMembership) ||
-						m.getPatient().getPatientId().equals(cohortMembership.getPatient().getPatientId()))
+		cohortMembership.setEndDate(new Date());
+	}
+
+	/**
+	 * @since 2.1.0
+	 * @param includeVoided boolean true/false to include/exclude voided memberships
+	 * @return ArrayList of cohort memberships
+	 */
+	public List<CohortMembership> getMemberships(boolean includeVoided) {
+		return getMemberships().stream()
+				.filter(m -> m.getVoided() == includeVoided)
 				.collect(Collectors.toList());
-		memberToRemoveList.forEach(m -> m.setEndDate(new Date()));
 	}
 
-	public List<CohortMembership> getMemberships(Date asOf) {
-		return getMembers().stream()
-				.filter(m -> !m.getStartDate().before(asOf)).collect(Collectors.toList());
+	/**
+	 * @since 2.1.0
+	 */
+	public Collection<CohortMembership> getMemberships() {
+		if (memberships == null) {
+			memberships = new TreeSet<CohortMembership>();
+		}
+		return memberships;
 	}
 
-	public void purgeMemberships(List<CohortMembership> cohortMembershipList) {
-		List<CohortMembership> membersToPurge = getMembers().stream()
+	/**
+	 * @since 2.1.0
+	 * @param asOfDate date used to return active memberships
+	 * @return ArrayList of cohort memberships
+	 */
+	public List<CohortMembership> getActiveMemberships(Date asOfDate) {
+		return getMemberships().stream()
+				.filter(m -> m.isActive(asOfDate))
+				.collect(Collectors.toList());
+	}
+
+	public List<CohortMembership> getActiveMemberships() {
+		return getActiveMemberships(new Date());
+	}
+
+	/**
+	 * @since 2.1.0
+	 */
+	public CohortMembership getActiveMembership(Patient patient) {
+		return getMemberships().stream()
+				.filter(m -> m.isActive() && m.getPatientId().equals(patient.getPatientId()))
+				.collect(Collectors.toList())
+				.get(0);
+	}
+	
+	/**
+	 * @since 2.1.0
+	 */
+	public boolean purgeMemberships(List<CohortMembership> cohortMembershipList) {
+		List<CohortMembership> membershipsToPurge = getMemberships().stream()
 				.filter(m -> cohortMembershipList.stream()
-						.anyMatch(c -> m.getPatient().getPatientId().equals(c.getPatient().getPatientId())))
+						.anyMatch(c -> m.getPatientId().equals(c.getPatientId())))
 				.collect(Collectors.toList());
-		membersToPurge.forEach(m -> getMembers().remove(m));
+		if (membershipsToPurge != null) {
+			membershipsToPurge.forEach(m -> getMemberships().remove(m));
+			return true;
+		}
+		return false;
 	}
 
 
 	public int size() {
-		return getMembers().size();
+		return getMemberships().stream()
+				.filter(m -> !m.getVoided() && m.getEndDate() == null).collect(Collectors.toList()).size();
 	}
 	
 	public int getSize() {
@@ -227,10 +286,10 @@ public class Cohort extends BaseOpenmrsData  {
 	public static Cohort union(Cohort a, Cohort b) {
 		Cohort ret = new Cohort();
 		if (a != null) {
-			ret.getMembers().addAll(a.getMembers());
+			ret.getMemberships().addAll(a.getMemberships());
 		}
 		if (b != null) {
-			ret.getMembers().addAll(b.getMembers());
+			ret.getMemberships().addAll(b.getMemberships());
 		}
 		if (a != null && b != null) {
 			ret.setName("(" + a.getName() + " + " + b.getName() + ")");
@@ -249,8 +308,8 @@ public class Cohort extends BaseOpenmrsData  {
 		Cohort ret = new Cohort();
 		ret.setName("(" + (a == null ? "NULL" : a.getName()) + " * " + (b == null ? "NULL" : b.getName()) + ")");
 		if (a != null && b != null) {
-			ret.getMembers().addAll(a.getMembers());
-			ret.getMembers().retainAll(b.getMembers());
+			ret.getMemberships().addAll(a.getMemberships());
+			ret.getMemberships().retainAll(b.getMemberships());
 		}
 		return ret;
 	}
@@ -265,9 +324,9 @@ public class Cohort extends BaseOpenmrsData  {
 	public static Cohort subtract(Cohort a, Cohort b) {
 		Cohort ret = new Cohort();
 		if (a != null) {
-			ret.getMembers().addAll(a.getMembers());
+			ret.getMemberships().addAll(a.getMemberships());
 			if (b != null) {
-				ret.getMembers().removeAll(b.getMembers());
+				ret.getMemberships().removeAll(b.getMemberships());
 				ret.setName("(" + a.getName() + " - " + b.getName() + ")");
 			}
 		}
@@ -302,29 +361,22 @@ public class Cohort extends BaseOpenmrsData  {
 	}
 
 	public Set<Integer> getMemberIds() {
-		memberIds = new TreeSet<Integer>();
-		for (CohortMembership member : getMembers()) {
-			memberIds.add(member.getPatient().getPatientId());
+		Set<Integer> memberIds = new TreeSet<Integer>();
+		for (CohortMembership member : getMemberships()) {
+			memberIds.add(member.getPatientId());
 		}
 		return memberIds;
 	}
 
 	public void setMemberIds(Set<Integer> memberIds) {
-		this.memberIds = new TreeSet<Integer>(memberIds);
-		for (Integer id : memberIds) {
-			addMembership(new CohortMembership(new Patient(id)));
+		Set<Integer> ids = new TreeSet<Integer>(memberIds);
+		for (Integer id : ids) {
+			addMembership(new CohortMembership(id));
 		}
 	}
 
-	public Collection<CohortMembership> getMembers() {
-		if (members == null) {
-			members = new TreeSet<CohortMembership>();
-		}
-		return members;
-	}
-
-	public void setMembers(Collection<CohortMembership> members) {
-		this.members = members;
+	public void setMemberships(Collection<CohortMembership> members) {
+		this.memberships = members;
 	}
 
 	/**
--- a/api/src/main/java/org/openmrs/CohortMembership.java
+++ b/api/src/main/java/org/openmrs/CohortMembership.java
@@ -10,10 +10,15 @@
 package org.openmrs;
 
 import java.util.Date;
+import java.util.Objects;
 
+import org.openmrs.util.OpenmrsUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * @since 2.1.0
+ */
 public class CohortMembership extends BaseOpenmrsData implements Comparable<CohortMembership> {
 	
 	public static final long serialVersionUID = 0L;
@@ -24,7 +29,7 @@ public class CohortMembership extends BaseOpenmrsData implements Comparable<Coho
 	
 	private Cohort cohort;
 	
-	private Patient patient;
+	private Integer patientId;
 	
 	private Date startDate;
 	
@@ -34,17 +39,33 @@ public class CohortMembership extends BaseOpenmrsData implements Comparable<Coho
 	public CohortMembership() {
 	}
 
-	public CohortMembership(Patient patient, Date startDate) {
-		this.patient = patient;
+	public CohortMembership(Integer patientId, Date startDate) {
+		this.patientId = patientId;
 		this.startDate = startDate;
 	}
 
-	public CohortMembership(Patient patient) {
-		this(patient, new Date());
+	public CohortMembership(Integer patientId) {
+		this(patientId, new Date());
 	}
 	
-	public boolean isMemberActive() {
-		return this.getStartDate() != null && !getStartDate().after(new Date()) && this.getEndDate() == null;
+	/**
+	 * @param asOfDate date to compare if membership is active or inactive
+	 * @return boolean true/false if membership is active/inactive
+	 */
+	public boolean isActive(Date asOfDate) {
+		Date date;
+		if (asOfDate == null) {
+			date = new Date();
+		} else {
+			date = asOfDate;
+		}
+		return !this.getVoided()
+				&& (date.equals(this.getStartDate()) || date.after(this.getStartDate()))
+				&& (this.getEndDate() == null || date.before(this.getEndDate()));
+	}
+
+	public boolean isActive() {
+		return isActive(null);
 	}
 
 	@Override
@@ -73,12 +94,12 @@ public class CohortMembership extends BaseOpenmrsData implements Comparable<Coho
 		this.cohort = cohort;
 	}
 	
-	public Patient getPatient() {
-		return patient;
+	public Integer getPatientId() {
+		return patientId;
 	}
 	
-	public void setPatient(Patient patient) {
-		this.patient = patient;
+	public void setPatientId(Integer patientId) {
+		this.patientId = patientId;
 	}
 	
 	public Date getStartDate() {
@@ -99,6 +120,20 @@ public class CohortMembership extends BaseOpenmrsData implements Comparable<Coho
 
 	@Override
 	public int compareTo(CohortMembership o) {
-		return this.getPatient().getPatientId() - o.getPatient().getPatientId();
+		int ret = -1;
+		if (Objects.equals(this.getPatientId(), o.getPatientId())
+				&& Objects.equals(this.getCohort().getCohortId(), o.getCohort().getCohortId())
+				&& this.getStartDate().equals(o.getStartDate())
+				&& OpenmrsUtil.compare(this.getStartDate(), o.getStartDate()) == 0
+				&& ((this.getEndDate() != null && o.getEndDate() != null
+				&& OpenmrsUtil.compare(this.getEndDate(), o.getEndDate()) == 0)
+				|| (this.getEndDate() == null && o.getEndDate() == null))) {
+			ret = 0;
+		} else if (this.isActive() && !o.isActive()) {
+			ret = -1;
+		} else if (!this.isActive() && o.isActive()) {
+			ret = 1;
+		}
+		return ret;
 	}
 }
--- a/api/src/main/java/org/openmrs/api/CohortService.java
+++ b/api/src/main/java/org/openmrs/api/CohortService.java
@@ -140,14 +140,15 @@ public interface CohortService extends OpenmrsService {
 	 *
 	 * @since 2.1
 	 * @param patient patient used to find the cohorts
-	 * @return Cohorts that contain the given patient
+	 * @param includeVoided voided true/false whether or not to include voided cohorts
+	 * @param asOfDate  @return Cohorts that contain the given patient
 	 * @throws APIException
 	 * @should not return voided cohorts
 	 * @should return cohorts that have given patient
 	 */
 	@Authorized( { PrivilegeConstants.GET_PATIENT_COHORTS })
-	public List<Cohort> getCohortsContainingPatient(Patient patient, Boolean voided) throws APIException;
-	
+	public List<Cohort> getCohortsContainingPatient(Patient patient, boolean includeVoided, Date asOfDate);
+
 	/**
 	 * Find all Cohorts that contain the given patient. (Not including voided Cohorts)
 	 * 
@@ -232,7 +233,7 @@ public interface CohortService extends OpenmrsService {
 	 * @should removes Membership from Cohort by setting end date and save the Cohort
 	 */
 	@Authorized( {PrivilegeConstants.EDIT_COHORTS })
-	Cohort removeMemberShipFromCohort(Cohort cohort, CohortMembership cohortMembership);
+	Cohort removeMembershipFromCohort(Cohort cohort, CohortMembership cohortMembership);
 
 	/**
 	 * Void membership of Cohort that contain the voided Patients
@@ -252,5 +253,5 @@ public interface CohortService extends OpenmrsService {
 	 * @should unvoid the membership for the patient that was passed in
 	 */
 	@Authorized( {PrivilegeConstants.EDIT_COHORTS })
-	void patientUnvoided(Patient patient, User voidedBy, Date dateVoided, String voidReason);
+	void patientUnvoided(Patient patient, User voidedBy, Date dateVoided);
 }
--- a/api/src/main/java/org/openmrs/api/db/CohortDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/CohortDAO.java
@@ -9,6 +9,7 @@
  */
 package org.openmrs.api.db;
 
+import java.util.Date;
 import java.util.List;
 
 import org.openmrs.Cohort;
@@ -52,12 +53,12 @@ public interface CohortDAO {
 	 * Finds all cohorts that contain the given patientId
 	 *
 	 * @param patientId patient id to get the cohorts containing the patient
-	 * @param voided voided true/false whether or not to include voided patients
-	 * @return List&lt;Cohort&gt; object of matching Cohorts
+	 * @param includeVoided voided true/false whether or not to include voided cohorts
+	 * @param asOfDate  @return List&lt;Cohort&gt; object of matching Cohorts
 	 * @throws DAOException
 	 */
-	public List<Cohort> getCohortsContainingPatientId(Integer patientId, Boolean voided) throws DAOException;
-
+	public List<Cohort> getCohortsContainingPatientId(Integer patientId, boolean includeVoided, Date asOfDate);
+	
 	/**
 	 * Finds all cohorts that contain the given patientId
 	 *
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateCohortDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateCohortDAO.java
@@ -20,7 +20,6 @@ import org.hibernate.criterion.MatchMode;
 import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Restrictions;
 import org.openmrs.Cohort;
-import org.openmrs.Patient;
 import org.openmrs.api.db.CohortDAO;
 import org.openmrs.api.db.DAOException;
 import org.slf4j.Logger;
@@ -61,19 +60,23 @@ public class HibernateCohortDAO implements CohortDAO {
 	 */
 	@Override
 	@SuppressWarnings("unchecked")
-	public List<Cohort> getCohortsContainingPatientId(Integer patientId, Boolean voided) throws DAOException {
-		Disjunction orDisjunction = Restrictions.disjunction();
-		orDisjunction.add(Restrictions.isNull("m.endDate"));
-		orDisjunction.add(Restrictions.gt("m.endDate", new Date()));
-		Patient p = new Patient(patientId);
-		return (List<Cohort>) sessionFactory.getCurrentSession().createCriteria(Cohort.class)
-				.add(Restrictions.eq("voided", voided))
-				.createAlias("members", "m")
-				.add(Restrictions.le("m.startDate", new Date()))
-				.add(orDisjunction)
-				.add(Restrictions.eq("m.patient", p))
-				.setResultTransformer(CriteriaSpecification.DISTINCT_ROOT_ENTITY)
-				.list();
+	public List<Cohort> getCohortsContainingPatientId(Integer patientId, boolean includeVoided,
+	                                                  Date asOfDate) throws DAOException {
+		Disjunction orEndDate = Restrictions.disjunction();
+		orEndDate.add(Restrictions.isNull("m.endDate"));
+		orEndDate.add(Restrictions.gt("m.endDate", asOfDate));
+
+		Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Cohort.class);
+		criteria.createAlias("memberships", "m");
+		criteria.add(Restrictions.le("m.startDate", asOfDate));
+		criteria.add(orEndDate);
+		criteria.add(Restrictions.eq("m.patientId", patientId));
+		criteria.setResultTransformer(CriteriaSpecification.DISTINCT_ROOT_ENTITY);
+		
+		if (!includeVoided) {
+			criteria.add(Restrictions.eq("voided", includeVoided));
+		}
+		return criteria.list();
 	}
 
 	/**
@@ -81,7 +84,7 @@ public class HibernateCohortDAO implements CohortDAO {
 	 */
 	@Override
 	public List<Cohort> getCohortsContainingPatientId(Integer patientId) throws DAOException {
-		return getCohortsContainingPatientId(patientId, false);
+		return getCohortsContainingPatientId(patientId, false, new Date());
 	}
 
 	/**
@@ -90,7 +93,7 @@ public class HibernateCohortDAO implements CohortDAO {
 	@Override
 	public Cohort getCohortByUuid(String uuid) {
 		return (Cohort) sessionFactory.getCurrentSession().createQuery("from Cohort c where c.uuid = :uuid").setString(
-		    "uuid", uuid).uniqueResult();
+			"uuid", uuid).uniqueResult();
 	}
 
 	/**
--- a/api/src/main/java/org/openmrs/api/handler/PatientDataUnvoidHandler.java
+++ b/api/src/main/java/org/openmrs/api/handler/PatientDataUnvoidHandler.java
@@ -38,7 +38,7 @@ import org.openmrs.parameter.EncounterSearchCriteriaBuilder;
  * @see Patient
  * @since 1.9
  */
-@Handler(supports = Patient.class)
+@Handler(supports = {Patient.class}, order = 50)
 public class PatientDataUnvoidHandler implements UnvoidHandler<Patient> {
 	
 	@Override
@@ -74,7 +74,7 @@ public class PatientDataUnvoidHandler implements UnvoidHandler<Patient> {
 			}
 
 			CohortService cs = Context.getCohortService();
-			cs.patientUnvoided(patient, patient.getVoidedBy(), patient.getDateVoided(), patient.getVoidReason());
+			cs.patientUnvoided(patient, originalVoidingUser, origParentVoidedDate);
 		}
 	}
 }
--- a/api/src/main/java/org/openmrs/api/impl/CohortServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/CohortServiceImpl.java
@@ -102,9 +102,8 @@ public class CohortServiceImpl extends BaseOpenmrsService implements CohortServi
 	@Override
 	public Cohort addPatientToCohort(Cohort cohort, Patient patient) {
 		if (!cohort.contains(patient.getPatientId())) {
-			CohortMembership cohortMembership = new CohortMembership(patient);
+			CohortMembership cohortMembership = new CohortMembership(patient.getPatientId());
 			Context.getCohortService().addMembershipToCohort(cohort, cohortMembership);
-			Context.getCohortService().saveCohort(cohort);
 		}
 		return cohort;
 	}
@@ -116,8 +115,8 @@ public class CohortServiceImpl extends BaseOpenmrsService implements CohortServi
 	@Override
 	public Cohort removePatientFromCohort(Cohort cohort, Patient patient) {
 		if (cohort.contains(patient.getPatientId())) {
-			cohort.removeMember(patient);
-			Context.getCohortService().saveCohort(cohort);
+			CohortMembership membership = cohort.getActiveMembership(patient);
+			Context.getCohortService().removeMembershipFromCohort(cohort, membership);
 		}
 		return cohort;
 	}
@@ -127,8 +126,8 @@ public class CohortServiceImpl extends BaseOpenmrsService implements CohortServi
 	 */
 	@Override
 	@Transactional(readOnly = true)
-	public List<Cohort> getCohortsContainingPatient(Patient patient, Boolean voided) {
-		return dao.getCohortsContainingPatientId(patient.getPatientId(), voided);
+	public List<Cohort> getCohortsContainingPatient(Patient patient, boolean includeVoided, Date asOfDate) {
+		return dao.getCohortsContainingPatientId(patient.getPatientId(), includeVoided, asOfDate);
 	}
 	
 	@Override
@@ -194,17 +193,16 @@ public class CohortServiceImpl extends BaseOpenmrsService implements CohortServi
 	@Override
 	public Cohort addMembershipToCohort(Cohort cohort, CohortMembership cohortMembership) throws APIException {
 		cohort.addMembership(cohortMembership);
-		Context.getCohortService().saveCohort(cohort);
-		return cohort;
+		return Context.getCohortService().saveCohort(cohort);
 	}
 
 	/**
-	 * @see org.openmrs.api.CohortService#removeMemberShipFromCohort(org.openmrs.Cohort,
+	 * @see org.openmrs.api.CohortService#removeMembershipFromCohort(org.openmrs.Cohort,
 	 * 		org.openmrs.CohortMembership)
 	 */
 	@Override
-	public Cohort removeMemberShipFromCohort(Cohort cohort, CohortMembership cohortMembership) throws APIException {
-		if (cohort.contains(cohortMembership.getPatient())) {
+	public Cohort removeMembershipFromCohort(Cohort cohort, CohortMembership cohortMembership) throws APIException {
+		if (cohort.contains(cohortMembership.getPatientId())) {
 			cohort.removeMembership(cohortMembership);
 			Context.getCohortService().saveCohort(cohort);
 		}
@@ -218,33 +216,31 @@ public class CohortServiceImpl extends BaseOpenmrsService implements CohortServi
 	public void patientVoided(Patient patient) throws APIException {
 		List<Cohort> cohorts = Context.getCohortService().getCohortsContainingPatient(patient);
 		for (Cohort cohort : cohorts) {
-			List<CohortMembership> membersToVoid = cohort.getMembers().stream()
-					.filter(m -> !m.getVoided() && patient.getVoided() && m.getPatient().getPatientId().equals(patient.getPatientId()))
-					.collect(Collectors.toList());
-			for (CohortMembership member : membersToVoid) {
-				member.setVoided(patient.getVoided());
-				member.setDateVoided(patient.getDateVoided());
-				member.setVoidedBy(patient.getVoidedBy());
-				member.setVoidReason(patient.getVoidReason());
+			if (patient.getVoided()) {
+				List<CohortMembership> memberships = cohort.getMemberships(false);
+				memberships.forEach(m -> {
+					m.setVoided(patient.getVoided());
+					m.setDateVoided(patient.getDateVoided());
+					m.setVoidedBy(patient.getVoidedBy());
+					m.setVoidReason(patient.getVoidReason());
+				});
 			}
-			cohort.removeMember(patient);
-			Context.getCohortService().saveCohort(cohort);
 		}
 	}
 	
 	/**
-	 * @see org.openmrs.api.CohortService#patientUnvoided(Patient, User, Date, String)
+	 * @see org.openmrs.api.CohortService#patientUnvoided(Patient, User, Date)
 	 */
 	@Override
-	public void patientUnvoided(Patient patient, User voidedBy, Date dateVoided, String voidReason) throws APIException {
-		List<Cohort> cohorts = Context.getCohortService().getCohortsContainingPatient(patient);
+	public void patientUnvoided(Patient patient, User voidedBy, Date dateVoided) throws APIException {
+		List<Cohort> cohorts = Context.getCohortService()
+				.getCohortsContainingPatient(patient, false, new Date());
 		for (Cohort cohort : cohorts) {
-			List<CohortMembership> membersToUnvoid = cohort.getMembers().stream()
-					.filter(m -> m.getVoided() &&
-							m.getPatient().getPatientId().equals(patient.getPatientId()) &&
-							m.getVoidedBy().equals(voidedBy) &&
-							DateUtils.isSameDay(m.getDateVoided(), dateVoided) &&
-							m.getVoidReason().equals(voidReason))
+			List<CohortMembership> membersToUnvoid = cohort.getMemberships().stream()
+					.filter(m -> m.getVoided()
+							&& m.getPatientId().equals(patient.getPatientId())
+							&& m.getVoidedBy().equals(voidedBy)
+							&& DateUtils.isSameDay(m.getDateVoided(), dateVoided))
 					.collect(Collectors.toList());
 			if (!membersToUnvoid.isEmpty()) {
 				for (CohortMembership member : membersToUnvoid) {
--- a/api/src/main/java/org/openmrs/validator/CohortValidator.java
+++ b/api/src/main/java/org/openmrs/validator/CohortValidator.java
@@ -16,6 +16,8 @@ import org.openmrs.Cohort;
 import org.openmrs.CohortMembership;
 import org.openmrs.Patient;
 import org.openmrs.annotation.Handler;
+import org.openmrs.api.context.Context;
+import org.openmrs.util.OpenmrsUtil;
 import org.springframework.validation.Errors;
 import org.springframework.validation.Validator;
 
@@ -41,13 +43,21 @@ public class CohortValidator implements Validator {
 
 
 		Cohort cohort = (Cohort) obj;
-		Collection<CohortMembership> members = cohort.getMembers();
-		if (!CollectionUtils.isEmpty(members)) {
-			for (CohortMembership member : members) {
-				Patient p = member.getPatient();
-				if (p.getVoided() && !member.getVoided()) {
-					String eMessage = "Patient " + p.getPatientId() + " is voided, cannot add voided members to a cohort";
-					errors.rejectValue("members", "Cohort.patientAndMemberShouldBeVoided", eMessage);
+		if (!cohort.getVoided()) {
+			Collection<CohortMembership> members = cohort.getMemberships();
+			if (!CollectionUtils.isEmpty(members)) {
+				for (CohortMembership member : members) {
+					Patient p = Context.getPatientService().getPatient(member.getPatientId());
+					int dateCompare = OpenmrsUtil.compareWithNullAsLatest(member.getStartDate(), member.getEndDate());
+					if (p != null && p.getVoided() && !member.getVoided()) {
+						String message = "Patient " + p.getPatientId()
+								+ " is voided, cannot add voided members to a cohort";
+						errors.rejectValue("memberships", "Cohort.patientAndMemberShouldBeVoided", message);
+					}
+					if (dateCompare == 1) {
+						String message = "Start date is null or end date is before start date";
+						errors.rejectValue("memberships", "Cohort.startDateShouldNotBeNullOrBeforeEndDate", message);
+					}
 				}
 			}
 		}
--- a/api/src/main/resources/liquibase-update-to-2.1.xml
+++ b/api/src/main/resources/liquibase-update-to-2.1.xml
@@ -81,7 +81,7 @@
 			</and>
 		</preConditions>
 		<comment>Updating cohort member ids</comment>
-		<customChange class="org.openmrs.util.databasechange.UpdateCohortMemberIds"/>
+		<customChange class="org.openmrs.util.databasechange.UpdateCohortMemberIdsChangeset"/>
 	</changeSet>
 	<changeSet author="vshankar" id="201609171146-2.2" dbms="mysql">
 		<preConditions onFail="MARK_RAN">
@@ -279,10 +279,20 @@
 		<addUniqueConstraint tableName="cohort_member" columnNames="uuid"/>
 	</changeSet>
 	<changeSet id="201610242135-1" author="vshankar">
+		<preConditions onError="MARK_RAN">
+			<not>
+				<foreignKeyConstraintExists foreignKeyName="parent_cohort" foreignKeyTableName="cohort_member"/>
+			</not>
+		</preConditions>
 		<comment>Adding foreign key constraint to cohort_member.cohort_id</comment>
 		<addForeignKeyConstraint baseColumnNames="cohort_id" baseTableName="cohort_member" constraintName="parent_cohort" deferrable="false" referencedColumnNames="cohort_id" referencedTableName="cohort"/>
 	</changeSet>
 	<changeSet id="201610242135-2" author="vshankar">
+		<preConditions onError="MARK_RAN">
+			<not>
+				<foreignKeyConstraintExists foreignKeyName="member_patient" foreignKeyTableName="cohort_member"/>
+			</not>
+		</preConditions>
 		<comment>Adding foreign key constraint to cohort_member.patient_id</comment>
 		<addForeignKeyConstraint baseColumnNames="patient_id" baseTableName="cohort_member" constraintName="member_patient" deferrable="false" initiallyDeferred="false" referencedColumnNames="patient_id" referencedTableName="patient"/>
 	</changeSet>
--- a/api/src/test/java/org/openmrs/CohortMembershipTest.java
+++ b/api/src/test/java/org/openmrs/CohortMembershipTest.java
@@ -1,0 +1,149 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+import org.junit.Test;
+
+public class CohortMembershipTest {
+
+	@Test
+	public void isMembershipActive_shouldReturnTrueIfAsOfDateIsAfterStartDate() throws Exception {
+		CohortMembership newMember = new CohortMembership(4);
+		SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+		Date startDate = dateFormat.parse("2017-01-01 00:00:00");
+		newMember.setStartDate(startDate);
+
+		Date dateToTest = dateFormat.parse("2017-01-01 12:00:00");
+		assertTrue(newMember.isActive(dateToTest));
+		assertTrue(newMember.isActive(new Date()));
+		assertTrue(newMember.isActive(startDate));
+	}
+
+	@Test
+	public void isActive_shouldReturnFalseIfStartDateIsAfterAsOfDate() throws Exception {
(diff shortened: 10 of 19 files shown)


dkayiwa:
@teleivo are these the new codacy rules at work? 😊

api/src/test/java/org/openmrs/api/handler/PatientDataUnvoidHandlerTest.java
```diff
@@ -144,4 +149,39 @@ public void handle_shouldNotUnvoidTheOrdersAndEncountersThatNeverGotVoidedWithTh
 		Assert.assertNotNull(testOrder.getVoidReason());
 		
 	}
+	
+	/**
+	 * @see PatientDataUnvoidHandler#handle(Patient,User,Date,String)
+	 */
+	@Test
+	@Verifies(value = "should unvoid the members associated with the patient", method = "handle(Patient,"
+			+ "User,Date,String)")
```
teleivo: @vshl you added tests the old style. please use
https://wiki.openmrs.org/display/docs/Generate+Test+Case+Plugin
to adhere to the new style

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -13,7 +13,11 @@
 import org.apache.commons.logging.LogFactory;
 
 import java.util.Date;
+import java.util.Objects;
```
teleivo: @vshl please organize your imports according to the https://wiki.openmrs.org/display/docs/Coding+Conventions

api/src/main/java/org/openmrs/api/CohortService.java
```diff
@@ -140,13 +140,16 @@
 	 *
 	 * @since 2.1
 	 * @param patient patient used to find the cohorts
-	 * @return Cohorts that contain the given patient
+	 * @param includeVoidedCohorts voided true/false whether or not to include voided cohorts
+	 * @param includeVoidedMemberships voided true/false whether or not to include voided memberships
+	 *@param asOfDate  @return Cohorts that contain the given patient
```
teleivo: @vshl fix formatting. use the IDE formatting settings as described here https://wiki.openmrs.org/display/docs/Coding+Conventions

api/src/main/java/org/openmrs/api/db/CohortDAO.java
```diff
@@ -52,12 +53,14 @@
 	 * Finds all cohorts that contain the given patientId
 	 *
 	 * @param patientId patient id to get the cohorts containing the patient
-	 * @param voided voided true/false whether or not to include voided patients
-	 * @return List&lt;Cohort&gt; object of matching Cohorts
+	 * @param includeVoidedCohorts voided true/false whether or not to include voided cohorts
+	 * @param includeVoidedMemberships voided true/false whether or not to include voided memberships
+	 *@param asOfDate  @return List&lt;Cohort&gt; object of matching Cohorts
```
teleivo: @vshl fix formatting

api/src/test/java/org/openmrs/api/CohortServiceTest.java
```diff
@@ -20,6 +20,8 @@
 import java.util.List;
 import java.util.stream.Collectors;
 
+import javax.validation.constraints.AssertTrue;
```
teleivo: @vshl please use the org.junit.Assert.assertTrue wich is what we use in tests

api/src/test/java/org/openmrs/api/CohortServiceTest.java
```diff
@@ -644,5 +654,195 @@ public void getMemberships_shouldGetMembershipsAsOfADate() throws Exception {
...
 		assertTrue(membersAsOfDate.stream().anyMatch(m -> m.getStartDate().equals(dateToTest)));
 	}
-
+	
+	/**
+	 * @verifies {@link Cohort#getMemberships(Date)}
+	 */
+	@Test
+	@Verifies(value = "should not get matching memberships of a cohort as of a date", method = "getMemberships(Date)")
+	public void getMemberships_shouldNotGetMatchingMembershipsAsOfADate() throws Exception {
```
teleivo: again old unit test style please update

api/src/test/java/org/openmrs/api/CohortServiceTest.java
```diff
@@ -644,5 +654,195 @@ public void getMemberships_shouldGetMembershipsAsOfADate() throws Exception {
...
+	 *
+	 */
+	@Test
+	@Verifies(value = "should not return voided memberships", method = "getCohortsContainingPatient(Patient)")
+	public void getCohortsContainingPatient_shouldNotReturnVoidedMemberships() throws Exception {
+		executeDataSet(COHORT_XML);
+		
+		Cohort cohort = service.getCohort(2);
+		
+		// add a patient to cohorts
```
teleivo: remove comments. they do not add information. the code speaks for itself

api/src/test/java/org/openmrs/api/CohortServiceTest.java
```diff
@@ -644,5 +654,195 @@ public void getMemberships_shouldGetMembershipsAsOfADate() throws Exception {
...
+		Patient patient = new Patient(7);
+		CohortMembership membership = new CohortMembership(patient);
+		cohort.addMembership(membership);
+		assertTrue(cohort.contains(patient.getPatientId()));
+		
+		// void the membership
+		membership.setVoided(true);
```
teleivo: good example of why this comment is not needed. the code reads almost exactly the same as your comment.  please remove your comments in all your tests! if you feel the need to explain your code think about a more clear way to write it. but in this case it is very clear :)

api/src/test/java/org/openmrs/api/handler/PatientDataUnvoidHandlerTest.java
```diff
@@ -144,4 +149,39 @@ public void handle_shouldNotUnvoidTheOrdersAndEncountersThatNeverGotVoidedWithTh
...
+		patient = Context.getPatientService().voidPatient(patient, "Void Reason");
+		Assert.assertTrue(patient.getVoided());
+		
+		Assert.assertTrue(membership.getVoided());
+		Assert.assertNotNull(membership.getDateVoided());
+		Assert.assertNotNull(membership.getVoidedBy());
+		Assert.assertNotNull(membership.getVoidReason());
+		
+		//TODO: unvoiding logic needs to be addressed
+		//Context.getPatientService().unvoidPatient(patient);
```
teleivo: @vshl who is this TODO for? Do not add TODOs, they will be left forever in the code. Nobody will feel responsible for it. if there is something left TODO your PR is not ready.

teleivo:
@dkayiwa yes, to strict? I agree with all of them.

api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -180,43 +181,68 @@ public void removeMember(Integer memberId) {
...
+	/**
+	 * @since 2.1.0
+	 */
 	public List<CohortMembership> getMemberships(Date asOf) {
-		return getMembers().stream()
-				.filter(m -> !m.getStartDate().before(asOf)).collect(Collectors.toList());
+		return getMemberships().stream()
+				.filter(m -> !m.getStartDate().before(asOf) &&
+						m.getStartDate().equals(asOf) &&
+						!m.getVoided() && m.getEndDate() == null)
```
wluyima: A few comments here:
- if asOfDate is null null, it default to current date
- the true condition for a membership to return should be 
`!voided && the asOfDate.after(startDate) && (endDate == null || asOfDate.before(endDate)`

api/src/main/java/org/openmrs/validator/CohortValidator.java
```diff
@@ -41,13 +41,15 @@ public void validate(Object obj, Errors errors) {
...
-				Patient p = member.getPatient();
-				if (p.getVoided() && !member.getVoided()) {
-					String eMessage = "Patient " + p.getPatientId() + " is voided, cannot add voided members to a cohort";
-					errors.rejectValue("members", "Cohort.patientAndMemberShouldBeVoided", eMessage);
+		if (!cohort.getVoided()) {
+			Collection<CohortMembership> members = cohort.getMemberships();
+			if (!CollectionUtils.isEmpty(members)) {
+				for (CohortMembership member : members) {
+					Patient p = member.getPatient();
+					if (p.getVoided() && !member.getVoided()) {
```
wluyima: Also reject memberships where startDate is null or if endDate is before startDate

api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -180,43 +181,68 @@ public void removeMember(Integer memberId) {
...
+				.collect(Collectors.toList());
 	}
-
+	
+	/**
+	 * @since 2.1.0
+	 */
+	public CohortMembership getActiveMembership(Patient patient) {
+		List <CohortMembership> membership = this.getMemberships().stream()
```
wluyima: Did you format your code? Because I agree with codacy here

api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -180,43 +181,68 @@ public void removeMember(Integer memberId) {
...
+				.filter(m -> m.getPatient().getPatientId().equals(patient.getPatientId()) &&
+						!m.getVoided() && m.getEndDate() == null && m.isMembershipActive())
```
wluyima: getMemberships() only returns active memberships therefore you don't need these other conditions, as long as the patients match that would be it.

wluyima: As long as you call getMemberships(null)

api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -180,43 +181,68 @@ public void removeMember(Integer memberId) {
...
+						m.getStartDate().equals(asOf) &&
+						!m.getVoided() && m.getEndDate() == null)
+				.collect(Collectors.toList());
 	}
-
+	
+	/**
+	 * @since 2.1.0
+	 */
+	public CohortMembership getActiveMembership(Patient patient) {
```
wluyima: You should be calling getActiveMemberships(null) , that way you don't have to check again the other conditions if the patient ids match

api/src/main/java/org/openmrs/Cohort.java
wluyima: In fact, you shouldn't have these condition here, just delegate to isMembershipActive(asOfDate), you always want to consolidate code you use several times in one place or method

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -43,8 +47,18 @@ public CohortMembership(Patient patient) {
 		this(patient, new Date());
 	}
 	
-	public boolean isMemberActive() {
-		return this.getStartDate() != null && this.getStartDate().before(new Date()) && this.getEndDate() == null;
+	public boolean isMembershipActive() {
+		return this.getStartDate() != null &&
```
wluyima: You should be delegating to isMembershipActive(null) to avoid code duplication

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -43,8 +47,18 @@ public CohortMembership(Patient patient) {
...
+				this.getStartDate().before(new Date()) &&
+				this.getEndDate() == null &&
+				!this.getVoided();
+	}
+	
+	public boolean isMembershipActive(Date asOfDate) {
+		return this.getStartDate() != null &&
+				!this.getStartDate().after(asOfDate) &&
```
wluyima: !voided && the asOfDate.after(startDate) && (endDate == null || asOfDate.before(endDate)

wluyima: In fact OpenmrsUtil has utility functions you can call that compare dates taking into account null values and how you wish to treat them

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -101,4 +115,16 @@ public void setEndDate(Date endDate) {
 	public int compareTo(CohortMembership o) {
 		return this.getPatient().getPatientId() - o.getPatient().getPatientId();
 	}
+	
+	@Override
```
wluyima: Codacy is correct here, you always have to implement hashCode when you implement eqauls and they should be consistent with each other, 2 objects that are equal should always have the same hashcode

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -101,4 +115,16 @@ public void setEndDate(Date endDate) {
...
+	public boolean equals(Object obj) {
+		if (obj instanceof CohortMembership) {
+			CohortMembership o = (CohortMembership) obj;
+			return Objects.equals(this.getPatient().getPatientId(), o.getPatient().getPatientId()) &&
+					Objects.equals(this.getCohort().getCohortId(), o.getCohort().getCohortId()) &&
+					this.getStartDate() == o.getStartDate() &&
+					this.getEndDate() == o.getEndDate();
+		}
+		return super.equals(obj);
```
wluyima: this should throw an exception since it isn't of the same type

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -101,4 +115,16 @@ public void setEndDate(Date endDate) {
...
 		return this.getPatient().getPatientId() - o.getPatient().getPatientId();
 	}
+	
+	@Override
+	public boolean equals(Object obj) {
+		if (obj instanceof CohortMembership) {
+			CohortMembership o = (CohortMembership) obj;
+			return Objects.equals(this.getPatient().getPatientId(), o.getPatient().getPatientId()) &&
+					Objects.equals(this.getCohort().getCohortId(), o.getCohort().getCohortId()) &&
+					this.getStartDate() == o.getStartDate() &&
```
wluyima: Use equals method to compare dates please

wluyima: You also need this to be consistent with equality semantics in OpenMRS, i.e if 2 CohortMemberships have the same uuid, they should be considered equal, basically you should delegate to super.equals first, if it returns false then you proceed to your code to further compare the memberships

api/src/main/java/org/openmrs/api/CohortService.java
```diff
@@ -140,13 +140,16 @@
...
 	 * @throws APIException
 	 * @should not return voided cohorts
 	 * @should return cohorts that have given patient
 	 */
 	@Authorized( { PrivilegeConstants.GET_PATIENT_COHORTS })
-	public List<Cohort> getCohortsContainingPatient(Patient patient, Boolean voided) throws APIException;
+	public List<Cohort> getCohortsContainingPatient(Patient patient, Boolean includeVoidedCohorts,
+													Boolean includeVoidedMemberships, Date asOfDate) throws APIException;
```
wluyima: You shouldn't have the includeVoidedMemberships on this method, instead you should add another  `getMemberships(boolean includeVoided)` method to the Cohort class

api/src/main/java/org/openmrs/api/CohortService.java
```diff
@@ -140,13 +140,16 @@
...
+	 * @param includeVoidedCohorts voided true/false whether or not to include voided cohorts
+	 * @param includeVoidedMemberships voided true/false whether or not to include voided memberships
+	 *@param asOfDate  @return Cohorts that contain the given patient
 	 * @throws APIException
 	 * @should not return voided cohorts
 	 * @should return cohorts that have given patient
 	 */
 	@Authorized( { PrivilegeConstants.GET_PATIENT_COHORTS })
-	public List<Cohort> getCohortsContainingPatient(Patient patient, Boolean voided) throws APIException;
+	public List<Cohort> getCohortsContainingPatient(Patient patient, Boolean includeVoidedCohorts,
```
wluyima: By convention the argument name should be `includeVoided`

api/src/main/java/org/openmrs/api/db/CohortDAO.java
```diff
@@ -52,12 +53,14 @@
...
 	 * @throws DAOException
 	 */
-	public List<Cohort> getCohortsContainingPatientId(Integer patientId, Boolean voided) throws DAOException;
-
+	public List<Cohort> getCohortsContainingPatientId(Integer patientId, Boolean includeVoidedCohorts,
```
wluyima: Remove and rename arguments based on the comments I added to the service method above

api/src/main/java/org/openmrs/api/db/hibernate/HibernateCohortDAO.java
```diff
@@ -59,35 +59,43 @@ public Cohort getCohort(Integer id) throws DAOException {
 	 * @see org.openmrs.api.db.CohortDAO#getCohortsContainingPatientId(java.lang.Integer)
 	 */
 	@SuppressWarnings("unchecked")
-	public List<Cohort> getCohortsContainingPatientId(Integer patientId, Boolean voided) throws DAOException {
+	public List<Cohort> getCohortsContainingPatientId(Integer patientId, Boolean includeVoidedCohorts,
+													  Boolean includeVoidedMemberships, Date asOfDate) throws DAOException {
 		Disjunction orDisjunction = Restrictions.disjunction();
 		orDisjunction.add(Restrictions.isNull("m.endDate"));
-		orDisjunction.add(Restrictions.gt("m.endDate", new Date()));
+		orDisjunction.add(Restrictions.gt("m.endDate", asOfDate));
```
wluyima: you should also exclude those that were ended on `asOfDate` too

api/src/main/java/org/openmrs/api/db/hibernate/HibernateCohortDAO.java
```diff
@@ -59,35 +59,43 @@ public Cohort getCohort(Integer id) throws DAOException {
...
-				.createAlias("members", "m")
-				.add(Restrictions.le("m.startDate", new Date()))
-				.add(orDisjunction)
-				.add(Restrictions.eq("m.patient", p))
-				.setResultTransformer(CriteriaSpecification.DISTINCT_ROOT_ENTITY)
-				.list();
+		
+		Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Cohort.class);
+		criteria.createAlias("memberships", "m");
+		criteria.add(Restrictions.not(Restrictions.gt("m.startDate", asOfDate)));
```
wluyima: Why did you change this from using `Restrictions.le("m.startDate), asOfDate)`? It's cleaner and more readable

api/src/main/java/org/openmrs/api/db/hibernate/HibernateCohortDAO.java
```diff
@@ -59,35 +59,43 @@ public Cohort getCohort(Integer id) throws DAOException {
...
+		criteria.add(orDisjunction);
+		criteria.add(Restrictions.eq("m.patient", p));
+		criteria.setResultTransformer(CriteriaSpecification.DISTINCT_ROOT_ENTITY);
+		
+		if (!includeVoidedCohorts) {
+			criteria.add(Restrictions.eq("voided", includeVoidedCohorts));
+		}
+		if (!includeVoidedMemberships) {
+			criteria.add(Restrictions.eq("m.voided", includeVoidedMemberships));
```
wluyima: You should always return voided memberships from the DB

api/src/main/java/org/openmrs/api/impl/CohortServiceImpl.java
```diff
@@ -200,27 +199,25 @@ public Cohort removeMemberShipFromCohort(Cohort cohort, CohortMembership cohortM
...
-				member.setDateVoided(patient.getDateVoided());
-				member.setVoidedBy(patient.getVoidedBy());
-				member.setVoidReason(patient.getVoidReason());
+			if (patient.getVoided()) {
+				CohortMembership membershipToVoid = cohort.getActiveMembership(patient);
+				membershipToVoid.setVoided(patient.getVoided());
+				membershipToVoid.setDateVoided(patient.getDateVoided());
+				membershipToVoid.setVoidedBy(patient.getVoidedBy());
+				membershipToVoid.setVoidReason(patient.getVoidReason());
+				Context.getCohortService().removeMemberShipFromCohort(cohort, membershipToVoid);
```
wluyima: Marking it as voided is enough, you don't have to remove it

api/src/main/java/org/openmrs/api/impl/CohortServiceImpl.java
```diff
@@ -200,27 +199,25 @@ public Cohort removeMemberShipFromCohort(Cohort cohort, CohortMembership cohortM
...
-			List<CohortMembership> membersToVoid = cohort.getMembers().stream()
-					.filter(m -> !m.getVoided() && patient.getVoided() && m.getPatient().getPatientId().equals(patient.getPatientId()))
-					.collect(Collectors.toList());
-			for (CohortMembership member : membersToVoid) {
-				member.setVoided(patient.getVoided());
-				member.setDateVoided(patient.getDateVoided());
-				member.setVoidedBy(patient.getVoidedBy());
-				member.setVoidReason(patient.getVoidReason());
+			if (patient.getVoided()) {
+				CohortMembership membershipToVoid = cohort.getActiveMembership(patient);
```
wluyima: You should call getMemberships() and void those that are not yet voided instead of only active ones, didn't you add a `getMemberships(includeVoided)` method that you can call?

api/src/main/java/org/openmrs/api/handler/PatientDataUnvoidHandler.java
```diff
@@ -68,14 +68,19 @@ public void handle(Patient patient, User originalVoidingUser, Date origParentVoi
...
 					}
 				}
 			}
-
+			
 			CohortService cs = Context.getCohortService();
-			cs.patientUnvoided(patient, patient.getVoidedBy(), patient.getDateVoided(), patient.getVoidReason());
+			if (patient.getId() != null) {
+				if (patient.getDateVoided().equals(origParentVoidedDate) &&
+						patient.getVoidedBy().equals(originalVoidingUser)) {
```
wluyima: This logic should be in the patientVoided method and not here

wluyima: Meant to say patientUnvoided method

api/src/main/resources/liquibase-update-to-2.1.xml
```diff
@@ -279,10 +279,28 @@
 		<addUniqueConstraint tableName="cohort_member" columnNames="uuid"/>
 	</changeSet>
 	<changeSet id="201610242135-1" author="vshankar">
+		<preConditions>
+			<not>
+				<customPrecondition className="com.ctp.liquibase.ForeignKeyExistsPrecondition">
```
wluyima: liquibase has a `<foreignKeyConstraintExists />` that you should use, it's good to use the standard features from a library

wluyima: Is is a standard liquibase class?

wluyima: liquibase has a ForeignKeyExistsPrecondition class too, why can't you use it?

api/src/test/java/org/openmrs/api/CohortServiceTest.java
```diff
@@ -644,5 +654,195 @@ public void getMemberships_shouldGetMembershipsAsOfADate() throws Exception {
...
+		CohortMembership newMember = new CohortMembership((new Patient(4)));
+		SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+		Date startDate = dateFormat.parse("2017-11-01 00:00:00");
+		newMember.setStartDate(startDate);
+		service.addMembershipToCohort(cohort, newMember);
+		
+		Date dateToTest = dateFormat.parse("2016-11-01 00:00:00");
+		List<CohortMembership> membersAsOfDate = cohort.getMemberships(dateToTest);
+		assertTrue(membersAsOfDate.isEmpty());
+		assertFalse(membersAsOfDate.stream().anyMatch(m -> m.getStartDate().equals(dateToTest)));
```
wluyima: This is a redundant call because isEmpty() already returned true, so this can never be true

api/src/test/java/org/openmrs/api/CohortServiceTest.java
```diff
@@ -644,5 +654,195 @@ public void getMemberships_shouldGetMembershipsAsOfADate() throws Exception {
...
+	}
+	
+	/**
+	 * @verifies {@link CohortMembership#isMembershipActive(Date)}
+	 */
+	@Test
+	@Verifies(value = "membership should be active as of a date", method = "isMembershipActive(Date)")
+	public void isMembershipActive_shouldBeActiveAsOfDate() throws Exception {
```
wluyima: Method name is not intuitive, should say something like 'shouldReturnTrueIfStartDateIsBeforeAsOfDate()'

api/src/test/java/org/openmrs/api/CohortServiceTest.java
```diff
@@ -644,5 +654,195 @@ public void getMemberships_shouldGetMembershipsAsOfADate() throws Exception {
...
+		assertTrue(newMember.isMembershipActive(new Date()));
+		assertTrue(newMember.isMembershipActive(startDate));
+	}
+	
+	/**
+	 * @verifies {@link CohortMembership#isMembershipActive(Date)}
+	 */
+	@Test
+	@Verifies(value = "membership should be inactive as of a date", method = "isMembershipActive(Date)")
+	public void isMembershipActive_shouldBeInactiveAsOfDate() throws Exception {
```
wluyima: Method name is not intuitive, should say something like shouldReturnFalseIfStartDateIsAfterAsOfDate()

api/src/test/java/org/openmrs/api/CohortServiceTest.java
```diff
@@ -644,5 +654,195 @@ public void getMemberships_shouldGetMembershipsAsOfADate() throws Exception {
...
+		SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+		Date startDate = dateFormat.parse("2017-01-01 00:00:00");
+		Date endDate = dateFormat.parse("2017-01-31 00:00:00");
+		newMember.setStartDate(startDate);
+		newMember.setEndDate(endDate);
+		newMember.setVoided(true);
+		
+		Date dateToTest = dateFormat.parse("2017-02-01 00:00:00");
+		assertFalse(newMember.isMembershipActive(dateToTest));
+		Date secondDateToTest = dateFormat.parse("2017-01-15 00:00:00");
```
wluyima: How different is testing this from 2017-02-01? I think it's a redundant check. You don't even need to pass in an as of date, the method you're testing should default to now

api/src/test/java/org/openmrs/api/CohortServiceTest.java
```diff
@@ -644,5 +654,195 @@ public void getMemberships_shouldGetMembershipsAsOfADate() throws Exception {
...
+		assertFalse(newMember.isMembershipActive(secondDateToTest));
+	}
+	
+	/**
+	 * @verifies {@link CohortMembership#compareTo(CohortMembership)}}
+	 */
+	@Test
+	@Verifies(value = "should compare two membership objects", method = "compareTo(CohortMembership)")
+	public void compareTo_shouldCompareMemberships() throws Exception {
```
wluyima: Are you testing compareTo() or equals()? The equals method checks the startDate too for equality but I see you are only testing with a patient, the test should also cover startDate. I was also expecting to see tests that check the method returns false if patient or startDate differ

api/src/test/java/org/openmrs/api/CohortServiceTest.java
```diff
@@ -644,5 +654,195 @@ public void getMemberships_shouldGetMembershipsAsOfADate() throws Exception {
...
+		Patient patient = new Patient(7);
+		CohortMembership membership = new CohortMembership(patient);
+		cohort.addMembership(membership);
+		assertTrue(cohort.contains(patient.getPatientId()));
+		
+		// void the membership
+		membership.setVoided(true);
+		
+		// call the method and it should not return the voided membership
+		List<Cohort> cohortsWithPatientAdded = service.getCohortsContainingPatient(patient);
```
wluyima: This test doesn't seem concrete enough, you need to save the cohort before you call this method otherwise.

api/src/test/java/org/openmrs/api/CohortServiceTest.java
wluyima: I agree, typically you add comments to areas where you're implementing some logic that isn't straight forward to understand, doesn't make sense to comment on method call like this unless you're saying why you are calling it.

teleivo:
@vshl did you get a chance to incorporate the comments? if so please update this PR by pushing your changes

vshl:
@teleivo Yes, I am still working on it.

api/src/test/java/org/openmrs/api/handler/PatientDataUnvoidHandlerTest.java
vshl: Interesting. Thanks. So should I move these tests into it's own class CohortMembershipTest.java?

api/src/test/java/org/openmrs/api/handler/PatientDataUnvoidHandlerTest.java
vshl: I wanted to clear a doubt with @wluyima. Basically to address [this comment](https://github.com/openmrs/openmrs-core/commit/0b662c88e8352da41bb3ac6525557b4801bf3e48#commitcomment-21178110) that I added in the earlier commit. I will remove it.

api/src/main/java/org/openmrs/Cohort.java
vshl: I believe I created this to get an active membership for a patient as suggested by you in the earlier commit. Since I was performing this query frequently, should I keep this method around? I have simplified the method to:

```
		return getMemberships().stream()
				.filter(m -> m.isMembershipActive() && m.getPatient().getPatientId().equals(patient.getPatientId()))
				.collect(Collectors.toList())
				.get(0);
```

api/src/main/java/org/openmrs/CohortMembership.java
vshl: Is this a better implementation?

```
	public boolean equals(Object obj) {
		if (obj instanceof CohortMembership) {
			CohortMembership o = (CohortMembership) obj;
			if (this.getUuid().equals(o.getUuid())) {
				return super.equals(obj);
			} else {
				return Objects.equals(this.getPatient().getPatientId(), o.getPatient().getPatientId())
						&& Objects.equals(this.getCohort().getCohortId(), o.getCohort().getCohortId())
						&& this.getStartDate().equals(o.getStartDate())
						&& this.getEndDate().equals(o.getEndDate());
			}
		}
		return false;
```

api/src/main/java/org/openmrs/api/impl/CohortServiceImpl.java
vshl: Thanks, I implemented `getMemberships(includeVoided)` right now.

api/src/main/resources/liquibase-update-to-2.1.xml
vshl: Sorry, <foreignKeyConstraintExists /> didn't work initially, however now it does. Not sure what went wrong before.

api/src/test/java/org/openmrs/api/CohortServiceTest.java
vshl: compareTo(). I wasn't entirely clear on how to perform this test. Should I be testing for equals() instead?

api/src/test/java/org/openmrs/api/CohortServiceTest.java
vshl: Since these test `CohortMembership.java`, should I move them to their own test class `CohortMembershipTest.java`?

api/src/main/java/org/openmrs/api/handler/PatientDataUnvoidHandler.java
vshl: For the condition to be in patientUnvoided(), shouldn't the patient be "unvoided" at the time? With the fields, dateVoided, voidedBy == null? This falls back to my earlier question about the unvoiding logic: [here](https://github.com/openmrs/openmrs-core/commit/0b662c88e8352da41bb3ac6525557b4801bf3e48#commitcomment-21178110). I will probably test again and get back to this comment.

vshl: Calling `Context.getPatientService().unvoidPatient(patient);`
gives error: `org.openmrs.api.ValidationException: 'Cohort id=2 name=Example Cohort size=2' failed to validate with reason: memberships: Cohort.patientAndMemberShouldBeVoided`
because of this in CohortValidator.java:
```
					if ((p.getVoided() && !member.getVoided())
							|| (member.getStartDate() == null || dateCompare == 1)) {
```

At this point, the patient is voided and the membership isn't. So my question, at what point the patient should be unvoided? Because at these stages: PatientDataUnvoidHandler.handle(), CohortServiceImpl.patientUnvoided(), the patient's voided fields are still set. Hence the validation fails in CohortValidator. I hope you get my question.

api/src/test/java/org/openmrs/api/handler/PatientDataUnvoidHandlerTest.java
wluyima: That's not what I meant, what I meant is that if you are testing code in PatientDataUnvoidHandler then the test should be in PatientDataUnvoidHandlerTest

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -0,0 +1,160 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.openmrs.api.CohortService;
+import org.openmrs.api.context.Context;
+import org.openmrs.test.BaseContextSensitiveTest;
+
+public class CohortMembershipTest extends BaseContextSensitiveTest {
```
wluyima: Why does this extend BaseContextSensitiveTest?

wluyima: I wouldn't expect any tests in this class that require inserting test data to make this be context sensitive

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -0,0 +1,160 @@
...
+		assertEquals(bMember.compareTo(aMember), 0);
+	}
+
+	/**
+	 * @verifies compare two objects for equality
+	 * @see CohortMembership#equals(Object)
+	 */
+	@Test
+	public void equals_shouldCompareTwoObjectsForEquality() throws Exception {
+		executeDataSet(COHORT_XML);
```
wluyima: So this is the test that requires adding test data? Do you really need to add it? Cant you just work with  new instances within this method? That way you don't have to make this class context sensitive?

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -101,4 +120,35 @@ public void setEndDate(Date endDate) {
...
 	}
+
+	/**
+	 * @should compare two objects for equality
+	 */
+	@Override
+	public boolean equals(Object obj) {
+		if (obj instanceof CohortMembership) {
+			CohortMembership o = (CohortMembership) obj;
+			if (this.getUuid().equals(o.getUuid())) {
```
wluyima: You should be calling super.equals(o), if the call to the super method returns false that's when you execute your logic.

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -101,4 +120,35 @@ public void setEndDate(Date endDate) {
...
+		}
+		return false;
+	}
+	
+	@Override
+	public int hashCode() {
+		int result = 0;
+		result = 71 * result + (this.getPatient().getPatientId() != null
+										? this.getPatient().getPatientId().hashCode() : 0);
+		result = 71 * result + (this.getCohort().getCohortId() != null
```
wluyima: Implementation of hashcode() needs to align with equals, in the equals method you do delegate to super and you would need to do the same here but this might complicate the logic. On the other hand, I think you could move this logic to the compareTo() method and not have to implement equals() and hashcode() since Set uses compareTo and not equals

api/src/main/java/org/openmrs/api/db/hibernate/HibernateCohortDAO.java
```diff
@@ -61,27 +61,36 @@ public Cohort getCohort(Integer id) throws DAOException {
...
-	public List<Cohort> getCohortsContainingPatientId(Integer patientId, Boolean voided) throws DAOException {
-		Disjunction orDisjunction = Restrictions.disjunction();
-		orDisjunction.add(Restrictions.isNull("m.endDate"));
-		orDisjunction.add(Restrictions.gt("m.endDate", new Date()));
+	public List<Cohort> getCohortsContainingPatientId(Integer patientId, Boolean includeVoided,
+	                                                  Date asOfDate) throws DAOException {
+		Disjunction orStartDate = Restrictions.disjunction();
+		Disjunction orEndDate = Restrictions.disjunction();
+		orStartDate.add(Restrictions.le("m.startDate", asOfDate));
+		orStartDate.add(Restrictions.eq("m.startDate", asOfDate));
```
wluyima: You don't need this second restriction because Restrictions.le is already inclusive of the specified date

api/src/main/java/org/openmrs/api/db/hibernate/HibernateCohortDAO.java
```diff
@@ -61,27 +61,36 @@ public Cohort getCohort(Integer id) throws DAOException {
...
+		orEndDate.add(Restrictions.isNull("m.endDate"));
+		orEndDate.add(Restrictions.gt("m.endDate", asOfDate));
```
wluyima: The restrictions for end date is null and greater than should be mutually exclusive i.e should be added to a single disjunction

api/src/main/java/org/openmrs/api/db/hibernate/HibernateCohortDAO.java
wluyima: What am saying is that Restrictions.le performs a less than or equal to so you don't need the second equal to

api/src/main/java/org/openmrs/api/db/hibernate/HibernateCohortDAO.java
```diff
@@ -61,27 +61,36 @@ public Cohort getCohort(Integer id) throws DAOException {
...
-				.createAlias("members", "m")
-				.add(Restrictions.le("m.startDate", new Date()))
-				.add(orDisjunction)
-				.add(Restrictions.eq("m.patient", p))
-				.setResultTransformer(CriteriaSpecification.DISTINCT_ROOT_ENTITY)
-				.list();
+		
+		Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Cohort.class);
+		criteria.createAlias("memberships", "m");
+		criteria.add(Restrictions.le("m.startDate", asOfDate));
```
wluyima: Aren't you repeating this?

api/src/main/java/org/openmrs/validator/CohortValidator.java
```diff
@@ -41,13 +42,18 @@ public void validate(Object obj, Errors errors) {
...
-					errors.rejectValue("members", "Cohort.patientAndMemberShouldBeVoided", eMessage);
+		if (!cohort.getVoided()) {
+			Collection<CohortMembership> members = cohort.getMemberships();
+			if (!CollectionUtils.isEmpty(members)) {
+				for (CohortMembership member : members) {
+					Patient p = member.getPatient();
+					int dateCompare = OpenmrsUtil.compareWithNullAsLatest(member.getStartDate(), member.getEndDate());
+					if ((p.getVoided() && !member.getVoided())
+							|| (member.getStartDate() == null || dateCompare == 1)) {
+						String message = "Patient " + p.getPatientId()
```
wluyima: You need to split up this so that you set a specific error message based on the failed condition

api/src/main/java/org/openmrs/validator/CohortValidator.java
```diff
@@ -41,13 +42,18 @@ public void validate(Object obj, Errors errors) {
...
-					String eMessage = "Patient " + p.getPatientId() + " is voided, cannot add voided members to a cohort";
-					errors.rejectValue("members", "Cohort.patientAndMemberShouldBeVoided", eMessage);
+		if (!cohort.getVoided()) {
+			Collection<CohortMembership> members = cohort.getMemberships();
+			if (!CollectionUtils.isEmpty(members)) {
+				for (CohortMembership member : members) {
+					Patient p = member.getPatient();
+					int dateCompare = OpenmrsUtil.compareWithNullAsLatest(member.getStartDate(), member.getEndDate());
+					if ((p.getVoided() && !member.getVoided())
+							|| (member.getStartDate() == null || dateCompare == 1)) {
```
wluyima: Do you realize that if startDate is null based on the method you called above then the dateCompare value would be 1? Meaning the `member.getStartDate() == null` check is redundant

api/src/test/java/org/openmrs/CohortMembershipTest.java
vshl: I wanted to use the existing dataset to insert Cohorts. I could do away with it and initialize them manually.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateCohortDAO.java
vshl: Sorry I wanted to remove it. Just left it there by mistake.

vshl:
@wluyima I have a question. If I have to use the REST API for the `memberships`, since the current behavior is /cohort/UUID/member/patientID, I do need to create new endpoints for memberships in particular? I have this requirement in a module that I am working on. I can theoretically create the endpoints in the module?

wluyima:
Actually the current behavior is `/ws/rest/v1/cohort/{uuid}/member,` in the updated rest module it would be `/ws/rest/v1/cohort/{uuid}/membership`, it means you would add a CohortMembershipResource2_1 subresource, I recall as if we already had a design discussion of what the rest service should be. May be we can bring up it again

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -99,6 +118,15 @@ public void setEndDate(Date endDate) {
...
+		if (Objects.equals(this.getPatient().getPatientId(), o.getPatient().getPatientId())
+				&& Objects.equals(this.getCohort().getCohortId(), o.getCohort().getCohortId())
+				&& this.getStartDate().equals(o.getStartDate())
+				&& ((this.getEndDate() != null && o.getEndDate() != null
+				&& this.getEndDate().equals(o.getEndDate()))
+				|| (this.getEndDate() == null && o.getEndDate() == null))) {
+			return 0;
+		} else {
+			return -1;
+		}
```
wluyima: I would imagine that there should be a condition where this returns 1

api/src/main/java/org/openmrs/api/CohortService.java
```diff
@@ -140,13 +140,15 @@
...
-	 * @return Cohorts that contain the given patient
+	 * @param includeVoided voided true/false whether or not to include voided cohorts
+	 * @param asOfDate  @return Cohorts that contain the given patient
 	 * @throws APIException
 	 * @should not return voided cohorts
 	 * @should return cohorts that have given patient
 	 */
 	@Authorized( { PrivilegeConstants.GET_PATIENT_COHORTS })
-	public List<Cohort> getCohortsContainingPatient(Patient patient, Boolean voided) throws APIException;
+	public List<Cohort> getCohortsContainingPatient(Patient patient,Boolean includeVoided, Date asOfDate)
```
wluyima: This should be boolean and not Boolean based on on similar existing methods

api/src/main/java/org/openmrs/api/CohortService.java
```diff
@@ -140,13 +140,15 @@
...
+			throws APIException;
```
wluyima: You can omit the throws clause

api/src/main/java/org/openmrs/api/db/CohortDAO.java
```diff
@@ -52,12 +53,13 @@
...
-	 * @param voided voided true/false whether or not to include voided patients
-	 * @return List&lt;Cohort&gt; object of matching Cohorts
+	 * @param includeVoided voided true/false whether or not to include voided cohorts
+	 * @param asOfDate  @return List&lt;Cohort&gt; object of matching Cohorts
 	 * @throws DAOException
 	 */
-	public List<Cohort> getCohortsContainingPatientId(Integer patientId, Boolean voided) throws DAOException;
-
+	public List<Cohort> getCohortsContainingPatientId(Integer patientId, Boolean includeVoided,
+	                                                  Date asOfDate) throws DAOException;
```
wluyima: Same here, you don't need the throws clause

api/src/main/java/org/openmrs/CohortMembership.java
vshl: What would that condition be? Do you mean instead of -1 or an additional condition?

api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -164,47 +165,92 @@ public void removeMember(Integer memberId) {
...
 	
+	/**
+	 * @since 2.1.0
+	 */
 	public void removeMember(Patient patient) {
-		List<CohortMembership> memberToRemoveList = getMembers().stream()
-				.filter(m -> m.getPatient().getPatientId().equals(patient.getPatientId())).collect(Collectors.toList());
-		memberToRemoveList.forEach(m -> m.setEndDate(new Date()));
+		CohortMembership memberToRemove = this.getActiveMembership(patient);
+		if (memberToRemove.getPatient().getPatientId().equals(patient.getPatientId())) {
```
djazayeri: what if memberToRemove is null?

api/src/main/java/org/openmrs/CohortMembership.java
wluyima: I guess my question should have been why you chose return -1 over 1, am not suggesting another over the other, implementations of compareTo typically return 1, 0 or -1 but I see you only return just two of them. You could  chose to return -1 if startDate is earlier otherwise 1 for active memberships. And always push stopped memberships to the end of the collections with the one with the latest start date coming dead last.

api/src/main/java/org/openmrs/Cohort.java
wluyima: Good question, if patient is null for methods that add or remove patients or memberships, the code should just return immediately. Same applies to similar methods in CohortService. 

Do we want these methods to have a boolean return type? Where true is returned only if the patient or member is actually added or removed.

api/src/test/java/org/openmrs/api/handler/PatientDataUnvoidHandlerTest.java
```diff
@@ -144,4 +149,36 @@ public void handle_shouldNotUnvoidTheOrdersAndEncountersThatNeverGotVoidedWithTh
...
+		
+		patient = Context.getPatientService().voidPatient(patient, "Void Reason");
+		Assert.assertTrue(patient.getVoided());
+		
+		Assert.assertTrue(membership.getVoided());
+		Assert.assertNotNull(membership.getDateVoided());
+		Assert.assertNotNull(membership.getVoidedBy());
+		Assert.assertNotNull(membership.getVoidReason());
+		
+		patient.setVoided(false);
```
wluyima: You should get rid of line 175 above because it's not the job of your code to call that, the reality is that the AOP around the unvoid method does reset those fields, this could be the reason the test is troubling you.

api/src/main/java/org/openmrs/Cohort.java
vshl: I see. Please let me know how to treat these methods.

api/src/main/java/org/openmrs/CohortMembership.java
vshl: > And always push stopped memberships to the end of the collections with the one with the latest start date coming dead last.

Do you want this in the compareTo() method? I was trying to find a way to do this: One method is to execute [Collections.rotate(List, -1) ](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#rotate(java.util.List,%20int)). Is there any other method?

api/src/test/java/org/openmrs/api/handler/PatientDataUnvoidHandlerTest.java
vshl: Removing line 175 I get the validation exception: ```org.openmrs.api.ValidationException: 'Cohort id=2 name=Example Cohort size=2' failed to validate with reason: memberships: Cohort.patientAndMemberShouldBeVoided```. That is my exact concern.

wluyima: You need to look at the stacktrace and see when the validation occurs that leads to the exception getting thrown, are you making any call to saveCohort() any where before calling PatientService.patientUnvoided()?

api/src/main/java/org/openmrs/Cohort.java
wluyima: You should be changing their return type boolean, i.e they should return true if a membership was actually added or removed otherwise false

api/src/test/java/org/openmrs/api/handler/PatientDataUnvoidHandlerTest.java
wluyima: I think what's happening is that the PatientDataUnvoidHandler is getting called before BaseUnVoidHandler which should be marking the patient as unvoided, you might want to add an order value to the @Handler annotations on both these handlers to guarantee that BaseUnVoidHandler is always called before PatientDataUnvoidHandler

api/src/main/java/org/openmrs/CohortMembership.java
wluyima: Doesn't matter really how you implement it, remember that the memberships are added to a TreeSet, so it would be nice to implement compareTo in a standard way where you return 1, 0, -1 and also have unit tests to back it up

api/src/test/java/org/openmrs/api/handler/PatientDataUnvoidHandlerTest.java
vshl: saveCohort() is called in CohortServiceImpl.patientUnvoided(). That's where the validation exception occurs. Removing the saveCohort() line actually fixes the test. Does patientUnvoided() need a saveCohort()?

Also, I added order of various magnitudes (10, 50, 100), the exception still happens:
```
@Handler(supports = {Patient.class}, order = 100)
public class PatientDataUnvoidHandler implements UnvoidHandler<Patient> {
```

wluyima: You need to set the appropriate order attribute to both handlers not just this one

wluyima: I think you missed one of my comments, you want this handler to be called before BaseUnvoidHandler, you would either need to set the appropriate order value on both BaseUnVoidHandler and PatientDataUnvoidedHandler, the order value works just the same as Spring's order value. 

Actually I think it can work when you set the order only on PatientDataUnvoidHandler but it would need to be the lowest precedence which I think is Integer.MAX_VALUE but I wouldn't recommend this since another module might correctly wish its handler to go last

wluyima: If only setting the order on PatientDataUnvoidHandler to Integer.MAX_VALUE works, then I guess it's okay not to add one to BaseUnvoidHandler

vshl: Adding order to both PatientDataUnvoidHandler and BaseUnvoidHandler works, but now there is an issue where the patient is already unvoided and I have these conditions in the CohortServiceImpl.patientUnvoided:

```
			List<CohortMembership> membersToUnvoid = cohort.getMemberships().stream()
					.filter(m -> m.getVoided() &&
							m.getPatient().getPatientId().equals(patient.getPatientId()) &&
							m.getVoidedBy().equals(voidedBy) &&
							DateUtils.isSameDay(m.getDateVoided(), dateVoided) &&
							m.getVoidReason().equals(voidReason))
```

I can only compare the patient id right now.

wluyima: What's the problem you're running into if the patient is already  unvoided? Is it that getMemberships() isn't returning the patient's memberships since they are voided? And I recall mentioning  that you'd need a `Cohort.getMemberships(boolean includeVoided)` method that returns memberships and possibly `Cohort.getMemberhips(Patient patient, boolean includeVoided)` that returns memberships for the specified patient, they'd be possibly need a boolean field that specifies whether to include voided memberships.

From what I know, the framework calls the handlers' unvoid method with the original voidReason before it's nullified, are you sure you aren't calling `CohortService. patientUnvoided(..)` with the wrong argument value?

vshl: When the patient is unvoided, the patient argument that go into PatientDataUnvoidHandler.handle method has the voidReason value nulled. I can only make use of the patientId, originalVoidingUser and origParentVoidDate values.
```
	@Override
	public void handle(Patient patient, User originalVoidingUser, Date origParentVoidedDate, String unused) {
```

The method signature for patientUnvoided has the nulled voidReason argument that I cannot use for comparison. Removing this argument fixes the entire unvoiding workflow.

vshl: I'll probably commit the changes to get a better understanding.

wluyima: You can ignore voidReason when comparing

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -99,6 +118,18 @@ public void setEndDate(Date endDate) {
...
-		return this.getPatient().getPatientId() - o.getPatient().getPatientId();
+		if (Objects.equals(this.getPatient().getPatientId(), o.getPatient().getPatientId())
+				&& Objects.equals(this.getCohort().getCohortId(), o.getCohort().getCohortId())
+				&& this.getStartDate().equals(o.getStartDate())
+				&& ((this.getEndDate() != null && o.getEndDate() != null
+				&& this.getEndDate().equals(o.getEndDate()))
+				|| (this.getEndDate() == null && o.getEndDate() == null))) {
+			return 0;
+		} else if ((this.isMembershipActive() && !o.isMembershipActive())
+			|| (!this.isMembershipActive() && o.isMembershipActive())) {
```
wluyima: You need to always push the voided memberships to the end of the list, i.e. if this membership is active and the other is inactive, you should return -1

wluyima: You should compare start dates, you could return -1 if startDate is earlier otherwise 1 for active memberships

vshl: Is this the method where I should move the voided memberships to the end of the collection? I don't have access to the collection here.

api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -164,47 +165,95 @@ public void removeMember(Integer memberId) {
...
 	}
 
-	public List<CohortMembership> getMemberships(Date asOf) {
-		return getMembers().stream()
-				.filter(m -> !m.getStartDate().before(asOf)).collect(Collectors.toList());
+	/**
+	 * @since 2.1.0
+	 * @should get memberships of a cohort as of a date
+	 */
+	public List<CohortMembership> getMemberships(Date asOfDate) {
```
bmamlin: Per design discussion, we'd like to rename this method to `getActiveMemberships(Date asOfDate)` (and we will add a `getActiveMemberships()` that calls this with today's date.

api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -164,47 +165,95 @@ public void removeMember(Integer memberId) {
...
+		return false;
 	}
-
-	public void addMembership(CohortMembership cohortMembership) {
+	
+	/**
+	 * @since 2.1.0
+	 */
+	public boolean addMembership(CohortMembership cohortMembership) {
 		if (!this.contains(cohortMembership.getPatient().getPatientId())) {
```
wluyima: You need to check that cohortMembership is not null

api/src/main/java/org/openmrs/CohortMembership.java
wluyima: compareTo is what gets called  when sorting since you are using a TreeSet, meaning that if you implement it this way, the voided memberships will automatically get pushed to the end, so you don't need to implement it anywhere else

vshl: Understood. Since I am already comparing startDates in the first if condition, won't it taken care of already. What will be the if condition to return 1? I didn't understand from your last comment. This is what I've got:

```
		if (Objects.equals(this.getPatient().getPatientId(), o.getPatient().getPatientId())
				&& Objects.equals(this.getCohort().getCohortId(), o.getCohort().getCohortId())
				&& this.getStartDate().equals(o.getStartDate())
				&& ((this.getEndDate() != null && o.getEndDate() != null
				&& this.getEndDate().equals(o.getEndDate()))
				|| (this.getEndDate() == null && o.getEndDate() == null))) {
			return 0;
		} else if ((this.isMembershipActive() && !o.isMembershipActive())
			|| (!this.isMembershipActive() && o.isMembershipActive())) {
			return -1;
		} else if () {
			return 1;
		}
```

wluyima: If this membership is active and the other is inactive, you should return -1, then return 1 for the opposite, in your code above you are returning -1 both cases which I think is incorrect.

The other thing is that the first condition isn't sorting by startDates since you always return 0 which means they are equal.

api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -164,47 +165,99 @@ public void removeMember(Integer memberId) {
...
+		return getMemberships().stream()
+				.filter(m -> m.isMembershipActive() && m.getPatient().getPatientId().equals(patient.getPatientId()))
+				.collect(Collectors.toList())
+				.get(0);
+	}
+	
+	/**
+	 * @since 2.1.0
+	 */
 	public void purgeMemberships(List<CohortMembership> cohortMembershipList) {
```
wluyima: I would also assume this should return a boolean value, i.e only return true if the member was found and removed by return the same value  from `getMemberships().remove(m)`

api/src/test/java/org/openmrs/api/CohortServiceTest.java
```diff
@@ -564,8 +574,194 @@ public void getMemberships_shouldGetMembershipsAsOfADate() throws ParseException
...
+		membership.setStartDate(startDate);
+		cohort.addMembership(membership);
+		assertTrue(cohort.contains(patient.getPatientId()));
+		
+		Date date = dateFormat.parse("2017-01-01 00:00:00");
+		assertEquals(service.getCohortsContainingPatient(patient, false, date).size(), 1);
+	}
+
+	@Test
+	public void union_shouldContainVoidedAndExpiredMemberships() throws Exception {
```
wluyima: These tests for union/intersection/subtract etc belong to CohortTest.java (add it if it doesn't exist) because the methods are found in Cohort.java.

api/src/test/java/org/openmrs/api/CohortServiceTest.java
```diff
@@ -564,8 +574,194 @@ public void getMemberships_shouldGetMembershipsAsOfADate() throws ParseException
...
+
+		Cohort cohortTwo = new Cohort(4);
+		CohortMembership membershipTwo = new CohortMembership(new Patient(8), startDate);
+		membershipTwo.setVoided(true);
+		membershipTwo.setEndDate(endDate);
+		cohortTwo.addMembership(membershipTwo);
+
+		Cohort cohortUnion = Cohort.union(cohortOne, cohortTwo);
+		Collection<CohortMembership> unionOfMemberships = cohortUnion.getMemberships();
+		assertTrue(unionOfMemberships.stream().anyMatch(m -> m.getVoided() || m.getEndDate() != null));
```
wluyima: It would be better to explicitly check that both members exist in the union cohort.

api/src/test/java/org/openmrs/api/CohortServiceTest.java
```diff
@@ -564,8 +574,194 @@ public void getMemberships_shouldGetMembershipsAsOfADate() throws ParseException
...
+
+		Cohort cohortTwo = new Cohort(4);
+		CohortMembership membershipTwo = new CohortMembership(new Patient(8), startDate);
+		membershipTwo.setVoided(true);
+		membershipTwo.setEndDate(endDate);
+		cohortTwo.addMembership(membershipTwo);
+
+		Cohort cohortSubtract = Cohort.subtract(cohortOne, cohortTwo);
+		Collection<CohortMembership> subtractOfMemberships = cohortSubtract.getMemberships();
+		assertTrue(subtractOfMemberships.stream().anyMatch(m -> m.getVoided() || m.getEndDate() != null));
```
wluyima: Would be nice to explicitly check that the resulting cohort only contains the expected members.

wluyima: I thought subtract should return members that are only in cohort one and not in two. Otherwise the test above isn't concrete enough since there is no common members.

wluyima: I think you're missing the test for the intersect method. These methods are expected to behave just like set logic in mathematics

api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -164,47 +165,99 @@ public void removeMember(Integer memberId) {
...
-				.filter(m -> m.equals(cohortMembership) ||
-						m.getPatient().getPatientId().equals(cohortMembership.getPatient().getPatientId()))
+		cohortMembership.setEndDate(new Date());
+	}
+	
+	/**
+	 * @since 2.1.0
+	 * @should return voided or unvoided memberships
+	 */
+	public List<CohortMembership> getMemberships(Boolean includeVoided) {
```
djazayeri: shouldn't this take a lower-case boolean as the parameter?

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -43,8 +47,23 @@ public CohortMembership(Patient patient) {
 		this(patient, new Date());
 	}
 	
-	public boolean isMemberActive() {
-		return this.getStartDate() != null && !getStartDate().after(new Date()) && this.getEndDate() == null;
+	/**
+	 * @should return active memberships as of current date
+	 */
+	public boolean isMembershipActive() {
```
djazayeri: personally I would have named these methods "isActive" (because it's already on a CohortMembership object). Not a huge deal.

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -10,10 +10,14 @@
...
 
 import java.util.Date;
+import java.util.Objects;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * @since 2.1.0
+ */
```
djazayeri: (copying a comment I made on the original commit)

@vshl @wluyima when I see this I wonder why CohortMembership has a "Patient patient" field instead of just an "Integer patientId". Wouldn't that more correctly capture the data that's available?

Cohort used to just have a Set; are we intentionally now using the fact that it's a Patient somewhere?
Otherwise I propose we change CohortMembership to just have the Integer patientId.

(Note that doing `new Patient(Integer)` will have a randomly-generated UUID, which could be misleading to consuming code, e.g. when I try to add this in the REST webservices module.

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -99,6 +118,20 @@ public void setEndDate(Date endDate) {
 
 	@Override
 	public int compareTo(CohortMembership o) {
-		return this.getPatient().getPatientId() - o.getPatient().getPatientId();
+		if (Objects.equals(this.getPatient().getPatientId(), o.getPatient().getPatientId())
+				&& Objects.equals(this.getCohort().getCohortId(), o.getCohort().getCohortId())
+				&& this.getStartDate().equals(o.getStartDate())
```
djazayeri: Prefer to use OpenmrsUtil.compare method for dates, since sometimes what you get back from hibernate is a java.sql.Date which is not directly comparable to a java.util.Date.

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -99,6 +118,20 @@ public void setEndDate(Date endDate) {
 
 	@Override
 	public int compareTo(CohortMembership o) {
-		return this.getPatient().getPatientId() - o.getPatient().getPatientId();
+		if (Objects.equals(this.getPatient().getPatientId(), o.getPatient().getPatientId())
```
djazayeri: This method is written in a confusing way. Further, because of the use of isMembershipActive() I'm not sure it actually behaves correctly for a compareTo method.

I suggest writing it more like:

    int temp = this.getPatient().getPatientId().compareTo(o.getPatient().getPatientId());
    if (temp == 0) {
        temp = OpenmrsUtil.compareWithNullAsEarliest(this.getStartDate(), o.getStartDate());
    }
    if (temp == 0) {
        temp = OpenmrsUtil.compareWithNullAsEarliest(this.getEndDate(), o.getEndDate());
    }
    return temp;

api/src/main/java/org/openmrs/Cohort.java
wluyima: Yap, good catch!

api/src/main/java/org/openmrs/CohortMembership.java
wluyima: I agree

api/src/main/java/org/openmrs/CohortMembership.java
wluyima: The use of isActive was to push voided memberships to end of the TreeSet

api/src/main/java/org/openmrs/CohortMembership.java
wluyima: All date comparisons need to delegate to Date.compareTo() or OpenmrsUtil.compare(Date, Date) to guarantee consistency

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -99,6 +118,20 @@ public void setEndDate(Date endDate) {
...
+				|| (this.getEndDate() == null && o.getEndDate() == null))) {
+			return 0;
+		} else if ((this.isMembershipActive() && !o.isMembershipActive())
+				|| (this.startDate.before(o.startDate))) {
+			return -1;
+		} else if ((!this.isMembershipActive() && o.isMembershipActive())
+				|| (this.startDate.after(o.startDate))) {
+			return 1;
+		}
+		return -1;
```
wluyima: You should split up each of the last 2 `else if` clauses above so that they only deal with active vs no active memberships, replace this last return statement  by delegating to OpenmrsUtil by calling `return OpenmrsUtil.compare(Date, Date)`

api/src/main/java/org/openmrs/CohortMembership.java
vshl: Will the hibernate change as well?
```
		<many-to-one name="patient" column="patient_id" class="Patient" not-null="true"/>
```

api/src/main/java/org/openmrs/CohortMembership.java
vshl: `return OpenmrsUtil.compare(Date, Date)` seems to be breaking some tests in CohortTest. 

screen shot 2017-04-14 at 6 05 29 pm

Assert of `cohort.contains(id)` fails. It can no longer match patient ids. Maybe because the ids are different and the start dates match?

api/src/main/java/org/openmrs/CohortMembership.java
vshl: I changed Patient patient to Integer patientId. However lot of tests are failing right now.

```
ERROR - BasicPropertyAccessor$BasicSetter.set(121) |2017-04-14 19:41:50,315| HHH000123: IllegalArgumentException in class: org.openmrs.CohortMembership, setter method of property: patientId
ERROR - BasicPropertyAccessor$BasicSetter.set(122) |2017-04-14 19:41:50,316| HHH000091: Expected type: java.lang.Integer, actual value: org.openmrs.Patient_$$_jvst4ea_28

IllegalArgumentException occurred while calling setter for property [org.openmrs.CohortMembership.patientId (expected type = java.lang.Integer)]; target = [CohortMembership[hashCode=c700c9ba,uuid=2bc2a1bf-bc54-410d-8721-7fb16375be91]], property value = [Patient#2] setter of org.openmrs.CohortMembership.patientId
IllegalArgumentException occurred while calling setter for property [org.openmrs.CohortMembership.patientId (expected type = java.lang.Integer)]; target = [CohortMembership[hashCode=c700c9ba,uuid=2bc2a1bf-bc54-410d-8721-7fb16375be91]], property value = [Patient#2]
```

It looks like something's a miss. Is it because of the <many-to-one> property in the hbm?

wluyima: You have to update the hbm file

wluyima: i.e CohortMembership.hbm.xml

api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -302,29 +361,22 @@ public void setName(String name) {
...
-		this.memberIds = new TreeSet<Integer>(memberIds);
-		for (Integer id : memberIds) {
-			addMembership(new CohortMembership(new Patient(id)));
-		}
-	}
-
-	public Collection<CohortMembership> getMembers() {
-		if (members == null) {
-			members = new TreeSet<CohortMembership>();
+		Set<Integer> ids = new TreeSet<Integer>(memberIds);
```
wluyima: Isn't this going to cause issues with hibernate if this is in the session? Because hibernate will have a different instance of the collection from the new one you are creating here. You should have this method just assign the new value. Yours and client code should never call this but instead addMember()

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -34,17 +39,33 @@
...
+	 */
+	public boolean isActive(Date asOfDate) {
+		Date date;
+		if (asOfDate == null) {
+			date = new Date();
+		} else {
+			date = asOfDate;
+		}
+		return !this.getVoided()
+				&& (date.equals(this.getStartDate()) || date.after(this.getStartDate()))
```
wluyima: Actually this isn't correct, if `date == startDate` then it's not active

wluyima: As a Darius suggested, you should using OpenmrsUti.compare(Date, Date) to compare dates

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -99,6 +120,20 @@ public void setEndDate(Date endDate) {
...
+				&& ((this.getEndDate() != null && o.getEndDate() != null
+				&& OpenmrsUtil.compare(this.getEndDate(), o.getEndDate()) == 0)
+				|| (this.getEndDate() == null && o.getEndDate() == null))) {
+			ret = 0;
+		} else if (this.isActive() && !o.isActive()) {
+			ret = -1;
+		} else if (!this.isActive() && o.isActive()) {
+			ret = 1;
+		}
+		return ret;
```
wluyima: This should not return an arbitrary value, you need to change it to:

`return OpenmrsUtil.compare(this.startDate(), o.startDate());`

djazayeri:
As I'm looking to implement the REST web service sub-resource for cohort member, I believe I need to have getCohortMembershipByUuid(String) and voidCohortMembership(CohortMembership) in CohortService.
(I can add these myself after this is merged if need be.)

wluyima:
I guess, I doubt though if we've had to do the same for all the other child collections

api/src/main/java/org/openmrs/CohortMembership.java
vshl: I'm confused, shouldn't it be equal or after the start date to be active?

api/src/main/java/org/openmrs/CohortMembership.java
vshl: _Copying my comment from an older commit:_

`return OpenmrsUtil.compare(Date, Date)` seems to be breaking some tests in CohortTest. 

screen shot 2017-04-14 at 6 05 29 pm

Assert of `cohort.contains(id)` fails sometimes. It can no longer match patient ids. 
```
		Arrays.stream(ids).forEach(id -> assertTrue(cohort.contains(new Patient(id))));
```

vshl: Is there a simpler implementation that could be done? Like how Darius suggested?

api/src/main/java/org/openmrs/CohortMembership.java
wluyima: Never mind, am merging the PR and will address any remaining issues, thanks!

wluyima: It was an oversight on my side

wluyima:
Merged at 9ae34fc55b87d25c9ce20967cafc926ba8cb0571

PR #2065, opened 2017-03-13
TRUNK-5110 | Fix NPE when a provider doesn't have a role

pushpa446:
Collectors.groupingby does not allow null as a key value.
Workaround is to use Collectors.toMap.

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-5110

## Checklist:

### DIFF
--- a/api/src/main/java/org/openmrs/Encounter.java
+++ b/api/src/main/java/org/openmrs/Encounter.java
@@ -14,6 +14,7 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.Deque;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -532,7 +533,14 @@ public class Encounter extends BaseOpenmrsData {
 		
 		return encounterProviders.stream()
 				.filter(ep -> includeVoided || !ep.getVoided())
-				.collect(Collectors.groupingBy(EncounterProvider::getEncounterRole, Collectors.mapping(EncounterProvider::getProvider, Collectors.toSet())));
+				.collect(Collectors.toMap((ep) -> (ep).getEncounterRole(),
+                                          (ep) -> {
+                                                    Set s = new HashSet<Provider>();
+                                                    s.add((ep).getProvider());
+                                                    return s; },
+                                          (ep1,ep2) -> {
+                                                         ep1.addAll(ep2);
+                                                         return  ep1; }));
 		
 	}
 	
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/Encounter.java
```diff
@@ -532,7 +533,14 @@ public void setVisit(Visit visit) {
 		
 		return encounterProviders.stream()
 				.filter(ep -> includeVoided || !ep.getVoided())
-				.collect(Collectors.groupingBy(EncounterProvider::getEncounterRole, Collectors.mapping(EncounterProvider::getProvider, Collectors.toSet())));
+				.collect(Collectors.toMap((ep) -> (ep).getEncounterRole(),
```
teleivo: you seem to have stumbled on a java bug https://bugs.openjdk.java.net/browse/JDK-8148463 :)

I feel that since we cant do anything about the bug we should not use streams here since its hard to read. but thats just IMHO

what about sticking to the traditional kind:

```java
		for (EncounterProvider ep : encounterProviders) {
			if (includeVoided || !ep.getVoided()) {
				if (!result.containsKey(ep.getEncounterRole())) {
					result.put(ep.getEncounterRole(), new HashSet<>());
				}
				result.get(ep.getEncounterRole()).add(ep.getProvider());
			}
		}
		return result;
```

maybe others @dkayiwa or @wluyima or our master of streams @lluismf want to comment as well

dkayiwa: I would like to comment after @lluismf has done so. 😊

teleivo: @lluismf can we please get your input on this one (please see comments).

lluismf: I wasn't aware role was optional, sorry! I also agree with Ivo, it's a bit hard to read. Can you please revert to the previous implementation without streams? I'd do it but I don't have permissions.

pushpa446: Can I modify it to previous implementation and add a commit in this pull request? @teleivo

teleivo: @pushpa446 what is the additional commit? just the test? than yes 👍

lluismf: Sorry to insist, but does it make any sense to add a NULL role to an encounter provider, even if the DB prevents so ?

teleivo: @djazayeri can you please join our discussion. by making use of streams here we ran into a NPE (due to a java bug) which @pushpa446 would fix with this change. see explanation starting from my 2 post. but this also lead to @lluismf fundamental question of why null is allowed in in the first place.

djazayeri: @pushpa446 and I talked about this. We thought this was a bug because it appeared when we updated bahmni-core to build against the latest openmrs-core, but we now realize that actually our Bahmni test data was invalid. We will cancel this PR and ticket, and just fix our Bahmni test data to include an encounter role.

PR #2067, opened 2017-03-15
TRUNK-5032 :Removed 4 unused private fields

Nuwantha:
## Description
The unused private variables were removed. Private Log variables were remained unchanged.

## Related Issue
see https://issues.openmrs.org/browse/TRUNK-5032

## Checklist:
- [ x] My pull request only contains one single commit.
- [ x] My pull request is based on the latest master branch
- [ x] My code follows the code style of this project.
- [ x] All new and existing tests passed.

### DIFF
--- a/api/src/main/java/org/openmrs/scheduler/Schedule.java
+++ b/api/src/main/java/org/openmrs/scheduler/Schedule.java
@@ -49,11 +49,6 @@ public class Schedule {
 	private boolean startOnStartup;
 	
 	/**
-	 * The date format pattern used to set the start date when start date is passed in as a string
-	 */
-	private String dateFormat;
-	
-	/**
 	 * Default TODO Move to constants or properties
 	 */
 	public static final String DEFAULT_DATE_FORMAT = "MM-dd-yyyy hh:mm:ss";
@@ -228,21 +223,6 @@ public class Schedule {
 	 */
 	public void setRepeatInterval(long repeatInterval) {
 		this.repeatInterval = repeatInterval;
-	}
-	
-	/**
-	 * Get the date format used to set the start time.
-	 */
-	public String getDateFormat() {
-		return this.dateFormat;
-	}
-	
-	/**
-	 * Sets the date format used to set the start time.
-	 */
-	public void setDateFormat(String dateFormat) {
-		this.dateFormat = dateFormat;
-		this.dateFormatter = new SimpleDateFormat(dateFormat);
 	}
 	
 	/**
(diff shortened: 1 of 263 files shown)


api/src/main/java/org/openmrs/scheduler/Schedule.java
```diff
@@ -231,21 +226,6 @@ public void setRepeatInterval(long repeatInterval) {
 	}
 	
 	/**
-	 * Get the date format used to set the start time.
-	 */
-	public String getDateFormat() {
```
teleivo: @dkayiwa should we deprecate this first? I mean its not used in the core but what if other modules use it? Or does it suffice to put this in a changelog that from 2.2.0 they do not exist anymore

web/src/main/java/org/openmrs/web/WebUtil.java
```diff
@@ -417,23 +416,20 @@ public boolean supportsPropertyName(String propertyName) {
 		return OpenmrsConstants.GP_SEARCH_DATE_DISPLAY_FORMAT.equals(propertyName);
 	}
 	
-	public static void setDefaultDateCache(String defaultDateCache) {
```
teleivo: @dkayiwa same here

api/src/main/java/org/openmrs/scheduler/Schedule.java
dkayiwa: Since it was not used, i do not think any module is using it for any meaningful work. And even if any module used it, it would just remove its usage without any effects.
So no need for deprecation. 😊

web/src/main/java/org/openmrs/web/WebUtil.java
dkayiwa: Same response as above.

api/src/main/java/org/openmrs/scheduler/Schedule.java
mogoodrich: It could potentially cause module compile errors, though, no?

Or, if we aren't careful and run a module against the latest version of core without testing that it compiles, could result in runtime errors?

That being said, these do look like little-used methods.

web/src/main/java/org/openmrs/web/WebUtil.java
Nuwantha: @teleivo , @dkayiwa 
Thank you for your comments. sorry for delay. Could I know what exact changes I need to do ?

api/src/main/java/org/openmrs/scheduler/Schedule.java
teleivo: @Nuwantha I think its best you exclude these for now from your PR. I am unsure if we should simply remove them.

web/src/main/java/org/openmrs/web/WebUtil.java
teleivo: @Nuwantha same here: I think its best you exclude these for now from your PR. I am unsure if we should simply remove them.

PR #2069, opened 2017-03-15
TRUNK-5114 Using different logging implementations instead of facade

teleivo:
## Description
we exclude the apache commons-logging dependency coming from springframework
and configure the slf4j facade with the log4j implementation but still
use commons-logging or log4j directly.

* replace usages by using the slf4j facade
* add "org.apache.commons.logging" to checkstyle's IllegalImport rule
so we get notified of anyone trying to use it
* replace log.fatal("message", exception)

with

log.error(MarkerFactory.getMarker("FATAL"), "message", e);

since slf4j does not provide the level FATAL and suggests the use of
markers instead

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-5114

### DIFF
--- a/api/src/main/java/org/openmrs/CohortMembership.java
+++ b/api/src/main/java/org/openmrs/CohortMembership.java
@@ -11,14 +11,14 @@ package org.openmrs;
 
 import java.util.Date;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class CohortMembership extends BaseOpenmrsData implements Comparable<CohortMembership> {
 	
 	public static final long serialVersionUID = 0L;
 	
-	protected static final Log log = LogFactory.getLog(CohortMembership.class);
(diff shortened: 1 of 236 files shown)


dkayiwa:
@teleivo this was a huge commit! 
Did you use some sort of tool to automate the changes? Or did you manually do them one by one? 😊

teleivo:
@dkayiwa I used Eclipse's file find/replace + the Eclipse "clean up" to organize the imports. This worked for most of the changes. I had to adapt a few of the `log.error(exception)` which needs to be `log.error("message", exception)` and `log.fatal()` which needs markers now. I will post a how to or write up a wiki for it tomorrow :)

dkayiwa:
I can see that you are an eclipse fun, just like me. Do you just hate intelliJ? 😊

teleivo:
@dkayiwa I think Eclipse UI is less bloated and has some nice clean up features, but IntelliJ is way better at importing projects and keeping it in sync with changes in maven pom's. Eclipse keeps crashing at the moment, so I move quickly from liking to disliking 😭 🤕 😄

PR #2076, opened 2017-03-17
TRUNK-5057 :Extracted private methods in OrderServiceImpl.saveOrder

Nuwantha:
## Description
TRUNK-5057 : Extracted private methods in OrderServiceImpl.saveOrder

## Related Issue

see https://issues.openmrs.org/browse/TRUNK-5057

## Checklist:

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
@@ -82,7 +82,7 @@ public class OrderServiceImpl extends BaseOpenmrsService implements OrderService
 	protected OrderDAO dao;
 	
 	private static OrderNumberGenerator orderNumberGenerator = null;
-	
+
 	public OrderServiceImpl() {
 	}
 	
@@ -126,75 +126,18 @@ public class OrderServiceImpl extends BaseOpenmrsService implements OrderService
 	public synchronized Order saveRetrospectiveOrder(Order order, OrderContext orderContext) {
 		return saveOrder(order, orderContext, true);
 	}
-	
+
 	private Order saveOrder(Order order, OrderContext orderContext, boolean isRetrospective) {
-		if (order.getOrderId() != null) {
-			throw new UnchangeableObjectException("Order.cannot.edit.existing");
-		}
-		if (order.getDateActivated() == null) {
-			order.setDateActivated(new Date());
-		}
-		boolean isDrugOrder = DrugOrder.class.isAssignableFrom(getActualType(order));
-		Concept concept = order.getConcept();
-		if (concept == null && isDrugOrder) {
-			DrugOrder drugOrder = (DrugOrder) order;
-			if (drugOrder.getDrug() != null) {
-				concept = drugOrder.getDrug().getConcept();
-				drugOrder.setConcept(concept);
-			}
-		}
-		if (isDrugOrder) {
-			((DrugOrder) order).setAutoExpireDateBasedOnDuration();
-		}
-		
-		if (concept == null) {
-			throw new MissingRequiredPropertyException("Order.concept.required");
-		}
+
+		failOnExistingOrder(order);
+		ensureDateActivatedIsSet(order);
+		ensureConceptIsSet(order);
+		ensureDrugOrderAutoExpirationDateIsSet(order);
+		ensureOrderTypeIsSet(order,orderContext);
+		ensureCareSettingIsSet(order,orderContext);
+		failOnOrderTypeMismatch(order);
 		
 		Order previousOrder = order.getPreviousOrder();
-		if (order.getOrderType() == null) {
-			OrderType orderType = null;
-			if (orderContext != null) {
-				orderType = orderContext.getOrderType();
-			}
-			if (orderType == null) {
-				orderType = getOrderTypeByConcept(concept);
-			}
-			//Check if it is instance of DrugOrder
-			if (orderType == null && order instanceof DrugOrder) {
-				orderType = Context.getOrderService().getOrderTypeByUuid(OrderType.DRUG_ORDER_TYPE_UUID);
-			}
-			//Check if it is an instance of TestOrder
-			if (orderType == null && order instanceof TestOrder) {
-				orderType = Context.getOrderService().getOrderTypeByUuid(OrderType.TEST_ORDER_TYPE_UUID);
-			}
-			
-			//this order's order type should match that of the previous
-			if (orderType == null) {
-				throw new OrderEntryException("Order.type.cannot.determine");
-			}
-			if (previousOrder != null && !orderType.equals(previousOrder.getOrderType())) {
-				throw new OrderEntryException("Order.type.does.not.match");
-			}
-			
-			order.setOrderType(orderType);
-		}
-		if (order.getCareSetting() == null) {
-			CareSetting careSetting = null;
-			if (orderContext != null) {
-				careSetting = orderContext.getCareSetting();
-			}
-			if (careSetting == null || (previousOrder != null && !careSetting.equals(previousOrder.getCareSetting()))) {
-				throw new OrderEntryException("Order.care.cannot.determine");
-			}
-			order.setCareSetting(careSetting);
-		}
-		
-		if (!order.getOrderType().getJavaClass().isAssignableFrom(order.getClass())) {
-			throw new OrderEntryException("Order.type.class.does.not.match", new Object[] {
-			        order.getOrderType().getJavaClass(), order.getClass().getName() });
-		}
-		
 		if (REVISE == order.getAction()) {
 			if (previousOrder == null) {
 				throw new MissingRequiredPropertyException("Order.previous.required", (Object[]) null);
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -82,7 +82,14 @@
 	protected OrderDAO dao;
 	
 	private static OrderNumberGenerator orderNumberGenerator = null;
-	
+
+	private Concept concept;
+
+	private boolean isDrugOrder;
+
+	private Order previousOrder;
```
teleivo: @Nuwantha why did you introduce these members? they are not needed

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -126,16 +133,19 @@ public OrderGroup saveOrderGroup(OrderGroup orderGroup) throws APIException {
 	public synchronized Order saveRetrospectiveOrder(Order order, OrderContext orderContext) {
 		return saveOrder(order, orderContext, true);
 	}
-	
-	private Order saveOrder(Order order, OrderContext orderContext, boolean isRetrospective) {
-		if (order.getOrderId() != null) {
-			throw new UnchangeableObjectException("Order.cannot.edit.existing");
-		}
+
+	private void failOnExistingOrder(Order order) {
```
teleivo: @Nuwantha I just checked the issue on jira, sorry for this but since the issue was created the line numbers must have changed.

the `failOnExistingOrder(Order order)` is of course not for setting the dateActivated but for the first if block that throws the UnchangeableObjectException

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -126,16 +133,19 @@ public OrderGroup saveOrderGroup(OrderGroup orderGroup) throws APIException {
...
 		if (order.getDateActivated() == null) {
 			order.setDateActivated(new Date());
 		}
-		boolean isDrugOrder = DrugOrder.class.isAssignableFrom(getActualType(order));
-		Concept concept = order.getConcept();
+	}
+
+	private void ensureDateActivatedIsSet(Order order) {
```
teleivo: `ensureDateActivatedIsSet` is for setting dateActivated

this should be clear from the method names

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -126,16 +133,19 @@ public OrderGroup saveOrderGroup(OrderGroup orderGroup) throws APIException {
...
+		isDrugOrder = DrugOrder.class.isAssignableFrom(getActualType(order));
+		concept = order.getConcept();
```
teleivo: this is for ensureConceptIsSet

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -189,11 +201,26 @@ private Order saveOrder(Order order, OrderContext orderContext, boolean isRetros
...
+
+	private void failOnOrderTypeMismatch(Order order) {
 		if (!order.getOrderType().getJavaClass().isAssignableFrom(order.getClass())) {
 			throw new OrderEntryException("Order.type.class.does.not.match", new Object[] {
-			        order.getOrderType().getJavaClass(), order.getClass().getName() });
+					order.getOrderType().getJavaClass(), order.getClass().getName() });
+		}
+	}
+
+	private Order saveOrder(Order order, OrderContext orderContext, boolean isRetrospective) {
```
teleivo: @Nuwantha currently I cannot access jira and adapt the issue. But it should be clear from the method names what the method does and since the code snippet in the issue of how it should look in the end is correct and should also help you. furthermore please put the private methods after the method `saveOrder` as stated in the issue. this makes reading of the code easier since it follow the natural read from top to bottom and simply flows :)

Nuwantha:
@teleivo I changed the code according to your review. could you check that ?

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -146,19 +205,21 @@ private Order saveOrder(Order order, OrderContext orderContext, boolean isRetros
...
 		if (concept == null) {
 			throw new MissingRequiredPropertyException("Order.concept.required");
 		}
-		
+
+	}
+
+	private void ensureOrderTypeIsSet(Order order, OrderContext orderContext) {
 		Order previousOrder = order.getPreviousOrder();
 		if (order.getOrderType() == null) {
```
teleivo: @Nuwantha we can simplify this.
turn this into a guard clause like
```java
if (order.getOrderType() != null) {
  return;
}
Order previousOrder = order.getPreviousOrder();
...
```

this way you remove one level of nesting in this method and you signal the reader that ok if the `orderType` is already set, nothing to do here, return.

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -146,19 +205,21 @@ private Order saveOrder(Order order, OrderContext orderContext, boolean isRetros
...
 			OrderType orderType = null;
 			if (orderContext != null) {
 				orderType = orderContext.getOrderType();
 			}
 			if (orderType == null) {
-				orderType = getOrderTypeByConcept(concept);
+				orderType = getOrderTypeByConcept(order.getConcept());
 			}
 			//Check if it is instance of DrugOrder
```
teleivo: remove all comments in this method, we see from the code that this is what is happening

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -146,19 +205,21 @@ private Order saveOrder(Order order, OrderContext orderContext, boolean isRetros
...
-		
 		if (concept == null) {
 			throw new MissingRequiredPropertyException("Order.concept.required");
 		}
-		
+
+	}
+
+	private void ensureOrderTypeIsSet(Order order, OrderContext orderContext) {
 		Order previousOrder = order.getPreviousOrder();
```
teleivo: move `previousOrder` declaration to right before its first use. see that we declare it here but there are a few cases where we might return before the variable is actually used.

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -146,19 +205,21 @@ private Order saveOrder(Order order, OrderContext orderContext, boolean isRetros
 		if (isDrugOrder) {
 			((DrugOrder) order).setAutoExpireDateBasedOnDuration();
 		}
-		
 		if (concept == null) {
 			throw new MissingRequiredPropertyException("Order.concept.required");
 		}
-		
+
```
teleivo: remove newline at the end of methods

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -168,17 +229,19 @@ private Order saveOrder(Order order, OrderContext orderContext, boolean isRetros
 			if (orderType == null && order instanceof TestOrder) {
 				orderType = Context.getOrderService().getOrderTypeByUuid(OrderType.TEST_ORDER_TYPE_UUID);
 			}
-			
 			//this order's order type should match that of the previous
```
teleivo: same comment doesnt add value, remove

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -126,14 +126,73 @@ public OrderGroup saveOrderGroup(OrderGroup orderGroup) throws APIException {
...
+	}
+
+	private void ensureDateActivatedIsSet(Order order) {
 		if (order.getDateActivated() == null) {
 			order.setDateActivated(new Date());
 		}
+	}
+
+	private void ensureConceptIsSet(Order order) {
 		boolean isDrugOrder = DrugOrder.class.isAssignableFrom(getActualType(order));
```
teleivo: extract this into a private method

```java
private boolean isDrugOrder(Order order)
```

this code is used twice already in this class and I have another suggestion which will make it 3 times, so its worth extracting it into a private method.

after extracting it call the method below in the `if(concept == null && isDrugOrder(order))`, no need for a local variable then

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -146,19 +205,21 @@ private Order saveOrder(Order order, OrderContext orderContext, boolean isRetros
 		if (isDrugOrder) {
```
teleivo: extract this if block into another method called

`ensureDrugOrderAutoExpirationDateIsSet`

reason is this: the method we extracted is `ensureConceptIsSet` but this block takes care of something else so keeping this in here is misleading of what the method is doing. Therefore extract it and with the new helper `isDrugOrder(order)` thats an easy task :)

call this method then in the saveOrder right after ensureConceptIsSet like so

```java
		failOnExistingOrder(order);
		ensureDateActivatedIsSet(order);
		ensureConceptIsSet(order);
		ensureDrugOrderAutoExpirationDateIsSet(order);
		ensureOrderTypeIsSet(order,orderContext);
		ensureCareSettingIsSet(order,orderContext);
		failOnOrderTypeMismatch(order);
```

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -168,17 +229,19 @@ private Order saveOrder(Order order, OrderContext orderContext, boolean isRetros
...
 			if (previousOrder != null && !orderType.equals(previousOrder.getOrderType())) {
 				throw new OrderEntryException("Order.type.does.not.match");
 			}
-			
 			order.setOrderType(orderType);
 		}
+	}
+
+	private void ensureCareSettingIsSet(Order order, OrderContext orderContext) {
+		Order previousOrder = order.getPreviousOrder();
```
teleivo: same here as before.

* make guard clause
```java
if (order.getCareSetting() != null) {
return;
}
```
* move `previousOrder` declaration right before its used.

teleivo:
@Nuwantha also please rebase on latest master

Nuwantha:
@teleivo  Thank you for valuable comments. I did all the requested changes. Could you check it ?

PR #2125, opened 2017-04-02
TRUNK-324:Extract private methods

suthagar23:
To reduce the complexity, increase readability and maintainability in EncounterServiceImple.saveEncounter(Encounter)
I modified related parts into smaller private methods, with longer meaningful names.

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
@@ -88,6 +88,36 @@ public class EncounterServiceImpl extends BaseOpenmrsService implements Encounte
 		    dao.getEncounters(query, null, null, null, includeVoided), null);
 	}
 	
+
+	private void failIfDeniedToEdit(Encounter encounter) throws APIException{
+		throw new APIException("Encounter.error.privilege.required.edit", new Object[] { encounter.getEncounterType()
+			     .getEditPrivilege() });
+	}
+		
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
```diff
@@ -88,6 +88,36 @@ public void setEncounterDAO(EncounterDAO dao) {
 		    dao.getEncounters(query, null, null, null, includeVoided), null);
 	}
 	
+
+	private void failIfDeniedToEdit(Encounter encounter) throws APIException {
```
teleivo: move the methods after they are first used. more natural to read top to bottom

api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
```diff
@@ -96,38 +126,20 @@ public Encounter saveEncounter(Encounter encounter) throws APIException {
 		
 		// if authenticated user is not supposed to edit encounter of certain type
 		if (!canEditEncounter(encounter, null)) {
```
gayanW: Why not move the whole block into failIfDeniedToEdit() instead of just moving APIException?

api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
suthagar23: Sorry for that mistake, I will change the methods.

PR #2130, opened 2017-04-05
TRUNK-4976 - Support obs.interpretation and obs.status, aligned with FHIR

djazayeri:
I will merge this myself. -Darius


wluyima:
Is client code the one expected to change the status to amended when an Obs is edited?

djazayeri:
@wluyima, Burke and I discussed this on the design call yesterday, and we think it would be better if the API sets this automatically. I'm viewing that as a nice-to-have for now, and as it is now the code doesn't do this, and the client is expected to set this field.

wluyima:
Okay, thanks!

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -64,6 +64,32 @@
  */
 public class Obs extends BaseOpenmrsData {
 	
+	public enum Interpretation {
```
wluyima: Wouldn't it be nice to have @since for these new enums?

djazayeri: Good point, I remembered for the getters/setters but I forgot it here. I will add this.

PR #2131, opened 2017-04-05
TRUNK-5140|Increase the size of the obs.value_complex column to be ~1000

ShruthiPitta:
Added migration task
https://issues.openmrs.org/browse/TRUNK-5140


ShruthiPitta:
In the pre condition we are just checking if the column exists or not. We are not checking the length of the column. Checked other existing similar migrations. They are doing the same.

PR #2241, opened 2017-09-12
TRUNK-5215: Replaced shouldIsValid test with multiple tests; added ch…

jerryb8:
Replaced shouldIsValid test with multiple tests; added character transposition test.

Added and replaced tests in VerhoeffIdentifierValidatorTest
Clarified Exception messaging in BaseHyphenatedIdentifierValidator

see https://issues.openmrs.org/browse/TRUNK-5215

### DIFF
--- a/api/src/main/java/org/openmrs/patient/impl/BaseHyphenatedIdentifierValidator.java
+++ b/api/src/main/java/org/openmrs/patient/impl/BaseHyphenatedIdentifierValidator.java
@@ -55,7 +55,14 @@ public abstract class BaseHyphenatedIdentifierValidator implements IdentifierVal
 	public boolean isValid(String identifier) throws UnallowedIdentifierException {
 		
 		if (identifier.indexOf("-") < 1) {
-			throw new UnallowedIdentifierException("Identifier must contain something besides the check digit.");
+			throw new UnallowedIdentifierException(
+					"Identifier \"" + identifier + "\" must contain a hyphen followed by a check digit character (A-J).");
+		}
+		if (identifier.endsWith("-")) {
+			throw new UnallowedIdentifierException(
+					"Identifier \""
+							+ identifier
+							+ "\" must not end with a hyphen - a check digit character (A-J) must follow.");
 		}
 		
 		String idWithoutCheckDigit = identifier.substring(0, identifier.indexOf("-"));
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/openmrs/patient/impl/VerhoeffIdentifierValidatorTest.java
```diff
@@ -33,87 +33,138 @@
...
+				}
+				// Shuffle the allowed identifier string around so that first character becomes last.
+				allowedIdentifier = allowedIdentifier.substring(1, allowedIdentifier.length()) + allowedIdentifier.charAt(0);
 			}
-			catch (Exception e) {}
-		}
-		
-		//Now test allowed identifiers that just have the wrong check digit.
-		for (int i = 0; i < allowedIdentifiers.length; i++) {
-			assertFalse(validator.isValid(allowedIdentifiers[i] + "-" + unusedCheckDigit));
```
dkayiwa: Did you intentionally get rid of these?

api/src/test/java/org/openmrs/patient/impl/VerhoeffIdentifierValidatorTest.java
```diff
@@ -33,87 +33,138 @@
...
-		//Make sure check digit can't be numeric
-		for (int j = 0; j < invalidIdentifiers.length; j++) {
-			try {
-				validator.isValid(allowedIdentifiers[j] + "-" + allowedIdentifiersCheckDigitsInt[j]);
-				fail("Identifier " + allowedIdentifiers[j] + " should have failed.");
+	}
+
+	@Test
+	public void checkDigit_shouldChangeWhenAdjacentCharsAreTransposed() {
+		VerhoeffIdentifierValidator v = new VerhoeffIdentifierValidator();
```
dkayiwa: I would rename v to validator.

api/src/main/java/org/openmrs/patient/impl/BaseHyphenatedIdentifierValidator.java
```diff
@@ -54,8 +54,15 @@ public String getValidIdentifier(String undecoratedIdentifier) throws UnallowedI
 	@Override
 	public boolean isValid(String identifier) throws UnallowedIdentifierException {
 		
-		if (identifier.indexOf("-") < 1) {
-			throw new UnallowedIdentifierException("Identifier must contain something besides the check digit.");
+		if (identifier.indexOf("-") <= 0) {
```
dkayiwa: Does <= 0 make it any better than < 1?

api/src/test/java/org/openmrs/patient/impl/VerhoeffIdentifierValidatorTest.java
jerryb8: This is covered in the test  isValid_shouldFailWithIncorrectlyCalculatedSuffixes()

api/src/test/java/org/openmrs/patient/impl/VerhoeffIdentifierValidatorTest.java
jerryb8: Agree, I will rename it.

api/src/main/java/org/openmrs/patient/impl/BaseHyphenatedIdentifierValidator.java
jerryb8: Sorry, I think I changed it to <0, then realised that it should fail on zero, so I added an "=" sign. I should have reinstated <1.

PR #2243, opened 2017-09-21
TRUNK-5027:  Add support for MIME types

FloBue:
Added support for saving and reading MIME types from filenames. However, the ConceptValidatorTest class only successfully runs when run as a single test, not with mvn clean install. Possible interference between the tests.

JIRA: https://issues.openmrs.org/browse/TRUNK-5027


FloBue:
I just took a look at the codacy remarks. Should I tackle them first, then do another pull request? Or should I wait for you to check the problem with the parallel tests?

dkayiwa:
You are free to address them as i look at the parallel tests.

api/src/test/java/org/openmrs/obs/MediaHandlerTest.java
```diff
@@ -12,39 +12,157 @@
...
+@PrepareForTest({ AbstractHandler.class, OpenmrsUtil.class, Context.class })
 
-    @Test
-    public void shouldSupportRawView() {
-        MediaHandler handler = new MediaHandler();
-
-        assertTrue(handler.supportsView(ComplexObsHandler.RAW_VIEW));
-    }
-
-    @Test
```
dkayiwa: Did you intentionally remove these?

api/src/test/java/org/openmrs/obs/ImageHandlerTest.java
```diff
@@ -12,39 +12,150 @@
...
-    @Test
-    public void shouldReturnSupportedViews() {
-        ImageHandler handler = new ImageHandler();
-        String[] actualViews = handler.getSupportedViews();
-        String[] expectedViews = { ComplexObsHandler.RAW_VIEW };
-
-        assertArrayEquals(actualViews, expectedViews);
-    }
-
-    @Test
```
dkayiwa: Did you intentionally remove these?

api/src/test/java/org/openmrs/obs/BinaryStreamHandlerTest.java
```diff
@@ -12,39 +12,152 @@
...
-
-    @Test
-    public void shouldReturnSupportedViews() {
-        BinaryStreamHandler handler = new BinaryStreamHandler();
-        String[] actualViews = handler.getSupportedViews();
-        String[] expectedViews = { ComplexObsHandler.RAW_VIEW };
-
-        assertArrayEquals(actualViews, expectedViews);
-    }
-
```
dkayiwa: Did you intentionally remove these?

api/src/test/java/org/openmrs/obs/BinaryDataHandlerTest.java
```diff
@@ -9,42 +9,149 @@
...
-        BinaryDataHandler handler = new BinaryDataHandler();
-        String[] actualViews = handler.getSupportedViews();
-        String[] expectedViews = { ComplexObsHandler.RAW_VIEW };
-
-        assertArrayEquals(actualViews, expectedViews);
-    }
-
-    @Test
-    public void shouldSupportRawView() {
-        BinaryDataHandler handler = new BinaryDataHandler();
```
dkayiwa: Did you intentionally remove these?

dkayiwa:
@FloBue three tests fail on my side. Is it the same with you?

Failed tests: 
  ImageHandlerTest.shouldGuessMimeTypeIfNotSpecified:154 null
  MediaHandlerTest.shouldGuessMimeTypeIfNotSpecified:162 null
  TextHandlerTest.shouldGuessMimeTypeIfNotSpecified:150 null

api/src/test/java/org/openmrs/obs/BinaryDataHandlerTest.java
FloBue: They aren't removed (all of the cases you commented as far as I can see). I'm not sure if it has to do with Eclipse's Auto Formatting (I took the OpenMRS Profile for it) or with me moving stuff. Just check the right side of the diff and you will see they should all still be there, just moved. I haven't changed them at all.

dkayiwa:
@FloBue can you do it without moving any of these tests?
That way i will see red only for code that you actually removed instead of just moved, and hence make it easier for me to figure out what is going on.

FloBue:
I'm kind of confused by your request. The things I've added make sense at the start of the test class, why would I add them after the existing tests?

FloBue:
Codacy checks out fine now. Let me know your answer to my previous reply. Have you found out anything about the null exceptions that don't occur on my side?

api/src/test/java/org/openmrs/obs/BinaryDataHandlerTest.java
dkayiwa: It takes more time to review such a pull request because one needs to differentiate between what was actually removed (red) and what was added back (green). How about if you just added your changes at the very end of the file?

FloBue: I do understand your point of course. However, adding the @Before method before the @Test methods for ordering reasons is pretty understandable I think. I can't imagine other developers add code to files "chronologically"? The same goes for the new variables I introduced to the test classes, including the @Mock variable for the administration service. Why would I add class variables below?

FloBue:
For some reason codacy now shows new complaints after it greenlighted this yesterday. I will take care of this today.

api/src/test/java/org/openmrs/obs/BinaryDataHandlerTest.java
dkayiwa: In that case, you can start by just adding them at the end such that the pull request does not show the other methods as moved. After it is merged, you can create a new pull request for simply formatting purposes.

FloBue: What about imports and other declarations that go before the class, do they cause the same problems? I can't move them down.

dkayiwa: imports and variable declarations are fine.

FloBue: I will do that and see if the diffs come out clearer, hopefully I can manage to do it later today. Meanwhile, are the null errors gone now? And could you take a look at my last comment concerning the codacy remarks?

dkayiwa:
@FloBue can you check if we already have that here? https://wiki.openmrs.org/display/docs/Java+Conventions
Free free to add it if missing.

FloBue:
I added a line to the wiki to make that clear. Somebody would have to add that rule to the codacy checks.

dkayiwa:
@FloBue thanks for adding it. Do you also wanna volunteer to update codacy? 😊

FloBue:
I have updated all tests and the diff looks cleaner now. I've also changed some details about the tests to make their success more reliable, since I used random bytes before, that could potentially cause errors. Let me know if there is anything else.

Sure, I've never done anything with codacy but if you can point me in some direction I'd be willing to do it!

FloBue:
I should mention that I will be on vacation from October 1st to October 15th, so I won't be able to reply or fix any remaining issues until then. The open issues I can think about are those in my last comment. I don't really know why definitely tested lines now show up as untested, maybe you've got an idea. If you can point me somewhere concerning codacy, I'd get to it when I get back.

FloBue:
@dkayiwa I am back now. Please take a look at my previous 2 comments (specifically about the unexpected change in test coverage after restructuring and where to modify codacy rules).

dkayiwa:
@FloBue welcome back! 😊 
Can you also include the ticket url as advised at? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

FloBue:
@dkayiwa Thanks, I changed the PR title and added the JIRA link to the top comment

FloBue:
@dkayiwa Hello Daniel, what about the status of this PR? I would like to get it done if possible. I've commented on everything that still fails (and why I think it's not a problem on my side) in my previous comments. But if you find anything I should correct, please let me know.

dkayiwa:
@FloBue when you say "everything that still fails", do you mean failing tests?

FloBue:
@dkayiwa No, not at all. I'm just talking about the checks here, codacy (we talked about that already, the caps constants) and the coverage/coveralls. I really don't understand the coverage decrease, since there was a coverage increase with my original PR. I restructured it as you asked, which didn't change anything in terms of functionality and tests. It now marks some lines as untested which are actually tested (I've checked it manually that they still get run). I currently assume that is the reason for the coverage decrease. Everything else should be fine.

api/src/main/java/org/openmrs/obs/handler/TextHandler.java
```diff
@@ -84,7 +84,12 @@ public Obs getObs(Obs obs, String view) {
 		}
 		
 		Assert.notNull(complexData, "Complex data must not be null");
-		complexData.setMimeType("text/plain");
+		
+		// Get the Mime Type and set it
+		String mimeType = getMimeTypeFromFile(file);
+		mimeType = mimeType != null ? mimeType : "application/octet-stream";
```
dkayiwa: Of the two "application/octet-stream" and "text/plain", which is more appropriate for this class?

api/src/main/java/org/openmrs/obs/handler/MediaHandler.java
```diff
@@ -61,7 +62,10 @@ public Obs getObs(Obs obs, String view) {
 				FileInputStream mediaStream = new FileInputStream(file);
 				ComplexData complexData = new ComplexData(originalFilename, mediaStream);
 				
-				complexData.setMimeType(OpenmrsUtil.getFileMimeType(file));
+				// Get the Mime Type and set it
+				String mimeType = getMimeTypeFromFile(file);
+				mimeType = mimeType != null ? mimeType : "application/octet-stream";
```
dkayiwa: If mimeType is null, shouldn't the default be like it was on line 64 before this change?

api/src/main/java/org/openmrs/obs/handler/ImageHandler.java
```diff
@@ -82,23 +78,10 @@ public Obs getObs(Obs obs, String view) {
...
-				}
-			}
-			catch (FileNotFoundException e) {
-				log.error("Image " + file.getAbsolutePath() + " was not found", e);
-			}
-			catch (IOException e) {
-				log.error("Trying to determine MIME type of " + file.getAbsolutePath(), e);
-			}
+			// Get the Mime Type and set it
+			String mimeType = getMimeTypeFromFile(file);
```
dkayiwa: If mimeType is null, shouldn't the default be like it was before this change?

api/src/main/java/org/openmrs/obs/handler/TextHandler.java
FloBue: You're right, I'll change this.

api/src/main/java/org/openmrs/obs/handler/MediaHandler.java
FloBue: I will replace it.

It will probably return the same thing, because if it can't read the type from the name, getMimeTypeFromFile() also probes the content type with Files.probeContentType(). I believe this also looks at the file extension. OpenmrsUtil.getFileMimeType() does that as well, but via a MimetypesFileTypeMap. However, from some stackoverflow posts it seems the behaviour of probeContentType() can be somehow buggy or confusing, so it might be better to do it as you suggested.

api/src/main/java/org/openmrs/obs/handler/ImageHandler.java
FloBue: I will replace this as well. It's the same thing as I explained above, I'm not sure if the old code is even able to return something more accurate than probeContentType(), but you're right, it is better to keep it just in case.

FloBue:
@dkayiwa I pushed a new version to address your comments.

api/src/test/java/org/openmrs/obs/ImageHandlerTest.java
```diff
@@ -47,4 +77,87 @@ public void shouldNotSupportOtherViews() {
...
+		
+		// Execute save
+		handler.saveObs(obs);
+		
+		// Get observation
+		Obs complexObs = handler.getObs(obs, "RAW_VIEW");
+		
+		String probedMimetype = complexObs.getComplexData().getMimeType();
+		
+		assertTrue(probedMimetype.equals(mimetype));
```
dkayiwa: When i run this test locally, probedMimeType equals application/octet-stream instead of image/png which the above line expects.

FloBue: That's interesting, for me it runs fine. That might be what those complaints I read about the inconsistent behaviour of probeContentType() across different OS was about. That's what I was talking about previously.

There would be 2 ways to handle this: 
1. I could let the old code for ImageHandler execute if an image is given application/octet-stream (I would do this for TextHandler analogously then) because its probing might be more reliable? I'd also adjust the test to accept the return mime type of the old code.
or
2 . Adjust the test to pass if either the image/png is recognized or application/octet-stream is returned because it could not be probed. 

Would you prefer one over the other?

dkayiwa: Having a test which expects inconsistent results, does not sound right.

FloBue: The problem is that probing (at least with Files.probeContentType()) is somehow system dependent and sometimes shows weird behaviour, at least from what I read. Probing is only a last resort if the mime type isn't in the filename, as it should be with all future complex obs. That is why I suggested option 1 in my previous post, in hopes that the old code returns more consistent types than Java's own probeContentType(). It's hard for me to evaluate though, with just 1 system here to test.

Another way to have consistency is to just determine the mime type (if it is not in the name) by looking at the file extension, which should be more consistent, but is incapable of guessing it from the file contents. That would be option 3, I think there is a method for that as well. That's why I ask what would be best for the project.

Thinking about it, having to guess the type from the file contents is probably not very important  and option 3 (just guessing it from the file extension) might be the best way.

dkayiwa: Did you see any disadvantage for using file extension?

FloBue: The only disadvantage is if no file extension exists either, we have to use a default type. Probing could try to find it out by reading the file itself instead of just the name + extension. But as you see, that's still kind of problematic.

I'm currently going for the file extension approach. The best way I have found is to modify META-INF/mime.types to contain more extensions and types, and then use MimetypesFileTypeMap. I will try this now and comment if I make any progress that I can push.

FloBue:
@dkayiwa I've changed the tests to only expect 1 mime type, which works now because the MimetypesFileTypeMap is used along with an expanded META-INF/mime.types list now. The file content is not probed anymore, I hope this eliminates any inconsistent behaviour across systems.

FloBue:
@dkayiwa I think travis-ci fails because of the same parallel test problem with maven we had before. Building it with DskipTests=true works fine and if I run the tests, the same test that has nothing to do with the changes, fails (a ConceptValidator test). I'm not sure I can do anything about this.

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -56,49 +59,95 @@ public AbstractHandler() {
 	 * and the file name is determined off the current obs.getComplexData().getTitle().
 	 * 
 	 * @param obs the Obs with a non-null complex data on it
-	 * @return File that the complex data should be written to
+	 * @return file that the complex data should be written to
 	 */
 	public File getOutputFileToWrite(Obs obs) throws IOException {
-		// Get the title and remove the extension.
-		String t = obs.getComplexData().getTitle();
```
dkayiwa: Could you remind me why we removed these two lines?

FloBue: I just renamed the variable to title, it's some lines down, to make it easier to read. At least that's how it looks like, it's been some time.

FloBue:
@dkayiwa So I tried to test it today with the legay UI as suggested by you, but it seems the deployment doesn't work as I thought, it's still the old code that executes (and I don't know where it's coming from). How would you run the packaged openmrs.war together with the legacy ui? With mvn jetty:run I don't have any UI.

dkayiwa:
@FloBue look at the logs for the location of your runtime properties file. In there you will see the modules folder in which you need to drop the legacyui module.

FloBue:
@dkayiwa Thanks. There was no path in the runtime properties file, but I found in the log that he seems to look in openmrs-core/webapp/src/main/webapp/WEB-INF/bundledModules . I created the bundledModules folder, dropped the legacyui.omod there and it seems it recognized it. However, I still get this error: HTTP ERROR 500
Problem accessing /openmrs/login.htm. Reason:
    Server Error
Caused by:
org.apache.jasper.JasperException: /WEB-INF/view/module/legacyui/login.jsp (line: 1, column: 1) null

FloBue:
@dkayiwa I've checked everything with the hosted .war file as you suggested and it seems to work fine. I had to adjust another JUnit test (ObsServiceTest.java) that had the filename for the complex obs file hardcoded, so that test does not fail anymore with the new naming. Let me know if there is anything else I should do before we can finish this ticket.

PS. The following test fails when I do mvn clean install, but it fails on the openmrs-core master for me as well. Do you know why that is?:
ConceptValidatorTest.validate_shouldPassIfDifferentConceptsHaveTheSameShortNames:479 expected:<1> but was:<0>

dkayiwa:
@FloBue aren't the rest of the handlers working well apart from only BinaryDataHandler and BinaryStreamHandler?

FloBue:
@dkayiwa My JUnit Tests and the manual testing on a deployed OpenMRS Server all worked out fine, for every handler. I tested saving ComplexObs for each handler and the files where correctly saved with the new mimetype information. I think I don't understand what you mean, could you elaborate? Did you find another error?

dkayiwa:
@FloBue what i mean is that wasn't the reported problem only in BinaryDataHandler and BinaryStreamHandler?
Meaning that the rest did not have this bug and hence needed not any fix. 😊

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -56,49 +59,102 @@ public AbstractHandler() {
...
+		if (file == null) {
+			return null;
+		}
+		
+		String filenameWithoutExtension = FilenameUtils.removeExtension(file.getName());
+		
+		String mimeType = null;
+		// Try to parse MIME type from filename
+		if (filenameWithoutExtension.contains("-slash-")) {
+			String[] fileParams = filenameWithoutExtension.split("_");
```
dkayiwa: Do you have an example where MimetypesFileTypeMap in the else clause will not be enough to get the mime type? In other words, if we dropped the adding of "-slash-mimetype" to the file name, shouldn't MimetypesFileTypeMap.getContentType(file.getName()) be enough for do what we need?

FloBue: This is pretty much what I've meant also when we talked on IRC, when I said we can simply replace some lines instead of what the ticket suggested.

It's a bit of a complicated story - At first, tmarzeion suggested we probe the mimetype from the file content even if no file extension exists. The problem with this is, as you have seen, that it is inconsistent across different systems. That is why I've replaced it with the MimetypesFileTypeMap in my code as well, which only considers the file extension. But that is something, that was previously insufficient in OpenMRS, simply because META-INF/mime.types, which is where the MimetypesFileTypeMap gets its mapping from, didn't contain many mimetypes. That is why I expanded META-INF/mime.types. You can take a look at the old mime.types and you will see that it's missing even some of the most basic mime types.

If you are against saving the mimetypes in filenames, that is also what I would suggest: set the mimetype via the MimetypesFileTypeMap when the obs is created, and set it the same way again when it is loaded from harddisk. This is still an improvement and more flexible than before. I think this is what you mean?

dkayiwa: Yes that is exactly what i mean! 👍

FloBue: I will go through the code and check what changes are necessary or have to be reverted.

dkayiwa: Will merge as soon as you are done with that. Thanks again for the great work and patience! 👍

FloBue:
@dkayiwa Here is my new version. I haven't had the time to deploy it on a server to test it manually yet, but I will try that tomorrow. I was a bit held up with a merge conflict on rebase. The unit tests work though. 

In case you are wondering why I removed the mimeType == null checks in the handlers, and replaced some of them by mimeType.equals("application/octet-stream"):

By its documentation, OpenmrsUtil.getFileMimeType() should never return null but only application/octet-stream" if no other corresponding mimetype was found in the MimetypesFileTypeMap.

If this happens, I fall back on how the old code handled it.

api/src/main/java/org/openmrs/obs/handler/ImageHandler.java
```diff
@@ -81,24 +82,31 @@ public Obs getObs(Obs obs, String view) {
...
-				if (imgReader.hasNext()) {
-					complexData.setMimeType("image/" + imgReader.next().getFormatName().toLowerCase());
-				} else {
-					log.warn("MIME type of " + file.getAbsolutePath() + " is not known");
+			// Get the Mime Type and set it
+			String mimeType = OpenmrsUtil.getFileMimeType(file);
+			
+			// Image MIME type if type couldn't be found		
+			if(mimeType.equals("application/octet-stream")) {
+				try {
```
dkayiwa: Do we have to wrap this in an if statement? Why not just leave it as before, without the conditional if statement?

FloBue: It depends on what you want. I thought all mimetypes would now be uniformly determined by OpenmrsUtil.getFileMimeType(file) and I only fall back on the old code if that does not determine anything better than application/octet-stream. Do you want me to change it?

By the way: I found a problem in the current OpenMRS core, where AbstractHandler.getOutputFileToWrite() creates weird file names when the file has no extension. For example, the file "Test" becomes "Test.Test" when saved as a complex obs. I will correct this as well and push it in a few minutes.

dkayiwa: I would be as surgical as possible, just in case. 😊 

As for the found problem, nice catch 👍

FloBue:
@dkayiwa Okay, I pushed a new version. ImageHandler now uses the old code and just if this does not find any mimetype, I determine it via getFileMimeType(). I also fixed the AbstractHandler file naming problem I mentioned and tested everything on a platform server with the legacy UI. Let me know what you think.

dkayiwa:
@FloBue i like the changes! 👍 
And i love the fact that you have more tests than the actual code to fix the reported problem. If we could all do like you did for this pull request, our codebase would be much better than what it currently is. 😊

dkayiwa:
Thanks again for being awesomely patient and not giving up on this, despite the lengthy back and forth. 😊

FloBue:
@dkayiwa Thanks for your positive feedback, means a lot to me. I kind of felt bad about it myself for not being able to foresee and resolve problems better and faster, so this is nice to hear :).

PR #2245, opened 2017-09-22
TRUNK-4917 - Add Creatable and Changeable interfaces

wluyima:
I added the new Creatable and Changeable interfaces, deprecated the changedBy and dateChanged setters and getters in the relevant super classes

This pull request is just a partial commit so that I can get some feedback from other devs before I can go ahead to do the major refactoring of updating the domain objects to extend BaseChangeableOpenmrsData or BaseChangeableOpenmrsMetadata

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-4917


api/src/main/java/org/openmrs/Auditable.java
```diff
@@ -19,49 +17,9 @@
...
-	 * @return Date - the date the object was last changed
-	 */
-	public Date getDateChanged();
-	
-	/**
-	 * @param dateChanged - the date the object was last changed
-	 */
-	public void setDateChanged(Date dateChanged);
-	
+public interface Auditable extends Creatable, Changeable {
```
wluyima: Should we actually get rid of this interface? Or should we keep it around?

api/src/main/java/org/openmrs/BaseChangeableOpenmrsData.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.Date;
+
+/**
+ * Base superclass for all mutable OpenmrsData.
+ * 
+ * @since 2.2
+ */
+public abstract class BaseChangeableOpenmrsData extends BaseOpenmrsData {
+
```
wluyima: I added these overrides so that these setters/getters don't appear deprecated as an effect from the superclass since eventually the methods will still exist on this class.

wluyima:
@djazayeri and @bmamlin does this seem in line with option D [on this design call](https://notes.openmrs.org/2016-08-31-Design-Forum)?

api/src/main/java/org/openmrs/Auditable.java
```diff
@@ -19,49 +17,9 @@
  * this interface.
  * 
  * @since 1.5
- * @see OpenmrsData
- * @see OpenmrsMetadata
+ * @see MutableOpenmrsData
+ * @see MutableOpenmrsMetadata
```
bmamlin: Do `MutableOpenmrsData` and `MutableOpenmrsMetadata` classes exist outside of this PR? Or should these be references `Changeable`, `BaseChangeableOpenmrsData`, and `BaseChangeableOpenmrsMetadata` classes?

api/src/main/java/org/openmrs/Auditable.java
bmamlin: Unless we can scan core code and popular modules and ensure nobody is referencing it, I would keep it around for now.

api/src/main/java/org/openmrs/BaseChangeableOpenmrsData.java
bmamlin: Make sense.

api/src/main/java/org/openmrs/BaseOpenmrsData.java
```diff
@@ -70,72 +70,79 @@ public BaseOpenmrsData() {
...
+	 * @see org.openmrs.OpenmrsData#setDateCreated(java.util.Date)
 	 */
 	@Override
 	public void setDateCreated(Date dateCreated) {
 		this.dateCreated = dateCreated;
 	}
 	
 	/**
-	 * @see org.openmrs.Auditable#getChangedBy()
+	 * @deprecated
```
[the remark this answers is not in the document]
bmamlin: I assume @deprecated in javadoc would ideally include "as of version x.y".

api/src/main/java/org/openmrs/Auditable.java
wluyima: Definitely it's referenced in core, but in the follow up commit I was going to replace its usage. I wasn't going to remove it in this version, we would mark it as deprecated now and eventually remove it may be 2 versions later.

api/src/main/java/org/openmrs/Auditable.java
wluyima: Good catch! I had them and removed them and was should have updated these javadocs

api/src/main/java/org/openmrs/BaseOpenmrsData.java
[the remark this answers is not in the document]
wluyima: Yes, I will add it

wluyima:
In core we are going to change subclasses of `BaseOpenmrsData` to extend `BaseChangeableOpenmrsData` instead, likewise subclasses of `BaseOpenmrsMetadata` will extend `BaseChangeableOpenmrsMetadata`. The same will apply for all modules, although it doesn't have to be immediate but will have to once we remove the deprecated code.

wluyima:
@bmamlin I did address the review comments, I think we can merge this along with the change of all subclasses of BaseOpenmrsData and BaseOpenmrsMetadata to extend the new base classes.

wluyima:
@dkayiwa , @rkorytkowski do you mind merging this PR after you review it? I need to do the next step of changing domain objects to extend BaseChangeableOpenmrsData or BaseChangeableOpenmrsMetadata instead.

dkayiwa:
@wluyima will it take you much time to squash these commits into one? I find it easier to review one commit than multiple, especially when they are followup commits to address earlier review comments.

wluyima:
@dkayiwa you shouldn't be looking at individual commits, the PR shows them as one

wluyima:
Aren't you viewing all the changes as one in **Files changed** tab on the PR?

wluyima:
Still waiting on you @dkayiwa

api/src/main/java/org/openmrs/BaseOpenmrsMetadata.java
```diff
@@ -110,72 +110,79 @@ public void setDescription(String description) {
...
 	 */
 	@Override
 	public void setDateCreated(Date dateCreated) {
 		this.dateCreated = dateCreated;
 	}
 	
 	/**
-	 * @see org.openmrs.Auditable#getChangedBy()
+	 * @see org.openmrs.OpenmrsMetadata#getChangedBy()
+	 * @deprecated as of version 2.2
```
dkayiwa: Our convention has been to include what they should alternatively use. https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-Deprecation

wluyima: The methods are actually going away with no replacement, I think I just have to add more details as to why it was necessary to remove the method.

wluyima: And to mention that we introduced a new interface that one would need to implement to in case they want to have a mutable class

wluyima: Actually the method javadocs have the `@see` the superclass where information as to why it's deprecated is included

api/src/main/java/org/openmrs/BaseChangeableOpenmrsData.java
ravirajmane: public interface OpenmrsData extends OpenmrsObject, Auditable, Voidable {}
Which more methods will be added in OpenmrsData interface?
Or OpenmrsData interface will implement the methods of Creatable interface which contain getCreator, getDateCreated methods and Changeable interface contain getChangedBy, getDateChanged methods?

wluyima: OpenmrdsData for now will stay the way it is, in future it will implement Creatable instead of Auditable because Data should be immutable by default.

PR #2287, opened 2017-11-15
Trunk 4954-Checking for patient_id of both obs and encounter when saving obs

dilekamadushan:
'TRUNK-4954' - ObsService.saveObs should throw a validation exception if the obs and its encounter point to different persons

I modified the code to check whether the patient_id of obs and encounter are the same and unless throw an exception.

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
@@ -113,8 +113,6 @@ public class ObsServiceImpl extends BaseOpenmrsService implements ObsService {
 			else{
 				throw new APIException("Obs.error.PatientId.mismatch", (Object[]) null);
 			}
-
-
 		}
 	}
 
--- a/api/src/main/java/org/openmrs/Auditable.java
+++ b/api/src/main/java/org/openmrs/Auditable.java
@@ -19,5 +19,5 @@ package org.openmrs;
  * @since 1.5
  */
 public interface Auditable extends Creatable, Changeable {
-	
+
 }
--- a/api/src/main/java/org/openmrs/scheduler/Schedule.java
+++ b/api/src/main/java/org/openmrs/scheduler/Schedule.java
@@ -189,7 +189,7 @@ public class Schedule {
 	public void setStartTime(Date startTime) {
 		this.startTime = startTime;
 	}
-
+	
 	/**
 	 * Gets the number of seconds until task is executed again.
 	 * 


dkayiwa:
Can you squash the commits into one as advised at? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

api/src/test/java/org/openmrs/api/ObsServiceTest.java
```diff
@@ -1973,4 +1974,29 @@ public void shouldNotAutomaticallySetStatusWhenManuallyCopyingAnObs() throws Exc
...
+
+		Obs o = new Obs();
+		o.setConcept(cs.getConcept(3));
+		o.setDateCreated(new Date());
+		o.setCreator(Context.getAuthenticatedUser());
+		o.setLocation(new Location(1));
+		o.setObsDatetime(new Date());
+		o.setValueText("Testing");
+		o.setPerson(ps.getPerson(2)); // Setting this person (2) as different from that of encounter person (7)
+		o.setEncounter(es.getEncounter(2));
```
wluyima: Can you please a sanity check that ensures that encounter.person is different from obs.person

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -109,8 +105,14 @@ public Obs saveObs(Obs obs, String changeMessage) throws APIException {
 			setPersonFromEncounter(obs);
 			return saveObsNotDirty(obs, changeMessage);
 		} else {
-			setPersonFromEncounter(obs);
-			return saveExistingObs(obs,changeMessage);
+
+			if(Objects.equals(obs.getPersonId(), obs.getEncounter().getId())){
```
wluyima: The code should be in ObsValidator and you should be comparing person objects and not their personIds

api/src/test/java/org/openmrs/api/ObsServiceTest.java
```diff
@@ -1973,4 +1974,29 @@ public void shouldNotAutomaticallySetStatusWhenManuallyCopyingAnObs() throws Exc
 		assertThat(existing.getVoided(), is(true));
 		assertThat(newObs.getStatus(), is(Obs.Status.FINAL));
 	}
+
```
wluyima: Once you move the logic to ObsValidator, you need to move the test to ObsValidatorTest

api/src/test/java/org/openmrs/api/ObsServiceTest.java
dilekamadushan: Ok I will work on it..Should I change all the attributes of one person object from other ?

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
dilekamadushan: Ok I will do the necessary changes.  thanks

api/src/test/java/org/openmrs/api/ObsServiceTest.java
dilekamadushan: Ok... Thanks

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
dilekamadushan: @wluyima I wrote the code to check for person objects in ObsValidator and placed the test in ObsValidatorTest .Should I call the ObsValidator's method in ObsServiceImpl ?

wluyima: Why? Move all the code to the validator, there should no changed in ObsServiceImpl

api/src/test/java/org/openmrs/api/ObsServiceTest.java
wluyima: Not sure what you mean

dkayiwa:
@dilekamadushan have you seen the merge conflicts on this pull request? 😊

PR #2320, opened 2017-11-30
Added an private constructor

frosch000:
Created a private constructor according to this issue: https://issues.openmrs.org/plugins/servlet/mobile#issue/GCI-202
This was a Task for the Code-In Contest

### DIFF
--- a/web/src/main/java/org/openmrs/web/WebConstants.java
+++ b/web/src/main/java/org/openmrs/web/WebConstants.java
@@ -11,6 +11,10 @@ package org.openmrs.web;
 
 public class WebConstants {
 	
+	private WebConstants() {
+		
+        }
+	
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/openmrs/web/WebConstants.java
```diff
@@ -11,6 +11,10 @@
 
 public class WebConstants {
 	
+	private WebConstants() {
+
```
gayanW: I think we don't need the blank line here.

frosch000:
I will correct this and  make a new pull request

gayanW:
BTW you don't need mention _"This was a Task for the Code-In Contest"._

Just FYI: https://chris.beams.io/posts/git-commit/

PR #2366, opened 2017-12-04
[GCI-142] Check bad code practices through Findbugs or Sonar and fix those issues.

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/GCI-142

What I have done:

- Fixed naming conventions and syntax
- Changed all Loggers to private final static
- Optimised imports
- Changed StreamAPI Calls and Lambdas to support Java 8 features
- Added diamond syntax in generic collections
- Fixed few bugs with not-closed try-catch resources
- Replaced Vector class occurencies
- Remove redundant modifiers (public/static) from Java interfaces
- Simplified "if" statements
- Replaced iteration operations with bulk (e.g. addAll instead of foreach)
- Removed unused Loggers

PR #2385, opened 2017-12-05
[GCI-142] Simplified "if" statements

BartlomiejRasztabiga:
Related PR: #2366

I've simplified few "if" statements and deleted unnecessary "return" statements

### DIFF
--- a/api/src/main/java/org/openmrs/util/Format.java
+++ b/api/src/main/java/org/openmrs/util/Format.java
@@ -42,7 +42,7 @@ public class Format {
 	}
 	
 	public static String format(double d) {
-		return "" + ((d == (int) d) ? (int) d : d);
+		return "" + (d);
 	}
 	
 	public static String format(Double d) {
--- a/api/src/main/java/org/openmrs/util/LocaleUtility.java
+++ b/api/src/main/java/org/openmrs/util/LocaleUtility.java
@@ -105,14 +105,8 @@ public class LocaleUtility implements GlobalPropertyListener {
 	 * @should confirm matching language as compatible
 	 */
 	public static boolean areCompatible(Locale lhs, Locale rhs) {
-		if (lhs.equals(rhs)) {
-			return true;
-		} else if ((("".equals(lhs.getCountry())) || ("".equals(rhs.getCountry())))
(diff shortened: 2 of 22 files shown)


api/src/main/java/org/openmrs/util/LocaleUtility.java
```diff
@@ -105,14 +105,8 @@ public static Locale getDefaultLocale() {
...
 	public static boolean areCompatible(Locale lhs, Locale rhs) {
-		if (lhs.equals(rhs)) {
-			return true;
-		} else if ((("".equals(lhs.getCountry())) || ("".equals(rhs.getCountry())))
-		        && lhs.getLanguage().equals(rhs.getLanguage())) {
-			// no country specified, so language match is good enough
-			return true;
-		}
-		return false;
+		return lhs.equals(rhs) || (("".equals(lhs.getCountry())) || ("".equals(rhs.getCountry()))) && lhs.getLanguage()
```
dkayiwa: Shouldn't the right hand side of the first || be enclosed in ()?

api/src/main/java/org/openmrs/util/Format.java
```diff
@@ -42,7 +42,7 @@ public static String formatPercentage(Number pct) {
 	}
 	
 	public static String format(double d) {
-		return "" + ((d == (int) d) ? (int) d : d);
```
dkayiwa: Are you sure this does not change the return value?

BartlomiejRasztabiga: Yes, look. If d is integer type, then we return (int) d. Else we return just d. So it can be simplified to return d :) I can write tests so you can be sure.

api/src/main/java/org/openmrs/util/LocaleUtility.java
BartlomiejRasztabiga: It was simplified automatically using SonarLint plugin so I hope it's good.

BartlomiejRasztabiga: I've checked it once again and I believe it is simplified properly

dkayiwa:

PR #2391, opened 2017-12-06
[GCI-183]:  Introduce a new variable instead of reusing the parameter "filename".

Troy19:
Link to the issue: https://issues.openmrs.org/browse/GCI-183

### DIFF
--- a/web/src/main/java/org/openmrs/web/filter/util/CustomResourceLoader.java
+++ b/web/src/main/java/org/openmrs/web/filter/util/CustomResourceLoader.java
@@ -91,11 +91,12 @@ public class CustomResourceLoader {
 	private Locale parseLocaleFrom(String filename, String basename) {
 		Locale result = null;
 		
+		String tmp_filename=filename;
 		if (filename.startsWith(basename)) {
-			filename = filename.substring(basename.length());
+			tmp_filename = filename.substring(basename.length());
 		}
 		
-		String localespec = filename.substring(0, filename.indexOf('.'));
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/openmrs/web/filter/util/CustomResourceLoader.java
```diff
@@ -91,11 +91,12 @@ public static CustomResourceLoader getInstance(HttpServletRequest httpRequest) {
 	private Locale parseLocaleFrom(String filename, String basename) {
 		Locale result = null;
 		
+		String tmp_filename=filename;
```
dkayiwa: Your variable name does not match with our naming conventions https://wiki.openmrs.org/display/docs/Java+Conventions

You would also need to put some space before and after the assignment operator "="

collinsnji:
Hey @Troy19 
Can you please squash your commits together so we get a clean, reversible git history? :)

web/src/main/java/org/openmrs/web/filter/util/CustomResourceLoader.java
```diff
@@ -91,11 +91,12 @@ public static CustomResourceLoader getInstance(HttpServletRequest httpRequest) {
 	private Locale parseLocaleFrom(String filename, String basename) {
 		Locale result = null;
 		
+		String tempFilename=filename;
```
collinsnji: Please add some space before and after the assignment operator `"="`

PR #2398, opened 2017-12-06
[GCI-142] Removed redundant value unboxing

BartlomiejRasztabiga:
Related PR: #2366

I've removed redundant value unboxing, like: valueBoolean.booleanValue(), because valueBoolean is already type of Boolean. Reduces complexity

### DIFF
--- a/api/src/main/java/org/openmrs/Form.java
+++ b/api/src/main/java/org/openmrs/Form.java
@@ -147,8 +147,8 @@ public class Form extends BaseChangeableOpenmrsMetadata {
 				
 				for (FormField ff : fieldSet) {
 					if (ff.getFieldNumber() != null) {
-						if (ff.getFieldNumber().intValue() < fieldNum || fieldNum == 0) {
-							fieldNum = ff.getFieldNumber().intValue();
+						if (ff.getFieldNumber() < fieldNum || fieldNum == 0) {
+							fieldNum = ff.getFieldNumber();
 							next = ff;
 						}
 					} else {
--- a/api/src/test/java/org/openmrs/util/DoubleRangeTest.java
+++ b/api/src/test/java/org/openmrs/util/DoubleRangeTest.java
@@ -49,60 +49,60 @@ public class DoubleRangeTest {
 	@Test
 	public void DoubleRange_shouldReturnInfiniteLowAndHighIfCalledWithNullParameters() {
 		DoubleRange dr = new DoubleRange(null, null);
-		assertEquals(Double.POSITIVE_INFINITY, dr.getHigh().doubleValue(), 0);
-		assertEquals(Double.NEGATIVE_INFINITY, dr.getLow().doubleValue(), 0);
+		assertEquals(Double.POSITIVE_INFINITY, dr.getHigh(), 0);
+		assertEquals(Double.NEGATIVE_INFINITY, dr.getLow(), 0);
 	}
 	
 	@Test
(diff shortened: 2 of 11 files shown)


BartlomiejRasztabiga:
@dkayiwa 
Could you set my task to done? So I could work on another and you would review the rest later :)

dkayiwa:
Sure. Just done the needful! 😊

BartlomiejRasztabiga:
Thanks a lot! I will of course do more of refactoring work after you merge all my PR's :)

api/src/main/java/org/openmrs/Form.java
```diff
@@ -148,8 +148,8 @@ public void setEncounterType(EncounterType encounterType) {
 				
 				for (FormField ff : fieldSet) {
 					if (ff.getFieldNumber() != null) {
-						if (ff.getFieldNumber().intValue() < fieldNum || fieldNum == 0) {
```
dkayiwa: Are you sure the object comparison is the same as primitive value comparison?

BartlomiejRasztabiga: It is "<" operator, so int and Integer comparisons are the same. It is different just with "==" operator.

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -327,7 +327,7 @@ private String getContextPath(ServletContext servletContext) {
 		}
 		
 		// trim off initial slash if it exists
-		if (contextPath.contains("/")) {
```
dkayiwa: Do you mind explaining the reason behind replacing contains with indexOf?

BartlomiejRasztabiga: It's just old PR. I will update this with master :)

BartlomiejRasztabiga: @dkayiwa Fixed

api/src/test/java/org/openmrs/util/DoubleRangeTest.java
```diff
@@ -49,60 +49,60 @@ public void DoubleRange_shouldReturnNullLowAndHighIfAccessorsAreNotCalled() {
...
 		dr.setLow(null);
-		assertEquals(Double.NEGATIVE_INFINITY, dr.getLow().doubleValue(), 0);
+		assertEquals(Double.NEGATIVE_INFINITY, dr.getLow(), 0);
 	}
 	
 	@Test
 	public void setLow_shouldCauseLowToHaveTheSetValue() {
 		DoubleRange dr = new DoubleRange(null, null);
 		dr.setLow(8.0);
-		assertEquals(8.0, dr.getLow().doubleValue(), 0);
```
dkayiwa: What value does changing from primitive and primitive data type comparison to Object and primitive data type comparison add?

BartlomiejRasztabiga: This will call the same method assertEquals(double, double, double) because Java will unbox this value automatically. So my PR just removes redundant code

dkayiwa: Was this also a recommendation from sonar?

BartlomiejRasztabiga: This actually was recomendation of built-in IntelliJ IDEA Inspector 😊

dkayiwa: Oh i see!!! 😊

PR #2399, opened 2017-12-06
[GCI-142] Create List of one element by Arrays.singletonList()

BartlomiejRasztabiga:
Related PR: #2366

I've replaced Arrays.asList() (with one element) with .singletonList()

Using Arrays.asList() with one argument creates new List with allocated memory for 10! elements. Instead of this, .singletonList() allocates memory just for one element. So it saves a lot of memory :)


BartlomiejRasztabiga:
Travis has failed here, but that's his issue. If you can restart him I would be grateful.

dkayiwa:

PR #2404, opened 2017-12-06
GCI-142 Collapse identical catch blocks

BartlomiejRasztabiga:
Related PR: #2366

I've collapsed identicall "catch" blocks because there is no sense in writing the same thing twice :)


dkayiwa:

PR #2407, opened 2017-12-06
[GCI-142] Lambdas and Method references instead of anonymous classes

BartlomiejRasztabiga:
Related PR: #2366

It is cleaner way to deal with anonymous classes since JDK8 so why not use it? :)

### DIFF
--- a/api/src/main/java/org/openmrs/Concept.java
+++ b/api/src/main/java/org/openmrs/Concept.java
@@ -886,7 +886,7 @@ public class Concept extends BaseOpenmrsObject implements Auditable, Retireable,
 			}
 		} else {
 			shortNames = getNames().stream()
-							.filter(n -> n.isShort())
+							.filter(ConceptName::isShort)
 							.collect(Collectors.toList());
 		}
 		return shortNames;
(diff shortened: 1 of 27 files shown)


BartlomiejRasztabiga:
@dkayiwa Can you look at these old PR's that are not yet merged? :)
#2407 
#2398 
#2385 
#2410

dkayiwa:
Sure let me do so. :)

api/src/main/java/org/openmrs/Concept.java
```diff
@@ -886,7 +886,7 @@ public ConceptName getShortNameInLocale(Locale locale) {
 			}
 		} else {
 			shortNames = getNames().stream()
-							.filter(n -> n.isShort())
+							.filter(ConceptName::isShort)
```
dkayiwa: How does this change make the code better?

BartlomiejRasztabiga: Method references are believed to be more readable, because they remove redundant "n ->" :)

dkayiwa:

PR #2411, opened 2017-12-07
[GCI-142] Replace code with removeIf()

BartlomiejRasztabiga:

RemoveIf() is another great method from Stream API that deletes item if some condition is true :)


dkayiwa:

PR #2414, opened 2017-12-08
[GCI-142] Removed class fields shadowing

tomekdo3:
### Description of what I changed
Removed class fields shadowing

### Issue I worked on

https://issues.openmrs.org/browse/GCI-142

### DIFF
--- a/api/src/main/java/org/openmrs/Concept.java
+++ b/api/src/main/java/org/openmrs/Concept.java
@@ -1600,8 +1600,8 @@ public class Concept extends BaseOpenmrsObject implements Auditable, Retireable,
 		//sets have moved to the bottom and hence need to be reassigned
 		//higher sort weights than the non retired ones
 		double weight = 990.0;
-		for (ConceptSet set : sortedConceptSets) {
-			set.setSortWeight(weight += 10.0);
+		for (ConceptSet conceptSet : sortedConceptSets) {
+			conceptSet.setSortWeight(weight += 10.0);
 		}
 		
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/openmrs/Concept.java
```diff
@@ -1600,8 +1600,8 @@ public void addSetMember(Concept setMember, int index) {
 		//sets have moved to the bottom and hence need to be reassigned
 		//higher sort weights than the non retired ones
 		double weight = 990.0;
-		for (ConceptSet set : sortedConceptSets) {
-			set.setSortWeight(weight += 10.0);
+		for (ConceptSet conceptSet : sortedConceptSets) {
```
dkayiwa: Do we have a class field named conceptSet?

tomekdo3: No we don't have it as class field, only as local

dkayiwa: Do we have a class field named set?

tomekdo3: We do have one. It's a Boolean

dkayiwa: Can you point me to it with a github link?

tomekdo3: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/Concept.java#LN98
line 98

dkayiwa:
Thanks @tomekdo3 😊

PR #2420, opened 2017-12-08
[TRUNK-5241] Handle null pointer dereferences

BartlomiejRasztabiga:

I've replaced every dangerous call to nullable variable with Objects.requireNonNull() so it is guaranteed to be non-null.


BartlomiejRasztabiga:
New PR: https://github.com/openmrs/openmrs-core/pull/2440

PR #2431, opened 2017-12-09
[TRUNK-5240] Use isEmpty() than size() == 0

BartlomiejRasztabiga:

It's cleaner to call isEmpty() on a collection rather than checking if size() == 0.


BartlomiejRasztabiga:
New PR: https://github.com/openmrs/openmrs-core/pull/2448

PR #2432, opened 2017-12-09
[TRUNK-5240] Replace Long constructor with Long.valueOf call

BartlomiejRasztabiga:

I've replaced new Long() call with Long.valueOf() because it is more effcient


BartlomiejRasztabiga:
New PR: https://github.com/openmrs/openmrs-core/pull/2449

PR #2437, opened 2017-12-09
[GCI-236] Make sure file.listFiles() is not null before looping

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/GCI-236

### DIFF
--- a/web/src/main/java/org/openmrs/web/Listener.java
+++ b/web/src/main/java/org/openmrs/web/Listener.java
@@ -426,14 +426,16 @@ public final class Listener extends ContextLoader implements ServletContextListe
 					webAppLog.debug("Overriding file: " + absolutePath);
 					webAppLog.debug("Overriding file with: " + userOverridePath);
 					if (file.isDirectory()) {
-						for (File f : file.listFiles()) {
-							userOverridePath = f.getAbsolutePath();
-							if (!f.getName().startsWith(".")) {
-								String tmpAbsolutePath = absolutePath + "/" + f.getName();
-								if (!copyFile(userOverridePath, tmpAbsolutePath)) {
-									webAppLog.warn("Unable to copy file in folder defined by runtime property: " + prop);
-									webAppLog.warn("Your source directory (or a file in it) '" + userOverridePath
-									        + " cannot be loaded or destination '" + tmpAbsolutePath + "' cannot be found");
+						if (file.listFiles() != null) {
+							for (File f : file.listFiles()) {
+								userOverridePath = f.getAbsolutePath();
+								if (!f.getName().startsWith(".")) {
+									String tmpAbsolutePath = absolutePath + "/" + f.getName();
+									if (!copyFile(userOverridePath, tmpAbsolutePath)) {
+										webAppLog.warn("Unable to copy file in folder defined by runtime property: " + prop);
+										webAppLog.warn("Your source directory (or a file in it) '" + userOverridePath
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -517,14 +520,16 @@ public static void loadBundledModules(ServletContext servletContext) {
...
-		for (File f : folder.listFiles()) {
-			if (!f.getName().startsWith(".")) { // ignore .svn folder and the like
-				try {
-					Module mod = ModuleFactory.loadModule(f);
-					log.debug("Loaded bundled module: " + mod + " successfully");
-				}
-				catch (Exception e) {
-					log.warn("Error while trying to load bundled module " + f.getName() + "", e);
+		if (folder.listFiles() != null) {
+			for (File f : Objects.requireNonNull(folder.listFiles())) {
```
dkayiwa: Do you need Objects.requireNonNull when you are already sure that folder.listFiles() is not null?

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -426,14 +427,16 @@ private void copyCustomizationIntoWebapp(ServletContext servletContext, Properti
...
-						for (File f : file.listFiles()) {
-							userOverridePath = f.getAbsolutePath();
-							if (!f.getName().startsWith(".")) {
-								String tmpAbsolutePath = absolutePath + "/" + f.getName();
-								if (!copyFile(userOverridePath, tmpAbsolutePath)) {
-									webAppLog.warn("Unable to copy file in folder defined by runtime property: " + prop);
-									webAppLog.warn("Your source directory (or a file in it) '" + userOverridePath
-									        + " cannot be loaded or destination '" + tmpAbsolutePath + "' cannot be found");
+						if (file.listFiles() != null) {
+							for (File f : Objects.requireNonNull(file.listFiles())) {
```
dkayiwa: Do you need Objects.requireNonNull when you are already sure that folder.listFiles() is not null?

web/src/main/java/org/openmrs/web/Listener.java
BartlomiejRasztabiga: I've made two checks, because:
1) Sonar was still warning about nullable call after I made the "if" statement
2) I didn't know if this call should be guaranteed to be null (so we can throw NPE) or just ignore it by wrapping this code in (if != null) check

dkayiwa: Does this mean folder.listFiles() can be null on line 524?

BartlomiejRasztabiga: I think it cannot be.

BartlomiejRasztabiga: I've pushed a fix for that, leaving only the if statement

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -23,6 +23,7 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
```
dkayiwa: What is the above import used for?

BartlomiejRasztabiga: Sorry, forgot to delete :) Already fixed

PR #2441, opened 2017-12-09
[TRUNK-5242] Remove duplication of Logger

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5242

Deleted duplicated declaration of Logger variable, this class should have only one field Logger which should be private static final.

### DIFF
--- a/web/src/main/java/org/openmrs/web/Listener.java
+++ b/web/src/main/java/org/openmrs/web/Listener.java
@@ -62,9 +62,7 @@ import org.springframework.web.context.WebApplicationContext;
 import org.springframework.web.context.support.XmlWebApplicationContext;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
-import org.xml.sax.EntityResolver;
 import org.xml.sax.InputSource;
-import org.xml.sax.SAXException;
 
 /**
  * Our Listener class performs the basic starting functions for our webapp. Basic needs for starting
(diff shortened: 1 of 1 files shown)


dkayiwa:
@BartlomiejRasztabiga there are still more in this same class. Just search for org.slf4j.Logger

BartlomiejRasztabiga:
@dkayiwa Done :)

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -172,8 +166,7 @@ public void contextInitialized(ServletContextEvent event) {
 				
 				//ensure that we always log the runtime properties file that we are using
 				//since openmrs is just booting, the log levels are not yet set. TRUNK-4835
-				Logger.getLogger(getClass()).setLevel(Level.INFO);
```
dkayiwa: How about setting the log level?

BartlomiejRasztabiga: I think there is no setLevel in slf4j, it is replaced with info() method call

dkayiwa: So how are log levels set?

BartlomiejRasztabiga: I think slf4j doesn't allow dynamic level changes. Maybe it depends on implementation

dkayiwa: Can we then revert to the original one which allowed setting the log level?

BartlomiejRasztabiga: @dkayiwa  Fixed :)

dkayiwa: Ticket status says you are still doing more work on this. 😊

BartlomiejRasztabiga: Done :)

PR #2446, opened 2017-12-09
[TRUNK-5246] Thread-safety: corrected lazy initialization of static field

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5246

I've corrected lazy initialization of two static fields so they can be accesses from any thread.

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Context.java
+++ b/api/src/main/java/org/openmrs/api/context/Context.java
@@ -246,8 +246,12 @@ public class Context {
 	 */
 	static ServiceContext getServiceContext() {
 		if (serviceContext == null) {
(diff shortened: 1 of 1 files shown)


dkayiwa:
@BartlomiejRasztabiga do we need both synchronized and volatile? Or just one of the two?

BartlomiejRasztabiga:
@dkayiwa I've seen examples of Singleton where both volatile and synchronized block were used. I think it just prevents the instance field from having different values for different threads.

dkayiwa:
For the sake of just the reported problem, wouldn't one be enough?

BartlomiejRasztabiga:
I think we can leave just the synchronized blocks

PR #2447, opened 2017-12-09
[TRUNK-5247] Close Resources in finally block

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5247

I've closed one more Resource in finally block and changed if statement to close another resource every time. Sonar would warn otherwise.

### DIFF
--- a/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
+++ b/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
@@ -1174,10 +1174,8 @@ public class InitializationFilter extends StartupFilter {
 			
 			// run the sql statement
 			statement = connection.createStatement();
-			
-			int updateDelta = statement.executeUpdate(replacedSql);
-			statement.close();
-			return updateDelta;
+
+			return statement.executeUpdate(replacedSql);
 			
 		}
 		catch (SQLException sqlex) {
@@ -1192,7 +1190,7 @@ public class InitializationFilter extends StartupFilter {
 		}
 		finally {
 			try {
-				if (statement != null && !statement.isClosed()) {
+				if (statement != null) {
 					statement.close();
 				}
 			}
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1192,7 +1192,7 @@ private int executeStatement(boolean silent, String user, String pw, String sql,
 		}
 		finally {
 			try {
-				if (statement != null && !statement.isClosed()) {
```
dkayiwa: Does this mean that statement.isClosed() is useless?

BartlomiejRasztabiga: I see you are closing statement on this line https://github.com/openmrs/openmrs-core/blob/48386f10175bf557801706f76c79e1413067673b/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java#L1179

We can delete that and close it in finally block, so we do not duplicate code

dkayiwa: OK.

BartlomiejRasztabiga: I've pushed a fix :)

web/src/main/java/org/openmrs/web/filter/util/FilterUtil.java
```diff
@@ -104,6 +102,14 @@ public static String restoreLocale(String username) {
 						log.debug(DATABASE_CLOSING_ERROR, e);
 					}
 				}
+
+				if (results != null) {
+					try {
+						results.close();
+					} catch (SQLException e) {
+						log.warn("Error while closing ResultSet");
```
dkayiwa: Shouldn't you add the exception as the second argument in log.warn?

BartlomiejRasztabiga: Done :)

PR #2451, opened 2017-12-09
[GCI-247] Synchronize on Object rathen than Boolean

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/GCI-247

### DIFF
--- a/api/src/main/java/org/openmrs/hl7/HL7InQueueProcessor.java
+++ b/api/src/main/java/org/openmrs/hl7/HL7InQueueProcessor.java
@@ -30,6 +30,8 @@ public class HL7InQueueProcessor /* implements Runnable */{
 	private static final Logger log = LoggerFactory.getLogger(HL7InQueueProcessor.class);
 	
 	private static Boolean isRunning = false; // allow only one running
+
+	private static final Object lock = new Object();
 	
 	private static Integer count = 0;
 	
@@ -99,7 +101,7 @@ public class HL7InQueueProcessor /* implements Runnable */{
 	 * Starts up a thread to process all existing HL7InQueue entries
 	 */
 	public void processHL7InQueue() throws HL7Exception {
-		synchronized (isRunning) {
+		synchronized (lock) {
 			if (isRunning) {
 				log.warn("HL7 processor aborting (another processor already running)");
 				return;


JudeNiroshan:
How about removing the `isRunning` variable. As the JIRA task is referring to "replace". It would be better to leave a descriptive comment as the existing `isRunning` variable does. 
Please correct the finally block appropriately as well :)

BartlomiejRasztabiga:
@JudeNiroshan What do you think of it? :)

api/src/main/java/org/openmrs/hl7/HL7InQueueProcessor.java
```diff
@@ -28,8 +28,8 @@
 public class HL7InQueueProcessor /* implements Runnable */{
 	
 	private static final Logger log = LoggerFactory.getLogger(HL7InQueueProcessor.class);
-	
-	private static Boolean isRunning = false; // allow only one running
```
dkayiwa: I thought you would not touch the isRunning variable but only that you do not use it for synchronization.

BartlomiejRasztabiga: Look at what @JudeNiroshan said. But ok, I can rever that.

BartlomiejRasztabiga: @dkayiwa Fixed :)

JudeNiroshan: @dkayiwa I think leaving the isRunning variable untouched is not good. As that is a private variable which is only being used solely for synchronization purpose within this class, we are good to replace instead of introducing another new variable. (It's just my personal opinion :) )

dkayiwa: @JudeNiroshan if you take a second look, you will notice that using it as a flag for the running status is quite different from locking. Just that originally, it was used for both saving the running status and locking. You can use a lock for a block of code. Removing the code block and leaving only the lock is what you are suggesting. 😊  Does it make sense?

@BartlomiejRasztabiga how about you, what do you say? 😊

BartlomiejRasztabiga: @dkayiwa I totally agree with you!

dkayiwa:
@JudeNiroshan if you have any objections, i can always revert this. 😊

api/src/main/java/org/openmrs/hl7/HL7InQueueProcessor.java
JudeNiroshan: @dkayiwa "Removing the code block and leaving only the lock is what you are suggesting."
Yes, exactly! I don't understand the requirement of both having a lock and boolean flag together. I think the requirement is to run only one execution in JVM for a given time period. If that so, having a lock is necessary. Meantime, if we want to have only one instance for the entire JVM, using a AtomicBoolean is more appropriate.  https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/atomic/AtomicBoolean.html

What do you think? :)

dkayiwa: @JudeNiroshan one instance for the entire JVM is exactly what we do not want and hence the reason why we created this ticket. Did you get a chance to look at the background information on the ticket? 😊

JudeNiroshan: @dkayiwa In that case; we are good. BTW; you have already merged it  👍

dkayiwa: Was waiting for your feedback to revert :blush

PR #2454, opened 2017-12-10
[TRUNK-5251] Remove final from static methods

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5251

### DIFF
--- a/api/src/main/java/org/openmrs/logic/result/Result.java
+++ b/api/src/main/java/org/openmrs/logic/result/Result.java
@@ -299,7 +299,7 @@ public class Result extends ArrayList<Result> {
 	/**
 	 * @return null/empty result
 	 */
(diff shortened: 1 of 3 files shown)


dkayiwa:
final will prevent the method from being hidden by subclasses which is very good. Do you wanna get rid of this? 😊

BartlomiejRasztabiga:
As I said in ticket description, when a static method is overridden in a subclass it can still be accessed via the super class, making a final declaration not very necessary.

dkayiwa:
When you say "not very necessary", you mean "necessary but not so much".  Do you think we should remove final for something that is still necessary, at least to some extent? 😊

BartlomiejRasztabiga:
Okay, I agree with you :) Leaving final keyword will be better

PR #2460, opened 2017-12-13
[TRUNK-5254] Use getPreferred() instead of isPreferred()

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5254


api/src/main/java/org/openmrs/PatientIdentifier.java
```diff
@@ -204,17 +204,6 @@ public void setPreferred(Boolean preferred) {
 		this.preferred = preferred;
 	}
 	
-	/**
-	 * @return the preferred status
-	 * 
-	 * @deprecated as of 2.0, use {@link #getPreferred()}
-	 */
```
dkayiwa: We do not deprecate and remove at the same time. After deprecation, we still leave it in for a number of versions to come until when we eventually remove it.

BartlomiejRasztabiga: @dkayiwa Fixed :)

PR #2461, opened 2017-12-13
[TRUNK-5255] Use getDead() instead of isDead()

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5255


api/src/main/java/org/openmrs/Person.java
```diff
@@ -268,17 +268,6 @@ public Date getBirthDateTime() {
 	public Date getBirthtime() {
         return this.birthtime;
     }
-
-	/**
-	 * @return Returns the death status.
-	 * 
-	 * @deprecated as of 2.0, use {@link #getDead()}
```
dkayiwa: We do not deprecate and remove at the same time. After deprecation, we still leave it in for a number of versions to come until when we eventually remove it.

BartlomiejRasztabiga: @dkayiwa Fixed

PR #2462, opened 2017-12-13
[TRUNK-5256] Use getCohortByName() instead of getCohort()

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5256


api/src/main/java/org/openmrs/api/CohortService.java
```diff
@@ -102,13 +102,6 @@
 	@Authorized({ PrivilegeConstants.GET_PATIENT_COHORTS })
 	public Cohort getCohortByName(String name) throws APIException;
 	
-	/**
-	 * @deprecated use {@link #getCohortByName(String)}
-	 */
```
dkayiwa: We do not deprecate and remove at the same time. After deprecation, we still leave it in for a number of versions to come until when we eventually remove it.

BartlomiejRasztabiga: @dkayiwa Fixed

PR #2463, opened 2017-12-13
[TRUNK-5257] Use getCohortsContainingPatientId() instead of depreceated method

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5257

### DIFF
--- a/api/src/test/java/org/openmrs/api/CohortServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/CohortServiceTest.java
@@ -414,7 +414,7 @@ public class CohortServiceTest extends BaseContextSensitiveTest {
 		assertTrue(service.getCohort(2).contains(patientToAdd.getPatientId()));
 		
 		// call the method and it should not return the voided cohort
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/api/CohortService.java
```diff
@@ -143,21 +143,6 @@
...
-	 * Find all Cohorts that contain the given patient. (Not including voided Cohorts)
-	 * 
-	 * @param patient patient used to find the cohorts
-	 * @return All non-voided Cohorts that contain the given patient
-	 * @throws APIException
-	 * @should not return voided cohorts
-	 * @should return cohorts that have given patient
-	 */
-	@Deprecated
-	@Authorized({ PrivilegeConstants.GET_PATIENT_COHORTS })
```
dkayiwa: We do not deprecate and remove at the same time. After deprecation, we still leave it in for a number of versions to come until when we eventually remove it.

BartlomiejRasztabiga: @dkayiwa Fixed

PR #2467, opened 2017-12-15
[TRUNK-5260] Replace anonymous type with lambda expression

tomekdo3:
### Issue I worked on
https://issues.openmrs.org/browse/TRUNK-5260

### DIFF
--- a/api/src/main/java/org/openmrs/util/HandlerUtil.java
+++ b/api/src/main/java/org/openmrs/util/HandlerUtil.java
@@ -139,13 +139,7 @@ public class HandlerUtil implements ApplicationListener<ContextRefreshedEvent> {
 		}
 		
 		// Return the list of handlers based on the order specified in the Handler annotation
-		handlers.sort(new Comparator<H>() {
-
-			@Override
-			public int compare(H o1, H o2) {
-				return getOrderOfHandler(o1.getClass()).compareTo(getOrderOfHandler(o2.getClass()));
-			}
-		});
+		handlers.sort(Comparator.comparing(o -> getOrderOfHandler(o.getClass())));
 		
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/util/HandlerUtil.java
```diff
@@ -139,13 +139,7 @@ public static void clearCachedHandlers() {
...
 		
 		// Return the list of handlers based on the order specified in the Handler annotation
-		handlers.sort(new Comparator<H>() {
-
-			@Override
-			public int compare(H o1, H o2) {
-				return getOrderOfHandler(o1.getClass()).compareTo(getOrderOfHandler(o2.getClass()));
-			}
-		});
+		handlers.sort(Comparator.comparing(o -> getOrderOfHandler(o.getClass())));
```
dkayiwa: Did you intentionally do this differently from the other one? https://github.com/openmrs/openmrs-core/commit/9847f2928dd45c90d351c961a35507e814b8d874

tomekdo3: This sorting is simpler than in trunk-5261 so this way is's more readable 😊

dkayiwa: Is there any reason why you did both differently? 😊

tomekdo3: Only readability

dkayiwa: So you decided to make one less readable than the other?

tomekdo3: No. There is no way to simplify lambda in trunk-5261 because there are if statements. I made them both as readable as I could

PR #2470, opened 2017-12-15
[TRUNK-344] PersonName.getFullName() should use the configured nameLayout

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-344

I've made also few changes to NameSupport so it may be reusable and pass all tests

### DIFF
--- a/api/src/main/java/org/openmrs/PersonName.java
+++ b/api/src/main/java/org/openmrs/PersonName.java
@@ -26,6 +26,8 @@ import org.hibernate.search.annotations.Fields;
 import org.hibernate.search.annotations.Indexed;
 import org.hibernate.search.annotations.IndexedEmbedded;
 import org.openmrs.api.db.hibernate.search.LuceneAnalyzers;
+import org.openmrs.layout.name.NameSupport;
+import org.openmrs.layout.name.NameTemplate;
 import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.OpenmrsUtil;
 import org.slf4j.Logger;
@@ -399,6 +401,11 @@ public class PersonName extends BaseChangeableOpenmrsData implements java.io.Ser
 	 * @should not put spaces around an empty middle name
 	 */
 	public String getFullName() {
+		NameTemplate nameTemplate = NameSupport.getInstance().getDefaultLayoutTemplate();
+		if (nameTemplate != null) {
+			return nameTemplate.format(this);
+		}
+
 		List<String> temp = new ArrayList<>();
 		if (StringUtils.hasText(getPrefix())) {
 			temp.add(getPrefix());
--- a/api/src/test/java/org/openmrs/layout/name/NameTemplateTest.java
+++ b/api/src/test/java/org/openmrs/layout/name/NameTemplateTest.java
@@ -20,14 +20,15 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.openmrs.PersonName;
+import org.openmrs.test.BaseContextSensitiveTest;
 
-public class NameTemplateTest {
+public class NameTemplateTest extends BaseContextSensitiveTest {
 	
 	private NameSupport nameSupport;
 	
 	@Before
 	public void setup() {
-		nameSupport = new NameSupport();
+		nameSupport = NameSupport.getInstance();
 		nameSupport.setSpecialTokens(Arrays.asList("prefix", "givenName", "middleName", "familyNamePrefix",
 		    "familyNameSuffix", "familyName2", "familyName", "degree"));
 	}
--- a/api/src/test/resources/TestingApplicationContext.xml
+++ b/api/src/test/resources/TestingApplicationContext.xml
@@ -30,6 +30,10 @@
 		</property>
 	</bean>
 
+	<bean id="nameSupport"
+	      class="org.openmrs.layout.name.NameSupport">
+	</bean>
+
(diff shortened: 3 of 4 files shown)


BartlomiejRasztabiga:
In addition, I think we can replace call to NameSupport constructor with getInstance method, because Singleton shouldn't expose its' constructor publicly.

api/src/main/java/org/openmrs/layout/name/NameSupport.java
```diff
@@ -23,18 +23,15 @@
...
+	private NameSupport() {
+		// hiding default public constructor
 	}
 	
 	public static NameSupport getInstance() {
 		if (singleton == null) {
-			throw new RuntimeException("Not Yet Instantiated");
-		} else {
-			return singleton;
+			singleton = new NameSupport();
```
dkayiwa: Isn't this resulting into a bug caused by a bean different from this? https://github.com/openmrs/openmrs-module-legacyui/blob/master/omod/src/main/resources/webModuleApplicationContext.xml#L754

BartlomiejRasztabiga:
So we have to initialise NameSupport instance before calling getInstance() because otherwise it doesn't pass tests.

BartlomiejRasztabiga:
So If I understand correctly, during normal application run, NameSupport is created automatically by Spring?

dkayiwa:
Correct!

BartlomiejRasztabiga:
I have an issue. If I call NameSupport.getInstance() in getFullName() then I get "Not Yet Instantiated" exception in every test related to it. So we have to initialize this Singleton first, right?

BartlomiejRasztabiga:
@dkayiwa I think I've fixed it :)

api/src/main/java/org/openmrs/layout/name/NameSupport.java
```diff
@@ -36,6 +36,12 @@ public static NameSupport getInstance() {
 			return singleton;
 		}
 	}
+
+	public static void init() {
+		if (singleton == null) {
```
dkayiwa: Is this just for testing?

BartlomiejRasztabiga: This method is used to initialise first instance of NameSupport, otherwise it will throw Not Yet Instantiated"

api/src/main/java/org/openmrs/PersonName.java
```diff
@@ -11,6 +11,7 @@
 
 import static org.apache.commons.lang.StringUtils.defaultString;
 
+import java.lang.reflect.Method;
```
dkayiwa: Where is this import used?

BartlomiejRasztabiga: I can remove that :)

api/src/main/java/org/openmrs/layout/name/NameSupport.java
dkayiwa: Let me rephrase, did you add the init method simply for tests?

BartlomiejRasztabiga: I think so, because Spring doesn't initialise this bean only in unit tests :blush:

dkayiwa: Did you consider putting the bean in TestingApplicationContext.xml? 😊

BartlomiejRasztabiga: So if I add something like this into TestingApplicationContext it would autowire NameTemplate?

        ```<bean id="nameSupport"
	      class="org.openmrs.layout.name.NameSupport">
	```</bean>

BartlomiejRasztabiga: @dkayiwa Could you tell me how to initialise this class instance before running any test? 😊

BartlomiejRasztabiga: I've put something like this
https://github.com/openmrs/openmrs-core/pull/2470/files#diff-de63ef1337d1da40a8826bb63ee8fc92R33

dkayiwa: And did it work? 😊

BartlomiejRasztabiga: Unfortunately no, that's why I ask you

api/src/test/resources/TestingApplicationContext.xml
```diff
@@ -30,6 +30,11 @@
 		</property>
 	</bean>
 
+	<bean id="nameSupport"
+	      class="org.openmrs.layout.name.NameSupport"
+	      factory-method="getInstance">
```
dkayiwa: You can remove factory-method

api/src/test/java/org/openmrs/layout/name/NameTemplateTest.java
```diff
@@ -27,7 +27,8 @@
```
dkayiwa: Change this test's base class to BaseContextSensitiveTest

dkayiwa: And in BaseContextSensitiveTest add "classpath*:TestingApplicationContext.xml"

BartlomiejRasztabiga:
@dkayiwa Thanks! I've made appropiate changes 😊

PR #2488, opened 2017-12-23
[TRUNK-5266] Delete unnecessary 'continue' as the last statement in a loop

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5266

### DIFF
--- a/api/src/main/java/org/openmrs/util/OpenmrsClassLoader.java
+++ b/api/src/main/java/org/openmrs/util/OpenmrsClassLoader.java
@@ -109,6 +109,9 @@ public class OpenmrsClassLoader extends URLClassLoader {
 	 * can happen correctly.
 	 */
 	private static class OpenmrsClassLoaderHolder {
+
+		private OpenmrsClassLoaderHolder() {
+		}
 		
 		private static OpenmrsClassLoader INSTANCE = null;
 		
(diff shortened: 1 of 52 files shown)


api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -1206,7 +1206,6 @@ public static void saveDocument(Document doc, File outFile) {
 		for (String token : tokens) {
 			token = token.trim();
 			if (token.length() == 0) {
```
dkayiwa: Can't we simplify this by only having an if statement without else?

api/src/main/java/org/openmrs/util/OpenmrsClassLoader.java
```diff
@@ -590,7 +590,6 @@ protected static void nullInstance(Object instance) {
 				field.setAccessible(true);
 				if (Modifier.isStatic(mods) && Modifier.isFinal(mods)) {
 					// Doing something recursively is too risky
-					continue;
```
dkayiwa: Can't we simplify this by only having an if statement without else?

BartlomiejRasztabiga: Fixed 😊

PR #2497, opened 2017-12-25
[TRUNK-5271] Replace StringBuffer/StringBuilder with String in simple statements

tomekdo3:
#### Issue: https://issues.openmrs.org/browse/TRUNK-5271

### DIFF
--- a/api/src/main/java/org/openmrs/PersonAddress.java
+++ b/api/src/main/java/org/openmrs/PersonAddress.java
@@ -103,11 +103,10 @@ public class PersonAddress extends BaseChangeableOpenmrsData implements java.io.
 	 */
 	@Override
 	public String toString() {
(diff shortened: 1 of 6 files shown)


dkayiwa:
Can you add some more description to the ticket? In other words, make it convincing to someone who may not yet know why you are doing this.

tomekdo3:
@dkayiwa I think it's all clear now. I've changed ticket and PR titles too 😊

dkayiwa:
You need to explain a bit, in the ticket description, why performance is the same because some people know StringBuilder to be more performant than string concatenation. 😊

tomekdo3:
@dkayiwa is that ok?

dkayiwa:
Yes that is a better and more complete explanation! 👍

PR #2508, opened 2017-12-27
[TRUNK-5279] Access modifier could be private

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5279

### DIFF
--- a/api/src/main/java/org/openmrs/Allergies.java
+++ b/api/src/main/java/org/openmrs/Allergies.java
@@ -326,7 +326,7 @@ public class Allergies implements List<Allergy> {
 	 * @param allergies the allergies whose allergens to compare with
 	 * @return true if the same allergen exists, else false
 	 */
-	public boolean containsAllergen(Allergy allergy, Collection<? extends Allergy> allergies) {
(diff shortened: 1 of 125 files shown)


dkayiwa:
Did you see the codacy-bot comments?

BartlomiejRasztabiga:
Yes, working on it 😊

dkayiwa:
@BartlomiejRasztabiga do we still need this? 😊

BartlomiejRasztabiga:
I will fix that after I do GCI task :)

teleivo:
@BartlomiejRasztabiga I don't think we should make these changes as these would most probably break modules code: for example Allergy.setPatient() was public which means you dont know who is using it just because its not referenced in openmrs-core (longer explanation here https://issues.openmrs.org/browse/TRUNK-5279).

@dkayiwa @wluyima what do you think?

BartlomiejRasztabiga:
Okay, I didn't think about other code from outside of openmrs-core. I think i will just close the PR. Thanks for explanation :)

dkayiwa:
@teleivo i fully agree! 👍

PR #2509, opened 2017-12-27
[TRUNK-5280] Redundant default parameter value assignment in annotations

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5280


dkayiwa:
What happens if the default value is changed?

BartlomiejRasztabiga:
I know it would break down, but there are cases where it probably won't be changed, like a length of column (255).

dkayiwa:
And isn't the original more readable? In other words, not having to look for default value to know the length, to know if a transaction is read only, etc

BartlomiejRasztabiga:
You are right, closing 😊

PR #2510, opened 2017-12-27
[TRUNK-5281] Duplicated exceptions in throws clause

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5281

### DIFF
--- a/api/src/main/java/org/openmrs/module/ModuleUtil.java
+++ b/api/src/main/java/org/openmrs/module/ModuleUtil.java
@@ -67,7 +67,7 @@ public class ModuleUtil {
 	 *
 	 * @param props Properties (OpenMRS runtime properties)
 	 */
-	public static void startup(Properties props) throws ModuleMustStartException, OpenmrsCoreModuleException {
+	public static void startup(Properties props) throws OpenmrsCoreModuleException {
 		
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -67,7 +67,7 @@ private ModuleUtil() {
 	 *
 	 * @param props Properties (OpenMRS runtime properties)
 	 */
-	public static void startup(Properties props) throws ModuleMustStartException, OpenmrsCoreModuleException {
+	public static void startup(Properties props) throws ModuleMustStartException {
```
dkayiwa: Shouldn't instead be leaving the more specific exception like OpenmrsCoreModuleException?

BartlomiejRasztabiga:
@dkayiwa You are right, fixed :)

dkayiwa:
@BartlomiejRasztabiga though this is valid change, i have noticed that it will break existing API clients. Not so?

BartlomiejRasztabiga:
Which for example? 😊

dkayiwa:
This being a public API, you cannot tell because it could be in any module, including proprietary ones. 😊

BartlomiejRasztabiga:
Okay, got it :D Closing

PR #2523, opened 2017-12-28
[TRUNK-5294] Delete unused assignments

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5294


dkayiwa:
Why do you call it "unused assignment"?

dkayiwa:
@BartlomiejRasztabiga did you see the above question?

BartlomiejRasztabiga:
I call them so because they could be as well unassigned. It wouldn't make a difference and will reduce some redundant code.

PR #2532, opened 2017-12-30
[TRUNK-5299] Possible NPE due to use of Objects.requireNonNull()

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/TRUNK-5299

### DIFF
--- a/api/src/main/java/org/openmrs/module/ModuleFactory.java
+++ b/api/src/main/java/org/openmrs/module/ModuleFactory.java
@@ -171,9 +171,10 @@ public class ModuleFactory {
 		if (log.isDebugEnabled()) {
 			log.debug("Loading modules from: " + modulesFolder.getAbsolutePath());
 		}
-		
-		if (modulesFolder.isDirectory()) {
-			loadModules(Arrays.asList(Objects.requireNonNull(modulesFolder.listFiles())));
+
+		File[] fileList = modulesFolder.listFiles();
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -171,9 +171,10 @@ public static void loadModules() {
...
 		}
-		
-		if (modulesFolder.isDirectory()) {
-			loadModules(Arrays.asList(Objects.requireNonNull(modulesFolder.listFiles())));
+
+		File[] fileList = modulesFolder.listFiles();
+		if (modulesFolder.isDirectory() && fileList != null) {
+			loadModules(Arrays.asList(fileList));
 		} else {
 			log.error("modules folder: '" + modulesFolder.getAbsolutePath() + "' is not a valid directory");
```
teleivo: now the log message also includes a directory which are empty which I assume is still a valid directory.

BartlomiejRasztabiga: moduleFolder.listFiles() returns null only when directory is not valid

teleivo:
closing this since we should reopen the issues that introduced these changes and fix them

BartlomiejRasztabiga:
Okay, should I reopen them or you will do it and ping me after?

PR #2537, opened 2017-12-31
[GCI-237] Possible null pointer dereference due to return value of modu…

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/GCI-237

I've decided to log the same error message when files are null. Because listFiles() returns null only in two cases: 1) directory is not valid 2) there is IO Exception. And the first case is already checked by modulesFolder.isDirectory() so we're getting null only on IO Exception which is highly uncommon here (directory is checked already in ModuleUtil).


BartlomiejRasztabiga:
Done

PR #2564, opened 2018-01-29
RA-543: Patient name should not allow leading or trailing spaces

samuelmale:
ticket: https://issues.openmrs.org/browse/RA-543

### DIFF
--- a/api/src/main/java/org/openmrs/validator/PersonNameValidator.java
+++ b/api/src/main/java/org/openmrs/validator/PersonNameValidator.java
@@ -128,26 +128,42 @@ public class PersonNameValidator implements Validator {
 			errors.reject("error.name");
 			return;
 		}
+		
 		// Make sure they assign a name
-		if (StringUtils.isBlank(personName.getGivenName())
+		if (StringUtils.isBlank(personName.getGivenName()) 
 		        || StringUtils.isBlank(personName.getGivenName().replaceAll("\"", ""))) {
 			errors.rejectValue(getFieldKey("givenName", arrayInd, testInd), "Patient.names.required.given.family");
 		}
 
+		// Make sure entered name value has no leading and trailing spaces
+		if (startsOrEndsWithWhitespace(personName.getGivenName())) {
+			errors.rejectValue(getFieldKey("givenName", arrayInd, testInd), "GivenName.startsOrEndsWithWhitespace");
+		}
+		if (startsOrEndsWithWhitespace(personName.getMiddleName())) {
+			errors.rejectValue(getFieldKey("middleName", arrayInd, testInd), "MiddleName.startsOrEndsWithWhitespace");
+		}
+		if (startsOrEndsWithWhitespace(personName.getFamilyName())) {
+			errors.rejectValue(getFieldKey("familyName", arrayInd, testInd), "FamilyName.startsOrEndsWithWhitespace");
+		}
+		if (startsOrEndsWithWhitespace(personName.getFamilyName2())) {
+			errors.rejectValue(getFieldKey("familyName2", arrayInd, testInd), "FamilyName2.startsOrEndsWithWhitespace");
+		}
+		
+		
 		// Make sure the entered name value is sensible 
 		String namePattern = Context.getAdministrationService().getGlobalProperty(
 		    OpenmrsConstants.GLOBAL_PROPERTY_PATIENT_NAME_REGEX);
 		if (StringUtils.isNotBlank(namePattern)) {
-			if (StringUtils.isNotBlank(personName.getGivenName()) && !personName.getGivenName().matches(namePattern)) {
+			if (!matchNamePattern(personName.getGivenName(), namePattern)) {
 				errors.rejectValue(getFieldKey("givenName", arrayInd, testInd), "GivenName.invalid");
 			}
-			if (StringUtils.isNotBlank(personName.getMiddleName()) && !personName.getMiddleName().matches(namePattern)) {
+			if (!matchNamePattern(personName.getMiddleName(), namePattern)) {
 				errors.rejectValue(getFieldKey("middleName", arrayInd, testInd), "MiddleName.invalid");
 			}
-			if (StringUtils.isNotBlank(personName.getFamilyName()) && !personName.getFamilyName().matches(namePattern)) {
+			if (!matchNamePattern(personName.getFamilyName(), namePattern)) {
 				errors.rejectValue(getFieldKey("familyName", arrayInd, testInd), "FamilyName.invalid");
 			}
-			if (StringUtils.isNotBlank(personName.getFamilyName2()) && !personName.getFamilyName2().matches(namePattern)) {
+			if (!matchNamePattern(personName.getFamilyName2(), namePattern)) {
 				errors.rejectValue(getFieldKey("familyName2", arrayInd, testInd), "FamilyName2.invalid");
 			}
 		}
@@ -162,6 +178,18 @@ public class PersonNameValidator implements Validator {
 	 */
 	private String getFieldKey(String field, boolean arrayInd, boolean testInd) {
 		return testInd ? field : arrayInd ? "names[0]." + field : "name." + field;
+	}
+	
+	private boolean startsOrEndsWithWhitespace(String value) {
+		return (StringUtils.isNotBlank(value) && (Character.isSpaceChar(value.charAt(0)) || Character.isSpaceChar(value.charAt(value.length() -1))));		
+	}
+	
+	private boolean matchNamePattern(String value, String namePattern) {
+		boolean match = true;
+		if (StringUtils.isNotBlank(value)) {
+			match = value.matches(namePattern);
+		} 
+		return match;
 	}
 	
 }
--- a/api/src/main/resources/messages.properties
+++ b/api/src/main/resources/messages.properties
@@ -3303,3 +3303,9 @@ Condition.clinicalStatusShouldNotBeNull=Clinical status is required
 
 mail.passwordreset.subject=Password Reset
 mail.passwordreset.content=Hi {name} \n You recently requested a password reset for your OpenMRS account. To complete the process, click the link below.\n {link} \n  Please disregard this email if you didn't innitiate the password reset process. \n This link will automatically expire after a period of {time} minutes.
+
+GivenName.startsOrEndsWithWhitespace=GivenName may not start or end with blank spaces
+MiddleName.startsOrEndsWithWhitespace=MiddleName may not start or end with blank spaces
+FamilyName.startsOrEndsWithWhitespace=FamilyName may not start or end with blank spaces
+FamilyName2.startsOrEndsWithWhitespace=FamilyName2 may not start or end with blank spaces
+
--- a/api/src/test/java/org/openmrs/validator/PersonNameValidatorTest.java
+++ b/api/src/test/java/org/openmrs/validator/PersonNameValidatorTest.java
@@ -757,7 +757,6 @@ public class PersonNameValidatorTest extends BaseContextSensitiveTest {
 	 * @see PatientNameValidator#validatePersonName(java.lang.Object, org.springframework.validation.Errors, boolean, boolean)
 	 */
 	@Test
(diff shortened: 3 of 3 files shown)


samuelmale:
@dkayiwa  I fixed broken Tests. done some channges

djazayeri:
This is a commit to openmrs-core, but the ticket is for the Reference Application project. We'd need to have a ticket for openmrs-core (i.e. TRUNK-xyz) before merging this.

samuelmale:
@djazayeri  I have done more changes

api/src/main/resources/messages.properties
```diff
@@ -3291,3 +3291,8 @@ OrderFrequency.concept.shouldBeClassFrequency=Concept should be of class Frequen
 OrderFrequency.concept.shouldNotBeShared=Only one OrderFrequency should be mapped to a particular concept
 
 Module.error.cyclicDependencies=Could not start some modules because they depend on others which are not started. Please first start them or use the "Start All" button on the "Manage Modules" page: {0}
+
+GivenName.startsOrEndsWithWhitespace=Invalid GivenName, Trailing or Leading Whitespace detected
```
djazayeri: the error message should be something like "Given name may not start or end with blank spaces"

djazayeri:
Someone (whether @samuelmale or @dkayiwa) needs to check why this situation isn't already being handled by the pre-existing code that I mentioned, before we commit to merging this.

Ben-Waters:
@djazayeri   Maybe I'm missing something but I think you're right. Running Openmrs-core version 2.2.0-SNAPSHOT  I created a new patient with both leading and trailing white space around their first middle and last names. On save, all the whitespace was removed. I then tried to edit the patient and add whitespace. On save it was all removed again.

The ticket points to a devtest2 environment which seems to be down, so I'm not able to reproduce it there. Is there a way to either start up devtest2 or find out what version it was running to try to reproduce it locally?

dkayiwa:
@Ben-Waters did you get a chance to look at the ticket comments? I think there is one which seems to address your question. 😊

teleivo:
@Ben-Waters or @samuelmale is this issue already handled by the pre-existing code as @djazayeri mentioned?

If not @samuelmale please resolve the conflicts in the PR.

Ben-Waters:
When I investigated the existing code, it is fixing it as it saves to the DB but not as the user is editing and validating it.

samuelmale:
So think we should move this to the validation layer :)

I have just fixed the merge conflicts :)

teleivo:
@samuelmale I dont think that Validators are the right place. I would not expect a Validator to mutate something I pass into it, it should only tell me if that thing is valid and if it isnt what is invalid about it.

samuelmale:
@teleivo  But according to my implementation in the latest commit, Validators don't mutate any values. They tell us if a value is valid or not

api/src/main/java/org/openmrs/validator/PersonNameValidator.java
```diff
@@ -129,11 +129,26 @@ public void validatePersonName(PersonName personName, Errors errors, boolean arr
...
 		}
 		// Make sure they assign a name
-		if (StringUtils.isBlank(personName.getGivenName())
+		if (StringUtils.isBlank(personName.getGivenName()) 
 		        || StringUtils.isBlank(personName.getGivenName().replaceAll("\"", ""))) {
 			errors.rejectValue(getFieldKey("givenName", arrayInd, testInd), "Patient.names.required.given.family");
 		}
 
+		// Make sure entered name value has no leading and trailing spaces
+		if (StringUtils.isNotBlank(personName.getGivenName()) && startsOrEndsWithWhitespace(personName.getGivenName())) {
```
teleivo: please move `StringUtils.isNotBlank(personName.getGivenName())` into the private method so we dont have to repeat this for every field, plus if its blank it does not start or end with whitespace so this fits into the duty of the method

api/src/main/java/org/openmrs/validator/PersonNameValidator.java
```diff
@@ -164,4 +179,22 @@ private String getFieldKey(String field, boolean arrayInd, boolean testInd) {
 		return testInd ? field : arrayInd ? "names[0]." + field : "name." + field;
 	}
 	
+	/*
+	 * Checks that the given {@link PersonName} field has no trailing and leading spaces.
+	 * @param value the value name.
+	 * @return containsSpace
+	 * @should return true if given field contains trailing or leading space(s), otherwise false.
```
teleivo: remove javadocs of this private method, the method has a clear name which says it all

api/src/main/java/org/openmrs/validator/PersonNameValidator.java
```diff
@@ -164,4 +179,22 @@ private String getFieldKey(String field, boolean arrayInd, boolean testInd) {
...
+	 */
+	private boolean startsOrEndsWithWhitespace(String value) {
+		boolean startsOrEndsWithWhitespace = false;
+		int lastIndex = value.length() - 1;
+		if (value.charAt(0) == ' ') {
+			startsOrEndsWithWhitespace = true;
+		} else if (Character.isWhitespace(value.charAt(lastIndex))) {
```
teleivo: this also contains file separator, is this what you want? if not maybe thats better
https://docs.oracle.com/javase/7/docs/api/java/lang/Character.html#isSpaceChar(char)

api/src/main/java/org/openmrs/validator/PersonNameValidator.java
```diff
@@ -164,4 +179,22 @@ private String getFieldKey(String field, boolean arrayInd, boolean testInd) {
...
+	/*
+	 * Checks that the given {@link PersonName} field has no trailing and leading spaces.
+	 * @param value the value name.
+	 * @return containsSpace
+	 * @should return true if given field contains trailing or leading space(s), otherwise false.
+	 */
+	private boolean startsOrEndsWithWhitespace(String value) {
+		boolean startsOrEndsWithWhitespace = false;
+		int lastIndex = value.length() - 1;
+		if (value.charAt(0) == ' ') {
```
teleivo: why use space here and below `Character.isWhitespace(value.charAt(lastIndex)))`?

Use the same for beginning and end.

api/src/main/java/org/openmrs/validator/PersonNameValidator.java
```diff
@@ -164,4 +179,22 @@ private String getFieldKey(String field, boolean arrayInd, boolean testInd) {
...
+			startsOrEndsWithWhitespace = true;
+		} else if (Character.isWhitespace(value.charAt(lastIndex))) {
+			startsOrEndsWithWhitespace = true; 
+		}
+		
+		return startsOrEndsWithWhitespace;
+
```
teleivo: remove newline

api/src/test/java/org/openmrs/validator/PersonNameValidatorTest.java
```diff
@@ -864,4 +856,5 @@ public void validate_shouldFailValidationIfPersonNameFamilyName2HasTrailingSpace
 		
 		assertThat(errors, hasFieldErrors("familyName2"));
 	}
+
```
teleivo: remove newline

teleivo:
@djazayeri can you please check this PR again, it has been updated.

samuelmale:
Implemented requested changes. @teleivo

teleivo:
@samuelmale I would like to have @djazayeri to have a look at this PR as I dont have the context.

api/src/main/java/org/openmrs/validator/PersonNameValidator.java
```diff
@@ -164,4 +180,31 @@ private String getFieldKey(String field, boolean arrayInd, boolean testInd) {
 		return testInd ? field : arrayInd ? "names[0]." + field : "name." + field;
 	}
 	
-}
+	
+	private boolean startsOrEndsWithWhitespace(String value) {
+		boolean startsOrEndsWithWhitespace = false;
+		if (value != null && StringUtils.isNotBlank(value)) {
```
dkayiwa: Isn't isNotBlank already checking for null?

samuelmale: Just addressed your comment :)

api/src/main/java/org/openmrs/validator/PersonNameValidator.java
```diff
@@ -164,4 +180,31 @@ private String getFieldKey(String field, boolean arrayInd, boolean testInd) {
...
+			    startsOrEndsWithWhitespace = true; 
+		    }
+		} else {
+			startsOrEndsWithWhitespace = false;
+		}
+		return startsOrEndsWithWhitespace;		
+	}
+	
+	private boolean matchNamePattern(String value, String namePattern) {
+		//True by default because this shouldNotValidateAgainstRegexForBlankNames
```
dkayiwa: Is the above supposed to be an English sentence?

samuelmale: Literally meaning I get the comment off? @dkayiwa

samuelmale:
Just removed the comment

api/src/main/java/org/openmrs/validator/PersonNameValidator.java
```diff
@@ -164,4 +180,29 @@ private String getFieldKey(String field, boolean arrayInd, boolean testInd) {
...
+		    }
+		} else {
+			startsOrEndsWithWhitespace = false;
+		}
+		return startsOrEndsWithWhitespace;		
+	}
+	
+	private boolean matchNamePattern(String value, String namePattern) {
+		boolean match = true;
+		if (value != null && StringUtils.isNotBlank(value)) {
```
dkayiwa: Doesn't StringUtils.isNotBlank already compare for null?

api/src/main/java/org/openmrs/validator/PersonNameValidator.java
```diff
@@ -164,4 +180,29 @@ private String getFieldKey(String field, boolean arrayInd, boolean testInd) {
 		return testInd ? field : arrayInd ? "names[0]." + field : "name." + field;
 	}
 	
-}
+	
+	private boolean startsOrEndsWithWhitespace(String value) {
+		boolean startsOrEndsWithWhitespace = false;
```
suthagar23: I think, here you made more complex for this function. Can we think like this way to reduce the nesting? 

```
private boolean startsOrEndsWithWhitespace(String value) {
   return (StringUtils.isNotBlank(value) && (Character.isSpaceChar(value.charAt(0)) || Character.isSpaceChar(value.charAt(-1)));
}
```

api/src/main/java/org/openmrs/validator/PersonNameValidator.java
```diff
@@ -164,4 +180,29 @@ private String getFieldKey(String field, boolean arrayInd, boolean testInd) {
...
+			    startsOrEndsWithWhitespace = true; 
+		    }
+		} else {
+			startsOrEndsWithWhitespace = false;
+		}
+		return startsOrEndsWithWhitespace;		
+	}
+	
+	private boolean matchNamePattern(String value, String namePattern) {
+		boolean match = true;
```
suthagar23: Good to make this without the `if` condition. Just enough

`return StringUtils.isNotBlank(value) && value.matches(namePattern); `

api/src/main/java/org/openmrs/validator/PersonNameValidator.java
suthagar23: @dkayiwa  I think, the `StingUtils.isNotBlank` has the capability to check for `null` also. 
```
 StringUtils.isNotBlank(null)      = false
 StringUtils.isNotBlank("")        = false
 StringUtils.isNotBlank(" ")       = false
```

dkayiwa:
@samuelmale did you see the comments on this?

samuelmale:
Just addressed the comments. Thanks @suthagar23

samuelmale:
@dkayiwa ping

dkayiwa:
@samuelmale did you see the merge conflicts on this?

PR #2610, opened 2018-03-01
TRUNK-5353: Create an implementation of the DiagnosisService interface.

pluwum:
This PR is an implementation of the DiagnosisService interface named DiagnosisServiceImpl.java and delivers tests for the newly added functionality.

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-5353

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
@@ -1,0 +1,142 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.impl;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.openmrs.Diagnosis;
+import org.openmrs.Encounter;
+import org.openmrs.Patient;
+import org.openmrs.api.APIException;
+import org.openmrs.api.DiagnosisService;
+import org.openmrs.api.context.Context;
+import org.openmrs.api.db.DiagnosisDAO;
+import org.openmrs.parameter.EncounterSearchCriteria;
+import org.openmrs.parameter.EncounterSearchCriteriaBuilder;
+/**
+ * API functionality for Diagnoses.
+ *
+ * @since 2.2
+ */
+public class DiagnosisServiceImpl extends BaseOpenmrsService implements DiagnosisService {
+
+	private DiagnosisDAO diagnosisDAO;
+
+	public DiagnosisDAO getDiagnosisDAO() {
+		return diagnosisDAO;
+	}
+
+	public void setDiagnosisDAO(DiagnosisDAO dao) {
+		this.diagnosisDAO = dao;
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#save(Diagnosis)
+	 */
+	public Diagnosis save(Diagnosis diagnosis) {
+		return diagnosisDAO.saveDiagnosis(diagnosis);
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#voidDiagnosis(Diagnosis, String)
+	 */
+	public Diagnosis voidDiagnosis(Diagnosis diagnosis, String voidReason) {
+		return Context.getDiagnosisService().save(diagnosis);
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#getDiagnosisByUuid(String)
+	 */
+	public Diagnosis getDiagnosisByUuid(String uuid) {
+		return diagnosisDAO.getDiagnosisByUuid(uuid);
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#getDiagnoses(Patient, Date)
+	 */
+	public List<Diagnosis> getDiagnoses(Patient patient, Date fromDate) {
+
+		EncounterSearchCriteria encounterSearchCriteria = new EncounterSearchCriteriaBuilder().setPatient(patient)
+			.setFromDate(fromDate).createEncounterSearchCriteria();
+		List<Encounter> encounters = Context.getEncounterService().getEncounters(encounterSearchCriteria);
+
+		List<Diagnosis> diagnoses = new ArrayList<Diagnosis>();
+
+		for (Encounter encounter : encounters) {
+			diagnoses.addAll(encounter.getDiagnoses());
+		}
+		return diagnoses;
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#getPrimaryDiagnoses(Encounter)
+	 */
+	public List<Diagnosis> getPrimaryDiagnoses(Encounter encounter) {
+		List<Diagnosis> diagnoses = new ArrayList<Diagnosis>();
+		if (encounter != null && !encounter.getVoided()) {
+
+			for (Diagnosis diagnosis : encounter.getDiagnoses()) {
+				// check if the diagnosis is given the highest importance. 
+				if (diagnosis.getRank() == 1) {
+					diagnoses.add(diagnosis);
+				}
+			}
+		}
+		return diagnoses;
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#hasDiagnosis(Encounter, Diagnosis)
+	 */
+	public boolean hasDiagnosis(Encounter encounter, Diagnosis diagnosis) {
+
+		for (Diagnosis diagnosisFound : encounter.getDiagnoses()) {
+			if (diagnosis.equals(diagnosisFound) && !diagnosisFound.getVoided()) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#getUniqueDiagnoses(Patient, Date)
+	 */
+	public List<Diagnosis> getUniqueDiagnoses(Patient patient, Date fromDate) {
+
+		Set<Diagnosis> uniqueDiagnoses = new HashSet<Diagnosis>();
+		uniqueDiagnoses.addAll(getDiagnoses(patient, fromDate));
+		return new ArrayList(uniqueDiagnoses);
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#unvoidDiagnosis(Diagnosis)
+	 */
+	public Diagnosis unvoidDiagnosis(Diagnosis diagnosis) throws APIException {
+		return Context.getDiagnosisService().save(diagnosis);
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#purgeDiagnosis(Diagnosis)
+	 */
+	public void purgeDiagnosis(Diagnosis diagnosis) throws APIException {
+		diagnosisDAO.deleteDiagnosis(diagnosis);
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#getDiagnosis(Integer)
+	 */
+	public Diagnosis getDiagnosis(Integer diagnosisId) {
+		return diagnosisDAO.getDiagnosisById(diagnosisId);
+	}
+}
--- a/api/src/test/java/org/openmrs/api/impl/DiagnosisServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/impl/DiagnosisServiceTest.java
@@ -1,0 +1,202 @@
+
+package org.openmrs.api.impl;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.List;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.openmrs.ConditionVerificationStatus;
+import org.openmrs.Diagnosis;
+import org.openmrs.Encounter;
+import org.openmrs.Patient;
+import org.openmrs.api.DiagnosisService;
+import org.openmrs.api.EncounterService;
+import org.openmrs.api.PatientService;
+import org.openmrs.api.context.Context;
+import org.openmrs.test.BaseContextSensitiveTest;
+
+/**
+ * Test methods for {@link DiagnosisServiceImpl}
+ */
+public class DiagnosisServiceTest extends BaseContextSensitiveTest {
+
+	protected static final String DIAGNOSIS_XML = "org/openmrs/api/include/DiagnosisServiceImplTest-SetupDiagnosis.xml";
+
+	private DiagnosisService diagnosisService;
+
+	private EncounterService encounterService;
+
+	private PatientService patientService;
+
+	@Before
+	public void setup() {
+		if (diagnosisService == null) {
+			diagnosisService = Context.getDiagnosisService();
+		}
+		if (encounterService == null) {
+			encounterService = Context.getEncounterService();
+		}
+		if (patientService == null) {
+			patientService = Context.getPatientService();
+		}
+		executeDataSet(DIAGNOSIS_XML);
+	}
+
+	/**
+	 * Creates a diganosis object.
+	 *
+	 * @return diagnosis the diagnosis created
+	 */
+	private Diagnosis buildDiagnosis() {
+		Diagnosis diagnosis = new Diagnosis();
+		diagnosis.setEncounter(encounterService.getEncounter(1));
+		diagnosis.setRank(1);
+		diagnosis.setCertainty(ConditionVerificationStatus.PROVISIONAL);
+		diagnosis.setUuid("2cd6780e-2a46-11e4-9038-a6c5e4d26fc2");
+		diagnosis.setPatient(new Patient(2));
+		return diagnosis;
+	}
+
+	/**
+	 * @see DiagnosisService#save(Diagnosis)
+	 */
+	@Test
+	public void save_shouldSaveDiagnosisWithBasicDetails() {
+		Diagnosis diagnosis = buildDiagnosis();
+		diagnosisService.save(diagnosis);
+		Diagnosis newDiagnosis = diagnosisService.getDiagnosis(diagnosis.getId());
+		Integer expectedRank = 1;
+		Integer expectedPatientId = 2;
+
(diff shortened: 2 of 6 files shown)


api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
```diff
@@ -0,0 +1,170 @@
...
+		}
+		return dao.saveDiagnosis(diagnosis);
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#voidDiagnosis(Diagnosis, String)
+	 */
+	public Diagnosis voidDiagnosis(Diagnosis diagnosis, String voidReason) {
+		diagnosis.setVoided(true);
+		diagnosis.setVoidedBy(Context.getAuthenticatedUser());
```
dkayiwa: Did you get a chance to look at voidObs in ObsServiceImpl?

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
```diff
@@ -0,0 +1,170 @@
...
+		}
+		return diagnoses;
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#unvoidDiagnosis(Diagnosis)
+	 */
+	public Diagnosis unvoidDiagnosis(Diagnosis diagnosis) throws APIException {
+		if (diagnosis.getVoided()) {
+			diagnosis.setVoided(false);
```
dkayiwa: Did you get a chance to look at unvoidObs in ObsServiceImpl?

api/src/main/java/org/openmrs/util/PrivilegeConstants.java
```diff
@@ -27,507 +27,507 @@ private PrivilegeConstants() {
...
+
 	@AddOnStartup(description = "Able to edit visits")
 	public static final String EDIT_VISITS = "Edit Visits";
-	
+
 	@AddOnStartup(description = "Able to delete visits")
 	public static final String DELETE_VISITS = "Delete Visits";
-	
+
 	public static final String PURGE_VISITS = "Purge Visits";
```
dkayiwa: Can you discard the formatting changes you made to this file?

api/src/main/java/org/openmrs/api/db/DiagnosisDAO.java
```diff
@@ -21,7 +21,7 @@
  * @since 2.2
  */
 public interface DiagnosisDAO {
```
wluyima: This file seems not to have any real changes, you can just revert these format changes

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
```diff
@@ -0,0 +1,170 @@
...
+import org.openmrs.util.PrivilegeConstants;
+
+/**
+ * API functionality for Diagnoses.
+ *
+ * @since 2.2
+ */
+public class DiagnosisServiceImpl extends BaseOpenmrsService implements DiagnosisService {
+
+	public static final String DIAGNOSIS_ORDER_PRIMARY = "Primary";
```
wluyima: Is this supposed to be public? Because I'd think this is a wrong place to put it since the service Impl is actually hidden away from API consumer

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
wluyima: These values are set by AOP via some void handler for data therefore you don't have to set them again.

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
```diff
@@ -0,0 +1,170 @@
...
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#voidDiagnosis(Diagnosis, String)
+	 */
+	public Diagnosis voidDiagnosis(Diagnosis diagnosis, String voidReason) {
+		diagnosis.setVoided(true);
+		diagnosis.setVoidedBy(Context.getAuthenticatedUser());
+		diagnosis.setVoidReason(voidReason);
+		diagnosis.setDateVoided(new Date());
+		return save(diagnosis);
```
wluyima: You shouldn't be calling save directly because you will skip any logic around the spring proxy object.
You need to replace this with:

`Context.getDiagnosisService().save(diagnosis);`

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
```diff
@@ -0,0 +1,170 @@
...
+			diagnoses.addAll(encounter.getDiagnoses());
+		}
+		return diagnoses;
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#getPrimaryDiagnoses(Encounter)
+	 */
+	public List<Diagnosis> getPrimaryDiagnoses(Encounter encounter) {
+		List<Diagnosis> diagnoses = null;
```
wluyima: You probably don't want to have this return null, it should return an empty list worst case scenario.

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
```diff
@@ -0,0 +1,170 @@
...
+		if (encounter != null && !encounter.getVoided()) {
+			diagnoses = new ArrayList<Diagnosis>();
+
+			for (Diagnosis diagnosis :
+				encounter.getDiagnoses()) {
+				// check if the diagnosis is given the highiest importance. 
+				if (diagnosis.getOrder == DIAGNOSIS_ORDER_PRIMARY) {
```
wluyima: Strings in java are objects so '==' won't work for equality, use equals

wluyima: Did you add any unit tests for this method? Because if you did, you would have caught this.

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#hasDiagnosis(Encounter, Diagnosis)
+	 */
+	public boolean hasDiagnosis(Encounter encounter, Diagnosis diagnosis) {
+
+		for (Diagnosis diagnosisFound :
+			encounter.getDiagnoses()) {
+			if (diagnosis == diagnosisFound) {
```
wluyima: Do you have unit tests? Again you need to use equals

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
```diff
@@ -0,0 +1,170 @@
...
+	 */
+	public List<Diagnosis> getUniqueDiagnoses(Patient patient, Date fromDate) {
+		List<Diagnosis> diagnoses = getDiagnoses(patient, fromDate);
+
+		Set<CodedOrFreeText> uniqueDiagnoses = new HashSet<CodedOrFreeText>();
+		Iterator<Diagnosis> it = diagnoses.iterator();
+
+		while (it.hasNext()) {
+			Diagnosis diagnosis = it.next();
+			if (!uniqueDiagnoses.add(diagnosis.getDiagnosis())) {
```
wluyima: `uniqueDiagnoses` is a Set and by design it doesn't allow duplicates, all you need to get rid of duplicates is the call below:

`uniqueDiagnoses.addAll(diagnoses);`

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
```diff
@@ -0,0 +1,170 @@
...
+	 * @see org.openmrs.api.DiagnosisService#unvoidDiagnosis(Diagnosis)
+	 */
+	public Diagnosis unvoidDiagnosis(Diagnosis diagnosis) throws APIException {
+		if (diagnosis.getVoided()) {
+			diagnosis.setVoided(false);
+			diagnosis.setVoidedBy(null);
+			diagnosis.setDateVoided(null);
+			diagnosis.setVoidReason(null);
+		}
+		return save(diagnosis);
```
wluyima: Same here, you don't need to set anything here and you need to get the service from the Context and call save on that

api/src/test/java/org/openmrs/api/DiagnosisServiceTest.java
```diff
@@ -0,0 +1,74 @@
...
+	@Test
+	public void save_shouldSaveDiagnosisWithBasicDetails() {
+		Diagnosis diagnosis = buildDiagnosis();
+		DiagnosisService ds = Context.getDiagnosisService();
+		ds.save(diagnosis);
+
+		assertNotNull("The saved diagnosis should have an id", diagnosis.getId());
+
+		Diagnosis newDiagnosis = ds.getDiagnosis(diagnosis.getId());
+		assertNotNull("We expect a diagnosis object to be returned", newDiagnosis);
```
wluyima: It's good to check that the uuid matches the expected value since uuids are unique like ids

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
pluwum: No, I did not. let me check it out.

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
pluwum: @wluyima I have changed this as recommended.

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
pluwum: It makes soo much sense now. I hadn't noticed the handler. I have removed them now.

api/src/test/java/org/openmrs/api/DiagnosisServiceTest.java
```diff
@@ -0,0 +1,74 @@
...
+	@Test
+	public void save_shouldUpdateDiagnosisSuccessfully() {
+		DiagnosisService ds = Context.getDiagnosisService();
+		Diagnosis diagnosis = ds.getDiagnosis(1);
+
+		Integer rank = diagnosis.getRank();
+		diagnosis.setRank(2);
+
+		ds.save(diagnosis);
+
```
dkayiwa: You would need to fetch the diagnosis again from the database, before asserting that the rank has changed.

api/src/test/java/org/openmrs/api/DiagnosisServiceTest.java
dkayiwa: You need to assert each property that you set to confirm it is as expected

api/src/main/java/org/openmrs/DiagnosisOrder.java
```diff
@@ -0,0 +1,27 @@
...
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs;
+
+/**
+ * The diagnosis order shows if a diagnosis is primary or secondary diagnosis
+ * 
+ * @since 2.2
+ */
+public enum DiagnosisOrder {
```
wluyima: Why can't this be declared inside the Diagnosis class?

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
```diff
@@ -0,0 +1,148 @@
...
+	 */
+	public Diagnosis save(Diagnosis diagnosis) {
+		return dao.saveDiagnosis(diagnosis);
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#voidDiagnosis(Diagnosis, String)
+	 */
+	public Diagnosis voidDiagnosis(Diagnosis diagnosis, String voidReason) {
+		return dao.saveDiagnosis(diagnosis);
```
wluyima: You should call `Context.getDiagnosisService().save()`, this ensures that the AOP around the proxy object is invoked

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
```diff
@@ -0,0 +1,148 @@
...
+		return diagnoses;
+	}
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#hasDiagnosis(Encounter, Diagnosis)
+	 */
+	public boolean hasDiagnosis(Encounter encounter, Diagnosis diagnosis) {
+
+		for (Diagnosis diagnosisFound :
+			encounter.getDiagnoses()) {
```
wluyima: Did you format your code? Actually did you run `mvn clean install`?

wluyima: Does `encounter.getDiagnoses()` only return non voided diagnoses? Because I don't think you'd want to return true if the found diagnosis is voided

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
```diff
@@ -0,0 +1,148 @@
...
+
+	/**
+	 * @see org.openmrs.api.DiagnosisService#getUniqueDiagnoses(Patient, Date)
+	 */
+	public List<Diagnosis> getUniqueDiagnoses(Patient patient, Date fromDate) {
+		List<Diagnosis> diagnoses = getDiagnoses(patient, fromDate);
+
+		Set<Diagnosis> uniqueDiagnoses = new HashSet<Diagnosis>();
+		uniqueDiagnoses.addAll(diagnoses);
+		diagnoses = new ArrayList(uniqueDiagnoses);
```
wluyima: You could replace all these lines to 3 like below:

```
Set<Diagnosis> uniqueDiagnoses = new HashSet<Diagnosis>();
uniqueDiagnoses.addAll(getDiagnoses(patient, fromDate)); //assuming getDiagnoses() never returns null
return new ArrayList(uniqueDiagnoses);
```

api/src/test/java/org/openmrs/api/DiagnosisServiceTest.java
wluyima: When you update data in openmrs, the voided related fields get updated too, you need to check that they've been set

api/src/test/java/org/openmrs/api/DiagnosisServiceTest.java
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	//TODO: implement voidDiagnosis_shouldVoidDiagnosis
+	//TODO: implement getDiagnosisByUuid_shouldGetDiagnosisByUuid
+	//TODO: implement getDiagnoses_shouldGetDiagnoses
+	//TODO: implement getPrimaryDiagnoses_shouldGetPrimaryDiagnoses
+	//TODO: implement hasDiagnosis test
+	//TODO: implement getUniqueDiagnoses test
+	//TODO: implement unvoidDiagnosis tests
+	//TODO: implement purgeDiagnosis_shouldPurgeDiaganosis
```
wluyima: Can you actually add these tests please?

api/src/main/java/org/openmrs/api/impl/DiagnosisServiceImpl.java
pluwum: @wluyima sorry I totally missed that.

api/src/test/java/org/openmrs/api/DiagnosisServiceTest.java
pluwum: @dkayiwa I am changing that right now. @wluyima Thanks for the tip, I am implementing that right away.

api/src/main/java/org/openmrs/Diagnosis.java
```diff
@@ -17,6 +17,23 @@
  */
 public class Diagnosis extends BaseChangeableOpenmrsData {
 
+	/**
+	 * The diagnosis order shows if a diagnosis is primary or secondary diagnosis
+	 *
+	 * @since 2.2
+	 */
+	public enum DiagnosisOrder {
```
djazayeri: This should not be included

The old implementation in the emrapi module has an enum for this, but now that we're implementing it openmrs-core we want to be more flexible, so the "Integer rank" field replaces diagnosisOrder.

See the parent JIRA issue.

pluwum: Thanks @djazayeri. Also, how can I determine if a diagnosis is primary or secondary?

api/src/test/java/org/openmrs/api/impl/ConditionServiceImplTest.java
```diff
@@ -30,7 +29,7 @@
  */
 public class ConditionServiceImplTest extends BaseContextSensitiveTest {
 	
-	protected static final String CONDITION_XML = "org/openmrs/api/include/ConditionServiceImplTest-SetupCondition.xml";
+	protected static final String DIAGNOSIS_XML = "org/openmrs/api/include/ConditionServiceImplTest-SetupCondition.xml";
```
paradisekelechi: @pluwum Can you revert changes in this file?

pluwum: @paradisekelechi sorry about that :)

api/src/test/java/org/openmrs/api/impl/DiagnosisServiceTest.java
```diff
@@ -0,0 +1,202 @@
...
+		}
+		executeDataSet(DIAGNOSIS_XML);
+	}
+
+	/**
+	 * Creates a diganosis object.
+	 *
+	 * @return diagnosis the diagnosis created
+	 */
+	private Diagnosis buildDiagnosis() {
```
paradisekelechi: @pluwum 
What is the aim of this method? To instantiate a Diagnosis Object, right? 
I don't think it is necessary. That is what you do when you executeDataSet().

api/src/test/resources/org/openmrs/api/include/DiagnosisServiceImplTest-SetupDiagnosis.xml
```diff
@@ -0,0 +1,27 @@
...
+    v. 2.0. If a copy of the MPL was not distributed with this file, You can
+    obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+    the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+
+    Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+    graphic logo is a trademark of OpenMRS Inc.
+
+-->
+<dataset>
+	<diagnosis diagnosis_id="1" patient_id="2" encounter_id="1" creator="1" date_created="2015-01-12 00:00:00"
```
paradisekelechi: @pluwum  I think this will throw. Are you sure the table name is `diagnosis`?
Also, the rank is required.

pluwum: @paradisekelechi fixed that. 🙇

pluwum:
@djazayeriI am still unable to figure out how to get this working for primary diagnosis in openmrs-core. Do you have some time to elaborate on a video call?

dkayiwa:
@pluwum i guess this was abandoned in favor of https://github.com/openmrs/openmrs-core/pull/2614

PR #2562, opened 2018-01-26
TRUNK-5189: Used Daemon User to create an Alert

samuelmale:
Issue : https://issues.openmrs.org/browse/TRUNK-5189.

This initial commit is not yet perfect, I just need your guidance @dkayiwa to complete this issue. 
Currently the alert is made by the daemon User, but there are 2 broken tests. In the [AlertServiceTest](https://github.com/openmrs/openmrs-core/blob/master/api/src/test/java/org/openmrs/notification/AlertServiceTest.java) but what has beat my understanding is that, when I comment out other tests and run one by one, they all pass fine. This has made me to anticipate that there are some thread conflicts. Could you please review my changes and direct me of what to do.
cc. @ssmusoke , @djazayeri etc.

PR #2615, opened 2018-03-07
TRUNK-4253:Made PatientService.savePatient threadsafe

samuelmale:
Ticket: https://issues.openmrs.org/browse/TRUNK-4253

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -59,6 +59,7 @@ import org.openmrs.util.PrivilegeConstants;
 import org.openmrs.validator.PatientIdentifierValidator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.transaction.annotation.Isolation;
 import org.springframework.transaction.annotation.Transactional;
 
 import java.util.ArrayList;
@@ -116,23 +117,23 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 	/**
 	 * @see org.openmrs.api.PatientService#savePatient(org.openmrs.Patient)
 	 */
+	@Transactional(isolation = Isolation.SERIALIZABLE)
 	@Override
 	public Patient savePatient(Patient patient) throws APIException {
-		requireAppropriatePatientModificationPrivilege(patient);
+		synchronized(this) {
+			requireAppropriatePatientModificationPrivilege(patient);
+			if (!patient.getVoided() && patient.getIdentifiers().size() == 1) {
+				patient.getPatientIdentifier().setPreferred(true);
+			} 
+			if (!patient.getVoided()) {
+				checkPatientIdentifiers(patient);
+			}
+			setPreferredPatientIdentifier(patient);
+			setPreferredPatientName(patient);
+			setPreferredPatientAddress(patient);
 
-		if (!patient.getVoided() && patient.getIdentifiers().size() == 1) {
-			patient.getPatientIdentifier().setPreferred(true);
+			return dao.savePatient(patient);
 		}
-
-		if (!patient.getVoided()) {
-			checkPatientIdentifiers(patient);
-		}
-
-		setPreferredPatientIdentifier(patient);
-		setPreferredPatientName(patient);
-		setPreferredPatientAddress(patient);
-
-		return dao.savePatient(patient);
 	}
 
 	private void requireAppropriatePatientModificationPrivilege(Patient patient) {
--- a/api/src/test/java/org/openmrs/api/PatientServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/PatientServiceTest.java
@@ -51,6 +51,7 @@ import org.openmrs.RelationshipType;
 import org.openmrs.User;
 import org.openmrs.Visit;
 import org.openmrs.api.context.Context;
+import org.openmrs.api.context.UserContext;
 import org.openmrs.api.impl.PatientServiceImpl;
 import org.openmrs.api.impl.PatientServiceImplTest;
 import org.openmrs.comparator.PatientIdentifierTypeDefaultComparator;
@@ -77,6 +78,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
 /**
@@ -116,6 +118,8 @@ public class PatientServiceTest extends BaseContextSensitiveTest {
 	protected static AdministrationService adminService = null;
 	
 	protected static LocationService locationService = null;
+	
+	private static AtomicBoolean isSavePatientThreadUnSafe;
 	
 	@Rule
 	public ExpectedException expectedException = ExpectedException.none();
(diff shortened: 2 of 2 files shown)


api/src/test/java/org/openmrs/api/PatientServiceTest.java
```diff
@@ -1848,6 +1853,48 @@ public void savePatient_shouldUpdateAnExistingPatient() throws Exception {
...
+		       
+		Runnable r2 = () -> savePatientOnThread(ctx, patient2);
+				
+		Thread t1 = new Thread(r1);
+		Thread t2 = new Thread(r2);
+		
+		t1.start();
+		t2.start();
+		
+		assertTrue(isSavePatientThreadsafe.get());
```
suthagar23: You should check for `assertFalse` since the default value is set to `false`.

api/src/test/java/org/openmrs/api/PatientServiceTest.java
```diff
@@ -1848,6 +1853,48 @@ public void savePatient_shouldUpdateAnExistingPatient() throws Exception {
 		assertTrue("The gender should be new", patient2.getGender().equals("F"));
 	}
 	
+	@Test
+	public void savePatient_shouldFailSaveIfPatientsWithSameIdentifierAreSaved() {
+	    isSavePatientThreadsafe = new AtomicBoolean(false);
```
suthagar23: Just one opinion,
`isSavePatientThreadsafe` would be changed to `true` by default or `isSavePatientThreadUnSafe = false` by default will good for the naming. Because at the beginning all threads are in safe state and it will become the critical stage while you are performing the saving method.

dkayiwa:
@samuelmale does the unit test fail when you remove the changes in PatientServiceImpl?

samuelmale:
Yes. The test fails @dkayiwa

api/src/test/java/org/openmrs/api/PatientServiceTest.java
```diff
@@ -1881,7 +1934,7 @@ public void getAllPatients_shouldFetchVoidedPatientsWhenGivenIncludeVoidedIsTrue
 		List<Patient> allPatients = patientService.getAllPatients(true);
 		// there are 1 voided and 4 nonvoided patients in
 		// standardTestDataset.xml
-		assertEquals(6, allPatients.size());
+		assertEquals(7, allPatients.size());
```
wluyima: Why is this necessary?

samuelmale: Hi, <br/>I have been facing this challenge for a while. Its all about Threads saving the data to the "in-memory-database" and effects aren't reverted as the default behavior is to other tests. Could I be missing something when using Threads?   
The above happens because of this [line](https://github.com/samuelmale/openmrs-core/blob/ce0953e35f0ddfaff07c738156cbcac386ab82af/api/src/test/java/org/openmrs/api/PatientServiceTest.java#L1876). Meaning the patient is saved the effect ain't reverted hence I had to change the number of patients from 6 to 7.  Its exactly the same issue in this thread. Could you please advise me on this?

@dkayiwa also asked me about it [here](https://github.com/openmrs/openmrs-core/pull/2562#discussion_r177600261).

wluyima: All I know is that changes in one thread aren't visible in others in tests since the transactions get rollback.

What I don't understand is why it's happening now and not before because the test was passing before your changes

samuelmale:
@wluyima  I have added some changes

api/src/test/java/org/openmrs/api/PatientServiceTest.java
```diff
@@ -1848,6 +1852,55 @@ public void savePatient_shouldUpdateAnExistingPatient() throws Exception {
...
+				
+		Thread t1 = new Thread(r1);
+		Thread t2 = new Thread(r2);
+		
+		t1.start();
+		t2.start();
+		
+		try {
+			t1.join();
+			//t2.join();
```
dkayiwa: Did you take a look at this? https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-Donotcommentoutcode

samuelmale:
Just removed the comment @dkayiwa

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -116,14 +117,16 @@ public void onShutdown() {
 	/**
 	 * @see org.openmrs.api.PatientService#savePatient(org.openmrs.Patient)
 	 */
+	@Transactional(isolation = Isolation.SERIALIZABLE)
 	@Override
 	public Patient savePatient(Patient patient) throws APIException {
+		synchronized(this) {
 		requireAppropriatePatientModificationPrivilege(patient);
```
dkayiwa: Did you intentionally decide not to TAB the code within the synchronised phrase?

api/src/test/java/org/openmrs/api/PatientServiceTest.java
```diff
@@ -1848,6 +1852,56 @@ public void savePatient_shouldUpdateAnExistingPatient() throws Exception {
...
+		}	
+		
+		assertTrue("IdentifierNotUniqueException not thrown, method not Threadsafe", isSavePatientThreadsafe.get());
+		
+		//Cleans up the unrolled back Thread transactions.
+		deleteAllData();
+	}
+	public void savePatientOnThread(UserContext ctx, Patient patient) {
+		Context.setUserContext(ctx);
+        Context.openSessionWithCurrentUser();
```
dkayiwa: Can you also fix the above indention?

samuelmale: I wonder, but locally, the indention is fine. Just to prove, hit view full file and check.

I have highlighted the it here https://github.com/samuelmale/openmrs-core/blob/fd2aed2031b647f046562da4142851669c3c4061/api/src/test/java/org/openmrs/api/PatientServiceTest.java#L1890-L1894

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -116,23 +117,23 @@ public void onShutdown() {
...
-		}
-
-		if (!patient.getVoided()) {
-			checkPatientIdentifiers(patient);
-		}
-
-		setPreferredPatientIdentifier(patient);
-		setPreferredPatientName(patient);
-		setPreferredPatientAddress(patient);
+		synchronized(this) {
```
dkayiwa: Do you have a reason for using a synchronised block instead of method?

api/src/test/java/org/openmrs/api/PatientServiceTest.java
dkayiwa: @samuelmale did you see the above?

samuelmale:
I'm of the view that we use synchronized blocks, a sync'd method makes it not thread safe @dkayiwa 

check the CI failure

dkayiwa:
Can you share the link where you read that?

samuelmale:
I have just run 'mvn clean install' on this branch twice and it builds. I wonder whats happening with CI

@dkayiwa

dkayiwa:
I have restarted the build

api/src/test/java/org/openmrs/api/PatientServiceTest.java
dkayiwa: @samuelmale did you see the above?

samuelmale:
Cool, just waiting for yr review @dkayiwa

api/src/test/java/org/openmrs/api/PatientServiceTest.java
samuelmale: If this is the unaddressed comment, I'm pretty sure my indention is fine. Have you looked at it as a file. In otherwise have looked at the above link? @dkayiwa

samuelmale: If this is the unaddressed comment, I'm pretty sure my indention is fine. Have you looked at it as a file. In otherwise have looked at the above link? @dkayiwa

samuelmale:
Could you restart the CI build once more @dkayiwa ?

dkayiwa:
Even compiling it locally on your machine should fail too.

samuelmale:
Noo, it actually passes @Kayiwa Daniel <danielkayiwa@gmail.com>

dkayiwa:
@samuelmale i have restarted it several times and noticed that some times it fails while other times it passed. Whenever it fails, the error message is as below:

Results :
Failed tests: 
  PatientServiceTest.getAllPatients_shouldFetchVoidedPatientsWhenGivenIncludeVoidedIsTrue:1938 expected:<6> but was:<7>
  PatientServiceTest.savePatient_shouldFailSaveIfPatientsWithSameIdentifierAreSaved:1884 IdentifierNotUniqueException not thrown, method not Threadsafe
Tests run: 4022, Failures: 2, Errors: 0, Skipped: 36

dkayiwa:
I just run it another time and got this:

Results :
Failed tests: 
  PatientServiceTest.getAllPatients_shouldFetchVoidedPatientsWhenGivenIncludeVoidedIsTrue:1938 expected:<6> but was:<8>
  PatientServiceTest.savePatient_shouldFailSaveIfPatientsWithSameIdentifierAreSaved:1884 IdentifierNotUniqueException not thrown, method not Threadsafe

samuelmale:
Do you mind fetching this branch and testing it out locally?

dkayiwa:
Like i said, it is random. I have just rerun it and now you can see the failure.

samuelmale:
So whats really going on, but locally its not random @dkayiwa

dkayiwa:
Tru dat!
Looks like it depends on the processing power.

samuelmale:
So whats up @dkayiwa

dkayiwa:
@samuelmale the very fact that the test randomly fails in some instances, makes me feel like it needs some more work.

samuelmale:
This weird behaviour started when I changed from a blocked synchronized block to a method synchronized block @dkayiwa . Do we have reasons as to why we can't use the block synchronization?

dkayiwa:
I remember as if it still failed even with block synchronisation. Any way, you can change it back and will show you by simply running travis again.

samuelmale:
Hahaha okie @dkayiwa

samuelmale:
Whats going on here @dkayiwa .
I has failed again. Could you restart travis?

samuelmale:
ahhhhh!!!! But have you ever hit upon such a case? @dkayiwa

suthagar23:
@samuelmale any more ideas on the above Travis failures?

bejelith:
I think ur problem is that savePatient is not thread safe and you have no idea if ur executing inside an pre-existing transaction. Create a separate hibernate transaction/session and should work

PR #2618, opened 2018-03-10
TRUNK-5065: Add tests to AdministrationServiceImpl

PKatGITHUB:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5065

  No? -> execute above command


api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -12,31 +12,27 @@
...
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
-import java.util.Locale;
-import java.util.Set;
+import java.util.*;
```
teleivo: @PKatGITHUB can you please configure your IDE not to use star imports
see https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -74,25 +69,56 @@ public void runBeforeEachTest() {
 			adminService.setImplementationIdHttpClient(implementationHttpClient);
 			cacheManager = Context.getRegisteredComponent("apiCacheManager", CacheManager.class);
 		}
-		
+
+	}
+
+	/**
+	 * @see AdministrationService#getSystemInformation()
+	 */
```
teleivo: please remove all javadocs above your tests, they just clutter the test class without providing much information. we see what class/method you tested in the test.

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -74,25 +69,56 @@ public void runBeforeEachTest() {
...
+	@Test
+	public void getSystemInformation_shouldReturnSystemInformation() {
+		Map<String, Map<String, String>> systemInfoMap = Context.getAdministrationService().getSystemInformation();
+		assertEquals(5, systemInfoMap.size());
```
teleivo: is there something more specific we can assert on in addition? what is it that you expect to be in this systemInformation map?

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -74,25 +69,56 @@ public void runBeforeEachTest() {
...
+
+	/**
+	 * @see AdministrationService#getSystemInformation()
+	 */
+	@Test
+	public void getSystemInformation_shouldAddAllLoadedModulesInformation() {
+		Module module = new Module("name", "moduleId", "packageName", "author", "description", "version");
+		ModuleFactory.loadModule(module, false);
+		int numberOfModules = ModuleFactory.getLoadedModules().size();
+		Map<String, Map<String, String>> systemInfoMap = Context.getAdministrationService().getSystemInformation();
```
teleivo: can you please add newlines in between, like

```java
int numberOfModules = ModuleFactory.getLoadedModules().size();

 Map<String, Map<String, String>> systemInfoMap = Context.getAdministrationService().getSystemInformation(); 

assertEquals(...
```

This arranges your test in AAA, groups of arrange (what you need to setup your test), act (the class/method under test), assert (what do you expect to happen after calling the method).

It makes your tests more readable since I can easily spot what you are testing which is also why the javadoc is not needed.

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -74,25 +69,56 @@ public void runBeforeEachTest() {
...
+		assertEquals(numberOfModules + 1, systemInfoMap.get("SystemInfo.title.moduleInformation").size());
```
teleivo: you could assert that the module that you loaded is in there.

at the moment you assert that the number matches what you expect but it could as well have been the same number before you loaded the module. You did not check that.

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -74,25 +69,56 @@ public void runBeforeEachTest() {
...
 	}
 	
+	/**
+	 * @see AdministrationService#removeGlobalPropertyListener(GlobalPropertyListener)
+	 */
+	@Test
+	public void removeGlobalPropertyListener_shouldNotFailOnNullListener() {
+		GlobalPropertyListener globalPropertyListener = null;
+		assertNull(globalPropertyListener);
```
teleivo: no need to assert its null, you set it to null the line above.

you can simply inline this temp variable to

```java
adminService.removeGlobalPropertyListener(null)
```

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -74,25 +69,56 @@ public void runBeforeEachTest() {
...
+		adminService.removeGlobalPropertyListener(globalPropertyListener);
+	}
```
teleivo: can you also please add a test for when the globalPropertyListener is not null

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -74,25 +69,56 @@ public void runBeforeEachTest() {
...
+
 	/**
 	 * Tests the AdministrationService.executeSql method with a sql statement containing a valid
 	 * group by clause
-	 * 
-	 * @see AdministrationService#executeSQL(String,null)
+	 *
+	 * @see AdministrationService#executeSQL(String, boolean)
```
teleivo: please remove these javadocs (since :) they clutter)

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -74,25 +69,56 @@ public void runBeforeEachTest() {
...
 	 * Tests the AdministrationService.executeSql method with a sql statement containing a valid
 	 * group by clause
-	 * 
-	 * @see AdministrationService#executeSQL(String,null)
+	 *
+	 * @see AdministrationService#executeSQL(String, boolean) 
 	 */
 	@Test
 	public void executeSQL_shouldExecuteSqlContainingGroupBy() {
-
```
teleivo: and please check your IDE setup https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE
are you inserting spaces (we use tabs)? or removing the tabs here?

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -168,119 +194,127 @@ public void setImplementationId_shouldThrowAPIExceptionIfGivenAPipeInTheImplemen
...
 		validId.setDescription("This is the junit impl id used for testing of the openmrs API only.");
-		adminService.setImplementationId(validId);
-		
-		assertEquals(validId, adminService.getImplementationId());
+		try {
+			Context.getAdministrationService().setImplementationId(validId);
+			assertNotNull(Context.getAdministrationService().getImplementationId());
+		}
+		catch (APIException e) {
+			//expected exception
```
teleivo: if you expect an exception to happen either put the exception in `@Test(expected=APIException.class)` or if this specific expection is thrown multiple times in the class and you cannot ensure the specific case with this style than use https://github.com/junit-team/junit4/wiki/Rules#expectedexception-rules where you can also assert on the exception message.

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -168,119 +194,127 @@ public void setImplementationId_shouldThrowAPIExceptionIfGivenAPipeInTheImplemen
...
 		ImplementationId validId2 = new ImplementationId();
-		validId2.setImplementationId("JUNIT-TEST 2");
+		validId2.setImplementationId("JUNIT-TEST");
 		validId2.setName("JUNIT-TEST (#2) implementation id");
 		validId2.setPassphrase("This is the junit test passphrase 2");
 		validId2.setDescription("This is the junit impl id (2) used for testing of the openmrs API only.");
-		adminService.setImplementationId(validId2);
-		assertEquals(validId2, adminService.getImplementationId());
+		try {
+			adminService.setImplementationId(validId2);
```
teleivo: same as above

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -168,119 +194,127 @@ public void setImplementationId_shouldThrowAPIExceptionIfGivenAPipeInTheImplemen
...
 		validId.setName("JUNIT-TEST implementation id");
 		validId.setPassphrase("This is the junit test passphrase");
 		validId.setDescription("This is the junit impl id used for testing of the openmrs API only.");
-		adminService.setImplementationId(validId);
-		
+		try {
+			adminService.setImplementationId(validId);
+		}
+		catch (APIException e) {
+			//expected exception
```
teleivo: same here.
please search for other tests that use junits `@Rule` with exceptions

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -168,119 +194,127 @@ public void setImplementationId_shouldThrowAPIExceptionIfGivenAPipeInTheImplemen
...
 	 */
 	@Test
 	public void getGlobalProperty_shouldNotFailWithNullDefaultValue() {
 		adminService.getGlobalProperty("asdfsadfsafd", null);
 	}
-	
+
 	/**
-	 * @see AdministrationService#updateGlobalProperty(String,String)
+	 * @see AdministrationService#updateGlobalProperty(String, String)
```
teleivo: in the places you already change code, please remove all javadocs

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -289,218 +323,219 @@ public void updateGlobalProperty_shouldUpdateAGlobalPropertyWhoseTypedvalueIsHan
...
+
+	/**
+	 * @see AdministrationService#getAllowedLocales()
+	 */
+	@Test
+	public void getAllowedLocales_shouldNotFailIfGlobalLocaleListIsNull() {
+		AdministrationServiceImpl as = new AdministrationServiceImpl();
+		as.setGlobalLocaleList(null);
+		assertNotNull(adminService.getAllowedLocales());
+
```
teleivo: remove newline at end and arrange in AAA style as explained

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -289,218 +323,219 @@ public void updateGlobalProperty_shouldUpdateAGlobalPropertyWhoseTypedvalueIsHan
...
-	
+
+	/**
+	 * @see AdministrationService#getAllowedLocales()
+	 */
+	@Test
+	public void getAllowedLocales_shouldNotFailIfGlobalLocaleListIsNull() {
+		AdministrationServiceImpl as = new AdministrationServiceImpl();
+		as.setGlobalLocaleList(null);
+		assertNotNull(adminService.getAllowedLocales());
```
teleivo: so you know what locales you would expect?

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -289,218 +323,219 @@ public void updateGlobalProperty_shouldUpdateAGlobalPropertyWhoseTypedvalueIsHan
...
-	}
-	
-	/**
-	 * @see AdministrationService#getPresentationLocales()
-	 */
-	@Test
-	@Ignore
-	//TODO: This test fails for some reason
-	public void getPresentationLocales_shouldReturnAtLeastOneLocaleIfNoLocalesDefinedInDatabaseYet() {
-		Assert.assertTrue(Context.getAdministrationService().getPresentationLocales().size() > 0);
```
teleivo: you deleted the test. did you try to fix it?

PKatGITHUB:
Hey @teleivo thank you for the feedback. I will sure be working on the points you mentioned. 

And sorry for the late response, I was busy writing my proposal for GSoC-2018 these past days :)

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
PKatGITHUB: Hey @teleivo, Coveralls indicate that the lines are covered. Can you please elaborate why more assertions are needed?
Thank you :)

PKatGITHUB: Do we need to assert each member which the map contains?

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -74,25 +69,56 @@ public void runBeforeEachTest() {
...
-		
+
+	}
+
+	/**
+	 * @see AdministrationService#getSystemInformation()
+	 */
+	@Test
+	public void getSystemInformation_shouldReturnSystemInformation() {
+		Map<String, Map<String, String>> systemInfoMap = Context.getAdministrationService().getSystemInformation();
```
teleivo: please just use `adminService`, its already set see `@Before` method, instead of getting it from the context in your tests.

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
teleivo: you only assert that the map contains 5 elements, you do not know whats inside. your test might pass by accident. maybe now it is not an accident and really does contain the elements you expect but if we refactor the AdminService in the future we might break our promises without noticing it due to an assertion that is not specific enough.

PKatGITHUB: Ok. :) 
Thank you!

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
PKatGITHUB: should I remove all the javadocs?? or only for the tests I wrote??

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
PKatGITHUB: Hey @teleivo , I have got a problem here. I think when running the tests when I am calling the function setImplementationId() , You can see that this function internally calls another function checkImplementationIdValidity() , see [HERE](https://github.com/openmrs/openmrs-core/blob/6966e508c199e3b808718d155373fa695a961ee9/api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java#L399) !! 

And this function always fails resulting in the execution of the catch block. Now I am not actually able to find out why the function checkImplementationIdValidity() is failing. Can you please help??

PKatGITHUB: I think thats why the try block below the function call remains uncovered :)

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
PKatGITHUB: I saw that the code was covered. Can you please explain why this test is relevant?

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
teleivo: you can remove the javadocs on the tests you write or change and leave the others as is

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
teleivo: this test to me seems to be the "happy path" so when setting the implementation id actually works, however there is an exception thrown my guess is because the id is not valid according to `checkImplementationIdValidity`. What is the exception?

the solution is not to catch it but to fix whats wrong in setting the implementation id.

the other tests for setImplementation id are not a good example, they should be changed to using
https://github.com/junit-team/junit4/wiki/Rules#expectedexception-rules an example of where we us it is here
https://github.com/openmrs/openmrs-core/blob/master/api/src/test/java/org/openmrs/OrderTest.java#L524-L526

PKatGITHUB: thank you @teleivo ! :) will do the changes.

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -74,25 +93,128 @@ public void runBeforeEachTest() {
...
-	 * @see AdministrationService#executeSQL(String,null)
-	 */
+	@Test
+	public void removeGlobalPropertyListener_shouldNotFailOnNullListener() {
+		adminService.removeGlobalPropertyListener(null);
+	}
+
+	@Test
+	public void removeGlobalPropertyListener_shouldNotFailIfListenerNotNull() {
+		executeDataSet(ADMIN_INITIAL_DATA_XML);
```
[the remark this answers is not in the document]
PKatGITHUB: Hey @teleivo thanks for your support and appreciation. So you want me leave this pull request here only, start off with a new branch with a new pull request adding the test you told?

PKatGITHUB:
@teleivo  added new PR. See https://github.com/openmrs/openmrs-core/pull/2648

PR #2636, opened 2018-03-22
TRUNK-5093 - AbstractHandler allows blind rewriting of existing compl…

FloBue:
JIRA: TRUNK-5093 - AbstractHandler allows blind rewriting of existing complex obs file
Link: https://issues.openmrs.org/browse/TRUNK-5093

I modified the AbstractHandler to concatenate the UUID to each complex obs filename, so that no collisions happen and no files get lost. I also added a JUnit test file AbstractHandlerTest to check that the previous error of overwritten/lost files does not occur and the files are named properly and modified an existing test in ObsServiceTest to work with the new naming scheme.

Not directly related to this ticket: As I previously worked on the Complex Obs Handlers tests for my last ticket, I went back to them and improved some things. The main point is a more reliable file clean-up through the use of temporary folders. I also got rid of unnecessary set-up methods for variables and reduced the scope of some variables, which I previously used more often globally but forgot to declare only locally after changes in the code.

### DIFF
--- a/api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
+++ b/api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
@@ -12,10 +12,9 @@ package org.openmrs.obs.handler;
 import java.io.File;
 import java.io.IOException;
 import java.text.NumberFormat;
-import java.text.SimpleDateFormat;
 import java.util.Arrays;
-import java.util.Date;
 
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.openmrs.Obs;
 import org.openmrs.api.context.Context;
@@ -37,8 +36,6 @@ public class AbstractHandler {
 	
 	protected NumberFormat nf;
 	
-	protected SimpleDateFormat longfmt;
-	
 	/**
 	 * Constructor initializes formats for alternative file names to protect from unintentionally
 	 * overwriting existing files.
@@ -47,7 +44,6 @@ public class AbstractHandler {
 		nf = NumberFormat.getInstance();
 		nf.setMaximumFractionDigits(0);
 		nf.setMinimumIntegerDigits(2);
-		longfmt = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
 	}
 	
 	/**
@@ -59,65 +55,51 @@ public class AbstractHandler {
 	 * @return File that the complex data should be written to
 	 */
 	public File getOutputFileToWrite(Obs obs) throws IOException {
-		// Get the title and remove the extension.
 		String title = obs.getComplexData().getTitle();
-		String extension = "." + getExtension(title);
+		String titleWithoutExtension = FilenameUtils.removeExtension(title);
+		String extension = "." + StringUtils.defaultIfEmpty(FilenameUtils.getExtension(title), "dat");
+		String uuid = obs.getUuid();
+		String filename;
 		
-		// If getExtension returns the title, there was no extension
-		if (getExtension(title).equals(title)) {
-			extension = "";
-		}
-		
-		File dir = OpenmrsUtil.getDirectoryInApplicationDataDirectory(Context.getAdministrationService().getGlobalProperty(
-		    OpenmrsConstants.GLOBAL_PROPERTY_COMPLEX_OBS_DIR));
-		File outputfile;
-		
-		// Get the output stream
-		if (null == title) {
-			String now = longfmt.format(new Date());
-			outputfile = new File(dir, now);
+		if (StringUtils.isNotBlank(titleWithoutExtension)) {
+			filename = titleWithoutExtension + "_" + uuid + extension;
 		} else {
-			title = title.replace(extension, "");
-			
-			outputfile = new File(dir, title + extension);
+			filename = uuid + extension;
 		}
 		
-		int i = 0;
-		String tmp;
-		
-		// If the Obs does not exist, but the File does, append a two-digit
-		// count number to the filename and save it.
-		while (obs.getObsId() == null && outputfile.exists() && i < 100) {
-			// Remove the extension from the filename.
-			tmp = String.valueOf(outputfile.getAbsolutePath().replace(extension, ""));
-			// Append two-digit count number to the filename.
-			String filename = (i < 1) ? tmp + "_" + nf.format(Integer.valueOf(++i)) : tmp.replace(nf.format(Integer
-			        .valueOf(i)), nf.format(Integer.valueOf(++i)));
-			// Append the extension to the filename
-			outputfile = new File(filename + extension);
-		}
+		File dir = OpenmrsUtil.getDirectoryInApplicationDataDirectory(
+		    Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_COMPLEX_OBS_DIR));
+		File outputfile = new File(dir, filename);
 		
 		return outputfile;
-		
 	}
 	
 	/**
-	 * Get the extension for a given filename. <br>
-	 * If given "asdf.jpg", will return "jpg". <br>
-	 * If given "asdf", will return "asdf". <br>
+	 * Get the extension for a given filename if it exists, else return the filename. If there is no
+	 * filename in the input string, "raw" is returned. 
+	 * 
+	 * If given "asdf.jpg", will return "jpg".
+	 * If given "asdf", will return "asdf". 
+	 * If given "" or "a/b/c/" will return "raw".
 	 * 
 	 * @param filename
-	 * @return the filepart after the period in the given filename
+	 * @return the part after the period in the given filename, the filename, or "raw"
+	 * @deprecated since 2.1.3 use {@link org.apache.commons.io.FilenameUtils#getExtension(String)}
+	 *             instead.
 	 */
+	@Deprecated
 	public String getExtension(String filename) {
-		String[] filenameParts = filename.split("\\.");
+		String result = FilenameUtils.getExtension(filename);
 		
-		log.debug("titles length: " + filenameParts.length);
+		if (StringUtils.isEmpty(result)) {
+			result = FilenameUtils.getBaseName(filename);
+			
+			if (StringUtils.isEmpty(result)) {
+				result = "raw";
+			}
+		}
 		
-		String extension = (filenameParts.length < 2) ? filenameParts[0] : filenameParts[filenameParts.length - 1];
-		extension = StringUtils.isNotEmpty(extension) ? extension : "raw";
-		
-		return extension;
+		return result;
 	}
 	
 	/**
@@ -153,8 +135,8 @@ public class AbstractHandler {
 			return true;
 		}
 		
-		log.warn("Could not delete complex data object for obsId=" + obs.getObsId() + " located at "
-		        + file.getAbsolutePath());
+		log.warn(
+		    "Could not delete complex data object for obsId=" + obs.getObsId() + " located at " + file.getAbsolutePath());
 		return false;
 	}
 	
@@ -167,8 +149,8 @@ public class AbstractHandler {
 	public static File getComplexDataFile(Obs obs) {
 		String[] names = obs.getValueComplex().split("\\|");
 		String filename = names.length < 2 ? names[0] : names[names.length - 1];
-		File dir = OpenmrsUtil.getDirectoryInApplicationDataDirectory(Context.getAdministrationService().getGlobalProperty(
-		    OpenmrsConstants.GLOBAL_PROPERTY_COMPLEX_OBS_DIR));
+		File dir = OpenmrsUtil.getDirectoryInApplicationDataDirectory(
+		    Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_COMPLEX_OBS_DIR));
 		return new File(dir, filename);
 	}
 	
--- a/api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
+++ b/api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
@@ -1,0 +1,166 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.obs;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.mockito.Matchers.any;
+import static org.powermock.api.mockito.PowerMockito.mockStatic;
+import static org.powermock.api.mockito.PowerMockito.when;
+
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.nio.charset.StandardCharsets;
+import java.text.ParseException;
+
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.FilenameUtils;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.openmrs.Obs;
+import org.openmrs.api.APIException;
+import org.openmrs.api.AdministrationService;
+import org.openmrs.api.context.Context;
+import org.openmrs.obs.handler.AbstractHandler;
+import org.openmrs.util.OpenmrsUtil;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({ AbstractHandler.class, OpenmrsUtil.class, Context.class })
+
+public class AbstractHandlerTest {
+	
+	private final String FILENAME = "mytxtfile.txt";
+	
+	private final AbstractHandler handler = new AbstractHandler();
+	
+	@Mock
+	private AdministrationService administrationService;
+	
+	@Rule
+	public TemporaryFolder complexObsTestFolder = new TemporaryFolder();
+	
+	@Before
+	public void initializeContext() throws APIException, IOException {
+		mockStatic(Context.class);
+		when(Context.getAdministrationService()).thenReturn(administrationService);
+		when(administrationService.getGlobalProperty(any())).thenReturn(complexObsTestFolder.newFolder().getAbsolutePath());
+	}
+	
+	@Test
+	public void getOutputFileToWrite_shouldNeverOverwritePreviousFiles() throws IOException {
+		String content1 = "A";
+		String content2 = "B";
+		
+		File previousFile = null;
+		File currentFile = null;
+		
+		for (int i = 0; i <= 101; i++) {
+			String currentData = (i % 2 == 0) ? content1 : content2;
+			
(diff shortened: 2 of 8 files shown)


FloBue:
Addressed the Codacy complaints. 

I can't quite wrap my head around the small coverage decrease, as my tests reach every branch of the method in question. Had a similar issue with a previous PR where coveralls erroneously marked lines as untested, but I haven't worked with it before and I'm not sure why this happens.

dkayiwa:
https://talk.openmrs.org/t/test-coverage-reduced-after-adding-unit-tests/17034

FloBue:
@dkayiwa Ah, good to know I'm not the only one that's been confused :D. Thanks for the heads-up.

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -25,6 +25,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import liquibase.util.file.FilenameUtils;
```
mks-d: This is the wrong `FilenameUtils`, you should use the Apache one.

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -59,46 +61,30 @@ public AbstractHandler() {
...
+		// Generate the filename with the Obs' UUID
+		String uuid = obs.getUuid();
 		if (null == title) {
 			String now = longfmt.format(new Date());
-			outputfile = new File(dir, now);
+			outputfile = new File(dir, now + '_' + uuid);
 		} else {
-			title = title.replace(extension, "");
+			// If the extension is no empty string, it exists and we can add a dot in front for later use
+			if (!(extension.equals(""))) {
```
mks-d: I think it is always safer to do the reverse (just an hedge against NPE):
```java
if (!("".equals(extension)))
```
Or simply
```java
if (!StringUtils.isEmpty(extension))
```

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -59,46 +61,30 @@ public AbstractHandler() {
...
-		// count number to the filename and save it.
-		while (obs.getObsId() == null && outputfile.exists() && i < 100) {
-			// Remove the extension from the filename.
-			tmp = String.valueOf(outputfile.getAbsolutePath().replace(extension, ""));
-			// Append two-digit count number to the filename.
-			String filename = (i < 1) ? tmp + "_" + nf.format(Integer.valueOf(++i)) : tmp.replace(nf.format(Integer
-			        .valueOf(i)), nf.format(Integer.valueOf(++i)));
-			// Append the extension to the filename
-			outputfile = new File(filename + extension);
+			String titleWithoutExtension = title.replace(extension, "");
```
mks-d: I would have bet there an API to do things like that in (Apache's) `FilenameUtils`.
Just in case the actual string `extension` exists multiple time in the file. Imagine a:
```
my_txt_file.txt
```

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
```diff
@@ -0,0 +1,168 @@
...
+import org.openmrs.util.OpenmrsUtil;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({ AbstractHandler.class, OpenmrsUtil.class, Context.class })
+
+public class AbstractHandlerTest {
+	
+	private final String FILENAME = "TestingComplexObsNaming.txt";
```
mks-d: I guess that'd be the place to attempt a swap with `FILENAME = "my_txt_file.txt"` ;-)

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
teleivo: the latter one is safer when you want to exclude inputs like `"    "`

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
FloBue: I could have sworn I used Apache. I suppose I selected the wrong one accidentally in Eclipse when it asked me and did not notice.

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
FloBue: This particular example would not be a problem I think, as I did add a dot to the extension String beforehand. But you're right in that there could probably be malformed names an API could maybe handle better. I will look for an appropriate method.

FloBue:
@mks-d I'll be tackling your remarks tomorrow, thanks for the input.

FloBue:
@mks-d I pushed the changes. Let me know if there is anything else.

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
```diff
@@ -0,0 +1,168 @@
...
+	private final String FILENAME = "mytxtfile.txt";
+	
+	@Mock
+	private AdministrationService administrationService;
+	
+	@Rule
+	public TemporaryFolder complexObsTestFolder = new TemporaryFolder();
+	
+	@Test
+	public void shouldNeverOverwritePreviousFiles() throws IOException {
```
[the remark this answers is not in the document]
FloBue: @mks-d I named it the same way the other methods in the complex obs handler classes were named (the ones I didn't create) because I didn't want it to look inconsistent. Although, now that you say it, maybe these didn't have the test method name attached to it because they were not testing a particular method?

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -59,46 +60,30 @@ public AbstractHandler() {
...
-		// Get the title and remove the extension.
+		// Get the title and the extension
 		String title = obs.getComplexData().getTitle();
-		String extension = "." + getExtension(title);
-		
-		// If getExtension returns the title, there was no extension
-		if (getExtension(title).equals(title)) {
-			extension = "";
-		}
+		String extension = FilenameUtils.getExtension(title);
```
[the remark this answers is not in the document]
FloBue: @mks-d I'm not sure where else it is used either and I didn't want to break anything that relies on it. This method's return behaviour is different, so if any other method relied on this one, it could lead to bigger problems. I'd rather adjust code that uses the old method than "emulate" its behaviour with FilenameUtils, though I don't know if that's feasible. Can modules access these methods, or would dependencies be isolated to openmrs-core and therefore easy to find?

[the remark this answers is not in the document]
mks-d: >"emulate" its behaviour with FilenameUtils

This sounds like a great idea ;-) You know by now that I love `FilenameUtils`!

I think it is unlikely albeit possible that modules rely on this specific method. So rather be safe than sorry indeed.

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
```diff
@@ -0,0 +1,168 @@
...
+import org.openmrs.Obs;
+import org.openmrs.api.AdministrationService;
+import org.openmrs.api.context.Context;
+import org.openmrs.obs.handler.AbstractHandler;
+import org.openmrs.util.OpenmrsUtil;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({ AbstractHandler.class, OpenmrsUtil.class, Context.class })
```
teleivo: Is it possible to write the test without PowerMockRunner? I know that our extensive use of static methods sometimes makes this necessary but we should try to avoid this as much as possible. one annoying reason is that tests using PowerMockRunner do not show up in our coverage which is a known problem https://issues.openmrs.org/browse/TRUNK-5150 ; not sure if there is a solution yet. the reason number one is that we bury dependencies in our code using Context.getXXXService, which makes it hard to test/exchange different implementations.

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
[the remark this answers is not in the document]
teleivo: we do still write tests the way you described it `methodName_shouldDoThisOrThat` but if the test class only tests one method I don't think you have to (unless we enforce it somewhere, I believe the `should` on the method name is enforced somewhere in our build but I am unsure, you would see that once you run `mvn clean package` I guess). (for background info on our test style evolution see  https://talk.openmrs.org/t/rethink-unit-test-style-to-remove-duplication/10200 😄 )

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
FloBue: @teleivo I had the same concerns when I wrote the tests and already invested some time to try and find a way to make it work without PowerMock. I didn't find a good solution, but I'm not saying there isn't one. I'm open for suggestions! :P

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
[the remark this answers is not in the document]
FloBue: Yeah ;D. Alright, I'll rewrite the old getExtension tomorrow to provide the same returns for the same inputs but internally use FilenameUtils.

FloBue:
@mks-d Pushed a new version, I hope everything looks good now :).

mks-d:
@FloBue you are doing a great job. And as I can see that you're coming with valuable input I allow myself to dig deeper to make sure that we nail this "forever". For the record the original implementation has lead to a production system nightmare scenario (see [here](https://talk.openmrs.org/t/core-cannot-purge-anymore-a-complex-obs-that-has-lost-its-complex-data/10254/9?u=mksd) if you're interested.)

See also my last comment on JIRA.

FloBue:
@mks-d New push with all your proposed changes. I also extended the test for AbstractHandler#getOutputFileToWrite() to include the new case where a title exists but no extension (e.g. "title", which now becomes title.dat").

On a sidenote: I removed all the Date-related imports and objects from AbstractHandler, but not the NumberFormat member. I'm not sure if it does anything, but the Date did not use it either if I saw it correctly.

mks-d:
@FloBue two things: (1) thanks for the heads up, (2) thanks for the great and very useful work on this ticket!

No more comments from my end, @teleivo / @dkayiwa? Can we merge this in?

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -108,16 +89,21 @@ public File getOutputFileToWrite(Obs obs) throws IOException {
 	 * 
 	 * @param filename
 	 * @return the filepart after the period in the given filename
+	 * @deprecated use {@link org.apache.commons.io.FilenameUtils#getExtension(String)} instead.
```
teleivo: can you add since when its `@deprecated since x.x.x use ...`

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -59,46 +55,31 @@ public AbstractHandler() {
 	 * @return File that the complex data should be written to
 	 */
 	public File getOutputFileToWrite(Obs obs) throws IOException {
-		// Get the title and remove the extension.
+		// Get the title and the extension
```
teleivo: please remove comments, your code is nicely readable ;)

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
teleivo: make it a BaseContextSensitiveTest thats the only way it has the Context setup and will add to our coverage.

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -59,46 +55,31 @@ public AbstractHandler() {
...
-		File dir = OpenmrsUtil.getDirectoryInApplicationDataDirectory(Context.getAdministrationService().getGlobalProperty(
-		    OpenmrsConstants.GLOBAL_PROPERTY_COMPLEX_OBS_DIR));
-		File outputfile;
+		File dir = OpenmrsUtil.getDirectoryInApplicationDataDirectory(
+		    Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_COMPLEX_OBS_DIR));
+		File outputfile = null;
 		
-		// Get the output stream
+		// Generate the filename with the Obs' UUID
+		String uuid = obs.getUuid();
```
teleivo: @FloBue nice work! I love it when we remove so much code and deprecate methods :)

I have a suggestion I find a bit cleaner, what do you say?

```java
File dir = ...
String uuid = obs.getUuid();
// should also catch title being null since getExtension(null) returns null but `defaultIfEmpty(null, "dat")` returns "dat"
String extension = "." + StringUtils.defaultIfEmpty(FilenameUtils.getExtension(title), "dat");
String filename;
String titleWitoutExtension = FilenameUtils.removeExtension(title);
if (StringUtils.isNotBlank(titleWitoutExtension) {
  filename =  titleWitoutExtension + "_" + uuid + extension;
} else {
  filename =  uuid + extension;
}
File outputfile = new File(dir, filename);
```

teleivo: one thing I dont know which concerns your/my implementation is: can a title be like "a\b\c.jpg" ? removeExtension says it makes "a\b\c.jpg  --> a\b\c" but there is also `getBaseName` which handles such cases differently

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
```diff
@@ -0,0 +1,177 @@
...
+			currentFile = handler.getOutputFileToWrite(obs);
+			
+			// Write the file
+			try (BufferedWriter fout = new BufferedWriter(
+			        new OutputStreamWriter(new FileOutputStream(currentFile), StandardCharsets.UTF_8))) {
+				fout.write(currentData);
+			}
+			
+			// Read the file to check if it wrote properly
+			try (BufferedReader fin = new BufferedReader(
```
teleivo: can we make use of apache commons-io `FileUtils.contentEquals` ?

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
```diff
@@ -0,0 +1,177 @@
...
+		
+		// Mocked methods
+		mockStatic(Context.class);
+		when(Context.getAdministrationService()).thenReturn(administrationService);
+		when(administrationService.getGlobalProperty(any())).thenReturn(complexObsTestFolder.newFolder().getAbsolutePath());
+		
+		AbstractHandler handler = new AbstractHandler();
+		
+		// Create files	
+		File titledFile = handler.getOutputFileToWrite(obsWithTitle);
```
teleivo: am I correct that you test 3 different cases?
if so please split them up in 3 tests. tests should be independent and if one case fails only one test should fail.
plus this might give you the opportunity to remove duplication in your tests by using a setup method or helper methods for setup/assertions...

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
FloBue: @teleivo Would I add the current platform version here (2.1.3 as far as I can see?) or a future version number?

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
FloBue: @teleivo You're right, I'll take a look.

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
mks-d: @FloBue you're right that this expands the scope of this ticket quite a lot. Could you create an ad-hoc ticket for this? (And skip this effort for now.)

When it will come to it I can help you setting up Spring Test to test complex obs handlers. I have done something along those lines in Attachments already (see [here](https://github.com/openmrs/openmrs-module-attachments/blob/75b6dc2bdf4ff723a6a2cc9417e25801affe516e/api/src/test/java/org/openmrs/module/attachments/obs/ImageAttachmentHandlerContextSensitiveTest.java#L1) for example.)

FloBue: @mks-d The ad-hoc ticket can be found at https://issues.openmrs.org/browse/TRUNK-5392 . I wasn't sure whether this is more an "Enhancement" or a "Bug".

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
teleivo: I mean all related to this ticket. I mean these comments are just duplicating the code, therefore they should go. I personally only very rarely add comments.

FloBue: I can see why, although I like comments when reading and writing code because they provide some order, or "story", like headlines (even though headlines only condense the content of the following text as well). I also find code blocks I'm looking for faster by using the comments, even if it's not complicated code. I suppose that's a matter of personal preference. I'll delete them, of course.

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
FloBue: @teleivo For one case, I was able to use contentEquals without having to rewrite most of the test. I think it should be fine as it is now.

FloBue:
@teleivo @mks-d This new push should cover the remarks by @teleivo. I hope I haven't missed anything.

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -108,16 +81,22 @@ public File getOutputFileToWrite(Obs obs) throws IOException {
...
-		extension = StringUtils.isNotEmpty(extension) ? extension : "raw";
+		if (StringUtils.isEmpty(result)) {
+			result = FilenameUtils.getBaseName(filename);
+			
+			if (StringUtils.isEmpty(result)) {
+				result = "raw";
+			}
+		}
 		
-		return extension;
```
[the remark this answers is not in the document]
FloBue: I renamed that on purpose since, weirdly enough, that method does not always return the extension. It depends on what you put in, so the result may also just be the filename itself if no extension is found.

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -59,46 +55,23 @@ public AbstractHandler() {
...
-		File dir = OpenmrsUtil.getDirectoryInApplicationDataDirectory(Context.getAdministrationService().getGlobalProperty(
-		    OpenmrsConstants.GLOBAL_PROPERTY_COMPLEX_OBS_DIR));
-		File outputfile;
-		
-		// Get the output stream
-		if (null == title) {
-			String now = longfmt.format(new Date());
-			outputfile = new File(dir, now);
+		if (StringUtils.isNotBlank(titleWitoutExtension)) {
+			filename = titleWitoutExtension + "_" + uuid + extension;
```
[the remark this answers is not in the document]
FloBue: Oh, yeah I copied that part of the code from Ivo's proposed cleaner solution and did not catch the typo. I will fix that for the new push I'll do tomorrow.

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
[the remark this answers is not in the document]
mks-d: I agree with @FloBue here but perhaps the Javadoc could be improved to better reflect what this function actually does. It's already been marked as deprecated, which is already a step ahead.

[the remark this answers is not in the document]
FloBue: @mks-d That's a good idea, I will add/rewrite some information in the Javadoc comment to make clear what to expect as a result.

mks-d:
Merged.

Thanks so much @FloBue !

FloBue:
@mks-d @teleivo @dkayiwa Nice! Thanks for your input for improvements as well. I'll hopefully be back for more once my thesis is all set and done in June :).

dkayiwa:
@FloBue we anxiously look forward to that and wish you all the best in your thesis! 😊

PR #2642, opened 2018-03-29
TRUNK-5307: Create test case for getFormCriteria

Mahov-Sven:
## TRUNK 5307: Create test case for getFormCriteria
Link: https://issues.openmrs.org/browse/TRUNK-5307

There was discussion about some of the tests, and a pull was requested even though not all tests are passing. Here is the link to the Talk page about this.
Talk: https://talk.openmrs.org/t/help-setting-up-tests-for-trunk-5307/17093/2

## Testing

see https://issues.openmrs.org/browse/TRUNK-5307

  No? -> execute above command

### DIFF
--- a/api/src/test/java/org/openmrs/api/db/hibernate/HibernateFormDAOTest.java
+++ b/api/src/test/java/org/openmrs/api/db/hibernate/HibernateFormDAOTest.java
@@ -1,0 +1,448 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.db.hibernate;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.openmrs.EncounterType;
+import org.openmrs.Field;
+import org.openmrs.Form;
+import org.openmrs.FormField;
+import org.openmrs.api.EncounterService;
+import org.openmrs.api.FormService;
+import org.openmrs.api.context.Context;
+import org.openmrs.test.BaseContextSensitiveTest;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class HibernateFormDAOTest extends BaseContextSensitiveTest {
+	
+	protected static final String INITIAL_FIELDS_XML = "org/openmrs/api/db/hibernate/include/HibernateFormDAOTest-initialFieldTypes.xml";
+	
+	protected static final String FORM_XML = "org/openmrs/api/db/hibernate/include/HibernateFormDAOTest-formDataset.xml";
+	
+	@Autowired
+	HibernateFormDAO hibernateFormDAO;
+	
+	@Before
+	public void setUp() {
+		executeDataSet(INITIAL_FIELDS_XML);
+		executeDataSet(FORM_XML);
+	}
+	
+	@Test
+	public void shouldReturnAllFormsWhenGivenNoCriteria() {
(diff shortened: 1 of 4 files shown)


api/src/test/java/org/openmrs/api/db/hibernate/HibernateFormDAOTest.java
```diff
@@ -0,0 +1,448 @@
...
+		String partialName = "";
+		Boolean published = null;
+		Collection<EncounterType> encounterTypes = new HashSet<EncounterType>();
+		Boolean retired = null;
+		Collection<FormField> containingAnyFormField = new HashSet<FormField>();
+		Collection<FormField> containingAllFormFields = new HashSet<FormField>();
+		Collection<Field> fields = new HashSet<Field>();
+		
+		List<Form> formsActual = hibernateFormDAO.getForms(partialName, published, encounterTypes, retired,
+		    containingAnyFormField, containingAllFormFields, fields);
```
wluyima: I'd think passing in null values for collections works too rather than passing in empty collections

api/src/test/java/org/openmrs/api/db/hibernate/HibernateFormDAOTest.java
```diff
@@ -0,0 +1,448 @@
...
+		Integer formCountActual = hibernateFormDAO.getFormCount(partialName, published, encounterTypes, retired,
+		    containingAnyFormField, containingAllFormFields, fields);
+		
+		ArrayList<Form> expectedForms = new ArrayList<Form>();
+		expectedForms.add(formService.getForm(2));
```
wluyima: Setting include retired to true should return the same forms as those when you set retired to false with the addition of retired ones. So I'd think this should include the form with id 1 since it got returned in the previous test

api/src/test/java/org/openmrs/api/db/hibernate/HibernateFormDAOTest.java
```diff
@@ -0,0 +1,448 @@
...
+		expectedForms.add(formService.getForm(3));
+		Assert.assertEquals(expectedForms, formsActual);
+		Assert.assertEquals(Integer.valueOf(2), formCountActual);
+	}
+	
+	@Test
+	public void shouldReturnTheSameFormMultipleTimesWhenMatchingOnRepeatedFields() {
```
wluyima: I think this shouldn't be the case, it should return a single unique form even if the fields are different as long as they match the same form

api/src/test/java/org/openmrs/api/db/hibernate/HibernateFormDAOTest.java
Mahov-Sven: Regarding the actual functionality of this boolean, the tests we wrote already pass. That means that if the functionality you mention is correct, then there is a bug. We can of course modify tests to show that this is a bug, but I want to confirm it is instead of a reading mistake. If it is a bug, I do want to suggest refactoring the 'retired' variable name to 'includeRetiered' to improve readability.

api/src/test/java/org/openmrs/api/db/hibernate/HibernateFormDAOTest.java
Mahov-Sven: Unfortunately, passing null values does not work as none of the collections are checked to make sure they are non-null before a method is called on them. Most just use collection.isEmpty() right away. 

I did modify a test to ensure this was true, and a NullPointerException was thrown when a null collection was used.

api/src/test/java/org/openmrs/api/db/hibernate/HibernateFormDAOTest.java
Mahov-Sven: This test also passes, meaning this is the current implementation of the method. We agree that this is extremely odd, and shouldn't be the case. However, according to the documentation we read about how the getFormCritieria() method and the methods it calls works, this is the correct functionality as written. If this is wrong, then this is a bug/improper implementation of functionality.

Mahov-Sven:
I would like to start a thread talking about the 
`shouldReturnZeroFormsWhenMatchingOnAllFormFieldsWithNoMatch`
test and the
`shouldReturnOneFormWhenMatchingOnAllFormFieldsWithASingleMatch`
test. We are fairly sure that these tests do pass, however, they do not. Among all the other problems @wluyima saw, this is the only one which currently has a failing test. Is this a bug? Are the tests wrong?

dkayiwa:
@Mahov-Sven are you still interested in finishing this up?

dkayiwa:
Closing this because the author no longer works on it.

PR #2643, opened 2018-03-29
Trunk-5230: PatientProgram: When completing a program, the end date of the most recent state in each workflow should be set to the completion date of the program

shifona1:
Updated savePatientProgram() to set Enddate of all recent states of each workflow and added two testcases for it.  
Added getRecentStates() in PatientProgram.java which fetches recent states of each workflow.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5230

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/ProgramWorkflowServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ProgramWorkflowServiceImpl.java
@@ -261,7 +261,17 @@ public class ProgramWorkflowServiceImpl extends BaseOpenmrsService implements Pr
 				}
 			}
 		}
-		
+		// Makes sure that the end dates of most recent states in each workflow
+		// and the program end date are consistent
+		if (patientProgram.getDateCompleted() != null) {
+			for (PatientState state : patientProgram.getMostRecentStateInEachWorkflow()) {
+				// The EndDate of active states only should be updated
+				if (state.getEndDate() == null) {
+					state.setEndDate(patientProgram.getDateCompleted());
+				}
+			}
+		}
+
 		return dao.savePatientProgram(patientProgram);
 	}
 	
--- a/api/src/test/java/org/openmrs/api/ProgramWorkflowServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/ProgramWorkflowServiceTest.java
@@ -126,6 +126,237 @@ public class ProgramWorkflowServiceTest extends BaseContextSensitiveTest {
 		assertEquals(today, ptProg.getDateCompleted());
 		
 	}
+
+	/**
+	 * Tests if the savePatientProgram(PatientProgram) sets the EndDate of recent state of each workflow
+	 * on calling the setDateCompleted(Date date) method.
+	 *
+	 * @see ProgramWorkflowService#savePatientProgram(PatientProgram)
+	 */
+	@Test
+	public void savePatientProgram_shouldSetEndDateOfAllRecentStatesWhenCompletingTheProgram() throws Exception {
+		Date day3 = new Date();
+		Date day2_5 = new Date(day3.getTime() - 12*3600*1000);
+		Date day2 = new Date(day3.getTime() - 24*3600*1000);
+		Date day1 = new Date(day2.getTime() - 24*3600*1000);
+
+		// Program Architecture
+		Program program = new Program();
+		program.setName("TEST PROGRAM");
+		program.setDescription("TEST PROGRAM DESCRIPTION");
+		program.setConcept(cs.getConcept(3));
+
+		ProgramWorkflow workflow1 = new ProgramWorkflow();
+		workflow1.setConcept(cs.getConcept(4));
+		program.addWorkflow(workflow1);
+
+		ProgramWorkflow workflow2 = new ProgramWorkflow();
+		workflow2.setConcept(cs.getConcept(4));
+		program.addWorkflow(workflow2);
+
+		// workflow1
+		ProgramWorkflowState state1_w1 = new ProgramWorkflowState();
+		state1_w1.setConcept(cs.getConcept(5));
+		state1_w1.setInitial(true);
+		state1_w1.setTerminal(false);
+		workflow1.addState(state1_w1);
+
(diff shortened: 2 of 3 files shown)


api/src/test/java/org/openmrs/api/ProgramWorkflowServiceTest.java
```diff
@@ -126,6 +126,208 @@ public void savePatientProgram_shouldUpdatePatientProgram() {
...
+		// Patient Program Architecture
+		PatientProgram patientprogram = new PatientProgram();
+		patientprogram.setProgram(program);
+		patientprogram.setPatient(new Patient());
+		patientprogram.setDateEnrolled(day1);
+		patientprogram.setDateCompleted(null);
+
+		PatientState patientstate1_w1 = new PatientState();
+		patientstate1_w1.setStartDate(day1);
+		patientstate1_w1.setEndDate(day2);
```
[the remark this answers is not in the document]
shifona1: I've removed the End Date assignment.

mogoodrich:
Thanks @shifona1, this looks good!

Rethinking this now though I do have a question regarding the original design we came up with... I am going to post on the ticket for feedback before merging...

shifona1:
Thankyou so much @mogoodrich  for the review!
I'll be waiting for further updates and please do let me know incase of any modifications required.

api/src/main/java/org/openmrs/api/impl/ProgramWorkflowServiceImpl.java
```diff
@@ -261,7 +261,12 @@ public PatientProgram savePatientProgram(PatientProgram patientProgram) throws A
 				}
 			}
 		}
-		
+		// Makes sure that the end dates of most recent states in each workflow
+		// and the program end date are consistent
+		for (PatientState state : patientProgram.getMostRecentStateInEachWorkflow()) {
```
wluyima: I thought this should only be happening when completing a program, am not seeing the condition that ensures that.

api/src/test/java/org/openmrs/api/ProgramWorkflowServiceTest.java
```diff
@@ -126,6 +126,217 @@ public void savePatientProgram_shouldUpdatePatientProgram() {
...
 	}
+
+	/**
+	 * Tests if the savePatientProgram(PatientProgram) sets the EndDate of recent state of each workflow
+	 * on calling the setDateCompleted(Date date) method.
+	 *
+	 * @see ProgramWorkflowService#savePatientProgram(PatientProgram)
+	 */
+	@Test
+	public void savePatientProgram_onSetDateCompleted_shouldSetEndDateOfAllRecentStates() throws Exception {
```
wluyima: Would be better if this were named something like

`savePatientProgram_shouldSetEndDateOfAllRecentStatesWhenCompletingTheProgram`

I.e. have one underscore

api/src/test/java/org/openmrs/api/ProgramWorkflowServiceTest.java
```diff
@@ -126,6 +126,217 @@ public void savePatientProgram_shouldUpdatePatientProgram() {
...
+	}
+
+	/**
+	 * Tests if the savePatientProgram(PatientProgram) sets the EndDate of recent state of each workflow
+	 * when a patient transitions to a terminal state.
+	 *
+	 * @see ProgramWorkflowService#savePatientProgram(PatientProgram)
+	 */
+	@Test
+	public void savePatientProgram_onTransitionToTerminalState_shouldSetEndDateOfAllRecentStates() throws Exception {
```
wluyima: Same here, one underscore should be enough

api/src/main/java/org/openmrs/api/impl/ProgramWorkflowServiceImpl.java
mogoodrich: @wluyima this confused me for a second as well, but it's actually more or less as the ticket specifies, because it's handling these two cases:

* When completing a program, the end date of the most recent state in each workflow should be set to the end date of the program
* When uncompleting a program (setting end date to null) the end date of the most recent state in each workflow should be set to null

However, I'm not sure anymore if this is the correct behavior.

wluyima: This method is called to save or update a program so not all the time the caller is completing or uncompleting a program

mogoodrich: Fair enough... will add some more comments on the ticket.

mogoodrich: Based on our discussion on this ticket can we change this loop so that it only executes if patientProgram.getDateCompleted() == null, and then it only sets the endDate for state if state.setEndDate() == null?

I think that's the only change we will need, though we'd want to add some tests for these cases.  Sorry for the additional changes @shifona1, let me know if you have any questions.

shifona1: Thanks @mogoodrich @wluyima ,
I'll update the changes soon. 
So, just to confirm, It should be ensured that the loop executes only when the program is completed or uncompleted and only the recent states which have their enddate as null (that is they are not suspended) should be set. 

The completion of the program can be detected if the enddate is set i.e `patientProgram.getDateCompleted() != null`.
Also, from what i have understood currently, we are not handling the case when a program is uncompleted, so we will also need to remove the test cases related to the same, since, they will fail otherwise. 
So, there are roughly three major changes: one in this file, second to add tests for these cases and third to remove the tests for uncompletion.

mogoodrich: Yes, that sounds correct!  And to your other questions, it is fine to have the tests in ProgramWorkflowServiceTests... this does seem like the correct place for them.

shifona1:
I've updated the PR.
Please let me know, incase of any changes required.
Thanks

mogoodrich:
Looks good, thanks for the extended work on this @shifona1 !

PR #2667, opened 2018-05-13
TRUNK-4939 : update visit validator to check startdate is after birthdate

Ben-Waters:

Added validation to visits that the start date can not be before the birthdate of the patient.

In cases where the age of a patient was estimated, a 50% margin of error is included.

I also updated one of the existing person's test data. They had a birthdate of 2007 but in one of the tests they started a visit in 2005. I changed their birthdate so this is no longer an issue.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-4939

  No? -> execute above command


Ben-Waters:
Hmm strange error in the CI log, it's not clear to me what stalled the build. There also doesn't seem to be anything to try it again in case that was a one time issue.

Please advise what I should do to resolve this.

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -448,4 +448,52 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+		new VisitValidator().validate(visit, errors);
+		assertTrue(errors.hasFieldErrors("startDatetime"));
+	}
+	
+	/**
+	 * @see VisitValidator#validate(Object,Errors)
+	 */
+	@Test
+	public void validate_shouldFailValidationIfVisitDateBeforeEstimatedBirthDate() {
+		VisitService vs = Context.getVisitService();
```
teleivo: please use the `visitService` thats already a field of the test class and set up in the `@Before`

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -448,4 +448,52 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+		visit.setStartDatetime(c.getTime());
+		
+		Errors errors = new BindException(visit, "visit");
+		new VisitValidator().validate(visit, errors);
+		assertTrue(errors.hasFieldErrors("startDatetime"));
+	}
+	
+	/**
+	 * @see VisitValidator#validate(Object,Errors)
+	 */
```
teleivo: remove these javadocs as they do not provide value. we should not add them anymore. you can easily navigate or see javadocs of the method under test in the test itself

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -448,4 +448,52 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
 		new VisitValidator().validate(visit, errors);
 		assertEquals(true, errors.hasFieldErrors("voidReason"));
 	}
+	
+	/**
+	 * @see VisitValidator#validate(Object,Errors)
+	 */
+	@Test
+	public void validate_shouldFailValidationIfVisitDateBeforeBirthDate() {
+		VisitService vs = Context.getVisitService();
```
teleivo: same as below use test class field visitService

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -448,4 +448,52 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+		Visit visit = new Visit();
+
```
teleivo: remove newline between visit initialization and all the setters.

to get a style of arrange (setup what you need for the test), act (execute the method/thing under test), assert (AAA). this makes it easy to read and spot what you are testing

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -448,4 +448,52 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+		
+		visit.setVisitType(vs.getVisitType(1));
+		visit.setPatient(Context.getPatientService().getPatient(2));
+		Calendar c = Calendar.getInstance();
+		c.set(1974, 4, 8);
+		visit.setStartDatetime(c.getTime());
+		
+		Errors errors = new BindException(visit, "visit");
+		new VisitValidator().validate(visit, errors);
+		assertTrue(errors.hasFieldErrors("startDatetime"));
```
teleivo: newline before the assertion

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -448,4 +448,52 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+		visit.setPatient(Context.getPatientService().getPatient(7));
+		
+		//Set the calendar within the margin of error to make sure it passes
+		Calendar c = Calendar.getInstance();
+		c.set(1956, 12, 31);
+		visit.setStartDatetime(c.getTime());
+		
+		Errors errors = new BindException(visit, "visit");
+		new VisitValidator().validate(visit, errors);
+		assertEquals(false, errors.hasFieldErrors("startDatetime"));
```
teleivo: use the same AAA style as explained above. and split those into 2 separate tests. it seems to me that you are testing two cases. therefore they should be independent. if one case fails the other shouldnt necessarily fail as well.

teleivo: use `assertFalse` when you expect false, like you did with true `assertTrue`

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
teleivo: can you assert that the case you expect is actually occuring by using an assertion we built `hasFieldErrors` you can see an example here:

https://github.com/openmrs/openmrs-core/blob/master/api/src/test/java/org/openmrs/validator/ConceptValidatorTest.java#L89-L99

also this test example shows the AAA style and that there is the least clutter necessary to see whats tested. Maybe you can apply some more cleanup to this test class according to the example test.

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -107,6 +110,12 @@ public void validate(Object target, Errors errors) {
 				validateStopDatetime(visit, otherVisit, errors);
 			}
 		}
+		
+		// Check visit not before patient's birthdate
+		if (visit.getPatient() != null) {
+			validateBirthDate(visit, errors);
```
teleivo: how about renaming the private method you added to `validateVisitStartedBeforePatientBirthdate`, this lets you remove your comment as its clear from the method name what it does.
and move the `visit.getPatient() != null` inside the private method.

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -159,4 +168,28 @@ private void validateStopDatetime(Visit visit, Visit otherVisit, Errors errors)
 		}
 		
 	}
+	
+	private void validateBirthDate(Visit visit, Errors errors) {
+		
+		if (visit.getPatient().getBirthdate() != null && visit.getStartDatetime() != null) {
```
teleivo: turn this into a "guard clause" to remove nesting and make it easier to read. by inverting the checks.
(already included the if from above when patient is null)

```
if (visit.getPatient() == null || visit.getPatient().getBirthdate() == null || visit.getStartDatetime() == null) {
  return;
}

File: api/src/main/java/org/openmrs/validator/VisitValidator.java [hunk 9]
```diff
@@ -159,4 +168,28 @@ private void validateStopDatetime(Visit visit, Visit otherVisit, Errors errors)
...
+			
+			Date birthday = visit.getPatient().getBirthdate();
```
teleivo: can you please extract the logic of getting/calculating the patient birthdate into a private method

`getPatientBirthdateAdjustedIfEstimated`

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
Ben-Waters: When I refactored this, I actually ditched using a visitService at all and instead am calling the method "private Visit makeVisit(Integer patientId)". Hopefully that makes sense.

Ben-Waters:
Have another look, it's hopefully a bit cleaner now.

Ben-Waters:
@teleivo changes made. For the tests around estimations I added a death date so that the patients have a static age, then made the passing test exactly within the estimate margin and the failing test 1 day outside of it.

As for how permissive the error margin is, the older the patient gets the more extreme it is. I implemented it the way described on the ticket but maybe we should add a maximum error margin of 10 years for example?

api/src/test/resources/org/openmrs/include/standardTestDataset.xml
```diff
@@ -237,7 +237,7 @@
...
   <person person_id="1" gender="M" birthdate="1948-01-01 00:00:00.0" birthdate_estimated="0" dead="false" creator="1" date_created="2005-01-01 00:00:00.0" voided="false" uuid="ba1b19c2-3ed6-4f63-b8c0-f762dc8d7562"/>
   <person person_id="2" gender="M" birthdate="1975-04-08 00:00:00.0" birthdate_estimated="0" dead="false" creator="1" date_created="2005-09-22 00:00:00.0" voided="false" uuid="da7f524f-27ce-4bb2-86d6-6d1d05312bd5"/>
-  <person person_id="6" gender="M" birthdate="2007-05-27 00:00:00.0" birthdate_estimated="0" dead="false" creator="1" date_created="2006-01-18 00:00:00.0" voided="false" uuid="a7e04421-525f-442f-8138-05b619d16def"/>
+  <person person_id="6" gender="M" birthdate="1975-05-27 00:00:00.0" birthdate_estimated="0" dead="false" creator="1" date_created="2006-01-18 00:00:00.0" voided="false" uuid="a7e04421-525f-442f-8138-05b619d16def"/>
```
teleivo: do we need the change in the dataset now that you create patients/visits from scratch?

Ben-Waters: Yes it's still needed. Without this change the following tests fail because they make a visit in 2005.
mergePatients_shouldChangeUserRecordsOfNonPreferredPersonToPreferredPerson
mergePatients_shouldMergeVisitsFromNonPreferredToPreferredPatient
mergePatients_shouldVoidNonPreferredPersonObject

I could also look at changing that visit's start date but it seemed lower impact to change this patient's birthday to me.

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -29,6 +31,10 @@
 @Handler(supports = { Visit.class }, order = 50)
 public class VisitValidator extends BaseCustomizableValidator implements Validator {
 	
+	private static final double ESTIMATED_BIRTHDATE_ERROR_MARGIN = -0.5;
+	
+	private static final int ESTIMATED_BIRTHDATE_ERROR_MARGIN_MINIMUM = -1;
```
teleivo: rename to ESTIMATED_BIRTHDATE_ERROR_MARGIN_MINIMUM_YEARS

so its clear its unit is years

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -159,4 +167,29 @@ private void validateStopDatetime(Visit visit, Visit otherVisit, Errors errors)
...
+	}
+	
+	private Date getPatientBirthdateAdjustedIfEstimated(Patient patient) {
+		Date birthday = patient.getBirthdate();
+		
+		if (patient.getBirthdateEstimated()) {
+			Calendar cal = Calendar.getInstance();
+			cal.setTime(birthday);
+			cal.add(Calendar.YEAR, Math.min(ESTIMATED_BIRTHDATE_ERROR_MARGIN_MINIMUM,
+			        new Double(Math.ceil(patient.getAge() * ESTIMATED_BIRTHDATE_ERROR_MARGIN)).intValue()));
```
teleivo: extract small private method `private int calculateGracePeriodInYears(Integer age)`

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -448,4 +455,78 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
 		new VisitValidator().validate(visit, errors);
 		assertEquals(true, errors.hasFieldErrors("voidReason"));
 	}
+	
+	@Test
+	public void validate_shouldFailValidationIfVisitDateBeforeBirthDate() {
```
teleivo: we should have test cases

* `validate_shouldPassValidationIfVisitDateOnBirthDate` - with the same patient setup as here and visit calendar.set(1974, 4, 8)
* `validate_shouldPassValidationIfVisitDateAfterBirthDate` - with the same patient setup as here and visit calendar.set(1974, 4, 9)

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -448,4 +455,78 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+	
+	@Test
+	public void validate_shouldAcceptValidationIfVisitDateWithinEstimatedBirthDateMarginOfError() {
+		Visit visit = new Visit();
+		Patient patient = new Patient();
+		calendar.set(2000, 7, 25);
+		patient.setBirthdate(calendar.getTime());
+		patient.setBirthdateEstimated(true);
+		calendar.set(2010, 7, 25);
+		patient.setDeathDate(calendar.getTime());
```
teleivo: just to understand, we need the deathdate so the age will stay fixed?

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -448,4 +455,78 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+		visit.setStartDatetime(calendar.getTime());
+		Errors errors = new BindException(visit, "visit");
+		
+		new VisitValidator().validate(visit, errors);
+		
+		assertThat(errors, hasFieldErrors("startDatetime", "Visit.startDateCannotFallBeforeTheBirthDateOfTheSamePatient"));
+	}
+	
+	@Test
+	public void validate_shouldAcceptValidationIfVisitDateWithinEstimatedBirthDateMarginOfError() {
```
teleivo: use `Pass` instead of `Accept` for the positive test cases as this is what we have in other tests, rename to

`validate_shouldPassValidationIfVisitDateOnEstimatedBirthDatesGracePeriod`

to be close to naming of above test cases.

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -448,4 +455,78 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+		visit.setStartDatetime(calendar.getTime());
+		Errors errors = new BindException(visit, "visit");
+		
+		new VisitValidator().validate(visit, errors);
+		
+		assertThat(errors, not(hasFieldErrors("startDatetime")));
+	}
+	
+	@Test
+	public void validate_shouldFailValidationIfVisitDateBeforeEstimatedBirthDate() {
```
teleivo: rename to `validate_shouldFailValidationIfVisitDateBeforeEstimatedBirthDatesGracePeriod`

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
```diff
@@ -448,4 +455,78 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+		visit.setStartDatetime(calendar.getTime());
+		Errors errors = new BindException(visit, "visit");
+		
+		new VisitValidator().validate(visit, errors);
+		
+		assertThat(errors, hasFieldErrors("startDatetime", "Visit.startDateCannotFallBeforeTheBirthDateOfTheSamePatient"));
+	}
+	
+	@Test
+	public void validate_shouldAcceptValidationIfVisitDateWithinEstimatedBirthDateMinimumOneYear() {
```
teleivo: rename to `validate_shouldPassValidationIfVisitDateOnEstimatedBirthDatesMinimumOneYearGracePeriod`
and add the failing test case of `validate_shouldFailValidationIfVisitDateBeforeEstimatedBirthDatesMinimumOneYearGracePeriod`

api/src/test/resources/org/openmrs/include/standardTestDataset.xml
teleivo: its ok then, just wanted to know. thanks!

api/src/main/java/org/openmrs/validator/VisitValidator.java
teleivo: @djazayeri are you ok with `Math.ceil` (than for ex. Math.floor) here? This means we are more generous when computing the grace period

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
Ben-Waters: We do if the birth date is estimated. The reason for this is because the permissive calculation for it says that the visit start date is still valid within half the patients estimated age. So in this test for example, they were born in 2000, died 2010, age 10, so a buffer of 5 years. We can then check that this passes within exactly 5 years and fails exactly outside of 5 years. By contrast if the patient isn't dead their age will continue to grow when the tests are run in the future and as a result the edge for the edge cases will be moving.

Ben-Waters:
Changes made :)

api/src/test/java/org/openmrs/validator/VisitValidatorTest.java
Ben-Waters: Added below

teleivo:
@Ben-Waters thanks a lot for this awesome contribution, your explanations, quick response time. Looking forward to your next PRs :)

PR #2687, opened 2018-07-03
Addactivationkey

fanyui:

I added method to create a token hash it and persist it and also moved the activation key from the User object to the LoginCredential Object. This resulted to the failure of the method that was getting user by activation key as the mapping for activation key has been removed from the user. Hence i had to rewrite the new method getLoginCredential By token That uses the token to get the users login credential.

- [ x] My pull request only contains **ONE single commit**

  No? -> execute above command


fanyui:
@wluyima  I have done some changes will love you to review them.

api/pom.xml
```diff
@@ -13,7 +13,7 @@
    <parent>
       <groupId>org.openmrs</groupId>
       <artifactId>openmrs</artifactId>
-      <version>2.2.0-SNAPSHOT</version>
+      <version>2.2.1-SNAPSHOT</version>
```
wluyima: This version will conflict with master, try something unique like 2.2.0-passwordreset please

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -110,12 +111,12 @@
...
 	/**
-	 * Gets user using user activation key
-	 * @param activationKey User's activation(token) for password reset 
-	 * @return requested user
+	 * Gets Login Credential using user activation key
+	 * @param token User's (token) for password reset 
+	 * @return requested LoginCredential with associated hashed of token as activation key
 	 */
 	@Authorized( { PrivilegeConstants.GET_USERS })
-	public User getUserByActivationKey(String activationKey);
```
wluyima: You should keep getUserByActivationKey, LoginCredential is not part of the public facing API so you should not return it or expose it

wluyima: To be clear get User by activation key should stay, get login credential by key should be in the DAO, get User by key delegating to the DAO

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -118,16 +119,18 @@ public User getUserByEmail(String email) {
...
-		if(user != null) {
-			String[] tokens = user.getActivationKey().split(":");
-			if(tokens[0].equals(activationKey)) {
-				return user;
+	public LoginCredential getLoginCredentialByToken(String token) {
+		String key = Security.encodeString(token);
+		LoginCredential loginCred = (LoginCredential) sessionFactory.getCurrentSession().createCriteria(LoginCredential.class).add(Restrictions.like("activationKey", key, MatchMode.START)).uniqueResult();	
+		if(loginCred != null) {
+			String[] credTokens = loginCred.getActivationKey().split(":");
+			// if(credTokens[0].equals(key) && (System.currentTimeMillis() <= Long.parseLong(credTokens[1]) ))
```
wluyima: Don't leave commented out code behind please

api/src/main/java/org/openmrs/api/UserService.java
fanyui: The query for getting user by activation key started failing because the mapping was removed and as well the  activation key from the user object. Could you please provide a way of querying using the activation key>

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -652,6 +655,27 @@ private Query createUserSearchQuery(String name, List<Role> roles, boolean inclu
...
+			int validTime = 10*60*1000; //token should be valid for 10 minutes
+			String token = RandomStringUtils.randomAlphanumeric(20);
+			long time = System.currentTimeMillis() + validTime;
+			String hashedKey = Security.encodeString(token);
+			String activationKey = hashedKey+":"+time;
+			
+			LoginCredential credentials = getLoginCredential(user);
+			credentials.setActivationKey(activationKey);			
+			sessionFactory.getCurrentSession().merge(credentials);
+			saveUser(user,null);
```
wluyima: Why are you calling saveUser?

api/src/main/resources/org/openmrs/api/db/hibernate/LoginCredential.hbm.xml
```diff
@@ -28,6 +28,8 @@
 		<property name="salt" type="java.lang.String" column="salt" length="128" />
 		<property name="secretQuestion" type="java.lang.String" column="secret_question" length="255" />
 		<property name="secretAnswer" type="java.lang.String" column="secret_answer" length="255" />
+		<property name="activationKey" type="java.lang.String" column="activation_key" length="150" />
```
wluyima: Why did you change the length to 150?

wluyima:
Where is the logic that ensures that the ket hasn't expired so that it rejects it, shouldn't that be taken care of by getUserByActivationKey?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
fanyui: i see it's really not necessary.  i'll remove it.

api/src/main/java/org/openmrs/api/UserService.java
wluyima: Since you moved the field to LoginCredential then you have to change the query

api/src/main/resources/org/openmrs/api/db/hibernate/LoginCredential.hbm.xml
fanyui: The hashing generated a long string which could not be accomodated by the innitial 50 and also i increased it to 128 as it done with the password since they used thesame hasing algorithm and it still didn't hold because i append the timestamp to it so i had to increase it.

fanyui:
I added the logic for checking expired key but the test case that i was writing was failing because the time expired so i removed it temporally.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
fanyui: ok will fix it i disabled it because of testing purpose

api/src/main/java/org/openmrs/api/UserService.java
fanyui: I did change the query  just couldn't get one that uses creteria and get the user by activationkey.

api/src/main/resources/org/openmrs/api/db/hibernate/LoginCredential.hbm.xml
wluyima: Then I guess have it as 255 which is the standard

fanyui: ok

wluyima:
I thought the time is 10min, that is long enough for your test to pass, may be your are writing it wrongly

api/src/main/java/org/openmrs/api/UserService.java
wluyima: what is the query you are using?

wluyima: And recall what saying mostly here is you should not expose getLoginCredentialBykey and not how you should write the query

fanyui:
The issue was with the test dataset it did passed when i used worked within timeframe. but creating a dataset nothing worked after some time since i will have an outdated time in the dataset and the system using current time will always be more than 10 mins. So i have issues creating dataset for the test. hence i disabled it

wluyima:
Definitely that can't work because time isn't something static, you need to set up the key within the test itself relative to the current time rather in the flat xml test dataset where you can't really achieve the same.

fanyui:
Ok Let me try it that way.

fanyui:
I think i didn't get you well. This is how the logic goes. When you request for a reset. I create a timestamp and add 10 minures to it So when you supply your token next time I just use the system time at the instance and make sure that is is not later than the know time. So setting it i can't figure out what you really mean unless i have to change the current way i've implimented it

api/src/main/java/org/openmrs/api/UserService.java
fanyui: i tried this (User) sessionFactory.getCurrentSession().createQuery("from User u where u.activation_key = :activationKey").setString(
			    "activationKey", activationKey).uniqueResult(); it didn't also turn out to work

fanyui:
I thought of sending the time along side the token and verifying then i noticed it can be modified when it reaches the user through email. So i left the verifying aspect to the system.

wluyima:
I meant in the unit test

api/src/main/java/org/openmrs/api/UserService.java
wluyima: I just mentioned above that since you moved the field to LoginCredential class that that should what query needs to search from

fanyui: Ok When you said i should allow getUserByAcivationKey. My issue is i am unable to get user by activation key directly as you specified. Should  i just allow the getUserByActivation key and internally in the DAO layer i use getLoginCredentialByToken which i already implimented then using the id of the user from the LoginCredentialreturned i can now retreive the real user Instance which can be exposed.

wluyima: I don't seem to understand what you are asking. All am saying is that you need to have getUserByActivationKey only on the service and then you should have getLoginCredentialByKey in the DAO, your service method delegate to the DAO method. LoginCredential has a userId field that you should use to load the use within the service.

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -110,12 +111,12 @@
 	public User getUserByEmail(String email);
 	
 	/**
-	 * Gets user using user activation key
-	 * @param activationKey User's activation(token) for password reset 
-	 * @return requested user
+	 * Gets User using user activation key(hash of token)
```
wluyima: Not hash, it should be the token itself

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -110,12 +111,12 @@
...
+	 * @param token User's (token) for password reset 
+	 * @return requested User with associated hashed of token as activation key
 	 */
 	@Authorized( { PrivilegeConstants.GET_USERS })
-	public User getUserByActivationKey(String activationKey);
+	public User getUserByActivationKey(String token);
```
wluyima: method says key but argument says key, can you harmonize them please?

api/src/main/resources/liquibase-schema-only.xml
```diff
@@ -1831,7 +1831,7 @@
             <column name="email" type="varchar(255)">
             	<constraints unique="true"/>
             </column>
-            <column name="activation_key" type="varchar(50)"/>
+            <column name="activation_key" type="varchar(150)"/>
```
wluyima: You need to change length to 255

api/src/main/resources/liquibase-schema-only.xml
```diff
@@ -1831,7 +1831,7 @@
             <column name="email" type="varchar(255)">
             	<constraints unique="true"/>
             </column>
-            <column name="activation_key" type="varchar(50)"/>
```
wluyima: This changeset should be not be in the liquibase schema only file, should be in the 2.2 file

api/src/main/resources/org/openmrs/api/db/hibernate/LoginCredential.hbm.xml
wluyima: change the length

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1365,26 +1367,25 @@ public void getUserByEmail_shouldGetUserByUsingEmail() {
...
+	public void getUserByActivationKey_shouldGetUserUsingToken() {
 		executeDataSet(XML_FILENAME);
-		String key="XbklsieiskeuztecEnPf";
+		String key="h4ph0fpNzQCIPSw8plJI";
 		User user = userService.getUserByActivationKey(key);	
-		assertNotNull("User with activationKey ", user); 
-		String[] tokens = user.getActivationKey().split(":");
-		assertEquals(key,tokens[0]);		
+		assertNotNull("Login Credentials with activationKey ", user); 
+		String[] tokens = new HibernateUserDAO().getLoginCredential(user).getActivationKey().split(":");
```
wluyima: You should never do this. you should call the service method

wluyima: If you want to test this method, then you need to add the test to UserDAOTest

fanyui:
The method `public User getUserByActivationKey(String token);` of api/UserService uses token as argument at the service layer but my description is so we know that the activation key we are refering to is the hash of the token. so at this stage it is till the token that has been passed.

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -110,12 +111,12 @@
 	public User getUserByEmail(String email);
 	
 	/**
-	 * Gets user using user activation key
-	 * @param activationKey User's activation(token) for password reset 
-	 * @return requested user
+	 * Gets User using user activation key(hash of token)
+	 * @param token User's (token) for password reset 
+	 * @return requested User with associated hashed of token as activation key
```
wluyima: This method on the service layer should not mention anything to do with hashing, the caller provides the raw token and should only mention token

wluyima: It's also okay to call it an activation key

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -122,12 +123,21 @@ public User getUserByEmail(String email) {
...
-	public User getUserByActivationKey(String activationKey) {
-		User user = (User) sessionFactory.getCurrentSession().createCriteria(User.class).add(Restrictions.like("activationKey", activationKey, MatchMode.START)).uniqueResult();	
-		if(user != null) {
-			String[] tokens = user.getActivationKey().split(":");
-			if(tokens[0].equals(activationKey)) {
-				return user;
+	public User getUserByActivationKey(String token) {
+		LoginCredential loginCred = getLoginCredentialByToken(token);
+		if(loginCred != null) {
+			User user = getUser(loginCred.getUserId());
```
wluyima: please combine these into one line

wluyima: It always pointless to assign a value to a new variable when on the next line you return, should always just combine it into a a single statement that returns

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -122,12 +123,21 @@ public User getUserByEmail(String email) {
 	 */
 	@Override
 	@SuppressWarnings("unchecked")
-	public User getUserByActivationKey(String activationKey) {
-		User user = (User) sessionFactory.getCurrentSession().createCriteria(User.class).add(Restrictions.like("activationKey", activationKey, MatchMode.START)).uniqueResult();	
-		if(user != null) {
-			String[] tokens = user.getActivationKey().split(":");
-			if(tokens[0].equals(activationKey)) {
-				return user;
+	public User getUserByActivationKey(String token) {
```
wluyima: I don't think you need this method in the DAO

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -122,12 +123,21 @@ public User getUserByEmail(String email) {
...
+		LoginCredential loginCred = getLoginCredentialByToken(token);
+		if(loginCred != null) {
+			User user = getUser(loginCred.getUserId());
+			return user;
+		}
+		return null;
+	}
```
wluyima: Did you build the project? There should be a space between methods

wluyima: No javadocs?

api/src/main/resources/liquibase-update-to-2.2.xml
```diff
@@ -265,4 +265,9 @@
 			SELECT condition_id, previous_condition_id, patient_id, status, condition_non_coded, concept_id, onset_date, additional_detail, end_date, creator, date_created, voided, voided_by, date_voided, void_reason, uuid FROM `conditions`;
 		</sql>
 	</changeSet>
+	<changeSet id="20180706-passwordreset" author="harisu+fanyui">
```
wluyima: A changeset needs to have a precondition and comment, see others that add a column for an example

fanyui:
the getUserByActivationKey in the api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java file clearly doesnot do much and as well shouldn't have been in the dao but because of the fear of exposing LoginCredential i think that method has to be there

fanyui:
@wluyima  i have fixed the reviews. please do have a look again

wluyima:
Can you please comment when you are on the [files]( Files changed 19) tab and not conversation tab that way I know which part of the code you are referring to.

What am saying is that getUserByActivationKey(String token) should not be in the DAO, the DAO should only have getLoginCredentialByKey which needs to be called by the service method.

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -110,9 +111,9 @@
 	public User getUserByEmail(String email);
 	
 	/**
-	 * Gets user using user activation key
-	 * @param activationKey User's activation(token) for password reset 
-	 * @return requested user
+	 * Gets User using user activation key
+	 * @param token User's (token) for password reset
```
wluyima: param value needs to match the argument name

api/src/main/java/org/openmrs/api/db/UserDAO.java
```diff
@@ -48,7 +48,7 @@
 	/**
 	 * @see org.openmrs.api.UserService#getUserByActivationKey(java.lang.String)
 	 */
-	public User getUserByActivationKey(String activationKey);
+	public User getUserByActivationKey(String token);
```
wluyima: The DAO should have getLoginCredentialByActivationKey instead

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -122,12 +123,26 @@ public User getUserByEmail(String email) {
...
+	
+	/**
+	 * Gets User LoginCredential using token
+	 * @param token User's (token) for password reset 
+	 * @return LoginCredentail associated with token 
+	 */
+	public LoginCredential getLoginCredentialByToken(String token) {
```
wluyima: For consistency use activation key too for this method and not token

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -652,6 +667,26 @@ private Query createUserSearchQuery(String name, List<Role> roles, boolean inclu
...
+	 * @see org.openmrs.api.db.UserDAO#createActivationKey(org.openmrs.User)
+	 */
+	@Override
+	public void createActivationKey(User user) {
+		log.info("Setting activationKey for " + user.getUsername());
+			int validTime = 10*60*1000; //token should be valid for 10 minutes
+			String token = RandomStringUtils.randomAlphanumeric(20);
+			long time = System.currentTimeMillis() + validTime;
+			String hashedKey = Security.encodeString(token);
+			String activationKey = hashedKey+":"+time;
```
wluyima: Lines 676-681 should be in the service layer

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -652,6 +667,26 @@ private Query createUserSearchQuery(String name, List<Role> roles, boolean inclu
...
+			
+			LoginCredential credentials = getLoginCredential(user);
+			credentials.setActivationKey(activationKey);			
+			sessionFactory.getCurrentSession().merge(credentials);
+			//Send Email with unhashed  activation key and Date:
```
wluyima: Even sending of email should be in the service layer so remove this comment, the DAO should only be used to do database operations only.

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -40,6 +40,8 @@
 import org.openmrs.User;
 import org.openmrs.api.context.Context;
 import org.openmrs.api.db.DAOException;
+import org.openmrs.api.db.LoginCredential;
```
wluyima: These need to be removed since they are unused imports

api/src/test/java/org/openmrs/api/db/UserDAOTest.java
```diff
@@ -178,4 +181,46 @@ public void isSecretAnswer_shouldReturnFalseWhenTheAnswerDoesNotMatch() {
 		
 	}
 	
+	@Test
```
wluyima: I think these tests belong to UserServiceTest

api/src/test/java/org/openmrs/api/db/UserDAOTest.java
```diff
@@ -178,4 +181,46 @@ public void isSecretAnswer_shouldReturnFalseWhenTheAnswerDoesNotMatch() {
...
+		int validTime = 10*60*1000; //equivalent to 10 minutes for token to be valid
+		Long tokenTime = System.currentTimeMillis() + validTime;
+		LoginCredential credentials = dao.getLoginCredential(userJoe);
+		credentials.setActivationKey("b071c88d6d877922e35af2e6a90dd57d37ac61143a03bb986c5f353566f3972a86ce9b2604c31a22dfa467922dcfd54fa7d18b0a7c7648d94ca3d97a88ea2fd0:"+tokenTime);			
+		dao.updateLoginCredential(credentials);	
+		User user = dao.getUserByActivationKey(key);
+		assertNotNull("User with activationKey ", user); 
+		
+	}
+
```
wluyima: Do you already have a test for the method that sets the activation key?

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -566,5 +566,11 @@
 	 */
 	@Authorized
 	public void changePasswordUsingSecretAnswer(String secretAnswer, String pw) throws APIException;
+	
+	/*
+	 * Set A users activation key
```
wluyima: Rephrase to **Sets a user's activation key**

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -110,9 +110,9 @@
 	public User getUserByEmail(String email);
 	
 	/**
-	 * Gets user using user activation key
-	 * @param activationKey User's activation(token) for password reset 
-	 * @return requested user
+	 * Gets User using user activation key
```
wluyima: Rephrase to **Gets a user with the specified activation key**

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -566,5 +566,11 @@
 	 */
 	@Authorized
 	public void changePasswordUsingSecretAnswer(String secretAnswer, String pw) throws APIException;
+	
+	/*
+	 * Set A users activation key
+	 * @param User The user for which the activation key will be set
+	 */
+	public void setUserActivationKey(User user);
```
wluyima: You need to require a privilege on this method, possibly **Get Users**

wluyima: Ignore the comment I had made to require a privilege

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -118,16 +118,16 @@ public User getUserByEmail(String email) {
...
-	@SuppressWarnings("unchecked")
-	public User getUserByActivationKey(String activationKey) {
-		User user = (User) sessionFactory.getCurrentSession().createCriteria(User.class).add(Restrictions.like("activationKey", activationKey, MatchMode.START)).uniqueResult();	
-		if(user != null) {
-			String[] tokens = user.getActivationKey().split(":");
-			if(tokens[0].equals(activationKey)) {
-				return user;
+	public LoginCredential getLoginCredentialByActivationKey(String activationKey) {
+		String key = Security.encodeString(activationKey);
+		LoginCredential loginCred = (LoginCredential) sessionFactory.getCurrentSession().createCriteria(LoginCredential.class).add(Restrictions.like("activationKey", key, MatchMode.START)).uniqueResult();
```
wluyima: Did you format the code, why is this line so long?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -118,16 +118,16 @@ public User getUserByEmail(String email) {
...
-		if(user != null) {
-			String[] tokens = user.getActivationKey().split(":");
-			if(tokens[0].equals(activationKey)) {
-				return user;
+	public LoginCredential getLoginCredentialByActivationKey(String activationKey) {
+		String key = Security.encodeString(activationKey);
+		LoginCredential loginCred = (LoginCredential) sessionFactory.getCurrentSession().createCriteria(LoginCredential.class).add(Restrictions.like("activationKey", key, MatchMode.START)).uniqueResult();	
+		if(loginCred != null) {
+			String[] credTokens = loginCred.getActivationKey().split(":");
+			if(credTokens[0].equals(key) && (System.currentTimeMillis() <= Long.parseLong(credTokens[1]) )){
```
wluyima: The logic that checks that the token hasnt expired doesn't belong to the DAO

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -712,9 +714,31 @@ public User getUserByEmail(String email) throws APIException {
...
+	}
+	
+	/**
+	 * @see org.openmrs.api.UserService#setUserActivationKey(org.openmrs.User)
+	 * 
+	 */
+	@Override
+	public void setUserActivationKey(User user) {
+		log.info("Setting activationKey for " + user.getUsername());
+		int validTime = 10*60*1000; //token should be valid for 10 minutes
```
wluyima: I thought we said this should be a GP

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1361,30 +1367,56 @@ public void getUserByEmail_shouldGetUserByUsingEmail() {
...
 	@Test
-	public void getUserByActivationKey_shouldGetUserUsingActivationKey() {
-		executeDataSet(XML_FILENAME);
-		String key="XbklsieiskeuztecEnPf";
-		User user = userService.getUserByActivationKey(key);	
-		assertNotNull("User with activationKey ", user); 
-		String[] tokens = user.getActivationKey().split(":");
-		assertEquals(key,tokens[0]);		
-
+	public void verifyCreateUserActivationKey_shouldCreateUserActivationKey() {
```
wluyima: By convention the part of a test method before the underscore should be the name of the method you are testing

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1361,30 +1367,56 @@ public void getUserByEmail_shouldGetUserByUsingEmail() {
...
+		User u = new User();
+		u.setPerson(new Person());
+		u.addName(new PersonName("Benjamin", "A", "Wolfe"));
+		u.setUsername("bwolfe");
+		u.getPerson().setGender("M");
+		User createdUser = userService.createUser(u, "Openmr5xy");
+		userService.setUserActivationKey(createdUser);
+		LoginCredential credential = dao.getLoginCredential(createdUser);
+		String activationKey = credential.getActivationKey();
+		assertNotNull(activationKey);
```
wluyima: You can combine these into a line

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -712,9 +714,31 @@ public User getUserByEmail(String email) throws APIException {
...
+		}
+		return null;
+	}
+	
+	/**
+	 * @see org.openmrs.api.UserService#setUserActivationKey(org.openmrs.User)
+	 * 
+	 */
+	@Override
+	public void setUserActivationKey(User user) {
```
wluyima: This method needs to return the saved user

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1361,30 +1367,56 @@ public void getUserByEmail_shouldGetUserByUsingEmail() {
...
+	}
+	
+	@Test 
+	public void getUserByActivationKey_shouldGetUserByActivationKey(){
```
wluyima: How different is this test from the one above?

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1361,30 +1367,56 @@ public void getUserByEmail_shouldGetUserByUsingEmail() {
...
-	/**
-	 * @see UserService#getUserByActivationKey(String)
-	 */
 	@Test
-	public void getUserByActivationKey_shouldReturnNullIfNoUserFoundWithGivenKey() {
-		executeDataSet(XML_FILENAME);
-		String key="XbklsieiskNoMatchKey";
-		User user = userService.getUserByActivationKey(key);			
-		assertNull(user);
+	public void getUserByActivationKey_shouldReturnNullIfTokenTimeExpired(){
```
wluyima: The method should always return the user even when the token is expired

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -652,6 +652,16 @@ private Query createUserSearchQuery(String name, List<Role> roles, boolean inclu
...
 		return query;
 	}
+
+	/**
+	 * @see org.openmrs.api.db.UserDAO#createActivationKey(org.openmrs.User)
+	 */
+	@Override
+	public void createActivationKey(User user, String activationKey) {
+			LoginCredential credentials = getLoginCredential(user);
+			credentials.setActivationKey(activationKey);
```
wluyima: These 2 lines belong in the service layer

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
wluyima: Did you see this comment?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -69,6 +71,16 @@ public void setUserDAO(UserDAO dao) {
...
 	}
 	
+	
+	/**
+	 *  @see org.openmrs.api.UserService#setValidTime(java.lang.Integer)
+	 */
+	@Override
+	public void setValidTime(int validTime) {
+		//if valid time is less that a minute or greater thatn 12hrs reset valid time to 1 minutes else set it to the required time.
+		UserServiceImpl.validTime = (validTime < 60*1000) || (validTime > 12*60*60*1000) ? 60*1000 : validTime;
```
wluyima: Why are you prepending the classname?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
wluyima: Did you see this comment?

api/src/test/java/org/openmrs/api/UserServiceTest.java
wluyima: Did you see this comment?

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1361,30 +1367,55 @@ public void getUserByEmail_shouldGetUserByUsingEmail() {
...
-
+	public void setUserActivationKey_shouldCreateUserActivationKey() {
+		User u = new User();
+		u.setPerson(new Person());
+		u.addName(new PersonName("Benjamin", "A", "Wolfe"));
+		u.setUsername("bwolfe");
+		u.getPerson().setGender("M");
+		User createdUser = userService.createUser(u, "Openmr5xy");
+		userService.setUserActivationKey(createdUser);
+		String activationKey  = dao.getLoginCredential(createdUser).getActivationKey();
```
wluyima: You should instead be asserting that UserService.getUserActivationKey returns the same user and not this dao call

api/src/test/java/org/openmrs/api/UserServiceTest.java
wluyima: Did you see this comment?

wluyima:
Please go through all the comments on the pull request and make sure they are addressed

fanyui:
@wluyima  Please i will need your oppinion on those other reviews that i commented on.

wluyima:
Which reviews did you comment on?

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -566,5 +566,17 @@
...
+	 * @param User The user for which the activation key will be set
+	 */
+	public User setUserActivationKey(User user);
+	
+	
+	/*
+	 * Sets the time in seconds for password reset activation key to be valid 
+	 * @param validTime time for which the token is valid
+	 */
+	public void setValidTime(int validTime);
```
wluyima: Why is tis method public? This should not be a public method on the service please

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
wluyima: Did you see this comment?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -712,9 +726,35 @@ public User getUserByEmail(String email) throws APIException {
 	 * 
 	 */
 	@Override
-	@Transactional(readOnly = true)
-	public User getUserByActivationKey(String activationKey) throws APIException {	
-		return dao.getUserByActivationKey(activationKey);
+	public User getUserByActivationKey(String activationKey) {
```
wluyima: Add back @Transactional(readOnly = true)

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -712,9 +726,35 @@ public User getUserByEmail(String email) throws APIException {
...
+		LoginCredential loginCred = dao.getLoginCredentialByActivationKey(activationKey);
+		String[] credTokens = loginCred.getActivationKey().split(":");
+
+		if(loginCred != null  && (System.currentTimeMillis() <= Long.parseLong(credTokens[1]) )) {
```
wluyima: Don't do this in the dao, whether the token is expired or not, the dao should return the token, it's not its business to check for expired tokens

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -712,9 +726,35 @@ public User getUserByEmail(String email) throws APIException {
...
+		return null;
+	}
+	
+	/**
+	 * @see org.openmrs.api.UserService#setUserActivationKey(org.openmrs.User)
+	 * 
+	 */
+	@Override
+	public User setUserActivationKey(User user) {
+		log.info("Setting activationKey for " + user.getUsername());
```
wluyima: Do you really need to log this?

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -66,6 +69,9 @@
 	private UserService userService;
 
 	private MessageSourceService messages;
+	
+	@Autowired(required = false)
```
wluyima: Why do you have to add required = false? Please remove it because I don't think it's necessary

api/src/test/java/org/openmrs/api/UserServiceTest.java
wluyima: Did you see this comment?

api/src/test/java/org/openmrs/api/UserServiceTest.java
wluyima: Did you see this comment

fanyui:
@wluyima  ì have updated the reviews

fanyui:
I have a problem on where the logic which checks if the token has expired will go

wluyima:
In the service layer, as I mentioned earlier always keep business logic out of the DAO, the DAO's job is to make database operations

api/src/main/java/org/openmrs/api/db/UserDAO.java
```diff
@@ -205,4 +207,9 @@
 	 */
 	public Integer getCountOfUsers(String name, List<Role> roles, boolean includeRetired);
 	
+	/**
+	 * @see UserService#setUserActivationKey(LoginCredential)
+	 */
+	public void createActivationKey(LoginCredential credentials);
```
wluyima: Renamed this method to match the equivalent service method

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -67,6 +71,15 @@ public void setUserDAO(UserDAO dao) {
 		this.dao = dao;
 	}
 	
+	/**
+	 * Sets the time in seconds for password reset activation key to be valid
+	 * @param validTime the validTime to set
+	 */
+	public void setValidTime(int validTime) {
```
wluyima: I die mention that this method should not be public

wluyima: What is the use of this method? You should have a utility method that returns the valid time, i'm it should be something like getValidTime()

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -57,6 +59,8 @@
 	
 	protected UserDAO dao;
 	
+	private int validTime = 60*1000; //period of one minute
```
wluyima: Why do you need this?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -67,6 +71,15 @@ public void setUserDAO(UserDAO dao) {
 		this.dao = dao;
 	}
 	
+	/**
+	 * Sets the time in seconds for password reset activation key to be valid
+	 * @param validTime the validTime to set
+	 */
+	public void setValidTime(int validTime) {
+		//if valid time is less that a minute or greater thatn 12hrs reset valid time to 1 minutes else set it to the required time.
+		this.validTime = (validTime < 60*1000) || (validTime > 12*60*60*1000) ? 60*1000 : validTime;
```
wluyima: Can you please use constants for the min, default and max valid time?

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1361,30 +1367,33 @@ public void getUserByEmail_shouldGetUserByUsingEmail() {
...
-		assertEquals(key,tokens[0]);		
-
+	public void setUserActivationKey_shouldCreateUserActivationKey() {
+		User u = new User();
+		u.setPerson(new Person());
+		u.addName(new PersonName("Benjamin", "A", "Wolfe"));
+		u.setUsername("bwolfe");
+		u.getPerson().setGender("M");
+		User createdUser = userService.createUser(u, "Openmr5xy");
+		assertEquals(createdUser, userService.setUserActivationKey(createdUser));
```
wluyima: You need to instead test that you get back the same user when you call getUserByActivationKey

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1361,30 +1367,33 @@ public void getUserByEmail_shouldGetUserByUsingEmail() {
...
+		User u = new User();
+		u.setPerson(new Person());
+		u.addName(new PersonName("Benjamin", "A", "Wolfe"));
+		u.setUsername("bwolfe");
+		u.getPerson().setGender("M");
+		User createdUser = userService.createUser(u, "Openmr5xy");
+		String key="h4ph0fpNzQCIPSw8plJI";
+		int validTime = 10*60*1000; //equivalent to 10 minutes for token to be valid
+		Long tokenTime = System.currentTimeMillis() + validTime;
+		LoginCredential credentials = dao.getLoginCredential(createdUser);
```
wluyima: This is actually not the right thing to do, instead call setActivationKey and make sure getUserByActivation key returns the same user

api/src/test/java/org/openmrs/api/db/UserDAOTest.java
```diff
@@ -178,4 +179,28 @@ public void isSecretAnswer_shouldReturnFalseWhenTheAnswerDoesNotMatch() {
 		
 	}
 	
+	@Test
```
wluyima: These tests are not necessary, they are duplicates of UserServiceTest

api/src/test/java/org/openmrs/api/db/UserDAOTest.java
```diff
@@ -178,4 +179,28 @@ public void isSecretAnswer_shouldReturnFalseWhenTheAnswerDoesNotMatch() {
...
+		Long tokenTime = System.currentTimeMillis() + validTime;
+		LoginCredential credentials = dao.getLoginCredential(userJoe);
+		credentials.setActivationKey("b071c88d6d877922e35af2e6a90dd57d37ac61143a03bb986c5f353566f3972a86ce9b2604c31a22dfa467922dcfd54fa7d18b0a7c7648d94ca3d97a88ea2fd0:"+tokenTime);			
+		dao.updateLoginCredential(credentials);	
+		LoginCredential loginCred = dao.getLoginCredentialByActivationKey(key);
+		assertNotNull("LoginCredential with activationKey ", loginCred); 	
+	}
+	
+	@Test
+	public void verifyUserActivationKey_shouldReturnNullIfActivationKeyIsNotValid() {
```
wluyima: This test doesn't make sense for the DAO, there is no logic in the DAO that needs to be tested since you are already testing the service layer methods anyways

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1361,30 +1367,33 @@ public void getUserByEmail_shouldGetUserByUsingEmail() {
...
-		String[] tokens = user.getActivationKey().split(":");
-		assertEquals(key,tokens[0]);		
-
+	public void setUserActivationKey_shouldCreateUserActivationKey() {
+		User u = new User();
+		u.setPerson(new Person());
+		u.addName(new PersonName("Benjamin", "A", "Wolfe"));
+		u.setUsername("bwolfe");
+		u.getPerson().setGender("M");
+		User createdUser = userService.createUser(u, "Openmr5xy");
```
wluyima: You need a assertion here that checks that the user has no key yet, use dao.getLoginCredention.getActivationKey
And then after you set it, check that the user after that has a key set using dao.getLoginCredention.getActivationKey

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -67,6 +75,14 @@ public void setUserDAO(UserDAO dao) {
 		this.dao = dao;
 	}
 	
+	/**
+	 * @return the validTime for which the password reset activation key will be valid
+	 */
+	public int getValidTime() {
```
wluyima: I did say this method should not be public

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -67,6 +75,14 @@ public void setUserDAO(UserDAO dao) {
...
+		//if valid time is less that a minute or greater than 12hrs reset valid time to 1 minutes else set it to the required time.
+		return (validTime < MIN_VALID_TIME) || (validTime > MAX_VALID_TIME) ? DEFAULT_VALID_TIME : validTime;
```
wluyima: We said you should define a GP for this, that's why you have min, max and default values so that you can have some control on bad values that the admin sets

fanyui:
@wluyima  with a little help and review I would like to finalise this pull request tonight.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -67,6 +75,14 @@ public void setUserDAO(UserDAO dao) {
 		this.dao = dao;
 	}
 	
+	/**
+	 * @return the validTime for which the password reset activation key will be valid
+	 */
+	private int getValidTime() {
+		//if valid time is less that a minute or greater than 12hrs reset valid time to 1 minutes else set it to the required time.
+		return (validTime < MIN_VALID_TIME) || (validTime > MAX_VALID_TIME) ? DEFAULT_VALID_TIME : validTime;
```
wluyima: What is happening here? The valid time needs to be read from a global property that you will have to add, if the value is less or more than the min and max then you use the default.

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -66,6 +69,9 @@
 	private UserService userService;
 
 	private MessageSourceService messages;
+	
+	@Autowired
+	private UserDAO dao;
```
wluyima: Where is the test that ensures that if the key is expired, then you get back nothing

fanyui:
@wluyima  I've added the test and with respect to the global valid time I guess am not understanding you properly.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
wluyima: Recall on our call I said you need to use a global property to let an admin define their preferred valid time, and that way also in your test you set the global property value to 1min for testing purposes so that the test doesn't have to last for the default 10min, do you know what a global property is in OpenMRS and how to use it?

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1361,30 +1367,52 @@ public void getUserByEmail_shouldGetUserByUsingEmail() {
...
-	/**
-	 * @see UserService#getUserByActivationKey(String)
-	 */
 	@Test
-	public void getUserByActivationKey_shouldReturnNullIfNoUserFoundWithGivenKey() {
-		executeDataSet(XML_FILENAME);
-		String key="XbklsieiskNoMatchKey";
-		User user = userService.getUserByActivationKey(key);			
-		assertNull(user);
+	public void getUserByActivationKey_shouldReturnNullIfTokenTimeExpired(){
```
wluyima: Actually this is the test I was looking for, I see you have it, so ignore my comment

fanyui:
@wluyima  please with some material and possible explanation  on how to add and where I can get a grasp of global properties in openmrs. I don't know what openmrs refer to as global property I was thinking of something else. thats why I took the approach I did.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -67,6 +73,16 @@ public void setUserDAO(UserDAO dao) {
 		this.dao = dao;
 	}
 	
+	
+	/**
+	 * @return the validTime for which the password reset activation key will be valid
+	 */
+	private int getValidTime() {
+		int validTime= Integer.parseInt(Context.getAdministrationService().getGlobalProperty("validTime"));
```
wluyima: Use the constants here for validTime

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -597,6 +597,12 @@ public static String getOpenmrsProperty(String property) {
 	 * Specifies the uuid of the concept which represents drug non coded
 	 */
 	public static final String GP_DRUG_ORDER_DRUG_OTHER = "drugOrder.drugOther";
+	
+	/**
+	 * Name of property that stores the duration for which the password reset token is valid
+	 */
+	public static final String GP_PASSWORD_RESET_VALIDTIME = "validTime";
```
wluyima: There is a section for  other security related constants, please move this next to them and add a prefix similar to those other security constants

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
wluyima: What happens if the value isn't set? i.e this returns null or blank string, won't you get a NPE or cast exception

fanyui:
Hello @wluyima please what is still left on addressed on this pul request.

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -854,6 +860,8 @@ public static String getOpenmrsProperty(String property) {
 		props.add(new GlobalProperty(GP_PASSWORD_MINIMUM_LENGTH, "8",
 		        "Configure the minimum length required of all passwords"));
 		
+		props.add(new GlobalProperty(GP_PASSWORD_RESET_VALIDTIME, "600000", "Default of 10 minutes for which the password reset is valid"));
```
wluyima: The description should be better, it needs to say what the global property is for and not its default value, something like **Specifies the duration of time in seconds for which a password reset token is valid, the default value is 10 minutes and the allowed values range from 1 to 12hrs**

fanyui:
Ive addressed it.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -67,6 +73,19 @@ public void setUserDAO(UserDAO dao) {
 		this.dao = dao;
 	}
 	
+	
+	/**
+	 * @return the validTime for which the password reset activation key will be valid
+	 */
+	private int getValidTime() {
+		final int validTime = Context.getAdministrationService().getGlobalProperty("security.validTime") == null
```
wluyima: - Please use the constant here
- You need to take care of when the value is an empty string or a white space, use StringUtils.isBlank() because it takes care of all these

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -67,6 +73,19 @@ public void setUserDAO(UserDAO dao) {
...
+		        ? DEFAULT_VALID_TIME
+		        :
```
wluyima: This doesn't look good, did you format your code?

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -854,6 +860,9 @@ public static String getOpenmrsProperty(String property) {
 		props.add(new GlobalProperty(GP_PASSWORD_MINIMUM_LENGTH, "8",
 		        "Configure the minimum length required of all passwords"));
 		
+		props.add(new GlobalProperty(GP_PASSWORD_RESET_VALIDTIME, "600000",
+		        " Specifies the duration of time in seconds for which a password reset token is valid, the default value is 10 minutes and the allowed values range from 1 to 12hrs"));
```
wluyima: Should be range from 1 minute

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -67,6 +73,18 @@ public void setUserDAO(UserDAO dao) {
 		this.dao = dao;
 	}
 	
+	
+	/**
+	 * @return the validTime for which the password reset activation key will be valid
+	 */
+	private int getValidTime() {
+		final int validTime = StringUtils.isBlank(Context.getAdministrationService().getGlobalProperty("security.validTime"))
```
wluyima: Use the constant you declared for this global property name

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -67,6 +73,18 @@ public void setUserDAO(UserDAO dao) {
...
+		        ? DEFAULT_VALID_TIME
+		        : Integer.parseInt(Context.getAdministrationService().getGlobalProperty("security.validTime"));
```
wluyima: Same herte

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -67,6 +73,18 @@ public void setUserDAO(UserDAO dao) {
 		this.dao = dao;
 	}
 	
+	
+	/**
+	 * @return the validTime for which the password reset activation key will be valid
+	 */
+	private int getValidTime() {
+		final int validTime = StringUtils.isBlank(Context.getAdministrationService().getGlobalProperty("security.validTime"))
+		        ? DEFAULT_VALID_TIME
```
wluyima: This looks anti pattern, did you format the code?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -67,6 +74,18 @@ public void setUserDAO(UserDAO dao) {
 		this.dao = dao;
 	}
 	
+	
+	/**
+	 * @return the validTime for which the password reset activation key will be valid
+	 */
+	private int getValidTime() {
+		String contexTime = Context.getAdministrationService()
```
wluyima: typo here, besides I thought this is validTime, why did you rename it to contextTime which doesn't make sense

fanyui:
Hi @wluyima Please could you take a look?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
wluyima: Did you see this comment?

PR #2693, opened 2018-07-28
TRUNK-5405 - Support free-text cause of death in the data model

fruether:
Fixing TRUNK-5405 - Support free-text cause of death in the data model

https://issues.openmrs.org/browse/TRUNK-5405
I added the attribute cause_of_death_non_coded to the "Person" class and adapted the business logic as requested in the ticket

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5405

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/Person.java
+++ b/api/src/main/java/org/openmrs/Person.java
@@ -72,6 +72,8 @@ public class Person extends BaseChangeableOpenmrsData {
 	
 	private Concept causeOfDeath;
 	
+	private String causeOfDeathNonCoded;
+	
 	private User personCreator;
 	
 	private Date personDateCreated;
@@ -134,7 +136,7 @@ public class Person extends BaseChangeableOpenmrsData {
 		dead = person.getDead();
 		deathDate = person.getDeathDate();
 		causeOfDeath = person.getCauseOfDeath();
-		
+		causeOfDeathNonCoded = person.getCauseOfDeathNonCoded();
 		// base creator/voidedBy/changedBy info is not copied here
 		// because that is specific to and will be recreated
 		// by the subobject upon save
@@ -323,6 +325,29 @@ public class Person extends BaseChangeableOpenmrsData {
 	 */
 	public void setCauseOfDeath(Concept causeOfDeath) {
 		this.causeOfDeath = causeOfDeath;
+	}
+	
+	/**
+	 * This method returns the non coded cause of death
+	 * 
+	 * @return non coded cause of death
+	 * @since 2.2.0
+	 * 
+	 */
+	public String getCauseOfDeathNonCoded() {
+		return this.causeOfDeathNonCoded;
+	}
+	
+	/**
+	 * This method sets the non coded cause of death with the value given as parameter
+	 * 
+	 * @param causeOfDeathNonCoded is a String that describes as text the cause of death
+	 * @since 2.2.0
+	 * @should not fail with null causeOfDeathNonCoded
+	 * @should set the attribute causeOfDeathNonCoded with the given parameter
+	 */
+	public void setCauseOfDeathNonCoded(String causeOfDeathNonCoded) { 
+		this.causeOfDeathNonCoded = causeOfDeathNonCoded;
 	}
 	
 	/**
--- a/api/src/main/java/org/openmrs/validator/PersonValidator.java
+++ b/api/src/main/java/org/openmrs/validator/PersonValidator.java
@@ -53,7 +53,8 @@ public class PersonValidator implements Validator {
 	 * @should fail validation if deathdate is a future date
 	 * @should fail validation if birthdate is after death date
 	 * @should fail validation if voidReason is blank when patient is voided
-	 * @should fail validation if causeOfDeath is blank when patient is dead
+	 * @should fail validation if causeOfDeath and causeOfDeathNonCoded is blank when patient is dead
+	 * @should fail validation if causeOfDeath and causeOfDeathNonCoded is both set
 	 * @should pass validation if gender is blank for Persons
 	 * @should pass validation if field lengths are correct
 	 * @should fail validation if field lengths are not correct
@@ -103,7 +104,12 @@ public class PersonValidator implements Validator {
 			ValidationUtils.rejectIfEmptyOrWhitespace(errors, "voidReason", "error.null");
 		}
 		if (person.getDead()) {
-			ValidationUtils.rejectIfEmpty(errors, "causeOfDeath", "Person.dead.causeOfDeathNull");
+			if(person.getCauseOfDeath() != null && person.getCauseOfDeathNonCoded() != null) {
+				errors.rejectValue("causeOfDeath", "Person.dead.shouldHaveOnlyOneCauseOfDeathOrCauseOfDeathNonCodedSet");
+			}
+			else if(person.getCauseOfDeath() == null && person.getCauseOfDeathNonCoded() == null) {
+				errors.rejectValue("causeOfDeath", "Person.dead.causeOfDeathAndCauseOfDeathNonCodedNull");
+			}
 		}
 		
 		ValidateUtil.validateFieldLengths(errors, Person.class, "gender", "personVoidReason");
--- a/api/src/main/resources/liquibase-update-to-2.2.xml
+++ b/api/src/main/resources/liquibase-update-to-2.2.xml
@@ -305,4 +305,17 @@
 								 referencedTableName="concept"
 								 referencedColumnNames="concept_id" />
 	</changeSet>
+
+	<changeSet id="TRUNK-4505-20180804" author="fruether">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<columnExists tableName="person" columnName="cause_of_death_non_coded"/>
+			</not>
+		</preConditions>
+		<comment>Adding cause_of_death_non_coded to the table person</comment>
+		<addColumn tableName="person">
+			<column name="cause_of_death_non_coded" type="varchar(255)" />
+		</addColumn>
+	</changeSet>
+	
 </databaseChangeLog>
--- a/api/src/main/resources/messages.properties
+++ b/api/src/main/resources/messages.properties
@@ -1793,7 +1793,7 @@ Person.birthdate=Birthdate
 Person.birthdateEstimated=Estimated
 Person.dead=Deceased
 Person.dead.checkboxInstructions=Check if this person is deceased
-Person.dead.causeOfDeathNull=Cause of Death is required for people marked as deceased.
+Person.dead.causeOfDeathAndCauseOfDeathNonCodedNull=Cause of Death or Cause of Death (non coded) is required for people marked as deceased.
 Person.deathDate=Death Date
 Person.deathdateEstimated=Estimated
 Person.causeOfDeath=Cause of Death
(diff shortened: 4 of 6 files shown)


api/src/main/java/org/openmrs/Person.java
```diff
@@ -325,6 +327,21 @@ public void setCauseOfDeath(Concept causeOfDeath) {
 		this.causeOfDeath = causeOfDeath;
 	}
 	
+	/**
+	 * @return cause of person's death non coded
+	 * @since Trunk-5405
```
dkayiwa: since takes on the version number when this is going to be released

api/src/main/java/org/openmrs/Person.java
```diff
@@ -325,6 +327,21 @@ public void setCauseOfDeath(Concept causeOfDeath) {
...
+	 */
+	public String getCauseOfDeathNonCoded() {
+		return this.causeOfDeathNonCoded;
+	}
+	
+	/**
```
dkayiwa: This one also needs the since annotation

api/src/main/java/org/openmrs/Person.java
fruether: @dkayiwa thanks for the review!

I have no idea when it should be released. There is also no detail provided in the issue description as far as I see.

Do you know where I can determine the release number? Or do you know when it will be released?

api/src/main/java/org/openmrs/Person.java
fruether: I think you commented on the same code segment as above. 

Does the setter and getter need a since annotation. Is that the intention of the comment

fruether: I set the version to 2.9.0 since 2.8.0 is already relaesed and the discussion on Talks (https://talk.openmrs.org/t/mark-patient-as-dead/6279/29) suggested that it should be 2.8.0 or 2.9.0. Can you please verify @dkayiwa since I am not completely sure what the right value should be?

api/src/main/java/org/openmrs/Person.java
```diff
@@ -325,6 +327,22 @@ public void setCauseOfDeath(Concept causeOfDeath) {
 		this.causeOfDeath = causeOfDeath;
 	}
 	
+	/**
```
dkayiwa: Can you also put a method description?

api/src/main/java/org/openmrs/Person.java
```diff
@@ -325,6 +327,22 @@ public void setCauseOfDeath(Concept causeOfDeath) {
...
+	 * @return cause of person's death non coded
+	 * @since 2.9.0
+	 */
+	public String getCauseOfDeathNonCoded() {
+		return this.causeOfDeathNonCoded;
+	}
+	
+	/**
+	 * @return cause of person's death non coded
```
dkayiwa: Can you also put the method description?

api/src/main/java/org/openmrs/Person.java
dkayiwa: You can read the next version from here: https://github.com/openmrs/openmrs-core/blob/master/pom.xml#L17

fruether: So 2.2.0 is the next version?

dkayiwa: yes

api/src/main/java/org/openmrs/Person.java
fruether: Can you verify if the method description is sufficient now?

fruether:
@dkayiwa I guess I also have to change the .hbm and liquidbase file. Anything more to do?

dkayiwa:
That looks to be it.

api/src/main/resources/liquibase-update-to-2.2.xml
```diff
@@ -305,4 +305,17 @@
...
+
+	<changeSet id="TRUNK-4505-20180804" author="fruether">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<columnExists tableName="person" columnName="cause_of_death_non_coded"/>
+			</not>
+		</preConditions>
+		<comment>Adding cause_of_death_non_coded to the table person</comment>
+		<addColumn tableName="person">
+			<column name="cause_of_death_non_coded" type="varchar(1024)" />
```
wluyima: Does this need to be this long? I'd imagine 255 should be fine

api/src/test/java/org/openmrs/validator/PersonValidatorTest.java
```diff
@@ -197,5 +198,35 @@ public void shouldNotSetDeathBeforeBirth() {
...
+		Patient pa = new Patient(1);
+		pa.setDead(true);
+		
+		pa.setCauseOfDeathNonCoded("Some text describing Cause of Death");
+		pa.setCauseOfDeath(new Concept());
+		
+		Errors errors = new BindException(pa, "patient");
+		validator.validate(pa, errors);
+		
+		Assert.assertTrue(errors.hasFieldErrors("causeOfDeath"));
```
wluyima: You probably want to check the error code too

api/src/test/java/org/openmrs/validator/PersonValidatorTest.java
```diff
@@ -197,5 +198,35 @@ public void shouldNotSetDeathBeforeBirth() {
...
+		person.setPersonVoided(true);
+		person.setPersonVoidReason("voidReason");
+		person.setGender("g");
+		
+		Errors errors = new BindException(person, "patient");
+		PersonValidator personValidator = new PersonValidator();
+		personValidator.validate(person, errors);
+
+		Assert.assertFalse(errors.hasErrors());
+	}
```
wluyima: Where is the test that ensures the test fails if both fields are not set

fruether: I think this case was already implemented by:
validate_shouldFailValidationIfCauseOfDeathIsBlankWhenPatientIsDead

That's why I did not add it

api/src/main/resources/liquibase-update-to-2.2.xml
fruether: There was no specification in the ticket description. I made it long to be sure al potential descriptions would fit.

I can change it to.254 or any other number if it is more applicable

wluyima: 255 should be fine

api/src/test/java/org/openmrs/validator/PersonValidatorTest.java
wluyima: But that test is making the check for one field and not both

fruether: Depends what test you are refering to. The test _validate_shouldNotFailWhenDeathCauseNotCodedIsSet_ which you commented to is  making the check for one field.

However, there already existed a test case before which neither set _causeOfDeath_ nor **causeOfDeathNonCoded**. As far as I can see and understand this code, this test case is called: **validate_shouldFailValidationIfCauseOfDeathIsBlankWhenPatientIsDead** and should validate the case if both are not set.
---> https://github.com/openmrs/openmrs-core/blob/1327f91d828cfa81f9412f6cf29137d21de6a519/api/src/test/java/org/openmrs/validator/PersonValidatorTest.java#L216


Or how would you expect that test to check or behave if both fields are not set

api/src/main/resources/liquibase-update-to-2.2.xml
fruether: Then I will adapt it accordingly

api/src/test/java/org/openmrs/validator/PersonValidatorTest.java
wluyima: I think that test satisfies it then

dkayiwa:
@fruether can you change the pull request title to match what is advised here? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

fruether:
@dkayiwa @wluyima Thank you for the review. I was applying the changes you were commenting on. Can you please verify that the PR is good to to go now or point out what else is missing?


In another PR I am trying to create a DatabaseUpgrade Test for 2.2.0. I would like to add a test case for this model extension there as well. But since the process takes quite a bit I would prefer to merge this change to the data model and then in the next step we can create an according DatabaseUpgrade test. Would this be fine to you?

As test case I was starting the open:mrs with jetty as shown in the Screen shots above.

(Related PR: https://github.com/openmrs/openmrs-core/pull/2694)

wluyima:
My comments have been addressed, thanks!

api/src/main/java/org/openmrs/validator/PersonValidator.java
```diff
@@ -103,7 +104,12 @@ public void validate(Object target, Errors errors) {
 			ValidationUtils.rejectIfEmptyOrWhitespace(errors, "voidReason", "error.null");
 		}
 		if (person.getDead()) {
-			ValidationUtils.rejectIfEmpty(errors, "causeOfDeath", "Person.dead.causeOfDeathNull");
+			if(person.getCauseOfDeath() != null && person.getCauseOfDeathNonCoded() != null) {
+				errors.rejectValue("causeOfDeath", "Person.shouldHaveOnlyOneCauseOfDeathOrCauseOfDeathNonCodedSet");
```
dkayiwa: Do we have a message with the above code in the message properties files?

api/src/main/java/org/openmrs/validator/PersonValidator.java
```diff
@@ -103,7 +104,12 @@ public void validate(Object target, Errors errors) {
...
+			}
+			else if(person.getCauseOfDeath() == null && person.getCauseOfDeathNonCoded() == null) {
+				errors.rejectValue("causeOfDeath", "Person.dead.causeOfDeathNull");
```
dkayiwa: Do we have a message with the above code in the message properties files?

fruether:
@dkayiwa  Yes the message properties are set now.

api/src/main/resources/messages.properties
```diff
@@ -1793,7 +1793,7 @@ Person.birthdate=Birthdate
 Person.birthdateEstimated=Estimated
 Person.dead=Deceased
 Person.dead.checkboxInstructions=Check if this person is deceased
-Person.dead.causeOfDeathNull=Cause of Death is required for people marked as deceased.
```
dkayiwa: Is there any reason why you removed the above? I thought you were just going to reuse it or add your own new messages.

fruether: Yes @dkayiwa 

CauseOfDeathNull does not reflect the new business logic accordingly since it is completely fine for cause of death to be null as long as cause of death (non coded) is set. 
That's why I renamed this property to reflect the current implementation better. In addition,I changed the text of the error message to describe the just explained circumstances

 Does this make sense?

api/src/main/java/org/openmrs/Person.java
```diff
@@ -325,6 +327,29 @@ public void setCauseOfDeath(Concept causeOfDeath) {
 		this.causeOfDeath = causeOfDeath;
 	}
 	
+	/**
+	 * This method returns the value of the causeOfDeathNonCoded attribute
+	 * @return cause of person's death non coded
```
dkayiwa: It would be better not to make the description be a mere method name with spaces. For instance, instead of "cause of person's death non coded", why not something like "non coded cause of death"?

dkayiwa: Likewise, instead of "This method returns the value of the causeOfDeathNonCoded attribute" why not something like "Gets the non coded cause of death"?

dkayiwa: It would also be great to be a line after the method description.

fruether:
@dkayiwa I applied the comments of your review! Thanks for the feedback

PR #2699, opened 2018-08-07
TRUNK-5416: Added UserSessionListener

kaweesi:
see discussion at: https://talk.openmrs.org/t/context-hasprivilege-is-too-busy/19237/2
https://issues.openmrs.org/browse/TRUNK-5416

### DIFF
--- a/api/src/main/java/org/openmrs/UserSessionListener.java
+++ b/api/src/main/java/org/openmrs/UserSessionListener.java
@@ -21,13 +21,13 @@ import org.openmrs.api.context.Context;
  */
 public interface UserSessionListener {
 
-  public void loggedInOrOut(User user, Event event, Status status);
+	public void loggedInOrOut(User user, Event event, Status status);
 
-  public enum Event {
-    LOGIN, LOGOUT
-  }
+	public enum Event {
+		LOGIN, LOGOUT
+	}
 
-  public enum Status {
-    SUCCESS, FAIL
-  }
+	public enum Status {
+		SUCCESS, FAIL
+	}
 }
--- a/api/src/main/java/org/openmrs/UserSessionListener.java
+++ b/api/src/main/java/org/openmrs/UserSessionListener.java
@@ -1,0 +1,23 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs;
+
+public interface UserSessionListener {
+
+  public void loggedInOrOut(User user, Event event, Status status);
+  
+  public enum Event {
+    LOGIN, LOGOUT
+  }
+  
+  public enum Status {
+    SUCCESS, FAIL
+  }
+}
--- a/api/src/main/java/org/openmrs/UserSessionListener.java
+++ b/api/src/main/java/org/openmrs/UserSessionListener.java
@@ -9,14 +9,24 @@
  */
 package org.openmrs;
 
+import org.openmrs.api.context.Context;
+
+/**
+ * Allows tracking down both login and logout events including their status
+ * 
+ * Beans implementing this class are picked up by {@link Context#logout()} and
+ * {@link Context#authenticate(String, String)}
+ * 
+ * @since 2.2
+ */
 public interface UserSessionListener {
 
   public void loggedInOrOut(User user, Event event, Status status);
-  
+
   public enum Event {
     LOGIN, LOGOUT
   }
-  
+
   public enum Status {
     SUCCESS, FAIL
   }
--- a/api/src/main/java/org/openmrs/api/UserService.java
+++ b/api/src/main/java/org/openmrs/api/UserService.java
@@ -553,7 +553,6 @@ public interface UserService extends OpenmrsService {
 	 */
 	@Authorized
 	public void changePasswordUsingSecretAnswer(String secretAnswer, String pw) throws APIException;
-
-  void notifyUserSessionListener(User user, Event event, Status status);
-
+	
+	public void notifyUserSessionListener(User user, Event event, Status status);
 }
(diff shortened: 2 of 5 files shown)


wluyima:
I think it would be nice to have this behavior discussed on a design call and be sure we have a proper and well thought approach.

api/src/main/java/org/openmrs/UserSessionListener.java
```diff
@@ -0,0 +1,23 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs;
+
```
wluyima: It would be nice to add javadocs at the class level and method(s)

wluyima: Missing @ since

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -552,4 +554,6 @@
 	@Authorized
 	public void changePasswordUsingSecretAnswer(String secretAnswer, String pw) throws APIException;
 
+  void notifyUserSessionListener(User user, Event event, Status status);
```
wluyima: I'd think this being internal logic, this should be private

wluyima: And I'd expect it to be in the same class as the code that call it

api/src/test/java/org/openmrs/api/db/ContextDAOTest.java
```diff
@@ -329,4 +342,56 @@ public void should_mergeDefaultRuntimeProperties() {
 		dao.mergeDefaultRuntimeProperties(properties);
 		Assert.assertNotNull(properties.getProperty("hibernate.key"));
 	}
+	
+
+  @Component("testUserSessionListener")
+  public static class TestUserSessionListener implements UserSessionListener {
+    public Set<String> logins = new LinkedHashSet<>();
+    public Set<String> logouts = new LinkedHashSet<>();
+
```
wluyima: I think it would be nice for the tests to check that the correct Status is passed to the listeners

api/src/main/java/org/openmrs/api/UserService.java
kaweesi: its rather tricky injecting beans into a non spring UserService managed class

wluyima: You call `Context.getRegisteredComponents(UserSessionListener.class)`

dkayiwa:
We recommend including a link to the ticket as per https://wiki.openmrs.org/display/docs/Pull+Request+Tips

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -453,4 +456,40 @@ private void setUserLocation() {
...
+   * It is called by {@link UserContext#hasPrivilege(java.lang.String)}.
+   * 
+   * @see PrivilegeListener
+   * @param user the authenticated user or <code>null</code> if not authenticated
+   * @param privilege the checked privilege
+   * @param hasPrivilege <code>true</code> if the authenticated user has the required privilege or
+   *            if it is a proxy privilege
+   * @since 1.8.4, 1.9.1, 1.10
+   */
+  @Transactional(readOnly = true)
```
wluyima: Given that the method is now private, I believe this is now redundant

kaweesi: part of: https://github.com/openmrs/openmrs-core/pull/2699/commits/82aecd86b9d917b4443f24a7ca8ddbf6c886a4d8

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -96,8 +100,15 @@ public User authenticate(String username, String password, ContextDAO contextDAO
 		if (log.isDebugEnabled()) {
 			log.debug("Authenticating with username: " + username);
 		}
-		
-		this.user = contextDAO.authenticate(username, password);
+		try {
+		  this.user = contextDAO.authenticate(username, password);
```
suthagar23: Somehow, I thought the indentation is missed here.

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -96,8 +100,15 @@ public User authenticate(String username, String password, ContextDAO contextDAO
...
+		  notifyUserSessionListener(this.user, Event.LOGIN, Status.SUCCESS);
+		} catch(ContextAuthenticationException e) {
+		  User loggingInUser = new User();
+	    loggingInUser.setUsername(username);
```
suthagar23: Somehow, I thought the indentation is missed here.

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -191,6 +202,7 @@ public boolean isAuthenticated() {
 	 */
 	public void logout() {
 		log.debug("setting user to null on logout");
+    notifyUserSessionListener(user, Event.LOGOUT, Status.SUCCESS);
```
suthagar23: Indentation is missed here. Can you please recheck all the changes for the indentation. 
It caused default by the IDE also

kaweesi: o, i forgot, my eclipse was set on a 2 space indention in contrast to our tab indention. good catch @suthagar23

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -96,8 +100,15 @@ public User authenticate(String username, String password, ContextDAO contextDAO
 		if (log.isDebugEnabled()) {
 			log.debug("Authenticating with username: " + username);
 		}
-		
-		this.user = contextDAO.authenticate(username, password);
+		try {
+			this.user = contextDAO.authenticate(username, password);
+			notifyUserSessionListener(this.user, Event.LOGIN, Status.SUCCESS);
+		} catch(ContextAuthenticationException e) {
+			User loggingInUser = new User();
```
suthagar23: Actually, I couldn't understand this method. Why you initiate a new user when the `contextDAO.authenticate()` throws a `ContextAuthenticationException` ?

PR #2701, opened 2018-08-13
Added send mail functionality to send token via mail

fanyui:

Add possibility for sending email with activation key 
## Issue I worked on

- [ x] My pull request only contains **ONE single commit**

- [x ] I have **added tests** to cover my changes. (If you refactored

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/UserService.java
+++ b/api/src/main/java/org/openmrs/api/UserService.java
@@ -20,6 +20,7 @@ import org.openmrs.User;
 import org.openmrs.annotation.Authorized;
 import org.openmrs.annotation.Logging;
 import org.openmrs.api.context.UserContext;
+import org.openmrs.notification.MessageException;
 import org.openmrs.util.PersonByNameComparator;
 import org.openmrs.util.PrivilegeConstants;
 
@@ -571,7 +572,7 @@ public interface UserService extends OpenmrsService {
 	 * Sets a user's activation key
 	 * @param User The user for which the activation key will be set
 	 */
-	public User setUserActivationKey(User user);
+	public User setUserActivationKey(User user) throws MessageException;
 	
 	/**
 	 * Change user password given the activation key
--- a/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
@@ -761,25 +761,21 @@ public class UserServiceImpl extends BaseOpenmrsService implements UserService {
 		String token = RandomStringUtils.randomAlphanumeric(20);
 		long time = System.currentTimeMillis() + getValidTime();
 		String hashedKey = Security.encodeString(token);
-		String activationKey = hashedKey+":"+time;
+		String activationKey = hashedKey + ":" + time;
 		LoginCredential credentials = dao.getLoginCredential(user);
 		credentials.setActivationKey(activationKey);	
 		dao.setUserActivationKey(credentials);	
 		
-		String link = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_HOST_URL) + token;
+		String link = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_HOST_URL)
+		        .replace("activationkey", "activationkey=" + token);
 		String msg = "Hi " + user.getUsername()
-		        + ", \n It looks like you like to change your password, click on the following link to do so.\n" + link
-		        + " please disregard this email if you didn't innitiate the password reset process  The openmrs Team";
-		Message mail = new Message();
-		mail.setId(20);
-		mail.setRecipients(user.getEmail());
-		mail.setSender(Context.getAdministrationService().getGlobalProperty("mail.from"));
-		mail.setContentType(Context.getAdministrationService().getGlobalProperty("mail.default_content_type"));
-		mail.setSubject("Password Reset Request");
-		mail.setContent(msg);
-		mail.setSentDate(new Date());
+		        + ",\n It looks like you want to change your password, click on the following link to do so.\n"
+		        + link
+		        + "\n Please disregard this email if you didn't innitiate the password reset process.  \n The openmrs Team";
 		
 		try {
+			Message mail = Context.getMessageService().createMessage(user.getEmail(),
+			    Context.getAdministrationService().getGlobalProperty("mail.from"), "Password Reset Request", msg);
 			Context.getMessageService().sendMessage(mail);
 		}
 		catch (MessageException e) {
--- a/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
@@ -15,7 +15,11 @@ import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 import java.util.Set;
+
+import javax.mail.PasswordAuthentication;
+import javax.mail.Session;
 
 import org.apache.commons.lang.RandomStringUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -35,6 +39,9 @@ import org.openmrs.api.context.Context;
 import org.openmrs.api.db.DAOException;
 import org.openmrs.api.db.LoginCredential;
 import org.openmrs.api.db.UserDAO;
+import org.openmrs.notification.Message;
+import org.openmrs.notification.MessageException;
+import org.openmrs.notification.mail.MailMessageSender;
 import org.openmrs.patient.impl.LuhnIdentifierValidator;
 import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.OpenmrsUtil;
@@ -751,8 +758,8 @@ public class UserServiceImpl extends BaseOpenmrsService implements UserService {
 	}
 	
 	/**
+	 * @throws MessageException
 	 * @see org.openmrs.api.UserService#setUserActivationKey(org.openmrs.User)
-	 * 
 	 */
 	@Override
 	public User setUserActivationKey(User user) {
@@ -764,9 +771,58 @@ public class UserServiceImpl extends BaseOpenmrsService implements UserService {
 		credentials.setActivationKey(activationKey);	
 		dao.setUserActivationKey(credentials);	
 			
-		//Send Email with unhashed  activation key and Date:
-		
+		//Send Email with unhashed  activation key 
+		sendMail(user, token);
 		return user;
+	}
+	
+	public void sendMail(User user, String token) {
+		System.setProperty("https.protocols", "TLSv1,TLSv1.1,TLSv1.2");
+		Properties props = new Properties();
+		String mailSmtpAuth = Context.getAdministrationService().getGlobalProperty("mail.smtp_auth");
+		props.put("mail.smtp.auth", mailSmtpAuth);
+		props.put(OpenmrsConstants.GP_MAIL_SMTP_STARTTLS_ENABLE,
+		    Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_MAIL_SMTP_STARTTLS_ENABLE));
+		props.put("mail.smtp.host", Context.getAdministrationService().getGlobalProperty("mail.smtp_host"));
+		props.put("mail.smtp.port", Context.getAdministrationService().getGlobalProperty("mail.smtp_port"));
+		
+		String link = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_HOST_URL) + token;
+		String msg = "Hi " + user.getUsername()
+		        + ", \n It looks like you like to change your password, click on the following link to do so.\n" + link
+		        + " please disregard this email if you didn't innitiate the password reset process  The openmrs Team";
+		Message mail = new Message();
+		mail.setId(20);
+		mail.setRecipients(user.getEmail());
+		mail.setSender(Context.getAdministrationService().getGlobalProperty("mail.from"));
+		mail.setContentType(Context.getAdministrationService().getGlobalProperty("mail.default_content_type"));
+		mail.setSubject("Password Reset Request");
+		mail.setContent(msg);
+		mail.setSentDate(new Date());
+		
+		Session session;
+		MailMessageSender mailMessageSender = new MailMessageSender();
+		if (mailSmtpAuth.equalsIgnoreCase("false")) {
+			session = Session.getInstance(props);
+		} else {
+			session = Session.getInstance(props, new javax.mail.Authenticator() {
+				
+				@Override
+				protected PasswordAuthentication getPasswordAuthentication() {
+					return new PasswordAuthentication(Context.getAdministrationService().getGlobalProperty("mail.user"),
+					        Context.getAdministrationService().getGlobalProperty("mail.password"));
+				}
+			});
+		}
+		
+		mailMessageSender.setMailSession(session);
+		try {
+			mailMessageSender.send(mail);
+		}
+		catch (MessageException e) {
+			// Prevent propagation of exception to the webservice should in case message was not sent  
+			//e.printStackTrace();
+		}
+		
 	}
 
 	/**
--- a/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
@@ -9,14 +9,12 @@
  */
 package org.openmrs.api.impl;
 
-import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Properties;
 import java.util.Set;
 
 import org.apache.commons.lang.RandomStringUtils;
@@ -30,6 +28,7 @@ import org.openmrs.annotation.Authorized;
 import org.openmrs.annotation.Logging;
 import org.openmrs.api.APIAuthenticationException;
 import org.openmrs.api.APIException;
+import org.openmrs.api.AdministrationService;
 import org.openmrs.api.CannotDeleteRoleWithChildrenException;
 import org.openmrs.api.InvalidActivationKeyException;
 import org.openmrs.api.UserService;
@@ -37,9 +36,9 @@ import org.openmrs.api.context.Context;
 import org.openmrs.api.db.DAOException;
 import org.openmrs.api.db.LoginCredential;
 import org.openmrs.api.db.UserDAO;
+import org.openmrs.messagesource.MessageSourceService;
 import org.openmrs.notification.MessageException;
 import org.openmrs.patient.impl.LuhnIdentifierValidator;
-import org.openmrs.util.OpenmrsClassLoader;
 import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.OpenmrsUtil;
 import org.openmrs.util.PrivilegeConstants;
@@ -68,7 +67,7 @@ public class UserServiceImpl extends BaseOpenmrsService implements UserService {
 	private static final int MAX_VALID_TIME = 12*60*60*1000; //Period of 12 hours
 	private static final int MIN_VALID_TIME = 60*1000; //Period of 1 minute
 	private static final int DEFAULT_VALID_TIME = 10*60*1000; //Default time of 10 minute
-		
+	
 	@Autowired(required = false)
 	List<PrivilegeListener> privilegeListeners;
 	
@@ -755,11 +754,12 @@ public class UserServiceImpl extends BaseOpenmrsService implements UserService {
 	}
 	
 	/**
+	 * @throws APIException
 	 * @throws MessageException
 	 * @see org.openmrs.api.UserService#setUserActivationKey(org.openmrs.User)
 	 */
 	@Override
-	public User setUserActivationKey(User user) {
+	public User setUserActivationKey(User user) throws MessageException {
 		String token = RandomStringUtils.randomAlphanumeric(20);
 		long time = System.currentTimeMillis() + getValidTime();
 		String hashedKey = Security.encodeString(token);
@@ -768,23 +768,17 @@ public class UserServiceImpl extends BaseOpenmrsService implements UserService {
 		credentials.setActivationKey(activationKey);	
 		dao.setUserActivationKey(credentials);	
 		
-		Properties props = new Properties();
-		InputStream stream = OpenmrsClassLoader.getInstance().getResourceAsStream("emailTemplate.properties");
-		if (stream != null) {
-			OpenmrsUtil.loadProperties(props, stream);
-		}
-		
-		String link = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_HOST_URL)
+		MessageSourceService messages = Context.getMessageSourceService();
+		AdministrationService adminService = Context.getAdministrationService();
+		String link = adminService.getGlobalProperty(OpenmrsConstants.GP_HOST_URL)
 		        .replace("{activationKey}", token);
-		user.getUsername();
-		String msg = props.getProperty("mail.content").replaceAll(":name", user.getUsername()) + link;
-		try {
-			Context.getMessageService().sendMessage(user.getEmail(),
-			    Context.getAdministrationService().getGlobalProperty("mail.from"), props.getProperty("mail.subject"), msg);
-		}
-		catch (MessageException e) {
-			log.debug(e.getMessage());
-		}
+		String msg = messages.getMessage("mail.passwordreset.content").replace("{name}", user.getUsername())
+		        .replace("{link}", link)
+		        .replace("{time}", String.valueOf(getValidTime() / 60000));
+		Context.getMessageService().sendMessage(user.getEmail(),
+		    adminService.getGlobalProperty("mail.from"),
+		    messages.getMessage("mail.passwordreset.subject"), msg);
+		
 		return user;
 	}
 
--- a/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
@@ -15,11 +15,7 @@ import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Properties;
 import java.util.Set;
-
-import javax.mail.PasswordAuthentication;
-import javax.mail.Session;
 
 import org.apache.commons.lang.RandomStringUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -41,7 +37,6 @@ import org.openmrs.api.db.LoginCredential;
 import org.openmrs.api.db.UserDAO;
 import org.openmrs.notification.Message;
 import org.openmrs.notification.MessageException;
-import org.openmrs.notification.mail.MailMessageSender;
 import org.openmrs.patient.impl.LuhnIdentifierValidator;
 import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.OpenmrsUtil;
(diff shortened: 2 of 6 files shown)


fanyui:
@wluyima  This pull request sends the emai with generated activation key. so we can use it to test the rest webservice endpoint.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -764,10 +771,59 @@ public User setUserActivationKey(User user) {
 		credentials.setActivationKey(activationKey);	
 		dao.setUserActivationKey(credentials);	
 			
-		//Send Email with unhashed  activation key and Date:
-		
+		//Send Email with unhashed  activation key
```
wluyima: I think you can get rid of this comment

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -764,10 +771,59 @@ public User setUserActivationKey(User user) {
...
 		dao.setUserActivationKey(credentials);	
 			
-		//Send Email with unhashed  activation key and Date:
-		
+		//Send Email with unhashed  activation key 
+		sendMail(user, token);
 		return user;
 	}
+	
+	public void sendMail(User user, String token) {
```
wluyima: You don't need this method and all this code, I did mention there is a messaging service and you should use it to send the email.

wluyima: This should be a one line call

fanyui: Hello @wluyima  I've been looking at the MessageService to use  that service method and send the email but. when i tried it fails with errors of **Must issue a STARTTLS command first. v10-v6sm19344157wrm.18 - gsmtp** and i guess the MessageService is for only notification and alerts. as when i looked at the MailMessageSender. it has methods setMailSession which takes a session variable and creating a session requires the Propertites and an authenticator which is what i did in the code above. I was hopping to see some where. how they use the Global variables defined in the OpenmrsConstants. but i couldn't find any.
Thanks for any directions.

fanyui: I did use Context.getMessageService().sendMessage(mail) where mail is a message object ; and since no session is set message always fails

wluyima: There is a sendMessage method that doesn't take a message object, I would try that first.

fanyui: The failure is when am still building so by then am not yet logged in. i will do some digging again and see.

wluyima:
Let's not do localization for now, it will be done in the future

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -567,6 +567,7 @@ private static Session getMailSession() {
 					props.setProperty("mail.from", adminService.getGlobalProperty("mail.from"));
 					props.setProperty("mail.debug", adminService.getGlobalProperty("mail.debug"));
 					props.setProperty("mail.smtp.auth", adminService.getGlobalProperty("mail.smtp_auth"));
+					props.setProperty(OpenmrsConstants.GP_MAIL_SMTP_STARTTLS_ENABLE, adminService.getGlobalProperty(OpenmrsConstants.GP_MAIL_SMTP_STARTTLS_ENABLE));
```
wluyima: I said that you need to create a separate TRUNK ticket to address because it's a bug in core, then create a separate PR to openmrs:master instead to get it in, this change should not be in the branch for your GSoC project

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -763,9 +765,26 @@ public User setUserActivationKey(User user) {
 		LoginCredential credentials = dao.getLoginCredential(user);
 		credentials.setActivationKey(activationKey);	
 		dao.setUserActivationKey(credentials);	
-			
-		//Send Email with unhashed  activation key and Date:
 		
+		String link = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_HOST_URL) + token;
```
wluyima: this is not correct, the URL should have a placeholder to replace the token, you should not be appending it

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -763,9 +765,26 @@ public User setUserActivationKey(User user) {
...
+		String msg = "Hi " + user.getUsername()
+		        + ", \n It looks like you like to change your password, click on the following link to do so.\n" + link
```
wluyima: Somewhat a lot of repetition of **like**, can you rephrase this please?

wluyima: Do line breaks really work here? I would think html would

wluyima: For a start, add a file to the classpath containing this message and read in the contents when sending the email, in the feature it will be a template that can support localization too

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -763,9 +765,26 @@ public User setUserActivationKey(User user) {
...
+		        + " please disregard this email if you didn't innitiate the password reset process  The openmrs Team";
+		Message mail = new Message();
```
wluyima: You should not be constructing anything, please use the message service, all this should be like 2 calls to the service once you have the contents of the email.

api/src/main/java/org/openmrs/api/context/Context.java
fanyui: Ok i did it already this was just to permit me do my testing and ensure it works with respect to my gsoc project. pull request for the issue is https://github.com/openmrs/openmrs-core/pull/2703

wluyima: Then remove it please

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1390,12 +1392,17 @@ public void getUserByUsernameOrEmail_shouldFailIfEmailIsNull() {
...
 	
 	@Test
-	public void setUserActivationKey_shouldCreateUserActivationKey() {
+	public void setUserActivationKey_shouldCreateUserActivationKey() throws MessageException {
 		User u = new User();
 		u.setPerson(new Person());
 		u.addName(new PersonName("Benjamin", "A", "Wolfe"));
 		u.setUsername("bwolfe");
 		u.getPerson().setGender("M");
+		Context.getAdministrationService().setGlobalProperty("mail.smtp_auth", "false");
```
wluyima: Why is this necessary

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1390,12 +1392,17 @@ public void getUserByUsernameOrEmail_shouldFailIfEmailIsNull() {
...
+		Context.getAdministrationService().setGlobalProperty(OpenmrsConstants.GP_MAIL_SMTP_STARTTLS_ENABLE, "true");
+		Context.getAdministrationService().setGlobalProperty("mail.smtp_host", "localhost");
+		Context.getAdministrationService().setGlobalProperty("mail.smtp_port", "25");
```
wluyima: Why are these GP necessary?

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -689,6 +694,10 @@ public static String getOpenmrsProperty(String property) {
 		props.add(new GlobalProperty(GP_MAIL_SMTP_STARTTLS_ENABLE, "false",
 		        "Set to true to enable TLS encryption, else set to false"));
 		
+		props.add(new GlobalProperty(GP_HOST_URL,
+		        "http://localhost:8080/openmrs/admin/users/changePassword.form?activationkey=",
```
wluyima: You can remove this default value for now, it will be added once we update the form

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -691,9 +713,91 @@ public String getSecretQuestion(User user) throws APIException {
...
+	 * @see org.openmrs.api.UserService#getUserByActivationKey(java.lang.String)
+	 * 
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	public User getUserByActivationKey(String activationKey) {
+		LoginCredential loginCred = dao.getLoginCredentialByActivationKey(activationKey);
+		if (loginCred != null) {
+			String[] credTokens = loginCred.getActivationKey().split(":");
+			if (System.currentTimeMillis() <= Long.parseLong(credTokens[1])) {
```
suthagar23: What happens if the credTokens length is 0?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -691,9 +713,91 @@ public String getSecretQuestion(User user) throws APIException {
...
+	/**
+	 * @see org.openmrs.api.UserService#getUserByActivationKey(java.lang.String)
+	 * 
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	public User getUserByActivationKey(String activationKey) {
+		LoginCredential loginCred = dao.getLoginCredentialByActivationKey(activationKey);
+		if (loginCred != null) {
+			String[] credTokens = loginCred.getActivationKey().split(":");
```
suthagar23: what happens if the `getActivationKey()` return null for an instance?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -691,9 +713,91 @@ public String getSecretQuestion(User user) throws APIException {
...
+		mail.setContent(msg);
+		mail.setSentDate(new Date());
+		
+		try {
+			Context.getMessageService().sendMessage(mail);
+		}
+		catch (MessageException e) {
+			// Prevent propagation of exception to the webservice should in case message was not sent  
+		}
+		return user;
```
suthagar23: So up to now, How the method will indicate that the email was sent without any faults to the caller? Do we want to indicate the caller about the email success or not?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -691,9 +713,91 @@ public String getSecretQuestion(User user) throws APIException {
...
+		credentials.setActivationKey(activationKey);	
+		dao.setUserActivationKey(credentials);	
+		
+		String link = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_HOST_URL) + token;
+		String msg = "Hi " + user.getUsername()
+		        + ", \n It looks like you like to change your password, click on the following link to do so.\n" + link
+		        + " please disregard this email if you didn't innitiate the password reset process  The openmrs Team";
+		Message mail = new Message();
+		mail.setId(20);
+		mail.setRecipients(user.getEmail());
```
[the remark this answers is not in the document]
fanyui: This method expects a user and it the responsibility of the caller to provide the user. Should incase null is provided, the mail will fails with message exception which in either case we don't want to inform the user of the failure. so everything goes on well

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -691,9 +713,91 @@ public String getSecretQuestion(User user) throws APIException {
...
+		String token = RandomStringUtils.randomAlphanumeric(20);
+		long time = System.currentTimeMillis() + getValidTime();
+		String hashedKey = Security.encodeString(token);
+		String activationKey = hashedKey+":"+time;
+		LoginCredential credentials = dao.getLoginCredential(user);
+		credentials.setActivationKey(activationKey);	
+		dao.setUserActivationKey(credentials);	
+		
+		String link = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_HOST_URL) + token;
+		String msg = "Hi " + user.getUsername()
```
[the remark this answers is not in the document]
fanyui: Ok, Am working on this.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
fanyui: Users needs not know of the success or failure of the mails

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -691,9 +713,91 @@ public String getSecretQuestion(User user) throws APIException {
...
+	/**
+	 * @throws MessageException
+	 * @see org.openmrs.api.UserService#setUserActivationKey(org.openmrs.User)
+	 */
+	@Override
+	public User setUserActivationKey(User user) {
+		String token = RandomStringUtils.randomAlphanumeric(20);
+		long time = System.currentTimeMillis() + getValidTime();
+		String hashedKey = Security.encodeString(token);
+		String activationKey = hashedKey+":"+time;
```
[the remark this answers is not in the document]
fanyui: Ok

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
fanyui: Trusting the code that generation for the activation key will always be provided ie  setUserActivationKey(User user)  as that is it job.  and for this line of code to be executed the loginCred is not null which was gotten using the activation key hence we are guaranteed that if loginCred exist and is not null then activation key exist else null will be returned

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
fanyui: credTokens length cannot be null at this level because if loginCred is not null then there is an activation key of the form xddkkdud7i8d:55636737352 which ensures that credToken lenght is always greater than 0, other wise  null will be returned instead of a failure

api/src/main/java/org/openmrs/api/context/Context.java
fanyui: Ok

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
fanyui: Ok

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
fanyui: @wluyima Line breaks work instead of html <br /> tags. Please am having issues adding txt files to classpath.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -751,21 +753,34 @@ public User getUserByActivationKey(String activationKey) {
...
-		String activationKey = hashedKey+":"+time;
+		String activationKey = hashedKey + ":" + time;
 		LoginCredential credentials = dao.getLoginCredential(user);
 		credentials.setActivationKey(activationKey);	
 		dao.setUserActivationKey(credentials);	
-			
-		//Send Email with unhashed  activation key and Date:
 		
+		String link = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_HOST_URL)
+		        .replace("activationkey", "activationkey=" + token);
```
wluyima: Please try to use constants where necessary, the placeholder needs to look explicit, use something like  **{actvitationKey}**, and you should replace the placeholder with only the key, nothing else.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -751,21 +753,34 @@ public User getUserByActivationKey(String activationKey) {
...
+		        + link
+		        + "\n Please disregard this email if you didn't innitiate the password reset process.  \n The openmrs Team";
+		
+		try {
+			Message mail = Context.getMessageService().createMessage(user.getEmail(),
+			    Context.getAdministrationService().getGlobalProperty("mail.from"), "Password Reset Request", msg);
+			Context.getMessageService().sendMessage(mail);
+		}
+		catch (MessageException e) {
+			// Prevent propagation of exception to the webservice should in case message was not sent
```
wluyima: You shouldn't be swallowing the exception

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -751,21 +753,34 @@ public User getUserByActivationKey(String activationKey) {
...
+		        .replace("activationkey", "activationkey=" + token);
+		String msg = "Hi " + user.getUsername()
+		        + ",\n It looks like you want to change your password, click on the following link to do so.\n"
+		        + link
+		        + "\n Please disregard this email if you didn't innitiate the password reset process.  \n The openmrs Team";
+		
+		try {
+			Message mail = Context.getMessageService().createMessage(user.getEmail(),
+			    Context.getAdministrationService().getGlobalProperty("mail.from"), "Password Reset Request", msg);
+			Context.getMessageService().sendMessage(mail);
```
wluyima: There is actually a sendMessage method that takes in recipients, from, subject and content, I think you need to call that.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
fanyui: In the restwebservice A post for request of passwordreset  will trigger this method and that post is supposed to always return http.OK  and if the exception is not handled here it will propergate back as the response and we end up not having an ok response.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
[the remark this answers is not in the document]
suthagar23: I think, you will get NPE in L775?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
suthagar23: So if someone did not get the mail, then how can the user or System Administrator will get to know about that?
Is it good to have a debug log about this?

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -689,6 +694,9 @@ public static String getOpenmrsProperty(String property) {
 		props.add(new GlobalProperty(GP_MAIL_SMTP_STARTTLS_ENABLE, "false",
 		        "Set to true to enable TLS encryption, else set to false"));
 		
+		props.add(new GlobalProperty(GP_HOST_URL, "",
+		        "The activationKey is the parameter which will be generated and substituted. always make sure to provide a parameter name activationkey in the link you provide here "));
```
wluyima: A full stop can't be followed by a lower case letter

wluyima: This is actually wrong, don't dictate that the URL should have a parameter because that's a implementation detail

wluyima: You should provide a proper description of the GP, this doesn't feel like one

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
wluyima: You should not swallow the exception here, that is implementation logic in REST and should have nothing to do with the service layer

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -766,20 +768,22 @@ public User setUserActivationKey(User user) {
...
 		dao.setUserActivationKey(credentials);	
 		
-		String link = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_HOST_URL)
-		        .replace("activationkey", "activationkey=" + token);
-		String msg = "Hi " + user.getUsername()
-		        + ",\n It looks like you want to change your password, click on the following link to do so.\n"
-		        + link
-		        + "\n Please disregard this email if you didn't innitiate the password reset process.  \n The openmrs Team";
+		Properties props = new Properties();
+		InputStream stream = OpenmrsClassLoader.getInstance().getResourceAsStream("emailTemplate.properties");
```
wluyima: The file isn't supposed to be a properties file, otherwise you might as well move everything to the messages.properties file which already supports internationalization anyways and get rid of this file.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -751,21 +755,36 @@ public User getUserByActivationKey(String activationKey) {
...
+		
+		String link = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_HOST_URL)
+		        .replace("{activationKey}", token);
+		user.getUsername();
+		String msg = props.getProperty("mail.content").replaceAll(":name", user.getUsername()) + link;
+		try {
+			Context.getMessageService().sendMessage(user.getEmail(),
+			    Context.getAdministrationService().getGlobalProperty("mail.from"), props.getProperty("mail.subject"), msg);
+		}
+		catch (MessageException e) {
```
wluyima: Please don't catch he exception

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -751,21 +755,36 @@ public User getUserByActivationKey(String activationKey) {
...
 		String hashedKey = Security.encodeString(token);
-		String activationKey = hashedKey+":"+time;
+		String activationKey = hashedKey + ":" + time;
 		LoginCredential credentials = dao.getLoginCredential(user);
 		credentials.setActivationKey(activationKey);	
 		dao.setUserActivationKey(credentials);	
-			
-		//Send Email with unhashed  activation key and Date:
 		
+		Properties props = new Properties();
```
wluyima: The file isn't supposed to be a properties file, otherwise you might as well move everything to the messages.properties file which already supports internationalization anyways and get rid of this file.

api/src/main/resources/emailTemplate.properties
```diff
@@ -0,0 +1,2 @@
+mail.subject=password reset request
+mail.content=Dear :name \n You recently requested a password reset for your OpenMRS account. To complete the process, click the link below.\n  Please disregard this email if you didn't innitiate the password reset process.  \n
```
wluyima: Change **Dear** to **Hi**

Is name a place holder? I would rather have it as **{name}** rather than **:name**

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -571,7 +572,7 @@
 	 * Sets a user's activation key
 	 * @param User The user for which the activation key will be set
 	 */
-	public User setUserActivationKey(User user);
+	public User setUserActivationKey(User user) throws MessageException;
```
wluyima: Don't catch the exception

api/src/main/resources/emailTemplate.properties
```diff
@@ -0,0 +1,2 @@
+mail.subject=password reset request
```
wluyima: The file isn't supposed to be a properties file, otherwise you might as well move everything to the messages.properties file which already supports internationalization anyways and get rid of this file.

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1390,12 +1392,14 @@ public void getUserByUsernameOrEmail_shouldFailIfEmailIsNull() {
...
 	@Test
-	public void setUserActivationKey_shouldCreateUserActivationKey() {
+	public void setUserActivationKey_shouldCreateUserActivationKey() throws MessageException {
 		User u = new User();
 		u.setPerson(new Person());
 		u.addName(new PersonName("Benjamin", "A", "Wolfe"));
 		u.setUsername("bwolfe");
 		u.getPerson().setGender("M");
+		Context.getAdministrationService().setGlobalProperty(OpenmrsConstants.GP_HOST_URL,
+		    "http://localhost:8080/openmrs/admin/users/changePassword.form?activationkey");
```
wluyima: I thought this is supposed to be {activationKey}

api/src/main/resources/emailTemplate.properties
suthagar23: @fanyui  The `\n` character is for the terminal outputs, will it work for the HTML based email content? Can you check and confirm with `<br/>`?

This link will active for some amount of time (24hrs). Right? So better to mention that at the end of this content.

wluyima: That's what I think too, we should use html.

wluyima:
@fanyui are you still working on this?

fanyui:
@wluyima Yes am still working on it sorry for the absence, I had a fever(maleria) but have recovered. If am no longer working on a ticket i will identify.

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -689,6 +694,9 @@ public static String getOpenmrsProperty(String property) {
 		props.add(new GlobalProperty(GP_MAIL_SMTP_STARTTLS_ENABLE, "false",
 		        "Set to true to enable TLS encryption, else set to false"));
 		
+		props.add(new GlobalProperty(GP_HOST_URL, "",
+		        "The URL to redirect to after requesting for a passwordreset. Always provide a place holder in this url with name {activationKey} so that it will be substituted by the actual activation key."));
```
wluyima: The grammar is a little off around  **so that it will be substituted**,  how about **it will get substituted**?

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -689,6 +694,9 @@ public static String getOpenmrsProperty(String property) {
 		props.add(new GlobalProperty(GP_MAIL_SMTP_STARTTLS_ENABLE, "false",
 		        "Set to true to enable TLS encryption, else set to false"));
 		
+		props.add(new GlobalProperty(GP_HOST_URL, "",
+		        "The URL to redirect to after requesting for a passwordreset. Always provide a place holder in this url with name {activationKey} so that it will get substituted by the actual activation key."));
```
wluyima: I added a comment asking to rephrase this

api/src/main/resources/messages.properties
```diff
@@ -3299,3 +3299,6 @@ Module.error.cyclicDependencies=Could not start some modules because they depend
 
 Condition.conditionShouldNotBeNull=Condition is required
 Condition.clinicalStatusShouldNotBeNull=Clinical status is required
+
+mail.passwordreset.subject=Password reset request
```
wluyima: Password Reset should be enough

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
wluyima: Did you see my comment about rephrasing this?

fanyui:
@wluyima I have fixed this could you take a look?

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -689,6 +694,9 @@ public static String getOpenmrsProperty(String property) {
 		props.add(new GlobalProperty(GP_MAIL_SMTP_STARTTLS_ENABLE, "false",
 		        "Set to true to enable TLS encryption, else set to false"));
 		
+		props.add(new GlobalProperty(GP_HOST_URL, "",
+		        "The URL to redirect to after requesting for a password reset. Always provide a place holder in this url with name {activationKey} it will get substituted by the actual activation key."));
```
wluyima: You are missing a comma after {activationKey}

wluyima:
Thanks @fanyui !

fanyui:
@wluyima Am grateful for all the support, and eye opening  all-through.

wluyima:
👍

PR #2710, opened 2018-09-12
TRUNK-5424: Added Order search criteria

dsurrao:
https://issues.openmrs.org/browse/TRUNK-5424


api/src/main/java/org/openmrs/api/db/hibernate/HibernateOrderDAO.java
```diff
@@ -147,6 +150,40 @@ public Order getOrder(Integer orderId) throws DAOException {
...
+			crit.add(Restrictions.eq("careSetting", searchCriteria.getCareSetting()));
+		}
+		if (searchCriteria.getConcepts() != null && !searchCriteria.getConcepts().isEmpty()) {
+			crit.add(Restrictions.in("concept", searchCriteria.getConcepts()));
+		}
+		if (searchCriteria.getOrderTypes() != null && !searchCriteria.getOrderTypes().isEmpty()) {
+			crit.add(Restrictions.in("orderType", searchCriteria.getOrderTypes()));
+		}
+		if (searchCriteria.getActivatedOnOrBeforeDate() != null) {
+			crit.add(Restrictions.le("dateActivated", searchCriteria.getActivatedOnOrBeforeDate()));
```
mogoodrich: I didn't think about this until now, but since we called it the "getActivateOnOrBeforeDate", we should be interpreting this as a Date (without a time component) even though a Java Data includes the time component.

For instance, if we have an Order that was activated on "2018-02-01 15:52:50", and in the search criteria you passed in "2018-02-01 00:00:00", I would expect that order to be returned, but I think as the query is written currently it wouldn't?

One solution would be in this query to change the time component of the getActivatedOnOrBeforeDate variable to be the last millisecond of the current date.  And then for the getActivatedOnOrAfterDate set the time component to 00:00:00

dsurrao: Thanks @mogoodrich I just updated the PR

dsurrao:
@mogoodrich I modified these unit tests to properly test dateActivated search: getOrders_shouldGetOrdersByActivatedOnOrBeforeDate
getOrders_shouldGetOrdersByActivatedOnOrAfterDate

@Mikhail42 I added the final keyword to the fields of OrderSearchCriteria. The dates generated from the Calendar instance are to be used as date ranges in a search (ignoring their time component) and are not stored anywhere, so I don't think the time zone is necessary in this case.

Thanks to you both for your feedback!

PR #2713, opened 2018-09-17
TRUNK-5424: Post commit code review updates

dsurrao:
https://issues.openmrs.org/browse/TRUNK-5424

### DIFF
--- a/api/src/main/java/org/openmrs/parameter/OrderSearchCriteria.java
+++ b/api/src/main/java/org/openmrs/parameter/OrderSearchCriteria.java
@@ -50,8 +50,8 @@ public class OrderSearchCriteria {
 	 * Instead of calling this constructor directly, it is recommended to use {@link OrderSearchCriteriaBuilder}.
 	 * @param patient the patient the order is for
 	 * @param careSetting the care setting to match on
-	 * @param concepts the concepts defining the order must be in this collection
-	 * @param orderTypes the order types to match on must be in this collection
+	 * @param concepts the concepts to match on; if not specified, matches on all concepts
+	 * @param orderTypes the order types to match on; if not specified, matches all order types
 	 * @param activatedOnOrBeforeDate orders must have dateActivated on or before this date
 	 * @param activatedOnOrAfterDate orders must have dateActivated on or after this date
 	 * @param includeVoided whether to include the voided orders or not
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/openmrs/parameter/OrderSearchCriteria.java
```diff
@@ -50,8 +50,8 @@
 	 * Instead of calling this constructor directly, it is recommended to use {@link OrderSearchCriteriaBuilder}.
 	 * @param patient the patient the order is for
 	 * @param careSetting the care setting to match on
-	 * @param concepts the concepts defining the order must be in this collection
-	 * @param orderTypes the order types to match on must be in this collection
+	 * @param concepts the concepts defining the order must be in this collection if specified, matches on all concepts otherwise
```
wluyima: **Otherwise** seems to have been left hanging here

api/src/main/java/org/openmrs/parameter/OrderSearchCriteria.java
```diff
@@ -50,8 +50,8 @@
...
+	 * @param orderTypes the order types to match on must be in this collection if specified, matches on all order types otherwise
```
wluyima: Same here **otherwise** seems to have been left hanging here

api/src/main/java/org/openmrs/parameter/OrderSearchCriteria.java
dsurrao: @wluyima this is gramatically correct, unless you mean adding a period at the end?

api/src/main/java/org/openmrs/parameter/OrderSearchCriteria.java
dsurrao: @wluyima similarly here, "otherwise" can end a sentence.

api/src/main/java/org/openmrs/parameter/OrderSearchCriteria.java
wluyima: I feel like otherwise is redundant or gives the impression there is another behavior but this is the default, _**if specified, matches on all concepts**_ sounds enough to me, am not a native English speaker so I might be wrong.

dsurrao: how about this: **if specified, the concepts defining the order must be in this collection, otherwise matches on all concepts** ?

dsurrao: or simply: ** if specified, the concepts defining the order must be in this collection**

mogoodrich: the "otherwise" is key, because it is "otherwise"... the first phrase seems to be the one that is confusing to me, how abotu:
"the concepts to match on; if not specified, matches on all concepts"
"the order types to match on; if not specified, matches all order types"

dsurrao:
Thanks for the conciseness @mogoodrich , I've updated the PR with those comments.

mogoodrich:
Looks good, thanks @dsurrao !

PR #2714, opened 2018-09-18
TRUNK-5454: Refactor WebUtil class not to be a GlobalPropertyListener

samuelmale:
Ticket: https://issues.openmrs.org/browse/TRUNK-5127

### Note
Whats confusing me, <code>Webutil.defaultDateCache</code> has a public setter method. However, it so looks like this method ain't explicitly used anywhere I decided to get rid of it.

### DIFF
--- a/web/src/main/java/org/openmrs/web/WebUtil.java
+++ b/web/src/main/java/org/openmrs/web/WebUtil.java
@@ -16,8 +16,6 @@ import java.util.Date;
 import java.util.Locale;
 
 import org.apache.commons.lang3.StringUtils;
-import org.openmrs.GlobalProperty;
-import org.openmrs.api.GlobalPropertyListener;
 import org.openmrs.api.context.Context;
 import org.openmrs.util.Format.FORMAT_TYPE;
 import org.openmrs.util.LocaleUtility;
@@ -27,12 +25,10 @@ import org.owasp.encoder.Encode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class WebUtil implements GlobalPropertyListener {
+public class WebUtil {
 	
 	private static final Logger log = LoggerFactory.getLogger(WebUtil.class);
 	
-	private static String defaultDateCache = null;
-
 	/**
 	 * Encodes for (X)HTML text content and text attributes.
 	 *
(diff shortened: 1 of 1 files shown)


dkayiwa:
Are all these changes about using an unused field?

samuelmale:
Hhhhmm, there some things you can look at and you even feel ashamed. Since I was working on many issues at the same time, I possibly messed up somewhere with the branches. Will soon start using 'git diff master' before committing.

samuelmale:
Just addressed your comments @dkayiwa

web/src/main/java/org/openmrs/web/WebUtil.java
```diff
@@ -420,23 +418,19 @@ public boolean supportsPropertyName(String propertyName) {
...
-		WebUtil.defaultDateCache = defaultDateCache;
-	}
-	
 	/**
 	 * @see org.openmrs.api.GlobalPropertyListener#globalPropertyChanged(org.openmrs.GlobalProperty)
 	 */
 	@Override
 	public void globalPropertyChanged(GlobalProperty newValue) {
-		setDefaultDateCache(null);
+		throw new UnsupportedOperationException();
```
dkayiwa: Why throw an exception?

web/src/main/java/org/openmrs/web/WebUtil.java
```diff
@@ -420,23 +418,19 @@ public boolean supportsPropertyName(String propertyName) {
...
 	}
 	
 	/**
 	 * @see org.openmrs.api.GlobalPropertyListener#globalPropertyDeleted(java.lang.String)
 	 */
 	@Override
 	public void globalPropertyDeleted(String propertyName) {
-		setDefaultDateCache(null);
+		throw new UnsupportedOperationException();
```
dkayiwa: Why throw an exception?

web/src/main/java/org/openmrs/web/WebUtil.java
samuelmale: Well since the property `defaultDateCache` was removed prior to this issue requirements, this method was just initially nullifying that property. So this Operation is no longer supported.

web/src/main/java/org/openmrs/web/WebUtil.java
samuelmale: Well since the property `defaultDateCache` was removed prior to this issue requirements, this method was just initially nullifying that property. So this Operation is no longer supported.

web/src/main/java/org/openmrs/web/WebUtil.java
samuelmale: So since the property was removed, should I just put a comment place holder like

`// Do nothing`

samuelmale:
I see it makes no sense.

Since we removed the defaultDateCache, does it still make sense of this
class being a GlobalPropertyListener? according to my understanding, this
property should either remain could be it's used somewhere behind the
curtains, or this class ceases being a GlobalPropertyListener.

web/src/main/java/org/openmrs/web/WebUtil.java
samuelmale: @dkayiwa  I see it makes no sense.

Since we removed the defaultDateCache, does it still make sense of this
class being a GlobalPropertyListener? according to my understanding, this
property should either remain could be it's used somewhere behind the
curtains, or this class ceases being a GlobalPropertyListener.

How do you see?

dkayiwa: Other things can still be done in the globalPropertyChanged call back. Therefore it has nothing to do with defaultDateCache

samuelmale: Well I now know that globalPropertyChanged call back has nothing to do with defaultDateCache. Here I ask, why is just setting defaultDateCache to null yet it has nothing to do with it? 
And which other thing can be done with the callback?
And to which globalProp is this callback made?

samuelmale:
Lol thats what I was meaning.

dkayiwa:
@samuelmale any updates on this?

samuelmale:
@dkayiwa I made an update

dkayiwa:
I feel that removing the implemented interface is doing more than what the ticket requires.

dkayiwa:
That would mean creating a new ticket.

samuelmale:
Closing this. Made new PR https://github.com/openmrs/openmrs-core/pull/2785

PR #2723, opened 2018-10-24
Improve Stream usage.

merlinweber:

I changed the usage of Java 8 Streams to make the code more readable, effectiv
and safe. 

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5435

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/Cohort.java
+++ b/api/src/main/java/org/openmrs/Cohort.java
@@ -9,6 +9,8 @@
  */
 package org.openmrs;
 
+import java.util.NoSuchElementException;
+import java.util.function.Predicate;
 import org.apache.commons.lang3.StringUtils;
 
 import java.util.Arrays;
@@ -208,15 +210,29 @@ public class Cohort extends BaseChangeableOpenmrsData {
 	public Collection<CohortMembership> getActiveMemberships() {
 		return getActiveMemberships(new Date());
 	}
-	
+
 	/**
+	 * Searches for the {@code patients} membership and returns it.
+	 * <p>
+	 * If the patient has no membership, an empty Optional is returned.
+	 *
+	 * @param patient Holder of the searched membership.
+	 * @return Optional membership of the given patient.
+	 * @throws NoSuchElementException Thrown if the patient has no active membership.
 	 * @since 2.1.0
 	 */
 	public CohortMembership getActiveMembership(Patient patient) {
-		return getMemberships().stream().filter(m -> m.isActive() && m.getPatientId().equals(patient.getPatientId()))
-		        .collect(Collectors.toList()).get(0);
+		Predicate<CohortMembership> patientMatcher =
+			membership -> membership.getPatientId().equals(patient.getPatientId());
+
+		return getMemberships().stream()
+			.filter(CohortMembership::isActive)
+			.filter(patientMatcher)
+			.findAny()
+			.orElseThrow(NoSuchElementException::new); // Same as calling get()
 	}
-	
+
+
 	public int size() {
 		return getMemberships().stream().filter(m -> !m.getVoided() && m.getEndDate() == null).collect(Collectors.toList())
 		        .size();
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -208,15 +210,29 @@ public boolean removeMembership(CohortMembership cohortMembership) {
...
-		return getMemberships().stream().filter(m -> m.isActive() && m.getPatientId().equals(patient.getPatientId()))
-		        .collect(Collectors.toList()).get(0);
+		Predicate<CohortMembership> patientMatcher =
+			membership -> membership.getPatientId().equals(patient.getPatientId());
+
+		return getMemberships().stream()
+			.filter(CohortMembership::isActive)
+			.filter(patientMatcher)
+			.findAny()
+			.orElseThrow(NoSuchElementException::new); // Same as calling get()
```
samuelmale: The above comment looks redundant

api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -208,15 +210,29 @@ public boolean removeMembership(CohortMembership cohortMembership) {
...
 	}
-	
+
+
```
samuelmale: Why add this whitespace! Could we remove it?

api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -208,15 +210,29 @@ public boolean removeMembership(CohortMembership cohortMembership) {
...
+	 * @param patient Holder of the searched membership.
+	 * @return Optional membership of the given patient.
+	 * @throws NoSuchElementException Thrown if the patient has no active membership.
 	 * @since 2.1.0
 	 */
 	public CohortMembership getActiveMembership(Patient patient) {
-		return getMemberships().stream().filter(m -> m.isActive() && m.getPatientId().equals(patient.getPatientId()))
-		        .collect(Collectors.toList()).get(0);
+		Predicate<CohortMembership> patientMatcher =
+			membership -> membership.getPatientId().equals(patient.getPatientId());
```
samuelmale: Just for minimal and efficient code, why don't we do all this in the `#filter()`? With that we could get rid of the above.

api/src/main/java/org/openmrs/Cohort.java
samuelmale: Are you sure throwing the above Exception won't cause problems?

api/src/main/java/org/openmrs/Cohort.java
```diff
@@ -208,15 +210,29 @@ public boolean removeMembership(CohortMembership cohortMembership) {
...
+	 * <p>
+	 * If the patient has no membership, an empty Optional is returned.
+	 *
+	 * @param patient Holder of the searched membership.
+	 * @return Optional membership of the given patient.
+	 * @throws NoSuchElementException Thrown if the patient has no active membership.
 	 * @since 2.1.0
 	 */
 	public CohortMembership getActiveMembership(Patient patient) {
-		return getMemberships().stream().filter(m -> m.isActive() && m.getPatientId().equals(patient.getPatientId()))
```
samuelmale: Actually the above line you removed looks fine. You are just expected to change the way data is returned/collected after filtration. That instead of first collecting all members that qualify by `.collect(Collectors.toList())` and then get the first one, just do something like `findFirst()` the get.

api/src/main/java/org/openmrs/Cohort.java
merlinweber: Calling the `get()` method will also result
in an `NoSuchElementException` to be thrown.
The best thing would be to return the Optional
but I didn't want to make changes to the public API.

api/src/main/java/org/openmrs/Cohort.java
merlinweber: The line is actually somewhat longer and according to Joshua Blochs Effective Java, its a better practice to do it as I wrote.

api/src/main/java/org/openmrs/Cohort.java
merlinweber: This can be removed. I think it was just something that happend because of the formatting.

api/src/main/java/org/openmrs/Cohort.java
merlinweber: Its about formatting and readability which is actually rather poor in this line.

dkayiwa:
@merlinweber can you look into these merge conflicts?

PR #2741, opened 2018-10-29
GCI-142: "===" and "!==" should be used instead of "==" and "!="

sviho:
Here's a fix of one of the Sonar issues. I was asked to divide one PR into the several problems I solve. This one fixes the problem by using '===' and '!==' instead of '==' and '!=' in JS files.

## Issue I worked on
https://issues.openmrs.org/browse/GCI-142
No need in posting task instance since it's already approved

- [x] My IDE is configured to follow the [**code style**]

### DIFF
--- a/webapp/src/main/webapp/WEB-INF/view/scripts/initialsetup.js
+++ b/webapp/src/main/webapp/WEB-INF/view/scripts/initialsetup.js
@@ -15,7 +15,7 @@ function clickRadio(radioName, value) {
     var radioLength = radio.length;
     for(var i = 0; i < radioLength; i++) {
         radio[i].checked = false;
-        if(radio[i].value == value.toString()) {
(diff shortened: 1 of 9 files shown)


dkayiwa:
@sviho am a bit hesitant to merge this because of the possible bugs that we could introduce. Imagine a developer who assumed something like 2 == "2" to be true and then all the sudden we make it false by switching to the triple equals. That being said, i fully agree that this is generally a good practice for new code. Though this should not prevent your task from getting approved because you have definitely done a good job in the pull request.

Do you have any objections or thoughts in regards to the above?
I assume your GCI task is already approved. Correct?

sviho:
@dkayiwa Yes, my task was approved. It totally makes sense to me. We don't want any new bugs coming because of fixing some really small code issues. I'm closing the PR then.

PR #2791, opened 2018-12-18
Trunk 5452 - Adaption of equals for CohortMembership

fruether:
In this function I added the equal method to CohortMembership accordingly to the ticket description (https://issues.openmrs.org/browse/TRUNK-5452). In addition, I added several  test cases to CohortMembershipTest to prove that the method behaves in the intended way. A related talk topic can be found here: [Talk Thread](https://talk.openmrs.org/t/design-forum-2018-10-15-resolving-design-inconsistencies-in-cohort-membership-trunk-5379/20115/7)

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5452

- [x ] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

  No? -> execute above command

- [x ] My pull request is **based on the latest changes** of the master branch.


dkayiwa:
Do you mind making the commit message better than simply "resolving ticket id"? In other words, from the commit message, i should tell what the commit is about, without having to first look at the ticket.

fruether:
@dkayiwa I adapted the commit message. In the past I always did it like this. THat's why I did not do it in the first place.

dkayiwa:
Can you have the commit message include the ticket id as advised at? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

fruether:
@dkayiwa added message and issue number

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,73 @@ public void compareTo_shouldSort() throws Exception {
 		
 		assertThat(list, contains(startedRecently, startedLongAgo, noStartOrEnd, endedRecently, endedLongAgo, voided));
 	}
+	
+	@Test
+	public void equal_shouldReturnTrueIfObjectReferenceAreTheSame() {
+		CohortMembership cohortMembership = new CohortMembership(12);
+		
+		boolean result = cohortMembership.equals(cohortMembership);
```
dkayiwa: I guess we could just assert without the temporary variable.

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,73 @@ public void compareTo_shouldSort() throws Exception {
...
+		
+		assertTrue(result);
+	}
+	
+	@Test
+	public void equal_shouldReturnFalseIfObjectIsNull() {
+		// A null argument to equal should not lead to an exception
+		CohortMembership cohortMembership = new CohortMembership(12);
+		
+		boolean result = cohortMembership.equals(null);
```
dkayiwa: I guess we could just assert without the temporary result variable.

api/src/test/java/org/openmrs/CohortMembershipTest.java
fruether: I thought it is more OpenMRS agree on style to first execute the function under test and then perform the assertion? Anyways I changed it

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -144,4 +144,17 @@ public int compareTo(CohortMembership o) {
 		}
 		return ret;
 	}
+	
+	/**
+	 * Indicates if a given cohortMembership object is equal to this one
+	 *
```
dkayiwa: Did you just forget to add the @since annotation?

fruether: Yes. I just added it

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,71 @@ public void compareTo_shouldSort() throws Exception {
...
+		Date startDate2 = new Date(1545140935);
+		Date endDate2 = new Date(1577836800);
+		
+		
+		cohortMembershipOne.setEndDate(endDate1);
+		cohortMembershipOne.setStartDate(startDate1);
+		cohortMembershipTwo.setStartDate(startDate2);
+		cohortMembershipTwo.setEndDate(endDate2);
+		
+		boolean result = cohortMembershipOne.equals(cohortMembershipTwo);
```
dkayiwa: I guess we could just directly assert

fruether: I did so in the new commit

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,71 @@ public void compareTo_shouldSort() throws Exception {
...
```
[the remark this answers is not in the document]
fruether: I did so in the new commit

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,71 @@ public void compareTo_shouldSort() throws Exception {
...
+	public void equal_shouldReturnTrueIfStartDateEndDatePatientAreSame() {
+		
+		// Creating two logical identical CohortMemberships to test equal method
+		CohortMembership cohortMembershipOne = new CohortMembership(12);
+		CohortMembership cohortMembershipTwo = new CohortMembership(12);
+		
+		Date startDate1 = new Date(1545140935);
+		Date endDate1 = new Date(1577836800);
+		
+		Date startDate2 = new Date(1545140935);
```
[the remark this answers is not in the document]
fruether: Are you sure? Because the issue is that then we may do not see if equal() of Date works as aspected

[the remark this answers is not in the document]
fruether: Anyways I adapted it

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -144,4 +144,18 @@ public int compareTo(CohortMembership o) {
...
+	 * @since 2.3.0
+	 * Indicates if a given cohortMembership object is equal to this one
+	 * 
+	 * @param otherCohortMembership is a CohortMembership object that should be checked for equality with this object
+	 * @return true if both objects are logically equal. This is the case when endDate, startDate and patientId are equal  
+	 */
+	public boolean equals(CohortMembership otherCohortMembership) {
+		if(this == otherCohortMembership) return true;
+		
+		return otherCohortMembership != null && this.endDate.equals(otherCohortMembership.getEndDate()) && this.startDate.equals(otherCohortMembership.getStartDate())
```
dkayiwa: What happens if any of these properties is null?

fruether: You are right! I did not took this into account. I changed the code and added two new tests. Are two tests alright or should I add further test cases dealing with null values?

dkayiwa: Those tests would not hurt. 😊

fruether: All possible cases should be tested now!

api/src/main/java/org/openmrs/CohortMembership.java
```diff
@@ -144,4 +144,22 @@ public int compareTo(CohortMembership o) {
...
+	 * Indicates if a given cohortMembership object is equal to this one
+	 * 
+	 * @param otherCohortMembership is a CohortMembership object that should be checked for equality with this object
+	 * @return true if both objects are logically equal. This is the case when endDate, startDate and patientId are equal  
+	 */
+	public boolean equals(CohortMembership otherCohortMembership) {
+		if(this == otherCohortMembership) return true;
+		
+		return otherCohortMembership != null && 
+			((endDate != null ) ? this.endDate.equals(otherCohortMembership.getEndDate()) : otherCohortMembership.getEndDate() == null)
```
dkayiwa: Any reason why you sometimes use endDate and others you use this.endDate?
At first appearance, it looked as if endDate is a local variable different from this.endDate

fruether: "this." is not necessary since only the class attributes exist with this name (no local variable). 
At first I thought it would make it more obvious that "this" class member is checked for equality with the attribute. But since the statement is already quite long to avoid NullPointerExceptions I think the better way to go is to remove the "this" in this method.

This was done in the new commit

dkayiwa:
@fruether is this of any help? http://www.javapractices.com/topic/TopicAction.do?Id=17

fruether:
@dkayiwa 
> @fruether is this of any help? http://www.javapractices.com/topic/TopicAction.do?Id=17
I am not sure if I got correctly what the intension of this link is.
Do you want to point out that hashCode() has to be implemented as well? If so I did not do this so far since the issue just said equal and I am not sure what the best way would be to implement hashCode so that is will work


Regarding the comparission guideline: "possibly-null object fields : use both == and equals" since  -> I used equals() but this should for Date and Int also check the == part and last but not least an identical object should also be equal. The null thing is also catched.

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,146 @@ public void compareTo_shouldSort() throws Exception {
...
+		cohortMembershipOne.setStartDate(startDate);
+		cohortMembershipTwo.setStartDate(startDate);
+		cohortMembershipTwo.setEndDate(endDate);
+		
+		assertTrue(cohortMembershipOne.equals(cohortMembershipTwo));
+		
+	}
+	
+	@Test
+	public void equal_shouldReturnTrueIfBothPatientIdIsNull() {
```
dkayiwa: Should return true or false?

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,146 @@ public void compareTo_shouldSort() throws Exception {
...
+		cohortMembershipOne.setEndDate(endDate1);
+		cohortMembershipOne.setStartDate(startDate1);
+		cohortMembershipTwo.setStartDate(startDate2);
+		cohortMembershipTwo.setEndDate(endDate2);
+		
+		assertFalse(cohortMembershipOne.equals(cohortMembershipTwo));
+	}
+	
+	@Test
+	public void equal_shouldReturnFalseIfEndDateIsNull() {
```
dkayiwa: The method names seems to suggest that this would return true if the end date was not null

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,146 @@ public void compareTo_shouldSort() throws Exception {
...
+		cohortMembershipOne.setEndDate(endDate1);
+		cohortMembershipOne.setStartDate(startDate1);
+		cohortMembershipTwo.setStartDate(startDate2);
+		cohortMembershipTwo.setEndDate(endDate2);
+		
+		assertFalse(cohortMembershipOne.equals(cohortMembershipTwo));
+	}
+	
+	@Test
+	public void equal_shouldReturnFalseIfStartDateIsNull() {
```
dkayiwa: The method name seems to suggest that this would return true if start date was not null

fruether:
@dkayiwa rephrased the method names :)

api/src/test/java/org/openmrs/CohortMembershipTest.java
fruether: Fixed it

api/src/test/java/org/openmrs/CohortMembershipTest.java
fruether: Fixed it

api/src/test/java/org/openmrs/CohortMembershipTest.java
fruether: Fixed it

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,146 @@ public void compareTo_shouldSort() throws Exception {
...
+		cohortMembershipOne.setStartDate(startDate);
+		cohortMembershipTwo.setStartDate(startDate);
+		cohortMembershipTwo.setEndDate(endDate);
+		
+		assertTrue(cohortMembershipOne.equals(cohortMembershipTwo));
+		
+	}
+	
+	@Test
+	public void equal_shouldReturnTrueIfObjectsAreTheSameIncludingNullValues() {
```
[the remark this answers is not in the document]
fruether: The thing is I want to distinguish the equal tests that contain Null values in their objects and that have all attributes set properly.

That is why I wanted to add a phrase to point out that the test actually tests if equal works when some attributes are null.

Which name do you think would make the most sense:

1. equal_shouldWorkWhenNullValuesExistInObject()
1. equal_shouldReturnTrueWhenBothPatientIdsAreNull()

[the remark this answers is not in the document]
dkayiwa: The second one is more precise.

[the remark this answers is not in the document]
fruether: Done

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,146 @@ public void compareTo_shouldSort() throws Exception {
...
+		cohortMembershipOne.setStartDate(startDate);
+		cohortMembershipTwo.setStartDate(startDate);
+		cohortMembershipTwo.setEndDate(endDate);
+		
+		assertTrue(cohortMembershipOne.equals(cohortMembershipTwo));
+		
+	}
+	
+	@Test
+	public void equal_shouldReturnTrueWhenBothPatientIdsAreNull() {
```
dkayiwa: Does this mean it will return true if both patient ids are null even if the other fields are different?

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,146 @@ public void compareTo_shouldSort() throws Exception {
...
+		cohortMembershipOne.setEndDate(endDate1);
+		cohortMembershipOne.setStartDate(startDate1);
+		cohortMembershipTwo.setStartDate(startDate2);
+		cohortMembershipTwo.setEndDate(endDate2);
+		
+		assertFalse(cohortMembershipOne.equals(cohortMembershipTwo));
+	}
+	
+	@Test
+	public void equal_shouldReturnTrueIfAllDatesAreNull() {
```
dkayiwa: Should return true if all dates are null even if the patients are different?

dkayiwa: When you say "all dates" does it also include dateCreated, dateChanged and dateVoided?

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,146 @@ public void compareTo_shouldSort() throws Exception {
...
+		assertFalse(cohortMembershipOne.equals(cohortMembershipTwo));
+	}
+	
+	@Test
+	public void equal_shouldReturnFalseIfOnlyOneEndDateIsNull() {
+		
+		CohortMembership cohortMembershipOne = new CohortMembership(13);
+		CohortMembership cohortMembershipTwo = new CohortMembership(12);
+		
+		Date startDate1 = new Date(1577836801);
```
dkayiwa: Should we duplicate 1577836801?

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,146 @@ public void compareTo_shouldSort() throws Exception {
...
+		cohortMembershipOne.setEndDate(endDate1);
+		cohortMembershipOne.setStartDate(startDate1);
+		cohortMembershipTwo.setStartDate(startDate2);
+		cohortMembershipTwo.setEndDate(endDate2);
+		
+		assertFalse(cohortMembershipOne.equals(cohortMembershipTwo));
+	}
+	
+	@Test
+	public void equal_shouldReturnFalseIfOnlyOneStartDateIsNull() {
```
dkayiwa: Should return false when one start date is null, even if all the other fields are the same? Or did you want to mean something else?

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,146 @@ public void compareTo_shouldSort() throws Exception {
...
+		cohortMembershipOne.setEndDate(endDate1);
+		cohortMembershipOne.setStartDate(startDate1);
+		cohortMembershipTwo.setStartDate(startDate2);
+		cohortMembershipTwo.setEndDate(endDate2);
+		
+		assertFalse(cohortMembershipOne.equals(cohortMembershipTwo));
+	}
+	
+	@Test
+	public void equal_shouldReturnFalseIfOnlyOneEndDateIsNull() {
```
dkayiwa: Should return false when end date is null, even if all the other fields are the same? Or did you want to mean something else?

api/src/test/java/org/openmrs/CohortMembershipTest.java
fruether: @dkayiwa I corrected the name

api/src/test/java/org/openmrs/CohortMembershipTest.java
fruether: @dkayiwa I corrected the name

api/src/test/java/org/openmrs/CohortMembershipTest.java
fruether: Pointed out now that it is only False if end Date is null for one object:

equal_shouldReturnFalseIfOnlyEndDateOfOneObjectIsNull

api/src/test/java/org/openmrs/CohortMembershipTest.java
fruether: Done

api/src/test/java/org/openmrs/CohortMembershipTest.java
fruether: equal_shouldReturnFalseIfOnlyStartDateOfOneObjectIsNull

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,147 @@ public void compareTo_shouldSort() throws Exception {
...
+	}
+	
+	@Test
+	public void equal_shouldReturnTrueWhenBothPatientIdsAreNullAndEndDateAndStartDateAreEqual() {
+		
+		// Creating two logical identical CohortMemberships to test equal method
+		CohortMembership cohortMembershipOne = new CohortMembership();
+		CohortMembership cohortMembershipTwo = new CohortMembership();
+		
+		Date startDate1 = new Date(1545140935);
```
dkayiwa: Do wee need to duplicate 1545140935?

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,147 @@ public void compareTo_shouldSort() throws Exception {
...
+		cohortMembershipOne.setEndDate(endDate1);
+		cohortMembershipOne.setStartDate(startDate1);
+		cohortMembershipTwo.setStartDate(startDate2);
+		cohortMembershipTwo.setEndDate(endDate2);
+		
+		assertFalse(cohortMembershipOne.equals(cohortMembershipTwo));
+	}
+	
+	@Test
+	public void equal_shouldReturnFalseIfOnlyEndDateOfOneObjectIsNull() {
```
dkayiwa: end date of one object or two objects?

dkayiwa: Secondly, would it return true of the patient ids were the same?

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,147 @@ public void compareTo_shouldSort() throws Exception {
...
+		cohortMembershipOne.setEndDate(endDate1);
+		cohortMembershipOne.setStartDate(startDate1);
+		cohortMembershipTwo.setStartDate(startDate2);
+		cohortMembershipTwo.setEndDate(endDate2);
+		
+		assertFalse(cohortMembershipOne.equals(cohortMembershipTwo));
+	}
+	
+	@Test
+	public void equal_shouldReturnFalseIfOnlyStartDateOfOneObjectIsNull() {
```
dkayiwa: Only start date is null or even end date of one object?

api/src/test/java/org/openmrs/CohortMembershipTest.java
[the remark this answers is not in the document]
dkayiwa: i mean duplicating 1545140935

api/src/test/java/org/openmrs/CohortMembershipTest.java
fruether: I will make a variable out of it. And yeah I think it is helpful for testing purposes to duplicate it

dkayiwa: how would duplicating it be helpful?

api/src/test/java/org/openmrs/CohortMembershipTest.java
fruether: (1) Only EndDate of one object is null. That is why both objects are not equal since the EndDate is different
(2) I fixed it. It was actually a mistake of me

api/src/test/java/org/openmrs/CohortMembershipTest.java
fruether: Fixxd it

api/src/test/java/org/openmrs/CohortMembershipTest.java
fruether: The alternative would be to say: startDate2 = startDate1 correct?
But if I am doing so the objects point to the same area and the reference is already equal. I want to test/make sure that the equal() function of the Date type is executed. Hence to verify, if both dates have the same content. That is why I need to duplicate the value to create a new object which is equal but not the same

dkayiwa: Assigning that value to a constant always helps to avoid the duplication 😊

fruether: I always used a constant instead of duplicating the same value twice. But I am wondering: What is bad about that duplication? 

Because noone sees that these values were supposed to be the same? Does it has to be the same for the patientId in the constructors?

dkayiwa: If you changed the value of one, would you change the other too?

fruether: What do you mean? I can't follow your right now :(

dkayiwa: If you wanted to change that value from 1545140935 to 1545140936, would also change it for the second date?

fruether: Yes. it should change the value of both startDate's. The one of object1 and object2. Because I want to test if it fails, when only the endDate is different in both objects

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,156 @@ public void compareTo_shouldSort() throws Exception {
...
+		cohortMembershipTwo.setEndDate(endDate2);
+		
+		assertFalse(cohortMembershipOne.equals(cohortMembershipTwo));
+	}
+	
+	@Test
+	public void equal_shouldReturnTrueIfPatientIdIsEqualAndStarDateAndEndDateAreBothNull() {
+		
+		// Creating two logical identical CohortMemberships to test equal method
+		int patientId = 12;
```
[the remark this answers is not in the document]
fruether: made it final

api/src/test/java/org/openmrs/CohortMembershipTest.java
```diff
@@ -252,4 +252,156 @@ public void compareTo_shouldSort() throws Exception {
...
+	}
+	
+	@Test
+	public void equal_shouldReturnTrueWhenBothPatientIdsAreNullAndEndDateAndStartDateAreEqual() {
+		
+		// Creating two logical identical CohortMemberships to test equal method
+		CohortMembership cohortMembershipOne = new CohortMembership();
+		CohortMembership cohortMembershipTwo = new CohortMembership();
+		
+		long timestampStart = 1545140935;
```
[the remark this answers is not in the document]
fruether: "Any constants should be written in uppercase letters." I made it upper case and new words seperated by _

PR #2800, opened 2019-01-02
TRUNK-5468 Remove not needed casts

michalfotyga:

Removed not needed castings reported by IntelliJ.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5468

- [X] My pull request only contains **ONE single commit**

### DIFF
--- a/api/src/main/java/org/openmrs/util/OpenmrsUtil.java
+++ b/api/src/main/java/org/openmrs/util/OpenmrsUtil.java
@@ -624,10 +624,10 @@ public class OpenmrsUtil {
 			return d1.compareTo(d2);
 		}
 		if (d1 instanceof Timestamp) {
-			d1 = new Date(((Timestamp) d1).getTime());
+			d1 = new Date(d1.getTime());
 		}
 		if (d2 instanceof Timestamp) {
-			d2 = new Date(((Timestamp) d2).getTime());
+			d2 = new Date(d2.getTime());
 		}
 		return d1.compareTo(d2);
 	}
--- a/api/src/main/java/org/openmrs/ProviderAttribute.java
+++ b/api/src/main/java/org/openmrs/ProviderAttribute.java
@@ -31,7 +31,7 @@ public class ProviderAttribute extends BaseAttribute<ProviderAttributeType, Prov
 	}
 	
 	public Provider getProvider() {
-		return (Provider) getOwner();
+		return getOwner();
 	}
 	
 	public void setProvider(Provider provider) {
--- a/api/src/main/java/org/openmrs/api/PasswordException.java
+++ b/api/src/main/java/org/openmrs/api/PasswordException.java
@@ -32,7 +32,7 @@ public class PasswordException extends APIException {
 	}
 	
 	public PasswordException(String message, Throwable cause) {
-		super(message, (Object[]) null, cause);
+		super(message, null, cause);
 	}
 	
 	public PasswordException(String message) {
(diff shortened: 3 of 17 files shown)


dkayiwa:
Thanks @michalfotyga for working on this. 😊 
I have just made the ticket ready for work. can you take it through the workflow mentioned here? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -624,10 +624,10 @@ public static int compare(Date d1, Date d2) {
 			return d1.compareTo(d2);
 		}
 		if (d1 instanceof Timestamp) {
-			d1 = new Date(((Timestamp) d1).getTime());
```
dkayiwa: Is Timestap.getTime() the same as Date.getTime()?

michalfotyga: Locally I have run a test
```java
import java.sql.Timestamp;
import java.util.Date;

@Test
void getTimeOfSqlTimestampVsJavaUtilDate() {
	long time = System.currentTimeMillis();
	Timestamp timestamp = new Timestamp(time);
	Date date = new Date(time);
	assertThat(timestamp.getTime()).isEqualTo(date.getTime());
}
```
and it's green

dkayiwa: Did you look at the actual implementations of the two methods?

michalfotyga: Yes, I know that the `Timestamp` has nanoseconds.
I have a feeling this code is doing something different then it is supposed to do. Please have a look.

There are two things happening here:
- not needed casts,
- logic that possibly does a different thing than intended

This PR is about a not needed cast:
```java
	if (d1 instanceof Timestamp) {
        // if we are here then it means that  d1 is an instance of a Timestamp
        // so why casting it into a Timestamp? Even without the cast, a Timestamp's getTime() will be called - not the getTime() from a parent class(Date)
        // it is polymorphism in action. 
		d1 = new Date(((Timestamp) d1).getTime());
	}
```
  
  
Now about the other point.
There are four scenarios based on the parameters passed to this method:
- `Date, Date`
- `Timestamp, Timestamp`,
- `Date, Timestamp`,
- `Timestamp, Date`

For `Date, Date` and `Timestamp, Timestamp` everything is ok.

Now let's check `d1=Date, d2=Timestamp`
```java
public static int compare(Date d1, Date d2) {
	if (d1 instanceof Timestamp && d2 instanceof Timestamp) {
        // not entered
	}

	if (d1 instanceof Timestamp) {
        // not entered
		d1 = new Date(((Timestamp) d1).getTime());
	}
	if (d2 instanceof Timestamp) {
        //d2.getTime() will return nanoseconds and they will be passed to a Date's constructor. This I think may create a date in the future..
        d2 = new Date(((Timestamp) d2).getTime());
	}

    //and because d2 is now a date in the future, this comparison will fail.
	return d1.compareTo(d2);
}
```
Reversed scenario, the one for `Timestamp, Date` will work the same.

But that's just my theory. Tests are needed to verify this so maybe let's do this:
- exclude these two casts from PR
- create a new task to provide tests covering those 4 scenarios and then we can come back to the idea of removing the casts from it. That's the safe way to go 💪 
What do you think about adding a separate task for writing tests for this method?

dkayiwa: I like your reasoning. Makes a lot of sense to me. 😊

PR #2820, opened 2019-02-02
TRUNK-5087 Added tests for Openmrsutil.java.

Andu033:
-Added tests for classes untested on coveralls .
-Solved a TODO for conceptListHelper 
-Rewrite conceptSetHelper because was all duplicate code.

### DIFF
--- a/api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
+++ b/api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
@@ -19,21 +19,28 @@ import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
+
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.FileSystems;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Locale;
 import java.util.Properties;
+import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.Collection;
@@ -43,10 +50,12 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
+import org.openmrs.Concept;
 import org.openmrs.GlobalProperty;
 import org.openmrs.PatientIdentifier;
 import org.openmrs.PatientIdentifierType;
 import org.openmrs.User;
+import org.openmrs.api.ConceptService;
 import org.openmrs.api.InvalidCharactersPasswordException;
 import org.openmrs.api.ShortPasswordException;
 import org.openmrs.api.WeakPasswordException;
@@ -62,6 +71,19 @@ public class OpenmrsUtilTest extends BaseContextSensitiveTest {
 	private static GlobalProperty luhnGP = new GlobalProperty(
 	        OpenmrsConstants.GLOBAL_PROPERTY_DEFAULT_PATIENT_IDENTIFIER_VALIDATOR,
 	        OpenmrsConstants.LUHN_IDENTIFIER_VALIDATOR);
+	
+	private ConceptService cs;
+	private final String separator = FileSystems.getDefault().getSeparator();
+	
+	@Before
+	public void setup() {
+		cs=Context.getConceptService();
+	}
+	private File createFileInUserDirectory(String relativePath) {
+		String userDirectoryPath = System.getProperty("user.dir");
+		File file = new File(userDirectoryPath + separator + relativePath);
+		return file;
+	}
 	
 	/**
 	 * @throws Exception
@@ -666,8 +688,155 @@ public class OpenmrsUtilTest extends BaseContextSensitiveTest {
 	}
 	
 	/**
-	 * @see OpenmrsUtil#nullSafeEqualsIgnoreCase(String,String)
+	 * @see OpenmrsUtil#folderContains(File, String)
 	 */
+	@Test
+	public void folderContains_shouldReturnTrueWhenFolderContainsFileWithGivenName() {
+		File folder = createFileInUserDirectory("folder");
+		folder.mkdir();
+		File file = createFileInUserDirectory("folder" + separator + "file");
+		try {
+			file.createNewFile();
+			assertTrue(OpenmrsUtil.folderContains(folder, "file"));
+			file.delete();
+
+		} 
+		catch (IOException e) {Assert.fail("Unable to create file");}
+		folder.delete();
+	}
+	
+	/**
+	 * @see OpenmrsUtil#folderContains(File, String)
+	 */
+	@Test
+	public void folderContains_shouldReturnFalseWhenFolderDoesNotContainFileWithGivenName() {
+		File folder = createFileInUserDirectory("folder");
+		folder.mkdir();
+		assertFalse(OpenmrsUtil.folderContains(folder,"file"));
+		folder.delete();
+	}
+	
+	/**
+	 * @see OpenmrsUtil#folderContains(File, String)
+	 */
+	@Test
+	public void folderContains_shouldReturnFalseWhenGivenStringIsNull() {
+		File folder = createFileInUserDirectory("folder");
+		folder.mkdir();
+		assertFalse(OpenmrsUtil.folderContains(folder, null));
+		folder.delete();
+	}
+	
+	/**
+	 * @see OpenmrsUtil#folderContains(File, String)
+	 */
+	@Test
+	public void folderContains_shouldReturnFalseWhenGivenFolderIsNull() {
+		assertFalse(OpenmrsUtil.folderContains(null,"file"));
+	}
+	
+	/**
+	 * @see OpenmrsUtil#conceptListHelper(String)
+	 */
+	@Test
+	public void conceptListHelper_shouldReturnEmptyListWhenGivenDescriptorIsNullOrEmpty() {
+		assertTrue(OpenmrsUtil.conceptListHelper("").isEmpty());
+		assertTrue(OpenmrsUtil.conceptListHelper(null).isEmpty());
+	}
+
+	/**
+	 * @see OpenmrsUtil#conceptListHelper(String)
+	 */
+	@Test
+	public void conceptListHelper_shouldReturnListOfConceptsWithGivenNameInDescriptor() {
+		List<Concept> helperList = OpenmrsUtil.conceptListHelper("name:YES|name:NO|name:DIED|set:25|set:26");
+		List<Concept> serviceList = new ArrayList<>();
+		Concept concept_auxiliar;
+		//Adding the concepts in the list to compare with the first one
+		serviceList.add(cs.getConceptByName("YES"));
+		serviceList.add(cs.getConceptByName("NO"));
+		serviceList.add(cs.getConceptByName("DIED"));
+		
+		concept_auxiliar = cs.getConcept(25);
+		serviceList.addAll(cs.getConceptsByConceptSet(concept_auxiliar));
+		
+		concept_auxiliar = cs.getConcept(26);
+		serviceList.addAll(cs.getConceptsByConceptSet(concept_auxiliar));
+		
+		assertTrue("List should be same size to be equal",helperList.size() == serviceList.size());
+		//Removing common elements
+		Set<Concept> serviceSet = new HashSet<>(serviceList);
+		helperList.stream()
+				.forEach(concept -> {
+					serviceSet.remove(concept);
+				});
+		
+		assertTrue("The set should be empty" ,serviceSet.isEmpty());
+	}
+	
+	/**
+	 * @see OpenmrsUtil#getFileAsString(File)
+	 */
+	@Test
+	public void getFileAsString_shouldReturnTheStringContainedInFile() {
+		try {
+			File file = createFileInUserDirectory("file");
+			file.createNewFile();
+			PrintWriter writer = new PrintWriter(file);
+			String output = "The first line" + "\n" + "The second line";
+			writer.println("The first line");
+			writer.print("The second line");
+			writer.close();
+			assertTrue(output.equals(OpenmrsUtil.getFileAsString(file)));
+			file.delete();
+			
+		}
+		catch (IOException e) {}
+		
+	}
+
+	/**
+	 * @see OpenmrsUtil#conceptSetHelper(String) 
+	 */
+	@Test
+	public void conceptSetHelper_shouldReturnSetOfConceptsWithGivenNameInDescriptor() {
+		Set<Concept> helperSet = OpenmrsUtil.conceptSetHelper("name:YES|name:NO|name:DIED|set:25|set:26");
+		Set<Concept> serviceSet = new HashSet<>();
+		Concept concept_auxiliar;
+		//Adding the concepts in the list to compare with the first one
+		serviceSet.add(cs.getConceptByName("YES"));
+		serviceSet.add(cs.getConceptByName("NO"));
+		serviceSet.add(cs.getConceptByName("DIED"));
+
+		concept_auxiliar = cs.getConcept(25);
+		serviceSet.addAll(cs.getConceptsByConceptSet(concept_auxiliar));
+
+		concept_auxiliar = cs.getConcept(26);
+		serviceSet.addAll(cs.getConceptsByConceptSet(concept_auxiliar));
+
+		assertTrue("List should be same size to be equal",helperSet.size() == serviceSet.size());
+		//Removing common elements
+
+		helperSet.stream()
+			.forEach(concept -> {
+				serviceSet.remove(concept);
+			});
+
+		assertTrue("The set should be empty" ,serviceSet.isEmpty());
+	}
+
+	/**
+	 * @see OpenmrsUtil#conceptListHelper(String)
+	 */
+	@Test
+	public void conceptSetHelper_shouldReturnEmptyListWhenGivenDescriptorIsNullOrEmpty() {
+		assertTrue(OpenmrsUtil.conceptSetHelper("").isEmpty());
+		assertTrue(OpenmrsUtil.conceptSetHelper(null).isEmpty());
+	}
+	
+		/**
+         * @see OpenmrsUtil#nullSafeEqualsIgnoreCase(String,String)
+         */
 	@Test
 	public void nullSafeEqualsIgnoreCase_shouldBeCaseInsensitive() {
 		Assert.assertTrue(OpenmrsUtil.nullSafeEqualsIgnoreCase("equal", "Equal"));
@@ -680,7 +849,11 @@ public class OpenmrsUtilTest extends BaseContextSensitiveTest {
 	public void nullSafeEqualsIgnoreCase_shouldReturnFalseIfOnlyOneOfTheStringsIsNull() {
 		Assert.assertFalse(OpenmrsUtil.nullSafeEqualsIgnoreCase(null, ""));
 	}
-	
+
+	/**
+	 * 
+	 * @see OpenmrsUtil#storeProperties(Properties, File, String) 
+	 */
 	@Test
 	public void storeProperties_shouldEscapeSlashes() throws IOException {
 		Charset utf8 = StandardCharsets.UTF_8;


Andu033:
Used TDD to rewrite conceptSetHelper, so i wrote the tests verified that pass and then i rewrote the function , so everything works good.

samuelmale:
Can you add the link to ticket?

Andu033:
@samuelmale  of course.
Here: https://issues.openmrs.org/browse/TRUNK-5087?filter=-1

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -19,34 +19,21 @@
...
 import org.apache.commons.io.IOUtils;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
-import org.openmrs.GlobalProperty;
-import org.openmrs.PatientIdentifier;
-import org.openmrs.PatientIdentifierType;
-import org.openmrs.User;
+import org.openmrs.*;
```
samuelmale: Why use wildcard imports https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-Donotusewildcardimports

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -715,37 +715,8 @@ public static int compareWithNullAsLatest(Date d1, Date d2) {
...
-			if (c != null) {
-				if (isSet) {
-					List<Concept> inSet = cs.getConceptsByConceptSet(c);
-					ret.addAll(inSet);
-				} else {
-					ret.add(c);
-				}
-			}
-		}
+		Set<Concept> ret = new HashSet<>(conceptListHelper(descriptor));
```
samuelmale: Why did you remove the above code for this method and instead delegated `conceptListHelper(arg..)`; do you mean these two methods treat these collections in the same way?

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -812,16 +783,22 @@ public static int compareWithNullAsLatest(Date d1, Date d2) {
...
 		}
 		ConceptService cs = Context.getConceptService();
 		
 		for (StringTokenizer st = new StringTokenizer(descriptor, "|"); st.hasMoreTokens();) {
 			String s = st.nextToken().trim();
+			if(duplicates.contains(s)) {
+				continue;
+			}
+			else {
+			duplicates.add(s);
```
samuelmale: * Can you correct this indention?
* By the way, why do we need need `duplicates`, why do we even need to cache them?

samuelmale:
You also need to squash your comments into one.

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
Andu033: You are right. My IDE did that . i will fix

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
Andu033: i cached to solve that todo . The function didnt handle duplicates in descriptor .

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -666,9 +674,185 @@ public void shortenedStackTrace_shouldReturnNullIfStackTraceIsNull() {
...
 	/**
-	 * @see OpenmrsUtil#nullSafeEqualsIgnoreCase(String,String)
+	 * @see OpenmrsUtil#folderContains(File, String)
+	 */
+	@Test
+	public void folderContains_shouldReturnTrueWhenFolderContainsFileWithGivenName() {
+		String path = System.getProperty("user.dir");
+		File folder = new File(path+"/folder");
+		folder.mkdir();
+		File file = new File(path+"/folder/file");
```
samuelmale: Will this file path work on all OSs! Will it work properly on windows? Can you use Java's abstraction and get system path symbols ie "/" or "\"

Andu033: You are right . I will change with System file separator now.

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -666,9 +676,189 @@ public void shortenedStackTrace_shouldReturnNullIfStackTraceIsNull() {
...
+	
+	/**
+	 * @see OpenmrsUtil#folderContains(File, String)
+	 */
+	@Test
+	public void folderContains_shouldReturnFalseWhenGivenStringIsNull() {
+		String path = System.getProperty("user.dir");
+		String separator = FileSystems.getDefault().getSeparator();
+		File folder = new File(path + separator + "folder");
+		folder.mkdir();
```
samuelmale: Don't you think your doing alot of code duplication of the above four lines of code in many methods?

Andu033: Want me to use a private method?

samuelmale: Clean code should be reusable instead of 'copy & paste'. https://www.quora.com/How-to-reuse-a-code-in-Java

You may want to separate concerns by creating another method

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -715,37 +715,8 @@ public static int compareWithNullAsLatest(Date d1, Date d2) {
...
-			}
-			if (c != null) {
-				if (isSet) {
-					List<Concept> inSet = cs.getConceptsByConceptSet(c);
-					ret.addAll(inSet);
-				} else {
-					ret.add(c);
-				}
-			}
-		}
```
samuelmale: I think the above change is beyond the scope for this ticket.

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -812,16 +783,22 @@ public static int compareWithNullAsLatest(Date d1, Date d2) {
 		return ret;
 	}
 	
-	// TODO: properly handle duplicates
 	public static List<Concept> conceptListHelper(String descriptor) {
 		List<Concept> ret = new ArrayList<>();
+		Set<String> duplicates = new HashSet<>();
```
samuelmale: I believe implementing the TODO is also beyond the scope of this issue

Andu033: Ok. I will revert

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
Andu033: Ok. I will revert

Andu033:
@samuelmale  i did the changes. What is left to do ?

Andu033:
@samuelmale i updated the method name and the exception from catch block. The links was very useful . Is there anything left to adjust?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -666,9 +683,169 @@ public void shortenedStackTrace_shouldReturnNullIfStackTraceIsNull() {
...
+		File folder = createFolderInUserDirectory("folder");
+		String separator = FileSystems.getDefault().getSeparator();
+		File file = new File(folder.getPath()  + separator + "file");
+		try {
+			file.createNewFile();
+			assertTrue(OpenmrsUtil.folderContains(folder, "file"));
+			file.delete();
+
+		} 
+		catch (IOException e) {}
```
samuelmale: What happens if the IOException is caught here? 
Does that mean we shall have no test assertions?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -666,9 +683,169 @@ public void shortenedStackTrace_shouldReturnNullIfStackTraceIsNull() {
...
+			PrintWriter writer = new PrintWriter(file);
+			String output = "The first line" + "\n" + "The second line";
+			writer.println("The first line");
+			writer.print("The second line");
+			writer.close();
+			assertTrue(output.equals(OpenmrsUtil.getFileAsString(file)));
+			file.delete();
+			
+		}
+		catch (IOException e) {}
```
samuelmale: What happens if the IOException is caught here?
Does that mean we shall have no test assertions?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -666,9 +683,169 @@ public void shortenedStackTrace_shouldReturnNullIfStackTraceIsNull() {
 	}
 	
 	/**
-	 * @see OpenmrsUtil#nullSafeEqualsIgnoreCase(String,String)
+	 * @see OpenmrsUtil#folderContains(File, String)
+	 */
+	@Test
+	public void folderContains_shouldReturnTrueWhenFolderContainsFileWithGivenName() {
+		File folder = createFolderInUserDirectory("folder");
+		String separator = FileSystems.getDefault().getSeparator();
```
samuelmale: There is alot of code duplication of the above line. Can't we have it as a class constant?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -62,6 +71,14 @@
 	private static GlobalProperty luhnGP = new GlobalProperty(
 	        OpenmrsConstants.GLOBAL_PROPERTY_DEFAULT_PATIENT_IDENTIFIER_VALIDATOR,
 	        OpenmrsConstants.LUHN_IDENTIFIER_VALIDATOR);
+
+	private File createFolderInUserDirectory(String name) {
```
samuelmale: Well according to what I have realized we are testing files in folders not folders in directories. Lets make this method to create a file given it's relative path name eg. 

```java
String relativePath = "{folderName}" + seperator + "{fileName}"; // The relative path could be the folder name + the file we wanna test
```

That way we can make code more reusable

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -666,9 +683,169 @@ public void shortenedStackTrace_shouldReturnNullIfStackTraceIsNull() {
...
+		assertFalse(OpenmrsUtil.folderContains(null,"file"));
+	}
+	
+	/**
+	 * @see OpenmrsUtil#conceptListHelper(String)
+	 */
+	@Test
+	public void conceptListHelper_shouldReturnEmptyListWhenGivenDescriptorIsNullOrEmpty() {
+		List<Concept> list = OpenmrsUtil.conceptListHelper("");
+		assertTrue(list.size() == 0);
```
samuelmale: Doesn't `List` as a Collection have a defined method to check whether it's empty or not?

samuelmale: Do we even need the variable! Isn't this just enough?
`assertTrue(OpenmrsUtil.conceptListHelper("").isEmpty());`

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -666,9 +683,169 @@ public void shortenedStackTrace_shouldReturnNullIfStackTraceIsNull() {
...
+	}
+
+	/**
+	 * @see OpenmrsUtil#conceptListHelper(String)
+	 */
+	@Test
+	public void conceptListHelper_shouldReturnListOfConceptsWithGivenNameInDescriptor() {
+		List<Concept> helperList = OpenmrsUtil.conceptListHelper("name:YES|name:NO|name:DIED|set:25|set:26");
+		List<Concept> serviceList = new ArrayList<>();
+		List<Concept> aux;
```
samuelmale: What does `aux` mean? Can we use more descriptive variable names to self document your code?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -666,9 +683,169 @@ public void shortenedStackTrace_shouldReturnNullIfStackTraceIsNull() {
...
+		Concept con;
```
samuelmale: Same case to this one. More descriptive names needed

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
Andu033: I used it to retrieve the list of concepts by concept set . Is just an auxiliar to make code look better . I can get rid of it by adding the line directly as argument . Same with con ( which stands for concept).

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -666,9 +683,169 @@ public void shortenedStackTrace_shouldReturnNullIfStackTraceIsNull() {
...
+		//Adding the concepts in the list to compare with the first one
+		ConceptService cs = Context.getConceptService();
```
samuelmale: I see alot of code duplication of the above line. Can we have it as a class variable?

Andu033: ok

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
Andu033: Ok . I will do it like that .

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
Andu033: ok

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
Andu033: ok

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
Andu033: I can say something like catch (IOException e){Assert.fail();}

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
samuelmale: what I trying to say have more descriptive variable names. For `con` I understand it's `concept` but whats up with `concept`! But  if one names it like `drugConcept` it makes more sense!

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
samuelmale: Is this of help https://stackoverflow.com/questions/16596418/how-to-handle-exceptions-in-junit/16597164#16597164?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
Andu033: If i do that i get exception : no service found APIexception

Andu033:
@samuelmale  did the changes. Unless the one with Concept service because it thrown exception . But i think that it s ok . I use this like of code 2 times so it wont save much space in test class. This was the final ajustment ? Is the ready to merge?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
samuelmale: Make it a class level variable and call `Context.getConceptService();` from a method level. eg.

```java
@Before
public void setup() {
   // Initialization goes here
}
```

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -63,6 +72,14 @@
 	        OpenmrsConstants.GLOBAL_PROPERTY_DEFAULT_PATIENT_IDENTIFIER_VALIDATOR,
 	        OpenmrsConstants.LUHN_IDENTIFIER_VALIDATOR);
 	
+	private final String separator = FileSystems.getDefault().getSeparator();
+	
+	private File createFileAtRelativePath(String relative_path) {
```
samuelmale: For `relative_path`  and `current_path ` make them camel case sensitive.

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -63,6 +72,14 @@
...
+		String current_path = System.getProperty("user.dir");
```
samuelmale: By the way can you use a better naming for `current_path` ?

Andu033:
OK @samuelmale  . I pushed the changes . Anything else?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -63,6 +72,19 @@
...
 	        OpenmrsConstants.LUHN_IDENTIFIER_VALIDATOR);
 	
+	private ConceptService cs;
+	private final String separator = FileSystems.getDefault().getSeparator();
+	
+	@Before
+	public void setup() {
+		cs=Context.getConceptService();
+	}
+	private File createFileAtRelativePath(String relativePath) {
```
samuelmale: What do you mean by `createFileAtRelativePath` ? Doesn't `createFileInUserDirectory` sound better?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -63,6 +72,19 @@
...
+		String userPath = System.getProperty("user.dir");
```
samuelmale: Doesn't `userDirectoryPath` give more meaning?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -63,6 +72,19 @@
...
+		File file = new File(userPath + separator + relativePath);
+		return file;
```
samuelmale: How sure are you that the file is really created? from where do you invoke `File#createNewFile()` ?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -666,9 +688,156 @@ public void shortenedStackTrace_shouldReturnNullIfStackTraceIsNull() {
...
+		folder.mkdir();
+		File file = createFileAtRelativePath("folder" + separator + "file");
+		try {
+			file.createNewFile();
+			assertTrue(OpenmrsUtil.folderContains(folder, "file"));
+			file.delete();
+
+		} 
+		catch (IOException e) {Assert.fail("Unable to create file");}
+		folder.delete();
```
samuelmale: There is no test case(Behaviour assertions) once  IOException is thrown!

Andu033: If the exception is thrown the test will fail. Is it incorrect?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
Andu033: I create the file in the method of test. This way i can make both directories or file after called this method

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
Andu033: I think that testing this Exception is out of scope . If we are able to create file the test will do it s job . If not test fail.

Andu033:
@samuelmale ping

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -63,6 +72,19 @@
...
 	
+	private ConceptService cs;
+	private final String separator = FileSystems.getDefault().getSeparator();
+	
+	@Before
+	public void setup() {
+		cs=Context.getConceptService();
+	}
+	private File createFileInUserDirectory(String relativePath) {
+		String userDirectoryPath = System.getProperty("user.dir");
```
samuelmale: For running testcases, are you able to use this method to make a directory in the ${user} directory `OpenmrsUtil.getDirectoryInApplicationDataDirectory()` ?

samuelmale: It works in a way that when running tests, files are created in temp dummy files and when running the webapplication , the files are created in the expected directory.

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
samuelmale: Can you implement that?

Andu033: This is already implemented . Thats why i added this assert at catch. If the test go to the catch it will fail.

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
Andu033: When im doing that createnewfile() throw expection

samuelmale:
You have not addressed all the comments

Andu033:
@samuelmale can't create new file in the directory returned by this method. It throw IOException

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
Andu033: I can't use this method. It throw exception when I want to create file.

Andu033:
@samuelmale  can we finish this review?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
samuelmale: I believe that method was created for that! You should tweak around code and make it work!

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -666,9 +688,156 @@ public void shortenedStackTrace_shouldReturnNullIfStackTraceIsNull() {
...
+		catch (IOException e) {Assert.fail("Unable to create file");}
+		folder.delete();
+	}
+	
+	/**
+	 * @see OpenmrsUtil#folderContains(File, String)
 	 */
 	@Test
+	public void folderContains_shouldReturnFalseWhenFolderDoesNotContainFileWithGivenName() {
+		File folder = createFileInUserDirectory("folder");
```
samuelmale: Was this(createFileInUserDirectory) method introduced for creating a file/folder or both?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -666,9 +688,156 @@ public void shortenedStackTrace_shouldReturnNullIfStackTraceIsNull() {
...
+		assertFalse(OpenmrsUtil.folderContains(folder,"file"));
+		folder.delete();
+	}
+	
+	/**
+	 * @see OpenmrsUtil#folderContains(File, String)
+	 */
+	@Test
+	public void folderContains_shouldReturnFalseWhenGivenStringIsNull() {
+		File folder = createFileInUserDirectory("folder");
```
samuelmale: Was this(createFileInUserDirectory) method introduced for creating a file/folder or both?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
Andu033: This method create file. If I want to create directors I just call mkdir() on this file. I can't modify this to do both because this would add to much complexity (one more argument, at least one if etc). So it's better to add a line of code when I want folder (I thinks I used it two times.

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -666,9 +688,156 @@ public void shortenedStackTrace_shouldReturnNullIfStackTraceIsNull() {
...
+	 */
+	@Test
+	public void conceptSetHelper_shouldReturnEmptyListWhenGivenDescriptorIsNullOrEmpty() {
+		assertTrue(OpenmrsUtil.conceptSetHelper("").isEmpty());
+		assertTrue(OpenmrsUtil.conceptSetHelper(null).isEmpty());
+	}
+	
+		/**
+         * @see OpenmrsUtil#nullSafeEqualsIgnoreCase(String,String)
+         */
```
haripriya999: I think it would look neater if  @see at line 837 started below the ending curly brace of conceptSetHelper_shouldReturnEmptyListWhenGivenDescriptorIsNullOrEmpty() :)

PR #2826, opened 2019-02-07
changes on trunk-4742

gitcliff:
Add support for setting test database credentials externally on ticket TRUNK--4742

LINK: https://issues.openmrs.org/browse/TRUNK-4742?src=confmacro
I closed the two old pull request which was on TRUNK-4742 because when i was squashing commits they got it another ticket and i have created a new one and travis failure

### DIFF
--- a/api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
+++ b/api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
@@ -291,6 +291,7 @@ public abstract class BaseContextSensitiveTest extends AbstractJUnit4SpringConte
 		return "openmrs";
 	}
 	
+	
 	/**
 	 * Mimics org.openmrs.web.Listener.getRuntimeProperties() Overrides the database connection
 	 * properties if the user wants an in-memory database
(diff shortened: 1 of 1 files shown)


api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
```diff
@@ -532,17 +543,8 @@ public void run() {
...
-	/**
-	 * Override this method to turn on/off the in-memory database. The default is to use the
-	 * in-memory database. When this method returns false, the database defined by the runtime
-	 * properties is used instead
-	 * 
-	 * @return true/false whether or not to use an in memory database
-	 */
-	public Boolean useInMemoryDatabase() {
-		return !"false".equals(System.getProperty("useInMemoryDatabase"));
-	}
```
samuelmale: Why change position of this method?

gitcliff: had changed it so that it can be overridden to turn off the in memory DB hence sitting on the extern al DB credentials in the runtime properties

gitcliff: is it what could have caused the travis failure? coz i got a build success while compiling

samuelmale: > had changed it so that it can be overridden to turn off the in memory DB hence sitting on the extern al DB credentials in the runtime properties

Can't the method be overridden from the original position. Asn what's the logic behind this!

PR #2830, opened 2019-02-11
TRUNK-3343: insert the new application level privileges for each each…

mozzy11:
TICKET :https://issues.openmrs.org/browse/TRUNK-3343

### DIFF
--- a/api/src/main/resources/liquibase-update-to-latest.xml
+++ b/api/src/main/resources/liquibase-update-to-latest.xml
@@ -21,4 +21,79 @@
     <include file="liquibase-update-to-2.2.xml"/>
     <include file="liquibase-update-to-2.3.xml"/>
 
+
+<changeSet id="TRUNK-3343-assigning-new-app-privileges" author="mozzy">
+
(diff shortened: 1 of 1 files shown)


api/src/main/resources/liquibase-update-to-latest.xml
```diff
@@ -21,4 +21,73 @@
     <include file="liquibase-update-to-2.2.xml"/>
     <include file="liquibase-update-to-2.3.xml"/>
 
+
+<changeSet id="TRUNK-3343-assigning-new-app-privileges" author="mozzy">
+
```
dkayiwa: Shouldn't you add some preconditions?

mozzy11: oh ya but in this case i failed to figure out an appropriate pre condition. May be u can give me a hint

mozzy11: i added that pre-condition

ibacher:
This probably needs to be refactored to take into account our new Liquibase structure.

PR #2833, opened 2019-02-14
TRUNK-381: Support for alternate authentication schemes

ahammi:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-381

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Authenticated.java
+++ b/api/src/main/java/org/openmrs/api/context/Authenticated.java
@@ -1,0 +1,29 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.context;
+
+import org.openmrs.User;
+
+/**
+ * The result of an authentication process, see {@link AuthenticationScheme#authenticate(Credentials)}.
+ * At least the OpenMRS {@link User) is returned, and more if needed through further extensions/implementations.
+ */
+public interface Authenticated {
+	
+	/**
+	 * @return A string describing or identifying the authentication scheme that was used.
+	 */
+	public String getAuthenticationScheme();
+	
+	/**
+	 * @return The authenticated OpenMRS {@link User}.
+	 */
+	public User getUser();
+}
--- a/api/src/main/java/org/openmrs/api/context/Authenticated.java
+++ b/api/src/main/java/org/openmrs/api/context/Authenticated.java
@@ -14,6 +14,8 @@ import org.openmrs.User;
 /**
  * The result of an authentication process, see {@link AuthenticationScheme#authenticate(Credentials)}.
  * At least the OpenMRS {@link User) is returned, and more if needed through further extensions/implementations.
+ * 
+ * @since 2.3.0
  */
 public interface Authenticated {
 	
--- a/api/src/main/java/org/openmrs/api/context/AuthenticationScheme.java
+++ b/api/src/main/java/org/openmrs/api/context/AuthenticationScheme.java
@@ -11,6 +11,8 @@ package org.openmrs.api.context;
 
 /**
  * Every authentication scheme should implement its authentication method out of this API.
+ * 
+ * @since 2.3.0
  */
 public interface AuthenticationScheme {
 	
--- a/api/src/main/java/org/openmrs/api/context/AuthenticationScheme.java
+++ b/api/src/main/java/org/openmrs/api/context/AuthenticationScheme.java
@@ -1,0 +1,19 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.context;
+
+/**
+ * Every authentication scheme should implement its authentication method out of this API.
+ */
+public interface AuthenticationScheme {
+	
+	public Authenticated authenticate(Credentials credentials) throws ContextAuthenticationException;
+	
+}
--- a/api/src/main/java/org/openmrs/api/context/BasicAuthenticated.java
+++ b/api/src/main/java/org/openmrs/api/context/BasicAuthenticated.java
@@ -13,6 +13,8 @@ import org.openmrs.User;
 
 /**
  * A basic implementation of {@link Authenticated} that contains the necessary elements for an OpenMRS authentication.
+ * 
+ * @since 2.3.0
  */
 public class BasicAuthenticated implements Authenticated {
 	
--- a/api/src/main/java/org/openmrs/api/context/BasicAuthenticated.java
+++ b/api/src/main/java/org/openmrs/api/context/BasicAuthenticated.java
@@ -1,0 +1,43 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.context;
+
+import org.openmrs.User;
+
+/**
+ * A basic implementation of {@link Authenticated} that contains the necessary elements for an OpenMRS authentication.
+ */
+public class BasicAuthenticated implements Authenticated {
+	
+	protected User user;
+	
+	protected String authenticationScheme;
+	
+	public BasicAuthenticated(User user, String authenticationScheme) {
+		this.user = user;
+		this.authenticationScheme = authenticationScheme;
+	}
+	
+	public String getAuthenticationScheme() {
+		return this.authenticationScheme;
+	}
+	
+	public void setAuthenticationScheme(String authenticationScheme) {
+		this.authenticationScheme = authenticationScheme;
+	}
+	
+	public User getUser() {
+		return this.user;
+	}
+	
+	public void setUser(User user) {
+		this.user = user;
+	}
+}
--- a/api/src/main/java/org/openmrs/api/context/BasicAuthenticated.java
+++ b/api/src/main/java/org/openmrs/api/context/BasicAuthenticated.java
@@ -18,9 +18,9 @@ import org.openmrs.User;
  */
 public class BasicAuthenticated implements Authenticated {
 	
-	protected User user;
+	private User user;
 	
-	protected String authenticationScheme;
+	private String authenticationScheme;
 	
 	public BasicAuthenticated(User user, String authenticationScheme) {
 		this.user = user;
--- a/api/src/main/java/org/openmrs/api/context/Context.java
+++ b/api/src/main/java/org/openmrs/api/context/Context.java
@@ -267,8 +267,7 @@ public class Context {
 	}
 
 	/**
-	 * @deprecated
-	 * @see #authenticate(AuthenticationScheme, Credentials)
+	 * @deprecated as of 2.3.0, replaced by #authenticate(AuthenticationScheme, Credentials)
 	 * 
 	 * Used to authenticate user within the context
 	 *
--- a/api/src/main/java/org/openmrs/api/context/Context.java
+++ b/api/src/main/java/org/openmrs/api/context/Context.java
@@ -281,6 +281,7 @@ public class Context {
 	 * @should not authenticate with null password and proper username
 	 * @should not authenticate with null password and proper system id
 	 */
+	@Deprecated
 	public static void authenticate(String username, String password) throws ContextAuthenticationException {
 		authenticate( new UserPassAuthenticationScheme(), new UserPassCredentials(username, password) );
 	}
--- a/api/src/main/java/org/openmrs/api/context/Context.java
+++ b/api/src/main/java/org/openmrs/api/context/Context.java
@@ -267,7 +267,7 @@ public class Context {
 	}
 
 	/**
-	 * @deprecated since 2.3.0
+	 * @deprecated
 	 * @see #authenticate(AuthenticationScheme, Credentials)
 	 * 
 	 * Used to authenticate user within the context
@@ -283,7 +283,7 @@ public class Context {
 	 */
 	@Deprecated
 	public static void authenticate(String username, String password) throws ContextAuthenticationException {
-		authenticate( new UserPassAuthenticationScheme(), new UserPassCredentials(username, password) );
+		authenticate( new UsernamePasswordAuthenticationScheme(), new UsernamePasswordCredentials(username, password) );
 	}
 
 	/**
--- a/api/src/main/java/org/openmrs/api/context/Context.java
+++ b/api/src/main/java/org/openmrs/api/context/Context.java
@@ -36,7 +36,9 @@ import org.openmrs.api.APIException;
 import org.openmrs.api.AdministrationService;
 import org.openmrs.api.CohortService;
 import org.openmrs.api.ConceptService;
+import org.openmrs.api.ConditionService;
 import org.openmrs.api.DatatypeService;
+import org.openmrs.api.DiagnosisService;
 import org.openmrs.api.EncounterService;
 import org.openmrs.api.FormService;
 import org.openmrs.api.LocationService;
@@ -51,8 +53,6 @@ import org.openmrs.api.ProviderService;
 import org.openmrs.api.SerializationService;
 import org.openmrs.api.UserService;
 import org.openmrs.api.VisitService;
-import org.openmrs.api.ConditionService;
-import org.openmrs.api.DiagnosisService;
 import org.openmrs.api.db.ContextDAO;
 import org.openmrs.hl7.HL7Service;
 import org.openmrs.logic.LogicService;
@@ -267,6 +267,9 @@ public class Context {
 	}
 
 	/**
+	 * @deprecated since 2.3.0
+	 * @see #authenticate(AuthenticationScheme, Credentials)
+	 * 
 	 * Used to authenticate user within the context
 	 *
 	 * @param username user's identifier token for login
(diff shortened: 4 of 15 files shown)


api/src/test/java/org/openmrs/api/context/ContextTest.java
```diff
@@ -82,7 +82,40 @@ public void authenticate_shouldNotAuthenticateWithNullUsername() {
...
+	@Test(expected = ContextAuthenticationException.class)
+	public void authenticate_shouldThrowWhenUnfitCredentials() {
+		Context.authenticate(new UserPassAuthenticationScheme(), new TestUuidCredentials("c1d8f5c2-e131-11de-babe-001e378eb67e"));
+	}
+	
+	/**
+	 * @see Context#authenticate(AuthenticationScheme, Credentials)
+	 */
+	@Test
+	public void authenticate_shouldAuthenticateUserThoughAlternateScheme() {
```
mks-d: @ahammi super minor change here, typo left in the method name: `authenticate_shouldAuthenticateUserThoughAlternateScheme` → `authenticate_shouldAuthenticateUserWithAlternateScheme`.

api/src/main/java/org/openmrs/api/context/UserPassAuthenticationScheme.java
```diff
@@ -0,0 +1,43 @@
...
+		
+		UserPassCredentials userPassCreds = null;
+		try {
+			userPassCreds = (UserPassCredentials) credentials;
+		}
+		catch (ClassCastException e) {
+			throw new ContextAuthenticationException("The provided credentials could not be used to authenticated with the specified authentication scheme.", e);
+		}
+		
+
```
mks-d: @ahammi there is an extra line here, let's remove it.

api/src/main/java/org/openmrs/api/context/Authenticated.java
```diff
@@ -0,0 +1,29 @@
...
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.context;
+
+import org.openmrs.User;
+
+/**
+ * The result of an authentication process, see {@link AuthenticationScheme#authenticate(Credentials)}.
+ * At least the OpenMRS {@link User) is returned, and more if needed through further extensions/implementations.
+ */
```
dkayiwa: Can you put a since annotation on this class?

api/src/main/java/org/openmrs/api/context/AuthenticationScheme.java
```diff
@@ -0,0 +1,19 @@
...
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.context;
+
+/**
+ * Every authentication scheme should implement its authentication method out of this API.
+ */
```
dkayiwa: Can you put a since annotation on this interface?

api/src/main/java/org/openmrs/api/context/BasicAuthenticated.java
```diff
@@ -0,0 +1,43 @@
...
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.context;
+
+import org.openmrs.User;
+
+/**
+ * A basic implementation of {@link Authenticated} that contains the necessary elements for an OpenMRS authentication.
+ */
```
dkayiwa: Can you put a since annotation on this class?

api/src/main/java/org/openmrs/api/context/ContextAuthenticationException.java
```diff
@@ -23,6 +23,10 @@ public ContextAuthenticationException() {
 		super();
 	}
 	
+	public ContextAuthenticationException(String message, Throwable cause) {
```
dkayiwa: Can you put a since annotation on this?

api/src/main/java/org/openmrs/api/context/Credentials.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.context;
+
+/**
+ * Authentication schemes define and require their own credentials.
+ * Any client authenticating against a given scheme must supply appropriate credentials.
+ */
```
dkayiwa: Can you put a since annotation on this?

api/src/main/java/org/openmrs/api/context/DaoAuthenticationScheme.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.openmrs.api.context;
+
+import org.openmrs.api.db.ContextDAO;
+
+/**
+ * Base class for authentication schemes that intend to leverage OpenMRS' {@link ContextDAO}.
+ *
+ * @see {@link Context#authenticate(AuthenticationScheme, Credentials)}
+ */
```
dkayiwa: Can you put a since annotation on this?

api/src/main/java/org/openmrs/api/context/UserPassAuthenticationScheme.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * Default OpenMRS authentication scheme to login with OpenMRS' usernames and passwords.
+ * 
+ * @see {@link AuthenticationScheme}
+ */
+public class UserPassAuthenticationScheme extends DaoAuthenticationScheme {
```
dkayiwa: Can you put a since annotation on this?

api/src/main/java/org/openmrs/api/context/UserPassCredentials.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.context;
+
+/**
+ * User-pass credentials represent the usual OpenMRS authentication credentials made of a pair username + password.
+ */
+public class UserPassCredentials implements Credentials {
```
dkayiwa: Can you put a since annotation on this?

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -267,6 +267,9 @@ public static void setContext(ServiceContext ctx) {
 	}
 
 	/**
+	 * @deprecated since 2.3.0
```
dkayiwa: Can you take a look at this? https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-Deprecation

api/src/main/java/org/openmrs/api/context/Credentials.java
```diff
@@ -0,0 +1,29 @@
...
+	 * @return The authentication scheme that should be used with those credentials.
+	 */
+	public String getAuthenticationScheme();
+	
+	/**
+	 * A string that identifies the owner of the credentials by a name.
+	 * Typically in the case of a user the client would simply be identified with its username. 
+	 */
+	public String getClientName();
+
```
dkayiwa: How do you plan to deal with this? "Credentials will contains scheme-specific properties"

mks-d: @dkayiwa this quote comes from @bmamlin's suggested JavaDoc on the ticket's description. I think the point is that devs/implementers can keep extending `Credentials` implementations (such as extending `UserPassCredentials` for example) or can provide their own implementations of `Credentials`. Either way they will add members and methods that are useful to their custom auth. schemes. We can't anticipate as of yet what will be done, we can just leave the design open enough.

api/src/main/java/org/openmrs/api/context/Context.java
dkayiwa: Combining deprecation with since can be confusing to some users because of the @since annotation. So i would recommend doing exactly as stated at https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-Deprecation

api/src/main/java/org/openmrs/api/context/BasicAuthenticated.java
```diff
@@ -0,0 +1,45 @@
...
+import org.openmrs.User;
+
+/**
+ * A basic implementation of {@link Authenticated} that contains the necessary elements for an OpenMRS authentication.
+ * 
+ * @since 2.3.0
+ */
+public class BasicAuthenticated implements Authenticated {
+	
+	protected User user;
```
dkayiwa: Any reason why these are not private?

api/src/main/java/org/openmrs/api/context/UserPassAuthenticationScheme.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * Default OpenMRS authentication scheme to login with OpenMRS' usernames and passwords.
+ * 
+ * @see {@link AuthenticationScheme}
+ * 
+ * @since 2.3.0
+ */
+public class UserPassAuthenticationScheme extends DaoAuthenticationScheme {
```
dkayiwa: Is there any reason why this was not named UsernamePasswordAuthenticationScheme or NamePasswordAuthenticationScheme? Not a big deal though, if you do not feel like it. Just wanted to mention it. 😊

api/src/main/java/org/openmrs/api/context/UserPassAuthenticationScheme.java
```diff
@@ -0,0 +1,44 @@
...
+
+	protected static Log log = LogFactory.getLog(UserPassAuthenticationScheme.class);
```
dkayiwa: Any reason why the above is not private static final?

api/src/main/java/org/openmrs/api/context/UserPassAuthenticationScheme.java
```diff
@@ -0,0 +1,44 @@
...
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.context;
+
+import org.apache.commons.logging.Log;
```
dkayiwa: Any reason why you are not using org.slf4j.Logger just like the rest of the platform?

api/src/main/java/org/openmrs/api/context/UserPassAuthenticationScheme.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.commons.logging.LogFactory;
```
dkayiwa: Any reason why you are not using org.slf4j.LoggerFactory? Or did you take a look at this? https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-loggingLogging

api/src/main/java/org/openmrs/api/context/UserPassCredentials.java
dkayiwa: Any reason why this was not named UsernamePasswordCredentials or NamePasswordCredentials?

api/src/main/java/org/openmrs/api/context/UserPassCredentials.java
```diff
@@ -0,0 +1,47 @@
...
+package org.openmrs.api.context;
+
+/**
+ * User-pass credentials represent the usual OpenMRS authentication credentials made of a pair username + password.
+ * 
+ * @since 2.3.0
+ */
+public class UserPassCredentials implements Credentials {
+
+	protected String username;
```
dkayiwa: Any reason why these are not private?

api/src/test/java/org/openmrs/api/context/ContextTest.java
```diff
@@ -82,7 +82,40 @@ public void authenticate_shouldNotAuthenticateWithNullUsername() {
...
 	public void authenticate_shouldNotAuthenticateWithNullUsernameAndPassword() {
-		Context.authenticate(null, null);
+		Context.authenticate((String) null, null);
+	}
+	
+	/**
+	 * @see Context#authenticate(String,String)
+	 */
+	@Test(expected = ContextAuthenticationException.class)
+	public void authenticate_shouldNotAuthenticateWithNullCredentials() {
```
dkayiwa: Is this testing against only the credentials being null or both credentials and authentication scheme?

api/src/test/java/org/openmrs/api/context/TestUuidAuthenticationScheme.java
```diff
@@ -0,0 +1,28 @@
...
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.context;
+
+import org.openmrs.User;
+
+/*
+ * This test auth scheme logs a user simply based on asserting that its UUID is valid and points to an exisiting user.
+ */
+public class TestUuidAuthenticationScheme extends DaoAuthenticationScheme {
```
dkayiwa: Wouldn't it better to instead of use a test with a username and password? Some sort of admin/test combination. For this to me looks closer to the real basic authentication which we have that is based on a username and password combination.

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -87,30 +86,42 @@ public UserContext() {
...
 	 * @throws ContextAuthenticationException
+	 * 
+	 * @since 2.3.0
 	 */
-	public User authenticate(String username, String password, ContextDAO contextDAO) throws ContextAuthenticationException {
-		log.debug("Authenticating with username: {}", username);
+	public Authenticated authenticate(AuthenticationScheme authenticationScheme, Credentials credentials)
+			throws ContextAuthenticationException {
+
+		if (log.isDebugEnabled()) {
```
dkayiwa: Do you want to also log the client name?

dkayiwa: Do we need the log.isDebugEnabled() check?

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -87,30 +86,42 @@ public UserContext() {
...
+			loggingInUser.setUsername(credentials.getClientName());
 			notifyUserSessionListener(loggingInUser, Event.LOGIN, Status.FAIL);
 			throw e;
 		}
+		
 		setUserLocation();
-		log.debug("Authenticated as: {}", this.user);
 		
-		return this.user;
+		if (log.isDebugEnabled()) {
```
dkayiwa: Do we need the log.isDebugEnabled() check?

api/src/main/java/org/openmrs/api/context/UserPassAuthenticationScheme.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public class UserPassAuthenticationScheme extends DaoAuthenticationScheme {
+
+	protected static Log log = LogFactory.getLog(UserPassAuthenticationScheme.class);
+	
+	@Override
+	public Authenticated authenticate(Credentials credentials)
+			throws ContextAuthenticationException {
+
+		if (log.isDebugEnabled()) {
```
dkayiwa: Do we need the log.isDebugEnabled() check?

api/src/test/java/org/openmrs/api/context/TestUuidAuthenticationScheme.java
dkayiwa: Just to say this more explicitly, this test should also stay to ensure that we support more than just our basic username and password db based authentication.

dkayiwa:
Which API will a module use for the wiring of a new authentication scheme? Will that be in another pull request?

api/src/main/java/org/openmrs/api/context/UserContext.java
mks-d: @dkayiwa those checks are all over the code base though, so I will let them in for now. I agree that it's weird, I would have assumed that a logger knows whether it supports debugging or not. Please advise.

mks-d:
>Which API will a module use for the wiring of a new authentication scheme? Will that be in another pull request?

@dkayiwa no wiring per se, you just explicitly invoke `Context.authenticate(AuthenticationScheme, Credentials)` with your own pair scheme/credentials defined in your custom module.

We are working on a OAuth2 client module that authenticates with [Keycloak](https://www.keycloak.org/), we will be able to provide an example shortly with a custom scheme. No Spring magic as of yet.

api/src/test/java/org/openmrs/api/context/TestUuidAuthenticationScheme.java
mks-d: @dkayiwa yes absolutely, this test should have existed already with the method that we suggest to deprecate. Anyway I added it in the commit #3.

api/src/main/java/org/openmrs/api/context/UserContext.java
dkayiwa: Are you sure there is any such in the latest version of master?

dkayiwa:
@mks-d what i mean is when some one wants the existing unmodified openmrs reference application login page to use a custom authentication scheme provided by a module.

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -267,6 +267,9 @@ public static void setContext(ServiceContext ctx) {
 	}
 
 	/**
+	 * @deprecated
```
dkayiwa: Did you take a look at this? https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-Deprecation

api/src/main/java/org/openmrs/api/context/UsernamePasswordAuthenticationScheme.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public class UsernamePasswordAuthenticationScheme extends DaoAuthenticationScheme {
+
+	private static final Logger log = LoggerFactory.getLogger(UsernamePasswordAuthenticationScheme.class);
+	
+	@Override
+	public Authenticated authenticate(Credentials credentials)
+			throws ContextAuthenticationException {
+
+		if (log.isDebugEnabled()) {
```
dkayiwa: Do we need the log.isDebugEnabled() check?

api/src/main/java/org/openmrs/api/context/UserContext.java
mks-d: @dkayiwa yes if you do a search, simply via GitHub like [here](https://github.com/openmrs/openmrs-core/search?q=log.isDebugEnabled%28%29&unscoped_q=log.isDebugEnabled%28%29), you will find dozens. One example on `Concept`: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/Concept.java#L451

Permalink [here](https://github.com/openmrs/openmrs-core/blob/c9996a61e8d920c1cbe5c1c8c0c4b3b3dd534fd4/api/src/main/java/org/openmrs/Concept.java#L451).

dkayiwa: There was a pull request which removed those in a number of classes. I just noticed that it did not fully remove them from the entire codebase. 😊

mks-d:
> @mks-d what i mean is when some one wants the existing unmodified openmrs reference application login page to use a custom authentication scheme provided by a module.

@dkayiwa this PR does not yet cover such use case. It rather provides devs with an API to authenticate with other schemes if they need to.

Btw I would suggest that we follow up with a PR on the Ref App module so that it uses the new API rather than the one that will be deprecated.

dkayiwa:
I was looking at this text in the ticket description: " Ideally, substitution (or addition) of the authentication scheme would be protected (e.g., limited to the configuration file setting or not easily done through the API) so that random code could not easily substitute a bogus authentication scheme to subvert proper authentication."

My understanding of the above was not to necessarily hard code the UsernamePasswordAuthenticationScheme but provide it as the default with ability to be overridden by the same authenticate(String username, String password) method call.

dkayiwa:
Infact authenticate(AuthenticationScheme authenticationScheme, Credentials credentials) seems to do the opposite of the above. That is, makes it so easy for some random code to substitute a bogus authentication scheme to subvert proper authentication. 😊

api/src/main/java/org/openmrs/api/context/Context.java
mks-d: @dkayiwa I did this in my last commit:
```java
/**
 * @deprecated as of 2.3.0, replaced by #authenticate(AuthenticationScheme, Credentials)
 ...
 */
```
Is that what you expect? If not please point me to the exact JavaDoc that you would like to have in place.

mks-d:
> Infact the public authenticate(AuthenticationScheme authenticationScheme, Credentials credentials) seems to do the opposite of the above. That is, makes it so easy for some random code to substitute a bogus authentication scheme to subvert proper authentication. 😊

@dkayiwa I see, so authentication schemes should be Spring-wired, and the Spring config would use the add and remove `protected` methods. So that's how `UserContext` knows which scheme(s) to use.
But when using an alternate scheme, how does one pass to `UserContext` the credentials to use with the scheme, since the only method that could ever be called is `Context.authenticate(String username, String password)`?

Does that imply that the only user identification that could ever be passed downstream is the username?

mks-d:
@dkayiwa more precisely, deep down one will rely on `ContextDAO` to fetch the authenticated user, and it's only possible to fetch users by UUID (see [here](https://github.com/openmrs/openmrs-core/blob/c9996a61e8d920c1cbe5c1c8c0c4b3b3dd534fd4/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java#L214)). What will be needed by alternate schemes is a `getUserByUsername(String)` so this will be added in the next commit.

PR #2838, opened 2019-02-20
TRUNK-381: Authentication scheme is wired as a Spring bean.

mks-d:
This is still WIP to make the new authentication scheme being wired as a Spring bean. This leads to a number of changes:
- `ContextDAO` needs to expose a non-authenticated way to feth a user by username.
- `ContextDAO` needs to expose a non-authenticated way to create a new user. This is for cases where the SSO authority validates that a user _is_ authenticated while it doesn't exist yet in OpenMRS.
- `BaseContextSensitiveTest` should provide a way to not authenticate 'admin' in its base setup as this authentication might fail when an alternate authentication scheme is in place. This is for cases where the SSO doesn't know 'admin' or doesn't know of its OpenMRS password.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-381

  No? -> execute above command


api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
```diff
@@ -996,4 +998,20 @@ public void skipBaseSetup() throws Exception {
...
-}
+	/**
+	 * Instance variable used by the {@link #baseSetupWithStandardDataAndAuthentication()} method to
+	 * know whether the current "@Test" method has asked to be _not_ do the authentication.
+	 * 
+	 * @see SkipBaseSetup
+	 * @see SkipBaseSetupAnnotationExecutionListener
+	 * @see #baseSetupWithStandardDataAndAuthentication()
+	 */
+	private boolean skipAuthenticationInBaseSetup = false;
```
dkayiwa: @mks-d i would not add this in anticipation of future use cases. 😊 
In otherwords, i have not seen any part of your changes in the pull request that required any change in this class.

mks-d: @dkayiwa it is necessary as soon as you want to test an alternate auth scheme, have a look at `AlternateAuthenticationSchemeContextTest`.

This test wires a test scheme that requires its own credentials, that scheme is therefore unable to authenticate 'admin' with the OpenMRS credentials as it doesn't know about them and hence what do to with such credentials.

api/src/test/java/org/openmrs/api/context/TestUsernameAuthenticationScheme.java
```diff
@@ -12,16 +12,16 @@
...
+ * This test auth scheme logs a user simply based on asserting that its username is valid and points to an existing user.
  */
-public class TestUuidAuthenticationScheme extends DaoAuthenticationScheme {
+public class TestUsernameAuthenticationScheme extends DaoAuthenticationScheme {
 
 	@Override
 	public Authenticated authenticate(Credentials credentials) throws ContextAuthenticationException {
 		
-		TestUuidCredentials creds = (TestUuidCredentials) credentials;
+		TestUsernameCredentials creds = (TestUsernameCredentials) credentials;
```
dkayiwa: I do not see a need for this casting. I thought you would just do credentials.getClientName() on line 24 below and as a result, you do not need the skipAuthenticationInBaseSetup changes in BaseContextSensitiveTest.java

api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
dkayiwa: I have put a comment in TestUsernameAuthenticationScheme in regards to this.

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -637,6 +627,38 @@ public static boolean isAuthenticated() {
...
+			log.error("Multiple authentication schemes overrides are being provided, this is currently not supported. Sticking to OpenMRS default authentication scheme.");
+		}
+		catch(NoSuchBeanDefinitionException e) {
+			log.debug("No authentication scheme override was provided. Sticking to OpenMRS default authentication scheme.");
+		}
+		catch(BeansException e){
+			log.error("Fatal error encountered when injecting the authentication scheme override. Sticking to OpenMRS default authentication scheme.");
+		}
+		
+		if (scheme instanceof DaoAuthenticationScheme) {
```
dkayiwa: Why would a get something method have a side effect of setting something?

mks-d: @dkayiwa this method is not really a getter (it doesn't return a member of `Context`). It's more akin of a factory, it's basically a wrapper around a Spring injection.
We could do this if you find it less awkward:

```java
public static AuthenticationScheme getAuthenticationScheme() {

  AuthenticationScheme scheme = new UsernamePasswordAuthenticationScheme(getContextDAO());
  
  try {
    scheme = Context.getServiceContext().getApplicationContext().getBean(AuthenticationScheme.class); // manual autowiring from a module
    if (scheme instanceof DaoAuthenticationScheme) {
      ((DaoAuthenticationScheme) scheme).setContextDao(getContextDAO());
    }
    log.info("An authentication scheme override was provided. Using this one in place of the OpenMRS default authentication scheme.");
  }
  catch(NoUniqueBeanDefinitionException e) {
    log.error("Multiple authentication schemes overrides are being provided, this is currently not supported. Sticking to OpenMRS default authentication scheme.");
  }
  catch(NoSuchBeanDefinitionException e) {
    log.debug("No authentication scheme override was provided. Sticking to OpenMRS default authentication scheme.");
  }
  catch(BeansException e){
    log.error("Fatal error encountered when injecting the authentication scheme override. Sticking to OpenMRS default authentication scheme.");
  }
  
  return scheme;
}
```

dkayiwa: For even a factory method that gets me an instance of an object, i would expect it to create or return an existing instance but not do anything like setting something else. This side effect does not look intuitive to me. Unless some one else disagrees. Spring injection would be a via a set, not get. 😊

mks-d: Thanks @dkayiwa. So I made an attempt at splitting clearly between the auth. scheme's getter and setter. Setting the auth. scheme in the first place is something that must be done as part of `Context`'s initialisation process and this is why it is now used as its `init-method` (see application context XML).

However it is tricky to achieve to run something like that:
```java
@Autowired(required=false)
static private void setAuthenticationScheme(AuthenticationScheme moduleAuthScheme) {

  authenticationScheme = moduleAuthScheme; 
  if (authenticationScheme == null) {
    authenticationScheme = new UsernamePasswordAuthenticationScheme();
  }
  
  if (authenticationScheme instanceof DaoAuthenticationScheme) {
    ((DaoAuthenticationScheme) authenticationScheme).setContextDao(getContextDAO());
  }
}
```
While that would be more Spring-like, it doesn't quite allow to fully handle all possible exceptions as done until now. But as I said it is tricky to make that work on static methods or on private methods.
Note also that I made sure that `setAuthenticationScheme` is private to prevent any kind of override through ever consuming the API.

Finally I tested all this at runtime again and it works fine with and without a module that Spring-wires a custom auth. scheme.

web/src/test/java/org/openmrs/web/context/AltAuthSchemeContextTest.java
```diff
@@ -0,0 +1,54 @@
...
+ * 
+ * @see Context
+ */
+/*
+ * TODO Enable @ContextConfiguration on subclasses of BaseContextSensitiveTest
+ * 
+ * @see https://issues.openmrs.org/browse/TRUNK-381?focusedCommentId=254292&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-254292
+ */
+//@ContextConfiguration(locations = { "classpath*:AltAuthSchemeTestingApplicationContext.xml" })
+public class AltAuthSchemeContextTest extends BaseWebContextSensitiveTest {
```
dkayiwa: Did you intentionally abbreviate Alternate? 😊

api/src/main/java/org/openmrs/api/db/ContextDAO.java
```diff
@@ -62,6 +62,29 @@
 	 */
 	public User getUserByUuid(String uuid) throws ContextAuthenticationException;
 	
+	/**
+	 * Gets a user given the username. Privilege checks are not done here because this is used by the
```
dkayiwa: Instead of introducing these new methods, can't we just use the Daemon functionality? https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/context/Daemon.java#L28-L29

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -177,6 +182,32 @@ public static void setDAO(ContextDAO dao) {
...
+			log.error("Multiple authentication schemes overrides are being provided, this is currently not supported. Sticking to OpenMRS default authentication scheme.");
+		}
+		catch(NoSuchBeanDefinitionException e) {
+			log.debug("No authentication scheme override was provided. Sticking to OpenMRS default authentication scheme.");
+		}
+		catch(BeansException e){
+			log.error("Fatal error encountered when injecting the authentication scheme override. Sticking to OpenMRS default authentication scheme.");
+		}
+
+		if (authenticationScheme instanceof DaoAuthenticationScheme) {
```
dkayiwa: A special case of setting the contextDAO from the setAuthenticationScheme() method when the scheme is a DaoAuthenticationScheme does not look very intuitive to me. 😊 
Can't the contextDAO be dependency injected or set from somewhere else?

web/src/test/java/org/openmrs/web/context/AltAuthSchemeContextTest.java
mks-d: Yes actually, I found that all those test resources were quite longuishly named.

api/src/main/java/org/openmrs/api/db/ContextDAO.java
mks-d: Could we leave this as a further improvement tracked by new ticket? While you're right this was just following the current pattern of the existing `getUserByUuid(String)` that also did not use the Daemon thread (and weirdly reimplemented the same method form `UserDAO` btw...)

api/src/main/java/org/openmrs/api/context/Context.java
mks-d: Should make all schemes subclass `DaoAuthenticationScheme`, thereby perhaps renaming it `BaseAuthenticationScheme`?

Until here I sort of left the question open as to whether some schemes might not need `ContextDAO`, but maybe this will just never happen in OpenMRS. I guess this might one day be bypassed if OpenMRS was authenticating using Spring Security instead of handling all those things "manually". What do you think?

dkayiwa: Why would other schemes need ContextDAO?

api/src/main/java/org/openmrs/api/db/ContextDAO.java
dkayiwa: Since getUserByUuid() is not making any changes to the database, may be yes. But do you still think so even for saveUser()?

api/src/main/java/org/openmrs/api/context/Context.java
mks-d: While the actual authentication process will be delegated to a third-party authority (an OAuth2 provider such as Keycloak typically), OpenMRS still requires one of its own users. Think of the roles for example, it is not possible to just pass a `User` instance here and there saying "it's authenticated", OpenMRS will connect roles to it, and for this it is necessary to have a `User` entity to exist in OpenMRS' database as well.

The question is rather... will there be a point when it will be possible to not rely on `ContextDAO` anymore when delegating authentication elsewhere?

dkayiwa: Does this mean the roles will also need to be inserted into the OpenMRS database?

mks-d: No, we assume that the roles we need already exist. So we might add roles to the new user, but we won't create those roles.

dkayiwa: Sorry, that is exactly what i did mean. In other words, i should have asked like this: "Are you also inserting user to role mappings?"

api/src/test/java/org/openmrs/api/context/DaemonTest.java
```diff
@@ -52,6 +58,36 @@ public void executeScheduledTask_shouldNotThrowErrorIfCalledFromATimerSchedulerT
...
+	public void createUser_shouldCreateUserWithRolesInContextDAO() throws Throwable {
+		// setup
+		User u = new User();
+		u.setPerson(new Person());
+		u.addName(new PersonName("Jane", "X", "Doe"));
+		u.setUsername("jdoe");
+		u.getPerson().setGender("F");
+		
+		// replay
+		User createdUser = Context.getContextDAO().createUser(u, "P@ssw0rd", Arrays.asList("Provider", "Foobar Role"));
```
mks-d: @dkayiwa this is basically all that is needed. A way to create a new user to which a list of (existing) roles is added. This is exactly what a custom auth. scheme might do.

What is provided is a list of role names.
* If a role name does indeed correspond to an existing role in OpenMRS, then the role is added and the user is created with that role as part of its set of roles.
* If a role name is unknown to OpenMRS, that role is skipped and the user is created anyway.

So if all the roles names are bogus, or if no role names are provided, it is possible to end up with a user with no roles at all. I think that's fine, it's up to the implementers to ensure that they reference role names that make sense.

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -177,6 +182,32 @@ public static void setDAO(ContextDAO dao) {
 		contextDAO = dao;
 	}
 
+	/**
+	 * Spring init method that sets the authentication scheme.
+	 */
+	static private void setAuthenticationScheme() {
```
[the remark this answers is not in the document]
dkayiwa: Did you see the above automated review comment?

api/src/main/java/org/openmrs/api/context/Daemon.java
```diff
@@ -102,7 +108,68 @@ public void run() {
 
 		return (Module) startModuleThread.returnedObject;
 	}
-	
+
+	/**
+	 * This method should not be called directly, only {@link ContextDAO#createUser(User, String)} can
+	 * legally invoke {@link #createUser(User, String)}.
+	 * 
+	 * @param user
```
dkayiwa: Did you just forget to document these params?

api/src/main/java/org/openmrs/api/context/Daemon.java
```diff
@@ -302,32 +369,32 @@ public void run() {
...
-		
+
 		try {
 			daemonThread.join();
 		}
 		catch (InterruptedException e) {
 			//Ignore
 		}
 	}
-
```
dkayiwa: Did you intentionally do all the formatting changes in this class?

gitcliff:
hello @dkayiwa and @mks-d i think we might have forgotten about squashing the commits on this issue .Thanks

gitcliff:
hello @dkayiwa and @mks-d i think we might have forgotten about squashing the commits on this issue .Thanks

api/src/main/java/org/openmrs/api/context/Daemon.java
mks-d: Yes, I'll do that in the next commit.

api/src/main/java/org/openmrs/api/context/Context.java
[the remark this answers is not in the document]
mks-d: @dkayiwa this was intentional by design. It is left private for the specific purpose of discouraging its use. But in fact Spring does use it [here](https://github.com/mks-d/openmrs-core/blob/bb074a562876407a12d89391d3e22e92a7223f88/api/src/main/resources/applicationContext-service.xml#L118).

api/src/main/java/org/openmrs/api/context/Daemon.java
mks-d: No, I hope I have solved that in the next commit...

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -1116,10 +1146,10 @@ public static void checkCoreDataset() {
 		ValidateUtil.setDisableValidation(disableValidation);
 
 		PersonName.setFormat(Context.getAdministrationService().getGlobalProperty(
-		    OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT));
+				OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT));
```
dkayiwa: You seem to have lots of formatting changes. 😊

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -79,28 +79,31 @@
...
 	
+	/**
+	 * The authentication scheme for this user
+	 */
+	private AuthenticationScheme authenticationScheme;
+	
 	/**
 	 * Default public constructor
 	 */
-	public UserContext() {
```
dkayiwa: Do we still have a default public constructor?

api/src/test/java/org/openmrs/api/context/DaemonTest.java
```diff
@@ -52,6 +58,52 @@ public void executeScheduledTask_shouldNotThrowErrorIfCalledFromATimerSchedulerT
...
+		
+		// verif
+		Assert.assertNotNull(createdUser.getId());
+		Set<String> roleNames = createdUser.getRoles().stream().map(r -> r.getName()).collect(Collectors.toSet());
+		Assert.assertTrue(roleNames.contains("Provider"));
+		Assert.assertFalse(roleNames.contains("Foobar Role")); // a bogus role name has just no impact on the created user
+	}
+	
+	@Test
```
dkayiwa: Any reason why you did not use the expected exception annotation?

dkayiwa: The other alternative is this: https://github.com/openmrs/openmrs-core/blob/917caf2265c70e0e32cd5f92644c9dcc47f6b845/api/src/test/java/org/openmrs/api/AdministrationServiceTest.java#L69

api/src/test/java/org/openmrs/api/context/DaemonTest.java
```diff
@@ -52,6 +58,52 @@ public void executeScheduledTask_shouldNotThrowErrorIfCalledFromATimerSchedulerT
...
+		u.setUsername("jdoe");
+		u.getPerson().setGender("F");
+		
+		// replay
+		User createdUser = Context.getContextDAO().createUser(u, "P@ssw0rd", Arrays.asList("Provider", "Foobar Role"));
+		
+		// verif
+		Assert.assertNotNull(createdUser.getId());
+		Set<String> roleNames = createdUser.getRoles().stream().map(r -> r.getName()).collect(Collectors.toSet());
+		Assert.assertTrue(roleNames.contains("Provider"));
```
dkayiwa: Do you have a test asserting that the bogus role was not added to the user?

api/src/test/java/org/openmrs/api/context/DaemonTest.java
```diff
@@ -52,6 +58,52 @@ public void executeScheduledTask_shouldNotThrowErrorIfCalledFromATimerSchedulerT
 		Assert.assertTrue(new PrivateSchedulerTask(task).runTheTest());
 	}
 	
+	@Test 
+	public void createUser_shouldThrowWhenCalledOutsideContextDAO() throws Throwable {
+		try {
+			Daemon.createUser(new User(), "password", null);
```
dkayiwa: As below, could we use ExpectedException?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -226,6 +235,25 @@ public User getUserByUuid(String uuid) {
...
+	}
+	
+	/**
+	 * @throws Exception 
+	 * @see org.openmrs.api.db.ContextDAO#createUser(User, String)
+	 */
+	@Override
+	@Transactional
+	public User createUser(User user, String password, List<String> roleNames) throws Exception {
+		return Daemon.createUser(user, password, roleNames);
```
dkayiwa: Do we have a test for this?

api/src/main/java/org/openmrs/api/db/ContextDAO.java
```diff
@@ -62,6 +63,32 @@
...
+	 * @since 2.3.0
+	 */
+	public User getUserByUsername(String username);
+	
+	/**
+	 * Creates a new user.
+	 * When the users are managed by a third-party authentication provider, it will happen that a successfully authenticated user still needs to be created in OpenMRS.
+	 * This method is made available to authentication schemes to create new users on the fly.
+	 * 
+	 * @param user
```
dkayiwa: Did you just forget to document these params?

api/src/main/java/org/openmrs/api/db/ContextDAO.java
```diff
@@ -62,6 +63,32 @@
...
+	 * @param password
+	 * @param roleNames A list of role names to fetch the roles to add to the user.
```
dkayiwa: the above description looks a bit confusing.

api/src/main/java/org/openmrs/api/db/ContextDAO.java
```diff
@@ -62,6 +63,32 @@
...
+	/**
+	 * Gets a user given the username. Privilege checks are not done here because this is used by the
+	 * {@link Context#getAuthenticatedUser()} or {@link Context#authenticate(org.openmrs.api.context.Credentials)} methods.
+	 * 
+	 * @param username The username of the user to fetch
+	 * @return The matched user, null otherwise.
+	 * 
+	 * @since 2.3.0
+	 */
+	public User getUserByUsername(String username);
```
dkayiwa: Where is this method used?

api/src/main/java/org/openmrs/api/context/UserContext.java
mks-d: @dkayiwa the only public constructor that remains takes an auth. scheme as argument. This has been around for a few commits yet. `UserContext` is always instantiated via `new` (it's never Spring-wired). All places where this is done should provide the auth. scheme from `Context`:
```java
new UserContext(Context.getAuthenticationScheme())
```

api/src/test/java/org/openmrs/api/context/DaemonTest.java
mks-d: @dkayiwa great, I actually had never used this. Coming in the next commit!

api/src/main/java/org/openmrs/api/db/ContextDAO.java
mks-d: @dkayiwa auth. schemes will need the ability to fetch users (typically to see if they already exist or not) to populate `Authenticated` implementations that then are able to return the authenticated user (see [here](https://github.com/openmrs/openmrs-core/blob/28990338f2096229268ac668d367bedce96d3d63/api/src/main/java/org/openmrs/api/context/Authenticated.java#L30)). In all likelihood users will be known in third-party auth. providers through their SSO usernames, this is why this new API is needed.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
mks-d: Well it is in fact already tested in `DaemonTest`, but also `Context.getContextDAO()` is not accessible from `ContextDAOTest` as this method has the package-only visibility and
- `Context` is in `org.openmrs.api.context`, while
- `ContextDAO` is in `org.openmrs.api.db`.

Do we really need to test it again, and if yes, what would you suggest to workaround the package visibility issue?

dkayiwa: In that case there is no need to test it again. 😊

web/src/test/java/org/openmrs/web/context/AltAuthSchemeContextTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @After
```
dkayiwa: What happens if we get rid of this entire method?

web/src/test/java/org/openmrs/web/context/AltAuthSchemeContextTest.java
```diff
@@ -0,0 +1,54 @@
...
+    Context.logout();
+  }
+
+  /**
+   * @see Context#authenticate(Credentials)
+   */
+  @Test
+  public void authenticate_shouldAuthenticateUserWithAlternateScheme() {
+    // replay
+    Context.logout();
```
dkayiwa: Do we need the above line? What happens if we get rid of it?

web/src/test/java/org/openmrs/web/context/AltAuthSchemeContextTest.java
```diff
@@ -0,0 +1,46 @@
...
+ * A Context loaded with a test Spring config that overrides the default auth scheme.
+ * 
+ * @see Context
+ */
+/*
+ * TODO Enable @ContextConfiguration on subclasses of BaseContextSensitiveTest
+ * 
+ * @see https://issues.openmrs.org/browse/TRUNK-381?focusedCommentId=254292&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-254292
+ */
+//@ContextConfiguration(locations = { "classpath*:AltAuthSchemeTestingApplicationContext.xml" })
```
dkayiwa: Since our practice is not to have commented out code, would you instead create tickets for things you want addressed later?

dkayiwa: The same goes for TODO

PR #2841, opened 2019-02-26
PLAT-34: Set the default value of the Login url Global Property as th…

mozzy11:
Ticket : https://issues.openmrs.org/browse/PLAT-34

### DIFF
--- a/api/src/main/java/org/openmrs/util/OpenmrsConstants.java
+++ b/api/src/main/java/org/openmrs/util/OpenmrsConstants.java
@@ -22,6 +22,7 @@ import java.util.Properties;
 
 import org.apache.commons.io.IOUtils;
 import org.openmrs.GlobalProperty;
+import org.openmrs.api.context.Context;
 import org.openmrs.api.handler.ExistingVisitAssignmentHandler;
 import org.openmrs.customdatatype.datatype.BooleanDatatype;
 import org.openmrs.customdatatype.datatype.FreeTextDatatype;
@@ -1110,6 +1111,8 @@ public final class OpenmrsConstants {
 				.add(new GlobalProperty(GP_DRUG_ORDER_DRUG_OTHER, "", "Specifies the uuid of the concept which represents drug other non coded"));
 		props.add(new GlobalProperty(GP_LOGIN_URL, LOGIN_URL,
 			"Responsible for defining the Authentication URL "));
+		
+	    props.add(changeDefaultLoginUrl_GP());
 		props.addAll(ModuleFactory.getGlobalProperties());
 		
 		return props;
@@ -1306,6 +1309,14 @@ public final class OpenmrsConstants {
 	
 	/** Value for the long person name format */
 	public static final String PERSON_NAME_FORMAT_LONG = "long";
+	
+	public static final String PLATFORM_UI_LOGIN_URL = "owa/openmrs-owa-platformui-0.1.0/index.html#/login.htm";
+	
+	public static GlobalProperty changeDefaultLoginUrl_GP() {	
+	Context.getAdministrationService().updateGlobalProperty(GP_LOGIN_URL, PLATFORM_UI_LOGIN_URL);
+	return Context.getAdministrationService().getGlobalPropertyObject(GP_LOGIN_URL);
+	}
+	
 	
 	private OpenmrsConstants() {
 	}


api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -1307,6 +1308,13 @@ public static String getOpenmrsProperty(String property) {
 	/** Value for the long person name format */
 	public static final String PERSON_NAME_FORMAT_LONG = "long";
 	
+	public static final String PLATFORM_UI_LOGIN_URL = "owa/openmrs-owa-platformui-0.1.0/index.html#/login";
+	
+	public void setDefaultLoginUrl_GP() {
+	Context.getAdministrationService().getGlobalPropertyObject(GP_LOGIN_URL).setPropertyValue(PLATFORM_UI_LOGIN_URL);
+	}
+
```
Ruhanga: Thank you @mozzy11 for the work. Did you test this?

mozzy11: @Ruhanga  , you mean testing it on my local installtion?? yes i did

Ruhanga: Can you also attach screen shots reflecting your changes as advised here https://wiki.openmrs.org/display/docs/Pull+Request+Tips ?

mozzy11: hi @Ruhanga , i made a small update in the PR , thanks . i have attached the image showing the default Global property value changed to point to the platform-ui Login url

mozzy11: updated commit is here  https://github.com/openmrs/openmrs-core/pull/2841/commits/5907c1b9806582a74dc88957ab6d95ce3ae11b95

Ruhanga: @mozzy11 , what happens when a User logs in with the set set Login url? How does the platform UI login page look like? What happens after a user logs in?

mozzy11:
@Ruhanga , the platform-ui is a web app that is just installed in the platform to provide a UI on the platform without the Needing the Legacy Hi. Unless the user has the web app installed in their platform , that URL will point to no page at all. But any way the ticket required I change the default Login-url global property to the platform _ui_login url , which I have done , because the initial default value has been pointing to the usual Login Url of the Legacy UI

Ruhanga:
Do your changes make any effect? If so, where are the visual effects of this in action given a platform instance created with use-case you have mentioned?

mozzy11:
> Do your changes make any effect? If so, where are the visual effects of this in action given a platform instance created with use-case you have mentioned?

@Ruhanga , this does not necesarilly make any visal changes , it was a matter of setting a default value of the Login_url Global property to have a default value of the platorm-ui Login URL.  the issue is icase any one wants to use that global property , it will have a default value of the platform-ui login URl.

mozzy11:
When you take a keen look at the image i attached on the ticket , youll discover that the default value of the Login_url Global property is now the platform-ui Login url as required by the ticket

Ruhanga:
Does this mean the default platform url being used leads to a dead end or links to a page that doesn't exist or can't be diplayed?

mozzy11:
@Ruhanga , noo , this is not a url being used. Actually as for now , in all the platform code , they refer to the login url as ` request.getContextPath() + "/login.htm"`    , so its not that the value i have set is used any where as for now .  the Login_url is actually a new Global property that was created not long ago . so wat i set was just its default value, just in case any implementation needs to use the platform-ui Login url , they can acces it from that global property , but not that its currently being used in code

mozzy11:
in other words , this is just a global property value , since ,in the time to come , we plan to completely use react web apps instead of the legacy ui , that global property would probably be very useful

mozzy11:
i updated the PR , so that the platform-ui login url is

owa/openmrs-owa-platformui-0.1.0/index.html#/login.htm
not

 owa/openmrs-owa-platformui-0.1.0/index.html#/login
i had forgorten to append the .htm at the end

mozzy11:
@Ruhanga  , i attached the image of the owa-paltform-ui Login page on the ticket. you can take a look at it.

PR #2851, opened 2019-03-07
TRUNK-5119: Updated the openmrs-api dependency of apache velocity fro…

ODORA0:
…m 1.6.2 to 2.0, https://issues.openmrs.org/browse/TRUNK-5119


ODORA0:
Yes, but I don't clearly understand where the error could be, because I think I updated rightly or unless of-course I wrongly updated the API-dependency.

PR #2895, opened 2019-05-15
TRUNK-5550: Upgrade org.slf4j Libraries

ODORA0:
**Issue worked on:** https://issues.openmrs.org/browse/TRUNK-5550

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -318,17 +318,17 @@
 			<dependency>
 				<groupId>org.slf4j</groupId>
 				<artifactId>slf4j-api</artifactId>
-				<version>1.6.0</version>
+				<version>1.7.26 </version>
 			</dependency>
 			<dependency>
 				<groupId>org.slf4j</groupId>
 				<artifactId>jcl-over-slf4j</artifactId>
-				<version>1.6.0</version>
+				<version>1.7.26 </version>
 			</dependency>
 			<dependency>
 				<groupId>org.slf4j</groupId>
 				<artifactId>slf4j-log4j12</artifactId>
-				<version>1.6.0</version>
+				<version>1.7.26 </version>
 				<scope>runtime</scope>
 			</dependency>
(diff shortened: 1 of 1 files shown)


mozzy11:
thanks @ODORA0  for the work ? it looks quite good.

mozzy11:
cc @dkayiwa

dkayiwa:
A beta versions is not a full release. We depend only on full releases.

ODORA0:
OK!! I think that should have been changed in the ticket description as well @mozzy11, I'll have that changed non the less

ODORA0:
@dkayiwa @mozzy11  I don't really get the error message here, when it says `[WARNING] The POM for org.slf4j:slf4j-api:jar:1.8.0 is missing, no dependency information available`

mozzy11:
@ODORA0 , yes you can look for the latest full release of that lib from their website

mozzy11:
@ODORA0 , yes you can look for the latest full release of that lib from their website

The error arises because you used a library version that doent exist. google out the latest full released version

ODORA0:
So according to https://www.slf4j.org/news.html and some various google resources show the tool is still in it's beta release stages, I had to check this as well https://mvnrepository.com/artifact/org.slf4j/slf4j-api to clarify. Maybe unless of-course we are shifting to 1.7.26 rather than the 1.8.x series I think that would do.

mozzy11:
@ODORA0 have you confirmed that  1.7.0 is the latest Full version??

ODORA0:
The latest full version 1.7.26 gave a coveralls failure when I pushed the changes, the 1.8.x series is in the beta stages so not really sure now

mozzy11:
then youwould probablygo with **1.7.26** , you may ignore the coveralls failure

dkayiwa:
Yes i agree going with 1.7.26 👍

PR #2897, opened 2019-05-16
TRUNK-5503:Upgrade ca.uhn.hapi: Libraries

ODORA0:
**Issue worked on:** https://issues.openmrs.org/browse/TRUNK-5503

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -334,27 +334,27 @@
 			<dependency>
 				<groupId>ca.uhn.hapi</groupId>
 				<artifactId>hapi-base</artifactId>
-				<version>2.0</version>
+				<version>2.1</version>
 			</dependency>
 			<dependency>
 				<groupId>ca.uhn.hapi</groupId>
 				<artifactId>hapi-structures-v23</artifactId>
-				<version>2.0</version>
+				<version>2.1</version>
 			</dependency>
 			<dependency>
 				<groupId>ca.uhn.hapi</groupId>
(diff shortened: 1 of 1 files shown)


ODORA0:
I had it successful then tried it again and it was a failure giving the same error so pushing it seems like the error remains, so am not really sure how upgrading those libs results to that, but I tried understanding how the cannot find symbol error can come about, vain

pom.xml
```diff
@@ -334,27 +334,27 @@
 			<dependency>
 				<groupId>ca.uhn.hapi</groupId>
 				<artifactId>hapi-base</artifactId>
-				<version>2.0</version>
+				<version>2.3</version>
```
Ruhanga: I think the closes you can upgrade to is 2.1 because of the [error source](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/hl7/impl/HL7ServiceImpl.java#L66) .
You can refer to [this doc](http://javadox.com/ca.uhn.hapi/hapi-base/2.1/ca/uhn/hl7v2/app/MessageTypeRouter.html)

@dkayiwa, would this be good enough ?

ODORA0: Ok!! let me at-least try it locally and see if the build passes

ODORA0: @dkayiwa @Ruhanga I have a build success after changing the upgrade to 2.1 so I guess am pushing these changes

PR #2898, opened 2019-05-16
Beginnings of the master single page application.

joeldenning:
See https://github.com/openmrs/openmrs-rfc-frontend

There is a substantial risk to merging in this pull request, because it changes to default behavior for all routes in openmrs. I was unable to test that frontend routes for legacyui, adminui, refapp are still working, which is important. Reason is that I couldn't get those modules to boot up correctly locally.

## Checklist

  No? -> execute above command

### DIFF
--- a/web/src/main/java/org/openmrs/web/SinglePageApplicationServlet.java
+++ b/web/src/main/java/org/openmrs/web/SinglePageApplicationServlet.java
@@ -1,0 +1,33 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.web;
+
+import javax.servlet.RequestDispatcher;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SinglePageApplicationServlet extends HttpServlet {
+	
+  private static final Logger log = LoggerFactory.getLogger(SinglePageApplicationServlet.class);
+
+	@Override
+	protected void doGet(HttpServletRequest request, HttpServletResponse response) {
+    try {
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -306,6 +306,15 @@
 		<servlet-name>jsp</servlet-name>
 		<url-pattern>*.jsp</url-pattern>
 	</servlet-mapping>
+
+	<servlet>
+		<servlet-name>default</servlet-name>
```
joeldenning: See https://tomcat.apache.org/tomcat-7.0-doc/default-servlet.html.

This is why this PR has such a big impact. Any route not covered by any of the other servlets will end up hitting this route.

web/src/main/java/org/openmrs/web/SinglePageApplicationServlet.java
```diff
@@ -0,0 +1,33 @@
...
+import org.slf4j.LoggerFactory;
+
+public class SinglePageApplicationServlet extends HttpServlet {
+	
+  private static final Logger log = LoggerFactory.getLogger(SinglePageApplicationServlet.class);
+
+	@Override
+	protected void doGet(HttpServletRequest request, HttpServletResponse response) {
+    try {
+      RequestDispatcher dispatcher = request.getRequestDispatcher("/master-single-page-application.jsp");
```
mseaton: @joeldenning If we are worried about the impact of this in the short term, we could make this configurable.  eg. If openmrs-runtime.properties file contains "enable-single-page-application=true", then execute this dispatch, otherwise fall back to legacy behavior

bmamlin:
Does the SPA servlet _have_ to cast such a broad net – i.e., are there specific requests we are expecting that would not be made to `/master-single-page-application.jsp*` that need to be handled by the SPA servlet?

I was assuming a SPA, by design, would be using URLs in a pattern like `{PATH_TO_SINGLE_SPA_MASTER}#/the/path/for/spa/routers`. If all calls to the SPA are expected to fall under a predictable path (the master page), then it seems far less risky to introduce an OpenMRS SPA module that could own `{OPENMRS}/module/spa` and everything below it. In the future, we could support `{OPENMRS}/spa` (like we've done with `{OPENMRS}/ws` for web services).

@joeldenning, if Single SPA can work as long as it receives all calls to a specific path or below, then we could create a simple OpenMRS SPA module for you that could serve as the handler for all SPA traffic.

dkayiwa:
I have made a commit to the SPA module, which has an overridable spa.baseUrl setting that defaults to {OPENMRS}/spa
https://github.com/openmrs/openmrs-module-spa/commit/8df90c7a4fb9418f868e3a4deaed83371f96a261

@joeldenning if it works for you, then we can cancel this pull request.

joeldenning:
Yep 👍 sounds good

PR #2900, opened 2019-05-17
TRUNK-5492 Replace Person and Patient hbm mapping file with annotations

wluyima:
TRUNK-5492 Replace Patient hbm mapping file with annotations

TRUNK-5492 Fixed failing tests

TRUNK-5492 Added special auditable fields to Patient table

TRUNK-5492 Updated Auditable interceptor to set patient auditable fields instead of PatientSaveHandler

TRUNK-5492 Renamed properties back to their original names

TRUNK-5492 Removed unused code

TRUNK-5492 Removed unused code

TRUNK-5492 Removed changes in BaseOpenmrsData

TRUNK-5492 Removed changes in PersonServiceTest

TRUNK-5492 Removed explicit lazy loading from JPA mappings

TRUNK-5492 Removed more explicit lazy loading from JPA mappings

TRUNK-5492 Replaced @Sort with @SortNatural

TRUNK-5492 Removed more explicit lazy loading from JPA mappings

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5492

  No? -> execute above command


api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
```diff
@@ -583,6 +584,10 @@ public void initializeInMemoryDatabase() throws SQLException {
 			throw new RuntimeException(
 			        "You shouldn't be initializing a NON in-memory database. Consider unoverriding useInMemoryDatabase");
 
+		//Because creator property in the superclass is mapped as nullable by the autoddl tool, we need to first
```
wluyima: I need to rephrase this, meant to say:

Because creator property in the superclass is mapped with optional set to false, the autoddl tool marks the column as not nullable

api/src/main/java/org/openmrs/Patient.java
```diff
@@ -172,7 +222,13 @@ public void setPersonId(Integer personId) {
 	 * @see org.openmrs.PatientIdentifier
 	 */
 	public void setIdentifiers(Set<PatientIdentifier> identifiers) {
-		this.identifiers = identifiers;
+		if (identifiers == null) {
```
wluyima: Probably this is no longer necessary

dkayiwa:
The link to the JIRA ticket does not look correct.

wluyima:
Fixed

api/src/main/java/org/openmrs/Patient.java
```diff
@@ -99,6 +156,9 @@ public Patient(Patient patient) {
 	 * @return internal identifier for patient
 	 */
 	public Integer getPatientId() {
+		if (this.patientId == null) {
```
dkayiwa: Do you mind explaining why you needed to do this?

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -608,7 +608,7 @@ public boolean isIdentifierInUseByAnotherPatient(PatientIdentifier patientIdenti
 		        && patientIdentifier.getIdentifierType().getUniquenessBehavior() == UniquenessBehavior.LOCATION;
 		
 		// switched this to an hql query so the hibernate cache can be considered as well as the database
-		String hql = "select count(*) from PatientIdentifier pi, Patient p where pi.patient.patientId = p.patient.patientId "
```
dkayiwa: Is this a bug fix? Or is it a change caused by switching to annotations?

api/src/main/java/org/openmrs/api/handler/PatientSaveHandler.java
```diff
@@ -45,5 +46,10 @@ public void handle(Patient patient, User creator, Date dateCreated, String other
 				}
 			}
 		}
+
+		if(patient.getPatientId() == null){
+			patient.setCreator(Context.getAuthenticatedUser());
```
dkayiwa: Is this a bug fix? Or is it a change caused by switching to annotations?

api/src/main/java/org/openmrs/api/handler/PersonSaveHandler.java
```diff
@@ -42,13 +45,16 @@ public void handle(Person person, User creator, Date dateCreated, String other)
 		
 		// address collection
 		if (person.getAddresses() != null && !person.getAddresses().isEmpty()) {
+			Set<Address> blankAddresses = new HashSet<>();
 			for (PersonAddress pAddress : person.getAddresses()) {
-				if (pAddress.isBlank()){
-					person.removeAddress(pAddress);
+				if (pAddress.isBlank()) {
```
dkayiwa: Is this a bug fix? Or is it a change caused by switching to annotations?

api/src/main/java/org/openmrs/Patient.java
wluyima: In-memory, this field is null when you create the patient record since it is technically mapped to a FK column which is why it is mapped as not insertable and not updatable, I wonder why hibernate was setting it with xml but not with annotations though, my guess is there could be a hibernate specific annotations to make fix that JPA doesn't support, will look into it a little a bit to see if there is another work around

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
wluyima: This is a bug fix because it was wrong before, annotations just exposed it, p is an alias for patient, and patient for sure doesn't have a patient field

api/src/main/java/org/openmrs/api/handler/PatientSaveHandler.java
wluyima: Not really but sort of, more like bad design, this is needed because creator and dateCreated fields exist both in  BaseOpenmrsData and Patient, this is actually bad practice because it defeats the purpose of inheritance, and this where xml and annotations differ, because annotations are pure java whereas xml isn't, with xml you get away with bad things just like we were doing before which is why I say sort of a bug.

Anyways, long story short, hibernate is unable to set these fields via the AuditableInterceptor mechanism because based on the ClassMedata it has built for Patient class, it sees the same fields on the superclass and the subclass and treats them as different because they are declared by different classes anyways, it only sets the values of the fields in the superclass and not the subclass so we have to manually set them from here as we used to do before to get around it.

api/src/main/java/org/openmrs/api/handler/PersonSaveHandler.java
wluyima: Definitely a bug, it's java 101, you can't remove an element from the same Iterable you are looping over, you get a ConcurrentModificationException

api/src/test/java/org/openmrs/api/PatientServiceTest.java
```diff
@@ -2096,7 +2094,7 @@ public void savePatientIdentifier_shouldAllowLocationToBeNullWhenLocationBehavio
 	 */
 	@Test(expected = ValidationException.class)
 	public void savePatientIdentifier_shouldAllowLocationToBeNullWhenLocationBehaviourIsRequired() {
-		PatientIdentifier patientIdentifier = patientService.getPatientIdentifier(7);
+		PatientIdentifier patientIdentifier = patientService.getPatientIdentifier(9);
```
wluyima: This was a bad patient, they are voided but it worked before because they had a non-voided identifier which I changed to voided and it got broken, so I had to pick another non-voided patient with a non voided identifier

api/src/test/resources/org/openmrs/api/db/hibernate/include/HibernatePatientDAOTest-patients.xml
```diff
@@ -68,7 +68,7 @@
         Voided patient with non-voided name
     -->
     <person person_id="60" gender="M" birthdate="1998-09-30 12:34:56.7" birthdate_estimated="0" dead="false" creator="1" date_created="2014-01-09 00:00:00.0" voided="false" uuid="7443D9DD-CA51-4BD9-9E68-7C43628B82DF"/>
-    <person_name person_name_id="60" preferred="true" person_id="60" prefix="Mr." given_name="Meriadoc" middle_name="" family_name="Brandybuck" family_name2="" creator="1" date_created="2014-01-09 00:00:00.0" voided="false" void_reason="" uuid="61ACD71F-989B-4B4A-B00E-0503B756099F"/>
+    <person_name person_name_id="60" preferred="true" person_id="60" prefix="Mr." given_name="Meriadoc" middle_name="" family_name="Brandybuck" family_name2="" creator="1" date_created="2014-01-09 00:00:00.0" voided="true" void_reason="" uuid="61ACD71F-989B-4B4A-B00E-0503B756099F"/>
```
wluyima: This was a construed scenario that should never arise based on the semantics of our API because a voided patient should never have a non-voided voided name, If an admin tinkers with data in the DB, they better know what they are doing.

api/src/test/resources/org/openmrs/api/db/hibernate/include/HibernatePersonDAOTest-people.xml
```diff
@@ -56,7 +56,7 @@
         Voided person with non-voided name
     -->
     <person person_id="50" gender="M" birthdate="1998-09-30 12:34:56.7" birthdate_estimated="0" dead="false" creator="1" date_created="2013-04-25 12:34:56.7" voided="true" uuid="2B856CBF-23D4-4D37-B2F6-FF8BD23A0434"/>
-    <person_name person_name_id="50" preferred="true" person_id="50" prefix="Mr." given_name="voided-bravo" middle_name="voided-foxtrot" family_name="voided-kilo" family_name2="voided-oscar" creator="1" date_created="2013-04-18 12:34:56.7" voided="false" void_reason="" uuid="3CCEB428-2B96-48AF-AAA3-36DB80601C9B"/>
+    <person_name person_name_id="50" preferred="true" person_id="50" prefix="Mr." given_name="voided-bravo" middle_name="voided-foxtrot" family_name="voided-kilo" family_name2="voided-oscar" creator="1" date_created="2013-04-18 12:34:56.7" voided="true" void_reason="" uuid="3CCEB428-2B96-48AF-AAA3-36DB80601C9B"/>
```
wluyima: Same here, name has to be voided if patient is voided

api/src/test/resources/org/openmrs/include/standardTestDataset.xml
```diff
@@ -302,7 +302,7 @@
   <patient_identifier patient_identifier_id="4" patient_id="7" identifier="6TS-4" identifier_type="1" preferred="1" location_id="1" creator="1" date_created="2006-01-18 00:00:00.0" voided="false" void_reason="" uuid="489edb7f-0786-4726-88a3-f0fcd2505d1a"/>
   <patient_identifier patient_identifier_id="5" patient_id="8" identifier="7TU-8" identifier_type="1" preferred="0" location_id="1" creator="1" date_created="2006-01-18 00:00:00.0" voided="false" void_reason="" uuid="2c23455b-dee3-450c-914f-fe4bd0748d81"/>
   <patient_identifier patient_identifier_id="6" patient_id="8" identifier="ABC123" identifier_type="2" preferred="0" location_id="1" creator="1" date_created="2006-01-18 00:00:00.0" voided="true" void_reason="Incorrect"/>
-  <patient_identifier patient_identifier_id="7" patient_id="999" identifier="XYZ" identifier_type="2" preferred="1" location_id="1" creator="1" date_created="2006-01-18 00:00:00.0" voided="false" void_reason=""/>
+  <patient_identifier patient_identifier_id="7" patient_id="999" identifier="XYZ" identifier_type="2" preferred="1" location_id="1" creator="1" date_created="2006-01-18 00:00:00.0" voided="true" void_reason=""/>
```
wluyima: Same here, which is why I changed the patient for the test I mentioned earlier

dkayiwa:
I take the switch from xml to annotations as a non simple change, which may have un expected effects. And therefore, i would feel more comfortable if we separated bug fixes that have nothing to do with xml vs annotations, in their own ticket and pull request. That would also simplify the reviews by easily telling what exactly was needed to change for the sake of annotations.

wluyima:
I see what you are saying, and am not opposed to it, the reasons why I did it this way are:
- These are 'bugs', I want to think of them as flaws, that are not all easily reproducible before changing to annotations, xml was just masking them, I think in terms of TDD how will you create a failing test to reproduce the issue and fix the issue?
- If I create separate tickets which imply separate PR then you can't merge this PR before they all merged and you can't reproduce them before you merge this, kind of a chicken-egg dilemma there.

I made sure I added some quick notes to explain what each change attempts to fix to make it easier for the reviewer

wluyima:
@dkayiwa did you see my response?

api/src/main/java/org/openmrs/Patient.java
dkayiwa: What do you think of adding this comment to the code?

api/src/main/java/org/openmrs/api/handler/PatientSaveHandler.java
dkayiwa: What do you think of adding this comment to the code?

api/src/main/java/org/openmrs/Patient.java
wluyima: I can but I see that you already merged the PR

dkayiwa: You can do it as a followup commit.

wluyima:
Sounds good!

PR #2907, opened 2019-05-18
Replaced Obs hbm mapping file with annotations

reagan-meant:
TRUNK-5494 Replace Obs hbm mapping file with annotations
https://issues.openmrs.org/browse/TRUNK-5494?filter=-1

Description of what I changed
I replaced the Obs.hbm.xml with Annotation basis of db mapping between the POJO and database.

  My IDE is configured to follow the [code style]


api/src/main/java/org/openmrs/Obs.java
```diff
@@ -90,13 +112,17 @@
 	private static final String FORM_NAMESPACE_PATH_SEPARATOR = "^";
 	
 	private static final int FORM_NAMESPACE_PATH_MAX_LENGTH = 255;
-	
+	@Id
+	@GeneratedValue(strategy = GenerationType.SEQUENCE)
```
wluyima: I think AUTO generator type is a better option, it lets hibernate decide the best based on the underlying DB

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -106,49 +132,64 @@
...
-	
+	@JoinColumn(name = "value_coded_name_id")
+	@ManyToOne
 	protected ConceptName valueCodedName;
-	
+	@JoinColumn(name = "value_drug")
+	@ManyToOne
 	protected Drug valueDrug;
-	
+	@Column(name = "value_group_id", length = 11)
```
wluyima: Length is redundant for some types, I would only add it for Date fields and Strings where you want it to be different from the default value of 255

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -106,49 +132,64 @@
...
+	@Column(name = "value_numeric", length = 22)
 	protected Double valueNumeric;
-	
+	@Column(name = "value_modifier", length = 2)
 	protected String valueModifier;
-	
+	@Column(name = "value_text", length = 65535)
 	protected String valueText;
-	
+	@Column(name = "value_complex", length = 255)
```
wluyima: 255 is the default, so you don't have to set it

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -106,49 +132,64 @@
...
 	protected String valueComplex;
 	
 	// ComplexData is not persisted in the database.
 	protected transient ComplexData complexData;
-	
+	@Column(name = "comments", length = 255)
 	protected String comment;
-	
+	@Transient
```
wluyima: Is this really transient? Try to match the original configuration in the xml file

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -106,49 +132,64 @@
...
+	@Column(name = "person_id", nullable = false, length = 11)
 	protected transient Integer personId;
-	
+	@JoinColumn(name = "person_id", nullable = false)
```
wluyima: Using optional attribute of many to one is better than this nullable attribute because then hibernate can check null for values at runtime too

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -1161,6 +1202,9 @@ public Boolean hasPreviousVersion() {
 	 * @see Auditable#setCreator(User)
 	 */
 	@Override
+	@Access(AccessType.PROPERTY)
```
wluyima: This is already mapped in the superclass

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -1171,6 +1215,8 @@ public void setCreator(User creator) {
 	 * @see Auditable#setDateCreated(Date)
 	 */
 	@Override
+	@Access(AccessType.PROPERTY)
```
wluyima: This is already mapped in the superclass

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -1301,6 +1347,9 @@ public Interpretation getInterpretation() {
 	/**
 	 * @since 2.1.0
 	 */
+	@Access(AccessType.PROPERTY)
```
wluyima: This is already mapped in the superclass

api/src/main/java/org/openmrs/Obs.java
reagan-meant: I had used the default used in the hbm.xml file originally...so should i change to auto then?

api/src/main/java/org/openmrs/Obs.java
reagan-meant: i was trying to maintain the defaults used in the hbm.xml file

api/src/main/java/org/openmrs/Obs.java
reagan-meant: Alright then...

api/src/main/java/org/openmrs/Obs.java
reagan-meant: True.I have noticed.

api/src/main/java/org/openmrs/Obs.java
reagan-meant: True.I have noticed.

api/src/main/java/org/openmrs/Obs.java
reagan-meant: True.I have noticed.

api/src/main/java/org/openmrs/Obs.java
reagan-meant: It was an oversight...It is not transient but rather not insertable.....its generated by the db itself

api/src/main/java/org/openmrs/Obs.java
reagan-meant: Alright then

reagan-meant:
Hope you have seen the changes I have made.

wluyima:
@reagan-meant also remember to remove the mapping file and its entry from the hibernate.cfg.xml file

reagan-meant:
Am running into these test failures https://pastebin.com/M9JxkPRL once i removed the obs.hbm file and its entry in the cfg.xml file

wluyima:
I suggest run a single test in ObsServiceTest to minimise unnecessary logs and be able to pin point the actual issue, my guess is there is something wrong with one or more mappings

wluyima:
@reagan-meant the builds fails, did you run the tests locally to ensure nothing is broken?

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -106,49 +131,63 @@
...
+	@JoinColumn(name = "obs_group_id")
+	@ManyToOne
 	protected Obs obsGroup;
 	
 	/**
 	 * The list of obs grouped under this obs.
 	 */
 	@AllowDirectAccess
+	@JoinColumn(name = "obs_group_id")
+	@OneToMany(cascade = CascadeType.REMOVE)
```
wluyima: I think you need to set mappedBy attribute

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -106,49 +131,63 @@
...
 	 */
+	@JoinColumn(name = "obs_group_id")
+	@ManyToOne
 	protected Obs obsGroup;
 	
 	/**
 	 * The list of obs grouped under this obs.
 	 */
 	@AllowDirectAccess
+	@JoinColumn(name = "obs_group_id")
```
wluyima: Is this necessary?

reagan-meant:
I keep on getting the same dbunit error...I am stuck how to proceed from here...https://pastebin.com/M9JxkPRL
I have tried researching on it but am hitting a dead end.....any clues?

wluyima:
Replaced with https://github.com/openmrs/openmrs-core/pull/2965

PR #2953, opened 2019-05-31
TRUNK-5473 - Refactor PersonValidator#validate method

fruether:
TRUNK-5473 - Refactor PersonValidator#validate method
This PR is highly inspired by (https://github.com/openmrs/openmrs-core/pull/2804) Thanks for the good previous work.
- PersonValidator#validate  simplified by creating sub-methods for validatePersonAddresses, validatePersonNames and validatePersonHasAtLeastOneNonVoidedName
- I added all validate methods into PersonValidator and not as suggested by previous work into the Person class
- I am not sure how the PersonValidator should deal with a Null in a personAddress.  I made it to simply continue without failing. See here fore more details: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/validator/PersonAddressValidator.java#L67

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5473

- [X] My pull request only contains **ONE single commit**

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/validator/PersonValidator.java
+++ b/api/src/main/java/org/openmrs/validator/PersonValidator.java
@@ -69,23 +69,59 @@ public class PersonValidator implements Validator {
 		
 		Person person = (Person) target;
 		
-		int index = 0;
-		boolean atLeastOneNonVoidPersonNameLeft = false;
-		for (PersonName personName : person.getNames()) {
-			errors.pushNestedPath("names[" + index + "]");
-			personNameValidator.validate(personName, errors);
-			if (!personName.getVoided()) {
-				atLeastOneNonVoidPersonNameLeft = true;
-			}
-			errors.popNestedPath();
-			index++;
-		}
-		if (!person.getVoided() && !atLeastOneNonVoidPersonNameLeft) {
+		validatePersonNames(person, errors);
+		
+		if (!person.getVoided() && !validatePersonHasAtLeastOneNonVoidedName(person)) {
 			errors.rejectValue("names", "Person.shouldHaveAtleastOneNonVoidedName");
 		}
 		
-		// validate the personAddress
-		index = 0;
+		validatePersonAddresses(person, errors);
+		
+		validateBirthDate(errors, person.getBirthdate());
+		validateDeathDate(errors, person.getDeathDate(), person.getBirthdate());
+		if (person.getVoided()) {
+			ValidationUtils.rejectIfEmptyOrWhitespace(errors, "voidReason", "error.null");
+		}
+		
+		validateDeathCause(person, errors);
+		
+		ValidateUtil.validateFieldLengths(errors, Person.class, "gender", "personVoidReason");
+	}
+	
+	/**
+	 * Checks if a person contains any non voided PersonName attribute
+	 *
+	 * @param person The person that is analysed for voided names
+	 * @return true if at leas one personName is not voided
+	 */
+	private boolean validatePersonHasAtLeastOneNonVoidedName(Person person) {
+			return person.getNames().stream().anyMatch(personName -> !personName.getVoided());
+	}
+	
+	/**
+	 * Checks if the names of a person fulfill the expected criteria  
+	 *
+	 * @param person The person whose names should be validated
+	 * @param errors Stores information about errors encountered during validation.
+	 */
+	private void validatePersonNames(Person person, Errors errors) {
+		int index = 0;
+		for (PersonName personName : person.getNames()) {
+			errors.pushNestedPath("names[" + index + "]");
+			personNameValidator.validate(personName, errors);
+			errors.popNestedPath();
+			index++;
+		}
+	}
+	
+	/**
+	 * Checks if the addresses of a person fulfill the expected criteria  
+	 *
+	 * @param person The person whose addresses should be validated
+	 * @param errors Stores information about errors encountered during validation.
+	 */
+	private void validatePersonAddresses(Person person, Errors errors) {
+		int index = 0;
 		for (PersonAddress address : person.getAddresses()) {
 			try {
 				errors.pushNestedPath("addresses[" + index + "]");
@@ -96,13 +132,15 @@ public class PersonValidator implements Validator {
 				index++;
 			}
 		}
-		
-		validateBirthDate(errors, person.getBirthdate());
-		validateDeathDate(errors, person.getDeathDate(), person.getBirthdate());
-		
-		if (person.getVoided()) {
-			ValidationUtils.rejectIfEmptyOrWhitespace(errors, "voidReason", "error.null");
-		}
+	}
+	
+	/**
+	 * Checks if the death cause and death cause non coded fulfill the business logic
+	 *
+	 * @param person The person whose death attributes should be validated.
+	 * @param errors Stores information about errors encountered during validation.
+	 */
+	private void validateDeathCause(Person person, Errors errors){
 		if (person.getDead()) {
 			if(person.getCauseOfDeath() != null && person.getCauseOfDeathNonCoded() != null) {
 				errors.rejectValue("causeOfDeath", "Person.dead.shouldHaveOnlyOneCauseOfDeathOrCauseOfDeathNonCodedSet");
@@ -111,8 +149,6 @@ public class PersonValidator implements Validator {
 				errors.rejectValue("causeOfDeath", "Person.dead.causeOfDeathAndCauseOfDeathNonCodedNull");
 			}
 		}
-		
-		ValidateUtil.validateFieldLengths(errors, Person.class, "gender", "personVoidReason");
 	}
 	
 	/**
--- a/api/src/test/java/org/openmrs/validator/PersonValidatorTest.java
+++ b/api/src/test/java/org/openmrs/validator/PersonValidatorTest.java
@@ -11,12 +11,15 @@ package org.openmrs.validator;
 
 import java.util.Calendar;
 import java.util.Date;
+import java.util.HashSet;
+import java.util.Set;
 
 import org.junit.Assert;
 import org.junit.Test;
 import org.openmrs.Concept;
 import org.openmrs.Patient;
 import org.openmrs.Person;
+import org.openmrs.PersonAddress;
 import org.openmrs.api.context.Context;
 import org.openmrs.test.BaseContextSensitiveTest;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -230,6 +233,25 @@ public class PersonValidatorTest extends BaseContextSensitiveTest {
 		personValidator.validate(person, errors);
 
 		Assert.assertFalse(errors.hasErrors());
+	}
+	
+	@Test(expected = IllegalArgumentException.class)
+	public void validate_shouldThrowExceptionWhenAddressIsNull() {
+		Person person = new Patient(1);
+		Set<PersonAddress> addresses =  new HashSet<>();
+		addresses.add(null);
+		
+		person.setDead(true);
+		person.setCauseOfDeathNonCoded("Some text describing Cause of Death");
+		person.setPersonVoided(true);
+		person.setPersonVoidReason("voidReason");
+		person.setGender("g");
+		person.setAddresses(addresses);
+		
+		Errors errors = new BindException(person, "patient");
+		PersonValidator personValidator = new PersonValidator();
+		
+		personValidator.validate(person, errors);
 	}
 
 }


dkayiwa:
The link to the ticket does not look correct.

fruether:
@dkayiwa I changed the link. It works now

fruether:
@dkayiwa, @chankruze  any updates regarding this PR :) ?

api/src/main/java/org/openmrs/validator/PersonValidator.java
```diff
@@ -68,41 +68,79 @@ public void validate(Object target, Errors errors) {
 		}
 		
 		Person person = (Person) target;
+		if (person.getVoided()) {
```
dkayiwa: Is this refactoring or new functionality?

dkayiwa:
@fruether these changes seem to be dealing with more than what the ticket description requires.

api/src/main/java/org/openmrs/validator/PersonValidator.java
fruether: No new functionality. I just rearrange this line of code. It was previously in the end. Now I moved it up to make the code easier/more readable:

See line 103: https://github.com/openmrs/openmrs-core/pull/2953/commits/69fba31b11fd3454630bb643e9dcb1685833a079#diff-7014d6c9eb6084f32ffc5c659e8ee524L103

dkayiwa: We usually do such formatting changes in separate pull requests.

fruether:
@dkayiwa the issue requested a refactoring. I just did a refactoring. As part of the refactoring I moved code up/down and eventually created new methods to make everything more simple. Therefore,  did not change any behaviour. The code and functionality is the same. Just hopefully easier to read.

Can you explain more in depth why you think I did more?


Furthermore, I noticed a strange behaviour which I documented in the issue bur did not change at all: https://issues.openmrs.org/browse/TRUNK-5473?focusedCommentId=256840&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-256840

api/src/main/java/org/openmrs/validator/PersonValidator.java
fruether: I did so because for me it seemed to be a valid refactoring operation. Do you want this Pull Requests to be split up?

dkayiwa: Because any refactoring has potential to introduce unintended bugs, i find it easier and cleaner to focus on only what the ticket was meant for. You can always create more tickets for further improvements.

fruether: My understanding was that refactoring to make a method simpler would also mean to reduce over-complex if conditions and make the code easier by layout. Anyways I removed the difference in the if condition and change in order. It should now only contain the new methods.

api/src/test/java/org/openmrs/validator/PersonValidatorTest.java
```diff
@@ -231,5 +234,25 @@ public void validate_shouldNotFailWhenDeathCauseNotCodedIsSet() {
 
 		Assert.assertFalse(errors.hasErrors());
 	}
+	
+	@Test
+	public void validate_shouldNotFailWhenOneAddressIsNull() {
```
fruether: @dkayiwa should that test be removed as well?  Especially in context of this content I think it should be part of this PR: https://issues.openmrs.org/browse/TRUNK-5473?focusedCommentId=256840&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-256840

api/src/main/java/org/openmrs/validator/PersonValidator.java
```diff
@@ -111,8 +151,6 @@ else if(person.getCauseOfDeath() == null && person.getCauseOfDeathNonCoded() ==
 				errors.rejectValue("causeOfDeath", "Person.dead.causeOfDeathAndCauseOfDeathNonCodedNull");
 			}
 		}
-
```
dkayiwa: Did you intentionally remove this?

api/src/test/java/org/openmrs/validator/PersonValidatorTest.java
dkayiwa: Tests to catch possible refactoring bugs, can stay. 😊

api/src/main/java/org/openmrs/validator/PersonValidator.java
fruether: I did not remove line 114 and 115. I put it down where it was before I changed the code according to the previous pull request. Git Change Log does not recognize it this way though:
https://github.com/openmrs/openmrs-core/pull/2953/commits/4c40dade91b459e578231a2e593828832a5b51d9#diff-7014d6c9eb6084f32ffc5c659e8ee524R87

api/src/main/java/org/openmrs/validator/PersonValidator.java
```diff
@@ -69,40 +69,80 @@ public void validate(Object target, Errors errors) {
...
+	 */
+	private void validatePersonAddresses(Person person, Errors errors) {
+		int index = 0;
 		for (PersonAddress address : person.getAddresses()) {
 			try {
 				errors.pushNestedPath("addresses[" + index + "]");
-				ValidationUtils.invokeValidator(personAddressValidator, address, errors);
+				personAddressValidator.validate(address, errors);
+			}
+			catch (NullPointerException exception) {
```
dkayiwa: Are you intentionally ignoring the null pointer exception?

fruether: Yes. Reason (1) To make explicit that a exception is thrown if the value is null. That fact is pretty much hidden by the finally block. I think that overcomplicates the code understanding  (2) To not change the behaviour of the code. I also added a test case to prove:

The issue lies in the fact that personAddressValidator validator does check if the value is null and then throws an exception. That is not how the other classes behave: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/validator/PersonAddressValidator.java#L67

dkayiwa: I would avoid changing behavior in a refactoring pull request. Even with the test, you never know what bugs may be introduced which the test does not catch.

fruether: If I remove the catch() I also have to remove the test case I added. Since it is some weeks ago I forgot that I added this catch() since the newly added test case of me would not work otherwise. Which means I changed the behaviour. I assume you want the test case & try/catch removed?

dkayiwa: Whatever it takes to restore the original behaviour.

api/src/test/java/org/openmrs/validator/PersonValidatorTest.java
```diff
@@ -231,5 +234,28 @@ public void validate_shouldNotFailWhenDeathCauseNotCodedIsSet() {
...
+		person.setGender("g");
+		person.setAddresses(addresses);
+		
+		Errors errors = new BindException(person, "patient");
+		PersonValidator personValidator = new PersonValidator();
+		
+		try {
+			personValidator.validate(person, errors);
+		}catch(NullPointerException np) {
+			Assert.assertNotNull(np);
```
fruether: @dkayiwa I reverted the old state. I would suggest to create a new ticket to fix this behaviour if you agree that this is not intended.

dkayiwa: Feel free to create a the new ticket.

fruether: @dkayiwa are there any more issues now stopping the merging? As far as I am know all your comments are resolved now :)

fruether:
@dkayiwa All your comments should be applied. Anything more open regarding this issue?

dkayiwa:
@fruether did you check if sonar is now happy with this method?

fruether:
I do not have the Developer Edition of Sonar. That's why I can't analyze
this branch.
However, the new introduced functions move complex parts out of the method
and hence Cyclomatic Complexity is at least reduced by 3 points so it
should be <= 10.
It could have been reduced a bit more by the changes that were declined
during the review process since less conditions were in the if clause .

Am Sa., 20. Juli 2019 um 22:45 Uhr schrieb dkayiwa <notifications@github.com

dkayiwa:
@fruether did you get a chance to look at this? https://wiki.openmrs.org/display/docs/Sonar

fruether:
@dkayiwa I am not allowed to execute sonar:

> ERROR] Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.4.0.905:sonar (default-cli) on project openmrs: You're only authorized to execute a local (preview) SonarQube analysis without pushing the results to the SonarQube server. Please contact your SonarQube administrator. -> [Help 1]

api/src/test/java/org/openmrs/validator/PersonValidatorTest.java
```diff
@@ -231,5 +234,28 @@ public void validate_shouldNotFailWhenDeathCauseNotCodedIsSet() {
 
 		Assert.assertFalse(errors.hasErrors());
 	}
+	
+	@Test()
+	public void validate_shouldNotFailWhenOneAddressIsNull() {
+		Person person = new Patient(1);
+		Set<PersonAddress> addresses =  new HashSet<>();
+		addresses.add(null);
```
dkayiwa: What use case would lead to a null address added to the collection?

api/src/test/java/org/openmrs/validator/PersonValidatorTest.java
```diff
@@ -231,5 +234,28 @@ public void validate_shouldNotFailWhenDeathCauseNotCodedIsSet() {
...
+		person.setPersonVoidReason("voidReason");
+		person.setGender("g");
+		person.setAddresses(addresses);
+		
+		Errors errors = new BindException(person, "patient");
+		PersonValidator personValidator = new PersonValidator();
+		
+		try {
+			personValidator.validate(person, errors);
+		}catch(NullPointerException np) {
```
dkayiwa: Are you trying to test that a null pointer exception should be thrown when there is a null in the address collection?

fruether:
I created this test case to illustrate that in case of a null address added
the program terminates normally.

This is to show that the PersonAdress validation deals with null values in
this way.

dkayiwa <notifications@github.com> schrieb am So., 21. Juli 2019, 20:05:

fruether:
Yes. That is basically related to the ticket I created yesterday.

It shows that the validation throws an exception opposed to the other
attributes that log the error instead of throwing the exception.

I basically use this test to document a behaviour that seemed to be strange
to me and at the same time proof that I did not change the program semantics

dkayiwa <notifications@github.com> schrieb am So., 21. Juli 2019, 20:06:

dkayiwa:
I do not see value in a test for a null address passed in the addresses collection. I would expect tests for real expected functionality.

fruether:
Its not for me to decide what behaviour is intended and what is not. So I
just created the test case to document the current behaviour as of proof
that refactoring did not change anything.

 For newly created ticket this test case could be used as a false positive.

I will remove it later and re-commit

dkayiwa <notifications@github.com> schrieb am So., 21. Juli 2019, 20:26:

dkayiwa:
Is your asserting in an exception handler to confirm that a null pointer exception is thrown when null is passed in the address collection?

fruether:
Let me try to be a bit more clear:
_**Test Case:**_  This test case validates that in case a PersonAddress is null a NullPointerException is thrown. This is the behaviour of the code before and after my refactoring.

_**Normal behaviour of error cases**_: For me it seems to be odd that a NullPointerException is thrown since it is neither part of the validate method description nor of the method definition. In fact, all other validations errors are dealt with by  errors.hasFieldErrors

_**History of my refactoring**_: Since I found that behaviour very odd and un-expecting I added during the refactoring process a try_catch_ block to the method to catch the NullPointerException and log that it was failing. [Where I would expect a  catch](https://github.com/openmrs/openmrs-core/pull/2953/files/627fe668ebfd9259df0b4d148df5155c574fc440#diff-7014d6c9eb6084f32ffc5c659e8ee524R129)
However, we agreed that we should not change the methods behaviour in refactoring. That is why I removed the catch block but sticked with the "Invered" test case (it is successful if the exception is thrown) to document the behaviour and proof that refactoring did not change the business logic

dkayiwa:
If your test is about expecting a null pointer exception, did you evaluate using this? @Test(expected = NullPointerException.class) or @Rule ExpectedException

fruether:
@dkayiwa I used now: @test(expected = NullPointerException.class)

dkayiwa:
Was this test passing before your refactoring changes?

fruether:
Good that you asked. Before my refactoring it was: 				ValidationUtils.invokeValidator(personAddressValidator, address, errors);

Isn't it odd that sometimes ValidationUtils is used and sometimes it is not?


That is why IllegalArgumentException was thrown not NullPointerException. However, good that I added the test to log the behaviour

dkayiwa:
The fact that your test did not fail before your refactoring changes means you are not testing the original behavior

fruether:
Yes. That is unfortunately the case. It was an error that is fixed now.

Am So., 21. Juli 2019 um 23:38 Uhr schrieb dkayiwa <notifications@github.com

dkayiwa:
It gets even more confusing when your test name says should not fail but then expect an exception.

fruether:
The test that I added now is neither failing before nor after the
refactoring. The exception was added to the correct type as well as the
method call of validate was re-stored.

Am So., 21. Juli 2019 um 23:40 Uhr schrieb Frederik Rue <
freddy92.online@gmail.com>:

dkayiwa:
That does not resolve the confusion which comes from a test whose name says should not fail but at the same time expect an exception.

fruether:
I renamed it just now to: validate_shouldThrowExceptionWhenAddressIsNull

api/src/main/java/org/openmrs/validator/PersonValidator.java
```diff
@@ -69,23 +69,58 @@ public void validate(Object target, Errors errors) {
 		
 		Person person = (Person) target;
 		
+		validatePersonNames(person, errors);
+		validatePersonAddresses(person, errors);
+		
+		if (!person.getVoided() && !validatePersonHasAtLeastOneNonVoidedName(person)) {
```
dkayiwa: Did you intentionally change the order of the validation of names and addresses?

fruether: In Line 74 personName is validated. in Line 89 personAdress in the original code. I am not sure what you mean with change of order?

dkayiwa: You are validation person addresses on line 73 before the names from line 75

fruether: Reason was that for me the code is more readable if both method calls followed each other.

dkayiwa: Changing the order may result into unintended side effects from the caller. So i would be cautious not to do that in this pull request. Am sure you have ever seen bug or unexpected behavior introduced by a mere change of the order for a few lines of code. So let us be as surgical as possible. 😊

fruether: I reverted the order

PR #2959, opened 2019-06-10
TRUNK-5557: Improve performance of ConceptService.getConceptsByMapping

mseaton:

I created a sample form that resulted in over 1000 calls to ConceptService.getConceptByMapping by the htmlformentry module.  I added some timing logging, and found that this form took around 48 seconds to load and almost all of the time was spent in this method.

After modifying the DAO to utilize Lucene, this same test returns in 8-9 seconds.  This is still unacceptably slow, but is a significant improvement and I wanted to get this approach looked at to discuss if I'm on the right track and get this merged in if so.

I am very new to Lucene, so would appreciate any constructive feedback.  One thing I am not entirely sure about is how/whether the indexes will be automatically rebuilt if the concept source / reference term were to change, and what the implications of this might be.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -500,7 +500,7 @@
 			<dependency>
 				<groupId>org.dbunit</groupId>
 				<artifactId>dbunit</artifactId>
-				<version>2.5.4</version>
+				<version>2.6.0</version>
 			</dependency>
 			<dependency>
 				<groupId>javax.validation</groupId>


rkorytkowski:
It's a nice exercise with Lucene, but I'd try fixing the existing DB query first as I noted on the issue https://issues.openmrs.org/browse/TRUNK-5557

PR #2961, opened 2019-06-13
TRUNK-5557: Improve performance of ConceptService.getConceptsByMapping

mseaton:
…g using spring cache

Based on feedback from @rkorytkowski , here is an attempt at solving the same problem using a Spring cache.  This actually does perform better than the earlier Lucene approach (and could likely be improved further in the future).

This adds one new service method to retrieve concept ids by mapping, the results of which are cached.  This is done this way as only service methods have the cache annotation picked up, and also I wanted to avoid caching any Hibernate-managed objects to avoid any lazy loading exceptions.

This has the implementation of the List<Concept> getConceptsByMapping iterate across the conceptIds returned by List<Integer> getConceptIdsByMapping, and load the matching concepts.

Performance testing shows considerable improvement.  Testing the loading of a large htmlform with a lot of concept lookups by mapping shows:

Prior to change:  47-49 seconds
After change:  3-5 seconds

@rkorytkowski curious on your feedback on this - your prior comments were very helpful, thank you!

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/ConceptDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/ConceptDAO.java
@@ -324,6 +324,11 @@ public interface ConceptDAO {
 	 * @see org.openmrs.api.ConceptService#getConceptsByMapping(java.lang.String, java.lang.String)
 	 */
 	public List<Concept> getConceptsByMapping(String code, String sourceName, boolean includeRetired);
+
+	/**
+	 * @see org.openmrs.api.ConceptService#getConceptIdsByMapping(String, String, boolean)
+	 */
+	public List<Integer> getConceptIdsByMapping(String code, String sourceName, boolean includeRetired);
 	
 	/**
 	 * @param uuid
--- a/api/src/main/java/org/openmrs/api/db/ConceptDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/ConceptDAO.java
@@ -322,7 +322,10 @@ public interface ConceptDAO {
 	
 	/**
 	 * @see org.openmrs.api.ConceptService#getConceptsByMapping(java.lang.String, java.lang.String)
+	 * This method has been deprecated in favor of getConceptIdsByMapping, which
+	 * is used by the ConceptService to leverage caching of results
 	 */
+	@Deprecated
 	public List<Concept> getConceptsByMapping(String code, String sourceName, boolean includeRetired);
 
 	/**
--- a/api/src/main/java/org/openmrs/api/db/ConceptDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/ConceptDAO.java
@@ -322,8 +322,8 @@ public interface ConceptDAO {
 	
 	/**
 	 * @see org.openmrs.api.ConceptService#getConceptsByMapping(java.lang.String, java.lang.String)
-	 * This method has been deprecated in favor of getConceptIdsByMapping, which
-	 * is used by the ConceptService to leverage caching of results
+	 * 
+	 * @deprecated As of 2.5.0, this method has been deprecated in favor of {@link #getConceptIdsByMapping(String, String, boolean)}
 	 */
 	@Deprecated
 	public List<Concept> getConceptsByMapping(String code, String sourceName, boolean includeRetired);
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
@@ -471,7 +471,8 @@ public class HibernateConceptDAO implements ConceptDAO {
 	}
 	
 	/**
-	 * @see org.openmrs.api.db.ConceptDAO#getConceptDatatypes(java.lang.String)
+	 * @param name the name of the ConceptDatatype
+	 * @return a List of ConceptDatatype whose names start with the passed name
 	 */
 	@SuppressWarnings("unchecked")
 	public List<ConceptDatatype> getConceptDatatypes(String name) throws DAOException {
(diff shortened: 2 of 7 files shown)


rkorytkowski:
Good point about caching ids only.

Maybe worth to consider adding evict to https://github.com/openmrs/openmrs-core/blob/4a0feb8da351088f25fdc4e6d324a1f277aa3410/api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java#L1473 This method may be used from the admin UI after loading e.g. an update to concept dictionary via SQL to rebuild indexes without having to restart the server.

Well done!

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
```diff
@@ -1016,54 +1016,29 @@ public void remove() {
...
 	/**
 	 * @see org.openmrs.api.db.ConceptDAO#getConceptsByMapping(String, String, boolean)
 	 */
 	@Override
 	@SuppressWarnings("unchecked")
 	public List<Concept> getConceptsByMapping(String code, String sourceName, boolean includeRetired) {
-		Criteria criteria = sessionFactory.getCurrentSession().createCriteria(ConceptMap.class);
-		
-		// make this criteria return a list of concepts
+		Criteria criteria = createSearchConceptMapCriteria(code, sourceName, includeRetired);
```
rkorytkowski: I'd make use of cache here as well with:
```
List<Integer> ids = Context.getConceptService().getConceptIdsByMapping(code, sourceName, includeRetired); 
session.createCriteria(Concept.class).add(Restrictions.in("id",ids)).list(); 
if (includeRetired) {
			// sort retired concepts to the end of the list
			criteria.addOrder(Order.asc("retired"));
}
```

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
```diff
@@ -2068,4 +2043,43 @@ private Criteria createSearchDrugByMappingCriteria(String code, ConceptSource co
...
+		} else {
+			criteria.add(Restrictions.or(Restrictions.eq("source.name", sourceName), Restrictions.eq("source.hl7Code",
+					sourceName)));
+		}
+
+		criteria.createAlias("concept", "concept");
+
+		if (!includeRetired) {
+			// ignore retired concepts
+			criteria.add(Restrictions.eq("concept.retired", false));
```
rkorytkowski: Note that cache should be evicted also when concept.retired changes.

mogoodrich:
Thanks @mseaton this is great!

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
mseaton: I tried this out, and it was a lot slower (2x) than the way the service method is implemented.  I also am not a fan of services calling DAOs which then call different services.  I actually think we should leave this dao method implementation alone but Deprecate the method, as it is no longer used by the Service layer at all.

rkorytkowski: Interesting. Fair point to not call services in DAOs.

mseaton: I assume this is due to the fact that by iterating over the concept ids and calling getConcept(id), we also leverage hibernate's use of ehcache in a more efficient manner than by running an HQL query with an "in" clause on ids.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
mseaton: I just pushed a follow-up commit that:
* Deprecates the getConceptsByMapping DAO method in favor of future use of getConceptIdsByMapping
* Evicts from cache more liberally on saveConcept, purge/save/retire ConceptSource, updateConceptIndexes
Thanks for the support and feedback @rkorytkowski - much appreciated.

api/src/main/java/org/openmrs/api/db/ConceptDAO.java
```diff
@@ -322,8 +322,16 @@ public Integer getCountOfConcepts(String phrase, List<Locale> locales, boolean i
 	
 	/**
 	 * @see org.openmrs.api.ConceptService#getConceptsByMapping(java.lang.String, java.lang.String)
+	 * This method has been deprecated in favor of getConceptIdsByMapping, which
```
dkayiwa: I would follow this convention: https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-Deprecation

api/src/main/java/org/openmrs/api/db/ConceptDAO.java
```diff
@@ -322,8 +322,16 @@ public Integer getCountOfConcepts(String phrase, List<Locale> locales, boolean i
...
+	 * is used by the ConceptService to leverage caching of results
 	 */
+	@Deprecated
 	public List<Concept> getConceptsByMapping(String code, String sourceName, boolean includeRetired);
+
+	/**
+	 * @see org.openmrs.api.ConceptService#getConceptIdsByMapping(String, String, boolean)
+	 */
```
dkayiwa: I would also add a since annotation for the version when this method was introduced.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
dkayiwa: Am in favour of leaving this dao method implementation alone.

api/src/main/java/org/openmrs/api/db/ConceptDAO.java
dkayiwa: @mseaton are you still interested in the work you did in this pull request? 😊

gitcliff:
@mseaton some files are conflicting ..
Thanks

mseaton:
I just noticed this ticket still assigned to me and still in the review state.  Since there are no PR approvals, I'll leave it alone for now, but re-pinging @dkayiwa and @rkorytkowski and including @ibacher in case we can get this resolved one way or another.

dkayiwa:
Hopefully you are looking into the merge conflicts. 😊

ibacher (approved):

ibacher:
@mseaton Thanks for this... sorry it's taken so long to actually merge this!

PR #2965, opened 2019-06-16
TRUNK-5494 Replace Obs hbm mapping file with annotations

wluyima:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5494

  No? -> execute above command


api/src/main/java/org/openmrs/Obs.java
```diff
@@ -106,55 +133,91 @@
...
+	@Column(name = "value_complex")
 	protected String valueComplex;
 	
 	// ComplexData is not persisted in the database.
 	protected transient ComplexData complexData;
 	
+	@Column(name = "comments")
 	protected String comment;
 	
-	protected transient Integer personId;
```
wluyima: Having transient on this field as before was wrong because this is a persistent field.

wluyima:
There is a failing test I need to fix.

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -106,55 +133,91 @@
...
 	
+	@JoinColumn(name = "previous_version", unique = true)
+	@ManyToOne
 	private Obs previousVersion;
 	
+	@Column(name = "form_namespace_and_path")
+	@Access(AccessType.FIELD)
 	private String formNamespaceAndPath;
 	
-	private Boolean dirty = Boolean.FALSE;
```
HerbertYiga: so we are having this to be persisted after these changes?

mozzy11: @reagan-meant , can you squash the commits into one ?

reagan-meant: Currently it's @wluyima working on this issue

wluyima: Since @reagan-meant did some work on this, I think it's okay not to squash the commits to keep his commits in the history

dkayiwa:
I got this error while registering a new patient with these changes in the reference application: https://pastebin.com/hAg6gTE2

wluyima:
Well, that's exactly what I have been saying in https://talk.openmrs.org/t/issue-with-using-annotations-to-map-obs/23684/7

api/src/main/java/org/openmrs/Obs.java
dkayiwa: How was it able to persist values into the database when it was transient?

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -106,55 +140,92 @@
...
+	@JoinColumn(name = "encounter_id")
+	@ManyToOne
 	protected Encounter encounter;
 	
+	@JoinColumn(name = "previous_version", unique = true)
+	@ManyToOne
 	private Obs previousVersion;
 	
+	@Column(name = "form_namespace_and_path")
+	@Access(AccessType.FIELD)
```
dkayiwa: Could you remind me why we explicitly set the access type for this?

api/src/main/java/org/openmrs/Obs.java
wluyima: Recall that we were using xml before, the field was mapped in the hbm file and that's why.

api/src/main/java/org/openmrs/Obs.java
wluyima: I was just replicating the exact mapping in the hbm file, it was already accessed by field in xml file.

As to why it was historically mapped this way, if I recall well, it was because the class exposes namespace and formfield path as the public properties but both are stored in one column as concatenation separated by some character, therefore formNamespaceAndPath is internal logic.

api/src/main/java/org/openmrs/Obs.java
dkayiwa: This property duplicates the person property for for speed and simplicity reasons as you can see from the javadoc of the setter and getter methods. And this was done for serialisation purposes. That is why i feel like we should not remove it using this pull request.

wluyima: What am saying is that the field has always been persistent, meaning the transient keyword was redundant, you only add a transient keyword to none persistent fields. If you have an OpenMRS instance you will realize that it is written to the DB.

wluyima: See https://github.com/openmrs/openmrs-core/blob/1.11.x/api/src/main/resources/org/openmrs/api/db/hibernate/Obs.hbm.xml#L31

wluyima: I understand we added it for serialization purposes but persistence frameworks use the same keyword to determine which fields to persist vs not persist, if we don't remove it, the field is excluded from insert statements yet it is required, so you can't save an Obs to the DB.

If we don't want to serialize it, we would have to just update the serializer to explicitly exclude it.

dkayiwa: Does this mean that if we leave that transient keyword then we cannot save an Obs?

wluyima: Yes, we can't because just like a serializer, JPA frameworks also ignore the field in the insert statement yet it is not nullable in the DB. The reason why it was working before is because when using xml, hibernate builds the class metadata from the hbm file so it never discovers the transient keyword whereas with annotations it inspects the class' fields using reflection and ends up seeing the transient keyword.

PR #2977, opened 2019-07-22
GCI-241:Lazy initialization of static serviceContext should be synchr…

jwnasambu:
I manged to synchronize the methods.

## Issue I worked on
https://issues.openmrs.org/browse/GCI-241

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Context.java
+++ b/api/src/main/java/org/openmrs/api/context/Context.java
@@ -239,7 +239,7 @@ public class Context {
 	 *
 	 * @return the current ServiceContext
 	 */
-	static ServiceContext getServiceContext() {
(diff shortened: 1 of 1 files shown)


dkayiwa:
Tickets status shows you are still doing more work on this.

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -239,7 +239,7 @@ public static UserContext getUserContext() {
 	 *
 	 * @return the current ServiceContext
 	 */
-	static ServiceContext getServiceContext() {
+	static synchronized ServiceContext getServiceContext() {
```
dkayiwa: I can see that the method body was already synchronised. Can you instead remove the outer if (serviceContext == null)? Because i see it duplicated in the inner block.

PR #2997, opened 2019-09-03
Trunk 5438

cioan:
Hi @mogoodrich , thank you for reviewing these changes!

### DIFF
--- a/api/src/main/java/org/openmrs/Encounter.java
+++ b/api/src/main/java/org/openmrs/Encounter.java
@@ -264,6 +264,21 @@ public class Encounter extends BaseChangeableOpenmrsData {
 	public Set<Obs> getAllObs() {
 		return getAllObs(false);
 	}
+
+	public Set<Obs> getAllFlattenObs(boolean includeVoided) {
+
+		Set<Obs> ret = new LinkedHashSet<>();
+
+		if (this.obs != null) {
+			for (Obs o : this.obs) {
+				if(!o.getVoided()) {
+					ret.add(o);
+					ret.addAll(getObsLeaves(o));
+				}
+			}
+		}
+		return ret;
+	}
 	
 	/**
 	 * Returns a Set&lt;Obs&gt; of all root-level Obs of an Encounter, including obsGroups
--- a/api/src/main/java/org/openmrs/Encounter.java
+++ b/api/src/main/java/org/openmrs/Encounter.java
@@ -229,7 +229,28 @@ public class Encounter extends BaseChangeableOpenmrsData {
 		
 		return leaves;
 	}
-	
+
+	/**
+	 * Convenience method to recursively get all leaf obs for a given parent obs.
+	 * @param obsParent indicates the parent Obs
+	 * @param includedVoided indicates whether or not to include voided Obs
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/openmrs/Encounter.java
```diff
@@ -264,6 +285,26 @@ public void setLocation(Location location) {
...
+	 * @return a Set of all encounter' Obs
+	 */
+	public Set<Obs> getAllFlattenObs(boolean includeVoided) {
+
+		Set<Obs> ret = new LinkedHashSet<>();
+
+		if (this.obs != null) {
+			for (Obs o : this.obs) {
+				if (includeVoided || (!includeVoided && !o.getVoided())) {
+					ret.add(o);
```
mogoodrich: Per my comments above, I think we may want to add the parent obs in the above method instead of here, to make sure we get nested parents?  Maybe the above method would be called "getFlattenedObsHelper" since it would no longer be dealing just with leaves?

api/src/main/java/org/openmrs/Encounter.java
```diff
@@ -229,7 +229,28 @@ public void setLocation(Location location) {
...
+	/**
+	 * Convenience method to recursively get all leaf obs for a given parent obs.
+	 * @param obsParent indicates the parent Obs
+	 * @param includedVoided indicates whether or not to include voided Obs
+	 * @return a Set of all leaves Obs including the parent Obs
+	 */
+	private Set<Obs> getFlattenObsLeaves(Obs obsParent, boolean includedVoided) {
+		Set<Obs> leaves = new LinkedHashSet<>();
+
+		if (includedVoided || (!includedVoided && !obsParent.getVoided())) {
```
dilanthas: Just a thought, i think checking !includedVoided is not necessary . If its coming to the right operand that means its already false.

api/src/main/java/org/openmrs/Encounter.java
cioan: Thank you @mogoodrich . Those were great catches and suggestions. I just pushed an update that addresses all your points above. Thanks!

api/src/main/java/org/openmrs/Encounter.java
cioan: Thanks @dilanthas. That's a subtle and elegant improvement.

api/src/main/java/org/openmrs/Encounter.java
```diff
@@ -264,6 +284,25 @@ public void setLocation(Location location) {
 	public Set<Obs> getAllObs() {
 		return getAllObs(false);
 	}
+
+	/**
+	 * Returns all encounter obs as a flatten list(Set) of obs.
+	 * @param includeVoided indicates whether or not to include voided obs
+	 * @return a Set of all encounter' Obs
+	 */
```
dkayiwa: Can we add the since annotation? https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-NewPublicMethodsandClasses

mogoodrich: cool call @dkayiwa fyi @cioan

PR #2999, opened 2019-09-08
TRUNK-5188: create setters and getters for closedLow and closedHigh

ODORA0:
**issue worked on:** https://issues.openmrs.org/browse/TRUNK-5188

### DIFF
--- a/api/src/main/java/org/openmrs/util/DoubleRange.java
+++ b/api/src/main/java/org/openmrs/util/DoubleRange.java
@@ -23,7 +23,7 @@ public class DoubleRange implements Comparable<DoubleRange> {
 	
 	private Double high;
 	
-	private boolean closedLow = true; //TODO: add setters and getters for these
+	private boolean closedLow = true;
 	
 	private boolean closedHigh = false;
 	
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/util/DoubleRange.java
```diff
@@ -77,6 +77,39 @@ public void setLow(Double low) {
...
+	public boolean getclosedLow () {
+		return closedLow;
+	}
+	
+	/**
+	 * @param closedLow to set
+	 * 
+	 */
+	public void closedLow(boolean closedLow) {
+		this.closedLow = closedLow == true;
```
dilanthas: simply having `this.closedLow = closedLow` would be enough

api/src/main/java/org/openmrs/util/DoubleRange.java
```diff
@@ -77,6 +77,39 @@ public void setLow(Double low) {
...
+	public boolean getclosedHigh () {
+		return closedHigh;
+	}
+	
+	/**
+	 * @param closedHigh to set
+	 * 
+	 */
+	public void closedHigh(boolean closedHigh) {
+		this.closedHigh = closedHigh == false;
```
dilanthas: This seems wrong. when the method parameter closedHigh is false you will be assigning true to the this.closedHigh. Is that correct _

api/src/main/java/org/openmrs/util/DoubleRange.java
ODORA0: That's ok i can change this

PR #3006, opened 2019-09-14
TRUNK-381 : Follow up commit to fix bug due to context refresh to suc…

Ruhanga:
Ticket : https://issues.openmrs.org/browse/TRUNK-381

Pass the ContextDAO instance on than save it in an authenticationScheme instance which has been causing context refresh vs existing browser session issue, evident when logging out and then in without clean the browser session.

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Context.java
+++ b/api/src/main/java/org/openmrs/api/context/Context.java
@@ -202,10 +202,6 @@ public class Context {
 		catch(BeansException e){
 			log.error("Fatal error encountered when injecting the authentication scheme override. Sticking to OpenMRS default authentication scheme.");
 		}
-
-		if (authenticationScheme instanceof DaoAuthenticationScheme) {
-			((DaoAuthenticationScheme) authenticationScheme).setContextDao(getContextDAO());
-		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


Ruhanga:
The old `ContextDAO` is probably not invalidated by spring at the time the new one is created on context refresh. Maybe because it's being referenced and used with the old `authenticationScheme` instance. This leaves the old `ContextDAO` with the hibernate session open on some thread. The new `ContextDAO` can not access the hibernate session that's open in the old `ContextDAO` hence [this error](https://pastebin.com/yd6D23cN).

mogoodrich:
@Ruhanga  @mks-d when does this error occur?  (We just did a release and unfortunately it includes this update, so I was wondering how prevalent it is... I can log in and out without issue with the PIH EMR... 

fwiw, this is our authentication methods:

https://github.com/PIH/openmrs-module-mirebalais/commit/6aef1edd5956c4596e7ed00c0a97b5979a667a0f

Ruhanga:
@mogoodrich, you'd experience the error if you were using an OWA that leverages rest. Just after refreshing context, which could be triggered by probably uploading or uninstalling/deleting a module from the web interface, logging out then logging in again reproduces the error.

mogoodrich:
Thanks @Ruhanga ... we do use OWAs that leverage REST, but we rarely/never add or remove modules without restarting the server... I don't think there are cases anymore where we do a context refresh outside of a full server restart, so I think we are okay... ?

mks-d:
Also @Ruhanga to confirm but the problem goes away after a while anyway, and this would only show up if you were to logout through the OWA (so using REST).
But looks like you're safe if you don't add/remove modules through the UI (which is the case for most prod. deployments).

Btw we might have a much lighter fix than the one suggested here, but a fix is required nonetheless...

PR #3022, opened 2019-09-30
Trunk-5661:Replace Allergy hbm mapping file with annotations

gitcliff:
link:https://issues.openmrs.org/browse/TRUNK-5661
Replace Allergy hbm mapping file with annotations


HerbertYiga:
did you remove the mappings any were, i can only see the introduced  annotations in your code!

HerbertYiga:
did you remove the mappings any were, i can only see the introduced  annotations in your code!

gitcliff:
no i haven't removed the mappings

api/src/main/java/org/openmrs/Allergen.java
```diff
@@ -9,17 +9,26 @@
...
+
 import org.apache.commons.lang3.StringUtils;
 
 /**
  * Represent allergen
  */
+@Embeddable
 public class Allergen {
 	
 	private AllergenType allergenType;
```
wluyima: You forgot to map this

api/src/main/java/org/openmrs/Allergy.java
```diff
@@ -15,26 +15,54 @@
...
  * Represent allergy
  */
+@Entity
+@Table(name = "allergy")
 public class Allergy extends BaseChangeableOpenmrsData {
 	
 	public static final long serialVersionUID = 1;
 	
+	@Id
+	@GeneratedValue(strategy = GenerationType.AUTO)
```
wluyima: AUTO is the default, so you can omit the strategy attribute

api/src/main/java/org/openmrs/Allergy.java
```diff
@@ -15,26 +15,54 @@
...
 	private Patient patient;
 	
+	@Embedded
 	private Allergen allergen;
 	
+	@ManyToOne
+	@JoinColumn(name = "severity_concept_id")
 	private Concept severity;
 	
+	@Column(name = "comments")
```
wluyima: You need to set the column length

wluyima: See Allergy.hbm.xml

api/src/main/java/org/openmrs/Allergy.java
```diff
@@ -15,26 +15,54 @@
...
 	private String comments;
 	
+	@ElementCollection
+	@OneToMany(mappedBy = "allergy", cascade = CascadeType.ALL)
+	@OrderBy("voided asc")
 	private List<AllergyReaction> reactions = new ArrayList<>();
```
wluyima: Make sure these match the mapping values from the hbm file

wluyima:
- Please make sure the annotation mappings match those in the hbm file
- Make sure all fields are mapped
- Delete the allergy hbm file
- Update hibernate.cfg.xml to remove the hbm file for allergy

Remember to run unit tests before you push your changes.

api/src/main/java/org/openmrs/Allergen.java
gitcliff: i have fixed it

api/src/main/java/org/openmrs/Allergy.java
gitcliff: i have fixed it

api/src/main/java/org/openmrs/Allergy.java
gitcliff: i have fixed it

gitcliff:
@wluyima i have responded to your comments,,kindly review

api/src/main/java/org/openmrs/Allergy.java
```diff
@@ -15,26 +15,60 @@
...
+	@JoinColumn(name = "severity_concept_id")
 	private Concept severity;
 	
+	@Column(name = "comments", length = 1024)
 	private String comments;
 	
+
+	@OneToMany(mappedBy = "allergy", cascade = CascadeType.ALL, orphanRemoval = true)
+	@LazyCollection(LazyCollectionOption.FALSE)
+	@OrderBy("allergy_id")
```
wluyima: There was no ordering set in the hbm file

wluyima: And I think by default they are ordered by allergy_id anyways

api/src/main/java/org/openmrs/Allergen.java
```diff
@@ -9,17 +9,30 @@
...
  * Represent allergen
  */
+@Embeddable
 public class Allergen {
 	
+	@Enumerated(EnumType.STRING)
+	@Column(name = "allergen_type")
 	private AllergenType allergenType;
 	
+	@ManyToOne
```
wluyima: Keep this as not nullable like in the hbm file

api/src/main/java/org/openmrs/Allergy.java
```diff
@@ -15,26 +15,60 @@
...
+	@GeneratedValue
+	@Column(name = "allergy_id")
 	private Integer allergyId;
 	
+	@ManyToOne(optional = false)
+	@JoinColumn(name = "patient_id")
 	private Patient patient;
 	
+	@Embedded
+	@AttributeOverrides({ @AttributeOverride(name = "allergenType", column = @Column(name = "allergen_type")),
```
wluyima: These are redundant since you are not actually overriding the column names

gitcliff: @wluyima should i remove them?

gitcliff: @wluyima coz am seeing according to the hbm file the allergenType is part of the allergen component...am abit confused on this

wluyima: I'm saying only remove the @AtttributeOverrides annotation, keep @Embedded only

api/src/main/java/org/openmrs/Allergen.java
gitcliff: @wluyima i have fixed it

wluyima: I prefer to set optional = false for the @ManyToOne rather than nullable on JoinColumn because then the constraint can be enforced on the object model itself rather than only in the db

api/src/main/java/org/openmrs/Allergy.java
wluyima: Did you see this comment?

gitcliff: yes @wluyima  i have seen the comment n fixed...

api/src/main/java/org/openmrs/Allergen.java
gitcliff: okay @wluyima true...

i have fixed it

PR #3032, opened 2019-10-20
Trunk-5225:Get list of OrderGroups by patient or encounter is missing…

gitcliff:
link:https://issues.openmrs.org/browse/TRUNK-5225
Get List of OrderGroups by patient or encounter is missing in openmrs-core.

### DIFF
--- a/api/src/main/java/org/openmrs/api/OrderService.java
+++ b/api/src/main/java/org/openmrs/api/OrderService.java
@@ -812,4 +812,26 @@ public interface OrderService extends OpenmrsService {
 	 */
 	@Authorized({ PrivilegeConstants.EDIT_ORDERS, PrivilegeConstants.ADD_ORDERS })
 	public OrderGroup saveOrderGroup(OrderGroup orderGroup) throws APIException;
+	
+	/**
+	 * Fetches all order groups for the specified patient
+	 * 
+	 * @param patient the patient to match on
+	 * @return list of matching OrderGroups
+	 * @since 2.3
+	 * @throws APIException
+	 */
+	@Authorized(PrivilegeConstants.GET_ORDERS)
+	public List<OrderGroup> getOrderGroupsByPatient(Patient patient) throws APIException;
+	
+	/**
+	 * Fetches all order groups for the specified encounter
+	 *
+	 * @param encounter the encounter to match on
+	 * @return list of matching OrderGroups
+	 * @since 2.3
+	 * @throws APIException
+	 */
+	@Authorized(PrivilegeConstants.GET_ORDERS)
+	public List<OrderGroup> getOrderGroupsByEncounter(Encounter encounter) throws APIException;
 }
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/openmrs/api/OrderService.java
```diff
@@ -812,4 +812,26 @@ public Order discontinueOrder(Order orderToDiscontinue, String reasonNonCoded, D
 	 */
 	@Authorized({ PrivilegeConstants.EDIT_ORDERS, PrivilegeConstants.ADD_ORDERS })
 	public OrderGroup saveOrderGroup(OrderGroup orderGroup) throws APIException;
+	
+	/**
+	 * Fetches all order groups for the specified patient
+	 * 
+	 * @param patient the patient to match on
+	 * @return list of matching OrderGroup * @since 2.3
```
dkayiwa: Why are you putting the since annotation on the same line as the return?

api/src/main/java/org/openmrs/api/OrderService.java
```diff
@@ -812,4 +812,26 @@ public Order discontinueOrder(Order orderToDiscontinue, String reasonNonCoded, D
...
+	 */
+	@Authorized(PrivilegeConstants.GET_ORDERS)
+	public List<OrderGroup> getOrderGroupsByPatient(Patient patient) throws APIException;
+	
+	/**
+	 * Fetches all order groups for the specified encounter
+	 *
+	 * @param encounter the encounter to match on
+	 * @return list of matching OrderGroups
+	 * @since 2.3
```
dkayiwa: Are you sure version 2.3 will have these changes?

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -484,6 +484,7 @@ public Order unvoidOrder(Order order) throws APIException {
 		return saveOrderInternal(order, null);
 	}
 	
+	@Override
```
dkayiwa: Is this change part of the ticket?

api/src/main/java/org/openmrs/api/OrderService.java
gitcliff: @dkayiwa i think so because in the ticket description the fix version is 2.3.0

api/src/main/java/org/openmrs/api/OrderService.java
gitcliff: i have fixed it

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
gitcliff: @dkayiwa this change keeps coming when i make changes in the OrderServiceImpl class,tho i didn't make any change on this line of code..i have run 'git reset --hard upstream/master' to revert the changes but on re-building again am still getting the same issue here

api/src/main/java/org/openmrs/api/OrderService.java
dkayiwa: Looks like that version is being released without this fix. Look at the current version in master pom.xml

PR #3034, opened 2019-10-21
TRUNK-5672 | Add liquibase changesets to populate username for admin user

mddubey:
As part of this PR following things are done:
- [x] Add a new file `liquibase-update-to-2.4.xml`.
- [x] Included the `liquibase-update-to-2.4.xml` in `liquibase-update-to-latest.xml` file.
- [x] Add the changeset to `liquibase-update-to-2.4.xml`.
- [x] Add the changeset to `liquibase-update-to-2.3.xml`.

Jira ticket: https://issues.openmrs.org/browse/TRUNK-5672
Talk Thread: https://talk.openmrs.org/t/admin-user-doesnt-have-a-username/25145/

### DIFF
--- a/api/src/main/resources/liquibase-update-to-2.3.xml
+++ b/api/src/main/resources/liquibase-update-to-2.3.xml
@@ -108,4 +108,16 @@
 			<column name="description" type="varchar(1)" />
 		</createTable>
 	</changeSet>
+	<changeSet id="TRUNK-5672-201910211105" author="Mritunjay">
+		<preConditions onFail="MARK_RAN">
+			<sqlCheck expectedResult="0">
+				SELECT COUNT(*) FROM users where username = 'admin'
+			</sqlCheck>
+		</preConditions>
+		<comment>Populate username for admin user</comment>
+		<update tableName="users">
+			<column name="username" value="admin"/>
+			<where>system_id='admin'</where>
+		</update>
+	</changeSet>
 </databaseChangeLog>


api/src/main/resources/liquibase-update-to-2.3.xml
```diff
@@ -108,4 +108,16 @@
 			<column name="description" type="varchar(1)" />
 		</createTable>
 	</changeSet>
+	<changeSet id="TRUNK-5672-201910211105" author="Mritunjay">
```
dkayiwa: We do not put the changeset in both the 2.3 and 2.4 files. Just one of the two.

api/src/main/resources/liquibase-update-to-2.3.xml
```diff
@@ -108,4 +108,16 @@
...
+		<preConditions onFail="MARK_RAN">
+			<sqlCheck expectedResult="0">
+				SELECT COUNT(*) FROM users where username = 'admin'
```
dkayiwa: Wouldn't it better to search by uuid?

api/src/main/resources/liquibase-update-to-2.3.xml
mddubey: Sure, it makes sense! I just got confused. I will remove `2.4` and just add it to `2.3`  file

api/src/main/resources/liquibase-update-to-2.3.xml
mddubey: In the `pre-condition`, I wanted to say that if there is already a user with username `admin` I would not run this changeset and mark it as RAN.
If I will put the pre-condition using `uuid` the count will always come as true. In that case, I will never execute the `changeset`.

mddubey: The other part is I am not really sure, where are we populating `uuid` to `admin` user. In [this changeset](https://github.com/openmrs/openmrs-core/blob/a7600d359821debe312d1d14b49cd4714481d7c0/api/src/main/resources/liquibase-core-data.xml#L52), we are not populating any `uuid` at all. 

My worry would be if it is generated as `uuid()` function call, we will get different `UUIDs` on different instances.

I see we generate a hardcoded UUID for `daemon` user in [this changeset](https://github.com/openmrs/openmrs-core/blob/a7600d359821debe312d1d14b49cd4714481d7c0/api/src/main/resources/liquibase-core-data.xml#L1078), but not for `admin` user.

api/src/main/resources/liquibase-update-to-2.3.xml
mddubey: Fixed this.

api/src/main/resources/liquibase-update-to-2.3.xml
dkayiwa: Have you tried to test this change by running the compiled web application with it?

mddubey: A question on that, I have been trying to run it my local. This is what I tried
- `mvn clean install` 
- Copy the generated `openmrs.war` to docker
- Restart the container

For some reason, it is not running the changeset. I couldn't see it in the `liquibasechangelog` table.

Am I missing some step?

dkayiwa: Can you change to the `webapp` directory and then run this command? `mvn jetty:run`

mddubey: Thanks, I tried this and I see username for `admin` user is populated.
The query:
```
select system_id, username from users;
```
Gives result:
```
admin, admin
daemon, daemon
```

And the query:
```
select ID,DATEEXECUTED,Comments from liquibasechangelog where author = 'Mritunjay';
```
Gives result:
```
'TRUNK-5672-201910211105','2019-10-23 08:24:55','Populate username for admin user'
```

PR #3040, opened 2019-10-30
TRUNK-5675  Lucene Analyzer Defs should be provided programatically

brandones:
TRUNK-5675 Lucene Analyzer Defs should be provided programatically
https://issues.openmrs.org/browse/TRUNK-5675

Moves the Lucene Analyzer Definitions out of BaseOpenmrsObject annotations and into a factory class.

- [ ] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.  (maven does auto-formatting, we should just get that working if it isn't presently...)
- [ ] I have **added tests** to cover my changes. (this is a refactor)

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactory.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactory.java
@@ -1,0 +1,59 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.db.hibernate.search;
+
+import org.apache.lucene.analysis.core.KeywordTokenizerFactory;
+import org.apache.lucene.analysis.core.LowerCaseFilterFactory;
+import org.apache.lucene.analysis.core.WhitespaceTokenizerFactory;
+import org.apache.lucene.analysis.ngram.EdgeNGramFilterFactory;
+import org.apache.lucene.analysis.ngram.NGramFilterFactory;
+import org.apache.lucene.analysis.standard.ClassicFilterFactory;
+import org.hibernate.search.annotations.Factory;
+import org.hibernate.search.cfg.SearchMapping;
+
+/**
+ * Provides a Lucene SearchMapping for any objects in openmrs-core.
+ * 
+ * Objects such as PersonName can use the analyzers provided by this mapping to make their fields searchable.
+ * This class defines some default analyzers:
+ * 	phraseAnalyzer, which allows searching for an entire phrase, including whitespace
+ * 	startAnalyzer, which allows searching for tokens that match at the beginning
+ * 	exactAnalyzer, which allows searching for tokens that are identical
+ * 	anywhereAnalyzer, which allows searching for text within tokens
+ *
+ * @since 2.4.0
+ */
+public class LuceneAnalyzerFactory {
+	@Factory
+	public SearchMapping getSearchMapping() {
+		SearchMapping mapping = new SearchMapping();
+		mapping
+			.analyzerDef(LuceneAnalyzers.PHRASE_ANALYZER, KeywordTokenizerFactory.class)
+			.filter(ClassicFilterFactory.class)
+			.filter(LowerCaseFilterFactory.class);
+		mapping.analyzerDef(LuceneAnalyzers.EXACT_ANALYZER, WhitespaceTokenizerFactory.class)
+			.filter(ClassicFilterFactory.class)
+			.filter(LowerCaseFilterFactory.class);
+		mapping.analyzerDef(LuceneAnalyzers.START_ANALYZER, WhitespaceTokenizerFactory.class)
+			.filter(ClassicFilterFactory.class)
+			.filter(LowerCaseFilterFactory.class)
+			.filter(EdgeNGramFilterFactory.class)
+                .param("minGramSize", "2")
+                .param("maxGramSize", "20");
+		mapping.analyzerDef(LuceneAnalyzers.ANYWHERE_ANALYZER, WhitespaceTokenizerFactory.class)
+			.filter(ClassicFilterFactory.class)
+			.filter(LowerCaseFilterFactory.class)
+			.filter(NGramFilterFactory.class)
+			.param("minGramSize", "2")
+			.param("maxGramSize", "20");
+		return mapping;
+	}
+}
+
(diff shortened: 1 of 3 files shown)


gitcliff:
@brandones  there seems to be a travis ci failure

HerbertYiga:
do you have  org.apache.lucene:lucene-analyzers-phonetic  dependencies in your pom? this is what is causing breaks according to your logs,it looks to be missing

pom.xml
```diff
@@ -275,6 +275,11 @@
 				<artifactId>lucene-queries</artifactId>
 				<version>4.10.4</version>
 			</dependency>
+			<dependency>
+				<groupId>org.apache.lucene</groupId>
+				<artifactId>lucene-analyzers-phonetic</artifactId>
+				<version>4.10.4</version>
+			</dependency>
```
mogoodrich: Are we just adding these here so they are available and users that want to use phonetic mappings don't have to add them to their own modules?  I think we could just add this in PIH Core?

api/src/main/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactory.java
```diff
@@ -0,0 +1,47 @@
...
+			.filter(EdgeNGramFilterFactory.class)
+                .param("minGramSize", "2")
+                .param("maxGramSize", "20");
+		mapping.analyzerDef(LuceneAnalyzers.ANYWHERE_ANALYZER, WhitespaceTokenizerFactory.class)
+			.filter(ClassicFilterFactory.class)
+			.filter(LowerCaseFilterFactory.class)
+			.filter(NGramFilterFactory.class)
+			.param("minGramSize", "2")
+			.param("maxGramSize", "20");
+		return mapping;
```
mogoodrich: This looks right... I wonder if can come up with a good strategy for testing that things are still working correctly?

mogoodrich: Forgive my ignorance of Lucene, but is the idea that we define the analyzers here and then the annotations on the actual domain objects (like PersonName) reference these analyzers?  Can you mix these two ways of configuring?  (ie setting up the analyzers programmatically but then referencing them via annotations?)

mogoodrich:
Also, for reference, here's a PR for overriding the analyzers from a module (though this isn't currently working).

pom.xml
brandones: That's what I thought, and then I removed it, and the build broke. I don't know exactly what it's providing.

api/src/main/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactory.java
brandones: Good question! Yes, PersonName is referencing the analyzers using field annotations: see [PersonName:58](https://github.com/openmrs/openmrs-core/blob/640cd7dd47f65c02574458f4b25d8b114aa817ed/api/src/main/java/org/openmrs/PersonName.java#L58). I have read that you can't define analyzers both with annotations and programmatically, though.

brandones:
@mseaton Good call about tests, I [filed a Jira ticket](https://issues.openmrs.org/browse/TRUNK-5679). I'll open a draft PR soon to get feedback on my approach to testing this.

brandones:
Started writing tests, let me know if you think I'm on the wrong track: https://github.com/openmrs/openmrs-core/pull/3044/files#diff-533922d1dc5d42385e95b23751b773e7R32-R85

api/src/main/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactory.java
mogoodrich: So we are saying here "according to the docs, this shouldn't work, but it looks like it actually is?".  I'd be a little concerned with that which could hopefully be mitigated by adding some tests as @mseaton mentions.

pom.xml
mogoodrich: That's bit of a red flag... :)

api/src/main/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactory.java
brandones: No, I think the distinction between defining and using analyzers is important here.

> Today, the programmatic API cannot be used on a class annotated with Hibernate Search annotations, chose one approach or the other.

We're only using the programmatic API to define analyzers, we're not "using it on a class." Classes are continuing to use analyzers via annotations (`@Field`).

mogoodrich: Great, thanks for the clarification @brandones

brandones:
Found existing test coverage for Lucene configuration [in PersonServiceTest](https://github.com/openmrs/openmrs-core/blob/640cd7dd47f65c02574458f4b25d8b114aa817ed/api/src/test/java/org/openmrs/api/PatientServiceTest.java#L1908). Do we think that's sufficient?

mogoodrich:
Probably that coverage is enough @brandones 

@rkorytkowski  any chance you have time to do a quick sanity check before we merge? Or, if not, is there anyone else you'd recommend to review this?  Thanks!

pom.xml
rkorytkowski: What's the build failure message if you do not include that? Is it failing to compile or some tests failing? Could you please paste the full output?

api/src/main/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactory.java
```diff
@@ -0,0 +1,59 @@
...
+ * 	startAnalyzer, which allows searching for tokens that match at the beginning
+ * 	exactAnalyzer, which allows searching for tokens that are identical
+ * 	anywhereAnalyzer, which allows searching for text within tokens
+ *
+ * @since 2.4.0
+ */
+public class LuceneAnalyzerFactory {
+	@Factory
+	public SearchMapping getSearchMapping() {
+		SearchMapping mapping = new SearchMapping();
```
rkorytkowski: Didn't you want to have some way to inject custom analyzers from modules? How do you overwrite that factory from a module?

api/src/main/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactory.java
rkorytkowski: Good to know. I thought that all needs to be in the programmatic API. In the long run we still may want to move from annotations on a class to programmatic API, but it's okay to do it for analyzers only, if it works.

api/src/main/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactory.java
brandones: Definitely, yes. The work of figuring that out is [ticketed here](https://issues.openmrs.org/browse/TRUNK-5677).

mogoodrich:
Thanks for the review @rkorytkowski !

pom.xml
brandones: Whoops, I looked at the output and realized it's that I had conflicting POMs. I've removed the dependency.

brandones:
Yeah, thanks @rkorytkowski ! I've addressed feedback. CC @mogoodrich , if you want to approve and merge.

PR #3044, opened 2019-11-04
TRUNK-5679  Add tests for Lucene Analyzer Factory

brandones:
https://issues.openmrs.org/browse/TRUNK-5679

Seeking feedback. Not sure if this is the right approach or not, but I'm trying to follow https://dzone.com/articles/quick-intro-hibernate-search

### DIFF
--- a/api/src/test/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactoryTest.java
+++ b/api/src/test/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactoryTest.java
@@ -1,0 +1,86 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.db.hibernate.search;
+
+import org.hibernate.search.annotations.Analyzer;
+import org.hibernate.search.annotations.Boost;
+import org.hibernate.search.annotations.DocumentId;
+import org.hibernate.search.annotations.Field;
+import org.hibernate.search.annotations.Fields;
+import org.hibernate.search.annotations.Indexed;
+import org.hibernate.search.annotations.IndexedEmbedded;
+import org.hibernate.search.cfg.SearchMapping;
+import org.junit.Assert;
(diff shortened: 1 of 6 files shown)


api/src/test/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactoryTest.java
```diff
@@ -0,0 +1,86 @@
+/**
```
brandones: This is the relevant file in this PR; the others are from https://github.com/openmrs/openmrs-core/pull/3040

mogoodrich:
Thanks @brandones ... assumedly there's a way to test existing objects that have been annotated, like PersonName?  Or do those tests already exist?

brandones:
It seems like it should be possible to test any kind of Lucene search mappings using the technique described in the blog post, whether provided through annotations or programmatically.

There are no existing tests having to do with Hibernate Search / Lucene.

Yeah, actually, perhaps it would be more appropriate to write these tests for PatientName rather than some hypothetical TestObject. Since LuceneAnalyzerFactory doesn't actually have any logic it maybe doesn't make a lot of sense to test it yet. [Making it configurable](https://issues.openmrs.org/browse/TRUNK-5669) would warrant giving it some tests.

mogoodrich:
Sounds good @brandones !

brandones:
Actually, @mogoodrich , why don't I just write these tests against [HibernatePatientDAO](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java)? This is where the PersonName queries [actually happen](https://github.com/openmrs/openmrs-core/blob/640cd7dd47f65c02574458f4b25d8b114aa817ed/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java#L194). And right now there are [no tests](https://github.com/openmrs/openmrs-core/blob/640cd7dd47f65c02574458f4b25d8b114aa817ed/api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java#L28) for it.

Actually, should I be writing tests against HibernatePatientDAO directly? Or is there a way to write tests against the PatientService (or PatientDAO?), which could then run against the Hibernate DAO or any other implementation?

mogoodrich:
There should be tests against the PatientService already, there must be a PatientServiceTest class?  Would be interesting to see if those actually cover Lucene searches for patients... (they very well might).

brandones:
Oh whoops, you're right! The relevant tests are in PatientServiceTest. Including [coverage for Lucene stuff](https://github.com/openmrs/openmrs-core/blob/640cd7dd47f65c02574458f4b25d8b114aa817ed/api/src/test/java/org/openmrs/api/PatientServiceTest.java#L1908)! Should I close this, since it's apparently superfluous?

mogoodrich:
Sounds good @brandones , yeah feel free to close this one.

PR #3047, opened 2019-11-08
    TRUNK-5669  Lucene analyzer used for patient search should be configurable via Settings/GP

brandones:
This is a draft; it is not working. The error that Spring is producing is in pastebin (split up because it's >1MB of error -- [part 1](https://pastebin.com/g557SNS5), [part 2](https://pastebin.com/SZZ9PvVR), [part 3](https://pastebin.com/qTHuLHkn).

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactory.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/search/LuceneAnalyzerFactory.java
@@ -14,15 +14,22 @@ import org.apache.lucene.analysis.core.WhitespaceTokenizerFactory;
 import org.apache.lucene.analysis.ngram.EdgeNGramFilterFactory;
 import org.apache.lucene.analysis.ngram.NGramFilterFactory;
 import org.apache.lucene.analysis.util.TokenFilterFactory;
+import org.hibernate.SessionFactory;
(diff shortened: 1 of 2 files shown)


mseaton:
I agree Spring startup errors are very annoying.  The relevant parts to me look to me:

Error creating bean with name 'sessionFactory' defined in class path resource [applicationContext-service.xml]: Invocation of init method failed; nested exception is org.hibernate.search.exception.SearchException: Unable to call the factory method: org.openmrs.api.db.hibernate.search.LuceneAnalyzerFactory.getSearchMapping

and

Caused by: org.springframework.beans.factory.BeanCurrentlyInCreationException: Error creating bean with name 'sessionFactory': FactoryBean which is currently in creation returned null from getObject

My guess is that there is a circular dependency.

brandones:
This is still not working; I am going to abandon it in favor of https://issues.openmrs.org/browse/TRUNK-5681, which meets my immediate needs.

brandones:
For posterity, in case this gets lost from the comments in the branch:

I tried three approaches to getting the global properties.

1. Context.getAdministrationService().getGlobalProperty(gpName) directly in getSearchMapping
2. Autowired AdministrationService
3. Initializing the value in a PostConstruct method

All of these approaches produced a circular dependency with sessionFactory.

PR #3060, opened 2019-11-21
TRUNK-5687, filter out DISCONTINUE orders

cioan:
Hi @mogoodrich , here is the PR for filtering out DISCONTINUE orders. Thanks.

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateOrderDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateOrderDAO.java
@@ -9,6 +9,8 @@
  */
 package org.openmrs.api.db.hibernate;
 
+import static org.openmrs.Order.Action.DISCONTINUE;
+
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/openmrs/api/db/hibernate/HibernateOrderDAO.java
```diff
@@ -197,6 +199,11 @@ public Order getOrder(Integer orderId) throws DAOException {
         if (searchCriteria.getAction() != null) {
             crit.add(Restrictions.eq("action", searchCriteria.getAction()));
         }
+        if (searchCriteria.getExcludeDiscontinueOrders() == true) {
+            crit.add(Restrictions.or(
+                    Restrictions.ne("action", Order.Action.DISCONTINUE),
+                    Restrictions.isNull("action")));
```
mogoodrich: This line isn't needed since Order.action cannot be null.

mogoodrich:
Merging, thanks @cioan !

PR #3065, opened 2019-11-26
TRUNK-5113: Migrate from log4j to log4j2

ibacher:

This converts the central logging system from log4j to log4j2. It also adds the log4j-1.x bindings for log4j2 which should enable any modules that are using the log4j binding to continue to work as expected.

## Issue I worked on:
JIRA: https://issues.openmrs.org/browse/TRUNK-5113

## Checklist:
- [x] I have tested these changes against the RefApp version 2.9 and they don't appear to break anything

## Caveats:

Log4j allows for extensive customization and the available bindings do not try to replicate that level of customization. Instead, they primarily implement redirecting of logging calls themselves. This means that some modules may not work as expected. This is hopefully only a small number of cases.


api/src/test/java/org/openmrs/util/Log4JCompatibilityTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.List;
+
+import org.apache.log4j.Logger;
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.core.layout.PatternLayout;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
ODORA0: There are a couple of wild cards in your imports you may want to take a look at this https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-Donotusewildcardimports

ibacher: Thanks for the catch! I apparently skipped that step in setting up my editor. I'll fix that.

dkayiwa:
@ibacher did you get a chance to look at this? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

ibacher:
@dkayiwa: I did. I’ve assigned the ticket to myself and marked it for code review. Is there anything else I may have missed or misunderstood? Thanks.

webapp/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,55 @@
...
+			INFO == log all setters
+			DEBUG == log all setters & log all getters & log execution time
+		-->
+		<Logger name="org.openmrs.api" level="INFO" />
+		<Logger name="org.apache.fop" level="ERROR" />
+		<!-- Hide the useless MissingResourceException -->
+		<Logger name="org.springframework.context.support.ResourceBundleMessageSource" level="ERROR" />
+		<Logger name="org.springframework.beans.factory.support.DefaultListableBeanFactory" level="ERROR" />
+		<Root>
+			<AppenderRef ref="CONSOLE" />
```
dkayiwa: Did you intentionally remove the level WARN which was on line 83?

dkayiwa:
@ibacher have you tried to run the OpenMRS reference application with these changes?

ibacher:
@dkayiwa Yes, with the caveat that I haven’t tried the 2.10 prerelease.

dkayiwa:
And the log out put is the same as before?

ibacher:
@dkayiwa Yes, the log output is the same, with the exception that the Liquibase actions get logged through the log4j2 logger.

To test, I stood up a distribution of the Ref App using the SDK and then swapped out the WAR for a version of the WAR with these changes backported (so, e.g., RefApp 2.9.0 was tested on both 2.4.0-SNAPSHOT and 2.2.1-SNAPSHOT; 2.8.1 against 2.1.5-SNAPSHOT).

dkayiwa:
@ibacher did you try the `2.4.0-SNAPSHOT`?

ibacher:
@dkayiwa Yes; it works fine on platform 2.4.0-SNAPSHOT. I didn't try with the ref app 2.10.0-SNAPSHOT, but that's because it was going through that rather rapid phase of UI-related changes.

ibacher:
I should say that this requires an SDK version with [SDK-245](https://github.com/openmrs/openmrs-sdk/pull/157), since I created that ticket and PR in response to the need to deploy this module.

dkayiwa:
And all the RefApp 2.9.0 modules started without errors on platform 2.4.0-SNAPSHOT?

ibacher:
They start without errors that seem to be related to logging. There are errors related to the Spring configuration, but I think that’s a result of trying to run against a newer version of core rather than these changes. That’s why I tested by backporting to platform 2.2 and 2.1.

dkayiwa:
@ibacher i have just setup the 2.4.0-SNAPSHOT of the platform using the OpenMRS SDK. Then replaced the `openmrs-2.4.0-SNAPSHOT.war` file with the one compiled with your changes. On running, i get this as part of the log: https://pastebin.com/kWWRFg77

ibacher:
@dkayiwa Did you see my note [here](https://github.com/openmrs/openmrs-core/pull/3065#issuecomment-567106774)? Log4j2 uses the Java Module System so it needs to be run under a version of Tomcat that supports the Java Module System.

dkayiwa:
Does it mean that with this change, we shall no longer support those other versions of tomcat?

ibacher:
@dkayiwa: Yes, we'll have to change support to Tomcat 7.0.76 or later. But this is necessary if we're going to support any code compiled against Java 9 or later.

dkayiwa:
@ibacher when i change the value of the setting named `log.level` from `org.openmrs.api:info` to `org.openmrs.api:debug` i do not see the changes in the logs.

ibacher:
@dkayiwa I just want to acknowledge I'm looking into this

ibacher:
@dkayiwa I believe my last commit fixes the issue. The Travis build failure seems to be an issue on Travis. Would you be able to manually restart that build? Thanks!

dkayiwa:
@ibacher after changing the log level, we are now required to restart the application. Yet before this change, the changes would take effect immediately. Could you look into this too?

ibacher:
@dkayiwa I agree that is the expected behaviour, but I don't think the regression is caused by this PR. If I spin up a new version of ref app 2.9.0 on platform 2.1.4, I still need to reboot for changes to the `log.level` setting to take effect.

The problem, AFAICT, is that `OpenmrsUtil#applyLogLevels()` is only called during startup. I think it would be easy to add a GlobalPropertyListener to call the appropriate method when the `log.level` property is changed, but should that be done as part of this PR?

dkayiwa:
@ibacher am running the latest versions of the reference application modules on the master branch (2.4.0-SNAPSHOT) of the platform. When i change the log.level from org.openmrs.api:info to org.openmrs.api:debug i immediately see the effects in the log. And i have tested this several times today. 😊

api/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,29 @@
...
+
+<Configuration xmlns="http://logging.apache.org/log4j/2.0/config">
+	<Appenders>
+		<Console name="CONSOLE" target="SYSTEM_OUT">
+			<PatternLayout pattern="%p - %C{1}.%M(%L) |%d{ISO8601}| %m%n" />
+		</Console>
+	</Appenders>
+	<Loggers>
+		<Logger name="org.hibernate.tool.hbm2ddl.SchemaExport" level="OFF" />
+		<Logger name="org.openmrs.test" level="OFF" />
```
dkayiwa: Is there any reason why this was duplicated?

api/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,29 @@
...
+-->
+
+<Configuration xmlns="http://logging.apache.org/log4j/2.0/config">
+	<Appenders>
+		<Console name="CONSOLE" target="SYSTEM_OUT">
+			<PatternLayout pattern="%p - %C{1}.%M(%L) |%d{ISO8601}| %m%n" />
+		</Console>
+	</Appenders>
+	<Loggers>
+		<Logger name="org.hibernate.tool.hbm2ddl.SchemaExport" level="OFF" />
```
dkayiwa: Did you intentionally leave out `org.openmrs.api`?

api/src/test/resources/log4j2.xml
ibacher: Yes. I meant to change one of the two to be org.openmrs.api.

ibacher:
@dkayiwa those little things are all set. I'm still trying to figure out the setting thing.

ibacher:
I've updated this PR to add a GP listener that updates log levels when the logging level is changed.

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -475,11 +480,11 @@ public static void startup(Properties p) {
 			val = OpenmrsConstants.DATABASE_NAME;
 		}
 		OpenmrsConstants.DATABASE_BUSINESS_NAME = val;
+
```
dkayiwa: Did you intentionally reverse setupLogAppenders() and applyLogLevels()?

api/src/main/java/org/openmrs/util/LoggingListener.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package org.openmrs.util;
+
+import org.openmrs.GlobalProperty;
+import org.openmrs.api.GlobalPropertyListener;
+
+public class LoggingListener implements GlobalPropertyListener {
+
+	/**
+	 * Asks this listener whether it wants to be notified about the given property name
```
dkayiwa: How about using the `@see` annotation for the javadocs instead of duplicating them?

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
ibacher: Yes. Initially, this was part of getting the settings to override the XML file, it's not strictly-speaking necessary.

dkayiwa:
@ibacher when i go to the `webapp` folder and run these changes with `mvn jetty:run`, i get this: https://pastebin.com/Z8qyKaqf

ibacher:
@dkayiwa I didn't even know that feature existed! It's been fixed by [upgrading the Jetty version](https://github.com/openmrs/openmrs-core/pull/3065/files#diff-6d3089830b3c2fa8d1d8902c7e7c8c9cR109) from the 9.3 series to 9.4.

api/src/main/java/org/openmrs/util/LoggingListener.java
```diff
@@ -0,0 +1,40 @@
...
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.util;
+
+import org.openmrs.GlobalProperty;
+import org.openmrs.api.GlobalPropertyListener;
+
```
dkayiwa: It would be great to add a `@since` annotation. https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-NewPublicMethodsandClasses

webapp/pom.xml
```diff
@@ -106,7 +106,7 @@
          <plugin>
              <groupId>org.eclipse.jetty</groupId>
              <artifactId>jetty-maven-plugin</artifactId>
-             <version>9.3.3.v20150827</version>
```
dkayiwa: Do you think you can look into removing the extra things in the logs when we upgrade jetty? 
This is the log for running the platform off the master branch (without any modules): https://pastebin.com/XJYNQuVH
And this one is after the jetty upgrade: https://pastebin.com/53CJ63iZ

ibacher: Could you try with the latest update? I'm not certain I have things setup correctly on my environment, but I think I should've fixed that.

dkayiwa:
@ibacher does `mvn jetty:run` in the `webapp` folder work?

ibacher:
@dkayiwa it does, I just get somewhat different logs than you do (I didn’t have all the warnings).

dkayiwa:
@ibacher would it be too hard to maintain the logs just as they were?

ibacher:
@dkayiwa Sorry I was testing with the wrong branch of the platform. The last change gets rid of all those warnings and should match the previous output.

dkayiwa:
@ibacher the logs are still very different. This is the log for the platform on the master branch without your changes: https://pastebin.com/H2QAvLdL

And this is the one for the same master branch of the platform, but with your changes in this pull request: https://pastebin.com/2tUEDea4

ibacher:
@dkayiwa Thanks for your patience with this PR! The extra logging information was caused [by a bug in JBOSS logging](https://issues.redhat.com/browse/JBLOGGING-107) library. That's been fixed in the latest version.

dkayiwa:
@ibacher thank you too for your resilience! 👍 

When `log.level` is set to `org.openmrs.api:debug`, after starting the openmrs instance with `mvn jetty:run` to the point where the server has booted successfully and hence waiting for requests, that is after this log message `[INFO] Started Jetty Server`, i should keep getting multiple lines of logs which keep adding for as long as the openmrs instance is up, even without any user interface activity. However, with your changes in this pull request, i do not get the extra continuous logs. Can you look into this?

ibacher:
@dkayiwa I've been looking into this and while there are a bunch of messages that usually get logged when running using the SDK—and consequently Tomcat—I don't see the same messages logged when using Jetty.

[Here](https://pastebin.com/qUWfQnFp) is the output I get when running Jetty against the current master, and [here](https://pastebin.com/xJWVMp09) is the output I get when running with these changes.

I also don't get continual lines of logs without some additional action when using the SDK to run either the master branch of the platform or the master branch of the platform with these changes. However, I do get the additional log lines shown [in this paste](https://pastebin.com/pAfxyfXd) when going to `http://localhost:8080/openmrs`, but these seem to be related to the OWA module. Could you possibly provide me with a sample log of the messages that are missing?

dkayiwa:
@ibacher this is a sample log that i get when i just run the master branch of the platform, even without accessing the browser: https://pastebin.com/DuyhzYeF

And it keeps going on forever.

dkayiwa:
@ibacher i tried these changes on the latest master branch of the openmrs platform with `mvn jetty:run` while running the reference application 2.10.0-SNAPSHOT modules. On changing the log level, i do not see the effects until when i restart OpenMRS.

ibacher:
@dkayiwa that's interesting. For me, it works as expected when I toggle the setting. See [this output](https://pastebin.com/xj3rNvDu).

dkayiwa:
@ibacher can i see two logs? One where the log level was `warn`, and the the other where the level was changed to `debug`, all without restarting openmrs.

ibacher:
So a bit more detail: it seems there's some weirdness caused by the LoggingAdvice interceptor itself. While other classes seem to obey the logging level correctly, LoggingAdvice itself doesn't. I'm investigating why this is the case.

dkayiwa:
Thanks for the investigation!
Ping me when you break through.

deniger:
Hello, by default the logger is using a RollingFileAppender  with a policy set to 10 Mo:
https://github.com/openmrs/openmrs-core/blob/8f67100a2c1344b61e7660439d58d6d2c5a37307/api/src/main/java/org/openmrs/util/OpenmrsUtil.java#L538
Is it possible to configure this limit ?

dkayiwa:
@deniger feel free to [create a ticket](https://wiki.openmrs.org/display/docs/Tickets) and raise a pull request that makes it as configurable as you would like it. 😊

PR #3071, opened 2019-12-03
GCI-142: Check bad code practices through Sonar and fix those issues - 1

PrathameshMutkure:
GCI-142: Check bad code practices through Findbugs or Sonar and fix those issues.

I've fixed some memory leaks as indicated by the Sonar :

- [issue 1](https://sonar.openmrs.org/project/issues?id=org.openmrs%3Aopenmrs&issues=AWGtPQ4reC2CQ1jcfcWi&open=AWGtPQ4reC2CQ1jcfcWi)
- [issue 2](https://sonar.openmrs.org/project/issues?id=org.openmrs%3Aopenmrs&issues=AWHiqMHLs_CtjRdw4sjQ&open=AWHiqMHLs_CtjRdw4sjQ)
- [issue 3](https://sonar.openmrs.org/project/issues?id=org.openmrs%3Aopenmrs&issues=AWGtPQ4reC2CQ1jcfcWh&open=AWGtPQ4reC2CQ1jcfcWh)
- [issue 4](https://sonar.openmrs.org/project/issues?id=org.openmrs%3Aopenmrs&issues=AWGtPQ4reC2CQ1jcfcWj&open=AWGtPQ4reC2CQ1jcfcWj)
- [issue 5](https://sonar.openmrs.org/project/issues?id=org.openmrs%3Aopenmrs&issues=AWHiqMHLs_CtjRdw4sjP&open=AWHiqMHLs_CtjRdw4sjP)
- [issue 6](https://sonar.openmrs.org/project/issues?id=org.openmrs%3Aopenmrs&issues=AWHiqMHLs_CtjRdw4sjR&open=AWHiqMHLs_CtjRdw4sjR)

Explanation for the code : http://www.informit.com/articles/article.aspx?p=31551&seqNum=2
## Issue I worked on

see https://issues.openmrs.org/browse/GCI-142

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/Condition.java
+++ b/api/src/main/java/org/openmrs/Condition.java
@@ -105,8 +105,8 @@ public class Condition extends BaseChangeableOpenmrsData {
 		this.verificationStatus = verificationStatus;
 		this.previousVersion = previousVersion;
 		this.additionalDetail = additionalDetail;
-		this.onsetDate = onsetDate;
-		this.endDate = endDate;
+		this.onsetDate = onsetDate != null ? new Date(onsetDate.getTime()) : null;
+		this.endDate = endDate != null ? new Date(endDate.getTime()) : null;
 		this.patient = patient;
 	}
 	
@@ -246,7 +246,7 @@ public class Condition extends BaseChangeableOpenmrsData {
 	 *         set
 	 */
 	public Date getOnsetDate() {
-		return onsetDate;
+		return onsetDate != null ? (Date) onsetDate.clone() : null;
 	}
 	
 	/**
@@ -255,7 +255,7 @@ public class Condition extends BaseChangeableOpenmrsData {
 	 * @param onsetDate the onset date of the condition to be set
 	 */
 	public void setOnsetDate(Date onsetDate) {
-		this.onsetDate = onsetDate;
+		this.onsetDate = onsetDate != null ? new Date(onsetDate.getTime()) : null;
 	}
 	
 	/**
@@ -264,7 +264,7 @@ public class Condition extends BaseChangeableOpenmrsData {
 	 * @return endDate - a date object that shows the end date of the condition
 	 */
 	public Date getEndDate() {
-		return endDate;
+		return endDate != null ? (Date) endDate.clone() : null;
 	}
 	
 	/**
@@ -273,7 +273,7 @@ public class Condition extends BaseChangeableOpenmrsData {
 	 * @param endDate the end date to be set for the condition
 	 */
 	public void setEndDate(Date endDate) {
-		this.endDate = endDate;
+		this.endDate = endDate != null ? new Date(endDate.getTime()) : null;
 	}
 	
 	/**


api/src/main/java/org/openmrs/Condition.java
```diff
@@ -255,7 +255,7 @@ public Date getOnsetDate() {
 	 * @param onsetDate the onset date of the condition to be set
 	 */
 	public void setOnsetDate(Date onsetDate) {
-		this.onsetDate = onsetDate;
+		this.onsetDate = new Date(onsetDate.getTime());
```
suthagar23: What happens, if the `onsetDate` is null?

api/src/main/java/org/openmrs/Condition.java
```diff
@@ -273,7 +273,7 @@ public Date getEndDate() {
 	 * @param endDate the end date to be set for the condition
 	 */
 	public void setEndDate(Date endDate) {
-		this.endDate = endDate;
+		this.endDate = new Date(endDate.getTime());
```
suthagar23: What happens, if the `endDate` is null?

api/src/main/java/org/openmrs/Condition.java
```diff
@@ -105,8 +105,8 @@ public Condition(CodedOrFreeText condition, ConditionClinicalStatus clinicalStat
 		this.verificationStatus = verificationStatus;
 		this.previousVersion = previousVersion;
 		this.additionalDetail = additionalDetail;
-		this.onsetDate = onsetDate;
-		this.endDate = endDate;
+		this.onsetDate = new Date(onsetDate.getTime());
```
suthagar23: What happens, if the `onsetDate` and `endDate` is null?

api/src/main/java/org/openmrs/Condition.java
PrathameshMutkure: Well I have fixed that issue, we can simply set those dates to null and cause NullPointerException in case user passes null values to prevent them from doing so

api/src/main/java/org/openmrs/Condition.java
```diff
@@ -264,7 +264,7 @@ public void setOnsetDate(Date onsetDate) {
 	 * @return endDate - a date object that shows the end date of the condition
 	 */
 	public Date getEndDate() {
-		return endDate;
+		return endDate != null ? (Date) endDate.clone() : new Date();
```
suthagar23: Why you are using diffrent methods here? 
1 . `new Date(onsetDate.getTime())`
2.  `(Date) endDate.clone()`
Better to keep the consistency among the code, and provide your feedback why did you select that among others.

PrathameshMutkure: Thanks for pointing out that

I'm going for `(Date) endDate.clone()`, because returning `new Date(onsetDate.getTime())` is causing some tests to fail and also, using `.clone()` method makes it easy to read and understand the code.

**P.S.:** I've squashed all commits into 1

Thanks

api/src/main/java/org/openmrs/Condition.java
```diff
@@ -105,8 +105,8 @@ public Condition(CodedOrFreeText condition, ConditionClinicalStatus clinicalStat
 		this.verificationStatus = verificationStatus;
 		this.previousVersion = previousVersion;
 		this.additionalDetail = additionalDetail;
-		this.onsetDate = onsetDate;
-		this.endDate = endDate;
+		this.onsetDate = onsetDate != null ? (Date) onsetDate.clone() : null;
```
dkayiwa: Did you get a chance to look at this? http://www.informit.com/articles/article.aspx?p=31551&seqNum=2

PrathameshMutkure: Well, in my previous commit I used `new Data(onsetDate.getTime())` but suthagar reccommended me to maintain consistency between `new Data(onsetDate.getTime())` and `(Date) onsetDate.clone()` so I choosed the second one as the first method didn't work when returning the value

dkayiwa: After reading that page, what do you think is the recommended way forward?

PrathameshMutkure: The artical does not recommend to use the `clone()` method as it posses some security risks, I think we should go for `new Date(onsetDate.getTime())`

dkayiwa: Does it not recommend clone in all cases? Or just some?

PrathameshMutkure: `clone()` should be used only when returning values and new copy must be made while setting the values, so in my case I'll have to replace the clone with new copy in setters

PrathameshMutkure: I'll push those changes soon @dkayiwa

PrathameshMutkure:
Sorry I messed up with rebasing, I'm fixing it

api/src/main/java/org/openmrs/Condition.java
```diff
@@ -105,8 +105,8 @@ public Condition(CodedOrFreeText condition, ConditionClinicalStatus clinicalStat
 		this.verificationStatus = verificationStatus;
 		this.previousVersion = previousVersion;
 		this.additionalDetail = additionalDetail;
-		this.onsetDate = onsetDate;
-		this.endDate = endDate;
+		this.onsetDate = onsetDate != null ? new Date(onsetDate.getTime()) : null;
+		this.endDate = onsetDate != null ? new Date(endDate.getTime()) : null;
```
dilanthas: I think should be endDate != null

PrathameshMutkure: Yeah, sorry my bad

api/src/main/java/org/openmrs/Condition.java
```diff
@@ -246,7 +246,7 @@ public void setAdditionalDetail(String additionalDetail) {
 	 *         set
 	 */
 	public Date getOnsetDate() {
-		return onsetDate;
+		return onsetDate != null ? (Date) onsetDate.clone() : new Date();
```
dkayiwa: Aren't you changing the original logic of this method?

api/src/main/java/org/openmrs/Condition.java
dkayiwa: Same as above.

api/src/main/java/org/openmrs/Condition.java
PrathameshMutkure: That thing popped up in my mind too, but previous reviewers suggested me to check for null values and also can you check [this](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java#L70) statement here, it has similar behaviour for null values, so instead of letting the user to check for null values I thought it would be better if we do that in the method itself. What are your thoughts on this?

api/src/main/java/org/openmrs/Condition.java
PrathameshMutkure: Again [this](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java#L50) statement here

api/src/main/java/org/openmrs/Condition.java
dkayiwa: If you want to change the logic of this method, it would be better to use a different pull request and let this stick to its original intention. 😊

api/src/main/java/org/openmrs/Condition.java
dkayiwa: As said above

api/src/main/java/org/openmrs/Condition.java
PrathameshMutkure: Yeah, I think it would be better to keep the original intention of this PR :+1:

PrathameshMutkure:
Hii @dkayiwa, fixed everything :)

dkayiwa:
Thanks @PrathameshMutkure 😊

PR #3117, opened 2020-01-16
TRUNK-3401: deprecated method Encounter.setProvider against 1.9 needs…

ODORA0:
… to be adjusted

**Issue worked on:** https://issues.openmrs.org/browse/TRUNK-3401

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/ProviderServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ProviderServiceImpl.java
@@ -17,6 +17,7 @@ import org.openmrs.Person;
 import org.openmrs.Provider;
 import org.openmrs.ProviderAttribute;
 import org.openmrs.ProviderAttributeType;
+import org.openmrs.User;
 import org.openmrs.api.APIException;
 import org.openmrs.api.ProviderService;
 import org.openmrs.api.context.Context;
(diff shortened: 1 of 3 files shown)


ODORA0:
Still a WIP but some of the errors are not recieved locally

dkayiwa:
Did you see the travis build failure?

api/src/main/java/org/openmrs/api/impl/ProviderServiceImpl.java
```diff
@@ -293,4 +296,19 @@ public Provider getUnknownProvider() {
...
+
+	@Override
+	public Provider createProviderFromUser(User user){
+		Provider p = new Provider();
+		if (user == null || user.getPerson() == null){
+			throw new APIException("User cannot be null");
+		} else if (Context.getProviderService().getProvidersByPerson(user.getPerson()).isEmpty()){
+			p.setPerson(user.getPerson());
+			return Context.getProviderService().saveProvider(p);
+		} else {
```
ShivanshRakesh: Maybe you can avoid using `else` statement here, since it is unnecessary (refer to [this](https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-CodeStyle)). Rather you can simply write:
```
    if (user == null || user.getPerson() == null){
		throw new APIException("User cannot be null");
	} else if (Context.getProviderService().getProvidersByPerson(user.getPerson()).isEmpty()){
		p.setPerson(user.getPerson());
		return Context.getProviderService().saveProvider(p);
	} 
	log.warn("Provider already exists for user");
	return
	Context.getProviderService().getProvidersByPerson(user.getPerson()).iterator().next();
```

ShivanshRakesh: Also, this is the part (`else if` and `else`) where you're missing out on code coverage, and hence the check failure.

PR #3136, opened 2020-02-27
TRUNK-5655: Added a utility method to get a Concept by it's identifier, name or mapping

jnsereko:
https://issues.openmrs.org/browse/TRUNK-5655

### DIFF
--- a/api/src/main/java/org/openmrs/api/ConceptService.java
+++ b/api/src/main/java/org/openmrs/api/ConceptService.java
@@ -1923,4 +1923,23 @@ public interface ConceptService extends OpenmrsService {
 	@Authorized(PrivilegeConstants.GET_CONCEPTS)
 	boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType);
 	
+	/**
+	 * Get the concept by conceptRef where the conceptRef can either be: 1) an integer id like 5090
+	 * 2) a mapping type id like "XYZ:HT" 3) a uuid like "a3e12268-74bf-11df-9768-17cfc9833272" 4) a
+	 * name like "PLATELETS" 5) the fully qualified name of a Java constant that contains one of
+	 * above
+	 * 
+	 * @param conceptRef the concept string identifier
+	 * @return the concept if exist, else null
+	 * <strong>Should</strong> find a concept by its conceptId
+	 * <strong>Should</strong>  find a concept by its mapping
+	 * <strong>Should</strong> find a concept by its uuid
+	 * <strong>Should</strong> find a concept by its name
+	 * <strong>Should</strong> find a concept by a fully qualified name of a Java constant that contains an id,
+	 * 								uuid, name or mapping
+	 * <strong>Should</strong> return null when ConceptRef does not exist
+	 */
+	@Authorized(PrivilegeConstants.GET_CONCEPTS)
+	public Concept getConceptByNameOrIdOrMap(String conceptRef);
 }
+
--- a/api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
@@ -11,6 +11,7 @@ package org.openmrs.api.impl;
 
 import static org.apache.commons.lang3.StringUtils.contains;
 
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -27,6 +28,7 @@ import java.util.UUID;
 
 import org.apache.commons.beanutils.BeanUtils;
 import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang3.math.NumberUtils;
 import org.openmrs.Concept;
 import org.openmrs.ConceptAnswer;
 import org.openmrs.ConceptAttribute;
@@ -1961,6 +1963,72 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 	@Transactional(readOnly = true)
 	public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType) {
 		return dao.getConceptAttributeCount(conceptAttributeType) > 0;
+	}
+	
+	/**
+	 * @see org.openmrs.api.ConceptService#getConceptByNameOrIdOrMap(String conceptRef
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	public Concept getConceptByNameOrIdOrMap(String conceptRef) throws APIException {
+		if (conceptRef == null) {
+			return null;
+		}
+		Concept cpt = null;
+		//check if input is a valid Uuid
+		if (!(conceptRef.length() < 36 || conceptRef.length() > 38 || conceptRef.contains(" ") || conceptRef.contains("."))) {
+			cpt = Context.getConceptService().getConceptByUuid(conceptRef);
+			if (cpt != null)
+				return cpt;
+		}
+		//check if input is map
+		if (conceptRef.indexOf(":") != -1) {
+			int idx = conceptRef.indexOf(":");
+			String conceptSource, conceptCode;
+			conceptCode = conceptRef.substring(idx + 1);
+			conceptSource = conceptRef.substring(0, idx);
+			cpt = Context.getConceptService().getConceptByMapping(conceptCode, conceptSource);
+			if (cpt != null)
+				return cpt;
+		}
+		//check if input is a name or id
+		if (!conceptRef.isEmpty()) {
+			//handle id
+			if (NumberUtils.isParsable(conceptRef)) {
+				int conceptId = Integer.valueOf(conceptRef);
+				cpt = Context.getConceptService().getConcept(conceptId);
+			} else
+				//handle name
+				cpt = Context.getConceptService().getConceptByName(conceptRef);
+			if (cpt != null)
+				return cpt;
+		}
+		if (conceptRef.contains(".")) {
+			return getConceptByNameOrIdOrMap(evaluateStaticConstant(conceptRef));
+		}
+		return cpt;
+	}
+	
+	/**
+	 * Evaluates the specified Java constant using reflection
+	 * 
+	 * @param fqn the fully qualified name of the constant
+	 * @return the constant value
+	 */
+	private static String evaluateStaticConstant(String fqn) {
+		int lastPeriod = fqn.lastIndexOf(".");
+		String clazzName = fqn.substring(0, lastPeriod);
+		String constantName = fqn.substring(lastPeriod + 1);
+		
+		try {
+			Class<?> clazz = Context.loadClass(clazzName);
+			Field constantField = clazz.getField(constantName);
+			Object val = constantField.get(null);
+			return val != null ? String.valueOf(val) : null;
+		}
+		catch (Exception ex) {
+			throw new IllegalArgumentException("Unable to evaluate " + fqn, ex);
+		}
 	}
 
 	private List<ConceptClass> getConceptClassesOfOrderTypes() {
--- a/api/src/test/java/org/openmrs/api/ConceptServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/ConceptServiceTest.java
@@ -2255,7 +2255,7 @@ public class ConceptServiceTest extends BaseContextSensitiveTest {
 		
 		// verify
 		Assert.assertThat(actualConcepts.size(), is(6));
-		Assert.assertThat(actualConcepts, containsInAnyOrder(conceptService.getConcept(3), conceptService.getConcept(60), conceptService.getConcept(88), conceptService.getConcept(792)));
+		Assert.assertThat(actualConcepts, containsInAnyOrder(conceptService.getConcept(3), conceptService.getConcept(64), conceptService.getConcept(60), conceptService.getConcept(67), conceptService.getConcept(88), conceptService.getConcept(792)));
 	}
 	
 	/**
@@ -2804,7 +2804,7 @@ public class ConceptServiceTest extends BaseContextSensitiveTest {
 		//check desc order
 		allConcepts = conceptService.getAllConcepts("dateCreated", false, true);
 		
-		assertEquals(36, allConcepts.size());
+		assertEquals(38, allConcepts.size());
 		assertEquals(23, allConcepts.get(0).getConceptId().intValue());
 		assertEquals(88, allConcepts.get(allConcepts.size() - 1).getConceptId().intValue());
 	}
--- a/api/src/test/java/org/openmrs/api/ConceptServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/ConceptServiceTest.java
@@ -3598,7 +3598,7 @@ public class ConceptServiceTest extends BaseContextSensitiveTest {
 	 * @see ConceptService#getConceptByNameOrIdOrMap(String)
 	 */
 	@Test
-	public void getConceptByNameOrIdOrMap_shouldReturnNullOtherwise() {
+	public void getConceptByNameOrIdOrMap_shouldReturnNullWhenEitherConceptRefIsInvalidOrDoesNotMatchAnyConcept() {
 		String givenNull = null;
 		Assert.assertNull(conceptService.getConceptByNameOrIdOrMap(givenNull));
 		
--- a/api/src/test/java/org/openmrs/api/ConceptServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/ConceptServiceTest.java
@@ -101,6 +101,10 @@ public class ConceptServiceTest extends BaseContextSensitiveTest {
 
 	protected static final String CONCEPT_ATTRIBUTE_TYPE_XML = "org/openmrs/api/include/ConceptServiceTest-conceptAttributeType.xml";
 
+	// For testing concept lookups by static constant
+	public static final int TEST_CONCEPT_CONSTANT_ID = 3;
+	public static final String TEST_CONCEPT_CONSTANT_UUID = "0cbe2ed3-cd5f-4f46-9459-26127c9265ab";
+
 	@Rule
 	public ExpectedException expectedException = ExpectedException.none();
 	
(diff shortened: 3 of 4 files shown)


sherrif10:
@jnsereko  did you see the travis ci failure, can you add the ticket link very well, the link you provided it not collect

jnsereko:
> @jnsereko did you see the travis ci failure, can you add the ticket link very well, the link you provided it not collect

the correct link to the ticket is https://issues.openmrs.org/browse/TRUNK-5655

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -1922,5 +1922,57 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
+	 * @param idOrMap
+	 * @return the found Concept
+	 * @throws APIException
+	 */
+	@Authorized(PrivilegeConstants.GET_CONCEPTS)
+	public static Concept getConceptInfo(String idOrMap) throws APIException {
+
+		Concept cpt = null;
+
+		if (idOrMap != null) {
```
JudeNiroshan: I would do the opposite check and throw the exception right away. In that case, less indentation makes code more easy for your eyes. 👀 something like below(no need to write `else` case):
```
if(idOrName == null) {
  throw new APIException("null isn`t alloed as acceptable input");
}
```

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -1922,5 +1922,57 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
+
+			idOrMap = idOrMap.trim();
+
+			//if input contains at least one period handle recursively as a code constant
+			if (idOrMap.contains(".")) {
+				return getConcept(evaluateStaticConstant(idOrMap));
+			}
+
+			// handle  mapping id: xyz:ht
+			int index = id.indexOf(":");
```
JudeNiroshan: not sure why you are treating the input as a Map. According to your method description it is either **an ID or a Name**. Isn't it?

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -1922,5 +1922,57 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
+			if (index != -1) {
+				cpt = Context.getService(HtmlFormEntryService.class).getConceptByMapping(idOrMap);
+				if (cpt != null) {
+					return cpt;
+				}
+			}
+
+			// handle uuid id: "a3e1302b-74bf-11df-9768-17cfc9833272", if the id matches a uuid format
```
JudeNiroshan: don't comment uuids in real code. This is data violation. You shouldn't put this comment at all :)

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -1922,5 +1922,57 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
+				if (cpt != null) {
+					return cpt;
+				}
+			}
+			catch (Exception ex) {
+				throw new APIException("id is entered invalid");
+			}
+		}
+		else
+			throw new APIException("null isn`t alloed as acceptable input");
```
JudeNiroshan: shouldn't this be `IllegalArgumentException` wrapped in a `APIException`?

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -1922,5 +1922,57 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
+
+			// handle uuid id: "a3e1302b-74bf-11df-9768-17cfc9833272", if the id matches a uuid format
+			if (isValidUuidFormat(idOrMap)) {
+				cpt = Context.getConceptService().getConceptByUuid(idOrMap);
+				if (cpt != null) {
+					return cpt;
+				}
+			}
+
+			//finally, see if this is a parseable int; if so, try looking up concept by id
```
JudeNiroshan: it's always better not to mix the comment with code lines. In fact try, not to write comments unless it is really needed. Your code should express what it does. You can choose nice variable names and method names to achieve that. :-)

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -1922,5 +1922,57 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
+			try { //handle integer: idOrMap
+				int conceptId = Integer.parseInt(idOrMap);
+				cpt = Context.getConceptService().getConcept(conceptIdOrMap);
+				if (cpt != null) {
+					return cpt;
+				}
+			}
+			catch (Exception ex) {
```
JudeNiroshan: Catching `Exception` is not a good idea. Exception is more generic. Try to catch more specific exceptions by examine the calls you make inside the try statement.

api/src/main/java/org/openmrs/api/ConceptService.java
jnsereko: @JudeNiroshan
 ``if(idOrName == null) {
  throw new APIException("null isn`t alloed as acceptable input");
}``
this is way more easier to read, thanks

JudeNiroshan: https://gph.is/g/ap0YXzx

JudeNiroshan: There is a small typo -> `null isn't allowed as...`

api/src/main/java/org/openmrs/api/ConceptService.java
jnsereko: >shouldn't this be `IllegalArgumentException` wrapped in a `APIException`?

do you mean i should catch `IllegalArgumentException` and through an `APIException`?

api/src/main/java/org/openmrs/api/ConceptService.java
jnsereko: How best can i include a map as an input?

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -1922,5 +1922,57 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
-	
+	/**
+	 * Get Concept by id or name convenience method
+	 * 
+	 * @param idOrMap
+	 * @return the found Concept
+	 * @throws APIException
+	 */
+	@Authorized(PrivilegeConstants.GET_CONCEPTS)
+	public static Concept getConceptInfo(String idOrMap) throws APIException {
```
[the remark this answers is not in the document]
samuelmale: @jnsereko This should also be resilient enough to also get concept by concept mapping. eg. 'CIEL:1234' 

So the naming becomes tricky here, I would keep it at:
`getConcept(String conceptRef)` or `getConcept(String id)` or `getConcept(String phrase)`

Since this method is being added to the service layer, I would not make it static hence:

```java

public Concept getConcept(String conceptRef)throws APIException {
         // some logic here
        return concept;
}
```

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -1922,5 +1922,57 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
+
+		Concept cpt = null;
```
Am-Coder: In java object references are by default initialized to null.So I don't think there is a need to specifically assign null.

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -1922,5 +1922,57 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
+			idOrMap = idOrMap.trim();
+
+			//if input contains at least one period handle recursively as a code constant
+			if (idOrMap.contains(".")) {
+				return getConcept(evaluateStaticConstant(idOrMap));
+			}
+
+			// handle  mapping id: xyz:ht
+			int index = id.indexOf(":");
+			if (index != -1) {
```
Am-Coder: Is declaring a separate index variable necessary? Can't we directly use the required value here i.e. like 

`if( id.indexOf(":") != -1 ){`

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -1922,5 +1922,57 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
+			// handle uuid id: "a3e1302b-74bf-11df-9768-17cfc9833272", if the id matches a uuid format
+			if (isValidUuidFormat(idOrMap)) {
+				cpt = Context.getConceptService().getConceptByUuid(idOrMap);
+				if (cpt != null) {
+					return cpt;
+				}
+			}
+
+			//finally, see if this is a parseable int; if so, try looking up concept by id
+			try { //handle integer: idOrMap
```
Am-Coder: To avoid using try catch to check if string is parsable you can make use of isParsable method in NumberUtils class.
[https://commons.apache.org/proper/commons-lang/javadocs/api-3.4/org/apache/commons/lang3/math/NumberUtils.html#isNumber(java.lang.String)](url)

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1977,4 +1977,72 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+			return val != null ? String.valueOf(val) : null;
+		}
+		catch (Exception ex) {
+			throw new IllegalArgumentException("Unable to evaluate " + fqn, ex);
+		}
+	}
+	/**
+	 * @see org.openmrs.api.ConceptService#getConceptByMapping(java.lang.String, java.lang.String)
+	 */
+	public Concept getConceptByMapping(String code) throws APIException {
```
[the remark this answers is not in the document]
jnsereko: Code breaks if i skip adding the new method.  thought it would be better to add a new maybe i made things _worse_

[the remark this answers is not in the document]
samuelmale: How does code break? You don't need this method; I guess it already exist's in the service layer.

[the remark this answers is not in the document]
jnsereko: method `getConceptMapping( )` aready exists but it takes **two Strings** and **one boolean** value as i**ts parameters**.  I was a kinder confused so i decided to add a new method that that takes one string as its parameter.  But this method i added might result into infinite calling of the method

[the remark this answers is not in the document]
samuelmale: So you assume adding a new method `getConceptMapping(String)` without a DAO implementation would fix things? 

Of course to get a concept by it's mapping means you have two things, ie. `code` and `source`, so the existing API should sufficient enough to handle this. 

_Still confused?_

- Get `source` and `code` from the concept mapping value eg. CIEL:123 where `CIEL` is the `source` and `123` is the `code` delimited by `:`.

Also see: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/ConceptService.java#L957

samuelmale:
> You should also consider adding test coverage for your new changes.

@jnsereko did you see the above comment?

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1977,4 +1977,46 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
 		}
 		return mappedClasses;
 	}
+
+	public Concept getConceptByRef(String conceptRef) throws APIException {
```
samuelmale: @jnsereko I would keep the naming to:
```java
public Concept getConcept(String conceptRef)

```

Or anything else better?

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1977,4 +1977,46 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+		if(conceptRef == null) {
+			throw new APIException("null isn`t allowed as acceptable input");
```
samuelmale: We possibly don't need this, how are NPEs or such cases handled in the entire Service layer?

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1977,4 +1977,46 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+			cpt = Context.getConceptService().getConceptByMapping(conceptCode, conceptSource);
+			if (cpt != null)
+				return cpt;
+		}
+		if ( !(conceptRef.length() < 36 || conceptRef.length() > 38 || conceptRef.contains(" ") || conceptRef.contains("."))){
+			cpt = Context.getConceptService().getConceptByUuid(conceptRef);
+			if (cpt != null)
+				return cpt;
+		}
+		cpt = getConcept(conceptRef);
```
samuelmale: What does this handle? `conceptId` or `name` or both?

And btw, you would just do:

```java
return getConcept(conceptRef);

```

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
jnsereko: the method `getConcept(String)` already exists in the class it results into a conflict, unless when i delete the aready existing `getConcept(String)` method.

jnsereko: how is `getConceptByNameIdOrMap(String)`?

api/src/test/java/org/openmrs/api/ConceptServiceTest.java
```diff
@@ -3531,4 +3531,39 @@ public void getConcepts_shouldPassWithAndOrNotWords() {
...
+	public void getConceptByNameOrIdOrMap_shouldGetConceptByItsMap() {
+		Concept concept = conceptService.getConceptByNameOrIdOrMap("WGT:234");
+		Assert.assertEquals(5089, concept.getId().intValue());
+	}
+	/**
+	 * @see ConceptService#getConceptByNameOrIdOrMap(String)
+	 */
+	@Test
+	public void getConceptByNameOrIdOrMap_shouldGetConceptByConceptIdOrUuid() {
+		String uuid = "0cbe2ed3-cd5f-4f46-9459-26127c9265ab";
```
alinmihaila99: You should use self-descriptive varibales names.

api/src/test/java/org/openmrs/api/ConceptServiceTest.java
```diff
@@ -3531,4 +3531,39 @@ public void getConcepts_shouldPassWithAndOrNotWords() {
...
+	public void getConceptByNameOrIdOrMap_shouldReturnNullGivenNullParameter() {
+		Assert.assertNull(Context.getConceptService().getConceptByNameOrIdOrMap(null));
+	}
+	
+	/**
+	 * @see ConceptService#getConceptByNameOrIdOrMap(String)
+	 */
+	@Test
+	public void getConceptByNameOrIdOrMap_shouldReturnNullGivenBlankString() {
+		Concept concept = conceptService.getConceptByNameOrIdOrMap("Some wrong input that matches no concept");
```
alinmihaila99: There is no need for this variable. You should put this in an assert.notNull

api/src/test/java/org/openmrs/api/ConceptServiceTest.java
```diff
@@ -3531,4 +3531,38 @@ public void getConcepts_shouldPassWithAndOrNotWords() {
...
+		assertNull(conceptService.getConceptByNameOrIdOrMap("Some wrong input that matches no concept"));
+	}
+
+	/**
+	 * @see ConceptService#getConceptByNameOrIdOrMap(String)
+	 */
+	@Test
+	public void getConceptByNameOrIdOrMap_shouldGetConceptByItsMap() {
+		Concept concept = conceptService.getConceptByNameOrIdOrMap("WGT:234");
+		Assert.assertEquals(5089, concept.getId().intValue());
```
Am-Coder: I had recently gone through the test datasets and there the code for a concept_reference_term was "WGT234" and not "WGT:234". I think here your returned concept is null because of "WGT:234". Your tests might be failing because of it.

jnsereko:
hello @dkayiwa, @samuelmale, may you please look at this and checkout if its missing anything.

api/src/test/java/org/openmrs/api/ConceptServiceTest.java
```diff
@@ -3531,4 +3536,76 @@ public void getConcepts_shouldPassWithAndOrNotWords() {
...
+	}
+
+	/**
+	 * @see ConceptService#getConceptByNameOrIdOrMap(String)
+	 */
+	@Test
+	public void getConceptByNameOrIdOrMap_shouldFindAConceptByItsUuid() {
+			String uuid = "0cbe2ed3-cd5f-4f46-9459-26127c9265ab";
+			Concept concept = Context.getConceptService().getConceptByNameOrIdOrMap(uuid);
+			Assert.assertEquals(3, (int) concept.getConceptId());
```
Am-Coder, line 3576: Why here an explicit typecasting is required ?

api/src/test/java/org/openmrs/api/ConceptServiceTest.java
```diff
@@ -3531,4 +3536,76 @@ public void getConcepts_shouldPassWithAndOrNotWords() {
...
+		assertThat(conceptService.getConceptByNameOrIdOrMap("org.openmrs.api.ConceptServiceTest.TEST_CONCEPT_CONSTANT_ID"), notNullValue());
+		assertThat(conceptService.getConceptByNameOrIdOrMap("org.openmrs.api.ConceptServiceTest.TEST_CONCEPT_CONSTANT_UUID"), notNullValue());
+		assertThat(conceptService.getConceptByNameOrIdOrMap("org.openmrs.api.ConceptServiceTest.TEST_CONCEPT_CONSTANT_NAME"), notNullValue());
+	}
+
+	/**
+	 * @see ConceptService#getConceptByNameOrIdOrMap(String)
+	 */
+	@Test
+	public void getConceptByNameOrIdOrMap_shouldReturnNullOtherwise() {
```
Am-Coder, line 3603: I think you should name this test method like `getConceptByNameOrIdOrMap_shouldReturnNullWhen< write condtition here like ConceptRefDoesNotExist >` 

NullOtherwise is not informative.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1977,4 +1978,67 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+	 * @return the constant value
+	 */
+	private static String evaluateStaticConstant(String fqn) {
+		int lastPeriod = fqn.lastIndexOf(".");
+		String clazzName = fqn.substring(0, lastPeriod);
+		String constantName = fqn.substring(lastPeriod + 1);
+
+		try {
+			Class<?> clazz = Context.loadClass(clazzName);
+			java.lang.reflect.Field constantField = clazz.getField(constantName);
```
Am-Coder, line 2036: Import Field and then use it here.

api/src/test/java/org/openmrs/api/ConceptServiceTest.java
jnsereko: code fails without the casting the result of `concept.getConceptId()` to int 
i think this is because `concept.getConceptId()` returns an **Integer Object**, **not** an int (primitive data-type)

Am-Coder: Ok, then its fine

jnsereko:
i accidentally closed this

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -1923,4 +1923,23 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
+	 * <strong>Should</strong> find a concept by its conceptId
+	 * <strong>Should</strong>  find a concept by its mapping
+	 * <strong>Should</strong> find a concept by its uuid
+	 * <strong>Should</strong> find a concept by its name
+	 * <strong>Should</strong> find a concept by a fully qualified name of a Java constant that contains an id,
+	 * 								uuid, name or mapping
+	 * <strong>Should</strong> return null when ConceptRef does not exist
+	 */
+	@Authorized(PrivilegeConstants.GET_CONCEPTS)
+	public Concept getConceptByNameOrIdOrMap(String conceptRef);
```
gitcliff, line 1943: @jnsereko thanks for the work on this, i think the method name should be` getConcept()` and the docs above plus the added unit tests will explain more about what the method does

gitcliff:
@jnsereko kindly have alook at the travis ci failure and also don't forget to squash the commits

api/src/main/java/org/openmrs/api/ConceptService.java
jnsereko: there is an aready existing method called `getConcept(String)` that gets a concept by id and name

gitcliff: oh okay,,,,,,,,,,
kindly have a look at the travis ci failure and also don't forget to squash the commits

api/src/main/java/org/openmrs/api/ConceptService.java
JudeNiroshan: APIException is more generic. Yes, you should always catch narrowed scoped exceptions

PR #3202, opened 2020-05-15
Trunk-5739:Expand updateOrderFulfillerStatus to also set accession number

gitcliff:
link:https://issues.openmrs.org/browse/TRUNK-5739

### DIFF
--- a/api/src/main/java/org/openmrs/api/OrderService.java
+++ b/api/src/main/java/org/openmrs/api/OrderService.java
@@ -293,6 +293,23 @@ public interface OrderService extends OpenmrsService {
 	public Order updateOrderFulfillerStatus(Order order, Order.FulfillerStatus orderFulfillerStatus, String fullFillerComment);
 	
 	/**
+	 * Updates the fulfillerStatus of an order and the related comment and finally persists it
+	 *
+	 * @param order order whose fulfillerStatus should be changed
+	 * @param orderFulfillerStatus describes the new Order.FulfillerStatus the order should be set to
+	 * @param fullFillerComment is a string which describes a comment that is set while changing the FulfillerStatus 
+	 * @param accessionNumber is the accession number to set             
+	 * @return the Order that is updated with an according fulfillerStataus and fulFillerComment and accession number
+	 * <strong>Should</strong> set the new fulfillerStatus
+	 * <strong>Should</strong> set the new fulFillerComment
+	 * <strong>Should</strong> set the new accessionNumber
+	 * <strong>Should</strong> save the changed order
+	 */
+	@Authorized(PrivilegeConstants.EDIT_ORDERS)
+	public Order updateOrderFulfillerStatus(Order order, Order.FulfillerStatus orderFulfillerStatus, String fullFillerComment, String accessionNumber);
+	
+	
+	/**
 	 * Gets the order identified by a given order number
 	 * 
 	 * @param orderNumber the order number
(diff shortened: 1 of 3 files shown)


api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -1253,6 +1253,24 @@ public void updateOrderFulfillerStatus_shouldEditFulfillerStatusInOrder() {
...
+	/**
+	 * @see OrderService#updateOrderFulfillerStatus(org.openmrs.Order,
+	 *      Order.FulfillerStatus, String, String)
+	 */
+	@Test
+	public void updateOrderFulfillerStatus_shouldEditFulfillerStatusWithAccessioNumberInOrder() {
+		Order originalOrder = orderService.getOrder(111);
+		String commentText = "We got the new order";
+		String accessionNumber = "12345";
+		assertNotEquals(originalOrder.getAccessionNumber(), 0);
```
mogoodrich, line 1265: Minor point, but shouldn't we really be testing here that the accession number is not "12345"?

api/src/main/java/org/openmrs/api/OrderService.java
```diff
@@ -292,6 +292,23 @@
 	@Authorized(PrivilegeConstants.EDIT_ORDERS)
 	public Order updateOrderFulfillerStatus(Order order, Order.FulfillerStatus orderFulfillerStatus, String fullFillerComment);
 	
+	/**
+	 * Updates the fulfillerStatus of an order and the related comment and finally persists it
+	 *
+	 * @param order order whose fulfillerStatus should be changed
+	 * @param orderFulfillerStatus describes the new Order.FulfillerStatus the order should be set to
+	 * @param fullFillerComment is a string which describes a comment that is set while changing the FulfillerStatus 
+	 * @param accessionNumber is the accession number of the order
```
mogoodrich, line 301: should be "is the accession number to set"

api/src/main/java/org/openmrs/api/OrderService.java
```diff
@@ -292,6 +292,23 @@
...
 	public Order updateOrderFulfillerStatus(Order order, Order.FulfillerStatus orderFulfillerStatus, String fullFillerComment);
 	
+	/**
+	 * Updates the fulfillerStatus of an order and the related comment and finally persists it
+	 *
+	 * @param order order whose fulfillerStatus should be changed
+	 * @param orderFulfillerStatus describes the new Order.FulfillerStatus the order should be set to
+	 * @param fullFillerComment is a string which describes a comment that is set while changing the FulfillerStatus 
+	 * @param accessionNumber is the accession number of the order             
+	 * @return the Order that is updated with an according fulfillerStataus and fulFillerComment
```
mogoodrich, line 302: ... and accession number"

api/src/test/java/org/openmrs/api/OrderServiceTest.java
gitcliff: @mogoodrich sure..i have fixed it

api/src/main/java/org/openmrs/api/OrderService.java
gitcliff: it done

api/src/main/java/org/openmrs/api/OrderService.java
gitcliff: its done to here

PR #3204, opened 2020-05-17
TRUNK-4830 Avoid running legacy liquibase changesets

WolfSchlegel:
Log details of un-run change sets.

This pull request provides additional logging when the OpenMRS database is updated. The additional log entries are limited to no more than 10 and un-run change sets are logged for 1.9.x Liquibase snapshot files only.

The additional logging was used to debug the database initialisation of the reference application and will help when other users report similar issues.

- [ x ] My pull request only contains **ONE single commit**
- [ x ] My IDE is configured to follow the [**code style**]- [ ] I have **added tests** to cover my changes. (If you refactored
- [ x ] I ran `mvn clean package` right before creating this pull request and
- [ x ] All new and existing **tests passed**.
- [ x ] My pull request is **based on the latest changes** of the master branch.

### DIFF
--- a/api/src/main/java/org/openmrs/liquibase/ChangeLogDetective.java
+++ b/api/src/main/java/org/openmrs/liquibase/ChangeLogDetective.java
@@ -44,6 +44,8 @@ public class ChangeLogDetective {
 	
 	private static final String LIQUIBASE_CORE_DATA_1_9_X_FILENAME = "liquibase-core-data-1.9.x.xml";
 	
+	private static final String LIQUIBASE_SCHEMA_ONLY_1_9_X_FILENAME = "liquibase-schema-only-1.9.x.xml";
+	
 	private ChangeLogVersionFinder changeLogVersionFinder;
 	
 	public ChangeLogDetective() {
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/liquibase/ChangeLogDetective.java
```diff
@@ -157,13 +161,22 @@ boolean isVintageChangeSet(String filename, ChangeSet changeSet) {
...
-			for (ChangeSet changeSet : filteredUnrunChangeSets) {
-				log.debug("file '{}' contains un-run change set with id '{}' by author '{}'", filename, changeSet.getId(),
+
+	/**
+	 * Logs un-run change sets no more than a given number and only for the 1.9.x Liquibase snapshots.
+	 * 
+	 * @return a boolean value indicating whether the change sets were logged. The value is used for testing.
+	 */
+	boolean logUnRunChangeSetDetails( String filename, List<ChangeSet> ChangeSets) {
+		if (ChangeSets.size() < MAX_NUMBER_OF_CHANGE_SETS_TO_LOG
```
dkayiwa, line 171: I guess these are usually not many, from the performance perspective. Correct?

WolfSchlegel: Yes, I remembered our discussion about having extra logs at all and took care to keep the number small.

In the case of the reference application there were two un-run change sets only and the max number to be logged is 9. If, for some reason, all of 1.9.x went missing, we still get the error and a single log messages stating the number of un-run change sets.

It took a couple of hours to recreate the bug and confirm what went wrong with the reference application yesterday, this patch is buying as and the community time where we can respond faster to any more findings.

dkayiwa: Fair enough. 👍

PR #3208, opened 2020-05-19
TRUNK-5739: Expand updateOrderFulfillerStatus to also set accession n…

mogoodrich:
…umber

### DIFF
--- a/api/src/main/java/org/openmrs/api/OrderService.java
+++ b/api/src/main/java/org/openmrs/api/OrderService.java
@@ -9,10 +9,6 @@
  */
 package org.openmrs.api;
 
-import java.util.Date;
-import java.util.List;
-import java.util.Locale;
-
 import org.openmrs.CareSetting;
 import org.openmrs.Concept;
 import org.openmrs.ConceptClass;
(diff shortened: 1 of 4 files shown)


mogoodrich:
@mseaton... not 100% sure, but they are able to explicitly set it to the empty string, just not null... seemed like a fair way to allow both...

mseaton:
@mogoodrich how would that work for the FulfillerStatus enum?  I guess I'm just trying to understand how an API user would expect this to work.  If I pass an explicit value to a method called "setXYZ", I expect the value to be set to what I pass in.  If you don't want this to happen in your use case, then you need the calling application to do the right thing.  In this case you should likely handle that in the REST endpoint, not in the service API, right?

mogoodrich:
It's true you can't null out FulfillerStatus, which I thought was okay.

I've updated the javadocs to explicitly mention that if you pass in null as a parameter it doesn't update it.

Perhaps in hindsight a mistake to extend the updateFulfillerStatus method with a new parameter instead of creating a new updateAccessionNumber method. But as it with a single method, if you don't have the null check, then, say, a consumer could never update the accession number for an order without knowing the fulfiller status or fulfiller comment because all three need to be passed in.

PR #3209, opened 2020-05-20
TRUNK-5740 | Initialize lazy properties while caching for True,False and unknown concepts.

mddubey:
This has been tested by putting the `initializeLazyPropertiesForConcept` method in our module activator. After that we didn't face the issue for last two days.

Issue Link: https://issues.openmrs.org/browse/TRUNK-5740
OpenMRS Talk: https://talk.openmrs.org/t/random-lazy-initialization-while-making-encounter-post-with-full-representation-as-response/28427

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
@@ -27,6 +27,7 @@ import java.util.UUID;
 
 import org.apache.commons.beanutils.BeanUtils;
 import org.apache.commons.collections.CollectionUtils;
+import org.hibernate.Hibernate;
 import org.openmrs.Concept;
 import org.openmrs.ConceptAnswer;
 import org.openmrs.ConceptAttribute;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1225,15 +1229,30 @@ private void setBooleanConcepts() {
...
 		catch (NumberFormatException e) {
 			log.warn("Concept ids for boolean concepts should be numbers");
 		}
 	}
-	
+
+	private void initializeLazyPropertiesForConcept(Concept concept) {
+		Hibernate.initialize(concept.getRetiredBy());
+		Hibernate.initialize(concept.getCreator());
+		Hibernate.initialize(concept.getChangedBy());
```
dkayiwa, line 1247: You can get the lazily loaded property from Concept.hbm.xml

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1195,9 +1196,12 @@ public Concept getTrueConcept() {
...
 		if (unknownConcept == null) {
 			try {
-				ConceptServiceImpl.setStaticUnknownConcept(Context.getConceptService().getConcept(
-				    Integer.parseInt(Context.getAdministrationService().getGlobalProperty(
-				        OpenmrsConstants.GLOBAL_PROPERTY_UNKNOWN_CONCEPT))));
+				Concept unknownConcept = Context.getConceptService().getConcept(
+					Integer.parseInt(Context.getAdministrationService().getGlobalProperty(
+						OpenmrsConstants.GLOBAL_PROPERTY_UNKNOWN_CONCEPT)));
+				initializeLazyPropertiesForConcept(unknownConcept);
+
```
dkayiwa, line 1203: Do you think there is any value in the original code, of setting the unknown concept via a new static method, instead of doing it just the same way as for the true and false concepts?

dkayiwa:
Can you also include the ticket id in the commit message as recommended at? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
mddubey: I see in `concept.hbm.xml`, `Creator, ChangedBy and RetiredBy` are not marked as `lazy=true`. However these were throwing the LazyInitException. 
Please have a look at the [screenshot](https://talk.openmrs.org/uploads/default/original/2X/5/51b275ed4e3fb5d27c92cdddcb0b72eca8ea7940.png).

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
mddubey: I feel it would be better to do in the same way as the true and false concepts are done. However I don't know intention of creating a private stating method in the first time itself.
I can make this change if needed.

dkayiwa:
Can you include the ticket id in your commit message?

mddubey:
@dkayiwa 
I have added the card number in the commit message.

Thank you

PR #3215, opened 2020-05-21
Trunk-5738:Allow searching for orders based on Accession Number and O…

mogoodrich:
…rder Number


api/src/test/resources/org/openmrs/include/standardTestDataset.xml
```diff
@@ -357,7 +357,7 @@
...
   <order_type_class_map order_type_id="1" concept_class_id="3"/>
   <orders order_id="1" order_type_id="1" order_number="1" urgency="ROUTINE" order_action="NEW" concept_id="88" orderer="1" instructions="2x daily" date_activated="2008-08-08 00:00:00.0" date_stopped="2008-08-15 00:00:00.0" creator="1" date_created="2008-08-08 00:00:00.0" voided="false" patient_id="7" uuid="921de0a3-05c4-444a-be03-e01b4c4b9142" care_setting="1" encounter_id="3" />
-  <orders order_id="111" order_type_id="1" previous_order_id="1" order_number="111" orderer="1" urgency="ROUTINE" order_action="REVISE" concept_id="88" creator="1" date_activated="2008-08-15 00:00:00.0" date_created="2008-08-15 00:00:00.0" voided="false" patient_id="7" uuid="e1f95924-697a-11e3-bd76-0800271c1b75" care_setting="1" encounter_id="3" />
+  <orders order_id="111" order_type_id="1" previous_order_id="1" order_number="111" orderer="1" urgency="ROUTINE" order_action="REVISE" concept_id="88" creator="1" date_activated="2008-08-15 00:00:00.0" date_created="2008-08-15 00:00:00.0" voided="false" patient_id="7" uuid="e1f95924-697a-11e3-bd76-0800271c1b75" care_setting="1" encounter_id="3" accession_number="ACC-123"/>
```
mogoodrich, line 360: Normally I refrain from changing the standard test data, but this seemed safe.

api/src/main/java/org/openmrs/parameter/OrderSearchCriteria.java
```diff
@@ -81,6 +91,49 @@
 	private final boolean excludeDiscontinueOrders;
 
 	/**
+	 * Instead of calling this constructor directly, it is recommended to use {@link OrderSearchCriteriaBuilder}.
```
mseaton, line 94: Two questions:
- Is this second constructor even needed, if it isn't even the recommended way to use this?
- If it is, it should not be copying all of this setting logic, but instead first calling the other constructor, and then setting the additional properties.

api/src/main/java/org/openmrs/parameter/OrderSearchCriteria.java
```diff
@@ -137,6 +190,14 @@ public OrderSearchCriteria(Patient patient, CareSetting careSetting, Collection<
 	 */
 	public Collection<OrderType> getOrderTypes() { return orderTypes; }
 
+	public String getAccessionNumber() {
```
mseaton, line 193: Would be useful to an add an @since annotation to any new public methods that your are introducing (here and elsewhere in this PR).

mogoodrich: Ah, agreed, will do!

api/src/main/java/org/openmrs/parameter/OrderSearchCriteria.java
mogoodrich: I debated just removing it.  Seemed like it would be safe to remove since people shouldn't be using it directly, but since I'm planning on backporting I figured I'd keep it around to be safe.

But, agreed, I'll change it not to be repetitive... I was thinking that one constructor couldn't call another, but maybe I'm wrong about that...

PR #3221, opened 2020-06-02
TRUNK-5222:Extract Private method in locationServiceImpl.getDefaultLocation

sherrif10:
https://issues.openmrs.org/browse/TRUNK-5222
 Kindly review  thanks
 cc @dkayiwa @ibacher

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/LocationServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/LocationServiceImpl.java
@@ -119,20 +119,10 @@ public class LocationServiceImpl extends BaseOpenmrsService implements LocationS
 		if (StringUtils.hasText(locationGP)) {
 			location = Context.getLocationService().getLocation(locationGP);
 		}
-
-		if (location == null) {
-			location = getDefaultLocation(null, locationGP);
+		  return StringUtils.hasText(locationGP) ? Context.getLocationService().getLocation(locationGP) : getDefaultLocation(null, "");
 		}
-		
-		// If neither exist, get the first available location
-		if (location == null) {
-			location = Context.getLocationService().getLocation(1);
-		}
-		
-		return location;
-	}
-
-	private Location getDefaultLocation(Location location, String locationGP) {
+	
+    private Location getDefaultLocation(Location location, String locationGP) {
 		//Try to look up 'Unknown Location' in case the global property is something else
 		Location result = getDefaultLocationFromSting(location, locationGP, "Unknown Location");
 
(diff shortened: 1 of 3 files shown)


gitcliff:
hello @sherrif10 ,have u seen the travis failure ?

sherrif10:
Yeah am working on it unfortunately failing tests that i encounter seem to be different from the class being tackled on , still checking on it

sherrif10:
Am coming back it to very soon

sherrif10:
Let me fix some failing test

sherrif10:
Refacoring this code broke some tests ,Especially hl7ServiceTest

api/src/main/java/org/openmrs/api/impl/LocationServiceImpl.java
```diff
@@ -118,21 +118,13 @@ public Location getDefaultLocation() throws APIException {
 		
 		if (StringUtils.hasText(locationGP)) {
 			location = Context.getLocationService().getLocation(locationGP);
+		} else if(location == null){
+				location = getDefaultLocation(null, locationGP);	
+			}
+			return location;
```
ibacher, line 124: This is a pretty drastic change in the logic here and I'm not sure this is what we want to do. 

So before your change, this would first check if `locationGP` has text and if it does, try to load that location. Then if `locationGP` didn't have text or the the location specified by `locationGP` couldn't be found, it would try to get the default location by calling the private `getDefaultLocation()` function. If none of those methods returned a location, it would try to get Location 1 (whatever that is) and return it.

With this change, this will check if `locationGP` has text and if it does, it attempts to return that location. If `locationGP` does not have text (the check on `location == null` is actually redundant because at this point, we know `location` is always `null`), it will call the private `getDefaultLocation()` function. The function will then return the result of either `getLocation(locationGP)` (if `locationGP` has text) or the result of `getDefaultLocation(null, locationGP)` (if `locationGP` does not have text.

In other words, I could simplify the logic in your change to something like this:

```java
return StringUtils.hasText(locationGP) ? Context.getLocationService().getLocation(locationGP) : getDefaultLocation(null, "");
```

I don't think there's much to gain from trying to simplify this method much more than it already is... At a certain point, methods get too "simple" and the logic becomes hard to follow.

sherrif10: Haaa thanks for the logical testing and logical clarification,i will try to do some changes as you suggested

sherrif10: How does it look now,Am not so conviced of how this methods were called from the start btw

sherrif10:
Hello @dkayiwa  @ibacher @mks-d    ,could you have a review and guide me more for this, Do we still need it

sherrif10:
We can go ahead and close it, if its needed then we will back to its attention , thanks @ibacher

PR #3230, opened 2020-06-08
FM2-184: Set context classloader to be module's classloader.

LeeBreisacher:
Explicitly set the context classloader to be the module's classloader
before calling willRefreshContext.

## Issue I worked on
https://issues.openmrs.org/browse/FM2-184

- [x] My IDE is configured to follow the [**code style**]

  No? -> I'm not sure how this can be unit tested, tbh. 

### DIFF
--- a/api/src/main/java/org/openmrs/module/ModuleUtil.java
+++ b/api/src/main/java/org/openmrs/module/ModuleUtil.java
@@ -858,6 +858,7 @@ public class ModuleUtil {
 		for (Module module : startedModules) {
 			try {
 				if (module.getModuleActivator() != null) {
+					Thread.currentThread().setContextClassLoader(ModuleFactory.getModuleClassLoader(module));
 					module.getModuleActivator().willRefreshContext();
 				}
 			}


api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -858,6 +858,7 @@ public static AbstractRefreshableApplicationContext refreshApplicationContext(Ab
 		for (Module module : startedModules) {
 			try {
 				if (module.getModuleActivator() != null) {
+					Thread.currentThread().setContextClassLoader(ModuleFactory.getModuleClassLoader(module));
```
dkayiwa, line 861: How about doing it also for the rest of the call backs for the module activator?

LeeBreisacher: hm, good question. There's 3 more calls in the loop at the end of that method. otoh, if it isn't broken...  Note that the OpenmrsClassloader is carefully installed as the context classloader just before that last section. Someone that knows much more about modules and classloaders than I should weigh in.

wluyima: I'm still not sure how this is the solution? A don't think a module classloader should ever be set as the context classloader because that's the role of the OpenmrsClassloader, the problems to be with the module's setup or a module it depends on or the OpenmrsClassloader itself otherwise I smell bad things happening because of this change.

LeeBreisacher: I'm certainly new to this part of OpenMRS, but my reading of the OpenmrsClassLoader code is that it simply holds on to all the "real" classloaders -- the ModuleClassLoaders, and it chooses the right one to do the real work of loading things. The OpenmrsClassLoader doesn't do any actual loading of its own. Except, in some cases (like the one here), the OpenmrsClassloader does not have enough information in hand to be able to choose the right ModuleClassLoader, and it ends up just choosing an arbitrary one (which of course can be the wrong one). So, in this case, we bypass OpenmrsClassLoader and insert the one ModuleClassLoader that we (as the caller) know to be the right one.

dkayiwa: @wluyima based on the above, do you still have objections to merging this? 😊

mddubey: @wluyima @dkayiwa @mks-d 
If there are no more concerns on this PR, can we please merge this PR, As it is blocking our deployments. For development we are able to get our way by removing data-filter for now but we would need it for integration testings.

Thanks

mks-d: @dkayiwa can you validate this PR on the RHS reviewers menu? That'll be one in the box already :-)

@wluyima, any objections on your end?

dkayiwa:
I have gone ahead and merged this because i do not for-see any harm from this change. It is small and logical. Also, the earlier we merge it, the more time we shall have to test it, just in case of any surprises.

api/src/main/java/org/openmrs/module/ModuleUtil.java
wluyima: I still strongly believe the problem is with the module setup or the OpenMRS classloader itself, but am fine with this if it sort of works for now

mddubey:
Hello @dkayiwa 

Could this be also back-ported to `2.3.2-SNAPSHOT`.  That's the version we are targeting in our distribution as of now.
CC: @mks-d

PR #3240, opened 2020-06-11
TRUNK-5655: Added a utility method to get a Concept by it's identifier, name or map

jnsereko:
## Issue I worked on TRUNK-5655

see https://issues.openmrs.org/browse/TRUNK-5655

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/ConceptService.java
+++ b/api/src/main/java/org/openmrs/api/ConceptService.java
@@ -332,6 +332,17 @@ public interface ConceptService extends OpenmrsService {
 	 */
 	@Authorized(PrivilegeConstants.GET_CONCEPTS)
 	public Concept getConcept(String conceptIdOrName) throws APIException;
+
+	/**
+	 * Get the concept by conceptRef where the conceptRef can either be: 1) an integer id like 5090
+	 * 2) a mapping type id like "XYZ:HT" 3) a uuid like "a3e12268-74bf-11df-9768-17cfc9833272" 4) a
+	 * name like "PLATELETS" 5) the fully qualified name of a Java constant which contains one of the above
+	 * @param conceptRef the concept string identifier
+	 * @since 2.6.0
+	 * @return the concept if it exists otherwise null
+	 */
+	@Authorized(PrivilegeConstants.GET_CONCEPTS)
+	public Concept getConceptByReference(String conceptRef);
 	
 	/**
 	 * Get Drug by its UUID
--- a/api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
@@ -9,8 +9,7 @@
  */
 package org.openmrs.api.impl;
 
-import static org.apache.commons.lang3.StringUtils.contains;
-
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -27,6 +26,8 @@ import java.util.UUID;
 
 import org.apache.commons.beanutils.BeanUtils;
 import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.math.NumberUtils;
 import org.hibernate.Hibernate;
 import org.openmrs.Concept;
 import org.openmrs.ConceptAnswer;
@@ -70,7 +71,6 @@ import org.slf4j.LoggerFactory;
 import org.springframework.cache.annotation.CacheEvict;
 import org.springframework.cache.annotation.Cacheable;
 import org.springframework.transaction.annotation.Transactional;
-import org.springframework.util.StringUtils;
 
 /**
  * Default Implementation of ConceptService service layer classes
@@ -286,7 +286,7 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 	 */
 	@Override
 	public Concept retireConcept(Concept concept, String reason) throws APIException {
-		if (!StringUtils.hasText(reason)) {
+		if (StringUtils.isBlank(reason)) {
 			throw new IllegalArgumentException(Context.getMessageSourceService().getMessage("general.voidReason.empty"));
 		}
 		
@@ -335,6 +335,59 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 	@Transactional(readOnly = true)
 	public Concept getConcept(Integer conceptId) throws APIException {
 		return dao.getConcept(conceptId);
+	}
+
+	/**
+	 * @see org.openmrs.api.ConceptService#getConceptByReference(String conceptRef)
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	public Concept getConceptByReference(String conceptRef) {
+		if (StringUtils.isBlank(conceptRef)) {
+			return null;
+		}
+		Concept cpt = null;
+		//check if input is a valid Uuid
+		if (isValidUuidFormat(conceptRef)) {
+			cpt = Context.getConceptService().getConceptByUuid(conceptRef);
+			if (cpt != null) {
+				return cpt;
+			}
+		}
+		//handle mapping
+		int idx = conceptRef.indexOf(":");
+		if (idx >= 0 && idx < conceptRef.length() - 1) {
+			String conceptSource = conceptRef.substring(0, idx);
+			String conceptCode = conceptRef.substring(idx + 1);
+			cpt = Context.getConceptService().getConceptByMapping(conceptCode, conceptSource);
+			if (cpt != null) {
+				return cpt;
+			}
+		}
+		//handle id
+		int conceptId = NumberUtils.toInt(conceptRef, -1);
+		if (conceptId >= 0) {
+			cpt = Context.getConceptService().getConcept(conceptId);
+			if (cpt != null) {
+				return cpt;
+			}
+		} else {
+			//handle name
+			cpt = Context.getConceptService().getConceptByName(conceptRef);
+			if (cpt != null) {
+				return cpt;
+			}
+		}
+		//handle static constant
+		if (conceptRef.contains(".")) {
+			try {
+				return getConceptByReference(evaluateStaticConstant(conceptRef));
+			}
+			catch (APIException e) {
+				log.warn("Unable to translate '{}' into a concept", conceptRef, e);
+			}
+		}
+		return cpt == null ? null : cpt;
 	}
 	
 	/**
@@ -417,7 +470,7 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 	@Override
 	@Transactional(readOnly = true)
 	public Concept getConceptByName(String name) {
-		if (!StringUtils.hasText(name)) {
+		if (StringUtils.isBlank(name)) {
 			return null;
 		}
 		return dao.getConceptByName(name);
@@ -769,7 +822,7 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 		}
 		
 		ConceptName conceptName = null;
-		if (cp.getState().equals(OpenmrsConstants.CONCEPT_PROPOSAL_CONCEPT) || !StringUtils.hasText(cp.getFinalText())) {
+		if (cp.getState().equals(OpenmrsConstants.CONCEPT_PROPOSAL_CONCEPT) || StringUtils.isBlank(cp.getFinalText())) {
 			cp.setState(OpenmrsConstants.CONCEPT_PROPOSAL_CONCEPT);
 			cp.setFinalText("");
 		} else if (cp.getState().equals(OpenmrsConstants.CONCEPT_PROPOSAL_SYNONYM)) {
@@ -1454,7 +1507,7 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 			dao.deleteConceptStopWord(conceptStopWordId);
 		}
 		catch (DAOException e) {
-			if (contains(e.getMessage(), "Concept Stop Word not found or already deleted")) {
+			if (StringUtils.contains(e.getMessage(), "Concept Stop Word not found or already deleted")) {
 				throw new ConceptStopWordException("ConceptStopWord.error.notfound", e);
 			}
 			throw new ConceptStopWordException("general.cannot.delete", e);
@@ -1645,7 +1698,7 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 	@Override
 	public ConceptMapType retireConceptMapType(ConceptMapType conceptMapType, String retireReason) throws APIException {
 		String tmpRetireReason = retireReason;
-		if (!StringUtils.hasText(tmpRetireReason)) {
+		if (StringUtils.isBlank(tmpRetireReason)) {
 			tmpRetireReason = Context.getMessageSourceService().getMessage("general.default.retireReason");
 		}
 		conceptMapType.setRetireReason(tmpRetireReason);
@@ -1717,7 +1770,7 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 		//On addition of extra attributes to concept maps, terms that were generated from existing maps have 
 		//empty string values for the name property, ignore the search when name is an empty string but allow 
 		//white space characters
-		if (!StringUtils.hasLength(name)) {
+		if (StringUtils.isBlank(name)) {
 			return null;
 		}
 		return dao.getConceptReferenceTermByName(name, conceptSource);
@@ -1749,7 +1802,7 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 	public ConceptReferenceTerm retireConceptReferenceTerm(ConceptReferenceTerm conceptReferenceTerm, String retireReason)
 	        throws APIException {
 		String tmpRetireReason = retireReason;
-		if (!StringUtils.hasText(tmpRetireReason)) {
+		if (StringUtils.isBlank(tmpRetireReason)) {
 			tmpRetireReason = Context.getMessageSourceService().getMessage("general.default.retireReason");
 		}
 		conceptReferenceTerm.setRetireReason(tmpRetireReason);
@@ -2002,6 +2055,42 @@ public class ConceptServiceImpl extends BaseOpenmrsService implements ConceptSer
 		return dao.getConceptAttributeCount(conceptAttributeType) > 0;
 	}
 
+	/***
+	 * Determines if the passed string is in valid uuid format By OpenMRS standards, a uuid must be 36
+	 * characters in length and not contain whitespace, but we do not enforce that a uuid be in the
+	 * "canonical" form, with alphanumerics seperated by dashes, since the MVP dictionary does not use
+	 * this format (We also are being slightly lenient and accepting uuids that are 37 or 38 characters
+	 * in length, since the uuid data field is 38 characters long)
+	 */
+	public static boolean isValidUuidFormat(String uuid) {
+		if (uuid.length() < 36 || uuid.length() > 38 || uuid.contains(" ") || uuid.contains(".")) {
+			return false;
+		}
+		return true;
+	}
+
+	/**
+	 * Evaluates the specified Java constant using reflection: if input is org.openmrs.CLASS_NAME.CONSTANT_NAME
+	 * then, output will be CONSTANT_NAME
+	 * @param fqn the fully qualified name of the constant
+	 * @return the constant value or null
+	 */
+	private static String evaluateStaticConstant(String fqn) {
+		int lastPeriod = fqn.lastIndexOf(".");
+		String clazzName = fqn.substring(0, lastPeriod);
+		String constantName = fqn.substring(lastPeriod + 1);
+		try {
+			Class<?> clazz = Context.loadClass(clazzName);
+			Field constantField = clazz.getDeclaredField(constantName);
+			constantField.setAccessible(true);
+			Object val = constantField.get(null);
+			return val != null ? String.valueOf(val) : null;
+		}
+		catch (Exception ex) {
+			throw new APIException("Error while evaluating " + fqn + " as a constant" , ex);
+		}
+	}
+	
 	private List<ConceptClass> getConceptClassesOfOrderTypes() {
 		List<ConceptClass> mappedClasses = new ArrayList<>();
 		AdministrationService administrationService = Context.getAdministrationService();
--- a/api/src/test/java/org/openmrs/api/ConceptServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/ConceptServiceTest.java
@@ -100,7 +100,7 @@ import org.springframework.validation.Errors;
 public class ConceptServiceTest extends BaseContextSensitiveTest {
 	
 	protected ConceptService conceptService = null;
-	
+
 	protected static final String INITIAL_CONCEPTS_XML = "org/openmrs/api/include/ConceptServiceTest-initialConcepts.xml";
 	
 	protected static final String GET_CONCEPTS_BY_SET_XML = "org/openmrs/api/include/ConceptServiceTest-getConceptsBySet.xml";
@@ -111,6 +111,13 @@ public class ConceptServiceTest extends BaseContextSensitiveTest {
 
 	@Autowired
 	CacheManager cacheManager;
+
+	// For testing concept lookups by static constant
+	private static final String TEST_CONCEPT_CONSTANT_ID = "3";
+ 
+	private static final String TEST_CONCEPT_CONSTANT_UUID = "35d3346a-6769-4d52-823f-b4b234bac3e3";
+	
+	private static final String TEST_CONCEPT_CONSTANT_NAME = "COUGH SYRUP";
 	
 	/**
 	 * Run this before each unit test in this class. The "@Before" method in
@@ -2302,8 +2309,10 @@ public class ConceptServiceTest extends BaseContextSensitiveTest {
 		List<Concept> actualConcepts = conceptService.getConceptsByClass(new ConceptClass(3));
 		
 		// verify
-		assertThat(actualConcepts.size(), is(4));
-		assertThat(actualConcepts, containsInAnyOrder(conceptService.getConcept(3), conceptService.getConcept(60), conceptService.getConcept(88), conceptService.getConcept(792)));
+		assertThat(actualConcepts.size(), is(6));
+		assertThat(actualConcepts,
+		    containsInAnyOrder(conceptService.getConcept(3), conceptService.getConcept(60), conceptService.getConcept(64),
+		        conceptService.getConcept(67), conceptService.getConcept(88), conceptService.getConcept(792)));
 	}
 	
 	/**
@@ -2834,7 +2843,7 @@ public class ConceptServiceTest extends BaseContextSensitiveTest {
 	public void getAllConcepts_shouldExcludeRetiredConceptsWhenSetIncludeRetiredToFalse() {
 		final List<Concept> allConcepts = conceptService.getAllConcepts(null, true, false);
 		
-		assertEquals(34, allConcepts.size());
+		assertEquals(36, allConcepts.size());
 		assertEquals(3, allConcepts.get(0).getConceptId().intValue());
 	}
 	
@@ -2845,14 +2854,14 @@ public class ConceptServiceTest extends BaseContextSensitiveTest {
 	public void getAllConcepts_shouldOrderByAConceptField() {
 		List<Concept> allConcepts = conceptService.getAllConcepts("dateCreated", true, true);
 		
-		assertEquals(36, allConcepts.size());
+		assertEquals(38, allConcepts.size());
 		assertEquals(88, allConcepts.get(0).getConceptId().intValue());
 		assertEquals(27, allConcepts.get(allConcepts.size() - 1).getConceptId().intValue());
 		
 		//check desc order
 		allConcepts = conceptService.getAllConcepts("dateCreated", false, true);
 		
-		assertEquals(36, allConcepts.size());
+		assertEquals(38, allConcepts.size());
 		assertEquals(23, allConcepts.get(0).getConceptId().intValue());
 		assertEquals(88, allConcepts.get(allConcepts.size() - 1).getConceptId().intValue());
 	}
@@ -2883,7 +2892,7 @@ public class ConceptServiceTest extends BaseContextSensitiveTest {
 	public void getAllConcepts_shouldOrderByConceptIdAndIncludeRetiredWhenGivenNoParameters() {
 		final List<Concept> allConcepts = conceptService.getAllConcepts();
 		
-		assertEquals(36, allConcepts.size());
+		assertEquals(38, allConcepts.size());
 		assertEquals(3, allConcepts.get(0).getConceptId().intValue());
 	}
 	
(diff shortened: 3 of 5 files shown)


HerbertYiga:
@jnsereko  you have a breaking line in ConceptServiceImpl.java,could you check out your ci logs to solve this

api/src/test/java/org/openmrs/api/ConceptServiceTest.java
```diff
@@ -3531,4 +3535,80 @@ public void getConcepts_shouldPassWithAndOrNotWords() {
...
 		assertThat(searchResults.get(0).getWord(), is("SALBUTAMOL INHALER NOT"));
 	}
+
+	/**
+	 * @see ConceptService#getConceptByNameOrIdOrMap(String)
+	 */
+	@Test
+	public void getConceptByNameOrIdOrMap_shouldFindAConceptByItsConceptName() {
+		String conceptName = "COUGH SYRUP";
+		Assert.assertEquals("3", conceptService.getConceptByNameOrIdOrMap(conceptName).getConceptId().toString());
```
HerbertYiga, line 3545: i am worried of the 3 here,did you set it as a conceptId  any were up for COUGH SYRUP

jnsereko:
> @jnsereko you have a breaking line in ConceptServiceImpl.java,could you check out your ci logs to solve this

Yea sure I am working on it

api/src/test/java/org/openmrs/api/ConceptServiceTest.java
jnsereko: There is already a concept with the name "**COUGH SYRUP**" and its id is "**3**" in the file called [StandardDataSet.xml](https://github.com/openmrs/openmrs-core/blob/master/api/src/test/resources/org/openmrs/include/standardTestDataset.xml)
I decided to use this concept as i was testing the method in getting the concept by name.

jnsereko:
@samuelmale @dkayiwa @mozzy11 
May you please review and comment.

api/src/test/java/org/openmrs/api/ConceptServiceTest.java
jnsereko: hello **herbert**! should i resolve this!

HerbertYiga: that line is ok,i got to see the id in the data set

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1982,6 +1984,72 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	public Concept getConceptByNameOrIdOrMap(String conceptRef) throws APIException {
+		if (conceptRef == null) {
+			return null;
+		}
+		Concept cpt = null;
+		//check if input is a valid Uuid
+		if (!(conceptRef.length() < 36 || conceptRef.length() > 38 || conceptRef.contains(" ") || conceptRef.contains("."))) {
```
HerbertYiga, line 1998: should it be greater than 36 or 38?

ibacher: Ugh... I hate the 36 / 38 thing. The normal display format for UUIDs is 32 digits + 4 dashes (36 characters), e.g. 8-4-4-4-12. However, our database fields allow UUIDs to be up to 38 characters (which corresponds to the Microsoft GUID format, which is the same as the above with curly braces around it, even though we don't use that anywhere).

Maybe this could be simplified into just checking with a regex:

```java
private static final Pattern UUID_FORMAT = Pattern.compile("^\\d{8}-\\d{4}-\\d{4}-\\d{4}-\\d{12}$");

...

if (UUID_FORMAT.matcher(conceptRef).matches()) {
    ...
}
```

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -1923,4 +1923,23 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
+	 * <strong>Should</strong> find a concept by its conceptId
+	 * <strong>Should</strong>  find a concept by its mapping
+	 * <strong>Should</strong> find a concept by its uuid
+	 * <strong>Should</strong> find a concept by its name
+	 * <strong>Should</strong> find a concept by a fully qualified name of a Java constant that contains an id,
+	 * 								uuid, name or mapping
+	 * <strong>Should</strong> return null when ConceptRef does not exist
+	 */
+	@Authorized(PrivilegeConstants.GET_CONCEPTS)
+	public Concept getConceptByNameOrIdOrMap(String conceptRef);
```
ibacher, line 1943: I'd call this `getConceptByReference()`, This should have a `@since` comment in the Javadoc. Also, we don't need the `<should>` stuff, although I would expect these to be included in unit tests.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1982,6 +1984,72 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
 		return dao.getConceptAttributeCount(conceptAttributeType) > 0;
 	}
 
+	/**
+	 * @see org.openmrs.api.ConceptService#getConceptByNameOrIdOrMap(String conceptRef
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	public Concept getConceptByNameOrIdOrMap(String conceptRef) throws APIException {
+		if (conceptRef == null) {
```
ibacher, line 1993: Presumably, this should be something like `StringUtils.isBlank(conceptRef)`, which handles both `null` and the case where `conceptRef` is only whitespace.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1982,6 +1984,72 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+			return null;
+		}
+		Concept cpt = null;
+		//check if input is a valid Uuid
+		if (!(conceptRef.length() < 36 || conceptRef.length() > 38 || conceptRef.contains(" ") || conceptRef.contains("."))) {
+			cpt = Context.getConceptService().getConceptByUuid(conceptRef);
+			if (cpt != null)
+				return cpt;
```
ibacher, line 2001: Every `if` should have curly brackets.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1982,6 +1984,72 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+		}
+		//check if input is map
+		if (conceptRef.indexOf(":") != -1) {
+			int idx = conceptRef.indexOf(":");
```
ibacher, line 2005: Why are we calling `.indexOf()` twice?

```java
int idx = conceptRef.indexOf(":");
if (idx >= 0 && idx < conceptRef.length - 1) {
    String conceptSource = conceptRef.substring(0, idx);
    String conceptCode = conceptRef.substring(idx + 1);
}
```

Also note the extra safety check to avoid the case where conceptRef is something like `CIEL:`.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1982,6 +1984,72 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+			String conceptSource, conceptCode;
+			conceptCode = conceptRef.substring(idx + 1);
+			conceptSource = conceptRef.substring(0, idx);
+			cpt = Context.getConceptService().getConceptByMapping(conceptCode, conceptSource);
+			if (cpt != null)
+				return cpt;
+		}
+		//check if input is a name or id
+		if (!conceptRef.isEmpty()) {
```
ibacher, line 2014: This is redundant or, at the very least, should be done *before* we get here, since none of these tests are very functional if the `conceptRef` is empty.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1982,6 +1984,72 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+			//handle id
+			if (NumberUtils.isParsable(conceptRef)) {
+				int conceptId = Integer.valueOf(conceptRef);
```
ibacher, line 2017: Again, this parses the string twice for no real value.

```java
int conceptId = NumberUtils.toInt(conceptRef, -1);
if (conceptId >= 0) {
    cpt = Context.getConceptService().getConcept(conceptId);
}
```

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1982,6 +1984,72 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+				cpt = Context.getConceptService().getConcept(conceptId);
+			} else
+				//handle name
+				cpt = Context.getConceptService().getConceptByName(conceptRef);
```
ibacher, line 2021: Always have curly brackets surrounding an `if` or `else` statement.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1982,6 +1984,72 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+			if (cpt != null)
+				return cpt;
+		}
+		if (conceptRef.contains(".")) {
+			return getConceptByNameOrIdOrMap(evaluateStaticConstant(conceptRef));
+		}
```
ibacher, line 2027: Why are we adding this functionality? This doesn't actually seem very useful. What is the use-case where I know that the concept I want is stored in `org.openmrs.my.package.MyGreatClass.MY_GREAT_CONCEPT` but I can't just call `Context.getConceptService().getConceptByConceptReference(MyGreatClass.MY_GREAT_CONCEPT)`?

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
jnsereko: Wow, Thank you Ivan!  The information you have given has even made it understand it the more

api/src/main/java/org/openmrs/api/ConceptService.java
jnsereko: Thank you @ibacher 
> although I would expect these to be included in unit tests.

The way i am understanding this is that i should remove the **<strong>Should</strong>** statements but persist the corresponding unit tests.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
jnsereko: Yea, Sure this will be better.  it will even eliminate the code below `if (!conceptRef.isEmpty())`

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
jnsereko: Understood! thanks @ibacher

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
jnsereko: Ohhh Yes! this is really good!

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
jnsereko: This must be done at the start

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
jnsereko: You are magical!

api/src/main/java/org/openmrs/api/ConceptService.java
gitcliff: yes, the corresponding unit tests should be maintained

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
jnsereko: Hello @ibacher! i am using java 8, and it doesn't have the method isBank(). Should i install java 11, which supports isBlank() method, but i fear messing up my installations of OpenMRS.  
how about i do it the old way.
```suggestion
		if (conceptRef.trim().isEmpty()) {
```
       .....
}

ibacher: @jnsereko I was suggesting using the method from [`StringUtils` in commons-lang](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html), which is already included in OpenMRS.

jnsereko: Thank you @ibacher i have tool a look at the its documentation.  it is really good.

jnsereko: Thanks @ibacher for the guidance,  if you take a look at 
[ConceptServiceImpl.java](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java#L71) on line 71, you will see that the `StringUtils` class is imported using the statement `import org.springframework.util.StringUtils;`  But the version of this version does not define the `isBlank(String)` method we need.   
I tried importing the version of [StringUtils](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html) using the statement `import org.apache.commons.lang3.StringUtils;`  The problem is that this version of the class provides the required `IsBlank()` method needed but does not define the `hasText(String)`
having both the importation statements is also causing conflict..
I have decided to use `if (StringUtils.trimWhitespace(conceptRef).isEmpty()) { ... }`
Is this ok with you, or do you have any other better option?

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
jnsereko: Hello @ibacher @HerbertYiga Thanks for the support.  
I tried to implement @ibacher's suggestion of using a simpler UUID check  but i failed because i got some conflicts which i tried to solve, but failed.  
How i wish we leave the UUID as it was.  siince i have commented, it will be understood by all those that will use the method.  
cc @dkayiwa @samuelmale

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -24,9 +25,11 @@
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.regex.Pattern;
```
jnsereko, line 28:
```suggestion
```

jnsereko, line 28:
```suggestion
```

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
jnsereko: Hello @ibacher!   If don't use the method `evaluateStaticConstant(String)` and try to get a concept by static constant (with reference to `org.openmrs.my.package.MyGreatClass.MY_CLEAR_CONCEPT`) like;

> if (conceptRef.contains(".")) {
			return getConceptByReference(MyNiceClass.MY_CLEAR_CONCEPT);
     }
> 
i get a `java.lang.NoClassDefFoundError: org/apache/logging/log4j/core/Appender` Error.  Actually, the method `evaluateStaticConstant(String)` loads the class path and retrieves the required concept.

ibacher: @jnsereko Ok, but I'm unclear on why were trying to load something given the string "org.openmrs.my.package.MyGreatClass.MY_GREAT_CONCEPT" at all. Is there some use case for this?

jnsereko: Hello @ibacher! thank you for your support.  The truth is that i haven't seen it being used anywhere, except that the method,[ HtmlFormEntryUtil#getConcept (id)](https://github.com/openmrs/openmrs-module-htmlformentry/blob/master/api/src/main/java/org/openmrs/module/htmlformentry/HtmlFormEntryUtil.java#L564-L605) (which the description on the issue pointed to) had this functionality.  As a result, i thought that we could need it here.  
cc @samuelmale @bmamlin @dkayiwa

ibacher: @jnsereko Thanks! That makes things very clear!

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1981,6 +1985,77 @@ public ConceptAttribute getConceptAttributeByUuid(String uuid) {
...
+		} else {
+			//handle name
+			cpt = Context.getConceptService().getConceptByName(conceptRef);
+			if (cpt != null) {
+				return cpt;
+			}
+		}
+		if (conceptRef.contains(".")) {
+			return getConceptByReference(evaluateStaticConstant(conceptRef));
+		}
```
ibacher, line 2034:
```suggestion
		if (conceptRef.contains(".")) {
		    try {
			    return getConceptByReference(evaluateStaticConstant(conceptRef));
			} catch (IllegalArgumentException e) {
			    log.error("Error while evaluating '{}' as a constant", conceptRef, e);
			}
		}
```

jnsereko: @ibacher!  Thsnk you so much for making me realize that sometimes the static constant might be faulty.   And hence the `try-catch` is very necessary.

jnsereko: Thank you @ibacher!  But i think we have a similar functionality in method `evaluateStaticConstant(String)` though we catch `Exception`.  I think any Error that occurs while Loading and retrieving the concept should shows that the static constant passed to the method was not correct, e.g miss-spelled,   

      private static String evaluateStaticConstant(String fqn) {
		......
		try {
			Class<?> clazz = Context.loadClass(clazzName);
			Field constantField = clazz.getField(constantName);
			Object val = constantField.get(null);
			return val != null ? String.valueOf(val) : null;
		}
		catch (Exception ex) {
			throw new IllegalArgumentException("Error while evaluating '{}' as a constant " + fqn, ex);
		}
	}

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -2057,7 +2052,7 @@ private static String evaluateStaticConstant(String fqn) {
 			return val != null ? String.valueOf(val) : null;
 		}
 		catch (Exception ex) {
-			throw new IllegalArgumentException("Unable to evaluate " + fqn, ex);
+			throw new IllegalArgumentException("Error while evaluating '{}' as a constant " + fqn, ex);
```
jnsereko:
```suggestion
			throw new IllegalArgumentException("Error while evaluating '{}' as a constant ", fqn, ex);
```

gitcliff:
@jnsereko  can you please fix the travis failure and we get this ready for final reviews and merge ...thanks

jnsereko:
Hello @gitcliff! sorry for taking so long on this,  I have been haveing problems with my IDE

HerbertYiga:
What is blocking you with your IDE

jnsereko:
@herbert My IDE hard refused to test `ConceptTestClass`  Using JUnit. this happened after   deleting the .metadata folder in my workplace because the IDE hard refused to create a maven-archetype Project.  But i have now used the cmd for testing.  I think nothing will fail again.  Unless, otherwise

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -1923,4 +1923,17 @@ public Drug getDrugByMapping(String code, ConceptSource conceptSource,
...
 	boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType);
 	
+	/**
+	 * Get the concept by conceptRef where the conceptRef can either be: 1) an integer id like 5090
+	 * 2) a mapping type id like "XYZ:HT" 3) a uuid like "a3e12268-74bf-11df-9768-17cfc9833272" 4) a
+	 * name like "PLATELETS" 5) the fully qualified name of a Java constant that contains one of
+	 * above
+	 * 
+	 * @param conceptRef the concept string identifier
+	 * @since 2.4.0
```
dkayiwa, line 1933: Isn't this version already released?

jnsereko:
@dkayiwa i have improved the method description and also changed the version to 2.5.0

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1982,6 +1990,82 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+			return null;
+		}
+		Concept cpt = null;
+		//check if input is a valid Uuid
+		if (UUID_FORMAT.matcher(conceptRef).matches()) {
+			cpt = dao.getConceptByUuid(conceptRef);
+			if (cpt != null) {
+				return cpt;
+			}
+		}
```
ibacher, line 2009: I would use the same check as is used in the HTML form entry to determine whether or not something is a UUID, namely this:

```java
	/***
	 * Determines if the passed string is in valid uuid format By OpenMRS standards, a uuid must be 36
	 * characters in length and not contain whitespace, but we do not enforce that a uuid be in the
	 * "canonical" form, with alphanumerics seperated by dashes, since the MVP dictionary does not use
	 * this format (We also are being slightly lenient and accepting uuids that are 37 or 38 characters
	 * in length, since the uuid data field is 38 characters long)
	 */
	public static boolean isValidUuidFormat(String uuid) {
		if (uuid.length() < 36 || uuid.length() > 38 || uuid.contains(" ") || uuid.contains(".")) {
			return false;
		}
		
		return true;
	}
```

The reason is not every UUID in OMRS is a "canonical" UUID (in particular, CIEL uses 123456AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA as UUIDs and someone somewhere may still be using old fashioned GUIDs).

jnsereko: @ibacher Thanks for bringing this up.  I thought about using 		
```
if (uuid.length() < 36 || uuid.length() > 38 || uuid.contains(" ") || uuid.contains(".")) {
```
but it was considered as an overkill by @samuelmale

jnsereko:
@ibacher I really appreciate your efforts to resolve this. My machine gotten a hardware problem with the monitor. Am going to solve all Travis errors in the next 24 hours. 
Thank you so much.

jnsereko:
@ibacher I really appreciate your efforts to resolve this. My machine gotten a hardware problem with the monitor. Am going to solve all Travis errors in the next 24 hours. 
Thank you so much.

jnsereko:
@ibacher I really appreciate your efforts to resolve this. My machine gotten a hardware problem with the monitor. Am going to solve all Travis errors in the next 24 hours. 
Thank you so much.

jnsereko:
@ibacher Thanks for taking time looking at this. I have made some changes.  I request for review.
cc @dkayiwa @ibacher

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -332,6 +332,61 @@ public void purgeDrug(Drug drug) throws APIException {
...
+		if (cpt != null) {
+			return cpt;
+		}
+
+		if (conceptRef.contains(".")) {
+			try {
+				return Context.getConceptService().getConceptByReference(evaluateStaticConstant(conceptRef));
+			}
+			catch (IllegalArgumentException e) {
+				log.error("Error while evaluating '{}' as a constant", conceptRef, e);
```
dkayiwa, line 384: How will the caller know that something went wrong?

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1980,6 +2035,43 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+		return true;
+	}
+
+	/**
+	 * Evaluates the specified Java constant using reflection
+	 * 
+	 * @param fqn the fully qualified name of the constant
+	 * @return the constant value
+	 */
+	private static String evaluateStaticConstant(String fqn) {
```
dkayiwa, line 2059: What kind of evaluation is this? In your javadoc, can you give some examples of input and output values?

api/src/test/java/org/openmrs/api/ConceptServiceTest.java
```diff
@@ -102,6 +102,12 @@
 
 	protected static final String CONCEPT_ATTRIBUTE_TYPE_XML = "org/openmrs/api/include/ConceptServiceTest-conceptAttributeType.xml";
 
+	// For testing concept lookups by static constant
+	public static final int TEST_CONCEPT_CONSTANT_ID = 3;
```
dkayiwa, line 106: Why are these public? Are you using them anywhere?

api/src/test/java/org/openmrs/api/ConceptServiceTest.java
```diff
@@ -3520,4 +3528,78 @@ public void getConcepts_shouldPassWithAndOrNotWords() {
...
+		String conceptName = "COUGH SYRUP";
+		assertEquals("3", conceptService.getConceptByReference(conceptName).getConceptId().toString());
+	}
+	
+	/**
+	 * @see ConceptService#getConceptByReference(String)
+	 */
+	@Test
+	public void getConceptByReference_shouldFindAConceptByItsConceptId() {
+		String conceptId = "3";
```
dkayiwa, line 3546: Is this a constant?

jnsereko: The constant is an int. Wanted to bring out that the ID is of type int.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
ibacher: This is pretty much taken as-is from the HTMLFormEntry version. In particular, this is meant to deal with [HTML-524](https://issues.openmrs.org/browse/HTML-524).

api/src/test/java/org/openmrs/api/ConceptServiceTest.java
dkayiwa: Can you take a look at this? https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-NamingConventions

dkayiwa:
@jnsereko can you also take a look at the above build failure?

sherrif10:
@jnsereko  We requesting to address the above comments. Thanks.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
jnsereko: @dkayiwa I have been updating all my pending issues, but honestly speaking, I have failed to understand what is required here.
cc @tendomart @ibacher

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
jnsereko: @ibacher @dkayiwa Should I eliminate the functionality of getting concept using static constant.

jnsereko:
@sherrif10 I think i just need to come to a resolution about 
```
  catch (IllegalArgumentException e) {
  log.error("Error while evaluating '{}' as a constant", conceptRef, e);
```
unless otherwise
cc @ibacher @dkayiwa @tendomart

sherrif10:
cc @jnsereko  Could we work on merge conflicts.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1980,6 +2033,42 @@ public boolean hasAnyConceptAttribute(ConceptAttributeType conceptAttributeType)
...
+		String constantName = fqn.substring(lastPeriod + 1);
+		try {
+			Class<?> clazz = Context.loadClass(clazzName);
+			Field constantField = clazz.getDeclaredField(constantName);
+			constantField.setAccessible(true);
+			Object val = constantField.get(null);
+			return val != null ? String.valueOf(val) : null;
+		}
+		catch (Exception ex) {
+			throw new IllegalArgumentException("Unable to evaluate " + fqn, ex);
```
ibacher, line 2068: It's probably better to throw an APIException here.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -332,6 +332,59 @@ public void purgeDrug(Drug drug) throws APIException {
...
+	/**
+	 * @see org.openmrs.api.ConceptService#getConceptByReference(String conceptRef)
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	public Concept getConceptByReference(String conceptRef) {
+		if (StringUtils.isBlank(conceptRef)) {
+			return null;
+		}
+		Concept cpt;
```
ibacher, line 345: Probably best to initialise this to `null`

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
ibacher: I think this is mostly fine, but I'd move the logging level to `warn` rather than `error`. This method is kind of speculative (try to find a way to figure out what concept this is) so it's not *really* an error if it's not a static reference. Also, the message logged here seems like it would be a better message to attach to the exception.

@dkayiwa I don't know that there's value in throwing the exception from here. In most cases, this method would probably be called to see if a string can (somehow) be resolved to a concept and if it can't, `null` seems like an appropriate value. An error most likely only indicates that a string like "MAP:1234" is not a Java constant as well as not matching any of the previous exit conditions.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -332,6 +332,59 @@ public void purgeDrug(Drug drug) throws APIException {
...
+		//handle static constant
+		if (conceptRef.contains(".")) {
+			try {
+				return getConceptByReference(evaluateStaticConstant(conceptRef));
+			}
+			catch (IllegalArgumentException e) {
+				log.error("Error while evaluating '{}' as a constant", conceptRef, e);
+			}
+		}
+		return cpt;
```
ibacher, line 386: Might be best to return `null` here explicitly?

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
dkayiwa: Sounds fair enough! 😊

jnsereko: I have made more changes and switched to `log.warn` and many more

Thanks @ibacher for the detailed explanation..

ibacher (approved):

PR #3152, opened 2020-03-22
TRUNK-5724: Refactor getSimilarPeople to use Lucene

fruether:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5724

which is a sub-task of https://issues.openmrs.org/browse/TRUNK-5680 to make everything deployable easier and introduce smaller changes

The talk thread discussing approach logic and so on can be found here: https://talk.openmrs.org/t/soundex-search-in-lucenequeries/26068
Please have a look there before asking for my changes!

## Checklist: I completed these to help reviewers :)

- [X] My pull request only contains **ONE single commit**
(the number above, next to the 'Commits' tab is 1).

  No? -> [read here](https://wiki.openmrs.org/display/docs/Pull+Request+Tips) on how to squash multiple commits into one

- [X] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

  No? Unsure? -> [configure your IDE](https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE), format the code and add the changes with `git add . && git commit --amend`

- [X ] I have **added tests** to cover my changes. (If you refactored
  existing code that was well tested you do not have to add tests)

  No? -> write tests and add them to this commit `git add . && git commit --amend`

- [X] I ran `mvn clean package` right before creating this pull request and
  added all formatting changes to my commit.

  No? -> execute above command

- [X ] All new and existing **tests passed**.

  No? -> figure out why and add the fix to your commit. It is your responsibility to make sure your code works.

- [ ] My pull request is **based on the latest changes** of the master branch.

  No? Unsure? -> execute command `git pull --rebase upstream master`

PR #3241, opened 2020-06-11
TRUNK-5746: Refactor getSimilar to use Lucene for 3 names

fruether:

This PR is a successor of https://github.com/openmrs/openmrs-core/pull/3152 now targeting the 3 names component as well.  Please have a look to is before starting the review.

A discussion emerges in in case you need more details regarding the implementation strategy: https://talk.openmrs.org/t/soundex-search-in-lucenequeries/26068/37?u=fruether

@ibacher In case you start testing may use this PR since it as well contains the 3-names search functionality.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5746

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePersonDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePersonDAO.java
@@ -148,6 +148,30 @@ public class HibernatePersonDAO implements PersonDAO {
 		return people;
 	}
 	
+	
+	/**
+	 * This method executes a Lucene search on persons based on the soundex filter with three name elements given
+	 *
+	 * @param name1 basically represents the first name to be searched for in a person name
+	 * @param name2 basically represents the middle name to be searched for in a person name	 
+	 * @param name3 basically represents the family name to be searched for in a person name
+	 * @param birthyear the birthyear the searched person should have 
+	 * @param includeVoided true if voided person should be included 
+	 * @param gender of the person to search for 
+	 * @return the set of Persons that match the search criteria 
+	 */
+	private Set<Person> executeSoundexThreePersonNamesQuery(String name1, String name2, String name3, Integer birthyear, boolean includeVoided , String gender) {
+		PersonLuceneQuery personLuceneQuery = new PersonLuceneQuery(sessionFactory);
+		int maxResults = HibernatePersonDAO.getMaximumSearchResults();
+		LinkedHashSet<Person> people = new LinkedHashSet<>();
+		
+		LuceneQuery<PersonName> luceneQuery = personLuceneQuery.getSoundexPersonNameSearchOnThreeNames(name1, name2, name3, birthyear, false, gender);;
+		ListPart<Object[]> names = luceneQuery.listPartProjection(0, maxResults, "person.personId");
+		names.getList().forEach(x -> people.add(getPerson((Integer) x[0])));
+		
+		return people;
+	}	
+	
 	/**
 	 * This method executes a search on either firstName or lastName attributes of person
 	 * 
@@ -191,31 +215,12 @@ public class HibernatePersonDAO implements PersonDAO {
 		name = name.replace(", ", " ");
 		String[] names = name.split(" ");
 		
-		StringBuilder q = new StringBuilder(
-		        "select p from Person p left join p.names as pname where p.personVoided = false and pname.voided = false and ");
-		
 		if (names.length == 1) {
 			return  executeSoundexOnePersonNameQuery(name, birthyear, false, gender);
 		} else if (names.length == 2) {
 			return executeSoundexTwoPersonNameQuery(names[0], names[1], birthyear, false, gender);
 		} else if (names.length == 3) {
-			q.append("(").append(" case").append("  when pname.givenName is null then 0").append(
-			    "  when soundex(pname.givenName) = soundex(:n1) then 3").append(
-			    "  when soundex(pname.givenName) = soundex(:n2) then 2").append(
-			    "  when soundex(pname.givenName) = soundex(:n3) then 1").append("  else 0 ").append(" end").append(" + ")
-			        .append(" case").append("  when pname.middleName is null then 0").append(
-			            "  when soundex(pname.middleName) = soundex(:n1) then 2").append(
-			            "  when soundex(pname.middleName) = soundex(:n2) then 3").append(
-			            "  when soundex(pname.middleName) = soundex(:n3) then 1").append("  else 0").append(" end").append(
-			            " + ").append(" case").append("  when pname.familyName is null then 0").append(
-			            "  when soundex(pname.familyName) = soundex(:n1) then 1").append(
-			            "  when soundex(pname.familyName) = soundex(:n2) then 2").append(
-			            "  when soundex(pname.familyName) = soundex(:n3) then 3").append("  else 0").append(" end").append(
-			            " +").append(" case").append("  when pname.familyName2 is null then 0").append(
-			            "  when soundex(pname.familyName2) = soundex(:n1) then 1").append(
-			            "  when soundex(pname.familyName2) = soundex(:n2) then 2").append(
-			            "  when soundex(pname.familyName2) = soundex(:n3) then 3").append("  else 0").append(" end").append(
-			            ") >= 5");
+			return executeSoundexThreePersonNamesQuery(names[0], names[1], names[2], birthyear, false, gender);
 		} else {
 			
 			// This is simply an alternative method of name matching which scales better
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/openmrs/Person.java
```diff
@@ -59,6 +62,9 @@
 	@Field
 	private String gender;
 	
+
+	@Field(analyze = Analyze.YES)
+	@DateBridge(encoding = EncodingType.STRING, resolution = Resolution.YEAR)
```
ibacher, line 67: In any case, as a general rule, we'd actually want to store the resolution of the birth date as a DAY rather than a YEAR (even for the purposes of `getSimilarPeople()`, the resolution isn't quite a year, but a fuzzy "3 year period either side of this given year).

ibacher: I can see we do explicitly need a year. If we cannot construct a year query in Lucene, I think we should add a new calculated Lucene-only field of birthyear or something.

dkayiwa:
@fruether did you see the merge conflicts?

fruether:
Yeah I have to do a pull rebase. I will do this soon.
.

dkayiwa <notifications@github.com> schrieb am Di., 21. Juli 2020, 21:54:

.github/workflows/stale.yml
```diff
@@ -12,7 +12,6 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         days-before-stale: 150
         days-before-close: 30
-        debug-only: true
```
fruether, line 15: Must have happened by git pull rebase....

fruether:
@dkayiwa I fixed the merge conflicts by git pull rebase now!

gitcliff:
@dkayiwa we fixed the merge conflicts as you had requested ,,kindly requesting for your review ..Thanks 
cc @ibacher

api/src/main/java/org/openmrs/api/db/hibernate/PersonLuceneQuery.java
```diff
@@ -112,6 +112,28 @@ public PersonLuceneQuery(SessionFactory sessionFactory) {
...
 	 */
+	public LuceneQuery<PersonName> getSoundexPersonNameSearchOnThreeNames(String n1, String n2, String n3,  Integer birthyear, boolean includeVoided, String gender) {
+		String threeNameQuery = "";
+		String givenNameMatchCondition = "(givenNameSoundex:n1^6 OR givenNameSoundex:n2^2 OR givenNameSoundex:n3)";
+		String middleNameMatchCondition = "(middleNameSoundex:n1^2 OR middleNameSoundex:n2^6 OR middleNameSoundex:n3^1)";
+		String familyNameMatchCondition = "(familyNameSoundex:n1^1 OR familyNameSoundex:n2^2 OR familyNameSoundex:n3^6)";
+		String familyName2MatchCondition = "(familyName2Soundex:n1^1 OR familyName2Soundex:n2^2 OR familyName2Soundex:n3^6)";
+		
+		threeNameQuery = givenNameMatchCondition+ " OR " + middleNameMatchCondition + " OR " + familyNameMatchCondition + " OR " + familyName2MatchCondition;
+ 		threeNameQuery = "(" +  threeNameQuery.replace("n1", n1).replace("n2", n2).replace("n3", n3) + ")";
```
ibacher, line 123: It seems to me we should at least run `LuceneQuery.escapeQuery()` on these parameters. But is there a way we could do this more cleanly, i.e., by passing these as paramters instead of a string replace operation?

If not, we should at least have this query stored as a constant string since we really just need to do some replacements right before it's executed.

fruether: What I could do is to directly replace in the query definition the n1, n2, n3 parameter with the correct value but this would made the code less readable I guess.

I think I would prefer to make it a constant and then replacing the values. 

Can you elaborate a bit more on what your thoughts are on "passing these as parameters"? 

Anyways, the escapeQuery can be added easily. I will probably do that till next Sunday!
Thanks!

brandones:
Hi @fruether ! Looks like this is really close. Are you available to finish it?

fruether:
@brandones I think I won't be able to do before the next weekend! Feel free to update the PR if you got time before.

I am also a bit in doubt how I should interpret and implement the feedback of @ibacher . Any hints?

fruether:
In addition, the *else* block as well has to be adapted

ibacher:
@fruether Oh! Sorry I thought what you seemed to propose in the comment was right on. What I was thinking about in terms of "parameters" is the way we usually handle SQL queries (or, rather, the way we *should* handle SQL queries), e.g.,

```sql
SELECT *
FROM Patients
WHERE firstName = :firstName
```

Which ends up getting called something like this:

```java
em.createNativeQuery(sql).setParameter("firstName", "bob");
```

Since we integrate [Hibernate Search](http://hibernate.org/search/), it seemed to me we could do something similar using it's [querying](https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#search-query) functionality, but then I thought better of it, since the query that you have is probably fairly hard to replicate through an API. In which case, just doing:


```java
query.replace("n1", LuceneQuery.escapeQuery(n1))
	.replace("n2", LuceneQuery.escapeQuery(n2))
	.replace("n3", LuceneQuery.esacpeQuery(n3));
```

Is probably fine. Nevertheless, I would like to see the query stored as a constant so we're minimizing the number of string allocations, especially when the bulk of the query is going to be the same. Something like:

```java
private static final String THREE_NAME_QUERY;

static {
	THREE_NAME_QUERY = "((givenNameSoundex:n1^6 OR givenNameSoundex:n2^2 OR givenNameSoundex:n3) OR " +
		"(middleNameSoundex:n1^2 OR middleNameSoundex:n2^6 OR middleNameSoundex:n3^1) OR " +
		"(familyNameSoundex:n1^1 OR familyNameSoundex:n2^2 OR familyNameSoundex:n3^6) OR " +
		"(familyName2Soundex:n1^1 OR familyName2Soundex:n2^2 OR familyName2Soundex:n3^6))";
}
```

fruether:
@ibacher alright got you point now. Thanks for the clarification. 
I will implement it either this or next weekedn!

ibacher:
@fruether Great! And thanks for all your work on these queries! It'll be wonderful to have them in Lucene.

fruether:
@ibacher I think I have applied your feedback! Can you check if you go a quite minute

gitcliff:
@ibacher  does look well now ?

fruether:
@ibacher are we ready to go with this PR?

api/src/main/java/org/openmrs/api/db/hibernate/PersonLuceneQuery.java
```diff
@@ -112,6 +117,26 @@ public PersonLuceneQuery(SessionFactory sessionFactory) {
 	 * @param gender the gender of the person to search  
 	 * @return the LuceneQuery that returns Persons with a soundex representation of the firstName and other defined search criteria
 	 */
+	public LuceneQuery<PersonName> getSoundexPersonNameSearchOnThreeNames(String n1, String n2, String n3,  Integer birthyear, boolean includeVoided, String gender) {
+		
+		String threeNameQuery =  THREE_NAME_QUERY.replace("n1", LuceneQuery.escapeQuery(n1))
+			.replace("n2", LuceneQuery.escapeQuery(n2))
+			.replace("n3", LuceneQuery.escapeQuery(n3));
+		
+		threeNameQuery = "(" +  threeNameQuery.replace("n1", n1).replace("n2", n2).replace("n3", n3) + ")";
```
ibacher, line 126: The calls to `replace()` here are redundant aren't they? I.e., we already did this substitution above.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePersonDAO.java
```diff
@@ -202,23 +226,7 @@ public void setSessionFactory(SessionFactory sessionFactory) {
 		} else if (names.length == 2) {
```
ibacher, line 226: Some comments on this function now:

* We seem to have eliminated the need for the `result` variable at all, so that should probably be removed.
* The line `StringBuilder q = ...` should be moved to the final `else` block, since that's the only place it's necessary.
* Either the final `else` block should be removed or everything after the end of the final `else` block should be added into the final `else` block as that common logic only applies when we continue to support a case for multiple names.

ibacher (commented):
Last question: how does this rather simpler query compare with the rather more complex queries created for one and two names?

api/src/main/java/org/openmrs/api/db/hibernate/PersonLuceneQuery.java
fruether: You are right

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePersonDAO.java
fruether: Make sense! I am planing to also refactor the else block. But for nut it should be good to move everything in it!

fruether:
> Last question: how does this rather simpler query compare with the rather more complex queries created for one and two names?

@ibacher: 
That's a good question. I though about that as well. The difference is: This query does not replicate the result 1:1 whereas the approach of the 2 name query replicates the result identically. 
I could create a sub-ticket to investigate the same approach used here for the 2 name query. Then we can see if the result is sufficient in terms of ranking. 
In terms of the divergence for this query. This is documented on talk: https://talk.openmrs.org/t/soundex-search-in-lucenequeries/26068/44?u=fruether

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePersonDAO.java
fruether: Fixed all these issues in the new commit!

ibacher:
@fruether This looks good to me, but can you look into the Git conflicts?

fruether:
@ibacher the merge conflicts are solved now!

fruether:
@ibacher the conflicts are fixed now. My suggestion would be a) We are merging this PR b) we are having a look if we can simplify the 2 names query in a subtask and discussion c) Last but not least I create a subticket to work on the else branch. 

What do you think about that?

ibacher:
@fruether Sounds good to me!

PR #3266, opened 2020-06-19
TRUNK-5665: User should not extend extend BaseOpenmrsMetadata

jwnasambu:
I changing User to extend BaseOpenmrsObject from BaseOpenmrsMetadata   and added the Auditable and Retireable interfaces to implement. Then added the following fields : creator, dateCreated, changedBy, dateChanged, retiredBy, retireReason, retired, and dateRetired.

## Issue I worked on
https://issues.openmrs.org/browse/TRUNK-5665

### DIFF
--- a/api/src/main/java/org/openmrs/User.java
+++ b/api/src/main/java/org/openmrs/User.java
@@ -12,6 +12,7 @@ package org.openmrs;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -35,14 +36,13 @@ import org.slf4j.LoggerFactory;
  * key-value pairs for either quick info or display specific info that needs to be persisted (like
  * locale preferences, search options, etc)
  */
-public class User extends BaseChangeableOpenmrsMetadata implements java.io.Serializable, Attributable<User> {
+public class User extends BaseOpenmrsObject implements java.io.Serializable, Attributable<User>, Auditable, Retireable {
 	
-	public static final long serialVersionUID = 2L;
+	public static final long serialVersionUID = 2L ;
 	
 	private static final Logger log = LoggerFactory.getLogger(User.class);
 	
 	// Fields
-	
 	private Integer userId;
 	
 	private Person person;
@@ -60,6 +60,22 @@ public class User extends BaseChangeableOpenmrsMetadata implements java.io.Seria
 	private List<Locale> proficientLocales = null;
 	
 	private String parsedProficientLocalesProperty = "";
+	
+	private User creator;
+	
+	private Date dateCreated;
+	
+	private User changedBy;
+	
+	private Date dateChanged;
+	
+	private boolean retired;
+	
+	private User retiredBy;
+	
+	private Date dateRetired;
+	
+	private String retireReason;
 	
 	// Constructors
 	
@@ -594,4 +610,88 @@ public class User extends BaseChangeableOpenmrsMetadata implements java.io.Seria
 		setUserId(id);
 	}
 	
+	@Override
+	public User getCreator() {
+		return creator;
+	}
+
+	@Override
+	public void setCreator(User creator) {
+		this.creator = creator;
+	}
+	
+    @Override
+	public Date getDateCreated() {
+		return dateCreated;
+	}
+
+	@Override
+	public void setDateCreated(Date dateCreated) {
+		this.dateCreated = dateCreated;
+	}
+
+	@Override
+	public User getChangedBy() {
+		return changedBy;
+	}
+
+	@Override
+	public void setChangedBy(User changedBy) {
+		this.changedBy = changedBy;
+	}
+
+	@Override
+	public Date getDateChanged() {
+		return dateChanged;
+	}
+
+	@Override
+	public void setDateChanged(Date dateChanged) {
+		this.dateChanged = dateChanged;
+	}
+
+	@Override
+	public Boolean isRetired() {
+		return retired;
+	}
+	
+    @Override
+	public Boolean getRetired() {
+		return retired;
+	}
+
+	@Override
+	public void setRetired(Boolean retired) {
+		this.retired = retired;
+	}
+
+	@Override
+	public User getRetiredBy() {
+		return retiredBy;
+	}
+
+	@Override
+	public void setRetiredBy(User retiredBy) {
+		this.retiredBy = retiredBy;
+	}
+
+	@Override
+	public Date getDateRetired() {
+		return dateRetired;
+	}
+
+	@Override
+	public void setDateRetired(Date dateRetired) {
+		this.dateRetired = dateRetired;
+	}
+
+	@Override
+	public String getRetireReason() {
+		return retireReason;
+	}
+
+	@Override
+	public void setRetireReason(String retireReason) {
+		this.retireReason = retireReason;
+	}
 }


jwnasambu:
@ibacher, @dkayiwa , @mozzy11 ,  am getting 2 conflicts which I don't know how to go about them. Kindly, how can I go about it?

ibacher:
@jwnasambu I *think* I fixed the merge issues here. Could you kindly add some implementations to those method stubs?

jwnasambu:
Thanks @ibacher I had tried several ways to sold the issue but I wasn't successful. How did you go about it? because am bound to face the same challenge again in my programming journey.

ibacher:
@jwnasambu It looks like I didn't quite do it right. The right way to go forward is to make sure you update your `master` branch before starting a new branch. So, usually, I keep two remotes on my Git repositories:

`origin` – pointing to my clone of a repo
`upstream` — pointing to the OpenMRS version of a repo.

I ensure that my `master` branch tracks the branch in `upstream`. So something like this:

```
# add the upstream
git remote add upstream git@github.com:openmrs/openmrs-core.git
git fetch --all --prune

# set master to track upstream
git checkout master
git branch --set-upstream-to upstream/master

# update local copy of master
git pull
```

Then, when I create a branch, I base it off the latest master:

```
git checkout master
git pull

git checkout -b TRUNK-5665 master
```

Usually, that should be enough to minimize merge conflicts. Otherwise, before pushing or after pushing a discovering a conflict:

```
# update local master
git checkout master
git pull

# go to my branch
git checkout TRUNK-5665

# try to update to the latest master
git rebase master
```

Unfortunately, if there are conflicts that Git can't resolve, you have to manually resolve them, which usually involves this process:

1. Look at the first conflict
2. If my change should replace what's been changed in master since my last update, just overwrite it with my version.
3. If my change shouldn't replace what's there, I try to figure out where my changes fit "next to" the existing code and re-write it.
4. Go to the next conflict and keeping going until they are resolved.

jwnasambu:
@ibacher  I now know where I made a mistake in the process of resolving the conflicts. I stopped on this command 

# update local master
git checkout master
git pull 

and that is why I wasn't making a progress. Thanks so so much

api/src/main/java/org/openmrs/User.java
```diff
@@ -35,14 +36,31 @@
...
+*dummy table containing these columns to 'please' hibernate until we fix the Class hierarchy see TRUNK-5665
+*/
+@SecondaryTable(name = "users_unused_fields")
+@AttributeOverrides({
+        @AttributeOverride(name = "name", column = @Column(table = "users_unused_fields", name = "name", insertable = false, updatable = false)),
+        @AttributeOverride(name = "description", column = @Column(table = "users_unused_fields", name = "description", insertable = false, updatable = false)) })
+public class User extends BaseOpenmrsObject  implements java.io.Serializable, Attributable<User>, Auditable, Retireable {
+=======
+public class User extends BaseOpenmrsObject implements java.io.Serializable, Attributable<User>, Auditable, Retireable {
+>>>>>>> 7f1edc92d80496f5d1c86e9e0b968292c5ebf238
```
ibacher, line 57: @jwnasambu So here, we'll want to select this version:

```
public class User extends BaseOpenmrsObject implements java.io.Serializable, Attributable<User>, Auditable, Retireable {
```

This code above is a Git conflict message. If shows two different versions of this section that git cannot resolve. These sections are divided by the `=======`. So we can treat the top as option 1 and the part underneath as option 2. `<<<<<<<` is used to indicate where the conflict begins and `>>>>>>>` indicates where it ends, so we don't want to modify anything outside of that.

In this, as I said, we want to go with option 2, so we'll get rid of the text that makes up option 1 and the conflict markers:

```suggestion
public class User extends BaseOpenmrsObject implements java.io.Serializable, Attributable<User>, Auditable, Retireable {
```

api/src/main/java/org/openmrs/User.java
```diff
@@ -53,14 +71,74 @@
...
+	@SuppressWarnings("deprecation")
+	@ElementCollection
+	@CollectionTable(name = "user_property", joinColumns = @JoinColumn(name = "user_id", nullable = false))
+	@MapKeyColumn(name = "property", length = 100)
+	@Column(name = "property_value")
+	@Cascade({ CascadeType.SAVE_UPDATE, CascadeType.MERGE, CascadeType.EVICT })
+=======
+	private Set<Role> roles;
+	
+>>>>>>> 7f1edc92d80496f5d1c86e9e0b968292c5ebf238
```
ibacher, line 119: Here we have something similar. However, unlike the above, we don't just cleanly want option 1 or option 2. We primarily want what's in option 2, but we also want the fields (but not the annotations) defined in option 1. 

If you want to understand *why* you get this conflict, it's because the version of `master` you were working from had [this commit](https://github.com/openmrs/openmrs-core/commit/267f4df2ba3d683323e2938690f5bcef116318b6#diff-7325e970b68864df576bcb7c49e495c0), but it's [since been backed out](https://github.com/openmrs/openmrs-core/commit/4d53a02710df2fda97ca5579772b629022267677#diff-7325e970b68864df576bcb7c49e495c0); this is also why we want option 2 with some additions. While we don't want to add all the Hibernate annotations in, we do want the fields that used to be defined in `BaseChangeableOpenmrsMetadata` that are required to support the `Auditable` adn `Retireable` interfaces.

So in this case, we want to end up with something like this:

```suggestion
	private User creator;
	
	private Date dateCreated;

	private User changedBy;
	
	private Date dateChanged;
	
	private String retired;
	
	private User retiredBy;
	
	private String retireReason;
	
	private Date dateRetired;

	private Set<Role> roles;
```

api/src/main/java/org/openmrs/User.java
```diff
@@ -593,5 +671,127 @@ public Integer getId() {
...
+		return creator;
+	}
+
+	@Override
+	public void setCreator(User creator) {
+		this.creator = creator;
+<<<<<<< HEAD
+		
+=======
+>>>>>>> 7f1edc92d80496f5d1c86e9e0b968292c5ebf238
```
ibacher, line 686: These issues all arise because I think, of the version I committed here. Here and below, we should go with option 2.

```suggestion
```

api/src/main/java/org/openmrs/User.java
```diff
@@ -593,5 +671,127 @@ public Integer getId() {
...
+		return dateCreated;
+	}
+
+	@Override
+	public void setDateCreated(Date dateCreated) {
+		this.dateCreated = dateCreated;
+<<<<<<< HEAD
+		
+=======
+>>>>>>> 7f1edc92d80496f5d1c86e9e0b968292c5ebf238
```
ibacher, line 700:
```suggestion
```

api/src/main/java/org/openmrs/User.java
ibacher: Option 2 is, indentally, leaving that space blank.

api/src/main/java/org/openmrs/User.java
```diff
@@ -593,5 +671,127 @@ public Integer getId() {
...
+		return changedBy;
+	}
+
+	@Override
+	public void setChangedBy(User changedBy) {
+		this.changedBy = changedBy;
+<<<<<<< HEAD
+		
+=======
+>>>>>>> 7f1edc92d80496f5d1c86e9e0b968292c5ebf238
```
ibacher, line 714:
```suggestion
```

api/src/main/java/org/openmrs/User.java
```diff
@@ -593,5 +671,127 @@ public Integer getId() {
...
+		return dateChanged;
+	}
+
+	@Override
+	public void setDateChanged(Date dateChanged) {
+		this.dateChanged = dateChanged;
+<<<<<<< HEAD
+		
+=======
+>>>>>>> 7f1edc92d80496f5d1c86e9e0b968292c5ebf238
```
ibacher, line 728:
```suggestion
```

api/src/main/java/org/openmrs/User.java
```diff
@@ -593,5 +671,127 @@ public Integer getId() {
...
+	}
+
+	@Override
+	public Boolean isRetired() {
+<<<<<<< HEAD
+		return null;
+=======
+		return retired;
+>>>>>>> 7f1edc92d80496f5d1c86e9e0b968292c5ebf238
```
ibacher, line 737:
```suggestion
		return retired;
```

api/src/main/java/org/openmrs/User.java
```diff
@@ -593,5 +671,127 @@ public Integer getId() {
...
+		return retired;
+	}
+
+	@Override
+	public void setRetired(Boolean retired) {
+		this.retired = retired;
+<<<<<<< HEAD
+		
+=======
+>>>>>>> 7f1edc92d80496f5d1c86e9e0b968292c5ebf238
```
ibacher, line 751:
```suggestion
```

api/src/main/java/org/openmrs/User.java
```diff
@@ -593,5 +671,127 @@ public Integer getId() {
...
+		return retiredBy;
+	}
+
+	@Override
+	public void setRetiredBy(User retiredBy) {
+		this.retiredBy = retiredBy;
+<<<<<<< HEAD
+		
+=======
+>>>>>>> 7f1edc92d80496f5d1c86e9e0b968292c5ebf238
```
ibacher, line 765:
```suggestion
```

api/src/main/java/org/openmrs/User.java
```diff
@@ -593,5 +671,127 @@ public Integer getId() {
...
+		return dateRetired;
+	}
+
+	@Override
+	public void setDateRetired(Date dateRetired) {
+		this.dateRetired = dateRetired;
+<<<<<<< HEAD
+		
+=======
+>>>>>>> 7f1edc92d80496f5d1c86e9e0b968292c5ebf238
```
ibacher, line 779:
```suggestion
```

api/src/main/java/org/openmrs/User.java
```diff
@@ -593,5 +671,127 @@ public Integer getId() {
...
+	@Override
+	public void setRetireReason(String retireReason) {
+		this.retireReason = retireReason;
+<<<<<<< HEAD
+		
+	}
 	
+=======
+	}
+>>>>>>> 7f1edc92d80496f5d1c86e9e0b968292c5ebf238
```
ibacher, line 796:
```suggestion
		this.retireReason = retireReason;
	}
```

api/src/main/java/org/openmrs/User.java
```diff
@@ -35,14 +36,27 @@
...
+*User inherits name and description properties from BaseOpenmrsMetadata but they don't exist in the users table since 
+*they are not used, because these field exist in BaseOpenmrsData and they are marked as persistent, hibernate generates 
+*a query containing these columns which of course fails to execute because of missing columns, so we introduce this 
+*dummy table containing these columns to 'please' hibernate until we fix the Class hierarchy see TRUNK-5665
+*/
+@SecondaryTable(name = "users_unused_fields")
+@AttributeOverrides({
+        @AttributeOverride(name = "name", column = @Column(table = "users_unused_fields", name = "name", insertable = false, updatable = false)),
+        @AttributeOverride(name = "description", column = @Column(table = "users_unused_fields", name = "description", insertable = false, updatable = false)) })
+public class User extends BaseOpenmrsObject implements java.io.Serializable, Attributable<User>, Auditable, Retireable {
```
ibacher, line 53: So here we really did want to end up with replacing all of this with one line:

```suggestion
public class User extends BaseOpenmrsObject implements java.io.Serializable, Attributable<User>, Auditable, Retireable {
```

ibacher: Similarly, the rest of the Git merge issues really do need to be resolved before this will compile, because otherwise you end up with things like this that Java doesn't know how to process: `<<<<<<< HEAD`

jwnasambu:
@ibacher thanks so so much for the explanation. I spend the better part of today to solve the GIT conflict and am glad I have learned the skill and it was worthy my time. Am trying to fix the Travis CI though from a hospital bed of which it can't be that fast but feel free to leave a comment.

PR #3288, opened 2020-07-02
TRUNK-5826: Fix error when creating user on Postgres

Am-Coder:
…reSQL

In case of PostgreSQL, on use of session.saveOrUpdate(..) , only the sequence gets increased and no insert statement are issued by hibernate. So, I had to flush the session in order to avoid the null pointer exception being throw for LoginCredential.
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5826

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
@@ -77,7 +77,12 @@ public class HibernateUserDAO implements UserDAO {
 		sessionFactory.getCurrentSession().saveOrUpdate(user);
 		
 		if (isNewUser && password != null) {
-			// In case of PostgreSQL hibernate session.saveOrUpdate does not issues insert statements at the same time
+			/* In OpenMRS, we are using generation strategy as native which will convert to IDENTITY 
+			 for MySQL and SEQUENCE for PostgreSQL. When using IDENTITY strategy, hibernate directly 
+			 issues insert statements where as with  SEQUENCE strategy hibernate only increments 
+			 sequences and issues insert on session flush ( batching is possible) . 
+			 PostgreSQL behaves differently than MySQL because it makes use of SEQUENCE strategy. 
+			*/
 			sessionFactory.getCurrentSession().flush();
 			
 			//update the new user with the password
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
@@ -77,6 +77,9 @@ public class HibernateUserDAO implements UserDAO {
 		sessionFactory.getCurrentSession().saveOrUpdate(user);
 		
 		if (isNewUser && password != null) {
+			// In case of PostgreSQL hibernate session.saveOrUpdate does not issues insert statements at the same time
+			sessionFactory.getCurrentSession().flush();
+			
 			//update the new user with the password
 			String salt = Security.getRandomToken();
 			String hashedPassword = Security.encodeString(password + salt);


api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -77,6 +77,9 @@ public User saveUser(User user, String password) {
 		sessionFactory.getCurrentSession().saveOrUpdate(user);
 		
 		if (isNewUser && password != null) {
+			// In case of PostgreSQL hibernate session.saveOrUpdate does not issues insert statements at the same time
+			sessionFactory.getCurrentSession().flush();
```
HerbertYiga, line 81: @Am-Coder   Did you try running these changes under a running openmrs instance with a  PostgreSQL Db?

Am-Coder: Yes, I have verified the changes on PostgreSQL.

dkayiwa: @Am-Coder did you investigate why PostgreSQL behaves differently from MySQL in this scenario?

Am-Coder: Let me look into it and see if I can find something.

Am-Coder: @dkayiwa , I went through [docs](https://docs.jboss.org/hibernate/orm/3.5/javadocs/org/hibernate/Session.html) and a lot of articles but there is no specific piece of documentation that could point to this. I found this [article](https://dzone.com/articles/hibernate-identity-sequence) very useful. Based on this article, to me it appears that:
  
In OpenMRS, we are using generation strategy as native which will convert to IDENTITY for MySQL and SEQUENCE for PostgreSQL. When using IDENTITY strategy, hibernate directly issues insert statements( cannot make use of batching ) where as with SEQUENCE strategy hibernate only increments sequences and issues insert on session flush ( batching is possible) . PostgreSQL behaves differently than MySQL because it makes use of SEQUENCE strategy.

Here is the link to the SQL queries I logged that get executed with MySQL and PostgreSQL( with session flush and without session flush ) - https://pastebin.com/yTsKUgYP . This is in accordance with the article I mentioned above.

dkayiwa: Can you add this explanation to the code?

ibacher: Do we need this call to `flush()` to prevent an error of some kind? I'm not sure that flushing the session just so we get insert statements generated at the same time should be a requirement (and there can be some modest benefits to allowing Hibernate to defer issuing a query until the session is closed).

Am-Coder: @ibacher , The current logical implementation is according to MySQL which makes use of IDENTITY strategy where there is no defering. Let me see if I can do something without making use of flush. As of now I think it might require a different hibernate session for each saveOrUpdate and get calls. The problem here is that in the same session we persisting as well as then calling the peristed entity.

Am-Coder: > Can you add this explanation to the code?

Added the explanation

ibacher: @Am-Coder So, to be clear, I don't think we should be calling `flush()` if the only reason we're doing so is to generate an `INSERT` statement at the same point that Hibernate would generate an `INSERT` statement when being run on MySQL. We're using a different identifier generation strategy and that has different behaviour. We'd only want to call `flush()` here if *not* calling `flush()` results in some kind of error.

Am-Coder: @ibacher , if you do not call flush here then an error is raised . Link to logs https://pastebin.com/FmP9ekBp . Also I have attached the screenshot in the ticket description https://issues.openmrs.org/browse/TRUNK-5826 .

Am-Coder: @dkayiwa @ibacher , are the changes fine or something more needs to be done ?

PR #3316, opened 2020-07-08
TRUNK-5816: Migrate unit tests to Junit5

achilep:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5816

  No? -> execute above command

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -278,6 +278,24 @@
 			<groupId>com.sun.xml.bind</groupId>
 			<artifactId>jaxb-impl</artifactId>
 	    </dependency>
+	   <dependency>
+		   <groupId>org.junit.jupiter</groupId>
+		   <artifactId>junit-jupiter-api</artifactId>
+		   <version>${junitVersion}</version>
+		   <scope>test</scope>
+	   </dependency>
+	   <dependency>
+		   <groupId>org.junit.jupiter</groupId>
+		   <artifactId>junit-jupiter-engine</artifactId>
+		   <version>${junitVersion}</version>
+		   <scope>test</scope>
+	   </dependency>
+	   <dependency>
+		   <groupId>org.junit.vintage</groupId>
+		   <artifactId>junit-vintage-engine</artifactId>
+		   <version>${junitVersion}</version>
+		   <scope>test</scope>
+	   </dependency>
    </dependencies>
    <build>
      <resources>
@@ -373,4 +391,7 @@
 			</build>
 		</profile>
 	</profiles>
+	<properties>
+		<junitVersion>5.6.2</junitVersion>
+	</properties>
 </project>
--- a/api/src/test/java/org/openmrs/OpenmrsTestsTest.java
+++ b/api/src/test/java/org/openmrs/OpenmrsTestsTest.java
@@ -9,7 +9,8 @@
  */
 package org.openmrs;
 
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.File;
 import java.lang.reflect.Method;
@@ -18,9 +19,9 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.regex.Pattern;
 
-import org.junit.Assert;
 import org.junit.Ignore;
-import org.junit.Test;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
 import org.openmrs.util.OpenmrsUtil;
 
 /**
@@ -43,7 +44,7 @@ public class OpenmrsTestsTest {
 	public void shouldHaveAtLeastOneTest() {
 		List<Class<?>> classes = getTestClasses();
 		
-		assertTrue("There should be more than one class but there was only " + classes.size(), classes.size() > 1);
+		assertTrue(classes.size() > 1, "There should be more than one class but there was only " + classes.size());
 	}
 	
 	/**
@@ -61,12 +62,11 @@ public class OpenmrsTestsTest {
 				
 				// make sure every "test" method (determined by having 
 				// the @Test annotation) starts with "testShould"
-				if (method.getAnnotation(Test.class) != null) {
+				if (method.getAnnotation(Test.class) != null && method.getAnnotation(org.junit.Test.class) != null) {
 					String methodName = method.getName();
 					
 					boolean passes = methodName.startsWith("should") || methodName.contains("_should");
-					assertTrue(currentClass.getName() + "#" + methodName
-					        + " is supposed to either 1) start with 'should' or 2) contain '_should' but it doesn't", passes);
+					assertTrue(passes, currentClass.getName() + "#" + methodName + " is supposed to either 1) start with 'should' or 2) contain '_should' but it doesn't");
 				}
 			}
 		}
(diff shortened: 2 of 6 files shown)


api/pom.xml
```diff
@@ -278,6 +278,21 @@
 			<groupId>com.sun.xml.bind</groupId>
 			<artifactId>jaxb-impl</artifactId>
 	    </dependency>
+	   <dependency>
+		   <groupId>org.junit.jupiter</groupId>
+		   <artifactId>junit-jupiter-api</artifactId>
+		   <version>${junitVersion}</version>
```
teleivo, line 284: I believe we should add `<scope>test</scope>` to every junit dependency so these dependencies are not packaged into the api jar

teleivo: see for example https://github.com/junit-team/junit5-samples/blob/main/junit5-jupiter-starter-maven/pom.xml#L22

achilep: @teleivo updated

api/src/test/java/org/openmrs/OpenmrsTestsTest.java
```diff
@@ -9,7 +9,8 @@
  */
 package org.openmrs;
 
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertFalse;
```
teleivo, line 12: @achilep will you also add a Test? I believe thats a class testing whether an OpenMRS test is properly setup.

achilep: @teleivo there I don't understand well .

> will you also add a Test?

achilep: @teleivo surely I plan to increase the coverage of the tests. I plan to do it at the end of the migration.

teleivo: This file is like a metatest, going through all test classes and checking they have a test, ...
Can you please add another test class like ModuleFileParserUnitTest to this commit for example that is a unit test

teleivo: Sorry, I have to correct myself. ModuleFileParserUnitTest extends BaseContextMockTest so we should not include this one in the commit. So any other Test class that does not extend any of these `BaseContext` classes. You did already migrate a few previously.

achilep: thanks @teleivo . done

api/src/test/java/org/openmrs/OpenmrsTestsTest.java
```diff
@@ -111,18 +109,16 @@ public void shouldHaveClassNameEndWithTestIfContainsMethodTestAnnotations() {
 		for (Class<?> currentClass : getClasses("^.*(?<!Test|IT|PT)\\.class$")) {
 			
 			// skip over classes that are @Ignore'd
-			if (currentClass.getAnnotation(Ignore.class) == null) {
+			if (currentClass.getAnnotation(Ignore.class) == null || currentClass.getAnnotation(Disabled.class) == null) {
 				boolean foundATestMethod = false;
 				
 				for (Method method : currentClass.getMethods()) {
-					if (method.getAnnotation(Test.class) != null) {
+					if (method.getAnnotation(org.junit.Test.class) != null && method.getAnnotation(Test.class) != null) {
```
teleivo, line 116: Shouldnt this also be an `||`?
Since before it was setting `foundATestMethod` to true if there is a method with the `org.junit.Test` annotation. I think now it should also count the ones with the `org.junit.jupiter.api.Test` annotation as a test method. Both annotations will never be present so an `&&` cannot identify a test method then.

therefore

```suggestion
					if (method.getAnnotation(org.junit.Test.class) != null || method.getAnnotation(Test.class) != null) {
```

achilep: Screenshot from 2020-07-09 19-21-29
@teleivo  here is the error I got.

api/src/test/java/org/openmrs/OpenmrsTestsTest.java
```diff
@@ -111,18 +109,16 @@ public void shouldHaveClassNameEndWithTestIfContainsMethodTestAnnotations() {
 		for (Class<?> currentClass : getClasses("^.*(?<!Test|IT|PT)\\.class$")) {
 			
 			// skip over classes that are @Ignore'd
-			if (currentClass.getAnnotation(Ignore.class) == null) {
+			if (currentClass.getAnnotation(Ignore.class) == null || currentClass.getAnnotation(Disabled.class) == null) {
```
teleivo, line 112: oh, now I get it. So here it says `// skip over classes that are @Ignore'd`. So what we have to do is skip over classes that do not have the annotation `Ignore` **and** also do not have the annotation `Disabled`

```suggestion
			if (currentClass.getAnnotation(Ignore.class) == null && currentClass.getAnnotation(Disabled.class) == null) {
```

Otherwise you are going to check `CreateInitialDataSet`.

teleivo: So you get the error after my suggested change only because you are in fact checking a class that should not be checked since it is annotated with `@Ignore`. So you need to take both my suggestions and it should work if I am not mistaken.

teleivo:
I just tried it

```java
@Test
	public void shouldHaveClassNameEndWithTestIfContainsMethodTestAnnotations() {
		// loop over all methods that _don't_ end in Test.class
		for (Class<?> currentClass : getClasses("^.*(?<!Test|IT|PT)\\.class$")) {
			
			// skip over classes that are @Ignore'd
			if (currentClass.getAnnotation(Ignore.class) == null && currentClass.getAnnotation(Disabled.class) == null) {
				System.out.println("class: " + currentClass.getName());
				boolean foundATestMethod = false;
				
				for (Method method : currentClass.getMethods()) {
					if (method.getAnnotation(org.junit.Test.class) != null || method.getAnnotation(Test.class) != null) {
						foundATestMethod = true;
					}
				}
				
				assertFalse(foundATestMethod, currentClass.getName() + " does not end with 'Test' but contains @Test annotated methods");
			}
		}
	}
```

and it does not print the `CreateInitialDataSet` class name. If I also print the class name on current master in this test it also does not print the `CreateInitialDataSet`. So that seems to be the desired behavior. 👌
(please dont commit my println ;)

achilep:
thanks @teleivo  , done

api/src/test/java/org/openmrs/OpenmrsTestsTest.java
```diff
@@ -61,12 +62,11 @@ public void shouldStartWithShould() {
 				
 				// make sure every "test" method (determined by having 
 				// the @Test annotation) starts with "testShould"
-				if (method.getAnnotation(Test.class) != null) {
+				if (method.getAnnotation(Test.class) != null && method.getAnnotation(org.junit.Test.class) != null) {
```
teleivo, line 65:
```suggestion
				if (method.getAnnotation(Test.class) != null || method.getAnnotation(org.junit.Test.class) != null) {
```

teleivo: Tests will never have Junit 4 and Junit 5 annotations. Its always a good idea to make a test fail once you assume its good. `shouldStartWithShould` is supposed to fail if any method annotated with a Test annotation does not start with `should` or have `_should` in its name. So you can just open any test and delete a `should` and `_should` and see if that test here fails. Since you made. a change you should test that in a Junit 4 and 5 test method. That would have told you that with the `&&` its not checking any test method.

PR #3357, opened 2020-07-14
TRUNK-5862: Condition and Obs to implement FormRecordable.

icrc-loliveira:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5862

## Checklist: I completed these


api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.4.x.xml
```diff
@@ -142,4 +142,19 @@
...
 		<addForeignKeyConstraint baseColumnNames="order_id" baseTableName="test_order" constraintName="test_order_order_id_fk" deferrable="false" initiallyDeferred="false" onDelete="CASCADE" onUpdate="NO ACTION" referencedColumnNames="order_id" referencedTableName="orders"/>
 	</changeSet>
+
+	<changeSet id="20200714-Trunk-5862" author="loliveira">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<columnExists tableName="conditions" columnName="form_namespace_path"/>
+			</not>
+		</preConditions>
+		<comment>Adding "order_group_reason" column to order_group table</comment>
```
mks-d, line 152:
```xml
<comment>Adding "form_namespace_path" column to conditions table</comment>
```

api/src/test/java/org/openmrs/ConditionTest.java
```diff
@@ -0,0 +1,147 @@
...
+	 */
+	@Test
+	public void setFormField_shouldRejectANamepaceAndPathCombinationLongerThanTheMaxLength() throws Exception {
+		StringBuilder nsBuffer = new StringBuilder(125);
+		for (int i = 0; i < 125; i++) {
+			nsBuffer.append("n");
+		}
+		for (int i = 0; i < 130; i++) {
+			nsBuffer.append("p");
+		}
```
mks-d, line 120: There's tons of Java string utils to generate strings of a certain length no?

api/src/test/java/org/openmrs/ConditionTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * This class tests all methods that are not getter or setters in the Condition java object TODO: finish
+ * this test class for Condition
+ * 
+ * @see Condition
+ */
+public class ConditionTest {
```
mks-d, line 25: Rather than all this I'd add a couple of test cases in [`ConditionServiceImplTest`](https://github.com/icrc-loliveira/openmrs-core/blob/a83975187495d8e438dc2cc74b4da2fc00838adb/api/src/test/java/org/openmrs/api/impl/ConditionServiceImplTest.java#L37).

icrc-loliveira: Hi @mks-d,

I don't think that is the best approach. Since for that I will have to make changes in `ConditionServiceImplTest` to a test a code that is in another class.

Point can be made that I am not testing the correct class. So I made some search how o test an abstract class and find [this](https://www.baeldung.com/junit-test-abstract-class) and decide to try the `3.1.` approach. Let me know your thoughts.

api/src/main/java/org/openmrs/BaseFormRecordableOpenmrsData.java
```diff
@@ -0,0 +1,104 @@
...
+		if (nsAndPathTemp.length() > FORM_NAMESPACE_PATH_MAX_LENGTH) {
+			throw new APIException("Obs.namespaceAndPathTooLong", (Object[]) null);
+		}
+		if (StringUtils.countMatches(nsAndPathTemp, FORM_NAMESPACE_PATH_SEPARATOR) > 1) {
+			throw new APIException("Obs.namespaceAndPathNotContainSeparator", (Object[]) null);
+		}
+
+		markAsDirty(this.formNamespaceAndPath, nsAndPathTemp);
+		formNamespaceAndPath = nsAndPathTemp;
+	}
```
mks-d, line 94: This will probably be the implementation for `Obs`, by _this_ I mean when the setter that needs to call `markAsDirty(..)`.

The base implementation will be the same, without the calls to `markAsDirty(..)`.

api/src/main/java/org/openmrs/BaseFormRecordableOpenmrsData.java
```diff
@@ -0,0 +1,104 @@
...
+
+	protected abstract void markAsDirty(Object oldValue, Object newValue);
```
mks-d, line 96: In light of the above you'll be able to remove this. The reason being that from recordable OpenMRS data and dirty-able OpenMRS data are two different things, I don't think we should conflate them there. @dkayiwa, @ibacher thoughts?

api/src/main/java/org/openmrs/Condition.java
```diff
@@ -396,4 +396,10 @@ public boolean equals(Object o) {
 		}
 		return Objects.equals(endReason, conditionToBeChecked.endReason);
 	}
+	
+	@Override
+	protected void markAsDirty(Object oldValue, Object newValue) {
+		// Do nothing
+	}
```
mks-d, line 403: In light of the above, this will not be necessary either.

api/src/main/java/org/openmrs/FormRecordable.java
```diff
@@ -0,0 +1,61 @@
...
+ * <p>
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs;
+
+/**
+ * Defines the methods to make a bridge between an OpenMrsObject and is position in the form.
+ *
+ * @since 1.4
```
mks-d, line 15:
```java
@since 2.4.0
```

api/src/main/java/org/openmrs/FormRecordable.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public interface FormRecordable {
+
+	/**
+	 * Gets the namespace for the form field that was used to capture the obs details in the form
+	 *
+	 * @return the namespace
+	 * @since 1.11
```
mks-d, line 23: Same here, and same for other methods.

api/src/test/java/org/openmrs/BaseFormRecordableOpenmrsDataTest.java
```diff
@@ -0,0 +1,158 @@
...
+	 * @see BaseFormRecordableOpenmrsDataImpl#setFormField(String,String)
+	 */
+	@Test
+	public void setFormField_shouldRejectAPathContainingTheSeparator() throws Exception {
+		final String path = FORM_NAMESPACE_PATH_SEPARATOR + "my path";
+		BaseFormRecordableOpenmrsDataImpl impl = new BaseFormRecordableOpenmrsDataImpl();
+		assertThrows(APIException.class, () -> impl.setFormField("", path));
+	}
+
+	public class BaseFormRecordableOpenmrsDataImpl extends BaseFormRecordableOpenmrsData {
```
mks-d, line 141: Why not just use an `Obs`? That'd be the more backward compatible approach.

api/src/test/java/org/openmrs/ConditionTest.java
mks-d: @icrc-loliveira `ConditionServiceImplTest` would validate that all is ok with Spring + Hibernate. But the Liquibase changeset can only be tested at app runtime.

api/src/main/java/org/openmrs/BaseFormRecordableOpenmrsData.java
ibacher: I agree. We don't really use this pattern outside of Obs, so it's hard to see why it would be a requirement for the `formNamespaceAndPath` functionality.

dkayiwa: I also agree!

api/src/test/java/org/openmrs/ConditionTest.java
icrc-loliveira: I updated the test `saveCondition_shouldSaveNewCondition()` in `ConditionServiceImplTest`  to test the `formNamespaceAndPath`field.

api/src/main/java/org/openmrs/BaseFormRecordableOpenmrsData.java
icrc-loliveira: Removed the `markAsDirty(..)` from the `BaseFormRecordableOpenmrsData` and re-implemented the `obs.setFormField` as: 
```
public void setFormField(String namespace, String formFieldPath) {
	String oldValue = formNamespaceAndPath;
	super.setFormField(namespace, formFieldPath);
	markAsDirty(oldValue, formNamespaceAndPath);
}
```

api/src/main/java/org/openmrs/BaseFormRecordableOpenmrsData.java
icrc-loliveira: Removed.

api/src/main/java/org/openmrs/FormRecordable.java
icrc-loliveira: Done.

api/src/test/java/org/openmrs/BaseFormRecordableOpenmrsDataTest.java
icrc-loliveira: I prefer create a test class for this  (since is not to complicated) in order to the test be more independent.

api/src/main/java/org/openmrs/Condition.java
icrc-loliveira: Removed.

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -1177,91 +1171,13 @@ public void setDateCreated(Date dateCreated) {
...
-		}
-		if (StringUtils.countMatches(nsAndPathTemp, FORM_NAMESPACE_PATH_SEPARATOR) > 1) {
-			throw new APIException("Obs.namespaceAndPathNotContainSeparator", (Object[]) null);
-		}
-		
-		markAsDirty(this.formNamespaceAndPath, nsAndPathTemp);
-		formNamespaceAndPath = nsAndPathTemp;
+		String oldValue = formNamespaceAndPath;
+		super.setFormField(namespace, formFieldPath);
+		markAsDirty(oldValue, formNamespaceAndPath);
```
mks-d, line 1180: Are you sure it's 100% ok, unfortunately there is also [this piece](https://github.com/openmrs/openmrs-core/blob/de7bfbb8f84ccf82ea5fd4e934c7a9ba9e483f31/api/src/main/java/org/openmrs/Obs.java#L1242):
```java
if (namespace == null && formFieldPath == null) {
  markAsDirty(formNamespaceAndPath, null);
  formNamespaceAndPath = null;
  return;
}
```

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.4.x.xml
```diff
@@ -142,4 +142,19 @@
     	<dropForeignKeyConstraint baseTableName="test_order" constraintName="test_order_order_id_fk"/>
 		<addForeignKeyConstraint baseColumnNames="order_id" baseTableName="test_order" constraintName="test_order_order_id_fk" deferrable="false" initiallyDeferred="false" onDelete="CASCADE" onUpdate="NO ACTION" referencedColumnNames="order_id" referencedTableName="orders"/>
 	</changeSet>
+
+	<changeSet id="20200714-Trunk-5862" author="loliveira">
```
mks-d, line 146: @dkayiwa @ibacher any preference/convention for the changeset ID?

There seems to be none.

ibacher: This probably needs to be a longer conversation than just comments on a PR. We don't seem to have an established convention and this is at least in line with other IDs, so in that sense, it's fine.

Personally, I would prefer something like: Ticket-Short Description. So, for this case, something like `TRUNK-5862-add_form_namespace_and_path_to_condition`. The idea being that this is unique enough (i.e., every changeset is "namespaced" to a ticket), but descriptive enough that if it appears to an administrator during install, they might understand what failed.

mogoodrich: Yeah, we should come up with a standard convention, might be worth a Talk post.

But, also, yeah, as we don't currently have a convention so as long as this is in line with existing IDs we can move forward with this one while we figure things out.

mks-d:
@icrc-loliveira, can you perhaps rebase this on master? Since there seems to be conflicts being reported.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.4.x.xml
dkayiwa: I also agree with a talk post, but which should not block this pull request from getting merged.

api/src/main/java/org/openmrs/Obs.java
mogoodrich: I think this new code still takes care of that use case properly... ie marking as dirty if the formNamespaceAndPath goes from something to null, if I'm understanding it correctly.

api/src/test/java/org/openmrs/BaseFormRecordableOpenmrsDataTest.java
```diff
@@ -0,0 +1,153 @@
...
+ */
+package org.openmrs;
+
+import liquibase.util.StringUtils;
+import org.junit.jupiter.api.Test;
+import org.openmrs.api.APIException;
+
+import java.lang.reflect.Field;
+
+import static org.junit.jupiter.api.Assertions.*;
```
mogoodrich, line 18: Minor point, but by convention we don't use wildcard imports.

api/src/test/java/org/openmrs/ObsTest.java
```diff
@@ -35,6 +35,7 @@
 import java.util.Locale;
 import java.util.Set;
 
+import liquibase.util.StringUtils;
```
mogoodrich, line 38: Where is this being used?

api/src/test/java/org/openmrs/api/impl/ConditionServiceImplTest.java
```diff
@@ -9,40 +9,32 @@
...
 
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.openmrs.CodedOrFreeText;
-import org.openmrs.Condition;
-import org.openmrs.ConditionClinicalStatus;
-import org.openmrs.ConditionVerificationStatus;
-import org.openmrs.Encounter;
-import org.openmrs.Patient;
+import org.openmrs.*;
```
mogoodrich, line 16: More wildcard imports.

api/src/test/java/org/openmrs/api/impl/ConditionServiceImplTest.java
```diff
@@ -9,40 +9,32 @@
...
 import org.openmrs.api.ConditionService;
 import org.openmrs.api.PatientService;
 import org.openmrs.api.context.Context;
 import org.openmrs.test.jupiter.BaseContextSensitiveTest;
 
+import static org.junit.jupiter.api.Assertions.*;
```
mogoodrich, line 22: Another wildcard import

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -1177,91 +1171,13 @@ public void setDateCreated(Date dateCreated) {
...
 	/**
-	 * Gets the namespace for the form field that was used to capture the obs details in the form
-	 * 
-	 * @return the namespace
-	 * @since 1.11
-	 * <strong>Should</strong> return the namespace for a form field that has no path
-	 * <strong>Should</strong> return the correct namespace for a form field with a path
-	 * <strong>Should</strong> return null if the namespace is not specified
-	 */
-	public String getFormFieldNamespace() {
```
dkayiwa, line 1188: Instead of removing, shouldn't we deprecate these public methods? https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-Deprecation

mks-d: @dkayiwa this method doesn't really get removed from `Obs`, it's now coming from a new superclass.

dkayiwa: Oh cool 😊

mks-d: @dkayiwa here: https://github.com/openmrs/openmrs-core/pull/3357/files#diff-26312f3ab88527f26370de68fc508603R64

api/src/test/java/org/openmrs/api/impl/ConditionServiceImplTest.java
```diff
@@ -61,21 +62,31 @@ public void setup (){
...
 	public void saveCondition_shouldSaveNewCondition(){
+		
+		// Create Condition to test
+		String ns = "my ns";
+		String path = "my path";
 		Integer patientId = 2;
 		String uuid = "08002000-4469-12q3-551f-0339000c9a76";
 		CodedOrFreeText codedOrFreeText = new CodedOrFreeText();
 		Condition condition = new Condition();
+		condition.setFormField(ns, path);
```
dkayiwa, line 73: Modifying an existing test to set the form field creates an impression that it is required field. If not, i would rather create a new test.

dkayiwa:
Can you also update this to 864? https://github.com/openmrs/openmrs-core/blob/master/api/src/test/java/org/openmrs/util/DatabaseUpdaterDatabaseIT.java#L32

api/src/test/java/org/openmrs/BaseFormRecordableOpenmrsDataTest.java
dkayiwa: It feels safer doing tests with real existing classes. If independency is needed, then i would do such tests in addition to those that deal with the real existing classes, not minus. 😊

dkayiwa: FWIW, i have ever had a case where a dummy test class masked a bug in a real class. 😊

dkayiwa:
@mks-d i still have a problem with replacing obs with a dummy BaseFormRecordableOpenmrsDataImpl class in the tests.

api/src/test/java/org/openmrs/BaseFormRecordableOpenmrsDataTest.java
mks-d: @icrc-loliveira... you'll have to use `Obs` ;-)

See [here](https://github.com/openmrs/openmrs-core/pull/3357#issuecomment-660128032):
>@mks-d i still have a problem with replacing obs with a dummy BaseFormRecordableOpenmrsDataImpl class in the tests.

dkayiwa:
@mks-d oh yes we are! 😊

api/src/main/java/org/openmrs/BaseFormRecordableOpenmrsData.java
```diff
@@ -0,0 +1,100 @@
...
+package org.openmrs;
+
+import org.apache.commons.lang3.StringUtils;
+import org.openmrs.api.APIException;
+
+import javax.persistence.Column;
+import javax.persistence.MappedSuperclass;
+
+/**
+ * This is a super class to make a bridge between an OpenMrsObject and is position in the form.
```
samuelmale, line 19: Just minor comment here on the wording.

'This is a super class to make a bridge between an OpenMrsObject and **is** position in the form.' **->** 'This is a super class to make a bridge between an OpenMrsObject and **its** position in the form.'

api/src/test/java/org/openmrs/BaseFormRecordableOpenmrsDataTest.java
```diff
@@ -0,0 +1,140 @@
...
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs;
+
+import liquibase.util.StringUtils;
+import org.junit.Assert;
```
teleivo, line 13: please replace the JUnit 4 assertions in this test `import org.junit.Assert` since we migrated to JUnit 5. JUnit 5 ones are in the Jupiter package https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html

api/src/main/java/org/openmrs/BaseFormRecordableOpenmrsData.java
mks-d: @icrc-loliveira this maybe?
>`Base implementation of FormRecordable that bridges between a saved BaseChangeableOpenmrsData entity and the path in a form where it was recorded.`

PR #3387, opened 2020-07-21
Trunk 5678:Classloading issues in Openmrs installation and DB upgrades

gitcliff:
link:https://issues.openmrs.org/browse/TRUNK-5678?src=confmacro


web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1914,7 +1920,7 @@ private static boolean isDatabaseEmpty(Properties props) {
 				
 				String[] types = { "TABLE" };
 				
-				//get all tables
+				//get all tables.
```
HerbertYiga, line 1923: @gitcliff   well done on this,did we need this change?

PR #3392, opened 2020-07-22
TRUNK-5817: Migrate web and webapp package tests to Junit 5

achilep:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5817

  No? -> execute above command

### DIFF
--- a/web/src/test/java/org/openmrs/web/filter/update/UpdateFilterModelTest.java
+++ b/web/src/test/java/org/openmrs/web/filter/update/UpdateFilterModelTest.java
@@ -14,15 +14,9 @@ import java.util.Arrays;
 import java.util.List;
 import org.junit.Before;
 import org.junit.Test;
-import org.junit.runner.RunWith;
 import org.openmrs.liquibase.LiquibaseProvider;
-import org.openmrs.util.DatabaseUpdater;
 import org.openmrs.util.DatabaseUpdater.OpenMRSChangeSet;
 import org.openmrs.util.DatabaseUpdaterLiquibaseProvider;
-import org.powermock.api.mockito.PowerMockito;
-import org.powermock.core.classloader.annotations.PowerMockIgnore;
-import org.powermock.core.classloader.annotations.PrepareForTest;
-import org.powermock.modules.junit4.PowerMockRunner;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
@@ -34,24 +28,23 @@ import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 /**
  * Tests {@link UpdateFilterModel}.
  */
-@RunWith(PowerMockRunner.class)
-@PrepareForTest(DatabaseUpdater.class)
-@PowerMockIgnore({ "com.sun.org.apache.xerces.*", "javax.xml.*", "org.xml.*", "javax.management.*", "org.w3c.dom.*" })
 public class UpdateFilterModelTest {
 	
 	private DatabaseUpdaterLiquibaseProvider liquibaseProvider;
+	private DatabaseUpdaterWrapper databaseUpdaterWrapper;
 	
 	private UpdateFilterModel model;
 	
 	@Before
 	public void setUp() {
 		liquibaseProvider = new DatabaseUpdaterLiquibaseProvider();
-		PowerMockito.mockStatic(DatabaseUpdater.class);
+		databaseUpdaterWrapper = mock( DatabaseUpdaterWrapper.class );
 	}
 	
 	@Test
@@ -59,18 +52,16 @@ public class UpdateFilterModelTest {
 	        throws Exception {
 		List<OpenMRSChangeSet> changes = Arrays.asList(mock(OpenMRSChangeSet.class));
 		
-		when(DatabaseUpdater.getUnrunDatabaseChanges(any(LiquibaseProvider.class))).thenReturn(changes);
-		when(DatabaseUpdater.isLocked()).thenReturn(false);
+		when(databaseUpdaterWrapper.getUnrunDatabaseChanges(any(LiquibaseProvider.class))).thenReturn(changes);
+		when(databaseUpdaterWrapper.isLocked()).thenReturn(false);
 		
-		model = new UpdateFilterModel(liquibaseProvider);
+		model = new UpdateFilterModel(liquibaseProvider, databaseUpdaterWrapper);
 		
 		assertTrue("should require an update", model.updateRequired);
 		assertThat(model.changes, is(changes));
 		
-		PowerMockito.verifyStatic(DatabaseUpdater.class);
-		DatabaseUpdater.getUnrunDatabaseChanges(liquibaseProvider);
-		PowerMockito.verifyStatic(DatabaseUpdater.class, never());
-		DatabaseUpdater.updatesRequired();
+		verify( databaseUpdaterWrapper, times(1)).getUnrunDatabaseChanges( liquibaseProvider );
+		verify( databaseUpdaterWrapper, never()).updatesRequired();
 	}
 	
 	@Test
@@ -78,19 +69,17 @@ public class UpdateFilterModelTest {
 	        throws Exception {
 		List<OpenMRSChangeSet> changes = new ArrayList<>();
 		
-		when(DatabaseUpdater.getUnrunDatabaseChanges(any(LiquibaseProvider.class))).thenReturn(changes);
-		when(DatabaseUpdater.isLocked()).thenReturn(false);
-		when(DatabaseUpdater.updatesRequired()).thenReturn(true);
+		when(databaseUpdaterWrapper.getUnrunDatabaseChanges(any(LiquibaseProvider.class))).thenReturn(changes);
+		when(databaseUpdaterWrapper.isLocked()).thenReturn(false);
+		when(databaseUpdaterWrapper.updatesRequired()).thenReturn(true);
 		
-		model = new UpdateFilterModel(liquibaseProvider);
+		model = new UpdateFilterModel(liquibaseProvider, databaseUpdaterWrapper);
 		
 		assertTrue("should require an update", model.updateRequired);
 		assertThat(model.changes, is(empty()));
-		
-		PowerMockito.verifyStatic(DatabaseUpdater.class);
-		DatabaseUpdater.getUnrunDatabaseChanges(liquibaseProvider);
-		PowerMockito.verifyStatic(DatabaseUpdater.class);
-		DatabaseUpdater.updatesRequired();
+
+		verify( databaseUpdaterWrapper, times(1)).getUnrunDatabaseChanges( liquibaseProvider );
+		verify( databaseUpdaterWrapper, times(1)).updatesRequired();
 	}
 	
 	@Test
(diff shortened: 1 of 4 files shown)


teleivo:
@achilep you are trying to mock static methods with mockito. mockito does not allow you to do that. that's why we used powermock. now that we migrate to junit 5 and we cannot use powermock its not emough to just remove powermock. you need to either use reflection to set some fields that will then execute the code in a way you want in the specific test or make it an integration test. I suggest to try reflection first. for example using the spring reflectionutils you already used in the last test you migrated. if you need to get specific information on the code you are testing I suggest reaching out to @WolfSchlegel

achilep:
thanks @teleivo

web/src/test/java/org/openmrs/web/filter/update/UpdateFilterModelTest.java
```diff
@@ -78,18 +75,22 @@ public void createUpdateFilterModel_shouldRequiredAnUpdateIfChangesAreEmptyButDa
...
+		when(DatabaseUpdater.updatesRequired()).thenReturn(true);
+*/
+		assertTrue(model.updateRequired, "should require an update");
 		assertThat(model.changes, is(empty()));
 		
-		PowerMockito.verifyStatic(DatabaseUpdater.class);
+		verify(DatabaseUpdater.class);
 		DatabaseUpdater.getUnrunDatabaseChanges(liquibaseProvider);
-		PowerMockito.verifyStatic(DatabaseUpdater.class);
+		verify(DatabaseUpdater.class);
```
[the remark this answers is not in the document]
achilep: cc @ibacher  @mozzy11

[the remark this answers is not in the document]
teleivo: its always a good idea to read the javadocs of the methods you are using. if you are talking about the spring reflectionutils then yes the javadoc says it does not support final fields. for these you have to use powermock.reflection you can find a usage in the api package. I think something like `Whitebox.set...`

You will then also need to adjust the dependencies of the web package. Also see the api pom.XML where I added a test score dependency for powermock.reflection and excluded the other 2 powermock dependencies.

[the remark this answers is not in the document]
achilep: i am avoiding to use Whitebox.set since we want to remove powermock dependencies .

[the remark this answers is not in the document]
teleivo: did you have a look at the api pom? https://github.com/openmrs/openmrs-core/blob/master/api/pom.xml#L24-L45

I did exclude powermock mockito library and its junit 4 runner in the api package. what we are using is simply the powermock reflection library. since that one allows setting of static final fields.

teleivo:
@achilep any luck replacing the `when().then` with setting fields using reflection?

achilep:
thanks @teleivo , no luck. this is  the last error that i try to solve this error : `java.lang.IllegalStateException: identifying the snapshot version that had been used to initialize the OpenMRS database failed as no candidate change set resulted in zero un-run changes`

teleivo:
please push your latest changes and let me know which test is failing

web/src/test/java/org/openmrs/web/filter/update/UpdateFilterModelTest.java
```diff
@@ -9,49 +9,50 @@
...
+import org.powermock.reflect.Whitebox;
+
 /**
  * Tests {@link UpdateFilterModel}.
  */
-@RunWith(PowerMockRunner.class)
-@PrepareForTest(DatabaseUpdater.class)
-@PowerMockIgnore({ "com.sun.org.apache.xerces.*", "javax.xml.*", "org.xml.*", "javax.management.*", "org.w3c.dom.*" })
-public class UpdateFilterModelTest {
+public class UpdateFilterModelTest  extends BaseContextSensitiveTest {
```
achilep, line 45: i had some error with the database connection that is why i extends BaseContextSensitiveTest

web/src/test/java/org/openmrs/web/filter/update/UpdateFilterModelTest.java
```diff
@@ -77,19 +78,20 @@ public void createUpdateFilterModel_shouldrequireAnUpdateAndSetChangesToUnrunDat
...
+		FieldUtils.writeField(model, "changes", changes, true);
+		FieldUtils.writeField(changeLogDetective, "changeLogVersionFinder", changeLogVersionFinder, true);
+		Whitebox.setInternalState(DatabaseUpdater.class, "changeLogDetective", changeLogDetective);
 		
-		assertTrue("should require an update", model.updateRequired);
+		assertTrue(model.updateRequired, "should require an update");
 		assertThat(model.changes, is(empty()));
 		
-		PowerMockito.verifyStatic(DatabaseUpdater.class);
 		DatabaseUpdater.getUnrunDatabaseChanges(liquibaseProvider);
```
achilep, line 94: the error is with this method . i have some difficulty to manipulate the behavior of this method .

achilep: `java.lang.IllegalStateException: identifying the snapshot version that had been used to initialize the OpenMRS database failed as no candidate change set resulted in zero un-run changes
`

achilep: the exception is cause by this line 
https://github.com/openmrs/openmrs-core/blob/02ba3aed4e9df2de1261f284f8b4a7cab5eff789/api/src/main/java/org/openmrs/liquibase/ChangeLogDetective.java#L116

i have try to solve the error with this` FieldUtils.writeField(changeLogDetective, "changeLogVersionFinder", changeLogVersionFinder, true);` but it did  not work

teleivo: I am not sure I understand what you want to do. changeLogVersionFinder field is of type ChangeLogVersionFinder so you cannot set it to true since that value is a different type.

teleivo: @WolfSchlegel can you help us out here. So with JUnit 5 PowerMock and thus mocking static methods does not work anymore. We are migrating all test to JUnit 5. This might be the last one we have to migrate. What we can do is use reflection to set static fields. Can you help us out with replacing powermock in this test with reflection so we keep its original intent.

WolfSchlegel: Looking into this and apologies for the delay...

`org/openmrs/web/filter/update/UpdateFilterModelTest.java` is a unit test and extending `BaseContextSensitiveTest` makes it an integration test. I would advise against that and would revert that change first.

In the context of https://issues.openmrs.org/browse/TRUNK-4830 I wrote a few new integration tests and found that tests based on `BaseContextSensitiveTest`are not well isolated and understanding the behaviour of individual tests was quite hard. Eventually I created a new base class for integration tests.

@achilep could you please change the test so that it is a unit test again and ping me when this is done?

Thank you,
Wolf

achilep: sorry @WolfSchlegel for the delay . i restarted every thing back . this is the initial change .

achilep:
hello @WolfSchlegel  , i have done . it is now a unit test

web/src/test/java/org/openmrs/web/filter/update/UpdateFilterModelTest.java
WolfSchlegel: Thanks, I have a look

teleivo:
@WolfSchlegel just to reiterate what we try to achieve: migrate this test to JUnit 5. Constraint: PowerMock is not supported with JUnit 5.

So the tools we can most likely use to replace the use of powermock are reflection either from java itself or the libraries: spring test ReflectionUtils, or powermock.reflect Whitebox (which supports setting static final fields)

If you would keep the test as JUnit 4 but remove the use of PowerMock on the current master that would already solve our issue. We can then simply turn it into JUnit 5 in a few minutes.

web/src/test/java/org/openmrs/web/filter/update/UpdateFilterModelTest.java
WolfSchlegel: Hey there,

please have a look at https://github.com/WolfSchlegel/openmrs-core/commit/6a087975d80b5a592a9209d2ccd8b9c7e611cf6f.

I introduced a wrapper for the (mostly static) `org/openmrs/util/DatabaseUpdater.java` class so that PowerMock is no longer needed for mocking it. This is good news for `org/openmrs/web/filter/update/UpdateFilterModelTest.java`.

However, this comes at a cost as the next question is how to unit test the new wrapper without resorting to PowerMock. I gave JMockit a short try but bumped into issues that are (allegedly) resolved by rearranging the order of dependencies in the pom file (yuck) and it seems a java agent needs to be registered as well for JMockit (yuck-ish).

So I am back at square one looking for a unit test that proves that the all new `org/openmrs/util/DatabaseUpdaterWrapper.java` class plays nicely with the incumbent `org/openmrs/util/DatabaseUpdater.java` class.

To be frank, I am not sure how Java reflections or respective Spring utilities help here. At the end of the day I need to verify that the wrapper calls the (static) methods. For this purpose I introduced some static helper variables and a helper method in `org/openmrs/util/DatabaseUpdater.java`.

Please have a look and let me know what you think. I believe we are facing a trade-off between not testing something and adding test helpers in production code.

teleivo: thank you @WolfSchlegel !

I think your suggestion for adding a wrapper is good. I added a suggestion in the code at the commit you sent.

Our goal with the UpdateModelFilterTest is to test that the filter makes sure updates are made. Your wrapper enables us to do just that. The update logic should be tested else were.

I am ok with not testing the wrapper as its super thin. This tradeoff is also ok for me since it enables us to remove powermock which hopefully forces us to rethink our designs going forward. The proliferation of static methods with side effects in the code base are a sign of bad design for me that circumvent springs dependency injection and harm testability.

Can you please bring in your changes while keeping the UpdateModelFilterTest a JUnit 4 test. We can chat about my suggestion on the code. @achilep  will then do the migration to JUnit 5 once the changes are merged.

achilep: thanks @WolfSchlegel

WolfSchlegel: Thanks for your feedback, @teleivo . 

I removed the hand-rolled verification from the `DatabaseUpdater` class and moved the new wrapper to the same package that contains the `UpdateFilterModel` class. Making the wrapper package-private resulted in flaky Mockito behaviour when running `mvn clean install` so the wrapper remained a public class.

I created this [pull request](https://github.com/achilep/openmrs-core/pull/15) which is based on the current OpenMRS master. If you do not want to get an implicit rebase against master in mid flight, best cherry pick the relevant changes from [here](https://github.com/WolfSchlegel/openmrs-core/commit/0a5f9fe1d198ca2177a50bb44d3ecd5bf6f5c36f).

achilep:
@teleivo  you can merge it before we  can continue .

sherrif10:
Thanks @teleivo ,@achilep can you first move it to pull request not draft  if you dont mind

teleivo:
thank you very much @WolfSchlegel !!

achilep:
thank you very much @WolfSchlegel

PR #3398, opened 2020-07-26
TRUNK-5887 Validation error when starting Reference Metadata module on PostgreSQL

Am-Coder:
TRUNK-5887 Validation error when starting Reference Metadata

1. Mapped the sequence correctly
2. Trimmed the extra 2 spaces that are returned in case of PostgreSQL while doing the comparison. So that no validation error occurs.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5887

  No? -> execute above command

### DIFF
--- a/api/src/main/java/liquibase/ext/datatype/core/PostgreSQLUuidType.java
+++ b/api/src/main/java/liquibase/ext/datatype/core/PostgreSQLUuidType.java
@@ -19,7 +19,7 @@ import org.slf4j.LoggerFactory;
 /**
  * PostgreSQL returns white spaces for UUIDs since their actual length is 36 and we have set it
  * everywhere as 38 in consideration to ticket https://issues.openmrs.org/browse/TRUNK-1283. This
- * leads to validation issues. Thus we convert it to VARCHAR(38).
+ * leads to validation issues. Thus we convert it to CHAR(36).
  * 
  * @since 2.4
  */
--- a/api/src/main/java/liquibase/ext/datatype/core/PostgreSQLUuidType.java
+++ b/api/src/main/java/liquibase/ext/datatype/core/PostgreSQLUuidType.java
@@ -30,7 +30,7 @@ public class PostgreSQLUuidType extends CharType {
 	@Override
 	public DatabaseDataType toDatabaseDataType(Database database) {
 		if (database instanceof PostgresDatabase && getSize() == 38) {
-			DatabaseDataType result = new DatabaseDataType("varchar", 38);
+			DatabaseDataType result = new DatabaseDataType("CHARACTER", 36);
 			
 			log.debug("uuid type for PostgreSQL is '{}' ", result.getType());
 			
--- a/api/src/main/java/liquibase/ext/datatype/core/PostgreSQLUuidType.java
+++ b/api/src/main/java/liquibase/ext/datatype/core/PostgreSQLUuidType.java
@@ -18,7 +18,8 @@ import org.slf4j.LoggerFactory;
 
 /**
  * PostgreSQL returns white spaces for UUIDs since their actual length is 36 and we have set it
- * everywhere as 38 in consideration to ticket https://issues.openmrs.org/browse/TRUNK-1283
+ * everywhere as 38 in consideration to ticket https://issues.openmrs.org/browse/TRUNK-1283. This
+ * leads to validation issues. Thus we convert it to VARCHAR(38).
  * 
  * @since 2.4
  */
--- a/api/src/main/java/liquibase/ext/datatype/core/PostgreSQLUuidType.java
+++ b/api/src/main/java/liquibase/ext/datatype/core/PostgreSQLUuidType.java
@@ -1,0 +1,49 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package liquibase.ext.datatype.core;
+
+import liquibase.database.Database;
+import liquibase.database.core.PostgresDatabase;
+import liquibase.datatype.DatabaseDataType;
+import liquibase.datatype.core.CharType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * PostgreSQL returns white spaces for UUIDs since their actual length is 36 and we have set it
+ * everywhere as 38 in consideration to ticket https://issues.openmrs.org/browse/TRUNK-1283
+ * 
+ * @since 2.4
+ */
+public class PostgreSQLUuidType extends CharType {
+	
+	private static final Logger log = LoggerFactory.getLogger(PostgreSQLUuidType.class);
+	
+	@Override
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/validator/OrderFrequencyValidator.java
```diff
@@ -61,7 +61,7 @@ public void validate(Object obj, Errors errors) {
 			
 			Concept concept = orderFrequency.getConcept();
 			if (concept != null) {
-				if (!ConceptClass.FREQUENCY_UUID.equals(concept.getConceptClass().getUuid())) {
+				if (!ConceptClass.FREQUENCY_UUID.equals(concept.getConceptClass().getUuid().trim())) {
```
dkayiwa, line 64: Aren't we going to have to do this in many places? Did you evaluate other options? Like switching the file size to 36?

Am-Coder: @dkayiwa , yes and as I am testing and getting validation errors, I am making the changes. One another ticket I raised today holds a similar problem. 

Making uuid columns char(36) or varchar(38) did come in my mind, but isn't there any reason that why in OpenMRS we have set UUID columns as char(38) everywhere?

dkayiwa: https://issues.openmrs.org/browse/TRUNK-1283

dkayiwa: Can we write the custom liquibase class that automatically maps to char(36) for PostgreSQL?

Am-Coder: Let me look into it. I know something similar has been done for boolean conversion to tinyint(1) for MySQL.

Am-Coder:
@dkayiwa , converting to char(36) leads to other errors where we are having uuid as 38 length( logs: https://pastebin.com/3i3L2N7a ). VARCHAR(38) appears to me an ideal solution. So I have converted char(38) to varchar(38) for PostgreSQL and validation issues are resolved.

api/src/main/java/org/openmrs/validator/OrderFrequencyValidator.java
ibacher: I'd actually think that we'd want to just change all UUID columns to 36 characters. The likelihood that we need to contend with Microsoft-generated UUIDs seems to me exceedingly small (and in those cases, we should just be removing the curly-braces anyways).

Am-Coder: @ibacher , setting UUID to 36 chars gives an error : 

`
Caused by: java.sql.BatchUpdateException: Batch entry 0 insert into appointmentscheduling_appointment_type (name, description, duration, confidential, visit_type_id, uuid, date_created, date_changed, date_retired, retired, retire_reason, creator, changed_by, retired_by, appointment_type_id) values ('Urology (New Patient)', NULL, 20, 'FALSE', NULL, '4d85dda4-c437-11e4-a470-82b0ea87e2d8e', '2020-07-27 15:33:49+05:30', NULL, NULL, 'FALSE', NULL, 2, NULL, NULL, 22) was aborted: ERROR: value too long for type character(36)  Call getNextException to see other errors in the batch.
`

If you see length of '4d85dda4-c437-11e4-a470-82b0ea87e2d8e' is 37, even though there are no curly braces. Thus I shifted to using VARCHAR(38) instead. Link to logs: https://pastebin.com/3i3L2N7a

ibacher: @Am-Coder Do examples like that exist outside of the referencedemodata module? If not, shouldn't we just fix that module?

Am-Coder: @ibacher , Let me check after correcting this uuid if I get error refering to uuid in other modules while initializing Reference Application.

Am-Coder: @ibacher , I just corrected it [here](https://github.com/openmrs/openmrs-module-referencedemodata/blob/abbce3805f7c76c57a210cb8595352cca7812f34/api/src/main/java/org/openmrs/module/referencedemodata/bundle/AppointmentMetadata.java#L46) and there is no other error in any other module or reference demo data module related to uuid being more then 36 characters. But for correction I just removed the last extra character. So, we are good to go with char(36) for PostgreSQL after this correction.

@dkayiwa , what is your say on this to keep uuid to char(36) or varchar(38) and whether this be done for MySQL as well ?

dkayiwa: Doing a uuid of 36 characters makes more sense.

Am-Coder: Made the changes to convert char(38) to char(36) for PostgreSQL.

api/src/main/java/liquibase/ext/datatype/core/PostgreSQLUuidType.java
```diff
@@ -0,0 +1,50 @@
...
+ * 
+ * @since 2.4
+ */
+public class PostgreSQLUuidType extends CharType {
+	
+	private static final Logger log = LoggerFactory.getLogger(PostgreSQLUuidType.class);
+	
+	@Override
+	public DatabaseDataType toDatabaseDataType(Database database) {
+		if (database instanceof PostgresDatabase && getSize() == 38) {
```
dkayiwa, line 32: What is it is another char(38) field which is not a uuid?

Am-Coder: @dkayiwa  @ibacher , I couldn't find a way to get access to column name within the scope of this method. Apparently, what I have understood is we will have to make the changes in diferent change classes [here](https://github.com/liquibase/liquibase/tree/bce767333bd4867708a03ca5fb90df868433fb44/liquibase-core/src/main/java/liquibase/change/core) and this does not seem right.

If we hold a possibility of having any other fields with char(38), what if we apparently make change to using varchar(38) and if we finally are ready to revert https://issues.openmrs.org/browse/TRUNK-1283 , then we make char(36) directly into liquibase changesets for both MySQL and PostgreSQL rather then going this way ?

dkayiwa: Would the revert require touching all existing changesets?

Am-Coder: Yes, that is what I have in mind.

dkayiwa: Would that require a validation checksum for the so many changesets?

Am-Coder: Apparently, Yes.

dkayiwa: That would be too many changes.

ibacher: Rather than reverting (which isn't really the way to go) would it make sense to just add a new changeset downsizing the UUID fields?

Am-Coder: A changeset like this ( I took it from the docs ) : 

```
<changeSet  author="liquibase-docs"  id="modifyDataType-example">  
    <modifyDataType  catalogName="cat"  
            columnName="id"  
            newDataType="int"  
            schemaName="public"  
            tableName="person"/>  
</changeSet>
```
would then need to be added for each table present in core as well as all the ref. application modules. Wouldn't it be adding some overhead ?

PR #3405, opened 2020-07-29
TRUNK-5892 Global property setting referencedemodata.createDemoPatien…

Am-Coder:
Global property setting referencedemodata.createDemoPatientsOnNextStartup not working on PostgreSQL

Currently, the comparison settings were set wrong. For case insensitive search in hibernate criteria we use Restrictions.ilike and for case sensitive or default search we use Restrictiones.eq. This was causing the problem in setting up reference demodata.

Also, had to make some changes in failing tests due to case sensitive look for concepts by concept name( logs : https://pastebin.com/nN7795dp ).
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5892

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
@@ -1779,9 +1779,9 @@ public class HibernateConceptDAO implements ConceptDAO {
 		criteria.add(Restrictions.or(Restrictions.eq("locale", locale), Restrictions.like("locale", language)));
 		
 		if (Context.getAdministrationService().isDatabaseStringComparisonCaseSensitive()) {
-			criteria.add(Restrictions.ilike("name", name));
-		} else {
 			criteria.add(Restrictions.eq("name", name));
+		} else {
+			criteria.add(Restrictions.ilike("name", name));
 		}
 		
 		criteria.add(Restrictions.eq("voided", false));
--- a/api/src/test/java/org/openmrs/api/ConceptServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/ConceptServiceTest.java
@@ -1468,7 +1468,7 @@ public class ConceptServiceTest extends BaseContextSensitiveTest {
 		//ensure that the association between the conceptName and the obs has been established
 		assertTrue(conceptService.hasAnyObservation(conceptName));
 		
-		Concept concept = conceptService.getConceptByName("cd4 count");
+		Concept concept = conceptService.getConceptByName("CD4 COUNT");
 		//make sure the name concept name exists
 		assertNotNull(concept);
 		assertThrows(ConceptNameInUseException.class, () -> conceptService.purgeConcept(concept));
@@ -1479,7 +1479,7 @@ public class ConceptServiceTest extends BaseContextSensitiveTest {
 	 */
 	@Test
 	public void saveConcept_shouldCreateANewConceptNameWhenTheOldNameIsChanged() {
-		Concept concept = conceptService.getConceptByName("cd4 count");
+		Concept concept = conceptService.getConceptByName("CD4 COUNT");
 		assertEquals(3, concept.getNames(true).size());
 		ConceptName oldName = null;
 		for (ConceptName cn : concept.getNames()) {
@@ -1509,7 +1509,7 @@ public class ConceptServiceTest extends BaseContextSensitiveTest {
 	 */
 	@Test
 	public void saveConcept_shouldVoidTheConceptNameIfTheTextOfTheNameHasChanged() {
-		Concept concept = conceptService.getConceptByName("cd4 count");
+		Concept concept = conceptService.getConceptByName("CD4 COUNT");
 		assertFalse(conceptService.getConceptName(1847).getVoided());
 		for (ConceptName cn : concept.getNames()) {
 			if (cn.getConceptNameId().equals(1847)) {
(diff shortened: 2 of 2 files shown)


api/src/test/java/org/openmrs/api/ConceptServiceTest.java
```diff
@@ -1468,7 +1468,7 @@ public void purgeConcept_shouldFailIfAnyOfTheConceptNamesOfTheConceptIsBeingUsed
 		//ensure that the association between the conceptName and the obs has been established
 		assertTrue(conceptService.hasAnyObservation(conceptName));
 		
-		Concept concept = conceptService.getConceptByName("cd4 count");
+		Concept concept = conceptService.getConceptByName("CD4 COUNT");
```
dkayiwa, line 1471: Aren't you changing the behaviour of the original code?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
```diff
@@ -1779,9 +1779,9 @@ public Concept getConceptByName(final String name) {
 		criteria.add(Restrictions.or(Restrictions.eq("locale", locale), Restrictions.like("locale", language)));
 		
 		if (Context.getAdministrationService().isDatabaseStringComparisonCaseSensitive()) {
```
Am-Coder, line 1781: @dkayiwa , Yes the behavior is changed. But if this condition evaluates to true we are supposed to search case sensitively ? 
We don't get this issue on MySQL because its collation is set to utf8_general_ci that is it is case insensitive. So, if you search for one of the concept with name "Systolic blood pressure" in db with query : 
**SELECT * FROM concept_name where name='SYSTOLIC BLOOD PRESSURE';**
This will give result in case of MySQL but not in case of PostgreSQL. This is the reason why no demo data is being added in case of PostgreSQL.

dkayiwa: Can you search the codebase for other cases of `isDatabaseStringComparisonCaseSensitive` and see what they do when the value is true?

Am-Coder: I found it [here](https://github.com/openmrs/openmrs-core/blob/77e5e8c4315a98f079ecab9ad9429b62cfd461a7/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java#L351) , [here](https://github.com/openmrs/openmrs-core/blob/77e5e8c4315a98f079ecab9ad9429b62cfd461a7/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java#L1035) , [here](https://github.com/openmrs/openmrs-core/blob/77e5e8c4315a98f079ecab9ad9429b62cfd461a7/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java#L1043) and [here](https://github.com/openmrs/openmrs-core/blob/1c21cb563496bbfeb58133b0ad2ea8b39121783e/api/src/main/java/org/openmrs/api/db/hibernate/PatientSearchCriteria.java#L282) . And apparently it appears to me it is supposed to behave opposite to what it is named. If it returns true then we are doing a case insensitive search. So, my implementation is not right.

Am-Coder: There are two solutions coming in my mind : either I set the global property GP_CASE_SENSITIVE_DATABASE_STRING_COMPARISON this to true before Ref. Demo Data setup or I change all the values in Ref. Demo Data to match with DB ?

dkayiwa: That is exactly what i wanted you to see! 😊 

Which of the two do you think would be better?

Am-Coder: Apparently, I think setting global property would be a better way. This way there would be minimal changes. Also, default value of this property is supposed to be true. There must be somewhere that it is being switched to false while calling the ReferenceDemoData Activator.

dkayiwa: After logging in, does it have a value of false?

Am-Coder: Yes, you can check in manage global properties section. What do you think should I go ahead with setting global property approach ?

dkayiwa: Isn't setting the global property simply hiding a bug?

Am-Coder: If we hold a use case where concept with name say "DENGUE" and "dengue" need to be treated separately, then it will turn out to be a data inconsistency bug and may lead to wrong data insertions. Then correcting the data in Ref. Demo Data module will be the ideal approach.

dkayiwa: I agree. So can you go with the ideal approach?

Am-Coder: I am on it :slightly_smiling_face:

dkayiwa:
@Am-Coder do we still need this?

Am-Coder:
@dkayiwa , we can close this.

PR #3407, opened 2020-07-29
TRUNK-5875:Unit test files left in place

jwnasambu:
## Description of what I changed:
I worked on the deleteOnExit() in the BaseContextSensitiveTest to ensure it properly clean-up a directory after running the unit tests.

## Issue I worked on:
https://issues.openmrs.org/browse/TRUNK-5875

### DIFF
--- a/api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
+++ b/api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
@@ -43,6 +43,7 @@ import javax.swing.JPasswordField;
 import javax.swing.JTextField;
 import javax.swing.UIManager;
 
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -353,10 +354,17 @@ public abstract class BaseContextSensitiveTest extends AbstractJUnit4SpringConte
 		runtimeProperties.setProperty(ModuleConstants.IGNORE_CORE_MODULES_PROPERTY, "true");
 		
 		try {
-			File tempappdir = File.createTempFile("appdir-for-unit-tests-", "");
+			final File tempappdir = File.createTempFile("appdir-for-unit-tests-", "");
 			tempappdir.delete(); // so we can make it into a directory
 			tempappdir.mkdir(); // turn it into a directory
-			tempappdir.deleteOnExit(); // clean up when we're done with tests
+			// clean up when we're done with tests
+			Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+				try {
+					FileUtils.deleteDirectory(tempappdir);
+				}
+				catch (IOException ignored) {
+				}
+			}));
 			
 			runtimeProperties.setProperty(OpenmrsConstants.APPLICATION_DATA_DIRECTORY_RUNTIME_PROPERTY, tempappdir
 			        .getAbsolutePath());
--- a/api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
+++ b/api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
@@ -60,6 +60,7 @@ import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.io.TempDir;
 import org.mockito.InjectMocks;
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.openmrs.ConceptName;
@@ -92,6 +93,7 @@ import org.springframework.test.context.TestExecutionListeners;
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 import org.springframework.transaction.annotation.Transactional;
 import org.xml.sax.InputSource;
+
 
 /**
  * This is the base for spring/context tests. Tests that NEED to use calls to the Context class and
@@ -130,6 +132,9 @@ public abstract class BaseContextSensitiveTest {
 	 * cached runtime properties
 	 */
 	protected static Properties runtimeProperties;
+	
+	@TempDir
+	public static File tempappdir;
 	
 	/**
 	 * Used for username/password dialog
--- a/api/src/test/java/org/openmrs/util/databasechange/Database1_9_7UpgradeIT.java
+++ b/api/src/test/java/org/openmrs/util/databasechange/Database1_9_7UpgradeIT.java
@@ -122,7 +122,13 @@ public class Database1_9_7UpgradeIT extends BaseContextSensitiveTest {
 		testAppDataDir = File.createTempFile("appdir-for-unit-tests", "");
 		testAppDataDir.delete();// so we can make turn it into a directory
 		testAppDataDir.mkdir();
-		
+		Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+			try {
+				FileUtils.deleteDirectory(testAppDataDir);
+			}
+			catch (IOException ignored) {
+			}
+		}));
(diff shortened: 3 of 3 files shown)


api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
```diff
@@ -353,10 +354,17 @@ public Properties getRuntimeProperties() {
 		runtimeProperties.setProperty(ModuleConstants.IGNORE_CORE_MODULES_PROPERTY, "true");
 		
 		try {
-			File tempappdir = File.createTempFile("appdir-for-unit-tests-", "");
+			final File tempappdir = File.createTempFile("appdir-for-unit-tests-", "");
```
teleivo, line 357: did you try to use the approach I outlined at https://talk.openmrs.org/t/unit-test-files-left-in-place/29617/2?u=teleivo ?

The way we create the temp dir is just not the JUnit 4 and 5 way. I would prefer we try to stick with what is the preferred way of the testing libraries. Its less code and easier to read and understand.

If it does not work and/or we have legitimate reasons to do this our own way I am ok with it.

ibacher: I actually think that using a JUnit 4 rule might require more steps to make this work. The problem is that we seem to only invoke this `getRuntimeProperties()` method during class construction and it then assumes that the `APPLICATION_DATA_DIRECTORY_RUNTIME_PROPERTY` stays stable across tests in the same class. I'm not sure the TempFolder rule allows for that?

With the JUnit 5 tests, I think we could use `@TempDir` correctly, since that can be scoped to "all tests in the same class."

api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
```diff
@@ -346,10 +347,17 @@ public Properties getRuntimeProperties() {
...
 			tempappdir.mkdir(); // turn it into a directory
-			tempappdir.deleteOnExit(); // clean up when we're done with tests
+			// clean up when we're done with tests
+			Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+				try {
+					FileUtils.deleteDirectory(tempappdir);
+				}
+				catch (IOException ignored) {
+				}
+			}));
```
ibacher, line 360: So, here, based on @teleivo's comments, we'd want to replace this with a static class variable. After the line that looks like this: `protected static Properties runtimeProperties;`, we'd want to add

```junit
@TempDir
private static File tempappdir;
```

api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
teleivo: there is a `@ClassRule`
https://junit.org/junit4/javadoc/4.12/org/junit/ClassRule.html

that can probably help us achieve that.

And should it fail if it cannot delete the folder? Because by default it does not unless we use what is explained here

https://www.baeldung.com/junit-4-rules#1-the-temporaryfolder-rule

teleivo: so in JUnit 4 it seems that to achieve the same behavior as in JUnit 5 TempDir

we would have to do something like

```java
@ClassRule
public static TemporaryFolder folder = TemporaryFolder.builder().assureDeletion().build();
```

api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
jwnasambu: @ibacher, @teleivo I did as suggested but run into this error https://pastebin.com/uAtB1VhR 
	```
/**
	 * cached runtime properties
	 */
	protected static Properties runtimeProperties;
	
	@TempDir
	private static File tempappdir;
```

Comment (teleivo) [reply api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java:360]: @jwnasambu the error you get says the field must not be private. so try to make it protected and if that does not work make it public

File: api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java [hunk 3]
```diff
@@ -346,10 +352,17 @@ public Properties getRuntimeProperties() {
 		runtimeProperties.setProperty(ModuleConstants.IGNORE_CORE_MODULES_PROPERTY, "true");
 		
 		try {
-			File tempappdir = File.createTempFile("appdir-for-unit-tests-", "");
+			final File tempappdir = File.createTempFile("appdir-for-unit-tests-", "");
```
teleivo, line 355: I think this line should be removed. The tempappdir is now created for us by JUnit.

api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
```diff
@@ -346,10 +352,17 @@ public Properties getRuntimeProperties() {
...
 			tempappdir.delete(); // so we can make it into a directory
 			tempappdir.mkdir(); // turn it into a directory
-			tempappdir.deleteOnExit(); // clean up when we're done with tests
+			// clean up when we're done with tests
+			Runtime.getRuntime().addShutdownHook(new Thread(() -> {
```
teleivo, line 359: this is also not needed as JUnit will also delete the directory

api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
```diff
@@ -34,6 +34,7 @@
 import java.util.Timer;
 import java.util.TimerTask;
 
+import org.apache.commons.io.FileUtils;
```
teleivo, line 37: we do not need FileUtils as we let JUnit create and delete the tempappdir

jwnasambu: @teleivo When I remove `import org.apache.commons.io.FileUtils;` I get this error https://pastebin.com/TiKe5rPs. Is it okay for me to conclude we still need the  `FileUtils` to provide operation based around the JDK File class?

teleivo: have you removed the code/lines I suggested? You introduced the import with your changes that I believe need to be removed as well. JUnit 5 TempDir creates and deletes the tempappdir

api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
```diff
@@ -131,6 +133,9 @@
 	 */
 	protected static Properties runtimeProperties;
```
teleivo, line 134: Should we not also adapt https://github.com/openmrs/openmrs-core/blob/master/api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java#L348-L360

@ibacher ?

I thought we also remove

```java
File tempappdir = File.createTempFile("appdir-for-unit-tests-", "");
			tempappdir.delete(); // so we can make it into a directory
			tempappdir.mkdir(); // turn it into a directory
			tempappdir.deleteOnExit(); // clean up when we're done with tests
```

and the try catch. So we are only left with setting the runtime property and the `OpenmrsUtil.setAppDir`.

ibacher: @teleivo  I completely agree.

teleivo: please remove the try catch 

and only keep the lines

https://github.com/openmrs/openmrs-core/blob/master/api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java#L354-L356

that are inside the try catch.

We are doing this since JUnit creates and deletes the tempappdir. So the only thing we need to do for tests that extend this base class is to set the runtime property and save the app dir using the openmrs utils so tests/code under test can write to this temp dir.

teleivo:
@jwnasambu are you going to reopen the PR?

jwnasambu:
@teleivo  yes am going to open. It closed accidentally when I was reverting the commit

PR #3409, opened 2020-07-30
TRUNK-5897: Remove discriminator from OrderGroup.hbm.xml

mseaton:
Removes incorrect discriminator column from OrderGroup.hbm.xml

This is the same fix that we had to make to OrderSet.hbm.xml in TRUNK-5877, likely introduced incorrectly at the same time.


mseaton:
FYI - this PR is against the 2.3.x branch.  Once reviewed and merged I'll up-port to master.

PR #3430, opened 2020-08-04
Add some notes on dependencies

ibacher:
Also refactors Lucene to use a property.

@teleivo I don't know if this is the best way to document the dependencies. Please let me know what you think!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -270,12 +270,12 @@
 			<dependency>
 				<groupId>org.hibernate</groupId>
 				<artifactId>hibernate-search-orm</artifactId>
-				<version>5.11.5.Final</version>
+				<version>${hibernateSearchVersion}</version>
 			</dependency>
 			<dependency>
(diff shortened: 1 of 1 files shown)


teleivo:
I just browsed the dependably cobfig options to see if we could somehow configure it to only give us 5.x updates for a specific dependency.

There is an allow config but it does not let us set a version range. There is an ignore option https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#ignore

so we would have to add ranges for all major versions above the one we care about. Its not ideal but might be worth it so er do not have to decline PRs going forward for every minor/patch update of let's say 8.x but still get 5.x updates automatically. What do you say?

ibacher:
@teleivo That's less than ideal, but probably not a bad idea for the few libraries I mentioned in the comments on this PR.

teleivo:
I'll see what I can do :ok_hand:

PR #3471, opened 2020-08-07
TRUNK-5818 Enable module devs to write Junit 4 & 5 tests

achilep:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5818

  No? -> execute above command

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -285,22 +285,6 @@
 			<artifactId>jaxb-impl</artifactId>
 	    </dependency>
 	   <dependency>
-		   <groupId>org.junit.jupiter</groupId>
-		   <artifactId>junit-jupiter-api</artifactId>
-	   </dependency>
-	   <dependency>
-		   <groupId>org.junit.jupiter</groupId>
-		   <artifactId>junit-jupiter-engine</artifactId>
-	   </dependency>
-	   <dependency>
-		   <groupId>org.junit.vintage</groupId>
-		   <artifactId>junit-vintage-engine</artifactId>
-	   </dependency>
-	   <dependency>
-		   <groupId>org.mockito</groupId>
-		   <artifactId>mockito-junit-jupiter</artifactId>
-	   </dependency>
-	   <dependency>
 		   <groupId>org.apache.lucene</groupId>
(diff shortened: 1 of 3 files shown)


achilep:
Hello @teleivo to solve this  https://github.com/openmrs/openmrs-module-webservices.rest/pull/445 I have to remove `<scope>test</scope>`  , how do you find the  the rest of dependencies that I deleted .

teleivo:
are you sure its the scope? these are test dependencies and should be of scope test
I am unsure about the other deletions. On the one hand its more consistent with what we currently do. On the other hand I think what we currently do with test dependencies is wrong. I need some more time to think about this.

achilep:
@teleivo I also read that the test scope is not transitive https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope

teleivo:
> @teleivo I also read that the test scope is not transitive https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope

thanks for pointing to this  resource :) so what you are saying is that if we use test scope, this leads to openmrs-test not propagating the dependency to web services? and if you remove the test scope it should by default be a compile scope of openmrs-test and thus 

> compile
This is the default scope, used if none is specified. Compile dependencies are available in all classpaths of a project. Furthermore, those dependencies are propagated to dependent projects.

so only then its propagated. Correct?

And since modules declare openmrs-test as dependency with scope test it is not included in their production jar / omod.

I just want us to fully understand how it currently works. I still think its wrong but I will post a longer thread on talk about my theory.

I you acknowledge my deduction above I think I am ok with merging this as its consistent with what we currently have.

achilep:
@teleivo I think that what we are doing is correct. I look at previous version of openmrs core api, i notice that there is not  `<scope>test</scope>` with junit 4  And the test dependencies are only include in the test pom file .
https://github.com/openmrs/openmrs-core/blob/9c87e0a8f4719e7db1b291a8ed41878906e19e8b/pom.xml#L463
https://github.com/openmrs/openmrs-core/blob/31fa025f2bb0e71208c17a0b5745e4eb9eeef14b/pom.xml#L461

PR #3491, opened 2020-08-10
TRUNK-5875:Unit test files left in place

jwnasambu:
I worked on the BaseContextSensitiveTest for JUnit 4 and JUnit5 to ensure it properly clean-up a directory after running the unit tests.

## Issue I worked on
https://issues.openmrs.org/browse/TRUNK-5875

### DIFF
--- a/api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
+++ b/api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
@@ -65,6 +65,8 @@ import org.dbunit.operation.DatabaseOperation;
 import org.hibernate.SessionFactory;
 import org.hibernate.cfg.Environment;
 import org.hibernate.dialect.H2Dialect;
+import org.junit.ClassRule;
+import org.junit.rules.TemporaryFolder;
 import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.Assume;
@@ -122,6 +124,9 @@ public abstract class BaseContextSensitiveTest extends AbstractJUnit4SpringConte
 	protected static final String INITIAL_XML_DATASET_PACKAGE_PATH = "org/openmrs/include/initialInMemoryTestDataSet.xml";
 	
 	protected static final String EXAMPLE_XML_DATASET_PACKAGE_PATH = "org/openmrs/include/standardTestDataset.xml";
+	
+	@ClassRule
+	public static TemporaryFolder tempappdir = TemporaryFolder.builder().assureDeletion().build();
 	
 	/**
 	 * cached runtime properties
@@ -352,19 +357,9 @@ public abstract class BaseContextSensitiveTest extends AbstractJUnit4SpringConte
 		// we don't want to try to load core modules in tests
 		runtimeProperties.setProperty(ModuleConstants.IGNORE_CORE_MODULES_PROPERTY, "true");
 		
-		try {
-			File tempappdir = File.createTempFile("appdir-for-unit-tests-", "");
-			tempappdir.delete(); // so we can make it into a directory
-			tempappdir.mkdir(); // turn it into a directory
-			tempappdir.deleteOnExit(); // clean up when we're done with tests
-			
-			runtimeProperties.setProperty(OpenmrsConstants.APPLICATION_DATA_DIRECTORY_RUNTIME_PROPERTY, tempappdir
+			runtimeProperties.setProperty(OpenmrsConstants.APPLICATION_DATA_DIRECTORY_RUNTIME_PROPERTY, tempappdir.getRoot()
 			        .getAbsolutePath());
-			OpenmrsUtil.setApplicationDataDirectory(tempappdir.getAbsolutePath());
-		}
-		catch (IOException e) {
-			log.error("Unable to create temp dir", e);
-		}
+			OpenmrsUtil.setApplicationDataDirectory(tempappdir.getRoot().getAbsolutePath());
 		
 		return runtimeProperties;
 	}
--- a/api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
+++ b/api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
@@ -60,6 +60,7 @@ import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.io.TempDir;
 import org.mockito.InjectMocks;
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.openmrs.ConceptName;
@@ -130,6 +131,9 @@ public abstract class BaseContextSensitiveTest {
 	 * cached runtime properties
 	 */
 	protected static Properties runtimeProperties;
+	
+	@TempDir
+	public static File tempappdir;
 	
 	/**
 	 * Used for username/password dialog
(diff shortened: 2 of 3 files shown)


api/src/test/java/org/openmrs/util/databasechange/Database1_9_7UpgradeIT.java
```diff
@@ -64,7 +66,8 @@
 	
 	private DatabaseUpgradeTestUtil upgradeTestUtil;
 	
-	private static File testAppDataDir;
+	@ClassRule
+	public static TemporaryFolder testAppDataDir = TemporaryFolder.builder().assureDeletion().build();
```
teleivo, line 70: you need to remove 125-128. please remember that JUnit is now creating and deleting the directory. So we do not need any code that creates/deletes athw testAppDataDir anymore

api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
```diff
@@ -345,19 +349,9 @@ public Properties getRuntimeProperties() {
 		// we don't want to try to load core modules in tests
 		runtimeProperties.setProperty(ModuleConstants.IGNORE_CORE_MODULES_PROPERTY, "true");
 		
-		try {
```
tendomart, line 348: @jwnasambu  you are trying to use a variable of a non existing or deleted object.

File tempappdir = File.createTempFile("appdir-for-unit-tests-", ""); was deleted.

tendomart: I believe that's why Travis is throwing   "[ERROR] /home/travis/build/openmrs/openmrs-core/api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java:[361,33] cannot find symbol

  symbol:   method getAbsolutePath() "

jwnasambu: @tendomart thanks for the review but I stand to be corrected but I think we need .getRoot()
        .getAbsolutePath()); to sort out the error

api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
```diff
@@ -131,6 +132,9 @@
 	 */
 	protected static Properties runtimeProperties;
 	
+	@TempDir
+	public static File tempappdir;
```
teleivo, line 136: @tendomart it dies exist its defined here

api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
teleivo: you can also read https://junit.org/junit5/docs/current/user-guide/#writing-tests-built-in-extensions-TempDirectory ; it also supports the newer java.nio.file.Path on which you can call https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#toAbsolutePath()

I suggest you try using the type java.nio.file.Path.

jwnasambu: @teleivo thanks let me do it.

api/src/test/java/org/openmrs/util/databasechange/Database1_9_7UpgradeIT.java
```diff
@@ -119,12 +122,8 @@ public static void createOrderEntryUpgradeFileWithTestData(String propString) th
...
 	@BeforeAll
 	public static void beforeClass() throws IOException {
-		testAppDataDir = File.createTempFile("appdir-for-unit-tests", "");
-		testAppDataDir.delete();// so we can make turn it into a directory
-		testAppDataDir.mkdir();
-		
-		System.setProperty(OpenmrsConstants.APPLICATION_DATA_DIRECTORY_RUNTIME_PROPERTY, testAppDataDir.getAbsolutePath());
-		OpenmrsUtil.setApplicationDataDirectory(testAppDataDir.getAbsolutePath());
+		System.setProperty(OpenmrsConstants.APPLICATION_DATA_DIRECTORY_RUNTIME_PROPERTY, testAppDataDir.getRoot().getAbsolutePath());
+		OpenmrsUtil.setApplicationDataDirectory(testAppDataDir.getRoot());
```
ibacher, line 126:
```suggestion
		OpenmrsUtil.setApplicationDataDirectory(testAppDataDir.getRoot().getAbsolutePath());
```

jwnasambu: @ibacher thanks so much! let me try out

api/src/test/java/org/openmrs/util/databasechange/Database1_9_7UpgradeIT.java
```diff
@@ -64,7 +66,8 @@
 	
 	private DatabaseUpgradeTestUtil upgradeTestUtil;
 	
-	private static File testAppDataDir;
+	@ClassRule
```
teleivo, line 69: this test class is a JUnit 5 one so it needs the JUnit 5 `@TempDir`
please try the java.nio.file.Path here as well

api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
```diff
@@ -352,19 +357,9 @@ public Properties getRuntimeProperties() {
...
 		runtimeProperties.setProperty(ModuleConstants.IGNORE_CORE_MODULES_PROPERTY, "true");
 		
-		try {
-			File tempappdir = File.createTempFile("appdir-for-unit-tests-", "");
-			tempappdir.delete(); // so we can make it into a directory
-			tempappdir.mkdir(); // turn it into a directory
-			tempappdir.deleteOnExit(); // clean up when we're done with tests
-			
-			runtimeProperties.setProperty(OpenmrsConstants.APPLICATION_DATA_DIRECTORY_RUNTIME_PROPERTY, tempappdir
+			runtimeProperties.setProperty(OpenmrsConstants.APPLICATION_DATA_DIRECTORY_RUNTIME_PROPERTY, tempappdir.getRoot()
```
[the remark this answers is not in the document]
jwnasambu: @teleivo thanks for reviewing my work! Kindly help me understand the tab that you are referring too. I wouldn't want to send you to laughter by telling you what I did that I thought I had sorted this "tab" issue.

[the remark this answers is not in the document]
teleivo: I made the comments in this file on the lines that you need to remove the tab. In general if you remove a block like if/else,try/catch,... and you keep code that was previously inside that block you will need to remove one level if nesting. So here you removed the try/catch and keep 2 lines of code. These 2 lines need to be unnested, thus one tab needs to be removed.

Comments on the PR are mostly attached to lines of code. So you can take that information into account when trying to understand the comment.

ibacher:
@jwnasambu @teleivo So the issue here turns out to be much more subtle than I thought. It's not just that we weren't guaranteeing that the application data directory was cleaned up, but rather we seem to be caching the initial application data directory in the [HibernateSessionFactoryBean](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java), which seems to only be called once for all of the unit tests run. This class sets a number of properties, but the most important one for this context is the `hibernate.search.default.indexBase`, which defines where Hibernate Search tells Lucene to find it's index. This property is set to `%APPLICATION_DATA_DIRECTORY%/lucene/indexes` which [gets set on the first run-through](https://github.com/openmrs/openmrs-core/blob/899095a11cc88bca9f3a01f985fe9d06962c9d7b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java#L167) and never changed. For some reason, this construct only tries to create the directory once, and after that it fails, hence the failed build.

I'm not really sure what the path forward is on this one, but it seems like we might be stuck with the hacky thing that kind of works and an AtExit runner...

jwnasambu:
@ibacher  and @teleivo thanks so much for guiding me on this. Am sorry you haven't seen any changes pushed because am still running into one error after another .

teleivo:
thank you @ibacher for your analysis! I need time to think more deeply about this. I assume that this bug is not an urgent one. It was likely around for quite some time. I therefore would like us to take more time to see if there is an alternative to the atExit runner.

PR #3506, opened 2020-08-13
Trunk 5927 Removed liquibase dependency for sequence update

Am-Coder:

Removed liquibase dependency for sequence update
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5927

  No? -> execute above command


api/src/main/java/org/openmrs/postgres/db/hibernate/HibernatePostgresDAO.java
```diff
@@ -0,0 +1,85 @@
...
+			// All the required PostgreSQL sequences that need to be updated
+			String postgresSequences = "SELECT setval('person_person_id_seq', (SELECT MAX(person_id) FROM person)+1);"
+			        + "SELECT setval('person_name_person_name_id_seq', (SELECT MAX(person_name_id) FROM person_name)+1);"
+			        + "SELECT setval('person_attribute_type_person_attribute_type_id_seq', (SELECT MAX(person_attribute_type_id) FROM person_attribute_type)+1);"
+			        + "SELECT setval('relationship_type_relationship_type_id_seq', (SELECT MAX(relationship_type_id) FROM relationship_type)+1);"
+			        + "SELECT setval('users_user_id_seq', (SELECT MAX(user_id) FROM users)+1);"
+			        + "SELECT setval('care_setting_care_setting_id_seq', (SELECT MAX(care_setting_id) FROM care_setting)+1);"
+			        + "SELECT setval('concept_datatype_concept_datatype_id_seq', (SELECT MAX(concept_datatype_id) FROM concept_datatype)+1);"
+			        + "SELECT setval('concept_map_type_concept_map_type_id_seq', (SELECT MAX(concept_map_type_id) FROM concept_map_type)+1);"
+			        + "SELECT setval('concept_stop_word_concept_stop_word_id_seq', (SELECT MAX(concept_stop_word_id) FROM concept_stop_word)+1);"
```
dkayiwa, line 65: If these are the only lines that have led to all these changes, then we should just put it in a database utility class and that should be all.

Am-Coder: Oh I see. I was unaware of utility class. Let me look into that.

Am-Coder:
@dkayiwa , I need to update the sequence in the ReferenceMetadataActivator class [here](https://github.com/openmrs/openmrs-module-referencemetadata/blob/8dc640fb4e29e7b5200e9c79027e7c23050c5095/api/src/main/java/org/openmrs/module/referencemetadata/ReferenceMetadataActivator.java#L110) and it might be required in activators of other modules as well. If we go with the Database Utility approach, then we need to have access to current hibernate session in the activator classes order to recognize that the current database is PostgreSQL. 

Is there a way we can get access to current hibernate session in the modules activator class ? I think we can access the session only at the DAO level.

dkayiwa:
Do we get any errors if we run this against mysql?

Am-Coder:
> Do we get any errors if we run this against mysql?

No there are no issues. I got Ref. App set up with demo data and all is working fine.

dkayiwa:
Then why do we need to find if the current database is postgresql?

Am-Coder:
> Then why do we need to find if the current database is postgresql?

Because we want to execute this update of sequence only when the db is PostgreSQL and for that we required hibernate session. For MySQL, this will fail because it follows identifier strategy.

dkayiwa:
What error do you get when you run it on mysql?

api/src/main/java/org/openmrs/postgres/db/hibernate/HibernatePostgresDAO.java
```diff
@@ -0,0 +1,85 @@
...
+		this.sessionFactory = sessionFactory;
+	}
+	
+	/**
+	 * @see org.openmrs.postgres.db.PostgresDAO#updateSequence()
+	 */
+	@Override
+	public void updateSequence() {
+		String dialect = sessionFactory.getProperties().getOrDefault("hibernate.dialect", "").toString();
+		if (StringUtils.containsIgnoreCase(dialect, "postgresql")) {
```
Am-Coder, line 54: @dkayiwa , as you can see here I am making use session to check for current database that is being used by OpenMRS via the set dialect, so **the current change will never cause an issue on MySQL. I have checked it on both MySQL and PostgreSQL and everything is working fine.**

Now, to make use of database utility approach I am not able to figure out how to get access to this hibernate session in module's activator class. This session will be required so as to pass to this utility method.

Am-Coder: Why I need the current change ?

I need to update postgresql sequences [here](https://github.com/openmrs/openmrs-module-referencemetadata/blob/8dc640fb4e29e7b5200e9c79027e7c23050c5095/api/src/main/java/org/openmrs/module/referencemetadata/ReferenceMetadataActivator.java#L110) , after the insertion of all the concepts in Reference Metadata Activator. If I don't update the sequence here then an error with logs : https://pastebin.com/23mQFUk1 this is raised when inserting concept source via reference application on PostgreSQL. So, I need to make a call to the service I have created in this PR here.

dkayiwa: How about Context.getAdministrationService().executeSQL?

dkayiwa: How about Context.getAdministrationService().executeSQL?

Am-Coder: Thanks for pointing to the AdministrativeService. I don't think we can use the executeSQL directly without doing some major changes. But what I think can be done is shift this updateSequence method to AdministrativeService . Then there won't be any requirement of a separate service as I have done here. Just a call like **Context.getAdministrationService().updatePostgreSQLsequence** sequence will do the work. I think this will be right. @dkayiwa What do you think ?

dkayiwa: Why can't we use executeSQL?

Am-Coder: If we pass the update sequence queries directly to this method then error will arise for MySQL since it does not support sequences. Now if we are to make use of this method then we will have to pass an extra argument to decide whether the sent query is of update sequence or not and then handle accordingly. If it is of update sequence type then execute it only on PostgreSQL.

dkayiwa: What error do you get on MySQL?

Am-Coder: If you execute the query "SELECT setval('person_person_id_seq', (SELECT MAX(person_id) FROM person)+1);" on MySQL then you get an error : 

**5000	Error Code: 1305. FUNCTION openmrs164.setval does not exist	0.611 sec**

Am-Coder:
@dkayiwa , I have made the changes and now I am using Administrative Service instead of a new service. I have tested the changes both on MySQL and PostgreSQL and everything is working fine.

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1820,6 +1820,17 @@ public void executing(ChangeSet changeSet, int numChangeSetsToRun) {
 							return;
 						}
 						
+						try {
+							// Update PostgreSQL Sequences after insertion of core data
+							Context.getAdministrationService().updatePostgresSequence();
```
dkayiwa, line 1825: Shouldn't we make this call only if we are running PostgreSQL?

dkayiwa: How about the UpdateFIlter.java?

dkayiwa: Will this need to be called every time data is inserted?

Am-Coder: > Shouldn't we make this call only if we are running PostgreSQL?

Yes We are supposed to make the call only when DB is PostgreSQL but since we do not have access to session at this level we cannot determine the DB. Two things come to my mind for this:
1. If it is regarding code readability then we can rename the method to UpdateSequenceIfDbIsPostgreSQL
or
2. We can split this single call into two different calls to AdminService, one two determine if db is PostgreSQL and other to update the sequence if previous call returns true. To me two calls appeared overhead so I went with a single call approach.

Am-Coder: > Will this need to be called every time data is inserted?

If the data that we are inserting does not holds id values for different entities then this call won't be required. PostgreSQL automatically updates the sequence in this case. But if in the data that is getting inserted we specify the id values for different entities then PostgreSQL does not increment the sequences and then we need to make a call to this service. This is the case [here](https://github.com/openmrs/openmrs-module-referencemetadata/blob/0cbf10b8103b97b9e1e0cd018706c47651ea20ea/api/src/main/java/org/openmrs/module/referencemetadata/ReferenceMetadataActivator.java#L109)

Am-Coder: > How about the UpdateFIlter.java?

I'll look into it and see if there's a way.

Am-Coder: > How about the UpdateFIlter.java?

The UpdateFilter is only called when there are insertions via liquibase . But consider [here](https://github.com/openmrs/openmrs-module-referencemetadata/blob/0cbf10b8103b97b9e1e0cd018706c47651ea20ea/api/src/main/java/org/openmrs/module/referencemetadata/ReferenceMetadataActivator.java#L101) where insertions are not via liquibase of Covid-19 Numric Concepts, then we cannot do anything and will get issues. So, I don't think it will work.

dkayiwa: Could we automate this by making the call from `OpenmrsObjectSaveHandler`?

Am-Coder: @dkayiwa , OpenMRSObjectSaveHandler is called before the "save__" or "void__", "unvoid_", "retire__", or "unretire___" methods in all the service classes but not around the method [importData](https://github.com/openmrs/openmrs-module-dataexchange/blob/caf00edb7529136581f694640709a3bafb6af3e9/api/src/main/java/org/openmrs/module/dataexchange/DataImporter.java#L56) in dataexhchange module. This method is called to save concepts from the xml dataset files using dbunit (see [here](https://github.com/openmrs/openmrs-module-referencemetadata/blob/0cbf10b8103b97b9e1e0cd018706c47651ea20ea/api/src/main/java/org/openmrs/module/referencemetadata/ReferenceMetadataActivator.java#L97)). 

Also, will it be alright to have the sequences updated even when PostgreSQL updates them automatically ?. There are only two cases where we will want to update the sequences via this service - First, when data is inserted via liquibase and second when a dataset is imported from an xml file ( in these two cases only id of object is specified).

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -301,4 +302,43 @@ public boolean isDatabaseStringComparisonCaseSensitive() {
...
+			        + "SELECT setval('field_type_field_type_id_seq', (SELECT MAX(field_type_id) FROM field_type)+1);"
+			        + "SELECT setval('hl7_source_hl7_source_id_seq', (SELECT MAX(hl7_source_id) FROM hl7_source)+1);"
+			        + "SELECT setval('location_location_id_seq', (SELECT MAX(location_id) FROM location)+1);"
+			        + "SELECT setval('order_type_order_type_id_seq', (SELECT MAX(order_type_id) FROM order_type)+1);"
+			        + "SELECT setval('patient_identifier_type_patient_identifier_type_id_seq', (SELECT MAX(patient_identifier_type_id) FROM patient_identifier_type)+1);"
+			        + "SELECT setval('scheduler_task_config_task_config_id_seq', (SELECT MAX(task_config_id) FROM scheduler_task_config)+1);"
+			        + "SELECT setval('scheduler_task_config_property_task_config_property_id_seq', (SELECT MAX(task_config_property_id) FROM scheduler_task_config_property)+1)"
+			        + "";
+			Session session = sessionFactory.getCurrentSession();
+			for (String postgresSequence : postgresSequences.split(";"))
```
dkayiwa, line 340: instead of looping one by one, cant we submit them as a batch?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -301,4 +302,43 @@ public boolean isDatabaseStringComparisonCaseSensitive() {
...
+				session.createNativeQuery(postgresSequence).list();
```
dkayiwa, line 341: Why .list()?

Am-Coder: Because each of the query is a "SELECT" query I used .list() . We can also use .uniqueResult() since each of these queries return single value. I have not mad use of ALTER sequence statement here since it was not allowing the subqueries in it as it is a DDL statement.

dkayiwa: Did you explore `createSQLQuery().executeUpdate()`?

Am-Coder: Since the statements are of "SELECT" type , executeUpdate() raises exception. Link to logs: https://pastebin.com/WAYJznU2

dkayiwa: How about using it with `ALTER SEQUENCE`?

Am-Coder: We cannot make use of ALTER sequence statement here since it was not allowing the subqueries in it as it is a DDL statement. We require a subquery to get max id from table.

PR #3543, opened 2020-08-27
TRUNK-5898:Add support for JSON MIME Type

jwnasambu:
I changed the  `getFileMimeType()` to `getFileMimeType(File file)` in the OpenmrsUtil class and made changes to other  classes to use the same  by swapping the line String` mimeType = getServletContext().getMimeType(f.getName());`  to this:
String` mimeType = OpenmrsUtil.getFileMimeType(f);`

## Issue I worked on
https://issues.openmrs.org/browse/TRUNK-5898

### DIFF
--- a/api/src/main/java/org/openmrs/util/OpenmrsUtil.java
+++ b/api/src/main/java/org/openmrs/util/OpenmrsUtil.java
@@ -125,7 +125,7 @@ import org.w3c.dom.DocumentType;
  */
 public class OpenmrsUtil {
 
-	private OpenmrsUtil() {
+	private static volatile MimetypesFileTypeMap mimetypesFileTypeMap = null; {
 	}
 	
 	private static org.slf4j.Logger log = LoggerFactory.getLogger(OpenmrsUtil.class);
@@ -308,8 +308,12 @@ public class OpenmrsUtil {
 	 * @return mime type
 	 */
 	public static String getFileMimeType(File file) {
-		MimetypesFileTypeMap mimeMap = new MimetypesFileTypeMap();
-		return mimeMap.getContentType(file);
+		if (mimetypesFileTypeMap == null) {
+			synchronized (OpenmrsUtil.class) {
+				mimetypesFileTypeMap = new MimetypesFileTypeMap();
+			}
+		}
+		return mimetypesFileTypeMap.getContentType(file);
 	}
 	
 	/**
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -125,7 +125,7 @@
  */
 public class OpenmrsUtil {
 
-	private OpenmrsUtil() {
+	private static volatile MimetypesFileTypeMap mimetypesFileTypeMap = null; {
```
ibacher, line 128:
```suggestion
	private OpenmrsUtil() {
```

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -125,7 +125,7 @@
...
 	}
 	
 	private static org.slf4j.Logger log = LoggerFactory.getLogger(OpenmrsUtil.class);
```
ibacher, line 131: I'd add the new variable here after the `timeFormatCache` variable:

```java
private static volatile MimetypesFileTypeMap mimetypesFileTypeMap = null;
```

web/src/main/java/org/openmrs/module/web/ModuleResourcesServlet.java
```diff
@@ -25,13 +25,13 @@
 import org.slf4j.LoggerFactory;
 
 public class ModuleResourcesServlet extends HttpServlet {
-
```
ibacher, line 28: Can we get rid of most of these formatting changes, so that this is just showing the single line that's changed here.

ibacher (changes requested):
@jwnasambu Thanks for this. This does exactly what we need to do to core, just needs a few corrections.

webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -294,5 +294,10 @@
 	        <http-only>true</http-only>
 	    </cookie-config>
 	</session-config>
+	
+	<mime-mapping>
+	<extension>json</extension>
+	<mime-type>application/json</mime-type>
+</mime-mapping>
```
ibacher, line 301:
```suggestion
	<mime-mapping>
		<extension>json</extension>
		<mime-type>application/json</mime-type>
	</mime-mapping>
```

ibacher (commented):
@jwnasambu So the mapping part looks right, but we need to address:

* [This](https://github.com/openmrs/openmrs-core/pull/3543/files#r479385446)
* [This](https://github.com/openmrs/openmrs-core/pull/3543/files#r479385822)
* And [this](https://github.com/openmrs/openmrs-core/pull/3543/files#r480296301).

PR #3546, opened 2020-08-28
TRUNK-5409:Implement DosingInstructions to ChemoDosingInstructions

gitcliff:
link:https://issues.openmrs.org/browse/TRUNK-5409

### DIFF
--- a/api/src/main/java/org/openmrs/ChemoDosingInstructions.java
+++ b/api/src/main/java/org/openmrs/ChemoDosingInstructions.java
@@ -1,0 +1,191 @@
+
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License, 
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can 
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under 
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS 
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+
+package org.openmrs;
+
+import java.util.Date;
+import java.util.Locale;
+import java.util.StringTokenizer;
+import org.openmrs.api.APIException;
+import org.springframework.validation.Errors;
+import org.springframework.validation.ValidationUtils;
+
+/**
+ * @since 2.4.0
+ */
+public class ChemoDosingInstructions implements DosingInstructions {
+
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/ChemoDosingInstructions.java
```diff
@@ -0,0 +1,191 @@
...
+		}
+		ChemoDosingInstructions chemoDosingInstructions = new ChemoDosingInstructions();
+
+		// dosing instructions in drug order consist of several lines each having a
+		// different type of instruction
+		StringTokenizer tokenizeDosingInstructions = new StringTokenizer(order.getDosingInstructions(), "\n");
+		while (tokenizeDosingInstructions.hasMoreTokens()) {
+
+			// Every instruction line is of the form "instructionLineType :
+			// instructionLineValue"
```
JudeNiroshan, line 91: make it inline

api/src/main/java/org/openmrs/ChemoDosingInstructions.java
```diff
@@ -0,0 +1,191 @@
...
+			String instructionLine = tokenizeDosingInstructions.nextToken();
+			int indexOfColon = instructionLine.indexOf(':');
+			String instructionLineType = instructionLine.substring(0, indexOfColon - 1);
```
JudeNiroshan, line 94: why `indexOfColon - 1`? I assume you want to avoid space at the end. It's better to use `trim()` instead of making hard rules that there will be a space always.

api/src/main/java/org/openmrs/ChemoDosingInstructions.java
```diff
@@ -0,0 +1,191 @@
...
+			String instructionLineValue = instructionLine.substring(indexOfColon + 2);
```
JudeNiroshan, line 95: same indexing question here

api/src/test/java/org/openmrs/ChemoDosingInstructionsTest.java
```diff
@@ -0,0 +1,86 @@
...
+	private static final Double ADJUSTMENT_PERCENTAGE_CONSTANT = 20.04;
+
+	private static final String DELIVERED_TIMING_CONSTANT = "once in 2 hours";
+
+	private static final String ADJUSTMENT_PERCENTAGE = "Dosage Adjustment Percentage : ";
+
+	private static final String DELIVERED_TIMING = "Dosage Delivered Timing : ";
+
+	private static final String DILUTION_INSTRUCTIONS = "Dosage Dilution Instructions : ";
+
```
JudeNiroshan, line 33: consider figuring erroneous scenarios too.

PR #3572, opened 2020-09-16
TRUNK-5935: Utility classes should not have public constructors

sacull:

I have overwritten the default constructors to prevent the creation of an instance of the utility class.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5935

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Daemon.java
+++ b/api/src/main/java/org/openmrs/api/context/Daemon.java
@@ -48,6 +48,12 @@ public class Daemon {
 	protected static final ThreadLocal<User> daemonThreadUser = new ThreadLocal<>();
 	
 	/**
+	 * Protected constructor to override the default constructor to prevent it from being instantiated.
+	 */
+	protected Daemon() {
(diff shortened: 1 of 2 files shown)


dkayiwa:
Could you explain why one is protected but the other is private?

sacull:
@dkayiwa of course. 
A class that only has static properties and methods, as described on the SonarQube, should not be instantiated. All methods should be accessible from the class, not the object. Any class that does not have an explicit constructor gets a default constructor with the "public" access modifier, so anyone can create an object of that class. The easiest way to prevent this from happening is to override the default constructor with the "private" access modifier, which was done in the first class. However, the second class could not receive such a constructor because another class inherits from it, and such a constructor cannot be inherited. Fortunately, there is one more access modifier that prevents the object from being created while also being inherited. Such a modifier is "protected", which is why the second class received a constructor with this modifier.

sacull:
If anything else is unclear, I'll be happy to elaborate on it. ;)

PR #3579, opened 2020-09-17
TRUNK-5151:Code in voidRelationship is not executed due to BaseVoidHandler

jwnasambu:
 I removed 

```
if (relationship.getVoided()) {
			
return relationship;
		}
		
		relationship.setVoided(true);
		if (relationship.getVoidedBy() == null) {
			relationship.setVoidedBy(Context.getAuthenticatedUser());
		}
		if (voidReason != null) {
			relationship.setVoidReason(voidReason);
		}
		relationship.setDateVoided(new Date());
		
		return Context.getPersonService().saveRelationship(relationship);
```

 which is already  being called via AOP  the Relationship and  fixed the methods to effectively return null.

## Issue I worked on
https://issues.openmrs.org/browse/TRUNK-5151

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/PersonServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PersonServiceImpl.java
@@ -485,19 +485,10 @@ public class PersonServiceImpl extends BaseOpenmrsService implements PersonServi
 	 */
 	@Override
 	public Relationship voidRelationship(Relationship relationship, String voidReason) throws APIException {
-		if (relationship.getVoided()) {
-			return relationship;
+		if (relationship == null) {
+		    return null;
 		}
-		
-		relationship.setVoided(true);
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/api/impl/PersonServiceImpl.java
```diff
@@ -485,19 +485,10 @@ public Relationship saveRelationship(Relationship relationship) throws APIExcept
 	 */
 	@Override
 	public Relationship voidRelationship(Relationship relationship, String voidReason) throws APIException {
```
JudeNiroshan, line 487: `voidReason` is no longer being used. Maybe you can remove this parameter.
```suggestion
	public Relationship voidRelationship(@Nonnull final Relationship relationship) throws APIException {
```

JudeNiroshan (changes requested):

PR #3606, opened 2020-10-22
TRUNK-5970: Encounter to handle voided conditions.

icrc-loliveira:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5970

### DIFF
--- a/api/src/main/java/org/openmrs/Encounter.java
+++ b/api/src/main/java/org/openmrs/Encounter.java
@@ -555,6 +555,7 @@ public class Encounter extends BaseChangeableOpenmrsData {
 		Optional.ofNullable(conditions).orElse(new LinkedHashSet<>()).stream().filter(c -> !c.getVoided() && c.equals(condition)).forEach(c -> {
 			c.setVoided(true);
 			c.setDateVoided(new Date());
+			c.setVoidReason("Voided by the API");
 			c.setVoidedBy(Context.getAuthenticatedUser());
 		});
 	}
--- a/api/src/main/java/org/openmrs/Encounter.java
+++ b/api/src/main/java/org/openmrs/Encounter.java
@@ -557,7 +557,7 @@ public class Encounter extends BaseChangeableOpenmrsData {
 			conditions.add(condition);
 		}
 	}
-	
+
 	/**
 	 * Remove the given condition from the set of conditions for this Encounter
 	 *
@@ -569,14 +569,11 @@ public class Encounter extends BaseChangeableOpenmrsData {
 			return;
 		}
 
-		for (Condition encounterCondition : conditions) {
-			if (encounterCondition.equals(condition) && !encounterCondition.getVoided()) {
-				encounterCondition.setVoided(true);
-				encounterCondition.setDateVoided(new Date());
-				encounterCondition.setVoidedBy(Context.getAuthenticatedUser());
-				return;
-			}
-		}
+		conditions.stream().filter(c -> !c.getVoided()).filter(c -> c.equals(condition)).forEach(c -> {
+			c.setVoided(true);
+			c.setDateVoided(new Date());
+			c.setVoidedBy(Context.getAuthenticatedUser());
+		});
 	}
 	
 	/**
--- a/api/src/main/java/org/openmrs/Encounter.java
+++ b/api/src/main/java/org/openmrs/Encounter.java
@@ -491,32 +491,25 @@ public class Encounter extends BaseChangeableOpenmrsData {
 	}
 	
 	/**
-	 * Basic property getter for conditions
+	 * Basic property getter for the encounter's non-voided conditions.
 	 * 
 	 * @return all non-voided conditions
-	 * <strong>Should</strong> not get voided conditions
 	 * @since 2.4.0, 2.3.1
 	 */
 	public Set<Condition> getConditions() {
-		
 		return getConditions(false);
 	}
 
 	/**
-	 * Returns all conditions where conditions.encounterId = Encounter.encounterId In practice, this method should
-	 * not be used very often...
+	 * Returns all conditions where 'Condition.encounterId = Encounter.encounterId'.
 	 *
-	 * @param includeVoided specifies whether or not to include voided Conditions
-	 * @return Returns the all Conditions.
-	 * <strong>Should</strong> not return null with null conditions set
-	 * <strong>Should</strong> get conditions
+	 * @param includeVoided - Specifies whether or not to include voided conditions
+	 * @return Returns the set of conditions, or an empty set if there are no conditions to return
 	 * @since 2.5.0
 	 */
 	public Set<Condition> getConditions(boolean includeVoided) {
-
 		return Optional.ofNullable(conditions).orElse(new LinkedHashSet<>())
-			.stream().filter(o -> includeVoided || !o.getVoided()).collect(Collectors.toSet());
-
+			.stream().filter(c -> includeVoided || !c.getVoided()).collect(Collectors.toSet());
 	}
 		
 	/**
@@ -530,7 +523,7 @@ public class Encounter extends BaseChangeableOpenmrsData {
 	}
 
 	/**
-	 * Add the given Condition to the set of conditions for this Encounter
+	 * Add the given condition to the set of conditions for this encounter.
 	 *
 	 * @param condition - the condition to add
 	 */
@@ -546,12 +539,12 @@ public class Encounter extends BaseChangeableOpenmrsData {
 	}
 
 	/**
-	 * Remove the given condition from the set of conditions for this Encounter
+	 * Remove the given condition from the set of conditions for this encounter.
+	 * In practise the condition is not removed but rather voided.
 	 *
 	 * @param condition - the condition to remove
 	 */
 	public void removeCondition(Condition condition) {
-		
 		Optional.ofNullable(conditions).orElse(new LinkedHashSet<>()).stream().filter(c -> !c.getVoided() && c.equals(condition)).forEach(c -> {
 			c.setVoided(true);
 			c.setDateVoided(new Date());
--- a/api/src/main/java/org/openmrs/Encounter.java
+++ b/api/src/main/java/org/openmrs/Encounter.java
@@ -569,7 +569,7 @@ public class Encounter extends BaseChangeableOpenmrsData {
 			return;
 		}
 
-		conditions.stream().filter(c -> !c.getVoided()).filter(c -> c.equals(condition)).forEach(c -> {
+		conditions.stream().filter(c -> !c.getVoided() && c.getUuid().equals(condition.getUuid())).forEach(c -> {
 			c.setVoided(true);
 			c.setDateVoided(new Date());
 			c.setVoidedBy(Context.getAuthenticatedUser());
--- a/api/src/main/java/org/openmrs/Encounter.java
+++ b/api/src/main/java/org/openmrs/Encounter.java
@@ -9,7 +9,16 @@
  */
 package org.openmrs;
 
-import java.util.*;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.Deque;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import javax.persistence.Access;
--- a/api/src/main/java/org/openmrs/Encounter.java
+++ b/api/src/main/java/org/openmrs/Encounter.java
@@ -9,15 +9,7 @@
  */
 package org.openmrs;
 
-import java.util.ArrayDeque;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.Deque;
-import java.util.HashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
 
 import javax.persistence.Access;
@@ -492,12 +484,13 @@ public class Encounter extends BaseChangeableOpenmrsData {
 	/**
 	 * Basic property getter for conditions
 	 * 
-	 * @return conditions - set of conditions
+	 * @return all non-voided conditions
+	 * <strong>Should</strong> not get voided conditions
 	 * @since 2.4.0, 2.3.1
 	 */
 	public Set<Condition> getConditions() {
-
-		return getAllConditions();
+		
+		return getConditions(false);
 	}
 
 	/**
@@ -509,29 +502,13 @@ public class Encounter extends BaseChangeableOpenmrsData {
 	 * <strong>Should</strong> not return null with null conditions set
 	 * <strong>Should</strong> get conditions
 	 */
-	public Set<Condition> getAllConditions(boolean includeVoided) {
-		if (includeVoided && conditions != null) {
-			return conditions;
-		}
+	public Set<Condition> getConditions(boolean includeVoided) {
 
-		Set<Condition> ret = new LinkedHashSet<>();
+		return Optional.ofNullable(conditions).orElse(new LinkedHashSet<>())
+			.stream().filter(o -> includeVoided || !o.getVoided()).collect(Collectors.toSet());
 
-		if (this.conditions != null) {
-			ret = this.conditions.stream().filter(o -> includeVoided || !o.getVoided()).collect(Collectors.toSet());
-		}
-		return ret;
 	}
-
-	/**
-	 * Convenience method to call {@link #getAllConditions(boolean)} with a false parameter
-	 *
-	 * @return all non-voided conditions
-	 * <strong>Should</strong> not get voided conditions
-	 */
-	public Set<Condition> getAllConditions() {
-		return getAllConditions(false);
-	}
-	
+		
 	/**
 	 * Basic property setter for conditions
 	 *  
@@ -564,12 +541,8 @@ public class Encounter extends BaseChangeableOpenmrsData {
 	 * @param condition - the condition to remove
 	 */
 	public void removeCondition(Condition condition) {
-
-		if (conditions == null) {
-			return;
-		}
-
-		conditions.stream().filter(c -> !c.getVoided() && c.getUuid().equals(condition.getUuid())).forEach(c -> {
+		
+		Optional.ofNullable(conditions).orElse(new LinkedHashSet<>()).stream().filter(c -> !c.getVoided() && c.getUuid().equals(condition.getUuid())).forEach(c -> {
 			c.setVoided(true);
 			c.setDateVoided(new Date());
 			c.setVoidedBy(Context.getAuthenticatedUser());
--- a/api/src/main/java/org/openmrs/Encounter.java
+++ b/api/src/main/java/org/openmrs/Encounter.java
@@ -510,6 +510,7 @@ public class Encounter extends BaseChangeableOpenmrsData {
 	 * @return Returns the all Conditions.
 	 * <strong>Should</strong> not return null with null conditions set
 	 * <strong>Should</strong> get conditions
+	 * @since 2.5.0
 	 */
 	public Set<Condition> getConditions(boolean includeVoided) {
 
@@ -551,7 +552,7 @@ public class Encounter extends BaseChangeableOpenmrsData {
 	 */
 	public void removeCondition(Condition condition) {
 		
-		Optional.ofNullable(conditions).orElse(new LinkedHashSet<>()).stream().filter(c -> !c.getVoided() && c.getUuid().equals(condition.getUuid())).forEach(c -> {
+		Optional.ofNullable(conditions).orElse(new LinkedHashSet<>()).stream().filter(c -> !c.getVoided() && c.equals(condition)).forEach(c -> {
 			c.setVoided(true);
 			c.setDateVoided(new Date());
 			c.setVoidedBy(Context.getAuthenticatedUser());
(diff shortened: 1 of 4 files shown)


api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -1397,4 +1397,85 @@ public void hasDiagnosis_shouldReturnFalseIfEncounterDoesNotHaveDiagnosis(){
...
+		freeText01.setNonCoded("Free text");
+		voidedCondition.setCondition(freeText01);
+		encounter.addCondition(voidedCondition);
+
+
+		Set<Condition> a = encounter.getConditions();
+		Set<Condition> b = encounter.getAllConditions(true);
+		assertEquals(0,encounter.getConditions().size());
+		assertEquals(2,encounter.getAllConditions(true).size());
+	}
```
mks-d, line 1480: Let's cleanup this class a little. I believe that 1) it should not extend `BaseContextSensitiveTest` and that 2) the two mocked services are not needed.

Can you confirm those two points?

Then I'd simplify your test so that you setup your encounter with a bunch of conditions in a setup method so that you can work on a reference encounter across all your test cases.

api/src/main/java/org/openmrs/Encounter.java
```diff
@@ -496,12 +496,42 @@ public void setDiagnoses(Set<Diagnosis> diagnoses) {
...
+	/**
+	 * Returns all conditions where conditions.encounterId = Encounter.encounterId In practice, this method should
+	 * not be used very often...
+	 *
+	 * @param includeVoided specifies whether or not to include voided Conditions
+	 * @return Returns the all Conditions.
+	 * <strong>Should</strong> not return null with null conditions set
+	 * <strong>Should</strong> get conditions
+	 */
+	public Set<Condition> getAllConditions(boolean includeVoided) {
```
mks-d, line 512: Just name this `getCondition(Boolean)`, there is no need for a different name.

I know you borrowed this from the logic for the obs getters, but obs is a different situation, the distinction between "obs" and "all obs" is "top level" vs "including nested in groups" obs.

api/src/main/java/org/openmrs/Encounter.java
```diff
@@ -496,12 +496,42 @@ public void setDiagnoses(Set<Diagnosis> diagnoses) {
...
+		if (includeVoided && conditions != null) {
+			return conditions;
+		}
+
+		Set<Condition> ret = new LinkedHashSet<>();
+
+		if (this.conditions != null) {
+			ret = this.conditions.stream().filter(o -> includeVoided || !o.getVoided()).collect(Collectors.toSet());
 		}
```
mks-d, line 521: The whole thing could be done with a one-liner:
```java
return Optional.of(conditions).orElse(new LinkedHashSet<>())
  .stream().filter(o -> includeVoided || !o.getVoided()).collect(Collectors.toSet());
```

api/src/main/java/org/openmrs/Encounter.java
```diff
@@ -518,19 +548,32 @@ public void setConditions(Set<Condition> conditions) {
...
 	 */
 	public void removeCondition(Condition condition) {
-		if (conditions != null) {
-			conditions.remove(condition);
+
+		if (conditions == null) {
+			return;
 		}
+
+		conditions.stream().filter(c -> !c.getVoided() && c.getUuid().equals(condition.getUuid())).forEach(c -> {
```
mks-d, line 572: Use `Optional.of(conditions).orElse(new LinkedHashSet<>())`.

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -1397,4 +1401,80 @@ public void hasDiagnosis_shouldReturnFalseIfEncounterDoesNotHaveDiagnosis(){
...
+		activeCondition.setCondition(freeText);
+		encounter.addCondition(activeCondition);
+
+		Condition voidedCondition = new Condition();
+		voidedCondition.setVoided(true);
+		voidedCondition.setClinicalStatus(ConditionClinicalStatus.HISTORY_OF);
+		CodedOrFreeText freeText01 = new CodedOrFreeText();
+		freeText01.setNonCoded("Free text");
+		voidedCondition.setCondition(freeText01);
+		encounter.addCondition(voidedCondition);
```
mks-d, line 1423: Put all this in the `before()` method.

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -1397,4 +1401,80 @@ public void hasDiagnosis_shouldReturnFalseIfEncounterDoesNotHaveDiagnosis(){
...
+
+		assertEquals(2, encounter.getAllConditions(true).size());
```
mks-d, line 1425: Also assert that you get the conditions that you are supposed to get, not just the size of the resulting collection.

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -1397,4 +1401,80 @@ public void hasDiagnosis_shouldReturnFalseIfEncounterDoesNotHaveDiagnosis(){
...
+
+		Condition voidedCondition = new Condition();
+		voidedCondition.setVoided(true);
+		voidedCondition.setClinicalStatus(ConditionClinicalStatus.HISTORY_OF);
+		CodedOrFreeText freeText01 = new CodedOrFreeText();
+		freeText01.setNonCoded("Free text");
+		voidedCondition.setCondition(freeText01);
+		encounter.addCondition(voidedCondition);
+
+		assertEquals(1, encounter.getConditions().size());
```
mks-d, line 1449: Same remarks.

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -1397,4 +1401,80 @@ public void hasDiagnosis_shouldReturnFalseIfEncounterDoesNotHaveDiagnosis(){
...
+		voidedCondition.setClinicalStatus(ConditionClinicalStatus.HISTORY_OF);
+		CodedOrFreeText freeText01 = new CodedOrFreeText();
+		freeText01.setNonCoded("Free text");
+		voidedCondition.setCondition(freeText01);
+		encounter.addCondition(voidedCondition);
+
+		Set<Condition> a = encounter.getConditions();
+		Set<Condition> b = encounter.getAllConditions(true);
+		assertEquals(0, encounter.getConditions().size());
+		assertEquals(2, encounter.getAllConditions(true).size());
```
mks-d, line 1478: Same remarks.

mks-d (changes requested):

api/src/main/java/org/openmrs/Encounter.java
```diff
@@ -9,15 +9,7 @@
...
-import java.util.ArrayDeque;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.Deque;
-import java.util.HashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
mks-d, line 12: Avoid wildcard imports.

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -1406,75 +1425,31 @@ public void hasDiagnosis_shouldReturnFalseIfEncounterDoesNotHaveDiagnosis() {
...
-		voidedCondition.setVoided(true);
-		voidedCondition.setClinicalStatus(ConditionClinicalStatus.HISTORY_OF);
-		CodedOrFreeText freeText01 = new CodedOrFreeText();
-		freeText01.setNonCoded("Free text");
-		voidedCondition.setCondition(freeText01);
-		encounter.addCondition(voidedCondition);
-
-		assertEquals(2, encounter.getAllConditions(true).size());
+		
+		assertEquals(2, encounter.getConditions(true).size());
```
mks-d, line 1429: Do not just assert the size, assert that the content is the one that is expected.

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -1406,75 +1425,31 @@ public void hasDiagnosis_shouldReturnFalseIfEncounterDoesNotHaveDiagnosis() {
...
-		Condition voidedCondition = new Condition();
-		voidedCondition.setVoided(true);
-		voidedCondition.setClinicalStatus(ConditionClinicalStatus.HISTORY_OF);
-		CodedOrFreeText freeText01 = new CodedOrFreeText();
-		freeText01.setNonCoded("Free text");
-		voidedCondition.setCondition(freeText01);
-		encounter.addCondition(voidedCondition);
-
+		
 		assertEquals(1, encounter.getConditions().size());
```
mks-d, line 1440: Same here.

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -1406,75 +1425,31 @@ public void hasDiagnosis_shouldReturnFalseIfEncounterDoesNotHaveDiagnosis() {
...
-		assertEquals(2, encounter.getAllConditions(true).size());
+		
+		assertEquals(2, encounter.getConditions(true).size());
 	}
 
 	/**
 	 * @see Encounter#getConditions()
 	 */
 	@Test
 	public void getConditions_shouldReturnActiveConditions() {
```
mks-d, line 1438: `getConditions_shouldReturnNonVoidedConditionsByDefault`

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -1406,75 +1425,31 @@ public void hasDiagnosis_shouldReturnFalseIfEncounterDoesNotHaveDiagnosis() {
 	 */
 	@Test
 	public void getAllConditions_shouldReturnAllConditions() {
```
mks-d: `getConditions_shouldReturnAllConditionsWhenFlagIsTrue`

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -1406,75 +1425,31 @@ public void hasDiagnosis_shouldReturnFalseIfEncounterDoesNotHaveDiagnosis() {
...
-		voidedCondition.setCondition(freeText01);
-		encounter.addCondition(voidedCondition);
-
 		Set<Condition> a = encounter.getConditions();
-		Set<Condition> b = encounter.getAllConditions(true);
+		Set<Condition> b = encounter.getConditions(true);
 		assertEquals(0, encounter.getConditions().size());
-		assertEquals(2, encounter.getAllConditions(true).size());
+		assertEquals(2, encounter.getConditions(true).size());
 	}
```
mks-d, line 1458: What is this test actually testing?

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -43,10 +43,29 @@
...
+	private Condition voidedCondition;
 
 	@BeforeEach
 	public void before() {
 		encounter = new Encounter();
+
+		activeCondition = new Condition();
+		activeCondition.setClinicalStatus(ConditionClinicalStatus.ACTIVE);
+		CodedOrFreeText freeText = new CodedOrFreeText();
+		freeText.setNonCoded("Free text");
```
mks-d, line 58: Give it a real name, that helps people understand what things are about:
```java
freeText1.setNonCoded("Asthma non-coded");
```

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -43,10 +43,29 @@
...
+	
+	private Condition voidedCondition;
 
 	@BeforeEach
 	public void before() {
 		encounter = new Encounter();
+
+		activeCondition = new Condition();
+		activeCondition.setClinicalStatus(ConditionClinicalStatus.ACTIVE);
+		CodedOrFreeText freeText = new CodedOrFreeText();
```
mks-d, line 57: Let's name them `freeText1` and `freeText2`.

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -43,10 +43,29 @@
...
+		freeText.setNonCoded("Free text");
+		activeCondition.setCondition(freeText);
+		encounter.addCondition(activeCondition);
+
+		voidedCondition = new Condition();
+		voidedCondition.setVoided(true);
+		voidedCondition.setClinicalStatus(ConditionClinicalStatus.HISTORY_OF);
+		CodedOrFreeText freeText01 = new CodedOrFreeText();
+		freeText01.setNonCoded("Free text");
```
mks-d, line 66: Same here:
```java
freeText2.setNonCoded("Glaucoma non-coded");
```

api/src/test/java/org/openmrs/EncounterTest.java
icrc-loliveira: It's already been done in one of the older commits.

api/src/test/java/org/openmrs/EncounterTest.java
icrc-loliveira: It's already been done in one of the older commits.

api/src/test/java/org/openmrs/EncounterTest.java
icrc-loliveira: It's already been done in one of the older commits.

api/src/test/java/org/openmrs/EncounterTest.java
icrc-loliveira: That the conditions are voided and not deleted.

icrc-loliveira: Done.

api/src/test/java/org/openmrs/EncounterTest.java
icrc-loliveira: That the conditions are voided and not deleted.

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -1397,4 +1420,40 @@ public void hasDiagnosis_shouldReturnFalseIfEncounterDoesNotHaveDiagnosis(){
...
+	/**
+	 * @see Encounter#removeCondition(Condition)
+	 */
+	@Test
+	public void removeCondition_shouldRemoveConditions() {
+
+		// The condition should be voided
+		encounter.removeCondition(activeCondition);
+
+		Set<Condition> a = encounter.getConditions();
```
mks-d, line 1454: Rename `a` to `conditions` or `allConditions`.
Rename `b` to `voidedConditions`.

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -414,9 +414,9 @@ public void saveEncounter_shouldCascadeSaveToContainedConditions() {
 		edema = (Condition) conditionsArray[1];
 		assertEquals(2, conditionsArray.length);
 		assertNotNull(pregnancy.getId());
-		assertEquals("Pregnancy", pregnancy.getCondition().getNonCoded());
+		assertNotNull(pregnancy.getCondition().getNonCoded());
 		assertNotNull(edema.getId());
-		assertEquals("Edema", edema.getCondition().getNonCoded());
+		assertNotNull(edema.getCondition().getNonCoded());
```
mks-d, line 419: Did you need to touch that test?

mks-d (changes requested):

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
icrc-loliveira: Yes.
The way that it was done was inconsistent. 
Sometimes they would come flipped, Emma would be Pregnancy and Pregnancy would be Emma.

mks-d: @icrc-loliveira this is burying the issue. We need to understand why our changes are affecting an order that used to be stable.

mks-d (changes requested):
Can you also add a `@since` JavaDoc for all new method and all new method signatures:
```java
@since 2.5.0
```

api/src/main/java/org/openmrs/Encounter.java
```diff
@@ -492,16 +493,31 @@ public void setDiagnoses(Set<Diagnosis> diagnoses) {
...
 	 * 
-	 * @return conditions - set of conditions
+	 * @return all non-voided conditions
+	 * <strong>Should</strong> not get voided conditions
 	 * @since 2.4.0, 2.3.1
 	 */
 	public Set<Condition> getConditions() {
-		if (conditions == null) {
-			conditions = new LinkedHashSet<>();
-		}
```
ibacher, line 501: While the implementation here is simpler, it does potentially break existing code which relies on being able to do, say, `getConditions().add(newCondition)`.

api/src/main/java/org/openmrs/Encounter.java
```diff
@@ -518,19 +534,28 @@ public void setConditions(Set<Condition> conditions) {
...
 	 * Remove the given condition from the set of conditions for this Encounter
 	 *
 	 * @param condition - the condition to remove
 	 */
 	public void removeCondition(Condition condition) {
-		if (conditions != null) {
-			conditions.remove(condition);
-		}
+		
+		Optional.ofNullable(conditions).orElse(new LinkedHashSet<>()).stream().filter(c -> !c.getVoided() && c.getUuid().equals(condition.getUuid())).forEach(c -> {
```
ibacher, line 554: Shouldn't we just be explicitly relying on the notion of equality here, e.g., `filter(c -> !c.getVoided && condition.equals(c))`?

api/src/main/java/org/openmrs/Encounter.java
```diff
@@ -518,19 +534,28 @@ public void setConditions(Set<Condition> conditions) {
...
+			c.setVoided(true);
+			c.setDateVoided(new Date());
+			c.setVoidedBy(Context.getAuthenticatedUser());
```
ibacher, line 557: We should have a void reason when voiding the condition, even if it's as simple as "Voided by the API"

api/src/main/java/org/openmrs/Encounter.java
ibacher: I know that `condition.equals()` doesn't work that way right now, but I thought it was agreed that we'd proceed with an implementation of `condition.equals()` that only verifies UUID.

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -1397,4 +1420,40 @@ public void hasDiagnosis_shouldReturnFalseIfEncounterDoesNotHaveDiagnosis(){
...
+	@Test
+	public void removeCondition_shouldRemoveConditions() {
+
+		// The condition should be voided
+		encounter.removeCondition(activeCondition);
+
+		Set<Condition> allConditions = encounter.getConditions();
+		Set<Condition> voidedConditions = encounter.getConditions(true);
+		assertEquals(0, encounter.getConditions().size());
+		assertEquals(2, encounter.getConditions(true).size());
```
ibacher, line 1457: Shouldn't we also verify that one of the voided conditions has the same UUID as the `activeCondition`?

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
icrc-loliveira: Used equal method.

api/src/main/java/org/openmrs/Encounter.java
icrc-loliveira: Done.

api/src/test/java/org/openmrs/EncounterTest.java
icrc-loliveira: done.

api/src/main/java/org/openmrs/Encounter.java
icrc-loliveira: Done.

api/src/main/java/org/openmrs/Encounter.java
mks-d: @ibacher you're right, but given that there is a `addCondition(Condition)` method, I guess we should assume that this is just bad practise no?

api/src/main/java/org/openmrs/Encounter.java
mks-d: @ibacher I didn't find any good example in Core, so we went for the message that you suggested.

api/src/main/java/org/openmrs/Encounter.java
ibacher: @mks-d I suppose that's fair. And realistically, I don't expect there to be much actual code out there that uses this.

ibacher (approved):

ibacher:
I've approved as my requested changes have been addressed. Thanks @icrc-loliveira! We still should ensure all of @mks-d's changes are made before moving this in.

api/src/test/java/org/openmrs/api/EncounterServiceTest.java
```diff
@@ -409,14 +409,10 @@ public void saveEncounter_shouldCascadeSaveToContainedConditions() {
...
 		// verify
-		Object[] conditionsArray = encounter.getConditions().toArray();
-		pregnancy = (Condition) conditionsArray[0];
-		edema = (Condition) conditionsArray[1];
-		assertEquals(2, conditionsArray.length);
-		assertNotNull(pregnancy.getId());
-		assertEquals("Pregnancy", pregnancy.getCondition().getNonCoded());
-		assertNotNull(edema.getId());
-		assertEquals("Edema", edema.getCondition().getNonCoded());
+		Set<Condition> savedConditions = encounter.getConditions();
```
ibacher, line 412: I'd honestly feel more comfortable with the changes here if we can prove that the same test still passes or, if it doesn't, that the reason it doesn't pass is because of an expected change in behaviour.

mks-d: @ibacher I had the same concern yesterday. But looking closer, it appears that there is no specific care being given at the order in which those conditions are returned (before or after this PR). So I suggested to @icrc-loliveira to simply test that those two conditions were part of the returned set.

ibacher: @mks-d That's fair. My main worry was that the condition non-coded value might not be being preserved (since that's no longer verified). Order-wise, it's probably not too important.

mks-d (approved):

PR #3619, opened 2020-10-28
TRUNK-4830: Fix a potential StackOverflowError

ibacher:

Small bugfix for TRUNK-4830, fixing a StackOverflowError. The downside to this is that we lose the ability to call Liquibase files with user-provided parameters. I'm not sure we were actually using that capability anywhere?

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-4830

  No? -> execute above command


ibacher:
@WolfSchlegel If you have a chance, I'd really appreciate you taking a look at this.

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -1189,11 +1189,26 @@ private static void checkForDatabaseUpdates(Properties props) throws DatabaseUpd
...
-	 * @throws InputRequiredException if user input is required
 	 * @since 1.5
+	 * @deprecated as of 2.4; see {@link #updateDatabase()}
+	 * 
+	 */
+	@Deprecated
+	public static void updateDatabase(Map<String, Object> userInput) throws DatabaseUpdateException {
+		DatabaseUpdater.executeChangelog();
+	}
+
```
WolfSchlegel, line 1200: Calling the deprecated method results in an `IllegalArgumentException`. I wonder if an `UnsupportedOperationException` would be more appropriate, with a message pointing out that the method called had been deprecated.

When overhauling the Liquibase setup it took some time to figure out all the places where things were broken by the change. I am not aware of situations where user input is used to set up the database but the only way to find out whether anyone is doing this is deprecating the method and breaking the respective builds (hopefully before they go live).

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -1189,11 +1189,26 @@ private static void checkForDatabaseUpdates(Properties props) throws DatabaseUpd
...
+	 * if that fails, call this method to get the database up to speed.
+	 * 
+	 * @throws DatabaseUpdateException if an error occurred while updating
+	 * @since 2.4
 	 */
-	public static void updateDatabase(Map<String, Object> userInput) throws DatabaseUpdateException, InputRequiredException {
-		DatabaseUpdater.executeChangelog(null, userInput);
+	public static void updateDatabase() throws DatabaseUpdateException {
+		DatabaseUpdater.executeChangelog();
 	}
```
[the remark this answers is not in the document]
ibacher: To be honest, I think this existed to run the database updates if either the initialization filter or update filter didn't get called. I really can't find any example of this actually being used anywhere. Good point on the illegal argument exception. I was mostly concerned with eliminating the endless recursion [here](https://github.com/openmrs/openmrs-core/pull/3619/files#diff-3f2d7ddb866d3834bf8e662b8c79a69a8e85808c2ddbf17c9493310d9d86c5a2L145). Maybe the thing to do is just remove these methods and see what breaks?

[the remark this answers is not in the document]
WolfSchlegel: `public static void updateDatabase(Map<String, Object> userInput)` existed before and is now deprecated which makes sense.

`public static void updateDatabase()` is new since 2.4 and I was wondering why it is added. 

Definitely a +1 on getting rid of the the strange recursion.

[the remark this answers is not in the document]
WolfSchlegel: >> Maybe the thing to do is just remove these methods and see what breaks?

Yes, but breaking it "gently" by throwing exceptions but still allowing any code using this method to compile.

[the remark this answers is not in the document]
ibacher: Well, I missed the fact that this code would throw an exception, so the new method was really just intended to replace `public static void updateDatabase(Map<String, Object> userInput)` with a version with the arguments we actually use (none). Obviously, that's also a bad idea!

api/src/main/java/org/openmrs/api/context/Context.java
ibacher: So, it looks like we call the `DatabaseUpdater.executeChangelog();` [here](https://github.com/openmrs/openmrs-core/blob/7492b7abc7dc228f656987e129fa46f87024fa12/api/src/main/java/org/openmrs/api/context/Context.java#L1176) as part of the [startup](https://github.com/openmrs/openmrs-core/blob/7492b7abc7dc228f656987e129fa46f87024fa12/api/src/main/java/org/openmrs/api/context/Context.java#L879), so we'll likely need to fix that too.

ibacher: I've updated this method to just throw an `UnsupportedOperationException`. On the other hand, I've now got an actual implementation for `DatabaseUpdater.executeChangelog()`, which may or may not be the desirable thing to do.

mogoodrich: For what it's worth, this method *was* being hit as part of the PIH EMR startup, but I didn't look into why...

mogoodrich: Or, actually, the 2-parameter version of this method...

WolfSchlegel: The implementation of `DatabaseUpdater.executeChangelog()` looks good to me. I tried to test it by initialising the database with OpenMRS 2.3.x and then restarting OpenMRS based on this branch but could not reach `DatabaseUpdater.executeChangelog()`.

So by looking at the code very hard we feel it is ok but there is a chance that it fails in the wild. To prepare for that it would be good to add logging to `DatabaseUpdater.executeChangelog()`, similar to `org.openmrs.liquibase.ChangeLogDetective` which borrows the log level from the `org.openmrs.api` package.

The log statement should inform us about the `version` and `changeLogs` variables.

ibacher: @WolfSchlegel I've added logging to the latest commit

WolfSchlegel: @ibacher thanks for nudging.

Anything logged at **debug** level will be invisible, so if things go wrong in the wild we can only wish we had more logging... 

I would suggest one of two things:

- log version and a limited number of un-run change logs at **info** level, regardless whether an exception is raised or not
- OR log these values in the exceptional case only

Please have a look [here](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/liquibase/ChangeLogDetective.java), implementing this approach in the `ChangeLogDetective` paid large dividends when we rolled out the Liquibase overhaul

WolfSchlegel: Any thoughts on the log settings please?

ibacher: @WolfSchlegel So I modeled the logging I added on what happens in the UpdateFilter, since that seems the most similar to where this is used. In particular, since we're just using calls to `changeLogDetective` to get the unrun changes, etc. and then passing the call to `executeChangelog()` to the same function, we should already be getting the same levels of logging we get with Liquibase currently in the case that this method gets called (I'm a little fuzzy on how this works, but it looks like we get to this chain on runs after the `openmrs-runtime.properties` file exists).

WolfSchlegel: That makes sense, I had not thought about the implicit logging that happens in the `ChangeLogDetective`.

The one thing that is still missing for me is a scenario where the changed methods are actually run. I had tried to see them in action during the initialisation and restart of OpenMRS but without success.

@ibacher and @mogoodrich are you aware of such a scenario? If so I am happy to run a few tests.

Apart from that, is there anything I can do for this pull request? This is the first time I am reviewing a PR for OpenMRS so I am not sure about the next steps.

ibacher: So I think to actually hit that method you'd have to:

1. Create a new server built on the 2.3 series of the platform.
2. Upgrade the platform to a version of the core with these changes.
3. Wait and definitely do not make any web requests to the server for a bit.

Essentially, the only place that uses the `executeChangelog()` call is [in `Context#checkForDatabaseUpdates()`](https://github.com/openmrs/openmrs-core/blob/4d7269120aef6f928900c87a9bd243b4a7572286/api/src/main/java/org/openmrs/api/context/Context.java#L1176), which is called during the start-up process, but is only executed if the [prior call to `DatabaseUpdater#updatesRequired()`](https://github.com/openmrs/openmrs-core/blob/4d7269120aef6f928900c87a9bd243b4a7572286/api/src/main/java/org/openmrs/api/context/Context.java#L1167) reports that there are unrun liquibase changes. So the idea is to cause there to be unrun liquibase changes and to *not* hit the update filter, which will also cause the changes to be run.

I'm not 100% certain that's right, but it's an educated guess.

You don't really have to do much else, unless there are other changes you think this code needs. Mostly I was checking in with you because you are currently the person most familiar with our liquibase processes 😁.

ibacher: Ok, so having dug a little deeper, I figured out how to actually trigger the code in question. Basically, you need to have:

1. Unrun liquibase changes
2. The `auto_update_database` runtime property set to `true`

It seems the default setup for the SDK has `auto_update_database` set to `false`, in which case, if you have unrun database changes, you're redirected to the UpdateFilter pathway. Setting it to `true` bypasses the update filter and directly calls `executeChangelog()`.

WolfSchlegel: This weekend I found some time to run another test:

- Initialise openmrs-core with 2.3.x
- set `auto_update_database=true` in `~/.OpenMRS/openmrs-runtime.properties`
- Build and restart openmrs-core from this branch in debug mode

`org.openmrs.util.DatabaseUpdater#executeChangelog()` was executed and all went well.

ibacher: Thanks @WolfSchlegel! That's what I've seen as well.

ibacher:
@mogoodrich Would you be able to test this change in your environment? I'd rather make sure we've cleared up any outstanding issues before I merge and back-port this but can do that if it's easier.

dkayiwa:
@ibacher do you have anything in particular that you would like to test with these changes?

ibacher:
@dkayiwa Not in particular; my thought was just that the PIH EMR has some slight variations from just the standard install of 2.3 and then upgrade to this version that I've tested locally and @WolfSchlegel indicated he tested above. Apart from any unknown factors, I'm quite happy to merge this change in.

PR #3631, opened 2020-11-06
TRUNK-5977 - getPatientOrPromotePerson leading to stack traces under …

mseaton:
TRUNK-5977 - getPatientOrPromotePerson leading to stack traces under some conditions

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -223,7 +223,7 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 		try {
 			patient = Context.getPatientService().getPatient(patientOrPersonId);
 		}
-		catch (ClassCastException ex) {
+		catch (Exception ex) {
 			// If the id refers to Person not Patient, it sometimes will cause class cast exception
 			// We will attempt to retrieve the Person and promote to Patient
 		}
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -9,6 +9,18 @@
  */
 package org.openmrs.api.impl;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
 import org.apache.commons.lang3.StringUtils;
 import org.openmrs.Allergen;
 import org.openmrs.Allergies;
@@ -46,6 +58,7 @@ import org.openmrs.api.UserService;
 import org.openmrs.api.VisitService;
 import org.openmrs.api.context.Context;
 import org.openmrs.api.db.PatientDAO;
+import org.openmrs.api.db.hibernate.HibernateUtil;
 import org.openmrs.parameter.EncounterSearchCriteria;
 import org.openmrs.parameter.EncounterSearchCriteriaBuilder;
 import org.openmrs.patient.IdentifierValidator;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -223,7 +223,7 @@ public Patient getPatientOrPromotePerson(Integer patientOrPersonId) {
 		try {
 			patient = Context.getPatientService().getPatient(patientOrPersonId);
 		}
-		catch (ClassCastException ex) {
```
dkayiwa, line 226: Just like the original exception, i would explicitly use `WrongClassException` instead of the generic `Exception`.

Secondly, even with the changes in this pull request, you would still get an `UnexpectedRollbackException` because of the above call `Context.getPatientService().getPatient(patientOrPersonId)` which marks the transaction for rollback. Therefore, you need to either change `Context.getPatientService().getPatient(patientOrPersonId)` to `getPatient(patientOrPersonId)` or add `noRollbackFor = WrongClassException.class` to the transaction definition for the `getPatient(Integer patientId)` method.

mseaton: @dkayiwa , thanks for that insight. I'd love to learn more about this from you.  Conceptually I'm confused about why any method annotated as a read-only transaction would ever result in an UnexpectedRollbackException - isn't the point of a read-only transaction that you want to prevent changes from getting written?  What is there to roll back?

From the perspective of our code, I'm not entirely sure why limiting our exception handling so narrowly will lead to better behavior.  Why not catch all Exceptions here and then attempt to load by person?  What type of error would we be catching that we feel we'd be better off throwing back to the calling code?

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -219,21 +220,17 @@ public Patient getPatient(Integer patientId) throws APIException {
 	@Override
 	@Transactional(readOnly = true)
 	public Patient getPatientOrPromotePerson(Integer patientOrPersonId) {
```
mseaton, line 222: @dkayiwa I was able to reproduce the issue and see the same errors that you reporting with my previous fix.  Here is my alternative solution, which fixes the issue for me.  Rather than try to load the patient, and then fall back to loading the person, this first loads the person, and checks to see if that person represents a valid patient or not.  This seems like the more correct behavior anyway, given our inheritance model.

To get this working in the refapp, I also needed to apply a related fix to idgen, which I raised a separate PR for, though you could also just disable the idgen module to test this standalone.

Thoughts on this approach?  This is a PR against the 2.3.x branch, which if accepted, we can up-port as needed.

mogoodrich: This generally seems correct, but I'd be leery of backporting it if it's just to fix the bug reported in TRUNK-5977... does this manifest itself it somewhere, or is there a reason this bug is more insidious than on might believe. It seems like a bit of an edge case (registering an existing provider or other relationship person as a patient)?

Another option would be to just change this to catch the other exception thrown in 2.3.x, and then make the larger change in 2.4.x?

mseaton: @mogoodrich - as previous PR comments (now marked as resolved) with @dkayiwa discovered, simply catching the other exception wasn't enough.  There were other exceptions that occurred.

The entire reason we are working on this and coding it against 2.3.x is because this is a bug that is manifesting in our soon-to-go-to-production distribution in Rwanda that is running 2.3.2, and we need the fix to go into 2.3.3.

mseaton: I think this is a more common use case than the ticket might have you believe - as in the Rwanda distribution it is common to create relationships and providers, including CHWs, on the patient dashboard, and then later if/when they become patients to promote these persons to patients via the process described here.

mogoodrich: Okay, fair enough... once we merge this we may actually want to upgrade the PIH EMR to 2.3.x to test...

mseaton: OK, will wait for approval or change requests until Monday and then look to commit.

mogoodrich (approved):

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
dkayiwa: @mseaton your solution is much better than the original code! 👍

PR #3632, opened 2020-11-06
TRUNK-5888: Fix business logic errors in saveCondition() method

ibacher:

This changes the `saveCondition()` method to conform with the rules defined [in this Talk thread](https://talk.openmrs.org/t/conditon-list-proposed-changes-to-savecondition-api/29451).

In order to maintain compatibility with EMR API, `saveCondition()` actually handles not just the logic of adding a new condition or modifying an existing condition, but also voiding or unvoiding an existing condition.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5888

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConditionDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConditionDAO.java
@@ -11,7 +11,7 @@ package org.openmrs.api.db.hibernate;
 
 import java.util.List;
 
-import org.hibernate.Query;
+import org.hibernate.query.Query;
 import org.hibernate.SessionFactory;
 import org.openmrs.Condition;
 import org.openmrs.Encounter;
@@ -42,18 +42,6 @@ public class HibernateConditionDAO implements ConditionDAO {
 	}
 	
 	/**
-	 * Saves the condition.
-	 *
-	 * @param condition the condition to save.
-	 * @return the saved condition.
-	 */
-	@Override
-	public Condition saveCondition(Condition condition) {
-		sessionFactory.getCurrentSession().saveOrUpdate(condition);
-		return condition;
-	}
-	
-	/**
 	 * Gets the condition with the specified id.
 	 *
 	 * @param conditionId the id to search for in the database.
@@ -61,7 +49,7 @@ public class HibernateConditionDAO implements ConditionDAO {
 	 */
 	@Override
 	public Condition getCondition(Integer conditionId) {
-		return (Condition) sessionFactory.getCurrentSession().get(Condition.class, conditionId);
+		return sessionFactory.getCurrentSession().get(Condition.class, conditionId);
 	}
 	
 	/**
@@ -72,8 +60,8 @@ public class HibernateConditionDAO implements ConditionDAO {
 	 */
 	@Override
 	public Condition getConditionByUuid(String uuid) {
-		return (Condition) sessionFactory.getCurrentSession().createQuery("from Condition c where c.uuid = :uuid")
-				.setString("uuid", uuid).uniqueResult();
+		return sessionFactory.getCurrentSession().createQuery("from Condition c where c.uuid = :uuid", Condition.class)
+				.setParameter("uuid", uuid).uniqueResult();
 	}
 	
 	/**
@@ -84,10 +72,22 @@ public class HibernateConditionDAO implements ConditionDAO {
 	 */
 	@Override
 	public List<Condition> getConditionHistory(Patient patient) {
-		Query query = sessionFactory.getCurrentSession().createQuery(
-				"from Condition con where con.patient.patientId = :patientId and con.voided = false " +
-						"order by con.onsetDate desc");
-		query.setInteger("patientId", patient.getId());
+		Query<Condition> query = sessionFactory.getCurrentSession().createQuery(
+				"from Condition con where con.patient.patientId = :patientId " +
+						"order by con.dateCreated desc", Condition.class);
+		query.setParameter("patientId", patient.getId());
+		return query.list();
+	}
+
+	/**
+	 * @see org.openmrs.api.ConditionService#getConditionsByEncounter(Encounter)
+	 */
+	@Override
+	public List<Condition> getConditionsByEncounter(Encounter encounter) throws APIException {
+		Query<Condition> query = sessionFactory.getCurrentSession().createQuery(
+			"from Condition c where c.encounter.encounterId = :encounterId and c.voided = false order "
+				+ "by c.dateCreated desc", Condition.class);
+		query.setParameter("encounterId", encounter.getId());
 		return query.list();
 	}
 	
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
```diff
@@ -117,37 +77,97 @@ public Condition getCondition(Integer conditionId) {
...
+	}
+
+	/**
+	 * Saves a condition
+	 *
+	 * @param condition - the condition to be saved
+	 * @return the saved condition
+	 */
+	@Override
+	public Condition saveCondition(Condition condition) {
```
JudeNiroshan, line 103:
```suggestion
	public Condition saveCondition(final Condition condition) throws APIException {
```

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
```diff
@@ -117,37 +77,97 @@ public Condition getCondition(Integer conditionId) {
...
 	@Transactional(readOnly = true)
 	public List<Condition> getAllConditions(Patient patient) {
 		return conditionDAO.getAllConditions(patient);
 	}
 
+	/**
+	 * @see ConditionService#getConditionsByEncounter(Encounter)
+	 */
+	@Override
+	public List<Condition> getConditionsByEncounter(Encounter encounter) throws APIException {
```
JudeNiroshan, line 92: This can also be `@Transactional(readOnly = true)`

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
```diff
@@ -117,37 +77,97 @@ public Condition getCondition(Integer conditionId) {
...
+		// If we got here, the updated condition and the existing condition are both live, so the updated condition is now
+		// replacing the existing condition
+		Condition newCondition = Condition.newInstance(condition);
+		newCondition.setPreviousVersion(existingCondition);
+
+		if (!existingCondition.getVoided()) {
+			existingCondition.setVoided(true);
+			existingCondition.setVoidedBy(Context.getAuthenticatedUser());
+			existingCondition.setVoidReason("Condition replaced");
+			conditionDAO.saveCondition(existingCondition);
```
JudeNiroshan, line 140: private void voidExistingCondition(Condition existingCondition) {
		existingCondition.setVoided(true);
		existingCondition.setVoidedBy(Context.getAuthenticatedUser());
		existingCondition.setVoidReason("Condition replaced");
		conditionDAO.saveCondition(existingCondition);
	}

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
ibacher: Could easily be, but I actually just moved the method rather than do anything to edit it. I'd prefer to stick that change as part of a different PR.

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
ibacher: I'm not sure I see the value of adding a method like that? Can you elaborate? To be clear, the reason that this logic is here is to void and replace an existing condition when updating the condition, hence the need to support voiding like that seems limited to this one method (other consumers of this service should continue to use `Context.getConditionService().voidCondition(myCondition, "My voiding reason");`.

JudeNiroshan (approved):

mogoodrich:
@ibacher  I missed this long ago... :)  this still good for review?

ibacher:
@mogoodrich Should be. Admittedly I haven't tried it since December...

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
mogoodrich: So this is just a move?  (Cool, fine with me)

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConditionDAO.java
```diff
@@ -84,10 +72,22 @@ public Condition getConditionByUuid(String uuid) {
...
 	public List<Condition> getConditionHistory(Patient patient) {
-		Query query = sessionFactory.getCurrentSession().createQuery(
-				"from Condition con where con.patient.patientId = :patientId and con.voided = false " +
-						"order by con.onsetDate desc");
-		query.setInteger("patientId", patient.getId());
+		Query<Condition> query = sessionFactory.getCurrentSession().createQuery(
+				"from Condition con where con.patient.patientId = :patientId " +
+						"order by con.dateCreated desc", Condition.class);
+		query.setParameter("patientId", patient.getId());
+		return query.list();
```
mogoodrich, line 79: So the idea here is that a condition history now *does* include voided conditions?

I guess the underlying question is, what is this condition history used for?  This has become more of an audit history of changes to the condition, right?

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
```diff
@@ -117,37 +77,97 @@ public Condition getCondition(Integer conditionId) {
...
+	 * @return the saved condition
+	 */
+	@Override
+	public Condition saveCondition(Condition condition) throws APIException {
+		Condition existingCondition = Context.getConditionService().getConditionByUuid(condition.getUuid());
+		// If there is no existing condition, then we are creating a condition
+		if (existingCondition == null) {
+			return conditionDAO.saveCondition(condition);
+		}
+
```
mogoodrich, line 109: Why do we need to support the void/unvoided functionality here?

mogoodrich (commented):
Thanks @ibacher ... added a few questions.

Also, I guess one we merge this will just be in master, correct?  We will have to think if we can backup it to 2.4.x or even 2.3.x.

Take care and thanks!
Mark

ibacher:
@mogoodrich Thanks for the review Mark! It actually jogged my memory, which is why I'm replying longer-form here rather than comment-by-comment.

So, one thing that struck me while doing this is that I ended up effectively treating "void" as a value for "active" / "inactive". Maybe that should be adjusted so that, e.g., we add a boolean indicating whether a condition is active or not and continue to treat "void" as it's more usual "deleted" meaning? (This is why `getConditionHistory()` returns voided results; so that we can populate the inactive tab of the UI).

The other thing this reminded me of is actually that the front-end changes kind of by design don't rely on the backend changes. That was so we don't have to add complications to get a new version of coreapps working correctly (i.e., you can deploy an updated version of coreapps against 2.1.x and the condition UI should continue to work). The result of that is that the front-end actually ends up just POSTing the updated data to the REST API exposed by EMRAPI, which ends up just calling `conditionService.saveCondition()`. So to maintain that flow, I had to adjust the `saveCondition()` method to handle the case where a condition had been voided. Again, this probably goes away if we move to some kind of "active" / "inactive" boolean.

mogoodrich:
So we can merge in the Core Apps change and the condition list will work without these changes?  That would be fantastic!

Having "voided" mean "inactive" seems wrong to me... that's what I think the "ACTIVE" and "INACTIVE" of the Condition Clinical Status field would be used for, and that the inactive UI would be driven by that.

For instance, say on January 21, 2010 you added an active condition for a patient, you'd end up with this row in the DB:
Condition A, Start Date: January 21, 2010, End Date: Null, ACTIVE

And then on January 21,2020 you determine the condition is no longer active, you'd want a new non-voided row:
Condition A, Start Date: January 11, 2020, End Date: January 21, 2020, INACTIVE

The only question would be would you just want to edit the first row in-line, or void the first row and create a second row to preserve an audit history.

Now, this audit history might also give some insight into the actual history, but that wouldn't be the primary use case.

Not sure if a "getConditionHistory(Patient)" DAO method actually makes sens as anything different from a getAllConditions(Patient) method. (And looking at the existing code the two methods appear to be identical?)  (It's also not used at all)

(I probably should review my previous write-ups on Talk to make sure I haven't contradicted something I wrote earlier, but please let me know if I did... :)

ibacher:
Oops! Sorry about that. 2 months away from this and we're out of coffee in my house so my brain isn't working properly. The void / unvoid in `saveCondition()` is to support deleting conditions via the UI. We use the status field for active / inactive. I'll have to dig into why I'm returning voided results from `getConditionHistory()`. I suspect it's to appease EMR API somehow.

mogoodrich:
Cool, thanks @ibacher !  Let me know... I'm not sure the getConditionHistory is even used?

ibacher:
@mogoodrich Yeah, it doesn't seem to be used anywhere. Let me try removing it altogether and seeing what breaks!

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
ibacher: Essentially this is here because CoreApps still talks to the platform through the EMR API module's REST interface which only calls the `saveCondition()` method regardless of if the condition is updated, voided, etc. So supporting this here was the best way to keep the changes pretty independent of one another.

PR #3649, opened 2020-11-30
TRUNK-5978 - Suppress errors when output stream is already closed

mseaton:
@dkayiwa  / @ibacher - thoughts on this?

### DIFF
--- a/web/src/main/java/org/openmrs/web/filter/GZIPResponseWrapper.java
+++ b/web/src/main/java/org/openmrs/web/filter/GZIPResponseWrapper.java
@@ -17,6 +17,7 @@ import javax.servlet.ServletOutputStream;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpServletResponseWrapper;
 
+import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -47,17 +48,12 @@ public class GZIPResponseWrapper extends HttpServletResponseWrapper {
 	}
 	
 	public void finishResponse() {
-		try {
-			if (writer != null) {
-				writer.close();
-			} else {
-				if (stream != null) {
-					stream.close();
-				}
+		if (writer != null) {
+			IOUtils.closeQuietly(writer);
+		} else {
+			if (stream != null) {
+				IOUtils.closeQuietly(stream);
 			}
-		}
-		catch (IOException e) {
-			log.error("Error during closing writer or stream", e);
 		}
 	}
 	


ibacher (approved):
Basically, this seems fine to me. It would be better to not have to use a deprecated method and even better if we could get out of the business of supporting gzip compression ourselves, but this looks to me like it works.

mseaton:
Thanks @ibacher - what's the deprecated method?  I ended up finding that the GZipFilter is the source of some problems on my instance that I have not yet debugged.  See more details in the ticket I raised here:
https://issues.openmrs.org/browse/TRUNK-5978
Do you have any insight / experience with this?

ibacher:
> what's the deprecated method?

[`IOUtils.closeQuietly()`](https://commons.apache.org/proper/commons-io/javadocs/api-release/org/apache/commons/io/IOUtils.html#closeQuietly-java.io.Closeable...-). Basically, they're recommending using try-with-resources instead. Doesn't really work for this scenario, though, so as long as it's there, we may as well use it.

>  Do you have any insight / experience with this?

Ooof... the issues you raised in that ticket look really troubling, especially when it's taking ~20 seconds to load a single smallish JS file.

My original thought had been that the Tomcat HTTP Connector [already supports GZip compression](https://tomcat.apache.org/tomcat-7.0-doc/config/http.html), and that's more likely to be performant than the servlet filter we're using. Spring Boot, for instance, relies on this to implement GZip compression when using Tomcat, see [here](https://github.com/spring-projects/spring-boot/blob/aef92b9295f62d008faa9ab79905a474bf3496f3/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/CompressionConnectorCustomizer.java#L34) (obviously, we're not going to be doing the Spring Boot configure Tomcat-in-code method, I was just pointing to it to indicate that a widely-used open-source framework also uses Tomcat's native compression).

Usually, though, when I'm building web apps, I try not to handle compression at all in the application and instead stick it behind a reverse proxy like Apache or Nginx and have compression enabled at that level (my reasoning being that efficient compression handling is critical for those web servers, so they're likely to be better at it than whatever I'd come up with).

dkayiwa:
Some history of a somehow related conversation: https://issues.openmrs.org/browse/TRUNK-4646

dkayiwa:
@mseaton these errors have always been around. It is just that previous versions did not log them as you can see here: https://github.com/openmrs/openmrs-core/blob/1.9.0/web/src/main/java/org/openmrs/web/filter/GZIPResponseWrapper.java#L63

So i have taken a different approach by first checking if the stream is not already closed, in this commit: https://github.com/openmrs/openmrs-core/commit/8d89376aa928ff8a04303dae9b992db78b27b28d

The setContentLength() method in that commit, probably fixes the slowing down problem.

PR #3708, opened 2021-03-03
RA-1865: Patch bugs discovered by NCSU team

Parth59:
@isears

Changes I made:
Added input-validation to the backend variables. Steps to reproduce can be found [here](https://docs.google.com/document/d/1lT8unCm2K1CzzmsX_1j44_vwm-5P2dKgk0i4JGRUTMo/edit?usp=sharing).

Link to ticket:
https://issues.openmrs.org/browse/RA-1865


api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -172,6 +173,24 @@ private OpenmrsUtil() {
...
+	 /**
+     * Verify if a string contains any HTML characters by comparing its
+     * HTML-escaped version with the original.
+     * @param String input  the input String
+     * @return boolean  True if the String contains HTML characters
+     */
+    public static boolean isHtml(String input) {
+        boolean isHtml = false;
+        if (input != null) {
+            if (!input.equals(HtmlUtils.htmlEscape(input))) {
```
HerbertYiga, line 186: entire pr looking good to me!

isears (commented):
@Parth59 it seems like this is not addressing EMPT-63 (xss in `/openmrs/admin/programs/conversion.form`) as described in the tracker. Is this a patch for another issue?

Parth59:
> @Parth59 it seems like this is not addressing EMPT-63 (xss in `/openmrs/admin/programs/conversion.form`) as described in the tracker. Is this a patch for another issue?

It's for 61 and 62.  Steps to reproduce can be found here. 
Apologies for my inconsistencies in the branch naming.

isears:
Ah, I see. I don't have access to that doc, but I can see 61 and 62 on my spreadsheet.

Parth59:
> Ah, I see. I don't have access to that doc, but I can see 61 and 62 on my spreadsheet.
Yupp. It's the same as 61 and 62 in spreadsheet ! 
Also, Have updated the link so that anyone with the link can view it for this month.

api/pom.xml
```diff
@@ -91,6 +91,10 @@
          <groupId>org.springframework</groupId>
          <artifactId>spring-context-support</artifactId>
       </dependency>
+      <dependency>
+				<groupId>org.springframework</groupId>
```
isears, line 95: This dependency is triggering a security alert. What do you think about using html-encoding functions from OWASP? These are already being used in the web section of openmrs-core: https://github.com/openmrs/openmrs-core/blob/722385734baabc0ee2e5a0d05fb1680d4ead8700/web/pom.xml#L127

isears (changes requested):
@Parth59 In general I think this is a good approach. My only concern is that there's a special case we're not thinking of where someone depends on being able to use HTML characters in their relationship types.

I'm going to open this issue up to the community on talk to see if anyone has any input before we merge this PR in: https://talk.openmrs.org/t/does-anyone-need-html-characters-in-relationship-types/32584

If we don't have any objections by Friday afternoon, I think we're safe to move forward.

Parth59:
Hi Team,
Have raised a new PR with better solution for the same. Here's the link for the same. 
NEW PR :- https://github.com/openmrs/openmrs-core/pull/3718
Openmrs talk : - https://talk.openmrs.org/t/does-anyone-need-html-characters-in-relationship-types/32584

isears:
Ok thanks @Parth59 

I'll close this one out and review #3718 shortly

PR #3717, opened 2021-03-11
TRUNK-5208:ModuleUtil compareVersions should sort a SNAPSHOT version as "less than"

jwnasambu:
I replaced the custom version ordering code with [SemVer ](https://github.com/vdurmont/semver4j)

## Issue I worked on
https://issues.openmrs.org/browse/TRUNK-5208

### DIFF
--- a/api/src/main/java/org/openmrs/module/ModuleUtil.java
+++ b/api/src/main/java/org/openmrs/module/ModuleUtil.java
@@ -59,18 +59,68 @@ public class ModuleUtil {
 
 	private ModuleUtil() {
 	}
-	
+
 	private static final Logger log = LoggerFactory.getLogger(ModuleUtil.class);
+
+    /**
+	 * This is the value used as the maximum version
+	 */
+	private static final String MAX_VERSION = String.valueOf(Integer.MAX_VALUE);
+
+	/**
+	 * This is the value used as the minimum version
+	 */
+	private static final String MIN_VERSION = "0";
+
+	/**
+	 * This pattern is adapted from SemVer.org's recommended pattern
+	 * 
+	 * However it has a few differences:
+	 *  1. Only a major version is required; minor, patch, and prerelease are all optional
+	 *  2. Build data is not supported
+	 *  3. The prerelease format is restricted and may not include a dash or period 
+	 */
+	private static final Pattern VERSION_PATTERN = Pattern.compile(
+		"^(?<major>0|[1-9]\\d*)" 
+			+ "(?:\\.(?<minor>0|[1-9]\\d*))?" 
+			+ "(?:\\.(?<patch>0|[1-9]\\d*))?" 
+			+ "(?:-(?<prerelease>[0-9a-zA-Z]*))?$"
+	);
+
+	/**
+	 * This is the same pattern as {@link #VERSION_PATTERN}, but it only provides a single group and permits the major,
+	 * minor, or patch version to be replaced with a <tt>*</tt> as a wildcard.
+	 * 
+	 * This is used to extract a reasonable facsimile of a SemVer from an OpenMRS version string
+	 */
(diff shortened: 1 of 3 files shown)


HerbertYiga:
@jwamalwa Thanks for working on this, i see some breaking point ,i picked this line from your travis logs ([ERROR] /home/travis/build/openmrs/openmrs-core/api/src/main/java/org/openmrs/module/ModuleUtil.java:[72,85] illegal character: '\')

api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -67,6 +67,11 @@ private ModuleUtil() {
 	 *
 	 * @param props Properties (OpenMRS runtime properties)
 	 */
+	// adapted from semver.org
+	private static final Pattern VERSION_PATTERN = Pattern.compile(
+		""^(?<major>0|[1-9]\\d*)\\.(?<minor>0|[1-9]\\d*)\\.(?<patch>0|[1-9]\\d*)(?:-(?<snapshot>SNAPSHOT))?$""
+	);
+
```
ibacher, line 74: This block should come before the Javadoc string.

api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -471,8 +488,26 @@ public static int compareVersion(String version, String value) {
...
+		final String[] versions = {
+			versionMatcher.group("major"),
+			versionMatcher.group("minor"),
+			versionMatcher.group("patch")
+		};
+		final String[] values = {
+			valueMatcher.group("major"),
+			valueMatcher.group("minor"),
+			valueMatcher.group("patch")
+		};
```
ibacher, line 510: Sorry, this block was intended to replace lines 426-450 above (the part in `compareVersion()` that is breaking the string down into it's component parts). The loop from lines 452-454 will also need to be adjusted.

api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -62,11 +62,17 @@ private ModuleUtil() {
 	
 	private static final Logger log = LoggerFactory.getLogger(ModuleUtil.class);
 	
+	// adapted from semver.org
+		private static final Pattern VERSION_PATTERN = Pattern.compile(
+			""^(?<major>0|[1-9]\\d*)\\.(?<minor>0|[1-9]\\d*)\\.(?<patch>0|[1-9]\\d*)(?:-(?<snapshot>SNAPSHOT))?$""
```
ibacher, line 67:
```suggestion
			"^(?<major>0|[1-9]\\d*)\\.(?<minor>0|[1-9]\\d*)\\.(?<patch>0|[1-9]\\d*)(?:-(?<snapshot>SNAPSHOT))?$"
```

You've got two sets of double quotes there, which doesn't work very well.

api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -418,31 +424,24 @@ public static int compareVersion(String version, String value) {
...
+				versionMatcher.group("minor"),
+				versionMatcher.group("patch")
+			};
+			final String[] values = {
+				valueMatcher.group("major"),
+				valueMatcher.group("minor"),
+				valueMatcher.group("patch")
+			};
 			
 			for (int x = 0; x < versions.size(); x++) {
```
ibacher, line 446: So now that we've changed `versions` and `values` from lists to arrays, we need adapt this look to use `versions.length` and, e.g., `versions[x]` to get teh value.

jwnasambu:
@ibacher  thanks for the review I have fixed the proposed changes.

jwnasambu:
@ibacher I kindly need your guidance on how to solve this error https://pastebin.com/QAKV87MC I tried fixing it on weekend but all in vain.

PR #3741, opened 2021-04-07
EMPT-111: Server logging for deleted patient

jmmacdo4:
Add a `log.info()` to patient deletion so removal can be tracked.

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
@@ -312,6 +312,7 @@ public class HibernatePatientDAO implements PatientDAO {
 	 */
         @Override
 	public void deletePatient(Patient patient) throws DAOException {
+		log.info("deleted patient " + patient.getId());
 		HibernatePersonDAO.deletePersonAndAttributes(sessionFactory, patient);
 	}
 	


jmmacdo4:
@isears another logging change here. I'm not entirely sure if I met all the requirements such as latest version and such, what do you think?

HerbertYiga:
@jmmacdo4  can you add  the ticket id to this pr as guided here https://wiki.openmrs.org/display/docs/Pull+Request+Tips

isears (changes requested):
@jmmacdo4 thanks for looking into this

I think this logging functionality already exists, it's just recorded as `voidPatient` rather than delete. When I do a simple delete patient test I see this in the logs (although admittedly, I don't know where/how exactly OpenMRS is logging the action, and maybe logging in the hibernate module is better):

```
INFO - LoggingAdvice.invoke(117) |2021-04-09T08:16:38,686| In method PatientService.voidPatient. Arguments: Patient=Patient#9, String=bla,
INFO - LoggingAdvice.invoke(157) |2021-04-09T08:16:38,816| Exiting method voidPatient
```

In addition, completely purging a patient from the advanced admin panel gets logged as:
```
INFO - LoggingAdvice.invoke(117) |2021-04-09T08:19:02,550| In method PatientService.purgePatient. Arguments: Patient=Patient#9,
INFO - LoggingAdvice.invoke(157) |2021-04-09T08:19:02,647| Exiting method purgePatient
```

If we're going to move forward with this PR, I think we need to at least identify where the logging is currently happening and remove it so that we're not logging redundantly.

isears:
So it looks like there are near-term plans to bring back the auditing module, so I'm going to close this PR for now.

In general let's postpone logging issues. If the audit module ever gets finalized and bundled as a refapp module, I don't want to get in a situation where every action in OpenMRS is needlessly double-logged.

Relevant talk thread here: https://talk.openmrs.org/t/consistent-logging-strategy/33062

PR #3754, opened 2021-04-27
TRUNK-5992: Order Service should always set autoExpireDate if duration is set.

mseaton:
#### Ticket:
https://issues.openmrs.org/browse/TRUNK-5992

#### What I have done:

See Talk thread here:  https://talk.openmrs.org/t/duration-and-auto-expire-date-on-orders/32006/2

The effect of this is to set the autoExpireDate if both duration and durationUnits are specified.  Unlike the previous behavior, this will happen for both Simple _and_ FreeText Dosing Instructions and will also happen if refills are specified on the order.

### DIFF
--- a/api/src/main/java/org/openmrs/BaseDosingInstructions.java
+++ b/api/src/main/java/org/openmrs/BaseDosingInstructions.java
@@ -1,0 +1,40 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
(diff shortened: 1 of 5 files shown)


mks-d (approved):

mogoodrich (commented):
From my limited knowledge, this looks good to me, but it really comes down to the question about what to do if refills are set... this does seem to change existing, intentional, behavior, even if it's not clear why... definitely interested in @bmamlin 's thoughts...

mseaton:
@mogoodrich - it's not just about the refills, it's about both refills and free-text.  What this PR is doing is correcting what (I think) @bmamlin is confirming as a misinterpretation of the original requirements, and ensuring that any explicit duration is always respected.

bmamlin (approved):
As discussed [here](https://talk.openmrs.org/t/duration-and-auto-expire-date-on-orders/32006/2), I would expect duration to include refills (e.g., a daily medicine dispensed as a month supply and 2 refills would have a duration of 3 months). If duration cannot be calculated and isn't manually set (e.g., PRN medications, chronic medications), then I'd expect duration  and auto-expiration date would be undefined (null).

Looks good to me. Thanks @mseaton!

PR #3758, opened 2021-04-28
TRUNK-6002: Encounters should appropriately save retrospective orders

mseaton:
This modified the logic within OrderService.saveOrder to consider the order to be retrospective if the date activated of that order is prior to the current date (time excluded).


mogoodrich (approved):
LGTM to me, thought I'm definitely not the authority on orders.

Although we wouldn't want to remove an existing API method, this could technically replace the " saveRetrospectiveOrder" method... ie instead of the user needed to decide whether or not an order was retro, it could be determined by the logic here?

bmamlin (approved):
Looks good to me as well.

PR #3759, opened 2021-04-28
TRUNK-5996:  Orders are not correctly voided when an Encounter is voided

mseaton:


mogoodrich (commented):
LGTM... do we want to do this for Obs as well?

PR #3768, opened 2021-05-05
TRUNK-6003: Make getLocationsByTag method null safe

cioan:


ibacher:
@cioan Thanks for this!

mogoodrich (commented):
LGTM... thanks @cioan ... and thanks @ibacher  for the quick merge!

PR #3829, opened 2021-07-29
RA-1945 Get ConceptNumeric with HibernateUtil

icrc-jfigueiredo:
Ticket:

https://issues.openmrs.org/browse/RA-1945


dkayiwa:
I thought the hibernate proxy thing was to be done from the coreapps module instead of openmrs core.

PR #3874, opened 2021-10-06
LUI-169 - Initial spike on fixing taglib loading issue

mseaton:
Spike on a patch that allows TLDs to be loaded from modules.  This is developed against 2.3.x, as this is the version I'd need to immediately target.

This extends JspServlet in order to apply an on-demand loading of TLDs at the time of the first page load, if they have not previously been loaded after the startup of OpenMRS.  Due to the nature of private methods on JspServlet, I needed to use reflection to get and update otherwise inaccessible fields.  The alternative would have been maintaining an entire custom copy of the JspServlet class (which is what my initial commit on this PR did, and which the subsequent commit addresses).

The reason I ended up here was that I found that it was not sufficient (at least in my test environment) to scan for TLDs, instantiate a new TldCache with these, and then simply update this on the ServletContext (via attribute).  I found that (or at least it appears thta) the JspServlet maintains an internal, private "Options" variable that stores the TldCache that it is initialized with at Tomcat startup, and this TldCache does not get updated when the TldCache in the ServletContext is replaced with a new instance.

@ibacher - I'm interested in your take on this.  Given that the OpenMRS web.xml file already referenced JspServlet directly, and now we are just directly referencing a subclass of it, I don't _think_ we should need to do any kind of "isTomcat" checking any more than we previously did.

### DIFF
--- a/web/src/main/java/org/openmrs/module/web/OpenmrsJspServlet.java
+++ b/web/src/main/java/org/openmrs/module/web/OpenmrsJspServlet.java
@@ -37,7 +37,7 @@ public class OpenmrsJspServlet extends JspServlet {
 	
 	private static final Logger log = LoggerFactory.getLogger(OpenmrsJspServlet.class);
 	
-	private transient boolean tldScanComplete = false;
+	public static final String OPENMRS_TLD_SCAN_NEEDED = "OPENMRS_TLD_SCAN_NEEDED";
 
 	@Override
 	public void init(ServletConfig config) throws ServletException {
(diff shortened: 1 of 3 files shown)


mseaton:
So, it seems that running this via the SDK doesn't work.  I need to investigate.

mseaton:
I don't think it has to do with the Tomcat version @dkayiwa .  I've tried running on the latest released SDK (which still uses Tomcat 7 embedded I believe), and the latest unreleased SNAPSHOT SDK (which @ibacher upgraded recently to use Tomcat 9 embedded from what I can see), and both exhibit the same behavior.

My guess is that the Embedded Tomcat has a different implementation of the org.apache.jasper.servlet.JspServet class which the OpenmrsJspServlet class I added in this PR is extending, and whatever it is doing differently is not being done properly with this new implementation.

Does anyone happen to know where I can find the source code for the tomcat embedded / tomcat embedded jasper codebases?

mseaton:
@ibacher / @dkayiwa are you guys happy for me to merge this into 2.3.x and upstream branches?  Any further review needed?

ibacher:
I'm happy with this. Thanks @mseaton!

PR #3875, opened 2021-10-06
LUI-169 - Initial spike on fixing taglib loading issue from Ian

mseaton:


mseaton:
@ibacher - this is just your original draft PR applied to the 2.3.x branch (with minimal changes) as this is what I was testing against.

PR #3949, opened 2022-01-04
TRUNK-6052: Enable log configuration via a log file

ibacher:

This PR changes how OpenMRS does logging configuration so that all configuration is done from a log4j2.xml file. The existing global properties (settings) that control logging should still work by default because they have been baked into our default log4j2.xml file.

The heart of this PR is the class `org.openmrs.logging.OpenmrsConfigurationFactory` which handles loading the configuration from a file and then applying the changes that OpenMRS expects. These changes are:

1. We must have an appender named `MEMORY_APPENDER` attached to the root logger. This appender uses the `org.openmrs.logging.MemoryAppender` class to store a configurable number of messages in-memory that are displayed in the various places we display logging messages in the UI.
2. Logging levels set via the `log.level` GP.

The `OpenmrsConfigurationFactory` will scan for any log configurations it knows how to handle in both the OpenMRS application directory and the configuration sub-directory of the application directory. These files should be named `log4j2` and should be in either the XML, YAML, or JSON dialects Log4J2 uses. Log4J1 style configurations aren't supported nor is configuration via a properties file, though those could be added in if necessary.

The other key facet of this PR is the `org.openmrs.logging.OpenmrsPropertyLookup` class, which exposes three values:

1. The OpenMRS application directory (`applicationDirectory`)
2. The `log.location` global property (`logLocation`)
3. the `log.layout` global property (`logLayout`)

Care should be taken when using these: because of the way OpenMRS starts up, the global property values will only be available when OpenMRS reloads the Spring context after loading all the modules, meaning these values will not always be available. However, they will work with Log4J2's expression syntax so that, e.g. `${openmrs:logLayout:-%m%n}` will default to `%m%n` and then change over to the configured value once the global properties are available.

There are some ancillary changes in this PR:

* `ServiceContext#getService()` now throws a `ServiceNotFound` exception instead of a generic `APIException`. This should be an invisible code change, but is useful if you want to detect specifically that a service wasn't found without digging into the message.
* All GP listeners in core are now defined outside of the context of defining the `openmrsEventListeners` bean. This ensures that they are only loaded once. Previously the core listeners could be loaded several different times (7 by default in the RefApp). This is because many modules use the following configuration to add a global property listener:

```xml
<bean id="AddressCacheResetSupport" parent="openmrsEventListeners">
    <property name="globalPropertyListeners">
        <list value-type="org.openmrs.api.GlobalPropertyListener" merge="true">
            <bean class="org.openmrs.module.addresshierarchy.AddressCacheResetSupport" />
        </list>
    </property>
</bean>	
```

This effectively creates a new bean that calls `setGlobalPropertyListeners()` with the list of both core event listeners and the added event listener. Since the core listeners are scoped to the `openmrsEventListener` bean, each child bean actually refers to a new instance of each core listener, so we ended up with several instances of core `openmrsEventListeners`. In the future, when adding beans like this, the `merge` property of the list should be set to `false`.

Outstanding things to be tested:

- [x] Does this actually load log4j2.xml from the application directory
- [x] Does this actually load log4j2.xml from the configuration sub-directory
- [x] Does everything work when using a customised log4j2.xml file

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6052

  No? -> execute above command


ibacher:
One outstanding question here: should this be backported to 2.5.x and 2.4.x? My gut feeling is "yes", but it's a reasonably extensive set of changes...

PR #3994, opened 2022-02-15
TRUNK-6061 - Support setting mail configuration in runtime or system …

mseaton:
…properties

This PR is issued against the 2.3.x branch, as this is where I plan to use this in the short term.  Once approved and merged, we can up-port to master and other later maintenance branches.

### DIFF
--- a/api/src/main/java/org/openmrs/util/ConfigUtil.java
+++ b/api/src/main/java/org/openmrs/util/ConfigUtil.java
@@ -1,0 +1,113 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/openmrs/util/ConfigUtil.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+	 * Returns the value of the given OpenMRS runtime property
+	 */
+	public static String getRuntimeProperty(String propertyName) {
+		return Context.getRuntimeProperties().getProperty(propertyName);
+	}
+
+	/**
+	 * Returns the value of the given OpenMRS runtime property
```
dkayiwa, line 50: You probably forgot to change the above description.

api/src/main/java/org/openmrs/util/ConfigUtil.java
```diff
@@ -0,0 +1,113 @@
...
+
+	/**
+	 * Returns the value of the given OpenMRS system property
+	 */
+	public static String getSystemProperty(String propertyName) {
+		return System.getProperty(propertyName);
+	}
+
+	/**
+	 * Returns the value of the given OpenMRS runtime property
```
dkayiwa, line 64: You probably forgot to change the above description.

mseaton: Good point, fixed

PR #3997, opened 2022-02-16
TRUNK-6061 - Support setting mail configuration in runtime or system …

mseaton:
This is a cherry-pick for the 2.4.x branch of the same features originally committed to 2.3.x here (073822f882d66301882ced310f34aae75707fa5e) but since the code is not 100% compatible, this has modifications to how this is setup in applicationContext-service.xml, and to making the Unit Test work with Junit 5 rather than Junit 4.  @dkayiwa I'm putting this in a new PR due to the above variations.  Would you mind having a look?


dkayiwa:
Looks good. Thanks @mseaton

PR #3998, opened 2022-02-16
TRUNK-6063 - Email configuration should support broader set of suppor…

mseaton:
…ted configuration parameters

This collects all configuration settings from global properties, runtime properties, and system properties that are prefixed with "mail.", and uses these to configure the mail session, in order to enable setting any current or future mail settings that are needed for java mail without hard-coding any specific or limited supported properties.

This is a PR against the 2.3.x branch, will need to be ported up if/when merged.

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Context.java
+++ b/api/src/main/java/org/openmrs/api/context/Context.java
@@ -9,21 +9,6 @@
  */
 package org.openmrs.api.context;
 
-import java.text.SimpleDateFormat;
-import java.util.Arrays;
-import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


ibacher (approved):
This seems fine. Shouldn't this target `master` though and then be backported?

ibacher:
Ah... I see TRUNK-6061... fair enough, I suppose.

mseaton:
@ibacher in theory yes, but our implementations are all on 2.3.x, so that's my main target and where I can easily test in the real world.  It's easier for me to up-port from there.

PR #3999, opened 2022-02-17
TRUNK-6063 - Fix to get broken unit test working in 2.4.x+

mseaton:
I don't know what changed between the 2.3.x branch and the 2.4.x branch to make this happen, but I get a test failure in 2.4.x without this commit.  It seems that the java mail implementation does not exist, just the api, and this leads to a failure in this unit test:

```
[ERROR] Failures: 
[ERROR]   UserServiceTest.setUserActivationKey_shouldCreateUserActivationKey:1518 Unexpected exception type thrown ==> expected: <org.openmrs.notification.MessageException> but was: <java.lang.NoClassDefFoundError>
```

@dkayiwa  / @ibacher - this should fix the 2.4.x build

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -277,6 +277,11 @@
 		   <groupId>org.apache.lucene</groupId>
 		   <artifactId>lucene-analyzers-phonetic</artifactId>
 	   </dependency>
+	   <dependency>
+		   <groupId>com.sun.mail</groupId>
+		   <artifactId>javax.mail</artifactId>
+		   <scope>test</scope>
+	   </dependency>
    </dependencies>
    <build>
      <resources>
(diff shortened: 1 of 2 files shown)


dkayiwa:
@mseaton do you know why that failure does not happen on bamboo?

mseaton:
@dkayiwa I think Bamboo is backed up - it doesn't look like it has run there yet.

mseaton:
To be honest, @dkayiwa I'm not entirely sure why we still build every branch in Bamboo when github actions can do the same at scale.

dkayiwa:
> To be honest, @dkayiwa I'm not entirely sure why we still build every branch in Bamboo when github actions can do the same at scale.

Could it be for releasing? Because we currently release from bamboo.

PR #4010, opened 2022-03-02
TRUNK-6069 - Password reset email should be localized to the user

mseaton:
This PR is against the 2.3.x branch, and ensures that the user's default locale is used to generate the localized email subject and message that is sent out in response to a password reset request.

### DIFF
--- a/api/src/main/java/org/openmrs/api/UserService.java
+++ b/api/src/main/java/org/openmrs/api/UserService.java
@@ -10,6 +10,7 @@
 package org.openmrs.api;
 
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 import org.openmrs.Person;
@@ -566,4 +567,10 @@ public interface UserService extends OpenmrsService {
 	 * @param newPassword the new password
 	 */
 	public void changePasswordUsingActivationKey(String activationKey, String newPassword);
+
+	/**
+	 * @param user the User whose Locale to retrieve
+	 * @return the default Locale of the given user, or the system locale if unspecified
+	 */
+	Locale getDefaultLocaleForUser(User user);
 }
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -566,4 +567,10 @@
 	 * @param newPassword the new password
 	 */
 	public void changePasswordUsingActivationKey(String activationKey, String newPassword);
+
+	/**
+	 * @param user the User whose Locale to retrieve
+	 * @return the default Locale of the given user, or the system locale if unspecified
```
dkayiwa, line 573: Would be great to add the @since annotation

mseaton: Yeah, I thought about that too @dkayiwa .  I figured could be added before merging, once the different versions are known.  Since it has to be changed (added to) when up-ported I would assume.

dkayiwa: By the way, do you intend to use this method outside of the user service?

mseaton: OK, I added a since 2.3.6.  Does it make sense to add the 2.4, 2.5, and 2.6 annotations at this point, or wait until porting up @dkayiwa ?

dkayiwa: > Does it make sense to add the 2.4, 2.5, and 2.6 annotations at this point, or wait until porting up @dkayiwa ?

Waiting until porting up makes sense!

dkayiwa: In that case, what do you think about adding a test for this public method?

mogoodrich (approved):
LGTM, thanks @mseaton !

api/src/main/java/org/openmrs/api/UserService.java
mseaton: @dkayiwa good point.  I've added a few.  Please have a look.

PR #4043, opened 2022-04-06
TRUNK-6071 - Add "MedicationDispense" domain object to core

mseaton:
Draft PR for adding new medication_dispense domain object and database table that will likely be combined with service/dao layer additions spec'd out in TRUNK-6072 in order to have unit test coverage and ensure that base set of related changes are added together.

@mogoodrich / @ibacher / @dkayiwa / @bmamlin / @mks-d  FYI.  Please review and confirm this matches your expectations.

This attempts to adhere mostly to what is laid out here:
https://wiki.openmrs.org/display/projects/Medication+Dispense+Data+Model

Some of the column names and property names might be slightly different, and we can discuss and debate.

I also have a few TODO comments for a few areas I want to confirm / sanity check before moving forward:

1. For Concept properties that relate to FHIR value sets which we may end up building decision logic around, do we want to consider enums or other programmatic means to identify these.

2. For 2 of the properties (provider and note) my read of the design is that the proper / FHIR way to do these would be to support multiple providers (with some sort of a role) and multiple notes (which contain additional data around who wrote the note and when it was written), but for which we are only planning to add a single "provider" foreign key column and single "note" text column.  I think this is worth sanity checking / pressure testing before we commit.


api/src/main/java/org/openmrs/parameter/MedicationDispenseCriteria.java
```diff
@@ -0,0 +1,89 @@
...
+ * @see MedicationDispenseCriteriaBuilder
+ */
+public class MedicationDispenseCriteria {
+	
+	private Patient patient;
+	private Encounter encounter;
+	private DrugOrder drugOrder;
+	private boolean includeVoided = false;
+	
+	// TODO: Consider what else we might want to add here to start out with, eg:
```
mseaton, line 36: @mogoodrich - question here.  We can leave all of these off for now, I just added them in as examples of ways we might want to search and filter the data.  Let's look at the use cases presented by the dispensing UI and consider what makes the most sense to support as search criteria upfront.

api/src/main/java/org/openmrs/api/impl/MedicationDispenseServiceImpl.java
```diff
@@ -0,0 +1,95 @@
...
+
+	@Override
+	@Transactional(readOnly = true)
+	public List<MedicationDispense> getMedicationDispenseByCriteria(MedicationDispenseCriteria criteria) {
+		return medicationDispenseDAO.getMedicationDispenseByCriteria(criteria);
+	}
+
+	@Override
+	public MedicationDispense saveMedicationDispense(MedicationDispense medicationDispense) {
+		// TODO: Do we want to void and re-create if represents an edit with changed data?
```
mseaton, line 63: @mogoodrich / @dkayiwa / @ibacher - question here.  Do we want to have some logic in the save method of either the service or DAO that enforces / magically voids and recreates on edit, or not?  We don't do this for most data, we do it most comprehensively for Obs.  I see some support for this in the Condition Service, but I also see some things that don't look quite right in that domain that I wouldn't want to replicate (i.e. I don't want to void and recreate if the MedicationDispense hasn't changed).  Possibly there is some dirty checking we can do in the DAO layer or with Hibernate Intercepting, but not sure if there are patterns established we can follow.  We should discuss.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateMedicationDispenseDAO.java
```diff
@@ -0,0 +1,85 @@
...
+		if (criteria.getDrugOrder() != null) {
+			predicates.add(criteriaBuilder.equal(md.get("drugOrder"), criteria.getDrugOrder()));
+		}
+		if (!criteria.isIncludeVoided()) {
+			predicates.add(criteriaBuilder.equal(md.get("voided"), false));
+		}
+		if (predicates.size() > 0) {
+			criteriaQuery.where(predicates.toArray(new Predicate[]{}));
+		}
+		// TODO: What should we order by here for consistency?
```
mseaton, line 71: I feel like we should have a consistent and sensible ordering of results - we do in most of the service methods.  Usually there is a straightforward date field that can be used for this, but I'm not sure what that is in this case.  We should look at the date fields in general and clarify what one can expect there - there are encounter dates (optional), order dates (optional), and a few optional dates on the medicationdispense for when certain things happen.  How do we plan to think about Medication Dispense items sorted by date?  @mogoodrich / @dkayiwa  / @ibacher / @bmamlin

mseaton (commented):
@bmamlin / @dkayiwa / @ibacher / @mogoodrich - a few comments on the working code that I committed in order to provide a means to ask a few questions in context.  Please have a look and let me know what you think.  Unit tests and such are coming next in this draft PR, but wanted to discuss this questions before getting too far along there.

api/src/main/java/org/openmrs/parameter/MedicationDispenseCriteria.java
mogoodrich: Yeah, makes sense... I believe I did intentionally derive the search specs from the basic use cases for Dispense that I knew we had so far, with the understanding that I inevitably would miss some.  In the UI design we have now, I believe all high-level searching happens at the Order/MedicationRequest level.  I don't think there's a case yet where we search for dispense events outside of orders.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateMedicationDispenseDAO.java
mogoodrich: Good question.  No immediate strong opinions.  Date prepared seems like my gut answer, but, as you pointed out, it's optional.  I'd be okay with making it or another date mandatory if we want to, it order to give consistent sorting... interested in what others think.

api/src/main/java/org/openmrs/api/impl/MedicationDispenseServiceImpl.java
mogoodrich: My gut would be no, but interested in others thoughts and adding @bmamlin .  Beyond Obs, don't we do so for Orders.  @bmamlin  @ibacher  does FHIR provide any guidance around this (not that I'm saying we 100% should follow what FHIR does, but interested in their thoughts toward it).

mogoodrich (commented):
Added some follow-ups to your questions, thanks @mseaton !

api/src/main/java/org/openmrs/api/db/hibernate/HibernateMedicationDispenseDAO.java
mseaton: I'm going to address this by ordering by "medicationDispenseId" as a starting point, which will at least ensure consistent results each time this method is called, and is a reasonable enough choice.  We can modify as we feel necessary moving forward.

api/src/main/java/org/openmrs/api/impl/MedicationDispenseServiceImpl.java
mseaton: OK, for now I'm going to leave this as-is, not not explicitly void and recreate, as it adds significant complexity. Happy to continue to discuss.

ibacher: Well, the reason for the void-and-recreate behaviour is to be sure we keep a historical record of previous values. FHIR has a different approach to it, which is, essentially, that every resource is versioned. So, e.g. if I request `/MedicationRequest/123`, I get the latest version of it, but if I request `/MedicationRequest/123/_history/2` I might get a previous version of it (at least assuming the FHIR server stores the previous version of the resource).

mseaton: Yeah, my gut is that we _should_ do this, but it would be great if there were some DAO utilities that the service layer could use to determine if a given object is dirty or not.  This object has like 35 properties, so doing a comparison (and maintaining it as properties change) feels a bit cumbersome and hard to maintain.  Thoughts @ibacher ?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateMedicationDispenseDAO.java
ibacher: Hmmm... that's a good question. There's a workflow consideration here, I think. To me, it would make the most sense to rely on the date handed over since it's a "dispense" event and that would be the field that stores "when the medication was given to the patient", which is probably the most clinically-relevant field. However, I can see a case where a pharmacist might want to setup all the data for a medication dispense before the drug was actually handed over... if that's a use-case we want to support than date prepared seems to make the most sense.

There's no reason we can't add constraints on top of the existing FHIR constraints, e.g., to make one or the other date fields required. FHIR resources are intended to be a bit general so they can be widely adopted with the idea that different business rules will actually restrict the contents of the resource.

api/src/main/java/org/openmrs/parameter/MedicationDispenseCriteria.java
ibacher: I like this having a criteria object pattern. Seems like it could save us quite a few override methods going forward.

api/src/main/java/org/openmrs/api/impl/MedicationDispenseServiceImpl.java
ibacher: Well there's what we do in Obs with a boolean `dirty` field and then every setter calls [`markAsDirty()`](https://github.com/openmrs/openmrs-core/blob/ed5eec88b4d3bfae584d9e6941ca6fb45cd54648/api/src/main/java/org/openmrs/Obs.java#L1198)...

It could be somewhat optimized by just making sure the `dirty` field defaults to `false`, e.g.,:

```java
@Transient
private transient boolean dirty = false;
```

Which would hopefully enable us to do away with the `ObsPostLoadEventListener`.

We could potentially use code generation to do the same thing (if you run Hibernate's newer "bytecode-enhanced dirty tracking", this is actually what it uses internally to check the state).

mseaton: Talking with @mogoodrich more, there are differences in intent behind MedicationDispense and Obs.  Obs typically, if you are editing them, means that you are correcting a mistake.  With MedicationDispense, there are several fields that seem like they'd get set progressively as you work - i.e. setting status, preparedBy, datePrepared, dateHandedOver, etc.  I don't know if there is value in retaining all of the permutations of these as voided data.

api/src/main/java/org/openmrs/api/impl/MedicationDispenseServiceImpl.java
```diff
@@ -0,0 +1,94 @@
...
+
+	@Override
+	public MedicationDispense voidMedicationDispense(
+		MedicationDispense medicationDispense, String reason) {
+		if (StringUtils.isBlank(reason)) {
+			throw new IllegalArgumentException("voidReason cannot be null or empty");
+		}
+		if (BooleanUtils.isNotTrue(medicationDispense.getVoided())) {
+			medicationDispense.setVoided(true);
+			medicationDispense.setVoidedBy(Context.getAuthenticatedUser());
```
dkayiwa, line 74: Isn't this automatically taken care of by `BaseVoidHandler`?

api/src/main/java/org/openmrs/api/impl/MedicationDispenseServiceImpl.java
```diff
@@ -0,0 +1,94 @@
...
+			medicationDispense.setDateVoided(new Date());
+			medicationDispense.setVoidReason(reason);
+		}
+		return saveMedicationDispense(medicationDispense);
+	}
+
+	@Override
+	public MedicationDispense unvoidMedicationDispense(MedicationDispense medicationDispense) {
+		medicationDispense.setVoided(false);
+		medicationDispense.setVoidedBy(null);
```
dkayiwa, line 84: Isn't this automatically taken care of by `BaseUnvoidHandler`?

api/src/test/java/org/openmrs/api/db/hibernate/HibernateMedicationDispenseDAOTest.java
```diff
@@ -0,0 +1,204 @@
...
+		l = medicationDispenseDAO.getMedicationDispenseByCriteria(b.build());
+		assertThat(l.size(), is(3));
+	}
+
+	/**
+	 * @see MedicationDispenseDAO#getMedicationDispenseByCriteria(MedicationDispenseCriteria)
+	 */
+	@Test
+	public void getMedicationDispenseByCriteria_shouldGetByDrugOrder() {
+		MedicationDispenseCriteriaBuilder b = new MedicationDispenseCriteriaBuilder();
```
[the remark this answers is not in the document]
mseaton: We could, but the value isn't clear to me, especially if code starts to span multiple lines due to length.

api/src/main/java/org/openmrs/api/impl/MedicationDispenseServiceImpl.java
mseaton: Yes, I believe it is, this is just being explicit about it, and the code here will ultimately never get called if the basevoidhandler properly does it's thing.  Do you want me to remove this code?

api/src/main/java/org/openmrs/api/impl/MedicationDispenseServiceImpl.java
mseaton: Same comment as above.

mseaton:
I've converted this from a draft to a normal PR for review.  There remain some potentially pending questions Talk that I've raised, but in the absence of opinions there to modify things, this represents the current state of changes that should be reviewed for merging.

api/src/main/java/org/openmrs/api/impl/MedicationDispenseServiceImpl.java
mogoodrich: What do other objects do?

If we do keep this, maybe add a comment here reflecting what Mike says above?

I guess the risk is that we could change the way voiding works, and update the BaseVoidHandler, but then forget to update the code here?

mogoodrich (commented):
Generally LGTM, just added some comments about the "Note" (which we will likely discuss in a separate Talk thread)

api/src/main/java/org/openmrs/MedicationDispenseNote.java
```diff
@@ -0,0 +1,149 @@
...
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+import java.util.Date;
+
+/**
+ * The MedicationDispenseNote class records extra information about the MedicationDispense
+ * that could not be conveyed in the other attributes
```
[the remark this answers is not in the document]
mseaton: Based on the discussion at the TAC today, I'm inclined to leave this in, though if we really don't think we'll need this, we can pull it out.

mseaton (commented):
Based on today's TAC, and reviewing the comments on this PR, I'm inclined to merge this in as is.

api/src/main/java/org/openmrs/MedicationDispenseNote.java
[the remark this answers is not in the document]
mogoodrich: Okay... not 100% crazy about it, as when I look at the specifications I'm thinking that we are going to need an OrdersNote more that a MedicationDispense note, but I could be wrong about that.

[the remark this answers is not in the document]
mseaton: OK @mogoodrich  et al, I have removed the MedicationDispenseNote from the PR and we can punt on whether to add support for this and how in a later phase.  I think this is likely good to merge.  @mogoodrich can you give a last once over?

api/src/main/java/org/openmrs/api/MedicationDispenseService.java
```diff
@@ -0,0 +1,81 @@
...
+	@Authorized({ PrivilegeConstants.EDIT_MEDICATION_DISPENSE })
+	MedicationDispense saveMedicationDispense(MedicationDispense medicationDispense);
+
+	/**
+	 * Voids a MedicationDispense
+	 * @param medicationDispense the MedicationDispense to be voided
+	 * @param reason the reason for voiding the MedicationDispense
+	 */
+	@Authorized({ PrivilegeConstants.EDIT_MEDICATION_DISPENSE })
+	MedicationDispense voidMedicationDispense(MedicationDispense medicationDispense, String reason);
```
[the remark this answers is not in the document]
mseaton: @ibacher I had used ConditionService as my template when I wrote this, and this is how it was done there.  If you think I should change it, I can.

api/src/main/java/org/openmrs/api/impl/MedicationDispenseServiceImpl.java
mseaton: OK, given the comments, I went ahead and removed the code.  See updated commit.

api/src/main/java/org/openmrs/api/impl/MedicationDispenseServiceImpl.java
mseaton: OK, given the comments, I went ahead and removed the code.  See updated commit.

PR #4045, opened 2022-04-07
TRUNK-5362: Issue with method in EncounterService causing extreme slowness

kdaud:
I have changed from vector class to using  ArrayList while implementing the search indexing criteria for patient identifier.

The search indexing criteria for 2.0.x version in the method `getEncountersByPatientIdentifier()` uses vector while searching for the patient identifier. Though the vector class maintains insertion order for the data but it is synchronized thus giving poor performance in searching, adding, delete and update of its elements.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5362

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
@@ -281,19 +281,20 @@ public class EncounterServiceImpl extends BaseOpenmrsService implements Encounte
 	/**
 	 * @see org.openmrs.api.EncounterService#getEncountersByPatientIdentifier(java.lang.String)
 	 */
+	@Override
 	@Transactional(readOnly = true)
 	public List<Encounter> getEncountersByPatientIdentifier(String identifier) throws APIException {
 		if (identifier == null) {
 			throw new IllegalArgumentException("The 'identifier' parameter is required and cannot be null");
 		}
 		
-		List<Encounter> encs = new Vector<Encounter>();
+		List<Encounter> encs = new ArrayList<>();
 		for (Patient p : Context.getPatientService().getPatients(identifier, null, null, false)) {
 			encs.addAll(Context.getEncounterService().getEncountersByPatientId(p.getPatientId()));
 		}
 		return Context.getEncounterService().filterEncountersByViewPermissions(encs, null);
 	}
-	
+
 	/**
 	 * @see org.openmrs.api.EncounterService#getEncounters(org.openmrs.Patient,
 	 *      org.openmrs.Location, java.util.Date, java.util.Date, java.util.Collection,


api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
```diff
@@ -281,19 +281,20 @@ public Encounter getEncounter(Integer encounterId) throws APIException {
...
 	 * @see org.openmrs.api.EncounterService#getEncountersByPatientIdentifier(java.lang.String)
 	 */
+	@Override
 	@Transactional(readOnly = true)
 	public List<Encounter> getEncountersByPatientIdentifier(String identifier) throws APIException {
 		if (identifier == null) {
 			throw new IllegalArgumentException("The 'identifier' parameter is required and cannot be null");
 		}
 		
-		List<Encounter> encs = new Vector<Encounter>();
```
dkayiwa, line 290: What do you base on to tell that this is the cause of the slowness?

dkayiwa:
Is there a reason why you are not using the master branch of openmrs-core?

api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
kdaud: The way how the search logic is implemented, the method searches actually  more criteria other than just patient identifier, which is not what the name of the method should be doing. And at the same time though the vector class maintains insertion order for the data but it is synchronized thus giving poor performance in searching, adding, delete and update of its elements.

kdaud:
The issue is reported after an upgrade to 2.0.x so that is why have redirected the fix to that branch other than master.

api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
dkayiwa: Did you in any way test this to confirm that it was slow before your changes but faster afterwards?

dkayiwa:
Is it faster on master?

sjmckee:
I don't believe just changing a Vector to an ArrayList fixes this issue.  In pre-2.0 versions of OpenMRS, this method ONLY searched by patient identifiers.  In post-2.0 versions, it searches across all aspects of the patient (patient identifiers, person names, person attributes), and this is what makes it extremely slow.  The method should only search based upon what the method name states...by identifier.

kdaud:
Master actually is using ArrayList but as mentioned above, there is less significance.

I notice  things get slow when during the search indexing criteria . It looks like 6 queries are executed every time HibernatePatientDAO.getPatients(String query, boolean includeVoided, Integer start, Integer length) is executed, which is what gets called by EncounterServiceImpl.getEncountersByPatientIdentifier(). 

Now since “getEncountersByIPatientIdentifier” does more than searching by identifier my work around is to deprecate this method and then refactor it to have the ability to search only by identifier.

api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
kdaud: I tried running against a deidentified copy of my database using a profiler but noticed a less significance.

dkayiwa: Then what you have done does not fix the reported problem.

PR #4064, opened 2022-04-19
Trunk 5954

enochmwanga:
Exception Fields 

TRUNK-5954 Exception classes should be immutable

Made all  exception class fields final

https://issues.openmrs.org/browse/TRUNK-5954


ibacher (changes requested):
@enochmwanga Thanks for this! However, I think some changes are needed. For example, how do I set the `format` field in the `InvalidIdentifierFormatException` class?

jwnasambu:
@enochmwanga kindly fix your ticket Id to something like this "TRUNK-5954:Exception classes should be immutable"

PR #4076, opened 2022-04-29
TRUNK-4988: Convenience method to allow change the configurations more  easily

kdaud:

Convenience method to allow change the configurations more  easily

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-4988

  No? -> execute above command

- [ x] My pull request is **based on the latest changes** of the master branch.

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Context.java
+++ b/api/src/main/java/org/openmrs/api/context/Context.java
@@ -70,6 +70,8 @@ import org.springframework.aop.Advisor;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.beans.factory.NoUniqueBeanDefinitionException;
+import org.springframework.context.support.AbstractApplicationContext;
+import org.springframework.context.support.ClassPathXmlApplicationContext;
 
 import javax.mail.Authenticator;
 import javax.mail.PasswordAuthentication;
@@ -576,6 +578,8 @@ public class Context {
 	 */
 	public static MessageService getMessageService() {
 		MessageService ms = getServiceContext().getMessageService();
+		AbstractApplicationContext context = new ClassPathXmlApplicationContext("applicationContext-service.xml");
+		ms = (MessageService) context.getBean("messageServiceBean", MessageService.class);
 		try {
 			// Message service dependencies
 			if (ms.getMessagePreparator() == null) {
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -576,6 +578,8 @@ public static ProgramWorkflowService getProgramWorkflowService() {
 	 */
 	public static MessageService getMessageService() {
 		MessageService ms = getServiceContext().getMessageService();
+		AbstractApplicationContext context = new ClassPathXmlApplicationContext("applicationContext-service.xml");
+		ms = (MessageService) context.getBean("messageServiceBean", MessageService.class);
```
ibacher, line 582: Can you walk me through what these two lines do and how they differ from the existing line 580?

api/src/main/resources/applicationContext-service.xml
```diff
@@ -597,6 +597,14 @@
 
 	<!--  ******************************  MESSAGE SERVICES  ****************************** -->
 
+    <bean id="messageServiceBean" class="org.openmrs.notification.MessageService">
```
ibacher, line 600: How is this bean different from the `messageService` defined below and why do we need it?

ibacher (changes requested):
@kdaud So the ticket mentions some TODO comments which don't seem to be addressed by this PR. It's unfortunate that it calls out specific lines because, of course line numbers change, especially in the almost 6 years since that ticket was created, but this PR doesn't seem connected to any TODO comments. Have you made an attempt to find what those TODO comment are / were and what the ticket actually asks to fix?

kdaud:
> Have you made an attempt to find what those TODO comment are / were and what the ticket actually asks to fix?

I went through the file and got to find out the line numbers changed and based on the ticket description, I get to understand the the ticket requires using Spring's method injection to set the dependencies for the message service.

Am not so familiar with spring method injection but my initial work was to add a bean in the applicationContext-service.xml so that the Message service dependencies are set via lookup-method provided by the spring injection  mechanism. 

Could it be that the TODO is no-longer existing as of now? or  I needed to approach it differently?

ibacher:
@kdaud They're still there. They've just moved to lines [668](https://github.com/kdaud/openmrs-core/blob/8db47aad4c28db1e7ff68c2f9733670fc03ec3ea/api/src/main/java/org/openmrs/api/context/Context.java#L668) and [678](https://github.com/kdaud/openmrs-core/blob/8db47aad4c28db1e7ff68c2f9733670fc03ec3ea/api/src/main/java/org/openmrs/api/context/Context.java#L678). The first comment is a bit unfortunate because I think it directed you to the wrong place.

With Spring, we already have a Spring context created and we should basically never create a new one.

PR #4094, opened 2022-05-25
TRUNK-6083: Build and deploy docker image for openmrs-core

rkorytkowski:
https://issues.openmrs.org/browse/TRUNK-6083


tendomart:
I will enjoy taking a look at this too

docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+#    v. 2.0. If a copy of the MPL was not distributed with this file, You can
+#    obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+#    the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+#    
+#    Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+#    graphic logo is a trademark of OpenMRS Inc.
+
+services:
+    db:
+        image: mariadb:10.3
```
tendomart, line 13: @rkorytkowski  @dkayiwa  just asking for learning's sake .
Are we not including **postgres** , How do we include **postgres** as an alternative for those wishing not wishing to run **mariadb** / mysql /

Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+WORKDIR /app
+
+# Copy source
+COPY . .
+
+# Build the app using maven cache
+RUN --mount=type=cache,target=/root/.m2 mvn clean package -DskipTests
+
+### Run stage
+FROM tomcat:7-jdk8-adoptopenjdk-hotspot
```
corneliouzbett, line 22: Just a thought,  is it possible to make tomcat and JDK configurable, or maybe we should avail different image flavors i.e `openmrs-core:tc9-jdk11`, `openmrs-core:tc10-jdk14`

rkorytkowski: When we decide to provide such flavours of images, we need to make sure it is compatible and test accordingly, which is not free... I don't think we can afford that and we should focus on one tomcat version. It should not be 7 since it reached the end of line. Ideally we pick 10, if openmrs-core runs on it correctly. I will also check Tomcat version in spring-boot since we may want to benefit from their dev tools at some point.

rkorytkowski: I found this thread: https://talk.openmrs.org/t/openmrs-update-to-latest-java-tomcat-and-mysql-versions/36505/3 which implies 10.x is not supported yet. I will upgrade to 9.x.

gitcliff: yeah 9.x is more suitable and stable for the current openmrs versions ,,,thanks raf

docker-compose.yml
bmamlin: I think it's safe to target mariadb over msyql. I'd actually like to migrate to postgresql as our first choice, but I think the majority of implementing teams have MySQL experience and that translates more easily (hopefully seamlessly) to mariadb.

docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+        command: "mysqld --character-set-server=utf8 --collation-server=utf8_general_ci"
+        environment:
+            MYSQL_DATABASE: openmrs
```
bmamlin, line 16: Does this need to be hardcoded? Or could it default to `openmrs` with ability to override via an environment variable? While we might be assuming "openmrs" for the database schema in some places, we shouldn't be. 🙂

Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+WORKDIR /app
+
+# All environment variables that are available to configure on this container are listed here
+# for clarity. These list the variables supported, and the default values if not overridden
+
+# These environment variables are appended to configure the Tomcat JAVA_OPTS
+ENV OMRS_JAVA_MEMORY_OPTS="-XX:NewSize=128m"
+ENV OMRS_JAVA_SERVER_OPTS="-Dfile.encoding=UTF-8 -server -Djava.security.egd=file:/dev/./urandom -Djava.awt.headless=true -Djava.awt.headlesslib=true"
+
+# These environment variables are used to create the openmrs-server.properties file, which controls how OpenMRS initializes
```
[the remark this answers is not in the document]
ibacher: So, unlike Compose files, Dockerfiles don't support the `${VAR:-defaultValue}` syntax. In effect, though, all of these declarations are the default value. If you were to run this container, with, e.g., `docker run -e OMRS_CONFIG_ADD_DEMO_DATA="true"` that will overwrite the value defined here. Similarly, these values can be overridden in the `environment` section of a compose file.

I think the naming convention is up for debate, but it's worth noting that `OMRS_` variables only apply to settings processed by `startup.sh`, leaving `OPENMRS_` variables for the things actually processed by OpenMRS Core. But maybe that's a consistency no one needs?

Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+RUN sed -i '/Connector port="8080"/a URIEncoding="UTF-8" relaxedPathChars="[]|" relaxedQueryChars="[]|{}^&#x5c;&#x60;&quot;&lt;&gt;"' /usr/local/tomcat/conf/server.xml
+
+# This is run at the end primarily so that all the previous steps in building the image remain cacheable
+RUN apt-get update && apt-get install -y \
+       zip dumb-init \
+    && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
+    && rm -rf /usr/local/tomcat/webapps/* 
+    
+# Copy the app
+COPY --from=build /app/webapp/target/openmrs.war /usr/local/tomcat/webapps/openmrs.war
```
ibacher, line 84: So the original idea of the `/openmrs/` directory structure was to have a convention for how OpenMRS distros are laid out that the Docker image would consume, so, e.g., by placing the WAR in the `/openmrs/openmrs_core` directory and then copying it over on startup, we effectively have a re-usable image that can deploy any OpenMRS instance without a deep knowledge of where things ended up in the final Docker file tree so, core goes in `/openmrs/openmrs_core`, modules go in `/openmrs/openmrs_modules`, etc. and the Docker image works out that the WAR goes in `/usr/local/tomcat/webapps` while the OMODs go in `/openmrs/data/modules`, etc. The original thought was that we could built tooling around this standardised distribution layout that could be deployed both to Docker and to bare metal.

Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+
+### Build Stage
+FROM maven:3.8-jdk-11 as build
+WORKDIR /app
+
+# Copy source
+COPY . .
+
+# Build the app using maven cache
+RUN --mount=type=cache,target=/root/.m2 mvn clean package -DskipTests
```
mseaton, line 19: I'm surprised we are building the war during Docker image creation.  I would have thought we would use pre-built artifacts.  Ideally we'd have 1 official version of each build (whether snapshot or release) in our Maven repository, and those versions are what would be bundled.  Then there are no surprises.

Dockerfile
mseaton: 9.x would be the right maximum version currently.

Dockerfile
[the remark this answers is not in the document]
mseaton: There OMRS_ prefix was just a convention we used to make it clear that these environment variables were the ones that, if you set them, would actually be used by the container/image, when we put together [the version I assume this is based off](https://github.com/PIH/docker-openmrs-server/blob/master/openmrs-server/Dockerfile) of here.

Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+# OMRS_CONFIG_CONNECTION_PORT
+# OMRS_CONFIG_CONNECTION_ARGS
+# OMRS_CONFIG_CONNECTION_EXTRA_ARGS
+#
+# If you really need complete control, you can just set
+# OMRS_CONFIG_CONNECTION_URL to whatever the URL should be
+
+# These environment variables are meant to enable developer settings
+# OMRS_DEV_DEBUG_PORT
+
```
mseaton, line 63: Not entirely sure why all of the above were commented out.  Are we assuming that no one should need to customize beyond the standard we are providing?

Dockerfile
mseaton: 100%.  It also makes it really easy if someone wants to mount in a volume and replace the "distribution" with a different one, or replace/modify certain components of it, they are in a standard location.  It also means that we can always clean up the working directories, webapps, etc. and replace them with fresh instances (remove work and temp dirs, exploded war dirs, etc) as needed, and add back in the artifacts.

docker-compose.yml
mseaton: My guess if you were to poll the community is that 95% of implementations are using MySQL 5.6 or 5.7.  MariaDB might be the right choice here, but I would be surprised to learn if it is widely used over MySQL.  If we want people to have the _most_ familiar environment to start out with, and the most well-tested / supported setup, it is definitely MySQL.

rkorytkowski: @mseaton the last time I checked there was no mysql build for ARM architecture (Apple's M1) so it's not ideal for devs (requires emulation). It most likely has performance implications, but if devs are not testing against a huge dataset (which is most of the time) then we should be fine. I haven't assessed differences between the two in terms of replication / sharding / partitioning. This is where I envisioned for us to provide configuration and it would be good to have it for the most used DB or the DB that we want most of implementations to use.

docker-compose.yml
rkorytkowski: I'll make it read an environment variable.

Dockerfile
rkorytkowski: Good point! As much as I agree with having a convention to store war, modules, configs, etc. under /openmrs/ dir so it's easier to build distro images and know where to put artifacts I would advise against mounting this dir and replacing components. The correct way is to build a new image and test.  If you need to adjust configuration between environments then do this via environment variables.  If you want to clean up, then drop the container and start fresh. It's okay to do it differently for development, but for production it's walking on thin ice.

docker-compose.yml
mseaton: @rkorytkowski - why do devs need a MySQL build?  Isn't running MySQL in Docker the way to go?  In any event, I'm not necessarily advocating here, just saying that if we are trying to reflect what is _actually_ used and best supported, then my guess is that it's still MySQL by far.

Dockerfile
mseaton: I'm definitely on board with moving all of our CI builds to use Docker.  That's something I've been working on personally in the PIH CI.  But that's not what we are talking about here, is it?  This container is about running OpenMRS in Tomcat, not building, testing, and deploying (otherwise you wouldn't be skipping tests below, and you'd be deploying to maven, etc).  Plus, there are a lot more things than OpenMRS core that we need to reliably build, test, and deploy to Maven, and we likely want some base images that strictly deal with building and deploying.  There are nuances there like how to setup Maven credentials, SSH and GPG keys, and run Selenium, Junit, and Jasmine tests in some cases.  What this seems to be doing is just building OpenMRS core because it is more convenient than downloading from Maven.

rkorytkowski: If you look at the Dockerfile it has 2 stages. First is the build stage, which is responsible for building and testing openmrs-core. It will be also used for pushing to maven repo and releasing. Tests are indeed skipped here for now, but I will enable them and make it possible to pass maven parameters by devs to skip them. Second is the runtime stage, which is building the tomcat image. It uses the results of the build stage to create the runtime environment (runtime image). The runtime image has none of the dev dependencies (not even maven). It is the one that will be pushed to dockerhub. The build stage and resulting image will be used be devs and CI only.

mseaton: That seems cool.  These stages are not something I'm familiar with.  Need to do some reading up.  Thanks!

docker-compose.yml
rkorytkowski: By MySQL build I mean a MySQL image built to be run on the given CPU, which can be used on devs' machines. When using Docker you still need to get the right image for your CPU architecture x86, AMR, etc. to run it or else it needs to be emulated.

PR #4128, opened 2022-08-18
TRUNK-6129: Explicitly allow Obs "groupMembers" to be mutable

mogoodrich:


mseaton (commented):
I don't think a separate PR is needed for backporting.  I'd do this by cherry-picking once the first one is merged.

PR #4143, opened 2022-09-08
TRUNK-6133 - Expand Listener to implement HttpSessionListener

mseaton:
See:  https://issues.openmrs.org/browse/TRUNK-6133

Interested in your thoughts on this PR @dkayiwa / @ibacher  / @mogoodrich

### DIFF
--- a/web/src/main/java/org/openmrs/web/Listener.java
+++ b/web/src/main/java/org/openmrs/web/Listener.java
@@ -9,35 +9,6 @@
  */
 package org.openmrs.web;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.io.OutputStreamWriter;
-import java.io.StringReader;
-import java.lang.reflect.Field;
-import java.nio.charset.StandardCharsets;
-import java.nio.file.Paths;
-import java.sql.Driver;
-import java.sql.DriverManager;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Enumeration;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-
-import javax.servlet.ServletContext;
-import javax.servlet.ServletContextEvent;
-import javax.servlet.ServletContextListener;
-import javax.servlet.ServletException;
-import javax.xml.parsers.DocumentBuilder;
-import javax.xml.parsers.DocumentBuilderFactory;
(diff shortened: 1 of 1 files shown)


dkayiwa:
@mseaton is it supposed to be an enhancement for this? https://issues.openmrs.org/browse/TRUNK-5416

mseaton:
> @mseaton is it supposed to be an enhancement for this? https://issues.openmrs.org/browse/TRUNK-5416

In a sense.  I'm aware of that feature in OpenMRS, the limitation of it is that it is purely an API-level feature and isn't actually tied into the session lifecycle in the servlet container (i.e. Tomcat).  It only really knows when the API has been actively logged into or logged out of, and what user did it.  There is no ability to know what session this is tied to, there is no knowledge of any data in the session or attached to the request, like the user's IP address, and there is no way to detect if the user is logged out passively (due to inactivity and session expiry by the servlet container - which for Tomcat defaults to 30 minutes), etc.  

So I think it would would be difficult/impossible to build certain features - for example, one that could accurately say how many users are currently logged in (because you wouldn't know about users who were passively logged out by the servlet container).

So it's complementary to the existing feature.  It doesn't really duplicate it.  I could see adding an HttpSessionListener implementation to core (or even just adding a method call directly in the Listener.sessionDestroyed() method that calls Context.logout() if the session has been destroyed and Context.isAuthenticated() is true.  This would then cascade to notify the UserSessionListener of the logout event.  But I left that as outside the scope of this feature.  Something to consider though.

dkayiwa:
> I could see adding an HttpSessionListener implementation to core (or even just adding a method call directly in the Listener.sessionDestroyed() method that calls Context.logout() if the session has been destroyed and Context.isAuthenticated() is true. This would then cascade to notify the UserSessionListener of the logout event. But I left that as outside the scope of this feature. Something to consider though.

Yes i would consider adding that because it makes lots of sense! 👍

ibacher:
Yes, this is an extremely useful feature to have, if a bit niche. As I recall, the chief difficulty replicating the 1.x "View Logged In Users" functionality was the inability to track expired sessions accurately.

mseaton:
Cool, good to hear this is useful.  Wondering on thoughts about backporting to 2.5.x.  Thoughts on that?  @dkayiwa / @ibacher ?

dkayiwa:
> Wondering on thoughts about backporting to 2.5.x

No objection from me. 😊

mogoodrich:
Belated LGTM... I guess I'm always a little fearful of backporting functionality that touches "everything", no matter how benign... but we will be using this right away and hopefully in the near term testing on a large scale implementation, so we should be able to debug as need be.

PR #4154, opened 2022-09-18
TRUNK-6133: Only run listeners after OpenMRS initialized

ibacher:

Just tries to delay running any of the registered `HttpSessionListener`s until the context has been initialized. There's no `ApplicationContext` until setup is done, so there isn't much point. One thing this leaves hanging is that if Liquibase updates are needed, we'll still try to load `HttpSessionListener` before a context is available, but this is a rare occurrence and there's ultimately no actual harm done.

@mseaton Interested in your thoughts on this

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6133

  No? -> execute above command

### DIFF
--- a/web/src/main/java/org/openmrs/web/Listener.java
+++ b/web/src/main/java/org/openmrs/web/Listener.java
@@ -91,7 +91,7 @@ public final class Listener extends ContextLoader implements ServletContextListe
 	
 	private static boolean setupNeeded = false;
 	
-	private static boolean initialized = false;
+	private static boolean openmrsStarted = false;
 	
 	/**
 	 * Boolean flag set on webapp startup marking whether there is a runtime properties file or not.
@@ -172,7 +172,7 @@ public final class Listener extends ContextLoader implements ServletContextListe
 	private List<HttpSessionListener> getHttpSessionListeners() {
 		List<HttpSessionListener> httpSessionListeners = Collections.emptyList();
 		
-		if (initialized) {
+		if (openmrsStarted) {
 			try {
 				httpSessionListeners = Context.getRegisteredComponents(HttpSessionListener.class);
 			}
@@ -257,8 +257,6 @@ public final class Listener extends ContextLoader implements ServletContextListe
 			setErrorAtStartup(e);
 			log.error(MarkerFactory.getMarker("FATAL"), "Failed to obtain JDBC connection", e);
 		}
-		
-		initialized = true;
 	}
 	
 	private void loadCsrfGuardProperties(ServletContext servletContext) throws FileNotFoundException, IOException {	
@@ -323,6 +321,7 @@ public final class Listener extends ContextLoader implements ServletContextListe
 	 * @throws ServletException
 	 */
 	public static void startOpenmrs(ServletContext servletContext) throws ServletException {
+		openmrsStarted = false;
 		// start openmrs
 		try {
 			// load bundled modules that are packaged into the webapp
@@ -360,6 +359,7 @@ public final class Listener extends ContextLoader implements ServletContextListe
 		finally {
 			Context.closeSession();
 		}
+		openmrsStarted = true;
 	}
 	
 	/**
@@ -629,6 +629,7 @@ public final class Listener extends ContextLoader implements ServletContextListe
 			}
 			// remove the user context that we set earlier
 			Context.closeSession();
+			openmrsStarted = false;
 		}
 		try {
 			for (Enumeration<Driver> e = DriverManager.getDrivers(); e.hasMoreElements();) {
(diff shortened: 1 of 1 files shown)


dkayiwa:
@ibacher under what conditions would `sessionCreated` or `sessionDestroyed` be called before `contextInitialized`? 😊

ibacher:
@dkayiwa That's actually a good point... It's actually [this locale part](https://github.com/openmrs/openmrs-core/blob/6e5a44642663152d2460024f56615691791d5264/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java#L204) in the `InitializationFilter`

dkayiwa:
@ibacher i thought `contextInitialized` would be called before any filter. Do you happen to have any pointers to how i can reproduce the situation where the `InitializationFilter` is run before `contextInitialized`? 😊

ibacher:
@dkayiwa Well, for me it was happening every time I started up the 3.x RefApp Docker images...

ibacher:
I suppose it may be that the [5s rate here](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/698d3c4621a8163ee56c0aa9da5377327bcbccbd/docker-compose.yml#L42) is too fast...

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -167,13 +170,17 @@ public void sessionDestroyed(HttpSessionEvent se) {
...
 	private List<HttpSessionListener> getHttpSessionListeners() {
-		List<HttpSessionListener> httpSessionListeners = new ArrayList<>();
-		try {
-			httpSessionListeners = Context.getRegisteredComponents(HttpSessionListener.class);
-		}
-		catch (Exception e) {
-			log.warn("An error occurred trying to retrieve HttpSessionListener beans from the context", e);
+		List<HttpSessionListener> httpSessionListeners = Collections.emptyList();
+		
+		if (initialized) {
```
mseaton, line 175: Putting a check around this seems fine with me.  I don't know if this will miss any particular session events, but if we are relying on listeners registered in the application context, we don't really have any choice but to wait for the application context to be fully initialized

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -251,6 +258,7 @@ public void contextInitialized(ServletContextEvent event) {
 			log.error(MarkerFactory.getMarker("FATAL"), "Failed to obtain JDBC connection", e);
 		}
 		
+		initialized = true;
```
mseaton, line 261: I wonder if this is actually sufficient.  What about instead adding a boolean flag for "openmrsStarted", and setting this to true at the end of Listener.startOpenmrs(ServletContext).  And actually exposing a static public method "isOpenmrsStarted()" that returns the value of this.   And use this value to determine whether the application context is refreshed sufficiently to notify the listeners.  Because it seems like it isn't until this point that the application context is fully refreshed with all of the modules?

mseaton: NOTE:  Although I still think this is a reasonable feature to have added in, I recently saw that in the legacyui module, this module users a Component that implements ServletContextAware to achieve much the same behavior - basically registering a custom Session Listener through that mechanism.  So if it is better to roll back this change and just recommend that modules take this approach, that's probably fine too.  See:  https://github.com/openmrs/openmrs-module-legacyui/blob/master/omod/src/main/java/org/openmrs/web/WebComponentRegistrar.java#L50

@dkayiwa do you have opinions on how well this approach has worked?

dkayiwa: I have not heard of any complaints. It has the advantage of not requiring upgrading the openmrs platform version. And since it is already based on spring classes, it does not require checking whether the spring application context is initialised. But that being said, i have no objection to bringing this in.

mseaton: @dkayiwa  / @ibacher  - I've committed an update to this PR that applies the idea I laid out above.  Can you have a look and let me know what you think?  @ibacher did you test the previous fix anywhere in particular?

ibacher: I think you can just build and start the Docker image to try it out. It consistently logs a warning when starting up so if the warning is gone, it should be good.

mseaton: @ibacher is just the Docker file included with core sufficient, or do I need to use the refapp image as described on the ticket?

ibacher: It should be fine as long as we backport this to the 2.5.x branch (since the original change is also in 2.5.x).

PR #4176, opened 2022-10-24
TRUNK-6145: Allow validation to be disabled only within the current t…

mseaton:
…hread

### DIFF
--- a/api/src/main/java/org/openmrs/validator/ValidateUtil.java
+++ b/api/src/main/java/org/openmrs/validator/ValidateUtil.java
@@ -9,9 +9,6 @@
  */
 package org.openmrs.validator;
 
-import java.util.LinkedHashSet;
-import java.util.Set;
-
 import org.apache.commons.lang3.StringUtils;
 import org.openmrs.OpenmrsObject;
 import org.openmrs.api.ValidationException;
@@ -22,6 +19,9 @@ import org.springframework.validation.BindException;
 import org.springframework.validation.Errors;
 import org.springframework.validation.FieldError;
 import org.springframework.validation.ObjectError;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 /**
  * This class should be used in the *Services to validate objects before saving them. <br>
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/validator/ValidateUtil.java
```diff
@@ -146,4 +149,31 @@ public static void setDisableValidation(Boolean disableValidation) {
...
+	 * @return true if validation has been disabled for the current thread, false otherwise
+	 * @since 2.5.8
+	 */
+	public static boolean isValidationDisabledForThread() {
+		return disableValidationForThread.get() == Boolean.TRUE;
+	}
+
+	/**
+	 * Sets a ThreadLocal variable to indicate that validation should be disabled for the current thread
+	 * NOTE: This should always be used in conjunction with the resumeValidationForThread method to
```
dkayiwa, line 162: Just a minor comment, i would not include the implementation details `Sets a ThreadLocal variable` in the method description.

api/src/main/java/org/openmrs/validator/ValidateUtil.java
```diff
@@ -146,4 +149,31 @@ public static void setDisableValidation(Boolean disableValidation) {
...
+	 * NOTE: This should always be used in conjunction with the resumeValidationForThread method to 
+	 * ensure the ThreadLocal instance is cleaned up at the end of the operation
+	 * @since 2.5.8
+	 */
+	public static void disableValidationForThread() {
+		disableValidationForThread.set(Boolean.TRUE);
+	}
+
+	/**
+	 * Removes a ThreadLocal variable to indicate that validation should be re-enabled for the current thread
```
dkayiwa, line 171: Same as above.

mseaton: Changed

api/src/main/java/org/openmrs/validator/ValidateUtil.java
mseaton: Changed

PR #4212, opened 2022-12-08
TRUNK-6152 - Support multiple log4j2 configuration files

mseaton:

### DIFF
--- a/api/src/main/java/org/openmrs/logging/OpenmrsConfigurationFactory.java
+++ b/api/src/main/java/org/openmrs/logging/OpenmrsConfigurationFactory.java
@@ -9,11 +9,7 @@
  */
 package org.openmrs.logging;
 
-import javax.validation.constraints.NotNull;
-import java.io.File;
-import java.net.URI;
-import java.util.Locale;
-
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.Level;
@@ -24,6 +20,7 @@ import org.apache.logging.log4j.core.config.ConfigurationFactory;
 import org.apache.logging.log4j.core.config.ConfigurationSource;
 import org.apache.logging.log4j.core.config.LoggerConfig;
 import org.apache.logging.log4j.core.config.Order;
+import org.apache.logging.log4j.core.config.composite.CompositeConfiguration;
 import org.apache.logging.log4j.core.config.json.JsonConfiguration;
 import org.apache.logging.log4j.core.config.plugins.Plugin;
 import org.apache.logging.log4j.core.config.xml.XmlConfiguration;
@@ -34,6 +31,14 @@ import org.openmrs.api.context.Context;
 import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.OpenmrsUtil;
 import org.slf4j.LoggerFactory;
+
+import javax.validation.constraints.NotNull;
+import java.io.File;
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Locale;
 
 /**
  * {@link ConfigurationFactory} to handle OpenMRS's logging configuration.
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/logging/OpenmrsConfigurationFactory.java
```diff
@@ -103,6 +114,33 @@ protected String[] getSupportedTypes() {
 		return SUFFIXES;
 	}
 	
+	public List<File> getConfigurationFiles() {
```
mseaton, line 117: The idea of this method is to find anything in the .OpenMRS directory or .OpenMRS/configuration directory (non-recursively) that is named log4j2*.{extension}.

It returns these sorted by name (without the directory), so inheritance will be based solely on filename sorting across these 2 directories.

Not sure if this is the best way to determine inheritance / which override which, but this is the most straightforward.

api/src/main/java/org/openmrs/logging/OpenmrsConfigurationFactory.java
```diff
@@ -103,6 +114,33 @@ protected String[] getSupportedTypes() {
...
+		List<File> configurationFiles = new ArrayList<>();
+
+		// Get all extensions except for the wildcard extension
+		List<String> extensionList = new ArrayList<>();
+		for (String suffix : SUFFIXES) {
+			if (!suffix.equals("*")) {
+				extensionList.add(suffix.substring(1)); // Add the suffix without the leading "."
+			}
+		}
+		String[] extensions = extensionList.toArray(new String[] {});
```
ibacher, line 127: Couldn't this be done statically, e.g.,

```
public static final String[] EXTENSIONS = Arrays.stream(SUFFIXES).filter(s -> !s.equals("*")).map(s -> s.substring(1)).toArray(String[]::new);
```

The suffixes shouldn't really change at runtime....

api/src/main/java/org/openmrs/logging/OpenmrsConfigurationFactory.java
```diff
@@ -53,33 +58,39 @@ public class OpenmrsConfigurationFactory extends ConfigurationFactory {
...
+				else {
+					List<AbstractConfiguration> abstractConfigurations = new ArrayList<>();
+					for (File configFile : configurationFiles) {
+						Configuration configuration = super.getConfiguration(loggerContext, name, configFile.toURI());
+						if (configuration instanceof AbstractConfiguration) {
+							System.out.println("Adding log4j2 configuration file: " + configFile.getPath());
+							abstractConfigurations.add((AbstractConfiguration) configuration);
+						}
+						else {
+							System.out.println("Unable to add log4j2 configuration file: " + configFile.getPath());
```
ibacher, line 85: Since this is an error, it might make sense to write it to stderr, e.g.,

```suggestion
							System.err.println("Unable to add log4j2 configuration file: " + configFile.getPath());
```

ibacher (approved):
A couple of small things, but this LGTM!

api/src/main/java/org/openmrs/logging/OpenmrsConfigurationFactory.java
mseaton: Made this change, let me know how this looks @ibacher

api/src/main/java/org/openmrs/logging/OpenmrsConfigurationFactory.java
mseaton: Done.

ibacher:
Looks goo to me now. I'm also unsure what the right pattern for overriding should be... I think just going by order of name is probably fine, though if we had a lot of these, e.g.,

log4j2_1.xml
log4j2_2.xml
...
log4j2_10.xml

The ordering might be a little unexpected... That said, I imagine the use-case for this is to have maybe two log configurations (global + implementation specific) not ten.

api/src/main/java/org/openmrs/logging/OpenmrsConfigurationFactory.java
```diff
@@ -53,33 +59,43 @@ public class OpenmrsConfigurationFactory extends ConfigurationFactory {
...
-						continue;
-					}
-					
-					File configFile = new File(applicationDirectory, getDefaultPrefix() + suffix);
-					if (configFile.exists() && configFile.canRead()) {
-						return super.getConfiguration(loggerContext, name, configFile.toURI());
+			List<File> configurationFiles = getConfigurationFiles();
+			if (!configurationFiles.isEmpty()) {
+				if (configurationFiles.size() == 1) {
+					System.out.println("Adding log4j2 configuration file: " + configurationFiles.get(0).getPath());
```
mogoodrich, line 78: Was it intentional to use System.out here?

mogoodrich (commented):
LGTM... It would be good to add some tests around this, but I guess that would be tough in this case?

PR #4214, opened 2022-12-09
TRUNK-6153 - Email messages are not sent with correct from address

mseaton:
This PR was authored, and will be initially merged against, the 2.5.x branch.  We can then up-port it to 2.6.x+.


mogoodrich (approved):
LGTM, thanks @mseaton !

PR #4264, opened 2023-03-06
Temporary fix for 2.5.x CI job

mseaton:
This is the kind of thing I'm looking at doing in the absence of properly fixing the 2.5.x build in Bamboo, which is currently down.  See the #infra Channel in slack for commentary / description of the issues around this.  Hopefully we can get a proper fix in place and this is unnecessary, but if we can't get the build issues resolved, then this is something I may look to try.  Feedback and input please.

### DIFF
--- a/bamboo-specs/bamboo.yml
+++ b/bamboo-specs/bamboo.yml
@@ -11,232 +11,46 @@ stages:
           final: false
           jobs:
               - Build
-    - Test:
-          manual: false
-          final: false
-          jobs:
-              - Unit Test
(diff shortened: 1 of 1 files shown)


dkayiwa:
@mseaton the current 2.5.x bamboo CI build failure looks exactly like the failure which happened to the master branch, last week. Doesn't the 2.5.x branch just need the same changes that were done for master?

dkayiwa:
@mseaton do you still need this?

PR #4265, opened 2023-03-07
TRUNK-6165 - ConceptService getConceptsByMapping should only return d…

mseaton:
…istinct concepts

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
@@ -2088,6 +2088,9 @@ public class HibernateConceptDAO implements ConceptDAO {
 			criteria.addOrder(Order.asc("concept.retired"));
 		}
 
+		// we only want distinct concepts
+		criteria.setResultTransformer(DistinctRootEntityResultTransformer.INSTANCE);
+
 		return criteria;
 	}
 }
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
```diff
@@ -2085,9 +2088,6 @@ private Criteria createSearchConceptMapCriteria(String code, String sourceName,
 			criteria.addOrder(Order.asc("concept.retired"));
 		}
 
-		// we only want distinct concepts
-		criteria.setResultTransformer(DistinctRootEntityResultTransformer.INSTANCE);
```
dkayiwa, line 2089: Did you intentionally remove this?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
```diff
@@ -1027,7 +1028,8 @@ public void remove() {
 	public List<Concept> getConceptsByMapping(String code, String sourceName, boolean includeRetired) {
 		Criteria criteria = createSearchConceptMapCriteria(code, sourceName, includeRetired);
 		criteria.setProjection(Projections.property("concept"));
-		return (List<Concept>) criteria.list();
```
dkayiwa, line 1030: Did you intentionally leave out a test for this?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
mseaton: Yes, but interested in your thoughts.  I think this was a holdover from an earlier time.  The root entity here is a ConceptMap, not a Concept.  I think the assumption was that this was leading the method to return distinct Concepts, but this was not actually the case.  This just resulted in distinct ConceptMaps...

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
mseaton: Well, I didn't intentionally add a test.  There was no existing test, and the method is deprecated, so I didn't really worry about it.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
dkayiwa: I would be cautious about removing this and do so only when am very sure that it does not have unintended side effects. At the very least, i would remove it from a commit different for that of fixing duplicate concept ids.

mseaton: OK, I added it back in @dkayiwa .  I don't think it has any impact, so it's not really a problem (though I think it is likely misleading, assuming I'm understanding it correctly).  Unit test I added passes either way...

dkayiwa: Am fine removing it from a separate commit.

PR #4285, opened 2023-03-24
TRUNK-7521: Complex obs handlers to return obs as is when underlying file is null.

jona42-ui:
Added a null check to ensure that the file variable is not null before attempting to read its contents:
This will return the original obs object if the file is null, rather than attempting to construct a ComplexData object from a null file and throwing a NullPointerException.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5721

### DIFF
--- a/api/mytxtfile.txt
+++ b/api/mytxtfile.txt
@@ -1,0 +1,1 @@
+test content
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -296,6 +296,12 @@
 			<groupId>org.testcontainers</groupId>
 			<artifactId>mysql</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-web</artifactId>
+			<version>5.3.13</version>
+		</dependency>
+
 	</dependencies>
 	<build>
 		<resources>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -75,6 +75,12 @@
 			<artifactId>reflectutils</artifactId>
 		</dependency>
 		<dependency>
+		<groupId>org.springframework</groupId>
+		<artifactId>spring-web</artifactId>
+		<version>5.3.9</version>
+		</dependency>
+
+		<dependency>
 			<groupId>org.apache.velocity</groupId>
 			<artifactId>velocity</artifactId>
 		</dependency>
@@ -296,12 +302,19 @@
 			<groupId>org.testcontainers</groupId>
 			<artifactId>mysql</artifactId>
 		</dependency>
+<<<<<<< HEAD
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-web</artifactId>
 			<version>5.3.13</version>
 		</dependency>
 
+=======
+<<<<<<< HEAD
+=======
+
+>>>>>>> e09e68135 (TRUNK-7521: Complex obs handlers to return obs as is when underlying file is null)
+>>>>>>> 58a93d278 (TRUNK-7521: Complex obs handlers to return obs as is when underlying file is null)
 	</dependencies>
 	<build>
 		<resources>
--- a/api/src/main/java/org/openmrs/obs/ComplexData.java
+++ b/api/src/main/java/org/openmrs/obs/ComplexData.java
@@ -9,6 +9,7 @@
  */
 package org.openmrs.obs;
 
+import java.io.File;
 import java.io.InputStream;
 
 /**
@@ -36,11 +37,11 @@ public class ComplexData implements java.io.Serializable {
 	/**
 	 * Default constructor requires title and data.
 	 * 
-	 * @param title Name or brief description of ComplexData.
+	 * @param fILENAME Name or brief description of ComplexData.
 	 * @param data The complex data for an Obs
 	 */
-	public ComplexData(String title, Object data) {
-		setTitle(title);
+	public ComplexData(String fILENAME, Object data) {
+		setTitle(fILENAME);
 		setData(data);
 	}
 	
@@ -119,6 +120,10 @@ public class ComplexData implements java.io.Serializable {
 	 */
 	public Long getLength() {
 		return this.length;
+	}
+	
+	public File[] getFiles() {
+		return null;
 	}
 	
 }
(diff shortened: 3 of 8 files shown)


jona42-ui:
cc/ @dkayiwa @

mks-d (commented):
This should most likely come with test case(s) that clearly illustrate the [given/when/then](https://www.newthings.co/blog/structuring-tests-using-given-when-then/) of what is intended here.

jona42-ui:
great, thanks @mks-d on  my radar, next week

ibacher:
I'm not sure I see how this actually addresses the NPEs mentioned in the ticket...

jona42-ui:
@ibacher does this handle all NPEs you woul xpect such that I can do test cases for the scenarios?

ibacher:
Do you have test cases as @mks-d suggested? It looks like this handles some NPEs; I'm not convinced that the `IllegalArgumentException`s are a good idea, though...

jona42-ui:
> Do you have test cases as @mks-d suggested? It looks like this handles some NPEs; I'm not convinced that the `IllegalArgumentException`s are a good idea, though...

sure, still fighting with error handling in test cases.
sure, i may not be 100% certain of suitability of IllegalArgumentExceptions but i provided it to quickly catch errors.  A more  optimised exception handler would be glad to use.

api/mytxtfile.txt
```diff
@@ -0,0 +1 @@
+test content
```
ibacher, line 1: Test data always goes in `src/test` (and then an appropriate sub-directory.

api/pom.xml
```diff
@@ -296,6 +296,12 @@
 			<groupId>org.testcontainers</groupId>
 			<artifactId>mysql</artifactId>
 		</dependency>
+		<dependency>
```
ibacher, line 299: The API project should not depend on spring-web

api/src/main/java/org/openmrs/obs/ComplexData.java
```diff
@@ -36,11 +37,11 @@ public class ComplexData implements java.io.Serializable {
 	/**
 	 * Default constructor requires title and data.
 	 * 
-	 * @param title Name or brief description of ComplexData.
+	 * @param fILENAME Name or brief description of ComplexData.
 	 * @param data The complex data for an Obs
 	 */
-	public ComplexData(String title, Object data) {
-		setTitle(title);
+	public ComplexData(String fILENAME, Object data) {
```
ibacher, line 43: I'm not sure why this parameter needs to be renamed. ComplexData can store more than just files, though one use of this field is for filenames.

api/src/main/java/org/openmrs/obs/ComplexData.java
```diff
@@ -121,4 +122,8 @@ public Long getLength() {
 		return this.length;
 	}
 	
+	public File[] getFiles() {
```
ibacher, line 125: We shouldn't need this...

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -103,8 +107,12 @@ public String getExtension(String filename) {
 	/**
 	 * @see org.openmrs.obs.ComplexObsHandler#getObs(Obs, String)
 	 */
-	public Obs getObs(Obs obs, String view) {
+	public Obs getObs(Obs obs, String fILENAME) {
```
ibacher, line 110: This change is also unnecessary...

api/src/main/java/org/openmrs/obs/handler/BinaryDataHandler.java
```diff
@@ -52,6 +52,10 @@ public BinaryDataHandler() {
 	@Override
 	public Obs getObs(Obs obs, String view) {
 		File file = getComplexDataFile(obs);
+		//will return the original obs object if the valueComplex field is null or empty
+		if (file == null) {
```
ibacher, line 56: Probably should check if the obs is null too right?

api/src/main/java/org/openmrs/obs/handler/BinaryDataHandler.java
```diff
@@ -60,8 +64,9 @@ public Obs getObs(Obs obs, String view) {
 		if (ComplexObsHandler.RAW_VIEW.equals(view)) {
 			// to handle problem with downloading/saving files with blank spaces or commas in their names
 			// also need to remove the "file" text appended to the end of the file name
-			String[] names = obs.getValueComplex().split("\\|");
-			String originalFilename = names[0];
+			// null check
+			String[] names = (obs.getValueComplex()!=null)? obs.getValueComplex().split("\\|"):new String[0];
+			String originalFilename =  (names.length > 0) ? names[0] : "";
```
ibacher, line 69: The parentheses here are unnecessary.

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
```diff
@@ -43,6 +44,8 @@ public class AbstractHandlerTest extends BaseContextSensitiveTest {
 	private final String FILENAME = "mytxtfile.txt";
 	
 	private  AbstractHandler handler;
+
+	private Obs obs;
```
ibacher, line 48: It's better not to introduce a class-level variable in tests where it can be avoided.

api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
```diff
@@ -119,7 +119,7 @@
 @ExtendWith(MockitoExtension.class)
 public abstract class BaseContextSensitiveTest {
 	
-	private static final Logger log = LoggerFactory.getLogger(BaseContextSensitiveTest.class);
+	public static final Logger log = LoggerFactory.getLogger(BaseContextSensitiveTest.class);
```
ibacher, line 122: Please don't change this visibility if you need a logger in your test, create a class-specific version.

ibacher (changes requested):

jona42-ui:
sorry, have to clean this up

jona42-ui:
un intended!!

PR #4290, opened 2023-03-28
TRUNK-6171: Address inconsistency issues with ConditionService

mseaton:
This makes the following changes to conditions:
* Fixes the ConditionService#saveCondition method to only void and recreate if a Condition has changed
* Fixes the ConditionService#saveCondition method to not exclude updates to properties when a Condition is voided
* Fixes the ConditionService#saveCondition method to not exclude updates to properties when a Condition is unvoided
* Fixes the ConditionService#getAllConditions(Patient) method to exclude voided Conditions
* Adds missing hashcode and equals methods to CodedOrFreeText
* Add new method to Condition to compare whether it matches another Condition

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-6171

### DIFF
--- a/api/src/main/java/org/openmrs/CodedOrFreeText.java
+++ b/api/src/main/java/org/openmrs/CodedOrFreeText.java
@@ -9,9 +9,6 @@
  */
 package org.openmrs;
 
-import org.apache.commons.lang.builder.EqualsBuilder;
-import org.apache.commons.lang.builder.HashCodeBuilder;
-
 import javax.persistence.Embeddable;
 import javax.persistence.ManyToOne;
 
@@ -51,24 +48,6 @@ public class CodedOrFreeText {
 		this.coded = coded;
 		this.specificName = specificName;
 		this.nonCoded = nonCoded;
-	}
-
-	@Override
-	public int hashCode() {
-		return new HashCodeBuilder().append(coded).append(specificName).append(nonCoded).toHashCode();
-	}
-
-	@Override
-	public boolean equals(Object object) {
-		if (object == null || !(object instanceof CodedOrFreeText)) {
-			return false;
-		}
-		CodedOrFreeText that = (CodedOrFreeText) object; 
-		EqualsBuilder b = new EqualsBuilder();
-		b.append(this.coded, that.coded);
-		b.append(this.specificName, that.specificName);
-		b.append(this.nonCoded, that.nonCoded);
-		return b.isEquals();
 	}
 
 	/**
--- a/api/src/main/java/org/openmrs/CodedOrFreeText.java
+++ b/api/src/main/java/org/openmrs/CodedOrFreeText.java
@@ -9,6 +9,9 @@
  */
 package org.openmrs;
 
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+
 import javax.persistence.Embeddable;
 import javax.persistence.ManyToOne;
 
@@ -49,7 +52,25 @@ public class CodedOrFreeText {
 		this.specificName = specificName;
 		this.nonCoded = nonCoded;
 	}
-	
+
+	@Override
+	public int hashCode() {
+		return new HashCodeBuilder().append(coded).append(specificName).append(nonCoded).toHashCode();
+	}
+
+	@Override
+	public boolean equals(Object object) {
+		if (object == null || !(object instanceof CodedOrFreeText)) {
+			return false;
+		}
+		CodedOrFreeText that = (CodedOrFreeText) object; 
+		EqualsBuilder b = new EqualsBuilder();
+		b.append(this.coded, that.coded);
+		b.append(this.specificName, that.specificName);
+		b.append(this.nonCoded, that.nonCoded);
+		return b.isEquals();
+	}
+
 	/**
 	 * Gets the coded concept
 	 *
--- a/api/src/main/java/org/openmrs/CodedOrFreeText.java
+++ b/api/src/main/java/org/openmrs/CodedOrFreeText.java
@@ -49,7 +49,7 @@ public class CodedOrFreeText {
 		this.specificName = specificName;
 		this.nonCoded = nonCoded;
 	}
-
+	
 	/**
 	 * Gets the coded concept
 	 *
--- a/api/src/main/java/org/openmrs/Condition.java
+++ b/api/src/main/java/org/openmrs/Condition.java
@@ -165,13 +165,18 @@ public class Condition extends BaseFormRecordableOpenmrsData {
 		if (c == null) {
 			return false;
 		}
+		CodedOrFreeText coft1 = getCondition() == null ? new CodedOrFreeText() : getCondition();
+		CodedOrFreeText coft2 = c.getCondition() == null ? new CodedOrFreeText() : c.getCondition();
+		
 		boolean ret = (OpenmrsUtil.nullSafeEquals(getPreviousVersion(), c.getPreviousVersion()));
 		ret = ret && (OpenmrsUtil.nullSafeEquals(getPatient(), c.getPatient()));
 		ret = ret && (OpenmrsUtil.nullSafeEquals(getEncounter(), c.getEncounter()));
 		ret = ret && (OpenmrsUtil.nullSafeEquals(getFormNamespaceAndPath(), c.getFormNamespaceAndPath()));
 		ret = ret && (OpenmrsUtil.nullSafeEquals(getClinicalStatus(), c.getClinicalStatus()));
 		ret = ret && (OpenmrsUtil.nullSafeEquals(getVerificationStatus(), c.getVerificationStatus()));
-		ret = ret && (OpenmrsUtil.nullSafeEquals(getCondition(), c.getCondition()));
+		ret = ret && (OpenmrsUtil.nullSafeEquals(coft1.getCoded(), coft2.getCoded()));
+		ret = ret && (OpenmrsUtil.nullSafeEquals(coft1.getSpecificName(), coft2.getSpecificName()));
+		ret = ret && (OpenmrsUtil.nullSafeEquals(coft1.getNonCoded(), coft2.getNonCoded()));
 		ret = ret && (OpenmrsUtil.nullSafeEquals(getOnsetDate(), c.getOnsetDate()));
 		ret = ret && (OpenmrsUtil.nullSafeEquals(getAdditionalDetail(), c.getAdditionalDetail()));
 		ret = ret && (OpenmrsUtil.nullSafeEquals(getEndDate(), c.getEndDate()));
--- a/api/src/main/java/org/openmrs/Condition.java
+++ b/api/src/main/java/org/openmrs/Condition.java
@@ -9,6 +9,8 @@
  */
 package org.openmrs;
 
+import org.openmrs.util.OpenmrsUtil;
+
 import javax.persistence.AssociationOverride;
 import javax.persistence.AssociationOverrides;
 import javax.persistence.AttributeOverride;
@@ -25,7 +27,6 @@ import javax.persistence.JoinColumn;
 import javax.persistence.ManyToOne;
 import javax.persistence.Table;
 import javax.persistence.Transient;
-
 import java.util.Date;
 
 /**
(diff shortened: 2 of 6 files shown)


api/src/main/java/org/openmrs/api/db/hibernate/HibernateConditionDAO.java
```diff
@@ -106,8 +106,10 @@ public List<Condition> getActiveConditions(Patient patient) {
 	@Override
 	public List<Condition> getAllConditions(Patient patient) {
 		Query<Condition> query = sessionFactory.getCurrentSession().createQuery(
-				"from Condition con where con.patient.patientId = :patientId " +
-						"order by con.dateCreated desc", Condition.class);
+				"from Condition c " +
+					"where c.patient.patientId = :patientId " +
+					"and c.voided = false " +
```
ibacher, line 111: IIRC, at some point the condition implementation was using `voided` as meaningful data, which is why it was left in here.

ibacher (commented):
Thanks @mseaton! Lots of good stuff here! A couple of little niggles.

api/src/main/java/org/openmrs/CodedOrFreeText.java
```diff
@@ -49,7 +52,25 @@ public CodedOrFreeText(Concept coded, ConceptName specificName, String nonCoded)
 		this.specificName = specificName;
 		this.nonCoded = nonCoded;
 	}
-	
+
+	@Override
+	public int hashCode() {
```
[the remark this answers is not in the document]
mseaton: Yeah, I wasn't really sure about this either, but I'm also not sure about what you are saying.  Why is a core part of `CodedOrFreeText`s identity the parent instance?  A core part of a Date, String, or Integer property is not the parent instance?  Isn't `CodedOrFreeText` just a shortcut for mapping `coded`, `nonCoded`, and `specificName` into this Object and also into certain database columns?  

It's easy enough to change, though I would have like to keep things consistent as seen in the below where I use `OpenmrsUtil.nullSafeEquals` to compare the condition property just like every other property.

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
```diff
@@ -101,46 +103,53 @@ public List<Condition> getConditionsByEncounter(Encounter encounter) throws APIE
 	 */
 	@Override
 	public Condition saveCondition(Condition condition) throws APIException {
-		Condition existingCondition = Context.getConditionService().getConditionByUuid(condition.getUuid());
+		
 		// If there is no existing condition, then we are creating a condition
-		if (existingCondition == null) {
+		Integer existingConditionId = condition.getConditionId();
```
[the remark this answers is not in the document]
mseaton: Can you explain this a bit more @ibacher ?  I believe the code I have below is accomplishing the same thing, but more reliably.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConditionDAO.java
ibacher: Oh, yeah!

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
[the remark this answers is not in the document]
ibacher: [This bit of code](https://github.com/openmrs/openmrs-module-emrapi/blob/79a5af9dac23f651b2073b413b871dd4444cc773/web-2.2/src/main/java/org/openmrs/module/emrapi/web/controller/ConditionController.java#L257). Basically, if a page POSTs to the `/rest/emrapi/condition` endpoint, it will create a "new" condition without setting the `id` on that condition, even if the condition was already an existing one; the only unique identifier it uses is the UUID.

api/src/main/java/org/openmrs/CodedOrFreeText.java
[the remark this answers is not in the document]
ibacher: Yeah, you know, I don't have anything more than a vague feeling. Ultimately, it probably affect anything.

api/src/main/java/org/openmrs/Condition.java
```diff
@@ -136,6 +150,43 @@ public static Condition copy(Condition fromCondition, Condition toCondition) {
...
+		
+		boolean ret = (OpenmrsUtil.nullSafeEquals(getPreviousVersion(), c.getPreviousVersion()));
+		ret = ret && (OpenmrsUtil.nullSafeEquals(getPatient(), c.getPatient()));
+		ret = ret && (OpenmrsUtil.nullSafeEquals(getEncounter(), c.getEncounter()));
+		ret = ret && (OpenmrsUtil.nullSafeEquals(getFormNamespaceAndPath(), c.getFormNamespaceAndPath()));
+		ret = ret && (OpenmrsUtil.nullSafeEquals(getClinicalStatus(), c.getClinicalStatus()));
+		ret = ret && (OpenmrsUtil.nullSafeEquals(getVerificationStatus(), c.getVerificationStatus()));
+		ret = ret && (OpenmrsUtil.nullSafeEquals(coft1.getCoded(), coft2.getCoded()));
+		ret = ret && (OpenmrsUtil.nullSafeEquals(coft1.getSpecificName(), coft2.getSpecificName()));
+		ret = ret && (OpenmrsUtil.nullSafeEquals(coft1.getNonCoded(), coft2.getNonCoded()));
```
mseaton, line 179: OK, @ibacher .  I removed the changes that added in hashcode and equals methods to CodedOrFreeText, and am now just comparing each property directly here.

api/src/main/java/org/openmrs/CodedOrFreeText.java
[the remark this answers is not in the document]
mseaton: Better to be safe than sorry.  I removed it.

api/src/main/java/org/openmrs/Condition.java
```diff
@@ -136,6 +150,43 @@ public static Condition copy(Condition fromCondition, Condition toCondition) {
...
+	 * Compares properties with those in the given Condition to determine if they have the same meaning
+	 * This method will return true immediately following the creation of a Condition from another Condition
+	 * @see Condition#newInstance(Condition)
+	 * This method will return false if any value is different, excepting identity data (id, uuid)
+	 * If the given instance is null, this will return false
+	 * @param c the Condition to compare against 	
+	 * @return true if the given Condition has the same meaningful properties as the passed Condition
+	 * @since 2.6.1
+	 */
+	public boolean matches(Condition c) {
```
mogoodrich, line 164: Annoying, but would good to have a couple quick test cases around this.

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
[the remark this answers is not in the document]
mseaton: I see.  Yeah that'll need to get fixed.  That's an odd implementation, though consistent with the way the unit tests that I changed were operating.  Basically, you should need to fetch an existing object from the DB and edit it to update it. Creating a new object and giving it the same uuid of an existing object to reflect the desire to do an update is not a pattern I've ever seen nor would I endorse.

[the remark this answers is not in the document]
mogoodrich: I haven't dug into the reference above, but, yes, if an OpenMRS object is incoming via REST (and therefore won't have an id), is it the REST processing code that is expected to handle whether there's an existing object with the same uuid, or is this API supposed to handle it?  I think the answer is that the REST processing code generally should, and so the fix (if any is needed?) should be made in EMR-API?

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
```diff
@@ -101,46 +103,53 @@ public List<Condition> getConditionsByEncounter(Encounter encounter) throws APIE
...
-			if (!existingCondition.getVoided()) {
-				return Context.getConditionService().voidCondition(existingCondition,
-					StringUtils.isNotBlank(condition.getVoidReason()) ? condition.getVoidReason() : "Condition deleted");
-			} else {
-				return existingCondition;
-			}
+		
+		// If there is an existing condition, create a new condition from it and reset the existing instance state
+		Condition newCondition = Condition.newInstance(condition);
+		Context.refreshEntity(condition);
```
mogoodrich, line 115: what does this do again? loads it in from the DB/cache again? (ie prior to any changes?)

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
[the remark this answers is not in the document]
ibacher: Not how I would've written it either! I was trying to keep the EMRAPI working the same way so that my two Condition PRs didn't directly depend on each other.

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
```diff
@@ -101,46 +103,53 @@ public List<Condition> getConditionsByEncounter(Encounter encounter) throws APIE
...
+		
+		// Determine how the existing state and new state have changed
+		boolean conditionHasChanged = !newCondition.matches(condition);
+		boolean existingVoided = BooleanUtils.isTrue(condition.getVoided());
+		boolean newVoided = BooleanUtils.isTrue(newCondition.getVoided());
+		boolean unVoidOriginal = existingVoided && !newVoided;
```
mogoodrich, line 121: we unvoid the original even if the new one differs from the old? wouldn't we just create a new condition in this case that we both want to unvoid and modify?

mogoodrich (commented):
A few questions, but generally looks good!

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
mseaton: It's a good point.  I had originally re-written that way, and it was failing, which is why I preserved the existing behavior.  I'll have a look.

api/src/main/java/org/openmrs/Condition.java
mseaton: Yeah

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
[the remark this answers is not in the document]
mseaton: OK, that is fixed and merged here https://github.com/openmrs/openmrs-module-emrapi/commit/09ddcb6132d547247d8eb0de15361ca4e8eb4350.  Thanks @ibacher  for catching and recognizing that, and for the super quick reviews.

api/src/test/java/org/openmrs/ConditionTest.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Tests of methods within the Condition class
+ * @see Condition
+ */
+public class ConditionTest {
```
mseaton, line 27: Unit test added here for the Condition matches method @mogoodrich

api/src/main/java/org/openmrs/Condition.java
mseaton: Added, see comment below @mogoodrich

api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
mseaton: OK, per @mogoodrich comment and suggestion, I've removed the special "unvoid" handling, which I agree is unnecessary and wrong.  Unvoiding shouldn't eliminate the audit trail of the previously voided data, it should just be treated like an edit, linking back to voided row.  That is now working.

mseaton:
@ibacher and @mogoodrich - I _think_ I've addressed all of your comments.  Let me know how things are looking...

mogoodrich (approved):
LGTM, thanks @mseaton

ibacher (approved):
LGTM

PR #4299, opened 2023-04-07
Refactoring the code changes 

SankeerthSharmaR:
I was not working on the any existing issues.
Refactored the existing code base which has the code smells.

Did couple of refactoring for the existing classes:
1) for instance introducing explaining variable (Like magic no smell where directly take the length of /web/src to string.length()
2)And Decompose conditional if the method has more no of conditions in if statement.
3)Extract method if the method is too complex.(having large cyclomatic complexity)

Along with that:

1)Replace  conditional with polymorphism by adding extract classes where we have if else conditions and any future changes to made for that class
2)Extract Class refactoring by adding a new class and moving the method to it which are cohesive to another class
3)Move method (private static boolean isValid(Locale locale) ) to webutil java as it is used here.

### DIFF
--- a/api/src/main/java/liquibase/ext/sqlgenerator/core/DB2Database.java
+++ b/api/src/main/java/liquibase/ext/sqlgenerator/core/DB2Database.java
@@ -1,0 +1,8 @@
+package liquibase.ext.sqlgenerator.core;
+
+public class DB2Database implements DatabaseModifier {
+	
+	public String getModifyString() {
+		return "ALTER COLUMN";
+	}
+}
--- a/api/src/main/java/liquibase/ext/sqlgenerator/core/DatabaseModifier.java
+++ b/api/src/main/java/liquibase/ext/sqlgenerator/core/DatabaseModifier.java
@@ -1,0 +1,7 @@
+package liquibase.ext.sqlgenerator.core;
+
+public interface DatabaseModifier {
+	
+	String getModifyString();
+	
+}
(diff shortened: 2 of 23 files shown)


jwnasambu:
@SankeerthSharmaR kindly include the description of what you changed please. Besides, kindly does the ticket exist on the  the issue tracker  for this PR? if yes, kindly fix the correct link else remove the link you shared before its pointing to a wrong page.

SankeerthSharmaR:
@jwnasambu 
I was not working on the any existing issues.
Refactored the existing code base which has the code smells and fixed few of them
And removed the link which is pointing to wrong page.

jwnasambu:
Oh sorry you have a build failure. Kindly the error message you encountered is related to the license-maven-plugin in a Maven project. This plugin is used to enforce license headers in source code files. The error indicates that some files in the openmrs-api project do not have the expected license header.

To fix this error, you can follow these steps:

Identify the files that are causing the error. 

Open each file mentioned in the error message and ensure that it contains the correct license header. The license header is typically a comment at the top of the file that specifies the license under which the code is distributed.

If the files are missing the license header or have an incorrect one, you need to add or update the license header. The specific license requirements may vary depending on the project. You can usually find the correct license header format in the project's documentation or by examining other files in the project that have the correct license header.

Once you have updated the license headers in the problematic files, save the changes.

Rebuild the project using Maven to verify if the error has been resolved. You can do this by running the following command in the terminal or command prompt:

`mvn clean install`

This command will clean the project, recompile the code, and run the necessary Maven goals, including the license check.

If the error persists, ensure that the license-maven-plugin is properly configured in your project's pom.xml file. Check that the plugin version is correct and that the configuration specifies the expected license headers.

Save the changes to the pom.xml file and rebuild the project again.

By following these steps, you should be able to fix the error related to the missing or incorrect license headers in your Maven project.






Regenerate response

tendomart:
And add a link to  the Ticket you're working on.

PR #4322, opened 2023-05-11
TRUNK-6179 - Authenticating should not override chosen location

mseaton:
see https://issues.openmrs.org/browse/TRUNK-6179

Note:  This PR is against 2.5.x, and will be up-ported to all later versions once approved and committed.


mogoodrich (commented):
LGTM, thanks @mseaton !

PR #4369, opened 2023-08-19
TRUNK-6185: Easier configuration of csrfguard.properties

wikumChamith:

This commit updates the loadCsrfGuardProperties method to enhance the loading of CSRF Guard properties. It first checks for the existence of the csrfguard.properties file in the `.OpenMRS` directory. If the file exists, its contents are loaded. Otherwise, the default values are loaded from the `/WEB-INF/csrfguard.properties` file. Additionally, properties starting with `org.owasp.csrfguard` from `openmrs-runtime.properties` are merged into the loaded properties. 

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6185

  No? -> execute above command

### DIFF
--- a/web/src/main/java/org/openmrs/web/Listener.java
+++ b/web/src/main/java/org/openmrs/web/Listener.java
@@ -264,20 +264,26 @@ public final class Listener extends ContextLoader implements ServletContextListe
 		File csrfGuardFile = new File(OpenmrsUtil.getApplicationDataDirectory(), "csrfguard.properties");
 		InputStream csrfGuardInputStream = null;
 		Properties csrfGuardProperties = new Properties();
-		if(csrfGuardFile.exists()) {
+		try {
 			csrfGuardInputStream = Files.newInputStream(csrfGuardFile.toPath());
-		}
-		else {
+		} 
+		catch (IOException ioException) {
 			final String fileName = servletContext.getRealPath("/WEB-INF/csrfguard.properties");
-			csrfGuardInputStream = Files.newInputStream(Paths.get(fileName));
-			Properties runtimeProperties = OpenmrsUtil.getRuntimeProperties(WebConstants.WEBAPP_NAME);
-			runtimeProperties.stringPropertyNames().forEach(property -> {
-				 if(property.contains("org.owasp.csrfguard")) {
-					 csrfGuardProperties.setProperty(property, runtimeProperties.getProperty(property));
-				 }
-			 });
+			try {
+				csrfGuardInputStream = Files.newInputStream(Paths.get(fileName));
+			} 
+			catch (IOException ioException1) {
+				ioException1.printStackTrace();
+				throw ioException1;
+			}
 		}
 		csrfGuardProperties.load(csrfGuardInputStream);
+		Properties runtimeProperties = OpenmrsUtil.getRuntimeProperties(WebConstants.WEBAPP_NAME);
+		runtimeProperties.stringPropertyNames().forEach(property -> {
+			if(property.startsWith("org.owasp.csrfguard")) {
+				csrfGuardProperties.setProperty(property, runtimeProperties.getProperty(property));
+			}
+		});		
 		IOUtils.closeQuietly(csrfGuardInputStream);
 		CsrfGuard.load(csrfGuardProperties);
 
--- a/web/src/main/java/org/openmrs/web/Listener.java
+++ b/web/src/main/java/org/openmrs/web/Listener.java
@@ -63,6 +63,7 @@ import java.io.OutputStreamWriter;
 import java.io.StringReader;
 import java.lang.reflect.Field;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.sql.Driver;
 import java.sql.DriverManager;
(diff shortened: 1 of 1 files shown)


mseaton (changes requested):

wikumChamith:
@mseaton I addressed the issues you pointed out.

mseaton:
@wikumChamith , no this doesn't look right.  You shouldn't be putting logic into the catch block, and you still don't have any finally blocks where you close the input streams.

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -258,29 +259,34 @@ public void contextInitialized(ServletContextEvent event) {
...
-		    
-		    //Moved to EOF by the copy operation. So open it again.
-		    inputStream = new FileInputStream(file);
-		}
-		Properties properties = new Properties();
-		properties.load(inputStream);
-		IOUtils.closeQuietly(inputStream);
-		CsrfGuard.load(properties);
-		
+			try {
```
ibacher, line 272: Doesn't seem right to do this _unconditionally_ in the catch block. There are numerous reasons an `IOException` could be thrown above, but only one where we'd want to fallback: a `FileNotFound` exception. On top of which, the error here is always swallowed.

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -258,29 +259,34 @@ public void contextInitialized(ServletContextEvent event) {
...
+				csrfGuardInputStream = Files.newInputStream(Paths.get(fileName));
+			} 
+			catch (IOException ioException1) {
+				ioException1.printStackTrace();
```
ibacher, line 276: Log, rather than printing.

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -258,29 +259,34 @@ public void contextInitialized(ServletContextEvent event) {
...
+				throw ioException1;
+			}
+		}
+		csrfGuardProperties.load(csrfGuardInputStream);
+		Properties runtimeProperties = OpenmrsUtil.getRuntimeProperties(WebConstants.WEBAPP_NAME);
```
ibacher, line 281: It's preferable to use `Context.getRuntimeProperties()`

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -258,29 +259,34 @@ public void contextInitialized(ServletContextEvent event) {
...
+		runtimeProperties.stringPropertyNames().forEach(property -> {
+			if(property.startsWith("org.owasp.csrfguard")) {
+				csrfGuardProperties.setProperty(property, runtimeProperties.getProperty(property));
+			}
+		});		
+		IOUtils.closeQuietly(csrfGuardInputStream);
```
ibacher, line 287: This really needs to be done in a finally block. There are way to many ways an exception can be thrown in the above code and so the IO stream never closed.

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -258,29 +259,34 @@ public void contextInitialized(ServletContextEvent event) {
 			log.error(MarkerFactory.getMarker("FATAL"), "Failed to obtain JDBC connection", e);
 		}
 	}
-	
-	private void loadCsrfGuardProperties(ServletContext servletContext) throws FileNotFoundException, IOException {	
-		File file = new File(OpenmrsUtil.getApplicationDataDirectory(), "csrfguard.properties");
-		InputStream inputStream = null;
+
+	private void loadCsrfGuardProperties(ServletContext servletContext) throws IOException {
```
ibacher, line 263: Note that `loadCsrfGuardProperties()` is currently called _before_ the runtime properties are setup, which means `OpenmrsUtil.getApplicationDirectory()` (inter alia) can return the wrong folder. Ideally, this call should be moved to after the runtime properties are loaded.

wikumChamith:
@mseaton, @ibacher I refactored the code according to your feedback.

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -258,28 +257,35 @@ public void contextInitialized(ServletContextEvent event) {
...
+		File csrfGuardFile = new File(OpenmrsUtil.getApplicationDataDirectory(), "csrfguard.properties");
+		Properties csrfGuardProperties = new Properties();
+		if(csrfGuardFile.exists()) {
+			try(InputStream csrfGuardInputStream = Files.newInputStream(csrfGuardFile.toPath())) {
+				csrfGuardProperties.load(csrfGuardInputStream);
+			}
+			catch (Exception e) {
+				log.error(e.getMessage());
```
mseaton, line 269: I'm not sure about just logging an error here and continuing.  Why catch this Exception if you aren't going to fall back to other behavior?  If we want to log this here (which is probably a good idea), then I'd add some additional context in the logging message to make it easier for someone to know where the problem is and what went wrong, and then re-throw the Exception.

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -258,28 +257,35 @@ public void contextInitialized(ServletContextEvent event) {
...
+			}
+		}
+		else {
+			String fileName = servletContext.getRealPath("/WEB-INF/csrfguard.properties");
+			try (InputStream csrfGuardInputStream = Files.newInputStream(Paths.get(fileName))) {
+				csrfGuardProperties.load(csrfGuardInputStream);
+			}
+			catch (Exception e) {
+				log.error(e.getMessage());
```
mseaton, line 278: Same comment here.  Just log, with a more helpful message for implementers to better understand the error, and then re-throw the Exception

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -258,28 +257,35 @@ public void contextInitialized(ServletContextEvent event) {
...
+			}
+		}
+		
+		Properties runtimeProperties = getRuntimeProperties();
+		runtimeProperties.stringPropertyNames().forEach(property -> {
+			if(property.startsWith("org.owasp.csrfguard")) {
```
mseaton, line 284: Style comment - please put a space between the `if` and the `(`.  Apply this elsewhere in this PR as well.

mseaton (changes requested):
Looking much better - a few additional comments to consider.

wikumChamith:
@mseaton I addressed the issues you pointed out. Please review the new code.

mseaton (commented):
Just a few minor style comments @wikumChamith and then I think it looks good.  The only other thing that I'd love to see are some unit tests around this, if this is possible to do.  Not sure if @ibacher has any additional feedback.

wikumChamith:
@mseaton I addressed the suggestions you made. Any suggestions on a viable way that I can write tests for this method?

mseaton (approved):

PR #4374, opened 2023-08-27
TRUNK-6189:Provider Service class should provide ability to getProviderAttributeTypeByName()

mherman22:
…erAttributeTypeByName()

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6189

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/ProviderService.java
+++ b/api/src/main/java/org/openmrs/api/ProviderService.java
@@ -225,6 +225,16 @@ public interface ProviderService extends OpenmrsService {
 	public ProviderAttributeType getProviderAttributeType(Integer providerAttributeTypeId);
 	
 	/**
+	 * Get a provider attribute type by it's name
+	 *
+	 * @param name the name of the provider attribute type
+	 * @return the provider attribute type for the given name
+	 * 	<strong>Should</strong> get the provider attribute type by it's name
+	 */
+	@Authorized( { PrivilegeConstants.GET_PROVIDER_ATTRIBUTE_TYPES })
+	public List<ProviderAttributeType> getProviderAttributeTypeByName(String name);
+	
+	/**
 	 * Get a provider attribute type by it's uuid
 	 * 
 	 * @param uuid the uuid of the provider attribute type
--- a/api/src/main/java/org/openmrs/api/ProviderService.java
+++ b/api/src/main/java/org/openmrs/api/ProviderService.java
@@ -232,7 +232,7 @@ public interface ProviderService extends OpenmrsService {
 	 * 	<strong>Should</strong> get the provider attribute type by it's name
 	 */
 	@Authorized( { PrivilegeConstants.GET_PROVIDER_ATTRIBUTE_TYPES })
-	public List<ProviderAttributeType> getProviderAttributeTypeByName(String name);
+	public ProviderAttributeType getProviderAttributeTypeByName(String name);
 	
 	/**
 	 * Get a provider attribute type by it's uuid
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateProviderDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateProviderDAO.java
@@ -279,6 +279,16 @@ public class HibernateProviderDAO implements ProviderDAO {
 		return (ProviderAttributeType) getSession().get(ProviderAttributeType.class, providerAttributeTypeId);
 	}
 	
+	/**
+	 * @see org.openmrs.api.db.ProviderDAO#getProviderAttributeTypesByName(String) 
+	 */
+	@Override
+	public List<ProviderAttributeType> getProviderAttributeTypesByName(String name) {
+		Criteria criteria = getSession().createCriteria(ProviderAttributeType.class);
+		criteria.add(Restrictions.eq("name", name));
+		return criteria.list();
+	}
+	
 	/* (non-Javadoc)
 	 * @see org.openmrs.api.db.ProviderDAO#getProviderAttributeTypeByUuid(java.lang.String)
 	 */
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateProviderDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateProviderDAO.java
@@ -283,10 +283,10 @@ public class HibernateProviderDAO implements ProviderDAO {
 	 * @see org.openmrs.api.db.ProviderDAO#getProviderAttributeTypesByName(String) 
 	 */
 	@Override
-	public List<ProviderAttributeType> getProviderAttributeTypesByName(String name) {
+	public ProviderAttributeType getProviderAttributeTypeByName(String name) {
 		Criteria criteria = getSession().createCriteria(ProviderAttributeType.class);
 		criteria.add(Restrictions.eq("name", name));
-		return criteria.list();
+		return (ProviderAttributeType) criteria.uniqueResult();
 	}
 	
 	/* (non-Javadoc)
--- a/api/src/main/java/org/openmrs/api/impl/ProviderServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ProviderServiceImpl.java
@@ -205,6 +205,15 @@ public class ProviderServiceImpl extends BaseOpenmrsService implements ProviderS
 	}
 	
 	/**
+	 * @see org.openmrs.api.ProviderService#getProviderAttributeTypeByName(java.lang.String)
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	public List<ProviderAttributeType> getProviderAttributeTypeByName(String name) {
+		return dao.getProviderAttributeTypesByName(name);
+	}
+	
+	/**
 	 * @see org.openmrs.api.ProviderService#getProviderAttributeTypeByUuid(java.lang.String)
 	 */
 	@Override
--- a/api/src/main/java/org/openmrs/api/impl/ProviderServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ProviderServiceImpl.java
@@ -209,8 +209,8 @@ public class ProviderServiceImpl extends BaseOpenmrsService implements ProviderS
 	 */
 	@Override
 	@Transactional(readOnly = true)
-	public List<ProviderAttributeType> getProviderAttributeTypeByName(String name) {
-		return dao.getProviderAttributeTypesByName(name);
+	public ProviderAttributeType getProviderAttributeTypeByName(String name) {
+		return dao.getProviderAttributeTypeByName(name);
 	}
 	
 	/**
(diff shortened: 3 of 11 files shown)


api/src/main/java/org/openmrs/api/ProviderService.java
```diff
@@ -224,6 +224,15 @@ public List<Provider> getProviders(String query, Integer start, Integer length,
...
 	public ProviderAttributeType getProviderAttributeType(Integer providerAttributeTypeId);
 	
+	/**
+	 * Get a provider attribute type by it's name
+	 *
+	 * @param name the name of the provider attribute type
+	 * @return the provider attribute type for the given uuid
+	 * <strong>Should</strong> get the provider attribute type by it's name
+	 */
+	public ProviderAttributeType getProviderAttributeTypeByName(String name);
```
dkayiwa, line 234: Shouldn't you have the @Authorized annotation on this?

api/src/main/java/org/openmrs/api/impl/ProviderServiceImpl.java
```diff
@@ -204,6 +204,14 @@ public ProviderAttributeType getProviderAttributeType(Integer providerAttributeT
 		return dao.getProviderAttributeType(providerAttributeTypeId);
 	}
 	
+	/**
+	 * @see org.openmrs.api.ProviderService#getProviderAttributeTypeByName(java.lang.String) 
+	 */
+	@Override
+	public ProviderAttributeType getProviderAttributeTypeByName(String name) {
```
dkayiwa, line 211: Shouldn't this be transactional readonly?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateProviderDAO.java
```diff
@@ -271,6 +271,12 @@ private <T> T getByUuid(String uuid, Class<T> clazz) {
 		return (T) criteria.uniqueResult();
 	}
 	
+	private <T> T getByName(String name, Class<T> tClass) {
```
dkayiwa, line 274: I do not like this method name.

api/src/main/java/org/openmrs/api/ProviderService.java
mherman22: added!

api/src/main/java/org/openmrs/api/impl/ProviderServiceImpl.java
mherman22: added!

api/src/main/java/org/openmrs/api/db/hibernate/HibernateProviderDAO.java
mherman22: changed!

api/src/main/java/org/openmrs/api/db/hibernate/HibernateProviderDAO.java
```diff
@@ -271,6 +271,12 @@ private <T> T getByUuid(String uuid, Class<T> clazz) {
 		return (T) criteria.uniqueResult();
 	}
 	
+	private <T> T getProviderAttributeTypeByName(String name, Class<T> tClass) {
+		Criteria criteria = getSession().createCriteria(tClass);
+		criteria.add(Restrictions.eq("name", name));
```
dkayiwa, line 276: Is there any reason why you added two methods in this class instead of just one?

api/src/main/java/org/openmrs/api/ProviderService.java
```diff
@@ -224,6 +224,16 @@ public List<Provider> getProviders(String query, Integer start, Integer length,
 	 */
 	public ProviderAttributeType getProviderAttributeType(Integer providerAttributeTypeId);
 	
+	/**
+	 * Get a provider attribute type by it's name
+	 *
+	 * @param name the name of the provider attribute type
+	 * @return the provider attribute type for the given uuid
```
dkayiwa, line 231: For the given uuid or name?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateProviderDAO.java
```diff
@@ -279,6 +279,16 @@ public ProviderAttributeType getProviderAttributeType(Integer providerAttributeT
...
 	}
 	
+	/**
+	 * @see org.openmrs.api.db.ProviderDAO#getProviderAttributeTypeByName(String)
+	 */
+	@Override
+	public ProviderAttributeType getProviderAttributeTypeByName(String name) {
+		Criteria criteria = getSession().createCriteria(ProviderAttributeType.class);
+		criteria.add(Restrictions.eq("name", name));
+		return (ProviderAttributeType) criteria.uniqueResult();
```
dkayiwa, line 289: What happens when we have more than one provider attribute type with the same name?

mherman22: @dkayiwa should i throw an exception or is having same names for the providerattributetype an acceptable occurance? or should i Set up a unique constraint on the "name" column of the database table storing the provider attribute types. This would prevent the insertion of duplicate names in the first place.

check out the next commit for my thought process on your question :smiling_face_with_tear:

dkayiwa: What do you think would be the best option?

mherman22: i think what i have done in the latest commit to retrieve a list of ProviderAttributeType objects from the database where the "name" attribute matches the provided name does the trick.

dkayiwa: And what happens at this code when we have more than one returned?

dkayiwa: And what if it is the one at index 1 that the caller wanted?

mherman22: i am blocked i guess! at this level i am thinking of modifying the method in the service class to something like `ProviderAttributeType getProviderAttributeTypeByNameAndIndex(String name, int index)` that uses `getProviderAttributeTypesByName(name)` to get the list and then iterate through with something like     
```
if (index >= 0 && index < attributeTypes.size()) {
        return attributeTypes.get(index);
    } else {
        return null;  // Or throw an exception to indicate an invalid index.
    }
```

dkayiwa: Why don't you do like the rest of the openmrs objects such as person attribute types?

ibacher: This is something we're somewhat inconsistent on. For locations, where we ensure that the attribute type names are unique on creation will actually throw a Hibernate `NonUniqueResultException`, because we use `uniqueResult()`; person attribute types will return result 0 regardless of the number of responses returned. Programs actually throw a `ProgramNameDuplicatedException`.

mherman22: Done!

dkayiwa: Did you also ensure that one is not able to create duplicate provider attribute types as it is for person attribute types?

mherman22: The prevention of duplication is achieved through the use of the `@Column` annotation on the `getName()` method as seen below. I did this because I could not directly make the 'name' column in BaseOpenmrsdata unique since it is extended by other classes that might not need the uniqueness of name

```java
@Column(name = "name", nullable = false, length = 255, unique = true)
@Field
public String getName() {
    return super.getName();
}
```

The `unique = true` attribute within the `@Column` annotation indicates that the values in the `name` column should be unique which means that each value in the `name` column must be distinct, and any attempt to insert or update a row with a duplicate name value will result in a constraint violation.

mherman22:
any last words here? @dkayiwa

api/src/main/java/org/openmrs/api/db/hibernate/HibernateProviderDAO.java
dkayiwa: @mherman22 did you test this and confirm that you are actually not able to create two provider attribute types with the same name?

mherman22: my latest commit addresses your question @dkayiwa

mherman22: well apparently the webapp still allows their creation, i need to figure out why

cc: @dkayiwa

ibacher:
@mherman22 The way we usually enforce this is via the appropriate validator class. [This is how we do it for LocationAttributeTypes](https://github.com/openmrs/openmrs-core/blob/e9aa464256ef33bca82b310813dcbeaf9b5145da/api/src/main/java/org/openmrs/validator/LocationAttributeTypeValidator.java#L46).

Just applying a Hibernate annotation only works in test, because in production, the database is generated from Liquibase scripts, not the Hibernate mappings.

ibacher:
@mherman22 I strongly prefer we don't use a unique index on the `name` field. This is because the rule we usually want to enforce is that there is only one _active_ attribute type with a given name, but if I retire an attribute type, I still want to keep it in the table.

mherman22:
> @mherman22 I strongly prefer we don't use a unique index on the `name` field. This is because the rule we usually want to enforce is that there is only one _active_ attribute type with a given name, but if I retire an attribute type, I still want to keep it in the table.

@ibacher does this mean if i retire an attribute type(e.g `type 2`) then i can create a new attribute with the same name(e.g `type 2` ?

ibacher:
Yes. More concretely, if I have a provider attribute "Phone Number" that's set to free text and I decide to move it to a different datatype (maybe I make a custom data type for phone numbers that reformats them to display a certain way), I still want the attribute to be called "Phone Number", but it's now a new attribute.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.6.x.xml
```diff
@@ -133,5 +133,17 @@
 		<comment>Adding not null constraint to encounter_diagnosis.dx_rank. TRUNK-6081</comment>
 		<addNotNullConstraint tableName="encounter_diagnosis" columnName="dx_rank" columnDataType="int"/>
 	</changeSet>
-	
+
+	<changeSet id="TRUNK-6189-2023-09-06" author="mherman22">
```
ibacher, line 137: I'd still avoid doing this at the database level...

mherman22: Not getting you. So the requirement to make the name unique is no more,right?.

ibacher: No, we want to make sure that the name is unique among non-retired provider attribute types. However, we don't normally do this via unique indexes on the database; we usually just work with the validator to do so, so this DB-level rule is unnecessary, but we should validate the uniqueness of the name in the validator.

mherman22:
closed to continue the work at https://github.com/openmrs/openmrs-core/pull/4403

PR #4399, opened 2023-10-06
TRUNK-6196: Email sending fails due to errors with velocity log conf…

mseaton:
…iguration

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-6196

### DIFF
--- a/api/src/main/java/org/openmrs/notification/mail/velocity/VelocityMessagePreparator.java
+++ b/api/src/main/java/org/openmrs/notification/mail/velocity/VelocityMessagePreparator.java
@@ -10,6 +10,7 @@
 package org.openmrs.notification.mail.velocity;
 
 import java.io.StringWriter;
+import java.util.Properties;
 
 import org.apache.velocity.VelocityContext;
 import org.apache.velocity.app.VelocityEngine;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/notification/mail/velocity/VelocityMessagePreparator.java
```diff
@@ -40,7 +41,11 @@ public class VelocityMessagePreparator implements MessagePreparator {
 	public VelocityMessagePreparator() throws MessageException {
 		try {
 			engine = new VelocityEngine();
-			engine.init();
+			Properties props = new Properties();
+			props.put("runtime.log.logsystem.class", "org.apache.velocity.runtime.log.SimpleLog4JLogSystem");
```
ibacher, line 45: It does look like Velocity recommends the `Log4JLogChute`class over this one.

mseaton: Changed.  Want to have a look?

mogoodrich (approved):
I don't really know much to weigh in too intelligently, but LGTM...

PR #4413, opened 2023-10-17
TRUNK-6187: Protect admin credentials with runtime property

mogoodrich:
add "changePassword" User Service API function back in

## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/UserService.java
+++ b/api/src/main/java/org/openmrs/api/UserService.java
@@ -331,6 +331,18 @@ public interface UserService extends OpenmrsService {
 	 */
 	@Logging(ignoredArgumentIndexes = { 0, 1 })
 	public void changePassword(String oldPassword, String newPassword) throws APIException;
+
+	/**
+	 * Changes password of {@link User} passed in
+	 * @param user user whose password is to be changed
+	 * @param newPassword new password to set
+	 * @throws APIException
+	 * <strong>Should</strong> update password of given user when logged in user has edit users password privilege
+	 * <strong>Should</strong> not update password of given user when logged in user does not have edit users password privilege
+	 */
+	@Deprecated
+	@Authorized({PrivilegeConstants.EDIT_USER_PASSWORDS})
+	public void changePassword(User user, String newPassword) throws APIException;
 	
 	/**
 	 * Changes the current user's password directly. This is most useful if migrating users from
--- a/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
@@ -660,19 +660,12 @@ public class UserServiceImpl extends BaseOpenmrsService implements UserService {
 		updatePassword(user, newPassword);
 	}
 	
-	/**
-	 * This is for internal use only. DO NOT CALL THIS METHOD.
-	 * 
-	 * @param user The user's password to change
-	 * @param newPassword The password to change it to
-	 */
-	@Authorized(PrivilegeConstants.EDIT_USER_PASSWORDS)
+	@Override
 	public void changePassword(User user, String newPassword) {
-		if (!Daemon.isDaemonThread() || !Context.getUserContext().getAuthenticatedUser().isSuperUser()) {
-			throw new APIAuthenticationException(Context.getMessageSourceService().getMessage("error.privilegesRequired",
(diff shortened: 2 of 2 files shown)


api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -331,6 +331,18 @@ public interface UserService extends OpenmrsService {
...
+
+	/**
+	 * Changes password of {@link User} passed in
+	 * @param user user whose password is to be changed
+	 * @param newPassword new password to set
+	 * @throws APIException
+	 * <strong>Should</strong> update password of given user when logged in user has edit users password privilege
+	 * <strong>Should</strong> not update password of given user when logged in user does not have edit users password privilege
+	 */
+	@Deprecated
```
mogoodrich, line 343: I flagged this as deprecated, but we could debate whether we really want to deprecate it.
Also we could potentially decide to deprecate it i the 2.5.x and 2.6.x branches you remove it in the 2.7.x

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -660,19 +660,12 @@ public void changePassword(User user, String oldPassword, String newPassword) th
...
-	/**
-	 * This is for internal use only. DO NOT CALL THIS METHOD.
-	 * 
-	 * @param user The user's password to change
-	 * @param newPassword The password to change it to
-	 */
-	@Authorized(PrivilegeConstants.EDIT_USER_PASSWORDS)
+	@Override
 	public void changePassword(User user, String newPassword) {
-		if (!Daemon.isDaemonThread() || !Context.getUserContext().getAuthenticatedUser().isSuperUser()) {
```
mogoodrich, line 671: I'm not sure why this was added, but from looking at the logic, it would break existing functionality, as it only allows access if the user is a system developer.

From a search, I believe this method was actually being used anywhere, but I could be missing something?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -660,19 +660,12 @@ public void changePassword(User user, String oldPassword, String newPassword) th
...
-			throw new APIAuthenticationException(Context.getMessageSourceService().getMessage("error.privilegesRequired",
-				new Object[] { "System Developer" }, Context.getLocale()));
+		if ("admin".equals(user.getUsername()) && Boolean.valueOf(Context.getRuntimeProperties()
```
mogoodrich, line 665: This check was removed from the updatePassword method, so I added it here to make the changePassword(User,String) method functionality equivalent to what is used to be.

mogoodrich (commented):
This PR intends to add back in the changePassword function with it's original functionality, but flagged as deprecated.

In the previous PR, the method was removed as a service method, and with the impl the implementation was slightly modified.  However, from a code search I don't think implementation was being used anywhere, so changing it seemed fine?

@ibacher  @mseaton

api/src/main/java/org/openmrs/api/UserService.java
dkayiwa: I would fully use the deprecation guidelines here: https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-Deprecation 😊

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
mogoodrich: (I ended up pulling this out because this turned out to be functionality that had only been added as part of this ticket)

api/src/main/java/org/openmrs/api/UserService.java
mogoodrich: Upon further reflection/discussion, I don't think we want to deprecate this method at all.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
mseaton: I agree.  There is already a privilege check around this.  Why is this additional check needed?  If we think most users should have the EDIT_USER_PASSWORDS privileged removed, that's reasonable, but we shouldn't restrict editing users to System Developer role specifically.

mogoodrich:
Thanks @mseaton , merging this in!

PR #4555, opened 2024-02-10
TRUNK-6218: Concept Validations should be more lenient with retired concepts

k4pran:

This PR skips validation of concept names and concept maps when the concept being validated is retired. As part of the test there was a requirement to handle these four scenarios:

- We should not validate any concept names for a concept if the concept is retired
- We should validate concept names for a concept when it is being un-retired
- The logic to check for duplicate mappings should ignore retired mappings
- Un-retiring a retired mapping should still check for duplicate mappings

I left a comment on the jira as it seemed to me that checking if the concept was retired also covered the cases where a concept was being unretired. If I am wrong in my understanding I can update this PR accordingly.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6218

  No? -> execute above command


api/src/main/java/org/openmrs/validator/ConceptValidator.java
```diff
@@ -113,6 +115,9 @@ public void validate(Object obj, Errors errors) throws APIException, DuplicateCo
 			Set<String> validNamesFoundInLocale = new HashSet<>();
 			Collection<ConceptName> namesInLocale = conceptToValidate.getNames(conceptNameLocale);
 			for (ConceptName nameInLocale : namesInLocale) {
+				if (conceptToValidate.getRetired()) {
+                    continue;
+                }
```
ibacher, line 120: It seems like it would be more efficient to move this if check before the outer loop, since if a concept is retired, we don't need to validate any of it's names.

api/src/main/java/org/openmrs/validator/ConceptValidator.java
```diff
@@ -209,6 +214,9 @@ else if (nameInLocale.getLocalePreferred() && preferredNameForLocaleFound) {
 			int index = 0;
 			Set<Integer> mappedTermIds = null;
 			for (ConceptMap map : conceptToValidate.getConceptMappings()) {
+				if (conceptToValidate.getRetired()) {
```
ibacher, line 217: Ditto for this.

ibacher (commented):
Thanks @k4pran! This is very close to perfect. I think the one other thing the ticket asked for was adding some leniency to the `ConceptReferenceTermValidator` so that we also can implement this part: "The logic to check for duplicate mappings should ignore retired mappings".

This may need some additional work because [getConceptReferenceTermByCode()](https://github.com/openmrs/openmrs-core/blob/9c950adeb5788e30a0e9300139050ddcc51edc29/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java#L1750) doesn't ignore retired mappings, which it probably should. (I guess the more sophisticated implementation, which may be better, would return an unretired mapping for a code if one exists and a retired mapping for a code if only a retired mapping for the code exists).

k4pran:
> This may need some additional work because [getConceptReferenceTermByCode()](https://github.com/openmrs/openmrs-core/blob/9c950adeb5788e30a0e9300139050ddcc51edc29/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java#L1750) doesn't ignore retired mappings, which it probably should. (I guess the more sophisticated implementation, which may be better, would return an unretired mapping for a code if one exists and a retired mapping for a code if only a retired mapping for the code exists).

Got it thanks. To clarify, if we want to implement a more sophisticated approach is it safe to do this at the DAO level (HibernateConceptDAO.getConceptReferenceTermByCode) ? Then I would update the method to do this following

1. Get an unretired `ConceptReferenceTerm` if an unretired one exists
2. If no `ConceptReferenceTerm` exists then get a retired one if it exists
3. If neither exists the code doesn't change

Is that correct?

ibacher (approved):
LGTM! Thanks @k4pran!

PR #4559, opened 2024-02-15
TRUNK-6223 - RequiredDataAdvice should validate after save handlers e…

mseaton:
…xecute

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-6223

### DIFF
--- a/api/src/test/java/org/openmrs/api/handler/RequiredReasonVoidSaveHandlerTest.java
+++ b/api/src/test/java/org/openmrs/api/handler/RequiredReasonVoidSaveHandlerTest.java
@@ -9,19 +9,19 @@
  */
 package org.openmrs.api.handler;
 
-import static org.junit.jupiter.api.Assertions.assertThrows;
-
-import java.util.Date;
-
 import org.junit.jupiter.api.Test;
 import org.openmrs.Encounter;
 import org.openmrs.Patient;
-import org.openmrs.Person;
+import org.openmrs.PersonAddress;
 import org.openmrs.User;
 import org.openmrs.Voidable;
 import org.openmrs.api.APIException;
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/openmrs/api/handler/RequiredReasonVoidSaveHandlerTest.java
```diff
@@ -77,10 +77,9 @@ public void handle_shouldNotThrowExceptionIfVoidReasonIsNotBlank() {
...
 	public void handle_shouldNotThrowExceptionIfVoidReasonIsNullForUnsupportedTypes() {
-		Person p = Context.getPersonService().getPerson(1);
-		p.setVoided(true);
-		p.setVoidReason(null);
-		p.setVoidReason("voidReason");
-		Context.getPersonService().savePerson(p);
+		PersonAddress pa = Context.getPersonService().getPersonAddressByUuid("3350d0b5-821c-4e5e-ad1d-a9bce331e118");
+		pa.setVoided(true);
+		pa.setVoidReason(null);
+		Context.getPersonService().savePersonAddress(pa);
```
mseaton, line 83: This test as written didn't make a lot of sense.  Not sure y it is setting void reason twice.  In any event, the PersonValidator has an explicit check to ensure the void reason is non-null, and based on the name of this test method, the idea is to ensure you can leave void reason null for types other than Patient and Encounter, so I changed this to PersonAddress and things work fine.

mogoodrich (approved):
LGTM, though see my comments on the ticket.

dkayiwa (approved):
I am currently not aware of any problems that may result from this change. So looks good to me.

PR #4581, opened 2024-03-11
TRUNK-5952: Order validation is bypassed if certain fields are set from the OrderContext.

IamMujuziMoses:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5952

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
@@ -46,6 +46,7 @@ import org.openmrs.api.OrderNumberGenerator;
 import org.openmrs.api.OrderService;
 import org.openmrs.api.RefByUuid;
 import org.openmrs.api.UnchangeableObjectException;
+import org.openmrs.api.ValidationException;
 import org.openmrs.api.context.Context;
 import org.openmrs.api.db.OrderDAO;
 import org.openmrs.customdatatype.CustomDatatypeUtil;
@@ -61,6 +62,9 @@ import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.util.StringUtils;
+import org.springframework.validation.BeanPropertyBindingResult;
+import org.springframework.validation.Errors;
+import org.springframework.validation.Validator;
 
 import java.lang.reflect.Field;
 import java.util.ArrayList;
@@ -169,7 +173,6 @@ public class OrderServiceImpl extends BaseOpenmrsService implements OrderService
 		ensureOrderTypeIsSet(order,orderContext);
 		ensureCareSettingIsSet(order,orderContext);
 		failOnOrderTypeMismatch(order);
-		validateFieldsForOrderContextOutPatientCareSettingType(order, orderContext);
 		
 		// If isRetrospective is false, but the dateActivated is prior to the current date, set isRetrospective to true
 		if (!isRetrospective) {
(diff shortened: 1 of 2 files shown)


IamMujuziMoses:
@mseaton  @dkayiwa @ibacher could you please review this PR

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -292,6 +293,28 @@ private void ensureCareSettingIsSet(Order order, OrderContext orderContext) {
 		order.setCareSetting(careSetting);
 	}
 
+	private void validateFieldsForOrderContextOutPatientCareSettingType(Order order, OrderContext orderContext) {
```
mseaton: I don't really like that this duplicates validation logic already happening elsewhere (and would need to be kept consistent with it), or that this is limited to this very specific issue that was discovered in the ticket.

My suggestion in the ticket was to try invoking the validator in the saveOrderInternal method, before saving with the DAO.  Would be good to know if this was attempted or not and if it caused problems, or otherwise why that approach was not taken?

mseaton (changes requested):
See comment

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -362,6 +366,21 @@ private Order saveOrderInternal(Order order, OrderContext orderContext) {
 				}
 			}
 		}
+
+		if (orderContext != null) {
```
mseaton, line 370: I don't think we necessarily need this restriction.

mseaton (approved):
This looks ok to me, though see my comment.  @dkayiwa thoughts?

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
IamMujuziMoses: Running validation unconditionally breaks existing behaviour/tests that never received post-validation before. Therefore we need to guard with orderContext != null so that this new validation only applies to the code paths where an OrderContext is actually provided and expected

mseaton: What tests does it break?  Do the tests seem correct and do the new failures look like they are exposing additional bugs?  Can you be more specific?

IamMujuziMoses: The tests seem correct:
```stacktrace
org.opentest4j.AssertionFailedError: expected: java.util.Date@4c92384f<Mon Nov 10 14:20:07 EAT 2025> but was: java.util.Date@543e126<Mon Nov 10 14:20:07 EAT 2025>
Expected :Mon Nov 10 14:20:07 EAT 2025
Actual   :Mon Nov 10 14:20:07 EAT 2025
```
Just  a classic Java Date object equality vs. value equality issue in tests that happens even though the two Date objects print exactly the same

mseaton: Are you sure?  `Sun Nov 09 15:16:07 EAT 2025` does not look the same as 1`Mon Nov 10 14:20:07 EAT 2025`,  Not sure what the "Expected" and "Actual" methods look like that.  But the assertion failure message shows differences.  Point is - if the test is failing in a situation where it should not be failing, let's figure out why and fix it and remove that unnecessary constraint on orderContext != null.

IamMujuziMoses: I think that the ```orderContext != null``` check scopes the fix to the path where this inconsistency actually occurs, saveOrder(Order, OrderContext). without it, we’d start running post-validation on legacy calls that never had it, causing unrelated test failures and changing existing behavior, this keeps the fix focused on the API scenario described in the ticket, preventiing Order validation from being bypassed if certain fields are set from the OrderContext.

PR #4585, opened 2024-03-13
TRUNK-5964: Hard-coded uuids for drug order type and test order type.

IamMujuziMoses:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5964

  No? -> execute above command


ojwanganto:
@IamMujuziMoses - thanks for this work. How different is this approach from the current/existing implementation? What gaps are we addressing?

IamMujuziMoses:
@ojwanganto thanks for the feedback, this PR is intended to implement a better and more correct approach to remove any dependency on the specific hard-coded uuids for drug order type and test order type as stated in this [ticket](https://openmrs.atlassian.net/browse/TRUNK-5964)

IamMujuziMoses:
@mseaton @dkayiwa @ibacher could you please review this PR, your feedback will be appreciated

ojwanganto:
@IamMujuziMoses got it. I have also checked the linked ticket. Thanks

agrpranjal07 (commented):
**Issue addressed by PR:**
This PR addresses the issue of hard-coded UUIDs for "Drug Order" and "Test Order" types in the OpenMRS API service.

**Key Changes:**
- Added getOrderTypesByClassName(String javaClassName) method in OrderService, OrderDAO, and HibernateOrderDAO
- Replaced hard-coded UUID lookups with javaClassName lookups for "Drug Order" and "Test Order" in OrderServiceImpl
- Added tests to check whether the method works properly.

**Suggestions:**
Consider adding more test cases to cover edge cases, such as empty strings or invalid class names.

api/src/main/java/org/openmrs/api/OrderService.java
```diff
@@ -724,6 +724,17 @@ public Order discontinueOrder(Order orderToDiscontinue, String reasonNonCoded, D
...
+	/**
+	 * Get order types by java class name
+	 *
+	 * @param javaClassName the class name used to get the order types
+	 * @since 2.7.0
+	 * @return return the order types associated with given class name
+	 * <strong>Should</strong> find order types with the specified class name
+	 */
+	@Authorized(PrivilegeConstants.GET_ORDER_TYPES)
+	public List<OrderType> getOrderTypesByClassName(String javaClassName) throws APIException;
```
mseaton: These service and DAO methods seem like a good addition.  We will need to update the since annotation to reflect the right versions at this point.  However, I think we should also consider how these methods support inheritance (or don't currently).  For example, maybe I have an Order class in my module that extends TestOrder.  If I call this method with "org.openmrs.TestOrder", should that return just OrderTypes that have a class of "org.openmrs.TestOrder", or should it also return OrderTypes that have a class of something that extends "org.openmrs.TestOrder"?  Do we want to add that as a possible option via an argument?

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -315,7 +315,7 @@ private boolean areDrugOrdersOfSameOrderableAndOverlappingSchedule(Order firstOr
 		        && !OpenmrsUtil.nullSafeEquals(firstOrder.getPreviousOrder(), secondOrder)
 		        && OrderUtil.checkScheduleOverlap(firstOrder, secondOrder)
 		        && firstOrder.getOrderType().equals(
-		            Context.getOrderService().getOrderTypeByUuid(OrderType.DRUG_ORDER_TYPE_UUID));
+		            Context.getOrderService().getOrderTypeByName(OpenmrsConstants.DRUG_ORDER_TYPE_NAME));
```
mseaton: I don't understand why we are making this change.  We still want to look up by these uuids - and if they are found, continue as before.  But if not found, we should try to find a unique match by class.  So if orderType is null by looking up by DRUG_ORDER_TYPE_UUID, then use the new methods you have added to look up OrderType based on "org.openmrs.DrugOrder" class, and if only one is found, then return it.  The "name" of the order type is not something we should be looking up.

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -233,6 +233,10 @@ public static final Collection<String> AUTO_ROLES() {
 		return roles;
 	}
 	
+	public static final String DRUG_ORDER_TYPE_NAME = "Drug order";
+	
+	public static final String TEST_ORDER_TYPE_NAME = "Test order";
+
```
mseaton: These should be removed.

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -270,10 +270,10 @@ private void ensureOrderTypeIsSet(Order order, OrderContext orderContext) {
...
 		if (orderType == null && order instanceof DrugOrder) {
-			orderType = Context.getOrderService().getOrderTypeByUuid(OrderType.DRUG_ORDER_TYPE_UUID);
+			orderType = Context.getOrderService().getOrderTypeByName(OpenmrsConstants.DRUG_ORDER_TYPE_NAME);
 		}
 		if (orderType == null && order instanceof TestOrder) {
-			orderType = Context.getOrderService().getOrderTypeByUuid(OrderType.TEST_ORDER_TYPE_UUID);
+			orderType = Context.getOrderService().getOrderTypeByName(OpenmrsConstants.TEST_ORDER_TYPE_NAME);
 		}
 		if (orderType == null && order instanceof ReferralOrder) {
 			orderType = Context.getOrderService().getOrderTypeByUuid(OrderType.REFERRAL_ORDER_TYPE_UUID);
```
wikumChamith: Shouldn’t these be replaced like this for the purpose of this ticket?

```
		if (orderType == null) {
			orderType = Context.getOrderService().getOrderTypeByName(order.getClass().getName());
		}
```

api/src/main/java/org/openmrs/api/db/hibernate/HibernateOrderDAO.java
```diff
@@ -778,6 +778,64 @@ public OrderType getOrderTypeByConceptClass(ConceptClass conceptClass) {
...
+		cq.where(cb.equal(root.get("javaClassName"), javaClassName));
+		
+		return session.createQuery(cq).getResultList();
+		
+	}
+
+	/**
+	 * @see org.openmrs.api.OrderService#getOrderTypesByClassName(String, boolean)
+	 */
+	@Override
```
mseaton: Based on what I see here, this does not need to be a DAO method, it should just be a service method, where you either call the dao.getOrderTypesByClassName, or you call "dao.getAllOrderTypes" and do the classname comparisons in Java like you are doing.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateOrderDAO.java
```diff
@@ -778,6 +778,64 @@ public OrderType getOrderTypeByConceptClass(ConceptClass conceptClass) {
...
+			return getOrderTypesByClassName(javaClassName);
+		}
+		
+		if (StringUtils.isBlank(javaClassName)) {
+			throw new APIException("javaClassName cannot be null");
+		}
+
+		Class<?> superClass;
+		try {
+			superClass = Class.forName(javaClassName);
```
mseaton: This should use Context.loadClass(javaClassName)

api/src/main/java/org/openmrs/api/db/hibernate/HibernateOrderDAO.java
```diff
@@ -778,6 +778,64 @@ public OrderType getOrderTypeByConceptClass(ConceptClass conceptClass) {
...
+		} catch (ClassNotFoundException e) {
+			throw new APIException("Invalid javaClassName: " + javaClassName, e);
+		}
+
+		Session session = sessionFactory.getCurrentSession();
+		CriteriaBuilder cb = session.getCriteriaBuilder();
+		CriteriaQuery<OrderType> cq = cb.createQuery(OrderType.class);
+		Root<OrderType> root = cq.from(OrderType.class);
+		
+		List<OrderType> all = session.createQuery(cq.select(root)).getResultList();
```
mseaton: Per my comment above, there is already a DAO method to getOrderTypes, you should just use that.  You either need to also take in the "includeRetired" parameter, or always exclude retired I would think.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateOrderDAO.java
```diff
@@ -778,6 +778,64 @@ public OrderType getOrderTypeByConceptClass(ConceptClass conceptClass) {
...
+
+		return all.stream()
+			.filter(ot -> {
+				try {
+					Class<?> c = Class.forName(ot.getJavaClassName());
```
mseaton: This should also use Context.loadClass

api/src/main/java/org/openmrs/api/db/OrderDAO.java
```diff
@@ -202,6 +202,16 @@ public List<OrderFrequency> getOrderFrequencies(String searchPhrase, Locale loca
...
+
+	/**
+	 * @see org.openmrs.api.OrderService#getOrderTypesByClassName(String)
+	 */
+	public List<OrderType> getOrderTypesByClassName(String javaClassName) throws DAOException;
+
+	/**
+	 * @see org.openmrs.api.OrderService#getOrderTypesByClassName(String, boolean)
+	 */
+	public List<OrderType> getOrderTypesByClassName(String javaClassName, boolean includeSubclasses) throws DAOException;
```
mseaton: Per above, this method should be removed from DAO and only needed in service, as no additional database queries are needed beyond what the DAO can already do

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -270,10 +270,12 @@ private void ensureOrderTypeIsSet(Order order, OrderContext orderContext) {
 			orderType = getOrderTypeByConcept(order.getConcept());
 		}
 		if (orderType == null && order instanceof DrugOrder) {
-			orderType = Context.getOrderService().getOrderTypeByUuid(OrderType.DRUG_ORDER_TYPE_UUID);
```
mseaton, line 273: I think it is fine to keep this original behavior, and to look up by this UUID, particularly since this will match many existing implementation and is backwards-compatible.  We just should not _assume_ this will always be the uuid for this order type

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -270,10 +270,12 @@ private void ensureOrderTypeIsSet(Order order, OrderContext orderContext) {
...
+			List<OrderType> types = Context.getOrderService().getOrderTypesByClassName(DrugOrder.class.getName(), true);
```
mseaton: No need for `Context.getOrderService()` here - just call `getOrderTypesByClassName` directly.  We only need to use that full method when we are needing to ensure AOP interceptors are hit, which we don't in this case.

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -270,10 +270,12 @@ private void ensureOrderTypeIsSet(Order order, OrderContext orderContext) {
...
+			orderType = !types.isEmpty() ? types.get(0) : null;
 		}
 		if (orderType == null && order instanceof TestOrder) {
-			orderType = Context.getOrderService().getOrderTypeByUuid(OrderType.TEST_ORDER_TYPE_UUID);
+			List<OrderType> types = Context.getOrderService().getOrderTypesByClassName(TestOrder.class.getName(), true);
```
mseaton: Same comments here as above

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
mseaton: Agree with @wikumChamith - we need to do the same with this block for ReferralOrder as we do above with the others.  ReferralOrder didn't exist when this ticket was originally written

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -312,10 +314,20 @@ private void failOnOrderTypeMismatch(Order order) {
...
-		            Context.getOrderService().getOrderTypeByUuid(OrderType.DRUG_ORDER_TYPE_UUID));
+			&& !OpenmrsUtil.nullSafeEquals(firstOrder.getPreviousOrder(), secondOrder)
+			&& OrderUtil.checkScheduleOverlap(firstOrder, secondOrder)
+			&& firstOrder.getOrderType().equals(getDefaultDrugOrderType());
+	}
+
+	private OrderType getDefaultDrugOrderType() {
+		OrderType type = Context.getOrderService().getOrderTypeByUuid(OrderType.DRUG_ORDER_TYPE_UUID);
+		if (type == null) {
+			List<OrderType> types = Context.getOrderService().getOrderTypesByClassName(DrugOrder.class.getName(), true);
```
mseaton: Same comment here, no need for the `Context.getOrderService()` part.  Why aren't you using this method in the above `ensureOrderTypeIsSet` method?  The logic should match, correct?  And you can just try to retrieve the DrugOrder type using this method

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -1061,6 +1073,24 @@ public OrderType getOrderTypeByConceptClass(ConceptClass conceptClass) {
...
+		return dao.getOrderTypesByClassName(javaClassName);
+	}
+	
+	/**
+	 * @see org.openmrs.api.OrderService#getOrderTypesByClassName(String, boolean)
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	public List<OrderType> getOrderTypesByClassName(String javaClassName, boolean includeSubclasses) throws APIException {
+		return dao.getOrderTypesByClassName(javaClassName, includeSubclasses);
```
mseaton: Per my comments above, logic in the DAO should just be moved here.

IamMujuziMoses: right

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -1061,6 +1072,47 @@ public OrderType getOrderTypeByConceptClass(ConceptClass conceptClass) {
...
+	}
+	
+	/**
+	 * @see org.openmrs.api.OrderService#getOrderTypesByClassName(String, boolean, boolean)
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	public List<OrderType> getOrderTypesByClassName(String javaClassName, boolean includeSubclasses, boolean includeRetired) throws APIException {
+		if (!includeSubclasses) {
+			return getOrderTypesByClassName(javaClassName);
```
mseaton: This doesn't use the "includeRetired" parameter, which would be incorrect in this case, right?  Maybe we need to include the includeRetired parameter in the getOrderTypesByClassName method too?

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -1061,6 +1072,47 @@ public OrderType getOrderTypeByConceptClass(ConceptClass conceptClass) {
...
+		}
+
+		if (!StringUtils.hasText(javaClassName)) {
```
mseaton, line 1091: Shouldn't this check come before the includeSubclasses check?

IamMujuziMoses: getOrderTypesByClassName(javaClassName) has it's own check but yeah, makes sense

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
IamMujuziMoses: right

PR #4586, opened 2024-03-13
TRUNK-6192: Fix bug related to continuous listener additions

Ruhanga:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6192

### DIFF
--- a/api/src/main/java/org/openmrs/layout/name/NameSupport.java
+++ b/api/src/main/java/org/openmrs/layout/name/NameSupport.java
@@ -15,6 +15,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import javax.annotation.PostConstruct;
+
 import org.openmrs.GlobalProperty;
 import org.openmrs.api.APIException;
 import org.openmrs.api.GlobalPropertyListener;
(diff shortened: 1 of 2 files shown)


mogoodrich (changes requested):
Instead of this hack, I think all we need to do it move line 67 (initialized = true) outside of the if/then block, to confirm that the init method is only run once.  The "globalPropertyChanged" method should handle updated in the template if the global property changes.

Ruhanga:
Sorry I missed the notification here @mogoodrich. The challenge in that approach is, since the `NameSupport.java` is a Spring bean that's instantiated once, it would then miss an initialisation involving the custom configured template if one existed. This mean that the custom configured template would only come into effect only after its global property has been added/updated which, 99% won't happen. 

An alternative approach was to make the list of listeners a Set.

mogoodrich:
Interesting @Ruhanga , you mean because the Name Support bean could be initialized before the Admin Service is ready?  What makes this different from the AddressSupport template that appears to do the same thing?

Ruhanga:
The AddressSupport template is not a Spring bean actually. It gets instantiated if not at the first call to getInstance().

ibacher:
@Ruhanga If the issue is that it's a Spring bean, couldn't we take advantage of that to have it call `init()` as either a Spring initializer or `@PostConstruct` or similar and skip having to call `init()` in every call to `getInstance()`?

Ruhanga (commented):
Hi @ibacher, sorry I got overtaken by some other tasks and forgot to respond to your comment above. Hope we are good now...

Ruhanga:
Closing this in favour of this commit here:

https://github.com/openmrs/openmrs-core/commit/e878c6389843cb052b4d158afac2582112312939

PR #4650, opened 2024-05-27
TRUNK-6235: auto deactivate users after XX days of inactivity

dicksonmulli:

- Creating a task for auto retiring users after XX days of inactivity.
- Setting number-of-days-to-retire-users property in global properties.
- Adding last login time in user properties

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6235

  No? -> execute above command


Dockerfile
```diff
@@ -95,7 +95,7 @@ ENTRYPOINT ["/usr/bin/tini", "--", "/usr/local/bin/mvn-entrypoint.sh"]
 CMD ["/openmrs/startup-dev.sh"]
 
 ### Production Stage
-FROM tomcat:8.5-$RUNTIME_JDK
```
dkayiwa, line 98: Did you make the change in this file accidentally?

api/pom.xml
```diff
@@ -296,6 +296,12 @@
 			<groupId>org.junit.jupiter</groupId>
 			<artifactId>junit-jupiter-params</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.assertj</groupId>
+			<artifactId>assertj-core</artifactId>
```
dkayiwa, line 301: Was this required?

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -578,4 +578,10 @@ public List<User> getUsers(String name, List<Role> roles, boolean includeRetired
 	 * @since 2.3.6, 2.4.6, 2.5.4, 2.6.0
 	 */
 	Locale getDefaultLocaleForUser(User user);
+
+	/**
+	 * @param user the user to retrieve last-login-time
+	 * @return timestamp representing last login time
+	 */
+	String getLastLoginTime(User user);
```
dkayiwa, line 586: Did you get a chance to look at this? https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#JavaConventions-NewPublicMethodsandClasses

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -93,126 +93,135 @@ public void setUserDAO(UserDAO userDao) {
 	@Override
 	@Transactional(noRollbackFor = ContextAuthenticationException.class)
 	public User authenticate(String login, String password) throws ContextAuthenticationException {
-		String errorMsg = "Invalid username and/or password: " + login;
+		if (StringUtils.isBlank(login) || password == null) {
```
dkayiwa, line 96: I would not do this validation as part of this ticket. It would be safer to limit its scope to the ticket requirements. Any other improvements of the code base can have separate tickets and pull requests. 
Did you get a chance to look at this? https://opensource.com/article/18/6/anatomy-perfect-pull-request

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -93,126 +93,135 @@ public void setUserDAO(UserDAO userDao) {
...
+		candidateUser.setUserProperty(
+			OpenmrsConstants.USER_PROPERTY_LAST_LOGIN_TIMESTAMP,
+			String.valueOf(System.currentTimeMillis())
+		);
+	}
 
-				// skip out of the method early (instead of throwing the exception)
-				// to indicate that this is the valid user
-				return candidateUser;
-			} else {
```
dkayiwa, line 183: The changes in this file are way much more than i would expect. Can we limit them to just the ticket requirements?

api/src/main/resources/org/openmrs/liquibase/snapshots/core-data/liquibase-core-data-2.6.x.xml
```diff
@@ -5218,4 +5218,23 @@
       <column name="role" value="System Developer"/> 
     </insert> 
   </changeSet>
+	<changeSet author="isaiah (generated)" id="202405201900-1">
+		<insert tableName="scheduler_task_config">
+			<column name="task_config_id" valueNumeric="3"/>
```
dkayiwa, line 5223: Instead of automatically creating this task, i would prefer some sort of documentation for how to create the task via the user interface such that its gets configured by whoever needs this functionality.

pom.xml
```diff
@@ -660,7 +660,7 @@
 				<plugin>
 					<groupId>org.codehaus.mojo</groupId>
 					<artifactId>build-helper-maven-plugin</artifactId>
-					<version>3.5.0</version>
```
dkayiwa, line 663: The changes in this file do not look related to the ticket.

api/pom.xml
dicksonmulli: Yes, this is required in unit tests

Dockerfile
dicksonmulli: This change occurred automatically/accidentally. I will revert it

pom.xml
dicksonmulli: I will clean this up. Some changes were made automatically.

api/src/main/resources/org/openmrs/liquibase/snapshots/core-data/liquibase-core-data-2.6.x.xml
dicksonmulli: Could you suggest where I can add the documentation.

api/pom.xml
dkayiwa: What errors do you get when you do not have it?

api/src/main/resources/org/openmrs/liquibase/snapshots/core-data/liquibase-core-data-2.6.x.xml
dkayiwa: You can start by including it in this list: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25520634/Administering+Scheduled+Tasks\
And then provide a link to it that directs one to a page with its details.

api/pom.xml
dicksonmulli: org.assertj gives more options in assertions. I have used it in the new unit tests.

Dockerfile
dicksonmulli: @dkayiwa  This change was done in a different PR

dkayiwa: So let us remove it from this pull request.

api/pom.xml
dkayiwa: I would avoid adding new libraries at this stage.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
dicksonmulli: Reverted

api/src/main/resources/org/openmrs/liquibase/snapshots/core-data/liquibase-core-data-2.6.x.xml
dicksonmulli: Done

pom.xml
dicksonmulli: Reverted

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -108,8 +108,8 @@ public User authenticate(String login, String password) throws ContextAuthentica
 
 			try {
 				candidateUser = session.createQuery(
-					"from User u where (u.username = ?1 or u.systemId = ?2 or u.systemId = ?3) and u.retired = false",
-					User.class)
+						"from User u where (u.username = ?1 or u.systemId = ?2 or u.systemId = ?3) and u.retired = false",
```
dkayiwa, line 111: Can we also avoid formatting changes?

pom.xml
```diff
@@ -660,7 +660,7 @@
 				<plugin>
 					<groupId>org.codehaus.mojo</groupId>
 					<artifactId>build-helper-maven-plugin</artifactId>
-					<version>3.6.0</version>
```
dkayiwa, line 663: This change does not look related to the ticket.

api/pom.xml
dkayiwa: Let us also remove this change.

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -578,4 +578,11 @@ public List<User> getUsers(String name, List<Role> roles, boolean includeRetired
 	 * @since 2.3.6, 2.4.6, 2.5.4, 2.6.0
 	 */
 	Locale getDefaultLocaleForUser(User user);
+
+	/**
```
dkayiwa, line 582: Can we also add the method description?

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -578,4 +578,11 @@ public List<User> getUsers(String name, List<Role> roles, boolean includeRetired
...
+	 * @param user the user to retrieve last-login-time
+	 * @return timestamp representing last login time
```
dkayiwa, line 584: As part of the return description, can we also include an example value to show the timestamp format?

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -635,6 +634,11 @@ public static final Collection<String> AUTO_ROLES() {
 	 * Global property that stores the base url for the password reset.
 	 */
 	public static final String GP_PASSWORD_RESET_URL = "security.passwordResetUrl";
+
+	/**
+	 * Global property that stores the number of days for users to be deactivated.
+	 */
+	public static final String GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS = "users.numberOfDaysToRetire";
```
dkayiwa, line 641: Can we also add this to CORE_GLOBAL_PROPERTIES?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -710,4 +712,12 @@ private Query createUserSearchQuery(String name, List<Role> roles, boolean inclu
 	public void setUserActivationKey(LoginCredential credentials) {		
 			sessionFactory.getCurrentSession().merge(credentials);	
 	}
+
+	/**
+	 * @see
```
dkayiwa, line 717: Can we finish the above?

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1693,6 +1704,40 @@ public void getDefaultLocaleForUser_shouldReturnDefaultLocaleForUserIfAlreadySet
...
+	@Test
+	public void getLastLoginTimeForUser_shouldReturnLastLoginTimeStampIfPropertyIsSet() {
+		executeDataSet(XML_FILENAME);
+		User createdUser = createTestUser();
+		Context.authenticate(new UsernamePasswordCredentials("bwolfe", "Openmr5xy"));
+		
+		assertThat(createdUser.getUserProperty(OpenmrsConstants.USER_PROPERTY_LAST_LOGIN_TIMESTAMP))
+			.isNotNull()
+			.isNotEmpty();
+		assertThat(Long.parseLong(Context.getUserService().getLastLoginTime(createdUser)))
```
dkayiwa, line 1733: If this test is all about ensuring that the last login timestamp is set, as per the test name, shouldn't it all be about simply asserting that the return value is not empty?

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1693,6 +1704,40 @@ public void getDefaultLocaleForUser_shouldReturnDefaultLocaleForUserIfAlreadySet
...
+		assertEquals("", Context.getUserService().getLastLoginTime(createdUser));
+	}
+
+	@Test
+	public void getLastLoginTimeForUser_shouldReturnEmptyStringOnLastLoginTimeIfADifferentUserIsLoggedIn() {
+		executeDataSet(XML_FILENAME);
+		User createdUser = createTestUser();
+		Context.authenticate(getTestUserCredentials());
+		
+		assertEquals("", createdUser.getUserProperty(OpenmrsConstants.USER_PROPERTY_LAST_LOGIN_TIMESTAMP));
```
dkayiwa, line 1720: Wouldn't it be better to use StringUtils.isBlank?

dkayiwa:
Can we also look into the build failures?

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1693,6 +1704,40 @@ public void getDefaultLocaleForUser_shouldReturnDefaultLocaleForUserIfAlreadySet
 		assertEquals(Locale.FRENCH, locale);
 	}
 
+	@Test
+	public void getLastLoginTimeForUser_shouldReturnEmptyStringOnLastLoginTimeIfPropertyNotSet() {
+		User createdUser = createTestUser();
+		assertEquals("", createdUser.getUserProperty(OpenmrsConstants.USER_PROPERTY_LAST_LOGIN_TIMESTAMP));
```
dkayiwa, line 1710: Can we use StringUtils.isBlank

dkayiwa:
@dicksonmulli can you look into the build failures.

pom.xml
dicksonmulli: It's not related to this ticket, it was changed by a different branch.

dkayiwa:
@dicksonmulli have you also tested these changes by running the webapp, create some users, and ensure that their accounts are deactivated as expected?

dkayiwa:
We also seem to have some build failures.

api/pom.xml
dicksonmulli: Done

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
dicksonmulli: Done

dkayiwa:
@dicksonmulli can you also fix the ticket link on the pull request as advised at? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,81 @@
...
+		if (!isExecuting) {
+			log.debug("Starting auto-retiring users");
+			
+			startExecuting();
+			
+			try {
+				UserService userService = Context.getUserService();
+				Set<User> usersToRetire = getUsersToRetire(userService);
+				
+				usersToRetire.forEach(user -> userService.retireUser(user, "Retire reason"));
```
dkayiwa, line 51: Can we make the retire reason more meaningful?

api/src/main/java/org/openmrs/util/DateUtil.java
```diff
@@ -31,4 +31,12 @@ public static Date truncateToSeconds(Date date) {
 		return Date.from(instant);
 	}
 	
+	/**
+	 * This method converts days to milliseconds
+	 * @param days No. of days
+	 * @return
```
dkayiwa, line 37: Can you also document the return value?

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1662,7 +1664,14 @@ public void saveUserProperty_shouldAddANewPropertyWithAVeryLargeStringWithoutRun
 		final String USER_PROPERTY_KEY = liquibase.util.StringUtil.repeat("emrapi.lastViewedPatientIds,",10);
 		final String USER_PROPERTY_VALUE = liquibase.util.StringUtil.repeat("52345",9899);
 		User updatedUser = userService.saveUserProperty(USER_PROPERTY_KEY, USER_PROPERTY_VALUE);
-		assertEquals(280, updatedUser.getUserProperties().keySet().iterator().next().length());
+
+		Set<String> emrApiPropertyKeys = updatedUser.getUserProperties()
+			.keySet()
+			.stream()
+			.filter(key -> key.contains("emrapi.lastViewedPatientIds"))
```
dkayiwa, line 1671: Does this make it depended on the `emrapi` module?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,81 @@
...
+			} finally {
+				log.debug("Auto-retiring users task ended");
+				stopExecuting();
+			}
+		}
+	}
+
+	private Set<User> getUsersToRetire(UserService userService) {
+		List<User> allUsers = userService.getAllUsers();
+		String numberOfDaysToRetire = Context.getAdministrationService().getGlobalProperty(GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS);
```
dkayiwa, line 63: What happens when this global property is not yet set or has an invalid (non numeric) value?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,81 @@
...
+		long numberOfMillisecondsToRetire = DateUtil.daysToMilliseconds(Integer.parseInt(numberOfDaysToRetire));
+
+		return allUsers.stream()
+			.filter(user -> !user.isRetired() && userInactivityExceedsDaysToRetire(user, numberOfMillisecondsToRetire))
+			.collect(Collectors.toSet());
+	}
+
+	private boolean userInactivityExceedsDaysToRetire(User user, long numberOfMillisecondsToRetire) {
```
dkayiwa, line 71: Can we add some tests for this?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,81 @@
...
+			} catch (Exception e) {
+				log.error("Error occurred while auto-retiring users: ", e);
+			} finally {
+				log.debug("Auto-retiring users task ended");
+				stopExecuting();
+			}
+		}
+	}
+
+	private Set<User> getUsersToRetire(UserService userService) {
```
dkayiwa, line 61: Can we also add some tests for this?

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1326,7 +1328,7 @@ public void saveUserProperty_shouldAddNewPropertyToExistingUserProperties() {
 		Context.authenticate(user.getUsername(), "testUser1234");
 		
 		final int numberOfUserProperties = user.getUserProperties().size();
-		assertEquals(2, user.getUserProperties().size());
+		assertEquals(3, user.getUserProperties().size());
```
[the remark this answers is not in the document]
dicksonmulli: `lastLoginTimestamp` property is added once the user is authenticated.

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
dicksonmulli: I have added a null check incase the value is not set. Tested to be working well.

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.Set;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.not;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class AutoRetireUsersTaskTest extends BaseContextSensitiveTest {
+	protected static final String XML_FILENAME = "org/openmrs/api/include/UserServiceTest.xml";
```
dkayiwa, line 31: Is the any reason why the above is not private?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,130 @@
...
+		User inactiveUser = userService.getUser(1);
+		inactiveUser.setUserProperty(
+			OpenmrsConstants.USER_PROPERTY_LAST_LOGIN_TIMESTAMP, 
+			String.valueOf(System.currentTimeMillis() - (ONE_DAY_IN_MILLISECONDS))
+		); // 1 day ago
+
+		userService.saveUser(inactiveUser);
+
+		Set<User> usersToRetire = autoRetireUsersTask.getUsersToRetire(userService);
+
```
dkayiwa, line 78: Isn't it clearer to just check that the list is empty?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,130 @@
...
+		); // 2 days ago
+		
+		retiredUser.setRetired(true);
+		retiredUser.setRetireReason("User retired due to inactivity");
+
+		userService.saveUser(retiredUser);
+
+		Set<User> usersToRetire = autoRetireUsersTask.getUsersToRetire(userService);
+
+		assertThat(usersToRetire, not(contains(retiredUser)));
```
dkayiwa, line 100: Is the list of usersToRetire expected to be non empty?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dicksonmulli: I don't think this is a good idea because it will make the test brittle. For instance, if another is added (in XML_FILENAME) then this test will fail because the list is not empty.

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dicksonmulli: If another is added (in XML_FILENAME) then this test will fail because the list is not empty.

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dicksonmulli: No particular reason, let me change it.

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dkayiwa: Very very good point! 👍

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dkayiwa: Your reasoning makes a lot of sense! 👍

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -578,4 +578,13 @@ public List<User> getUsers(String name, List<Role> roles, boolean includeRetired
 	 * @since 2.3.6, 2.4.6, 2.5.4, 2.6.0
 	 */
 	Locale getDefaultLocaleForUser(User user);
+
+	/**
+	 * Retrieves the last login time of the user
+	 * 
+	 * @param user the user to retrieve last-login-time
```
dkayiwa, line 585: Can we improve the above by removing the hyphen and correcting the grammar?

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -578,4 +578,13 @@ public List<User> getUsers(String name, List<Role> roles, boolean includeRetired
 	 * @since 2.3.6, 2.4.6, 2.5.4, 2.6.0
 	 */
 	Locale getDefaultLocaleForUser(User user);
+
+	/**
+	 * Retrieves the last login time of the user
```
dkayiwa, line 583: Is the retrieved last login time in milliseconds?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,88 @@
...
+
+import static org.openmrs.util.OpenmrsConstants.GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS;
+
+/**
+ * A scheduled task that automatically retires user after the set number-of-days of inactivity. 
+ * The inactivity duration is set as a global property 
+ * <a href="https://openmrs.atlassian.net/wiki/spaces/docs/pages/25520634/Administering+Scheduled+Tasks">Documentation</a>
+ * {@link OpenmrsConstants#GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS}
+ */
+public class AutoRetireUsersTask extends AbstractTask {
```
dkayiwa, line 34: Can we put a since annotation on this class?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,88 @@
...
+import org.openmrs.util.OpenmrsConstants;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.openmrs.util.OpenmrsConstants.GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS;
```
dkayiwa, line 26: Can we remove this static import?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * A scheduled task that automatically retires user after the set number-of-days of inactivity.
```
dkayiwa, line 29: that automatically retires `users`

dkayiwa: afte the set `number of days` instead of `number-of-days`

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.openmrs.util.OpenmrsConstants.GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS;
+
+/**
+ * A scheduled task that automatically retires user after the set number-of-days of inactivity. 
+ * The inactivity duration is set as a global property 
+ * <a href="https://openmrs.atlassian.net/wiki/spaces/docs/pages/25520634/Administering+Scheduled+Tasks">Documentation</a>
```
dkayiwa, line 31: Shouldn't the link be the one that has the details of setting up this specific task?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,88 @@
...
+ * {@link OpenmrsConstants#GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS}
+ */
+public class AutoRetireUsersTask extends AbstractTask {
+	
+	private static final Logger log = LoggerFactory.getLogger(AutoRetireUsersTask.class);
+	private final String AUTO_RETIRE_REASON = "User retired due to inactivity";
```
dkayiwa, line 37: Shouldn't this be static?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,88 @@
...
+				log.error("Error occurred while auto-retiring users: ", e);
+			} finally {
+				log.debug("Auto-retiring users task ended");
+				stopExecuting();
+			}
+		}
+	}
+
+	Set<User> getUsersToRetire(UserService userService) {
+		List<User> allUsers = userService.getAllUsers();
```
dkayiwa, line 64: How about making this final?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
dkayiwa: How about if it is simply an empty string?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,88 @@
...
+		String numberOfDaysToRetire = Context.getAdministrationService().getGlobalProperty(GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS);
+		
+		if (numberOfDaysToRetire == null) {
+			return Collections.emptySet();
+		}
+		
+		long numberOfMillisecondsToRetire = DateUtil.daysToMilliseconds(Double.parseDouble(numberOfDaysToRetire));
```
dkayiwa, line 71: Why not just use a library or something like `TimeUnit.DAYS.toMillis(days)`?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	@Test
+	public void userInactivityExceedsThreshold_shouldReturnTrueIfInactivityExceedsThreshold() {
+		User user = userService.getUser(1);
+		user.setUserProperty(
+			OpenmrsConstants.USER_PROPERTY_LAST_LOGIN_TIMESTAMP, 
+			String.valueOf(System.currentTimeMillis() - ONE_DAY_IN_MILLISECONDS - 2000)
```
dkayiwa, line 108: Could we instead just use a named constant?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,130 @@
...
+		boolean result = autoRetireUsersTask.userInactivityExceedsDaysToRetire(user, ONE_DAY_IN_MILLISECONDS);
+		assertTrue(result);
+	}
+
+	@Test
+	public void userInactivityExceedsThreshold_shouldReturnFalseIfInactivityDoesNotExceedThreshold() {
+		User user = userService.getUser(1);
+		user.setUserProperty(
+			OpenmrsConstants.USER_PROPERTY_LAST_LOGIN_TIMESTAMP, 
+			String.valueOf(System.currentTimeMillis() - (23 * 60 * 60 * 1000))
```
dkayiwa, line 122: Could we instead just use a named constant? Instead of the comment?

api/src/main/java/org/openmrs/api/UserService.java
dicksonmulli: No, in Unix Timestamp (Epoch Unix Timestamp)

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	@Test
+	public void getUsersToRetire_shouldReturnUsersThatExceedInactivityThreshold() {
+		// Global property for number of days to retire
+		administrationService.saveGlobalProperty(new GlobalProperty(OpenmrsConstants.GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS, "1")); // 1 day
+
+		User inactiveUser = userService.getUser(1);
+		inactiveUser.setUserProperty(
+			OpenmrsConstants.USER_PROPERTY_LAST_LOGIN_TIMESTAMP,
```
dkayiwa, line 53: Could we just use a named constant? Instead of the comment?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
dicksonmulli: This import is used in the class, so we can't remove it. Or did I miss the point?

dkayiwa: You already have an import for `org.openmrs.util.OpenmrsConstants`

api/src/main/java/org/openmrs/api/UserService.java
dkayiwa: That answer could be in the code instead of github comment. 😊

dkayiwa:
@dicksonmulli can you also update the pull request description with the correct JIRA ticket link?

dkayiwa:
It also looks like you have not yet assigned yourself the ticket and taken it through the transition that is documented here: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477199/Pull+Request+Tips

dkayiwa:
@dicksonmulli do you get WIKI notifications? I put for you a comment at https://openmrs.atlassian.net/wiki/spaces/docs/pages/101318663/Creating+Auto-Deactivating+User+Task

dkayiwa:
@dicksonmulli what do you think about adding an exception for admin accounts, to avoid getting locked out of your system? 😊

dicksonmulli:
@dkayiwa  I had assigned myself to the ticket but you updated that. Was there any reason for that?

dicksonmulli:
I don't get WIKI notifications. I am also not able to reply to comment.

Regarding admin accounts, I thought restricting retirement of these users was to be done in another iteration. From this conversation, this feature would be in a separate module and probably include alerts : https://talk.openmrs.org/t/seeking-input-auto-deactivating-emr-user-accounts/38616/9

dkayiwa:
For the WIKI, are you logged in?

Regarding retiring admin accounts, it can lead to locking everyone out of the system. So without it, we are in a dangerous state. Don't you think so?

dicksonmulli:
Yes, I am logged into the WIKI (Actually I can edit).

Regarding admin users, we can find a way of skipping the admin users for now. Is there a way to identify admin users?

dkayiwa:
> Is there a way to identify admin users?

User.isSuperUser()

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,88 @@
...
+
+		return allUsers.stream()
+			.filter(user -> !user.isRetired() && userInactivityExceedsDaysToRetire(user, numberOfMillisecondsToRetire))
+			.collect(Collectors.toSet());
+	}
+
+	boolean userInactivityExceedsDaysToRetire(User user, long numberOfMillisecondsToRetire) {
+		String lastLoginTimeString = Context.getUserService().getLastLoginTime(user);
+
+		if (StringUtils.isNotBlank(lastLoginTimeString)) {
```
dkayiwa, line 81: What happens to users who have never logged in?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,99 @@
...
+		
+		long numberOfMillisecondsToRetire = DateUtil.daysToMilliseconds(Double.parseDouble(numberOfDaysToRetire));
+
+		return allUsers.stream()
+			.filter(user -> !user.isRetired() && userInactivityExceedsDaysToRetire(user, numberOfMillisecondsToRetire))
+			.collect(Collectors.toSet());
+	}
+
+	boolean userInactivityExceedsDaysToRetire(User user, long numberOfMillisecondsToRetire) {
+		if (!user.isSuperUser()) {
```
dkayiwa, line 80: Wouldn't it be cleaner to include the `!user.isSuperUser()` in the stream filter on line 75 above?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,191 @@
...
+		);
+
+		userService.saveUser(user);
+
+		boolean result = autoRetireUsersTask.userInactivityExceedsDaysToRetire(user, ONE_DAY_IN_MILLISECONDS);
+		assertFalse(result);
+	}
+
+	@Test
+	public void userInactivityExceedsDaysToRetire_shouldReturnTrueIfUserInactiveSinceCreation() {
```
dkayiwa, line 157: The method name has should return true but i see one of the return values being false.

dkayiwa:
@dicksonmulli can you also remember to improve on the wiki documentation that you started?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+		userService.saveUser(retiredUser);
+
+		Set<User> usersToRetire = autoRetireUsersTask.getUsersToRetire(userService);
+
+		assertThat(usersToRetire, not(contains(retiredUser)));
+	}
+
+	@Test
+	public void userInactivityExceedsThreshold_shouldReturnTrueIfInactivityExceedsThreshold() {
```
dkayiwa, line 122: Could there be a reason why you replaced `DaysToRetire` with `Threshold`? 😊

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,100 @@
...
+				stopExecuting();
+			}
+		}
+	}
+
+	Set<User> getUsersToRetire(UserService userService) {
+		final List<User> allUsers = userService.getAllUsers();
+		String numberOfDaysToRetire = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS);
+		
+		if (numberOfDaysToRetire == null || numberOfDaysToRetire.isEmpty()) {
```
dkayiwa, line 68: Why not just use `StringUtils.isNotBlank()`?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
dkayiwa: Did you see the above comment?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,100 @@
...
+			} catch (Exception e) {
+				log.error("Error occurred while auto-retiring users: ", e);
+			} finally {
+				log.debug("Auto-retiring users task ended");
+				stopExecuting();
+			}
+		}
+	}
+
+	Set<User> getUsersToRetire(UserService userService) {
```
dkayiwa, line 64: Did you intentionally set this to package-private access?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,100 @@
...
+			.collect(Collectors.toSet());
+	}
+
+	boolean userInactivityExceedsDaysToRetire(User user, long numberOfMillisecondsToRetire) {
+		String lastLoginTimeString = Context.getUserService().getLastLoginTime(user);
+
+		if (StringUtils.isNotBlank(lastLoginTimeString)) {
+			long lastLoginTime = Long.parseLong(lastLoginTimeString);
+
+			boolean b = System.currentTimeMillis() - lastLoginTime >= numberOfMillisecondsToRetire;
```
dkayiwa, line 88: Since b does not look like a good variable name, why not just directly `return System.currentTimeMillis() - lastLoginTime >= numberOfMillisecondsToRetire;`?

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,100 @@
...
+			return b;
+		} else {
+			Date dateCreated = user.getDateCreated();
+			if (dateCreated != null) {
+				long creationTime = dateCreated.getTime();
```
dkayiwa, line 93: I do not see a need of the `creationTime` local variable.

api/src/main/java/org/openmrs/util/DateUtil.java
```diff
@@ -31,4 +31,12 @@ public static Date truncateToSeconds(Date date) {
 		return Date.from(instant);
 	}
 	
+	/**
+	 * This method converts days to milliseconds
+	 * @param days No. of days
+	 * @return Equivalent time in milliseconds, e.g., 172,800,000 milliseconds for 2 days
+	 */
+	public static long daysToMilliseconds(double days) {
```
dkayiwa, line 39: Basing on my comments above, we can get rid of this.

dkayiwa:
@dicksonmulli can you also add the pull request link to the ticket as advised at? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

dicksonmulli:
Are you able to see my comments? Let me repost some comments here

This won't work properly since `TimeUnit.DAYS.toMillis` expects long but we cannot part the numberOfDaysToRetire to long if it's set with a decimal place e.g if numberOfDaysToRetire is 0.02083 then an error will occur

"Did you intentionally set this to package-private access?" Yes, because of the test

dkayiwa:
> if it's set with a decimal place e.g if numberOfDaysToRetire is 0.02083 then an error will occur

Do we have a requirement for number of days to have decimal places?

dicksonmulli:
Do we have a requirement for number of days to have decimal places?

Not really, but we need to allow this ability. For instance, when an admin wants to test this feature, it will come in handy to set say half a day or few hours.

dkayiwa:
> Not really, but we need to allow this ability. For instance, when an admin wants to test this feature, it will come in handy to set say half a day or few hours.

We generally tend to avoid being too permissive. Especially when it has not been explicitly requested for. 😊

dicksonmulli:
> We generally tend to avoid being too permissive. Especially when it has not been explicitly requested for.

I see, however, I genuinely think this ability will be beneficial not only to testing but also preventing errors when the user accidentally adds numbers with decimal places.

dkayiwa:
> I see, however, I genuinely think this ability will be beneficial not only to testing but also preventing errors when the user accidentally adds numbers with decimal places.

For preventing errors, our error handling should take care of it. 
Avoiding a builtin library and add our custom code that increases the maintenance burden, is not worth it, when all this is done in the name of supporting decimal places which is not part of the user requirements. 😊

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
dkayiwa: @dicksonmulli did you see the above comment?

dicksonmulli:
> Did you intentionally set this to package-private access?

Yes, I did this to enable us to call these methods from the unit tests.

[Note] It could be that you can't see my replies anymore. I am not sure what changed.

dkayiwa:
> Yes, I did this to enable us to call these methods from the unit tests.

Ummmmm, it does not generally smell good, to make methods more visible than they should, just for the sake of testing them. We naturally test them via their public contract. Or in the worst case, at least access them via reflection in the tests. 😊

api/src/main/java/org/openmrs/scheduler/tasks/AutoRetireUsersTask.java
```diff
@@ -0,0 +1,99 @@
...
+ * The inactivity duration is set as a global property. 
+ * <a href="https://openmrs.atlassian.net/wiki/spaces/docs/pages/101318663/Creating+Auto-Deactivating+User+Task">Documentation</a>
+ * {@link OpenmrsConstants#GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS}
+ * 
+ * @since 2.7.0
+ */
+public class AutoRetireUsersTask extends AbstractTask {
+	
+	private static final Logger log = LoggerFactory.getLogger(AutoRetireUsersTask.class);
+	static final String AUTO_RETIRE_REASON = "User retired due to inactivity";
```
dkayiwa, line 38: Shouldn't the above be private?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+		verify(userService, atLeastOnce()).retireUser(eq(adminUser), anyString());
+	}
+
+	@Test
+	public void shouldNotRetireSuperUsers() {
+		administrationService.saveGlobalProperty(new GlobalProperty(OpenmrsConstants.GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS, ONE_DAY_PROPERTY_VALUE));
+
+		User adminUser = getDefaultUser();
+		adminUser.addRole(new Role(RoleConstants.SUPERUSER));
```
dkayiwa, line 129: Will this test fail if i remove the above line of adding the super user role?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+		verify(userService, never()).retireUser(activeUser, AutoRetireUsersTask.AUTO_RETIRE_REASON);
+	}
+
+	@Test
+	public void shouldNotRetireAlreadyRetiredUsers() {
+		administrationService.saveGlobalProperty(new GlobalProperty(OpenmrsConstants.GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS, ONE_DAY_PROPERTY_VALUE));
+
+		User retiredUser = getDefaultUser();
+		retiredUser.setRetired(true);
```
dkayiwa, line 87: Will this test fail if i remove the above line of setRetired(true)?

dicksonmulli:
> Shouldn't the above be private?
 I was using it in tests but I can still use anyString() instead of passing the actual reason.

dicksonmulli:
> Will this test fail if i remove the above line of setRetired(true)?

Ideally it should fail but the reason it won't fail when you remove this line is because the algorithm will check if the inactivity exceeds the number of days to retire, of which this won't work unless we add unnecessary stubbings of a last login time or created date - e.g. `when(userService.getLastLoginTime(any())).thenReturn(String.valueOf(System.currentTimeMillis() - TWO_DAYS_IN_MILLISECONDS));`

> Will this test fail if i remove the above line of adding the super user role?

Same case for the above

dkayiwa:
> I was using it in tests but I can still use anyString() instead of passing the actual reason.

Would this catch a bug which changes the expected string to a different one?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dkayiwa: @dicksonmulli did you see the above comment?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dkayiwa: @dicksonmulli did you see the above comment?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dicksonmulli: Commenting should be working now. 

Regarding the above concern, kindly refer to the other reply that I shared with you on slack.

dkayiwa: It is better to share the reply here. These are public comments that are usually read by other community members who may join the discussion.

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dkayiwa: Do you have a response to the above comment?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dicksonmulli: https://github.com/openmrs/openmrs-core/pull/4650#issuecomment-2168116350

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dicksonmulli: https://github.com/openmrs/openmrs-core/pull/4650#issuecomment-2168116350

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dicksonmulli: Ideally it should fail but the reason it won't fail when you remove this line is because the algorithm will check if the inactivity exceeds the number of days to retire, of which it doesn't, unless we add unnecessary stubbings of a last login time or created date - e.g. when(userService.getLastLoginTime(any())).thenReturn(String.valueOf(System.currentTimeMillis() - TWO_DAYS_IN_MILLISECONDS));
Note that these unnecessary stubbing would cause an error.

dkayiwa: I have not properly understood your response. Let me ask differently, if i removed the line which adds the super user role to the user, will this test fail?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dkayiwa: If i do not retire the user in this test, and the test still passes, what then would i be testing for? 😊

dicksonmulli:
> Would this catch a bug which changes the expected string to a different one?

I have updated the tests with the actual string. This may have resulted to duplication of code but on the positive side, we achieve separation of concerns.

dkayiwa:
@dicksonmulli can you look at this merge conflicts?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,161 @@
...
+		autoRetireUsersTask.execute();
+
+		assertFalse(user.isRetired());
+	}
+
+	@Test
+	public void shouldNotRetireAlreadyRetiredUsers() {
+		administrationService.saveGlobalProperty(new GlobalProperty(OpenmrsConstants.GP_NUMBER_OF_DAYS_TO_AUTO_RETIRE_USERS, ONE_DAY_PROPERTY_VALUE));
+
+		String retireReason = "Retire Test users";
```
dkayiwa, line 88: Shouldn't this be a constant?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,161 @@
...
+		User user = getDefaultUser();
+
+		user.getAllRoles().forEach(user::removeRole);
+
+		user.setDateCreated(new Date(System.currentTimeMillis() - TWO_DAYS_IN_MILLISECONDS));
+
+		autoRetireUsersTask.execute();
+
+		assertTrue(user.isRetired());
+		assertEquals(user.getRetireReason(), AUTO_RETIRE_REASON);
```
dkayiwa, line 139: Of the two arguments, which is the expected? `AUTO_RETIRE_REASON` or `user.getRetireReason()`?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,161 @@
...
+		retiredUser.setRetired(true);
+		retiredUser.setRetireReason(retireReason);
+		
+		userService.saveUser(retiredUser);
+
+		autoRetireUsersTask.execute();
+
+		User fetchedUser = userService.getUser(retiredUser.getUserId());
+		assertTrue(fetchedUser.isRetired(), "User should remain retired after the task runs");
+		assertEquals(fetchedUser.getRetireReason(), retireReason);
```
dkayiwa, line 108: Of these two arguments, which is the expected?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
```diff
@@ -0,0 +1,161 @@
...
+			OpenmrsConstants.USER_PROPERTY_LAST_LOGIN_TIMESTAMP,
+			String.valueOf((System.currentTimeMillis() - TWO_DAYS_IN_MILLISECONDS))
+		);
+
+		user.getAllRoles().forEach(user::removeRole);
+
+		autoRetireUsersTask.execute();
+
+		assertTrue(user.isRetired());
+		assertEquals(user.getRetireReason(), AUTO_RETIRE_REASON);
```
dkayiwa, line 63: Which is the expected of these two arguments?

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dicksonmulli: Not really. This is because the variable's scope in within this method.

dkayiwa: I mean a constant within the method.

api/src/test/java/org/openmrs/scheduler/tasks/AutoRetireUsersTaskTest.java
dicksonmulli: `AUTO_RETIRE_REASON` is the expected. I will switch them around.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -43,6 +43,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.openmrs.util.OpenmrsConstants.ZERO_LOGIN_ATTEMPTS_VALUE;
```
dkayiwa, line 46: Our convention has not been to import static. Can we do it like we have done else where for OpenmrsConstants?

PR #4703, opened 2024-08-01
TRUNK-6247: Merging Patients should not copy Patient Programs on "losing" patients

mogoodrich:
…osing" patients

## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/person/PersonMergeLogData.java
+++ b/api/src/main/java/org/openmrs/person/PersonMergeLogData.java
@@ -38,9 +38,9 @@ public class PersonMergeLogData {
 	private List<String> movedEncounters;
 	
 	/**
-	 * List of UUIDs of patient programs copied from non-preferred to preferred
+	 * List of UUIDs of patient programs moved from non-preferred to preferred
 	 */
-	private List<String> createdPrograms;
+	private List<String> movedPrograms;
 	
 	/**
 	 * List of UUIDs of voided relationships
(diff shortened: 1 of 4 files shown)


ibacher (approved):

dkayiwa (approved):

api/src/main/java/org/openmrs/person/PersonMergeLogData.java
```diff
@@ -38,9 +38,9 @@ public class PersonMergeLogData {
 	private List<String> movedEncounters;
```
mseaton, line 38: Unfortunately this is a public API, and we really don't know how it is being used.  I think we should just deprecate the existing `createdPrograms` methods and add the new `movedPrograms` methods, rather than rename/replace them.  We can then remove them in some future release.

api/src/main/java/org/openmrs/person/PersonMergeLogData.java
```diff
@@ -323,8 +323,8 @@ public int computeHashValue() {
 		if (getCreatedOrders() != null) {
 			str += getCreatedOrders().toString();
 		}
-		if (getCreatedPrograms() != null) {
-			str += getCreatedPrograms().toString();
```
mseaton, line 327: Likewise.

api/src/main/java/org/openmrs/person/PersonMergeLogData.java
mogoodrich: Fair enough, will do.

ibacher: Yeah, I was kind of evaluating this as something added in 2.7, where removing a seldom-used API is kind of a non-issue (we don't use it anywhere else); however, for backporting to 2.6, we would need to retain (and deprecate) the "created programs" stuff.

FWIW, I couldn't find any uses of this class outside of forks of OpenMRS-Core, so it seems very unlikely to be an issue.

mogoodrich:
Whoops, good catch @ibacher !  I thought I had merged this, but evidently I hadn't!  I had even backported.

PR #4709, opened 2024-08-09
TRUNK-6257: Fix visit overlapping logic and improve performance

jayasanka-sack:

The main goal of this PR is to fix the validation logic. Previously, the logic was incorrect and didn’t cover all use cases, allowing users to start multiple visits at the same time, even when overlapping visits were disabled, leading to complaints like "cannot start overlapping visit" when trying to end a visit. 

Another focus is improving efficiency. Currently, the VisitValidator loads all non-voided visits for the patient and checks one by one to ensure the start and stop dates don't overlap with the visit being validated, which causes n+1 selects problems and unnecessarily loads entire Visit objects just to check two properties.

This change pushes the work down to the database. Instead of loading all visits for a patient, we load only those where the period between the visit start and stop dates overlaps with the visit being edited. If there's more than one visit or if the single visit isn’t the same as the visit being edited, we know there’s an overlap and can reject the request.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6257

  No? -> execute above command


api/src/main/java/org/openmrs/api/VisitService.java
```diff
@@ -295,6 +295,18 @@ public List<Visit> getVisits(Collection<VisitType> visitTypes, Collection<Patien
...
+	 * Gets all visits for the specified patient that started on or after the given start date
+	 * @param patient
+	 * @param maxStartDatetime
+	 * @param minEndDatetime
+	 * @return a list of visits
+	 * @throws APIException
+	 */
+	@Authorized(PrivilegeConstants.GET_VISITS)
+	public List<Visit> getVisitsByPatient(Patient patient, Date maxStartDatetime, Date minEndDatetime)
+	        throws APIException;
```
mseaton, line 308: I don't understand the search criteria or use case here, but I would much rather see us support a method like `getVisits(VisitSearchCriteria criteria)` where VisitSearchCriteria contains properties including `patient`, `minStartDatetime`, `maxStartDatetime`, `minEndDateTime`, `maxEndDateTime`, etc.  It seems odd to have a method that takes these very specific search critieria, and will be limiting in the future when we inevitably want to support limiting by other criteria.  Better to add additional possible, nullable attributes to `VisitSearchCriteria` that create lots of service methods with overlapping but slightly different arguments.

api/src/main/java/org/openmrs/api/VisitService.java
```diff
@@ -295,6 +295,18 @@ public List<Visit> getVisits(Collection<VisitType> visitTypes, Collection<Patien
 	public List<Visit> getVisitsByPatient(Patient patient, boolean includeInactive, boolean includeVoided)
 	        throws APIException;
 	
+	/**
+	 * Gets all visits for the specified patient that started on or after the given start date
```
mseaton, line 299: Comment does not fully match the method signature.

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -105,15 +105,15 @@ public void validate(Object target, Errors errors) {
 		// check attributes
```
mseaton, line 105: What does this have to do with TRUNK-6257 - Add a function to VisitService to get visits that overlap with a specified time period ?

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -105,15 +105,15 @@ public void validate(Object target, Errors errors) {
...
 		super.validateAttributes(visit, errors, Context.getVisitService().getAllVisitAttributeTypes());
 		
-		// check start and end dates
+		// check for overlapping visits
 		if (disallowOverlappingVisits()) {
```
mseaton, line 109: What exactly are you trying to achieve here?  There is no context as to what the problem is that you are trying to solve in the ticket description, and nothing here that indicates what the problem is that you are solving.  What was wrong before and how does this fix it?

mseaton (changes requested):

api/src/main/java/org/openmrs/validator/VisitValidator.java
ibacher: The goal here is basically efficiency (though the ticket may be poorly described). The `VisitValidator` currently works by loading _all_ non-voided visits for the patient and then checking one-by-one that the start date and stop date don't overlap with the visit being validated in Java code. This means that we hit the n+1 selects problem as well as unnecessarily materializing entire Visit objects just to check two properties.

The change here basically aims to push the work down to the database; i.e., instead of loading all visits for a given patient, we load all visits where the period between the visits start date and stop date overlaps with that of the visit being edited. If there is more than one visit or if the single visit is not the same visit as the visit being edited, then we know there's an overlapping visit and can reject the request.

jayasanka-sack: @mseaton @mogoodrich I'm really sorry about that. The PR was still a work in progress, so I hadn't included any details yet. Thanks @ibacher for explaining the issue.

api/src/main/java/org/openmrs/api/VisitService.java
jayasanka-sack: That's a great suggestion. I created a method with search criteria here: https://github.com/openmrs/openmrs-core/pull/4709/commits/1b54547d89d7da85f7c051cf01e7c0d0a9df72cd

There's a set of overloaded methods for searching visits, I'll update those in a separate PR.

api/src/main/java/org/openmrs/parameter/VisitSearchCriteria.java
```diff
@@ -0,0 +1,250 @@
...
+	private Collection<Patient> patients;
+	private Collection<Location> locations;
+	private Collection<Concept> indications;
+	private Date minStartDatetime;
+	private Date maxStartDatetime;
+	private Date minEndDatetime;
+	private Date maxEndDatetime;
+	private Map<VisitAttributeType, String> serializedAttributeValues;
+	private boolean includeInactive = true;
+	private boolean includeVoided = true;
```
ibacher, line 39: Default `false`

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -105,15 +109,25 @@ public void validate(Object target, Errors errors) {
...
-				if (visit.equals(otherVisit)) {
-					continue;
-				}
-				validateStartDatetime(visit, otherVisit, errors);
-				validateStopDatetime(visit, otherVisit, errors);
+		// Some tests are using non-existing patients. Skipping overlapping validation for such cases.
+		boolean nonExistingPatient = visit.getPatient() == null || visit.getPatient().getId() == null;
+		
+		// check for overlapping visits
+		if (disallowOverlappingVisits() && !nonExistingPatient) {
```
ibacher, line 116:
```suggestion
		if (!nonExistingPatient && disallowOverlappingVisits()) {
```

api/src/main/java/org/openmrs/validator/VisitValidator.java
```diff
@@ -105,15 +109,25 @@ public void validate(Object target, Errors errors) {
...
+			VisitSearchCriteria visitSearchCriteria = new VisitSearchCriteriaBuilder()
+				.setPatients(Collections.singletonList(visit.getPatient()))
```
ibacher, line 118: It seems like it would be helpful for the CriteriaBuilder to have a `setPatient()` handler...

api/src/test/resources/org/openmrs/include/standardTestDataset.xml
```diff
@@ -325,9 +325,9 @@
...
-  <visit visit_id="2" patient_id="2" visit_type_id="1" date_started="2005-01-02 00:00:00.0" creator="1" date_created="2005-01-01 00:00:00.0" voided="0" uuid="4e663d66-6b78-11e0-93c3-18a905e044dc" />
-  <visit visit_id="3" patient_id="2" visit_type_id="2" date_started="2005-01-03 00:00:00.0" creator="1" date_created="2005-01-01 00:00:00.0" voided="0" uuid="6f85b2a6-6b78-11e0-93c3-18a905e044dc" />
+  <visit visit_id="1" patient_id="2" visit_type_id="1" date_started="2005-01-05 00:00:00.0" location_id="1" indication_concept_id="5497" creator="1" date_created="2005-01-01 00:00:00.0" voided="0" uuid="1e5d5d48-6b78-11e0-93c3-18a905e044dc" />
+  <visit visit_id="2" patient_id="2" visit_type_id="1" date_started="2005-01-02 00:00:00.0" date_stopped="2005-01-02 10:00:00.0" creator="1" date_created="2005-01-01 00:00:00.0" voided="0" uuid="4e663d66-6b78-11e0-93c3-18a905e044dc" />
+  <visit visit_id="3" patient_id="2" visit_type_id="2" date_started="2005-01-03 00:00:00.0" date_stopped="2005-01-03 10:00:00.0" creator="1" date_created="2005-01-01 00:00:00.0" voided="0" uuid="6f85b2a6-6b78-11e0-93c3-18a905e044dc" />
```
ibacher, line 330: These changes should be new visits as this file is used by modules.

api/src/main/java/org/openmrs/parameter/VisitSearchCriteriaBuilder.java
```diff
@@ -0,0 +1,176 @@
...
+import java.util.Collection;
+import java.util.Date;
+import java.util.Map;
+
+/**
+ * A builder class for constructing instances of {@link VisitSearchCriteria}.
+ * This builder allows for the flexible creation of {@link VisitSearchCriteria}
+ * objects by providing a simple interface to set various fields.
+ */
+public class VisitSearchCriteriaBuilder {
```
ibacher, line 27: nit: I always prefer builders that don't explicitly use `get...` or `set...`. It looks better to write something like:

```java
getVisits(VisitSearchCriteriaBuilder.patient(xxxx));
```

api/src/main/java/org/openmrs/api/VisitService.java
```diff
@@ -258,6 +259,15 @@ public List<Visit> getVisits(Collection<VisitType> visitTypes, Collection<Patien
...
 	
+	/**
+	 * Gets the visits matching the specified search criteria
+	 * 
+ 	 * @param visitSearchCriteria
+	 * @return
+	 * @throws APIException
+	 */	
+	@Authorized(PrivilegeConstants.GET_VISITS)
+	public List<Visit> getVisits(VisitSearchCriteria visitSearchCriteria) throws APIException;
```
ibacher, line 270: nit: blank line between function definitions

```suggestion
	public List<Visit> getVisits(VisitSearchCriteria visitSearchCriteria) throws APIException;

```

api/src/main/java/org/openmrs/api/VisitService.java
```diff
@@ -258,6 +259,15 @@ public List<Visit> getVisits(Collection<VisitType> visitTypes, Collection<Patien
 	        Date minEndDatetime, Date maxEndDatetime, Map<VisitAttributeType, Object> attributeValues,
 	        boolean includeInactive, boolean includeVoided) throws APIException;
 	
+	/**
+	 * Gets the visits matching the specified search criteria
+	 * 
+ 	 * @param visitSearchCriteria
+	 * @return
+	 * @throws APIException
+	 */
```
dkayiwa, line 268: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#New-Public-Methods-and-Classes

api/src/main/java/org/openmrs/parameter/VisitSearchCriteria.java
```diff
@@ -0,0 +1,250 @@
...
+import java.util.Collection;
+import java.util.Date;
+import java.util.Map;
+
+/**
+ * The VisitSearchCriteria class encapsulates the criteria used for searching visits in the system.
+ * It includes various fields that allow filtering visits based on different parameters such as visit type,
+ * patient, location, and date ranges. It also provides options to include or exclude inactive and voided visits.
+ */
+public class VisitSearchCriteria {
```
dkayiwa, line 27: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#New-Public-Methods-and-Classes

api/src/main/java/org/openmrs/parameter/VisitSearchCriteriaBuilder.java
```diff
@@ -0,0 +1,176 @@
...
+import org.openmrs.VisitAttributeType;
+import org.openmrs.VisitType;
+
+import java.util.Collection;
+import java.util.Date;
+import java.util.Map;
+
+/**
+ * A builder class for constructing instances of {@link VisitSearchCriteria}.
+ * This builder allows for the flexible creation of {@link VisitSearchCriteria}
```
dkayiwa, line 24: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#New-Public-Methods-and-Classes

api/src/main/java/org/openmrs/parameter/VisitSearchCriteria.java
jayasanka-sack: I added `@since 2.6.8` Is it correct? or should it be `2.7.0`?

ibacher: Usually, for backported things, we list both.

jayasanka-sack: Since 2.6.8 is the latest snapshot and 2.7.0 hasn't been released, do we still need to include both? There's no gap, so I'm not sure.

ibacher: Yes.

jayasanka-sack: Sure, I'll include both. I'm still a bit unclear on the reason though, let's talk about it in our next 1:1.

ibacher: Here this is less of a big deal, but in some cases, we backport to multiple versions, so it's helpful to be able to say "this is in 2.7.0, 2.6.8, and 2.5.13". If we just said "2.5.13" that could leave the mistaken impression that its available in 2.6.0. It may be a little silly here if we only backport this to 2.6, but it also clearly documents that this was originally committed against 2.7 (which is what `master` is) and then backported.

jayasanka-sack: Makes sense. Thanks!

ibacher (approved):
LGTM

PR #4710, opened 2024-08-12
TRUNK-6256: Draft moving threading to a single threadpool

ibacher:
## Issue I worked on

  No? -> execute above command


api/src/main/java/org/openmrs/api/context/Daemon.java
```diff
@@ -390,30 +360,114 @@ public void run() {
...
+				}
+			}
+		});
+	}
+	
+	private static Future<?> runInDaemonThreadInternal(Runnable runnable) {
+		// for Threads, we used to guarantee that Thread.start() was called before the function returned
+		// since we cannot guarantee that the executor actually started executing the thread, we use a CountDownLatch
+		// to emulate this behaviour when the user submits a Thread. Other runnables are unaffected.
+		CountDownLatch countDownLatch = getCountDownLatch(runnable instanceof Thread);
```
ibacher, line 442: There are two differences between this function and the one that operates on `Callable`s: the obvious difference that a `Runnable` has no result, and the more subtle distinction that a `Thread` is a `Runnable`. Any `Thread` objects that are submitted will be guaranteed to resume _after_ the thread has been started and the session is open.

The reason for the above hack is to continue to support this idiom:

```java
Thread myThread = new Thread() {
	// snip
}

runInDaemonThread(myThread, token);

thread.join();
```

Basically if `runInDaemonThread()` returns control to consume code before the thread is actually started, it's easy to get in a state where the thread.join()` call returns immediately because the thread isn't running yet. The delay until we hit the `CountDownLatch` ensures that when `thread.join()` is called the thread will have been started (though, obviously we make no guarantee that the thread has not finished by the time it's returned.

api/src/main/java/org/openmrs/api/context/Daemon.java
```diff
@@ -390,30 +360,114 @@ public void run() {
...
+		
+		try {
+			countDownLatch.await();
+		} catch (InterruptedException ignored) {}
+		
+		return result;
+	}
+
+	private static CountDownLatch getCountDownLatch(boolean isThread) {
+		return isThread ? new CountDownLatch(1) : new CountDownLatch(0);
```
ibacher, line 468: If we're using a Thread, we need to wait for the CountDownLatch, otherwise, we don't.

web/src/main/java/org/openmrs/web/WebDaemon.java
```diff
@@ -10,63 +10,52 @@
...
 
+import java.util.concurrent.ExecutionException;
+
 /**
  * This class provides {@link Daemon} functionality in a web context.
  * 
  * @since 1.9
  */
-public class WebDaemon extends Daemon {
+public final class WebDaemon {
```
ibacher, line 27: Technically, this is a breaking change, since `WebDaemon` used to also export the public methods of `Daemon`. However, we never used that internally and I could not find any code on GitHub that made use of `WebDaemon` (other than various forks of this repo).

web/src/main/java/org/openmrs/web/filter/update/UpdateFilter.java
```diff
@@ -627,7 +630,7 @@ public synchronized Map<String, Object[]> getErrors() {
 		 */
 		public void start() {
 			setUpdatesRequired(true);
-			thread.start();
+			OpenmrsThreadPoolHolder.threadExecutor.submit(r);
```
ibacher, line 633: This is basically the idiom for starting a background thread without any privileges.

dkayiwa:
@ibacher do you happen to know, off head, why the changes in this pull request lead to the failure of `DeamonTest.createUser_shouldCreateUserWithRolesInContextDAO` when run on mysql? This is the command to run tests on mysql in a docker container: `mvn test -DuseInMemoryDatabase=false -DtestDb=mysql`

This is the error log: https://pastebin.com/cpaiNSf9

FWIW, the test passes when this method is reverted to what it was before the changes in the pull request: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/context/Daemon.java#L116-L153

ibacher:
@dkayiwa Not really, but I suggested that @wikumChamith might try something similar to the code I explained [here](https://github.com/openmrs/openmrs-core/pull/4710/files#r1719815597) to see if that fixes things. I added it to resolve a race condition that showed up in the normal tests. (Short version: add a count-down latch with a value of 1 and call it's `countDown()` method only once the session is loaded; this will effectively block the calling thread until the context is established on the new thread which may resolve the issue).

PR #4718, opened 2024-08-16
Resolve 3.3.2

slubwama:
## Issue I worked on

**Issue: 3.3.2**

  No? -> execute above command

### DIFF
--- a/webapp/src/main/webapp/WEB-INF/web.xml
+++ b/webapp/src/main/webapp/WEB-INF/web.xml
@@ -339,6 +339,7 @@
 	<session-config>
 	    <cookie-config>
 	        <http-only>true</http-only>
+			<secure>true</secure>
 	    </cookie-config>
 	</session-config>
 	


rkorytkowski:
Have you tested it locally? It is very likely that it won't work if you are not using https to access the web server, which is typical for localhost (development) and also may be for some simple implementations that run in LAN and didn't setup https.

slubwama:
@

> Have you tested it locally? It is very likely that it won't work if you are not using https to access the web server, which is typical for localhost (development) and also may be for some simple implementations that run in LAN and didn't setup https.

@rkorytkowski I tested in an HTTP environment and it seemed fine

ibacher:
So, I actually think this PR might be causing an issue described [here](https://talk.openmrs.org/t/failure-to-login-after-database-initialization-no-error-shown-at-frontend/43766). The problem is that the RFC leaves it up to the user-agent (browser) to determine what "Secure" means. Most browsers simply define "localhost" as being a secure context, so this means that this would work for HTTP requests run locally, but not for HTTP requests served over the network.

It's actually hard, given the nature of web proxies, to detect whether a request is made via HTTPS. For example, our infra serves everything from behind nginx proxies, which is also where we terminate SSL. This means that even if, e.g., a request to https://dev3.openmrs.org/openmrs/ uses "HTTPS", by the time it hits Tomcat, its just a plain HTTP request. However, this issue itself is probably better resolved at the nginx reverse proxy level rather than here.

ynurmahomed:
We are facing issues with our implementations that run in LAN, likely due to this change. Can it be made configurable?

PR #4806, opened 2024-10-27
TRUNK-6276: Fix ValidateHibernateMappingsDatabaseIT test errors when running with MySQL

wikumChamith:
## Issue I worked on

see https://openmrs.atlassian.net/browse/TRUNK-6276

  No? -> execute above command


api/src/main/java/org/openmrs/BaseOpenmrsMetadata.java
```diff
@@ -38,7 +39,8 @@ public abstract class BaseOpenmrsMetadata extends BaseOpenmrsObject implements O
 	@Field
 	private String name;
 	
-	@Column(name = "description", length = 255)
+	@Column(name = "description", columnDefinition = "text")
+	@Lob
```
dkayiwa, line 43: I do not think a description of length 255 is a Lob.

wikumChamith: Some tables use text for the description field, while others use varchar. Defining description as text in the superclass and overriding it as varchar in the child classes when necessary helps mitigate compatibility issues between H2 and MySQL during tests

dkayiwa: Because the majority of the descriptions are small, it does not make sense to me making it a LOB in the base class.

wikumChamith: If we don't include lob here we'll get an error when running ValidHibernateMappings with H2.
`
Schema-validation: wrong column type encountered in column [description] in table [order_type]; found [clob (Types#CLOB)], but expecting [text (Types#VARCHAR)]`

if we set this to "varchar" and try to override to "text" when necessary we still get the error as we can't add @Lob for overrides.

dkayiwa: Thinking about this again, which object would need text that is too big to require a LOB, for just a mere description?

wikumChamith: Tables like global_property, order_type, and form use `text` for `description`.
Should we move the tables that use `text` for description to something like `varchar(1024)`?

dkayiwa: I think we should do so. I cannot think of a reason why any of those would require a LOB. Let us change them.

api/src/main/java/org/openmrs/BaseOpenmrsMetadata.java
```diff
@@ -38,7 +39,7 @@ public abstract class BaseOpenmrsMetadata extends BaseOpenmrsObject implements O
 	@Field
 	private String name;
 	
-	@Column(name = "description", length = 255)
+	@Column(name = "description", length = 1024)
```
dkayiwa, line 42: Why don't we leave the default as 255?

api/src/main/java/org/openmrs/BaseOpenmrsMetadata.java
```diff
@@ -38,7 +39,7 @@ public abstract class BaseOpenmrsMetadata extends BaseOpenmrsObject implements O
 	@Field
 	private String name;
 	
-	@Column(name = "description", length = 255)
```
dkayiwa, line 41: Shouldn't the default length be 255?

api/src/main/java/org/openmrs/PatientIdentifierType.java
```diff
@@ -32,7 +33,8 @@
 @Table(name = "patient_identifier_type")
 @Audited
 @AttributeOverrides({
-	@AttributeOverride(name = "name", column = @Column(name = "name", nullable = false, length = 50))
+	@AttributeOverride(name = "name", column = @Column(name = "name", nullable = false, length = 50)),
+	@AttributeOverride(name = "description", column = @Column(name = "description", columnDefinition = "text"))
```
dkayiwa, line 37: Why should a patient identifier type description have a column definition of text?

api/src/main/java/org/openmrs/PersonAttributeType.java
```diff
@@ -31,6 +33,9 @@
 @Entity
 @Table(name = "person_attribute_type")
 @Audited
+@AttributeOverrides({
+	@AttributeOverride(name = "description", column = @Column(name = "description", columnDefinition = "text"))
```
dkayiwa, line 37: Same question as above

api/src/main/java/org/openmrs/BaseOpenmrsMetadata.java
wikumChamith: The default length of varchar is set to 255 by JPA, so there's no need to explicitly define it

wikumChamith:
@dkayiwa I updated the PR

api/src/main/java/org/openmrs/Location.java
```diff
@@ -47,7 +48,10 @@
 @Entity
 @Table(name = "location")
 @Cache(usage = CacheConcurrencyStrategy.READ_WRITE)
-@AttributeOverride(name = "attributes", column = @Column(name = "location_id"))
+@AttributeOverrides(value = {
+	@AttributeOverride(name = "attributes", column = @Column(name = "location_id")),
+	@AttributeOverride(name = "description", column = @Column(name = "description", columnDefinition = "varchar(255)", length = 255))
```
dkayiwa, line 53: do we need to duplicate the above length?

api/src/main/java/org/openmrs/PatientIdentifierType.java
```diff
@@ -32,7 +32,7 @@
 @Table(name = "patient_identifier_type")
 @Audited
 @AttributeOverrides({
-	@AttributeOverride(name = "name", column = @Column(name = "name", nullable = false, length = 50))
+	@AttributeOverride(name = "name", column = @Column(name = "name", nullable = false, length = 50)),
```
dkayiwa, line 35: Is this extra comma needed?

api/src/main/java/org/openmrs/api/db/SerializedObject.java
```diff
@@ -28,6 +31,9 @@
 @Entity
 @Table(name = "serialized_object")
 @Audited
+@AttributeOverrides(value = {
+	@AttributeOverride(name = "description", column = @Column(name = "description", length = 5000))
```
dkayiwa, line 35: Shouldn't this also be text?

dkayiwa: Oh wait, why should a description be that large?

wikumChamith: create table serialized_object
(
    serialized_object_id int auto_increment
        primary key,
    name                 varchar(255)         not null,
    description          varchar(5000)        null,
    type                 varchar(255)         not null,
    subtype              varchar(255)         not null,
    serialization_class  varchar(255)         not null,
    serialized_data      mediumtext           not null,
    date_created         datetime             not null,
    creator              int                  not null,
    date_changed         datetime             null,
    changed_by           int                  null,
    retired              tinyint(1) default 0 not null,
    date_retired         datetime             null,
    retired_by           int                  null,
    retire_reason        varchar(1000)        null,
    uuid                 char(38)             not null,
    constraint uuid
        unique (uuid),
    constraint serialized_object_changed_by
        foreign key (changed_by) references users (user_id),
    constraint serialized_object_creator
        foreign key (creator) references users (user_id),
    constraint serialized_object_retired_by
        foreign key (retired_by) references users (user_id)
);

DDL for this table looks like this. What do you suggest?

api/src/main/resources/org/openmrs/api/db/hibernate/DiagnosisAttribute.hbm.xml
```diff
@@ -46,7 +46,9 @@
 
 		<property name="voidReason" type="java.lang.String" column="void_reason" length="255" />
 
-		<property name="uuid" type="java.lang.String" column="uuid" length="38" unique="true" />
+		<property name="uuid" type="java.lang.String" unique="true">
```
dkayiwa, line 49: For the changes on uuids, were they failing on any of the databases?

api/src/main/java/org/openmrs/api/db/SerializedObject.java
dkayiwa: Looks like that was a typo 😊

wikumChamith: should it be `text` or `varchar` ?

PR #4818, opened 2024-11-06
Adding maven.development.version and build.release.type variables to Bamboo

wikumChamith:
## Issue I worked on

  No? -> execute above command

### DIFF
--- a/bamboo-specs/bamboo.yml
+++ b/bamboo-specs/bamboo.yml
@@ -252,62 +252,62 @@ Release:
               interpreter: SHELL
               scripts:
                   - |-
-                        #!/bin/bash -eux
-    
-                        set +x
-                        
-                        export OMRS_VERSION=${bamboo.maven.release.version}
-                        export IMAGE=${bamboo.docker.image.name}:${OMRS_VERSION}
-                        export DEV_IMAGE=${bamboo.docker.image.name}:${OMRS_VERSION}-dev
-                        export BUILD_IMAGE=${bamboo.docker.image.id}
+                      #!/bin/bash -eux
                       
-                        # Set to be able to push
-                        git remote set-url origin git@github.com:openmrs/openmrs-core.git
-                        
-                        docker login -u ${bamboo.dockerhub.username} -p ${bamboo.dockerhub.password}
-                        
-                        echo "Setting the release version"
-                        docker pull ${BUILD_IMAGE}
-                        docker run --rm -v m2-repo:/root/.m2/repository -v $(pwd):/openmrs_core \
-                        ${BUILD_IMAGE} mvn versions:set -DnewVersion=${OMRS_VERSION} -DgenerateBackupPoms=false
-                        git commit -am "[skip-ci] Releasing ${OMRS_VERSION}"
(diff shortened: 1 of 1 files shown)


dkayiwa:
@wikumChamith can we remove the formatting changes?

wikumChamith:
Yes. That error we got was not in the code I added.

dkayiwa:
Do you mind explaining a bit more why the error happened only after those changed were merged?

wikumChamith:
> Do you mind explaining a bit more why the error happened only after those changed were merged?

We are still getting that error. Here is the failure for the revert you did: https://ci.openmrs.org/browse/TRUNK-MASTER-3603/commit

dkayiwa:
I am still trying to understand why it was not failing before our attempt. 😊

wikumChamith:
Is it possible it did but somehow got overlooked?

wikumChamith:
@dkayiwa, we've encountered this error before as well. For instance, here’s the same issue in @ibacher's changes on this file: https://ci.openmrs.org/browse/TRUNK-MASTER-3482

dkayiwa:
Does the fix require all these formatting changes?

wikumChamith:
> Does the fix require all these formatting changes?

Yes

dkayiwa:
How do you tell so?

wikumChamith:
> How do you tell so?

I used a YML validator to check.

dkayiwa:
Ok let us see how this goes.

wikumChamith:
It passed :) https://ci.openmrs.org/browse/TRUNK-MASTER-3606

ibacher:
What's the point of adding the `maven.development.version` variable? We don't seem to actually be _using_ it since we always set the development version using this: `${DEV_IMAGE} mvn versions:set -DnextSnapshot=true -DgenerateBackupPoms=false` and we don't re-export that back to Bamboo.

Traditionally, `maven.development.version` gets passed as the second argument to [this script](https://github.com/openmrs/openmrs-contrib-bamboo/blob/master/release-prepare-perform.sh), and the use-case is to be able to specify the next SNAPSHOT version (which is not really something we do too much).

wikumChamith:
@ibacher, is it possible to handle cases like the following when doing a release without using the `maven.development.version` variable?

- When releasing Platform 2.7.0-alpha, we want the next development version to remain Platform 2.7.0-SNAPSHOT.
- After the full release of Platform 2.7.0, we want the next development version to be Platform 2.7.1-SNAPSHOT.


cc: @rkorytkowski"

dkayiwa:
@rkorytkowski would that be better than just setting the next development version variable in bamboo and have all that automatically done during the release task?

ibacher:
> is it possible to handle cases like the following when doing a release without using the maven.development.version variable?

Yes. There’s a job in the RefApp for O3 that does something very similar. But my real point was that `maven.development.version` is an OpenMRS convention, not a Bamboo or Maven thing. So just adding the variable to the CI job doesn’t actually do what you want. You need to swap out the line that sets the next development version with the value of the variable.

PR #4830, opened 2024-11-09
TRUNK-6252: Conditional resource path should be interpreted as a globs

k4pran:

This PR updates the logic to use globs instead of regex when matching conditional resources. This logic is only applied to config versions 2.0 and onwards, earlier config.xml versions will still use regex. 

The default pattern is `glob:**/%s` to make it agnostic to absolute paths, allowing users to focus on matching relative paths instead.

This change also includes adding the config.xml version to the `Module` class as `configVersion` to allow us to check if we should use glob or regex logic. This PR also uses the `ModuleUtil.matchRequiredVersions` function as this seems to work well for configVersions that only have two parts like 2.0. 

I tried to add parameterized tests to cover different glob input patterns and config versions, please let me know if  you notice any other patterns that should be included.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6252

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/Module.java
+++ b/api/src/main/java/org/openmrs/module/Module.java
@@ -83,6 +83,8 @@ public final class Module {
 	
 	private Set<String> packagesWithMappedClasses = new HashSet<>();
 	
+	private String configVersion;
+	
 	private Document config = null;
 	
 	private Document sqldiff = null;
@@ -116,13 +118,15 @@ public final class Module {
 	 * @param description
 	 * @param version
 	 */
-	public Module(String name, String moduleId, String packageName, String author, String description, String version) {
+	public Module(String name, String moduleId, String packageName, String author, String description, String version,
+				  String configVersion) {
 		this.name = name;
 		this.moduleId = moduleId;
 		this.packageName = packageName;
 		this.author = author;
 		this.description = description;
 		this.version = version;
+		this.configVersion = configVersion;
 		log.debug("Creating module " + name);
 	}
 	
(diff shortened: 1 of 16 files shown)


api/src/main/java/org/openmrs/module/Module.java
```diff
@@ -644,7 +648,15 @@ public Document getConfig() {
 	public void setConfig(Document config) {
 		this.config = config;
 	}
-	
+
+	public String getConfigVersion() {
```
dkayiwa, line 652: Can we put `@since` annotations on these? https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#New-Public-Methods-and-Classes

api/src/main/java/org/openmrs/module/ModuleClassLoader.java
```diff
@@ -405,6 +408,41 @@ static boolean shouldResourceBeIncluded(Module module, URL fileUrl, String openm
 		return include;
 	}
 	
+	static boolean isMatchingConditionalResource(Module module, URL fileUrl, ModuleConditionalResource conditionalResource) {
```
dkayiwa, line 411: Did you intentionally decide not to make this method private?

k4pran: Yes it was intentional, I think it was important to unit test this method separately from `shouldResourceBeIncluded` to make sure it is returning true/false based on the pattern matching rather than the logic of `shouldResourceBeIncluded`. I made it package-private for this reason.

PR #4920, opened 2025-02-12
TRUNK-6300 Add Storage Service

rkorytkowski:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6300

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/StorageService.java
+++ b/api/src/main/java/org/openmrs/api/StorageService.java
@@ -1,0 +1,176 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.time.format.DateTimeFormatter;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+import org.openmrs.api.storage.ObjectMetadata;
+import org.openmrs.api.stream.StreamDataWriter;
+import org.springframework.stereotype.Service;
+
+/**
+ * Storage service to persist data that does not fit well in DB.
+ * <p>
+ * Data objects are stored under generated or provided unique keys. Data and keys are immutable.
+ * <p>
+ * {@link ObjectMetadata} can be passed in save methods. It is implementation dependent on how the metadata is 
+ * used or persisted.
+ * 
+ * @since 2.8.0, 2.7.4, 2.6.16, 2.5.15
+ */
+public interface StorageService extends OpenmrsService {
+
+	Pattern moduleIdOrGroupPattern = Pattern.compile("[\\w-./]+");
+
+	/**
+	 * Get InputStream to read data for the given key.
+	 *
+	 * @param key unique key
+	 * @return data
+	 * @throws IOException wrong key or IO error
+	 */
+	InputStream getData(String key) throws IOException;
+
+
+	/**
+	 * Get InputStream to read temporary data for the given key.
+	 * 
+	 * @param key unique key
+	 * @return data
+	 * @throws IOException wrong key or IO error
+	 */
+	InputStream getTempData(String key) throws IOException;
+
+	/**
+	 * Returns metadata for the file associated under the given key.
+	 * 
+	 * @param key the key
+	 * @return the metadata
+	 * @throws IOException wrong key or IO error
+	 */
+	ObjectMetadata getMetadata(String key) throws IOException;
+	
+	/**
+	 * Returns keys starting with the given prefix (the order depends on the implementation).
+	 * <p>
+	 * Remember to close the stream when done iterating.
+	 *
+	 * @param moduleIdOrGroup module id or group (allowed characters <code>[\\w-./]+</code>)
+	 * @param keyPrefix key prefix
+	 * @return stream of keys
+	 * @throws IOException IO error
+	 */
+	Stream<String> getKeys(String moduleIdOrGroup, String keyPrefix) throws IOException;
+	
+	/**
+	 * Saves the given InputStream with auto-generated key (should be preferred to avoid key collision and use 
+	 * the best key structure).
+	 *
+	 * @param inputStream data
+	 * @param metadata object metadata or null
+	 * @param moduleIdOrGroup module id or group (allowed characters <code>[\\w-./]+</code>)
+	 * @return unique key
+	 * @throws IOException IO error
+	 */
+	String saveData(InputStream inputStream, ObjectMetadata metadata, String moduleIdOrGroup) throws IOException;
+	
+	/**
+	 * Saves the given InputStream as a temporary file. Temporary files are kept, if possible, in local
+	 * storage for fast IO. You shall assume that they are not accessible between HTTP requests and different replicas.
+	 *
+	 * @param inputStream data
+	 * @param metadata object metadata or null
+	 * @return unique key
+	 * @throws IOException IO error
+	 */
+	String saveTempData(InputStream inputStream, ObjectMetadata metadata) throws IOException;
+
+	/**
+	 * Saves data as a temporary file. Temporary files are kept, if possible, in local
+	 * storage for fast IO. You shall assume that they are not accessible between HTTP requests and different replicas.
+	 * <p>
+	 * It uses pipes internally to avoid copying data into memory and writes directly to storage in a separate thread.
+	 *
+	 * @param writer lambda to write to stream
+	 * @param metadata object metadata or null   
+	 * @return unique key
+	 * @throws IOException IO error
+	 */
+	String saveTempData(StreamDataWriter writer, ObjectMetadata metadata) throws IOException;
+	
+	/**
+	 * Saves the given InputStream under the given key suffix.
+	 * <p>
+	 * {@link #saveData(InputStream, ObjectMetadata, String)} should be preferred to avoid key collision and use the 
+	 * best key structure.
+	 * 
+	 * @param inputStream data
+	 * @param metadata object metadata or null      
+	 * @param moduleIdOrGroup module id or group (allowed characters <code>[\\w-./]+</code>)
+	 * @param keySuffix for the unique key
+	 * @return unique key
+	 * @throws IOException if key exists or IO error
+	 */
+	String saveData(InputStream inputStream, ObjectMetadata metadata, String moduleIdOrGroup,
+					String keySuffix) throws IOException;
+	
+	/**
+	 * Saves the data under the given key suffix.
+	 * <p>
+	 * {@link #saveData(StreamDataWriter, ObjectMetadata, String)} should be preferred to avoid key collision and use
+	 * the best key structure.
+	 * <p>
+	 * It uses pipes internally to avoid copying data into memory and writes directly to storage in a separate thread.
+	 *
+	 * @param writer lambda to write to stream
+	 * @param metadata object metadata or null                    
+	 * @param moduleIdOrGroup module id or group (allowed characters '[\w-\./]+')
+	 * @param keySuffix for the unique key
+	 * @return unique key
+	 * @throws IOException if key exists or IO error
+	 */
+	String saveData(StreamDataWriter writer, ObjectMetadata metadata, String moduleIdOrGroup, String keySuffix) 
+		throws IOException;
+
+	/**
+	 * Saves the data with auto-generated key (should be preferred to avoid key collision and use the best key 
+	 * structure).
+	 * <p>
+	 * It uses pipes internally to avoid copying data into memory and writes directly to storage in a separate thread.
+	 *
+	 * @param writer lambda to write to stream
+	 * @param metadata object metadata or null      
+	 * @param moduleIdOrGroup module id or group (allowed characters '[\w-\./]+')
+	 * @return unique key
+	 * @throws IOException if key exists or IO error
+	 */
+	String saveData(StreamDataWriter writer, ObjectMetadata metadata, String moduleIdOrGroup) throws IOException;
+
+	/**
+	 * Marks data for deletion. The key may be freed up after some period.
+	 *
+	 * @param key unique key
+	 * @return true if marked for deletion, false if not exists
+	 * @throws IOException IO error
+	 */
+	boolean purgeData(String key) throws IOException;
+	
+	/**
+	 * Returns true if object with the given key exists.
+	 * 
+	 * @param key unique key
+	 * @return true if exists
+	 */
+	boolean exists(String key);
+}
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/openmrs/api/StorageService.java
```diff
@@ -0,0 +1,94 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.stream.Stream;
+
+/**
+ * Storage service to persist data that does not fit well in DB.
+ * <p>
+ * Data objects are stored under generated or provided unique keys. Data and keys are immutable.
+ * 
+ * @since 2.8.0, 2.7.2, 2.6.16, 2.5.15
```
rkorytkowski, line 21: Ideally we backport the interface and implementation to all versions used by o3, o2, bahmni, ozone... It is to be able to easily add support in modules. We will only modify the master branch to use the new service in api. Does this list of versions cover what is used by implementations these days?

rkorytkowski:
Please do not merge until I refactor complex obs to use this service in https://openmrs.atlassian.net/jira/software/c/projects/TRUNK/issues/TRUNK-6304

rkorytkowski:
Also I need to add key encoding so it doesn't include special characters. It might be needed to add mimetype support as well as it's used in https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java#L117

api/src/main/java/org/openmrs/api/StorageService.java
```diff
@@ -0,0 +1,94 @@
...
+	 * Returns keys starting with the given prefix (the order depends on the implementation).
+	 * <p>
+	 * Remember to close the stream when done iterating.
+	 *
+	 * @param moduleId null for global or module id
+	 * @param prefix key prefix
+	 * @return stream of keys
+	 * @throws IOException IO error
+	 */
+	Stream<String> getKeys(String moduleId, String prefix) throws IOException;
```
dkayiwa, line 44: What is the use case for this?

rkorytkowski: It's pretty common to list files in a directory, e.g. https://github.com/openmrs/openmrs-module-reporting/blob/master/api/src/main/java/org/openmrs/module/reporting/definition/library/implementerconfigured/BaseImplementerConfiguredDefinitionLibrary.java#L125

dkayiwa: Oh i see! 😊

dkayiwa: So would a key be a folder name? For instance `lucene` as we currently have in the application data directory?

rkorytkowski: Key points to an actual file not a directory, but a key can include a directory e.g. lucene/data.dat

dkayiwa: Can you also give an example of how the `prefix` would differ from the key?

rkorytkowski: I call it prefix, because it matches any keys starting with the prefix e.g. having 'abc/ab', 'abc/abc', 'ab', 'abc/abcd', 'abc/ac' and the prefix 'abc/' we would get back keys 'abc/ab', 'abc/abc', 'abc/abcd' but not 'ab' and 'abc/ac'.

api/src/main/java/org/openmrs/api/impl/LocalStorageService.java
```diff
@@ -0,0 +1,158 @@
...
+	@Override
+	public Stream<String> getKeys(final String moduleId, final String prefix) throws IOException {
+		List<String> dirs = new ArrayList<>();
+		final Path storagePath;
+		if (moduleId != null) {
+			dirs.add("modules");
+			dirs.add(moduleId);
+			storagePath = Paths.get(storageDir, "modules", moduleId);
+		} else {
+			storagePath = Paths.get(storageDir);
```
ibacher, line 73: If `moduleId` is `null`, should we filter the `modules` directory from the returned results? Or else are we creating the contract that every storage service should have a top-level key called `modules` that stores the module-specific data?

ibacher (commented):
Thanks @rkorytkowski! A few questions / comments.

api/src/main/java/org/openmrs/api/StorageService.java
rkorytkowski: The thing is that if you want to correct your module to use storage service and it's only available in 2.8.0 then your module is incompatible with all older versions. It's not something module authors wanted to do historically. If we backport it to older branches then modules only need to require new maintenance branches of core, but do not have to suddenly drop support for core older than 2.8.0.

api/src/main/java/org/openmrs/api/impl/LocalStorageService.java
```diff
@@ -0,0 +1,158 @@
...
+	}
+	
+	@Override
+	public InputStream getData(final String key) throws IOException {
+		return Files.newInputStream(getPath(key));
+	}
+	
+	public Path getPath(final String key) {
+		Path path = Paths.get(key);
+		if (path.isAbsolute()) {
```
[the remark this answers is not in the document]
rkorytkowski: This is to support old storage location and new storage location as well as temp files. I'll add more clarification around that.

api/src/main/java/org/openmrs/api/impl/LocalStorageService.java
```diff
@@ -0,0 +1,158 @@
...
+		Path newPath = Paths.get(storageDir).resolve(key);
+		Files.createDirectories(newPath.getParent());
+		
+		return newPath;
+	}
+	
+	private Path newKey(String moduleId, String keySuffix) {
+		if (keySuffix == null) {
+			Date date = new Date();
+			keySuffix = dateFormat.format(date) + UUID.randomUUID().toString().substring(0, 8);
```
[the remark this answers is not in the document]
rkorytkowski: Fair enough.

api/src/main/java/org/openmrs/api/StorageService.java
```diff
@@ -0,0 +1,94 @@
...
+	 * @param inputStream data
+	 * @param moduleId null for global or module id
+	 * @return unique key
+	 * @throws IOException IO error
+	 */
+	String saveData(InputStream inputStream, String moduleId) throws IOException;
+	
+	/**
+	 * Saves the given InputStream as a temporary file. Temporary files are kept, if possible, in local
+	 * storage for fast IO. They are available only for the duration of a request and for the single
```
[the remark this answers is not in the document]
rkorytkowski: It's not tied to any request in particular. It's just to mind setting. Even in replication you can always expect that a request runs in a single VM and has access to local temp files.

api/src/main/java/org/openmrs/api/impl/LocalStorageService.java
rkorytkowski: I'll slightly change this in the next revision I push.

api/src/main/java/org/openmrs/api/StorageService.java
ibacher: > It's not something module authors wanted to do historically

If that's the problem we're trying to solve, we'll need to backport to a lot further. While I don't know of any major deployments running against 2.3.x or lower, for example, very few modules are written that target anything newer than that and many of them target substantially older versions. E.g., attachments targets 2.2.0, module-spa targets 2.4.1, reporting targets 1.9.9.

I think the best path here is to add this to 2.8.0 and let module authors figure out how they want to handle the migration, e.g., force the module to only be compatible with 2.8.0 or add their own abstraction layer that handles both.

api/src/main/java/org/openmrs/api/impl/LocalStorageService.java
```diff
@@ -0,0 +1,158 @@
...
+		if (moduleId == null) {
+			return Paths.get(keySuffix);
+		} else {
+			return Paths.get("modules", moduleId, keySuffix);
+		}
+	}
+	
+	@Override
+	public String saveTempData(InputStream inputStream) throws IOException {
+		Path tempFile = Files.createTempFile("openmrs-temp", ".tmp");
```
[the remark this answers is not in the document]
rkorytkowski: In such a case we become responsible for cleaning up the temp directory. We do mark temp files as deleteOnExit, but it only works if JVM exits gracefully. I'd rather have temp files land in default OS location and use standard OS cleanup. Also you need extra case when setting up your volumes to exclude a custom temp dir from a volume and be it part of the container.

api/src/main/java/org/openmrs/api/StorageService.java
rkorytkowski: It's tricky to add the abstraction layer... I'm still hoping to have it back ported to make the migration easier. Let's discuss on platform call.

api/src/main/java/org/openmrs/api/StorageService.java
rkorytkowski: Changed the var to keyPrefix. Hope it is more clear now.

api/src/main/java/org/openmrs/api/StorageService.java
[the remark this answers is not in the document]
rkorytkowski: Clarified the comment.

api/src/main/java/org/openmrs/api/impl/LocalStorageService.java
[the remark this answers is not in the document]
rkorytkowski: Please see https://github.com/openmrs/openmrs-core/pull/4944/files#r1971611823

api/src/main/java/org/openmrs/api/impl/LocalStorageService.java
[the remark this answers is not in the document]
rkorytkowski: Let's discuss on the platform call.

api/src/main/java/org/openmrs/api/storage/LocalStorageService.java
```diff
@@ -0,0 +1,195 @@
...
+
+/**
+ * Used to persist data in a local file system or volumes.
+ * <p>
+ * It is the default implementation of StorageService.
+ * 
+ * @since 2.8.0, 2.7.2, 2.6.16, 2.5.15
+ */
+@Conditional(StorageServiceCondition.class)
+@Qualifier("local")
```
rkorytkowski, line 47: In order to introduce e.g. "s3" storage one would annotate the class with:
{code}
@Conditional(StorageServiceCondition.class)
@Qualifier("s3")
{code}
Then at runtime you would need to provide storage_type=s3 property to switch from local to s3.

api/src/main/java/org/openmrs/api/impl/LocalStorageService.java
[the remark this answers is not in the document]
rkorytkowski: As discussed on the platform call I'll introduce getTempData to access temporary files and it will use OS default temp directory to go around using absolute paths.

api/src/main/java/org/openmrs/api/StorageService.java
dkayiwa: I actually like experimenting with this new approach as an alternative to the mess we have been going through to support multiple platform versions. How far to backport would not be a big deal for now because if at any one time we find that we need a lower back port, all we need is do it and release its corresponding maintenance version. As simple as that. 😊

rkorytkowski: All right. Could I backport it at least as far as 2.5.x for now so implementations on refapp 2.x can use it?

api/src/main/java/org/openmrs/api/impl/LocalStorageService.java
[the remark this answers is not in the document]
rkorytkowski: Changed to disallow absolute paths and resolve legacy paths on the fly.

api/src/main/java/org/openmrs/api/impl/LocalStorageService.java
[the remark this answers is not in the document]
rkorytkowski: Added getTempData.

api/src/main/java/org/openmrs/api/StorageService.java
dkayiwa: No objection from me. 👍

dkayiwa: `2.7.2` has moved to `2.7.4`

api/src/main/java/org/openmrs/api/storage/BaseStorageService.java
```diff
@@ -0,0 +1,71 @@
...
+		return saveData(dataWriter, metadata, moduleIdOrGroup, null);
+	}
+
+	public String saveData(StreamDataWriter dataWriter, ObjectMetadata metadata, String moduleIdOrGroup,
+						   String keySuffix) throws IOException {
+		return saveData(streamService.streamData(dataWriter, metadata != null ? metadata.getLength() : null), metadata, 
+			moduleIdOrGroup, keySuffix);
+	}
+
+	public String saveTempData(InputStream inputStream, ObjectMetadata metadata) throws IOException {
```
dkayiwa, line 61: What is the use of the `metadata` parameter?

api/src/main/java/org/openmrs/api/storage/LocalStorageService.java
```diff
@@ -0,0 +1,225 @@
...
+		
+		return ObjectMetadata.builder()
+				.setLength(attributes.size())
+				.setMimeType(mimetypes.getContentType(fileName))
+				.setFilename(fileName)
+				.setCreationTime(attributes.creationTime().toInstant()).build();
+	}
+
+	Path getPath(String key) {
+		key = key.replace('/', File.separatorChar); //MS Windows support
```
dkayiwa, line 95: For windows wouldn't you replacing `\` instead of `/`?

api/src/main/java/org/openmrs/api/storage/BaseStorageService.java
rkorytkowski: It allows to set e.g. the size of input. Based on the size we can choose the best method to move data e.g. copy in-memory or use pipes to write directly to disk. If no size is provided then we always write directly to disk, which is less efficient for smaller files.

api/src/main/java/org/openmrs/api/storage/LocalStorageService.java
rkorytkowski: Keys always have '/' as a separator. In case of Windows '/' will be replaced with File.separatorChar, which is '\'.

api/src/main/java/org/openmrs/api/storage/LocalStorageService.java
```diff
@@ -0,0 +1,225 @@
...
+		return Files.newInputStream(getPath(key));
+	}
+
+	/**
+	 * It needs to be evaluated each time as it changes over time in tests...
+	 * 
+	 * @return the legacy storage dir
+	 * @deprecated since 2.8.0
+	 */
+	@Deprecated
```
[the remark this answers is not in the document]
dkayiwa: Do you think we should add that explanation to the method description?

api/src/main/java/org/openmrs/api/storage/LocalStorageService.java
dkayiwa: Oh i see! 😊

rkorytkowski:
Addressed all review comments.

dkayiwa:
@rkorytkowski do we just close this after merged https://github.com/openmrs/openmrs-core/pull/4944?

rkorytkowski:
Yes. Thanks!

PR #4929, opened 2025-02-16
TRUNK-5208 ModuleUtil compareVersions should sort a SNAPSHOT version as "less than"

suniala:

TRUNK-5208 compareVersion orders SNAPSHOT version before regular one

That means a version string like "1.2.3-SNAPSHOT" is ordered before version string "1.2.3".

Also update javadoc and give more descriptive names to variables.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5208

- [ x ] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

- [ x ] I have **added tests** to cover my changes. (If you refactored

- [ x ] I ran `mvn clean package` right before creating this pull request and

  No? -> execute above command

- [ x ] All new and existing **tests passed**.

- [ x ] My pull request is **based on the latest changes** of the master branch.

### DIFF
--- a/api/src/main/java/org/openmrs/module/ModuleUtil.java
+++ b/api/src/main/java/org/openmrs/module/ModuleUtil.java
@@ -322,9 +322,9 @@ public class ModuleUtil {
 						upperBound = upperBound.replaceAll("\\*", Integer.toString(Integer.MAX_VALUE));
 					}
 					
-					int lowerReturn = compareVersion(version, lowerBound);
+					int lowerReturn = compareVersionIgnoringSnapshotQualifier(version, lowerBound);
 					
-					int upperReturn = compareVersion(version, upperBound);
+					int upperReturn = compareVersionIgnoringSnapshotQualifier(version, upperBound);
 					
 					if (lowerReturn < 0 || upperReturn > 0) {
 						log.debug("Version " + version + " is not between " + lowerBound + " and " + upperBound);
@@ -332,7 +332,7 @@ public class ModuleUtil {
 						return true;
 					}
 				} else {
-					if (compareVersion(version, range) < 0) {
+					if (compareVersionIgnoringSnapshotQualifier(version, range) < 0) {
 						log.debug("Version " + version + " is below " + range);
 					} else {
 						return true;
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -322,17 +322,17 @@ public static boolean matchRequiredVersions(String version, String versionRange)
 						upperBound = upperBound.replaceAll("\\*", Integer.toString(Integer.MAX_VALUE));
 					}
 					
-					int lowerReturn = compareVersion(version, lowerBound);
+					int lowerReturn = compareVersionIgnoringSnapshotQualifier(version, lowerBound);
```
mogoodrich, line 325: Is there a reason we want to use the legacy compare here that ignores the snapshot qualifier?

suniala: Good point. The reason for ignoring qualifiers here was this existing test case:
```
public void matchRequiredVersions_shouldMatchWhenVersionHasWildCardPlusQualifierAndIsWithinBoundary() {
	String openmrsVersion = "1.4.*-SNAPSHOT";
	String requiredVersion = "1.4.0 - 1.4.10";
	assertTrue(ModuleUtil.matchRequiredVersions(openmrsVersion, requiredVersion));
}
```
Now I see I did not think this through. Following the qualifier comparison logic described in the ticket, here `openmrsVersion` should match if `*` is 1-9 but not if it is 0. To me it is not obvious what `matchRequiredVersions` should return in this case.

We can also look at a simpler case. The following test should pass, right?
```
public void matchRequiredVersions_shouldNotMatchWhenVersionHasQualifierAndIsOnLowerBoundary() {
	String openmrsVersion = "1.4.0-SNAPSHOT";
	String requiredVersion = "1.4.0 - 1.4.10";
	assertFalse(ModuleUtil.matchRequiredVersions(openmrsVersion, requiredVersion));
}
```
However, it does not pass with my modifications.

Do you have suggestions on how to resolve the issue?

api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -392,65 +392,87 @@ public static void checkRequiredVersion(String version, String versionRange) thr
...
+
+	/**
+	 * Compare two version strings. Any version qualifiers are ignored in the comparison.
+	 *
+	 * @param versionA String like 1.9.2.0, may include a qualifier like "-SNAPSHOT", may be null
+	 * @param versionB String like 1.9.2.0, may include a qualifier like "-SNAPSHOT", may be null
+	 * @return the value <code>0</code> if versions are equal; a value less than <code>0</code> if first version is
+	 * 		   before the second one; a value greater than <code>0</code> if first version is after the second one.
+	 */
+	public static int compareVersionIgnoringSnapshotQualifier(String versionA, String versionB) {
```
suniala, line 416: Here `SnapshotQualifier` is misleading because the qualifier may be something else than "SNAPSHOT". I will rename this function.

api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -392,65 +392,87 @@ public static void checkRequiredVersion(String version, String versionRange) thr
...
+		return compareVersion(versionA, versionB, true);
+	}
+
+	private static int compareVersion(String versionA, String versionB, boolean ignoreSnapshotQualifier) {
```
suniala, line 420: Here `SnapshotQualifier` is misleading because the qualifier may be something else than "SNAPSHOT". I will rename the parameter.

api/src/main/java/org/openmrs/module/ModuleUtil.java
mogoodrich: Interesting.  Yes, this 1.4.0-SNAPSHOT does not fulfill the requirement for versions 1.4.0 to 1.4.10, you are correct.

I think the existing test you pointed out is somewhat nonsensical and I don't think we should have to worry about fulfilling it... @dkayiwa  thoughts?

dkayiwa: I think in the past we have expected `1.4.0-SNAPSHOT` to be within the `1.4.0 - 1.4.10` range.

mogoodrich: Okay, thanks @dkayiwa I'm fine leaving things that way they for legacy compatibility.

suniala: I hope I understood correctly that the implementation is fine. I added another test case to document the "we have expected 1.4.0-SNAPSHOT to be within the 1.4.0 - 1.4.10 range".

mogoodrich (approved):
This works for me... @dkayiwa  are you okay with this?

PR #4940, opened 2025-02-24
TRUNK-6309: Fix logging issue at Listener.java

wikumChamith:
## Issue I worked on

see https://openmrs.atlassian.net/browse/TRUNK-6309

  No? -> execute above command


web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -43,6 +45,7 @@
 import org.w3c.dom.Element;
 import org.xml.sax.InputSource;
 
+import javax.security.auth.login.Configuration;
```
samuelmale, line 48: Do you need this import?

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -9,7 +9,9 @@
  */
 package org.openmrs.web;
 
+import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.core.config.Configurator;
 import org.openmrs.api.context.Context;
 import org.openmrs.logging.OpenmrsLoggingUtil;
```
samuelmale, line 16: I guess we no longer need this?

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -231,9 +234,7 @@ public void contextInitialized(ServletContextEvent event) {
 					OpenmrsUtil.setApplicationDataDirectory(null);
 				}
 				
-				//ensure that we always log the runtime properties file that we are using
```
dkayiwa, line 234: Do we need to remove the comment?

dkayiwa (approved):

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -230,10 +231,9 @@ public void contextInitialized(ServletContextEvent event) {
 				if (StringUtils.hasLength(appDataRuntimeProperty)) {
 					OpenmrsUtil.setApplicationDataDirectory(null);
 				}
-				
 				//ensure that we always log the runtime properties file that we are using
 				//since openmrs is just booting, the log levels are not yet set. TRUNK-4835
-				OpenmrsLoggingUtil.applyLogLevel(getClass().toString(), "INFO");
```
ibacher, line 236: Uh... If this is broken, we have a much bigger problem, since this is also what we use to set the logging levels via GP. I suggest we try to work that out rather than this.

dkayiwa: @ibacher would you rather we just set this to `INFO`? https://github.com/openmrs/openmrs-core/blob/master/webapp/src/main/resources/log4j2.xml#L48
 For it would fix this problem?

wikumChamith: I first tried this, and it worked. However, I refrained from using it since it introduces additional log lines that were not present before.

ibacher: > For it would fix this problem?

Well, my point was more that if the `applyLogLevel()` call doesn't function correctly, then changing logging via the GP will no longer work.

How we go about changing the logging level here isn't too important.

dkayiwa: @wikumChamith can we create a ticket for this?

wikumChamith: ticket: https://openmrs.atlassian.net/browse/TRUNK-6325

wikumChamith: Looks like this only needs a simple fix :)

New PR: https://github.com/openmrs/openmrs-core/pull/4967

PR #4944, opened 2025-02-26
TRUNK-6304 Refactor code to use Storage Service

rkorytkowski:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6304

  No? -> execute above command

- [x ] All new and existing **tests passed**.

- [x ] My pull request is **based on the latest changes** of the master branch.

### DIFF
--- a/api/src/main/java/org/openmrs/api/StorageService.java
+++ b/api/src/main/java/org/openmrs/api/StorageService.java
@@ -1,0 +1,176 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.time.format.DateTimeFormatter;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+import org.openmrs.api.storage.ObjectMetadata;
+import org.openmrs.api.stream.StreamDataWriter;
+import org.springframework.stereotype.Service;
+
+/**
+ * Storage service to persist data that does not fit well in DB.
+ * <p>
+ * Data objects are stored under generated or provided unique keys. Data and keys are immutable.
+ * <p>
+ * {@link ObjectMetadata} can be passed in save methods. It is implementation dependent on how the metadata is 
+ * used or persisted.
+ * 
+ * @since 2.8.0, 2.7.4, 2.6.16, 2.5.15
+ */
+public interface StorageService extends OpenmrsService {
+
+	Pattern moduleIdOrGroupPattern = Pattern.compile("[\\w-./]+");
+
+	/**
+	 * Get InputStream to read data for the given key.
+	 *
+	 * @param key unique key
+	 * @return data
+	 * @throws IOException wrong key or IO error
+	 */
+	InputStream getData(String key) throws IOException;
+
+
+	/**
+	 * Get InputStream to read temporary data for the given key.
+	 * 
+	 * @param key unique key
+	 * @return data
+	 * @throws IOException wrong key or IO error
+	 */
+	InputStream getTempData(String key) throws IOException;
+
+	/**
+	 * Returns metadata for the file associated under the given key.
+	 * 
+	 * @param key the key
+	 * @return the metadata
+	 * @throws IOException wrong key or IO error
+	 */
+	ObjectMetadata getMetadata(String key) throws IOException;
+	
+	/**
+	 * Returns keys starting with the given prefix (the order depends on the implementation).
+	 * <p>
+	 * Remember to close the stream when done iterating.
+	 *
+	 * @param moduleIdOrGroup module id or group (allowed characters <code>[\\w-./]+</code>)
+	 * @param keyPrefix key prefix
+	 * @return stream of keys
+	 * @throws IOException IO error
+	 */
+	Stream<String> getKeys(String moduleIdOrGroup, String keyPrefix) throws IOException;
+	
+	/**
+	 * Saves the given InputStream with auto-generated key (should be preferred to avoid key collision and use 
+	 * the best key structure).
+	 *
+	 * @param inputStream data
+	 * @param metadata object metadata or null
+	 * @param moduleIdOrGroup module id or group (allowed characters <code>[\\w-./]+</code>)
+	 * @return unique key
+	 * @throws IOException IO error
+	 */
+	String saveData(InputStream inputStream, ObjectMetadata metadata, String moduleIdOrGroup) throws IOException;
+	
+	/**
+	 * Saves the given InputStream as a temporary file. Temporary files are kept, if possible, in local
+	 * storage for fast IO. You shall assume that they are not accessible between HTTP requests and different replicas.
+	 *
+	 * @param inputStream data
+	 * @param metadata object metadata or null
+	 * @return unique key
+	 * @throws IOException IO error
+	 */
+	String saveTempData(InputStream inputStream, ObjectMetadata metadata) throws IOException;
+
+	/**
+	 * Saves data as a temporary file. Temporary files are kept, if possible, in local
+	 * storage for fast IO. You shall assume that they are not accessible between HTTP requests and different replicas.
+	 * <p>
+	 * It uses pipes internally to avoid copying data into memory and writes directly to storage in a separate thread.
+	 *
+	 * @param writer lambda to write to stream
+	 * @param metadata object metadata or null   
+	 * @return unique key
+	 * @throws IOException IO error
+	 */
+	String saveTempData(StreamDataWriter writer, ObjectMetadata metadata) throws IOException;
+	
+	/**
+	 * Saves the given InputStream under the given key suffix.
+	 * <p>
+	 * {@link #saveData(InputStream, ObjectMetadata, String)} should be preferred to avoid key collision and use the 
+	 * best key structure.
+	 * 
+	 * @param inputStream data
+	 * @param metadata object metadata or null      
+	 * @param moduleIdOrGroup module id or group (allowed characters <code>[\\w-./]+</code>)
+	 * @param keySuffix for the unique key
+	 * @return unique key
+	 * @throws IOException if key exists or IO error
+	 */
+	String saveData(InputStream inputStream, ObjectMetadata metadata, String moduleIdOrGroup,
+					String keySuffix) throws IOException;
+	
+	/**
+	 * Saves the data under the given key suffix.
+	 * <p>
+	 * {@link #saveData(StreamDataWriter, ObjectMetadata, String)} should be preferred to avoid key collision and use
+	 * the best key structure.
+	 * <p>
+	 * It uses pipes internally to avoid copying data into memory and writes directly to storage in a separate thread.
+	 *
+	 * @param writer lambda to write to stream
+	 * @param metadata object metadata or null                    
+	 * @param moduleIdOrGroup module id or group (allowed characters '[\w-\./]+')
+	 * @param keySuffix for the unique key
+	 * @return unique key
+	 * @throws IOException if key exists or IO error
+	 */
+	String saveData(StreamDataWriter writer, ObjectMetadata metadata, String moduleIdOrGroup, String keySuffix) 
+		throws IOException;
+
+	/**
+	 * Saves the data with auto-generated key (should be preferred to avoid key collision and use the best key 
+	 * structure).
+	 * <p>
+	 * It uses pipes internally to avoid copying data into memory and writes directly to storage in a separate thread.
+	 *
+	 * @param writer lambda to write to stream
+	 * @param metadata object metadata or null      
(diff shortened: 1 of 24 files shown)


rkorytkowski:
It contains changes from both TRUNK-6300 and TRUNK-6304. They are in separate commits as I'm hoping to backport TRUNK-6300.

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -46,112 +59,144 @@ public AbstractHandler() {
...
+	 * @see org.openmrs.obs.ComplexObsHandler#getObs(Obs, String)
 	 */
-	@Deprecated
-	public String getExtension(String filename) {
-		String result = FilenameUtils.getExtension(filename);
+	public Obs getObs(Obs obs, String view) {
+		// Deprecated
+		// Check if the file exists in the legacy location, if not use obs value complex as key
+		File file = AbstractHandler.getComplexDataFile(obs);
+		String key = file.exists() ? file.getAbsolutePath() : parseDataKey(obs);
```
rkorytkowski, line 83: It's how we benefit from StorageService.getData(key) supporting absolute paths. We have much less deprecated code this way.

api/src/test/java/org/openmrs/api/ObsServiceTest.java
```diff
@@ -1351,17 +1312,14 @@ public void purgeObs_shouldDeleteTheGivenObsFromTheDatabase() {
 		
 		
 		executeDataSet(COMPLEX_OBS_XML);
+		File complexObsDir = OpenmrsUtil.getDirectoryInApplicationDataDirectory(adminService
+			.getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_COMPLEX_OBS_DIR));
+		File createdFile = createImage(complexObsDir, "openmrs_logo_small.gif");
+		
 		Obs complexObs = obsService.getComplexObs(44, ComplexObsHandler.RAW_VIEW);
```
rkorytkowski, line 1319: getCompexObs failed due to missing file with a new implementation. It appears that the old implementation didn't complain that the file didn't exist. I'm confused whether we should fail if the file is missing or silently ignore.

api/src/main/java/org/openmrs/api/storage/LocalStorageService.java
```diff
@@ -0,0 +1,195 @@
...
+@Qualifier("local")
+public class LocalStorageService extends BaseStorageService implements StorageService {
+	
+	private final String storageDir;
+	
+	private final DateTimeFormatter keyDateTimeFormat = DateTimeFormatter.ofPattern("yyyy/MM-dd/HH-mm/yyyy-MM-dd-HH-mm-ss-SSS-");
+	
+	private final MimetypesFileTypeMap mimetypes = new MimetypesFileTypeMap();
+	
+	public LocalStorageService(@Value("${storage_dir}") String storageDir, @Autowired StreamDataService streamService) {
```
wikumChamith, line 56: How can we define the storage_dir? Do we have a property source?

rkorytkowski: I don't think we have a property source properly configured in openmrs-core. In such a case the way to define it is to use environment or system variable storage_dir. It's a typical Docker approach of configuring an app.

We probably want to configure property source from openmrs-runtime.properties, but that deserves a separate issue.

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -36,6 +43,12 @@ public class AbstractHandler {
 	
 	protected NumberFormat nf;
 	
+	@Autowired
```
wikumChamith, line 46: Is there a specific reason for keeping these access modifiers as default?

api/src/main/java/org/openmrs/api/storage/LocalStorageService.java
rkorytkowski: It would first use value from the env property if not found try sys property and finally openmrs-runtime.properties. See also https://docs.spring.io/spring-boot/docs/2.6.1/reference/html/features.html#features.external-config.typesafe-configuration-properties.relaxed-binding

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
rkorytkowski: Default lets us easily mock them in tests that we put in the same package. I included them in a constructor so it's not needed in this case, but sometimes constructor would be too long. I don't know if we have a convention around that, but it's something I find quite common in spring code.  They could be protected as well, but then mocking is only feasible via the constructor.

api/src/main/java/org/openmrs/api/StorageService.java
```diff
@@ -0,0 +1,234 @@
...
+public interface StorageService extends OpenmrsService {
+
+	DateTimeFormatter dateTimeFormat = DateTimeFormatter.ISO_OFFSET_DATE_TIME;
+
+	Pattern moduleIdOrGroupPattern = Pattern.compile("[\\w-./]+");
+	
+	class ObjectMetadata {
+		private static final Logger log = LoggerFactory.getLogger(ObjectMetadata.class);
+		
+		private final Map<String, String> all = new HashMap<>();
```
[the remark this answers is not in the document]
rkorytkowski: Adjusted to be more informative.

api/src/test/java/org/openmrs/api/ObsServiceTest.java
rkorytkowski: @dkayiwa @ibacher do you have any opinion about that?

rkorytkowski:
It's no longer a draft. There's one final question I have around failing if file is missing. See above. Other than that I consider it ready. https://github.com/openmrs/openmrs-core/pull/4920 should be merged first.

api/src/test/java/org/openmrs/api/ObsServiceTest.java
dkayiwa: In order to reduce the possibility of unintended side effects of introducing a new storage service, i would preserve the existing behaviour. Then if we need to change it to fail if the file is missing, i would do that as a separate ticket and pull request. 😊

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -34,124 +40,154 @@ public class AbstractHandler {
...
-		log.debug("value complex: " + obs.getValueComplex());
-		log.debug("file path: " + file.getAbsolutePath());
-		ComplexData complexData = null;
+	public Obs saveObs(Obs obs) throws APIException {
 		try {
-			complexData = new ComplexData(file.getName(), OpenmrsUtil.getFileAsBytes(file));
+			byte[] data = (byte[]) obs.getComplexData().getData();
+			
+			String key = storageService.saveData(outputStream -> {
+				if (outputStream == null) System.out.println("outputStream is null");
```
dkayiwa, line 91: Did you just forget to replace the `System.out.println()` statements? 😊

dkayiwa:
@rkorytkowski did you get a chance to test these changes in running webapp? For instance, i have tried them locally on the equivalent of this page `https://qa-refapp2.openmrs.org/openmrs/admin/observations/obs.form` but iam not able to view the complex obs image that i attached to a patient.

rkorytkowski:
@dkayiwa Let me retest this. It's RAW_VIEW, which I think I tested...

rkorytkowski:
@dkayiwa it's an issue specific to ImageHandler. I added file size as length of data, but it's different after reading into BufferedImage, which is uncompressed and then compressing again in legacyui controller, thus content-length mismatch issue in the browser. It's by the way unnecessary decompressing and compressing, but it's legacyui so I won't be fixing that.

dkayiwa:
> I will just not include file size in the response as it was previously done.

Ok. Happy to test again when you are done with it. :)

api/src/test/java/org/openmrs/api/ObsServiceTest.java
rkorytkowski: Changed to old behaviour in ImageHandler. Other handlers did fail if file was missing so we may want to fix that indeed in a separate issue.

rkorytkowski:
Addressed all review comments. Tested all handlers through legacyui with new data and data saved before new logic.

dkayiwa:
Do you know what could be going on here? I have used the O2 Refapp to add an attachment and got this error: https://pastebin.com/pWXaadmH

rkorytkowski:
@dkayiwa  org.openmrs.module.attachments.obs.ImageAttachmentHandler extends AbstractHandler which has changed... ImageAttachmentHandler would have to be changed as well. Let me have a look if we can make changes in the attachments module in a way that works with backwards compatibility without introducing submodule to support 2.8.0.

rkorytkowski:
@dkayiwa I had a look at the attachment module and in fact it needs to be refactored to use the new storage service to run on 2.8.0. It's not only using AbstractHandler, but accessing files directly, which is no longer allowed in 2.8.0. Once TRUNK-6300 is merged and back-ported, we will bump the attachment module to require openmrs-core 2.5.15 and make it use the new storage service. Currently it's requiring openmrs-core 2.2.0.

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -34,124 +40,148 @@ public class AbstractHandler {
...
+		}
+		return key;
+	}
+
 	/**
 	 * Convenience method to create and return a file for the stored ComplexData.data Object
 	 * 
 	 * @param obs
 	 * @return File object
+	 * @deprecated since 2.8.0
```
dkayiwa, line 142: Our convention for deprecation is to mention the replacement: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#Deprecation

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
```diff
@@ -42,119 +32,75 @@ public class AbstractHandlerTest extends BaseContextSensitiveTest {
...
-		String uuidPartWithTitle = nameWithTitle[1];
-		String extensionPart = nameWithTitle[2];
 		
 		assertEquals(titlePart, FilenameUtils.removeExtension(FILENAME));
-		assertEquals(extensionPart, "txt");
-		assertEquals(uuidPartWithTitle, obsWithTitle.getUuid());
-	}
-	
-	@Test
-	public void getOutputFileToWrite_shouldCorrectlyNameTitledFileWithoutExtension() throws IOException, ParseException {
```
dkayiwa, line 121: Are we removing this because we changed the file naming convention?

api/src/main/java/org/openmrs/obs/handler/ImageHandler.java
```diff
@@ -131,60 +126,44 @@ public String[] getSupportedViews() {
...
-			
+			obs.setValueComplex(extension + " image |" + assignedKey);
+
 			// Remove the ComlexData from the Obs
 			obs.setComplexData(null);
-			
-		}
-		catch (IOException ioe) {
-			if (outfile != null && outfile.length() == 0) {
-				outfile.delete(); // OpenJDK 7 & 8 may leave a 0-byte file when ImageIO.write(..) fails.
```
dkayiwa, line 185: Should i assume that the above is no longer a possibility?

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -34,124 +40,148 @@ public class AbstractHandler {
...
+
+	/**
+	 * Should be replaced with just parseDataKey once migration is over.
+	 * 
+	 * @param obs complex obs
+	 * @return key
+	 * @deprecated 2.8.0
+	 * @since 2.8.0
+	 */
+	@Deprecated
```
dkayiwa, line 128: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#Deprecation

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
rkorytkowski: Correct. The outputfile is also no longer available as a file handle, just inputstream.

api/src/main/java/org/openmrs/obs/handler/ImageHandler.java
rkorytkowski: I'll add some logic and test to make sure that file is deleted if exception occurs.

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
rkorytkowski: I'll just add the logic to parseDataKey and remove this method altogether.

api/src/main/java/org/openmrs/api/storage/LocalStorageService.java
```diff
@@ -0,0 +1,248 @@
...
+	
+	@Override
+	public String saveData(InputStream inputStream, ObjectMetadata metadata, String moduleIdOrGroup, String keySuffix) throws IOException {
+		String key = newKey(moduleIdOrGroup, keySuffix, metadata != null ? metadata.getFilename() : null);
+		Path target = newPath(key);
+		try {
+			Files.copy(inputStream, target);
+		} catch (IOException e) {
+			purgeData(key);
+			throw e;
```
rkorytkowski, line 215: @dkayiwa here's the new logic for deleting file if copying failed...

rkorytkowski:
All review comments addressed :) Thanks @dkayiwa.

dkayiwa (approved):

PR #5022, opened 2025-04-27
TRUNK-6334: Define PropertySource so properties can be specified via property file or system property

mherman22:

This pull request updates the InitializationFilter to support automated setup by loading initialization properties from multiple sources, prioritizing:
- System properties (-D JVM arguments)
- Environment variables (with selective normalization)
- Installation script file (OPENMRS_INSTALLATION_SCRIPT)

## Key Changes
- Added resolveInitializationProperties() to merge properties respecting source priority.
- Environment variable normalization: underscores (_) replaced with dots (.) except for specific keys like install_method.
- Maintained full backward compatibility with existing installation script setups.
- Improved logging to show merged properties during initialization (at DEBUG level).

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6334

  No? -> execute above command

### DIFF
--- a/api/src/main/resources/applicationContext-service.xml
+++ b/api/src/main/resources/applicationContext-service.xml
@@ -678,6 +678,17 @@
 		<property name="queueCapacity" value="1000" />
 	</bean>
 
+	<bean class="org.springframework.context.support.PropertySourcesPlaceholderConfigurer">
+		<property name="locations">
+			<list>
+				<value>classpath:hibernate.default.properties</value>
+				<value>file:${OPENMRS_APPLICATION_DATA_DIRECTORY}/openmrs-runtime.properties</value>
+			</list>
+		</property>
+		<property name="ignoreResourceNotFound" value="true"/>
+		<property name="localOverride" value="true"/>
+	</bean>
+
 	<!--  ************************ END VALIDATOR CONFIGURATION  ************************  -->
 
 	<context:component-scan base-package="org.openmrs">
--- a/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
+++ b/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
@@ -22,14 +22,17 @@ import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.Base64.Encoder;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Random;
+import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.zip.ZipInputStream;
@@ -172,7 +175,7 @@ public class InitializationFilter extends StartupFilter {
 	 * this object are made available to all templates via reflection in the
 	 * {@link org.openmrs.web.filter.StartupFilter#renderTemplate(String, Map, HttpServletResponse)} method.
 	 */
-	private InitializationWizardModel wizardModel = null;
+	protected InitializationWizardModel wizardModel = null;
 	
 	private InitializationCompletion initJob;
 	
@@ -184,7 +187,10 @@ public class InitializationFilter extends StartupFilter {
 	
 	// the actual driver loaded by the DatabaseUpdater class
 	private String loadedDriverString;
-	
+
+	private static final Set<String> NON_NORMALIZED_KEYS = new HashSet<>(Arrays.asList(
+		"INSTALL_METHOD", "DATABASE_NAME", "HAS_CURRENT_OPENMRS_DATABASE", "CREATE_DATABASE_USER",
+		"CREATE_TABLES", "ADD_DEMO_DATA", "MODULE_WEB_ADMIN", "AUTO_UPDATE_DATABASE", "ADMIN_USER_PASSWORD"));
 	/**
 	 * Variable set at the end of the wizard when spring is being restarted
 	 */
@@ -204,7 +210,7 @@ public class InitializationFilter extends StartupFilter {
 	protected void doGet(HttpServletRequest httpRequest, HttpServletResponse httpResponse)
 		throws IOException {
 		SessionModelUtils.loadFromSession(httpRequest.getSession(), wizardModel);
-		loadInstallationScriptIfPresent();
+		initializeWizardFromResolvedPropertiesIfPresent();
 		
 		// we need to save current user language in references map since it will be used when template
 		// will be rendered
@@ -308,9 +314,24 @@ public class InitializationFilter extends StartupFilter {
 			renderTemplate(INSTALL_METHOD, referenceMap, httpResponse);
 		}
 	}
-	
-	private void loadInstallationScriptIfPresent() {
-		Properties script = getInstallationScript();
+
+	/**
+	 * Initializes the setup wizard model by resolving configuration properties from multiple sources.
+	 * <p>
+	 * Properties are loaded and prioritized from system properties, environment variables (normalized),
+	 * and the installation script file. Resolved values are then applied to the corresponding fields
+	 * of the {@link InitializationWizardModel}.
+	 */
+	protected void initializeWizardFromResolvedPropertiesIfPresent() {
+		Properties script = resolveInitializationProperties();
+
+		if (log.isDebugEnabled()) {
+			for (String key : script.stringPropertyNames()) {
+				String value = script.getProperty(key);
+				log.debug("{} = {}", key, key.toLowerCase().contains("password") ? "*******" : value);
+			}
+		}
+		
 		if (!script.isEmpty()) {
 			wizardModel.installMethod = script.getProperty("install_method", wizardModel.installMethod);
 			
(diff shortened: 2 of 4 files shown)


rkorytkowski:
Thanks! It's good direction though instead of System.getEnv() I'd like us to set <context:property-placeholder ... />  in applicationContext and then use https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/env/Environment.html instead of System. The benefit of that is it can resolve properties from different places including openmrs-runtime.properties, java env variables and global system variables. It also makes it very easy to inject properties with @Value annotation for example.

rkorytkowski:
Just saw your comment on talk. In such a case I'd use System in InitializationFilter, but still set property-placeholder in applicationContext to be used in other places.

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1936,14 +1957,39 @@ private static boolean goBack(HttpServletRequest httpRequest) {
...
+
+	private String normalizeEnvVariableKey(String envVarKey) {
+		if (NON_NORMALIZED_KEYS.contains(envVarKey)) {
+			return envVarKey.toLowerCase();
+		}
+		return envVarKey.toLowerCase().replace('_', '.');
+	}
+
+	protected Map<String, String> getEnvironmentVariables() {
+		return System.getenv();
```
rkorytkowski, line 1983: Are you only checking for OS env variables? We also want to combine them with System.getProperties() that are java environment variables. The order should be:
1. Java env variables
2. OS env variables
3. properties files
Meaning 1. overwrites 2 and 3.

mherman22:
> Thanks! It's good direction though instead of System.getEnv() I'd like us to set <context:property-placeholder ... /> in applicationContext and then use https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/env/Environment.html instead of System. The benefit of that is it can resolve properties from different places including openmrs-runtime.properties, java env variables and global system variables. It also makes it very easy to inject properties with `@Value("${connection.url}")` annotation for example.

Thanks for clarifying @rkorytkowski , so what should the default location be within the `<context:property-placeholder .../>`? see impl below:

```
<context:property-placeholder system-properties-mode="OVERRIDE" ignore-resource-not-found="true" location="file:${OPENMRS_INSTALLATION_SCRIPT}"/>
```

rkorytkowski:
The default should point to openmrs-runtime.properties. OpenMRS installation script is only for installation wizard, for which we cannot use property-placeholder anyway so we don't list that one there.

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -307,57 +318,79 @@ protected void doGet(HttpServletRequest httpRequest, HttpServletResponse httpRes
...
+					for (String key : ((MapPropertySource) propertySource).getPropertyNames()) {
+						String value = environment.getProperty(key);
+						log.debug("{} = {}", key, key.toLowerCase().contains("password") ? "*******" : value);
+					}
+				}
+			}
+		}
+
+		if (this.environment == null) {
+			log.warn("Spring Environment not properly initialized");
```
mherman22, line 342: i would like some guidance here, i am blocked on how to ensure the environment is not null. Since InitializationFilter is not a spring bean and the filter is declared in web.xml, not managed by Spring, then i could not inject Environment like below
```
@Autowired
private Environment environment
```

In that case, i have to get the ApplicationContext manually inside the Filter by having the snippet below within the init() method of the filter.
```
ServletContext servletContext = filterConfig.getServletContext();
XmlWebApplicationContext ctx = (XmlWebApplicationContext) WebApplicationContextUtils.getWebApplicationContext(servletContext);

this.env = ctx.getEnvironment();
```

Despite the above, environment is still not injected hence still throwing Spring `Environment not properly initialized` in the logs.

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1081,14 +1114,24 @@ protected static boolean isInstallationStarted() {
 	@Override
 	public void init(FilterConfig filterConfig) throws ServletException {
 		super.init(filterConfig);
+		ServletContext servletContext = filterConfig.getServletContext();
+		XmlWebApplicationContext ctx = (XmlWebApplicationContext) WebApplicationContextUtils.getWebApplicationContext(servletContext);
+
+		if (ctx == null) {
+			log.warn("Spring WebApplicationContext is NOT available at filter init");
```
mherman22, line 1121: this is still thrown, any pointers?

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
rkorytkowski: Don't worry about it and use System to get properties as you originally did. Let's just have Environment configured for code that runs after startup.

api/src/main/resources/applicationContext-service.xml
```diff
@@ -678,4 +678,9 @@
 		<context:exclude-filter type="custom" expression="org.openmrs.annotation.OpenmrsProfileExcludeFilter"/>
 	</context:component-scan>
 
+	<context:property-placeholder
+		location="file:${OPENMRS_APPLICATION_DATA_DIRECTORY}/openmrs-runtime.properties"
+		system-properties-mode="OVERRIDE"
+		ignore-resource-not-found="true" />
```
rkorytkowski, line 684: Could we please also include hibernate.default.properties and have a simple BaseContextSensitiveTest that uses e.g. @Value annotation or Environment to test if properties are loaded correctly?

For including extra locations with ordering you would probably need to define bean directly instead of using context:property-placeholder.

```
<bean class="org.springframework.context.support.PropertySourcesPlaceholderConfigurer">
  <property name="location">
  <list>
  <value>classpath:hibernate.default.properties</value>
  <value>file:${OPENMRS_APPLICATION_DATA_DIRECTORY}/openmrs-runtime.properties</value>
  </list>
  </property>
  ...
</bean>
```

rkorytkowski:
Thanks! Almost there! Let's just load hibernate.default.properties, which can be overwritten by openmrs-runtime.properties to have a full view of the environment and add a test.

rkorytkowski:
I don't want to debug the issue with CacheManager since I'm just changing ehcache to infinispan and the problem may go away by itself.

For now I'd just set ignoreResourceNotFound to true to go around the missing file issue without introducing the problematic initialiser. We don't mind about the file missing in tests.

Please also set localOverride to true to have the same logic as in initializer.

mherman22:
> I don't want to debug the issue with CacheManager since I'm just changing ehcache to infinispan and the problem may go away by itself.

i removed the initializer and instead added the two properties you mention above. The test passes as an individual test but when you run the entire directory or all tests, i still get the error related to this https://pastebin.com/rjN9SCbj

rkorytkowski:
Thanks! Bear with me a bit longer. I'm hoping to send a PR today that replaces ehcache with infinispan, which may fix the issue here.

mherman22:
> Just an update that my PR introducing infinispan is at #5027
> 
> There's still some issue with it running on GHA though... but I added PropertySource config as a test there and it worked so infinispan should address the issue with this PR indeed.

I wanted to checkout the code in that PR later today, git merge with this PR and test things out. I will provide feedback here or the other side.

rkorytkowski:
Please go ahead and update to the latest master to see if the issue is fixed.

mherman22:
@rkorytkowski any pointers on how to reproduce it? I have triggered `mvn clean install --batch-mode && mvn test -Pskip-default-test -Pintegration-test --batch-mode --file pom.xml` but the build is successful.

I have tested on both `temurin` and `corretto` distributions.

rkorytkowski:
It's going to be hard to reproduce locally unless you have windows VM... I'd try to look at tests and implementation and reason why it might be failing on Windows. I've re-triggered the build just to see if it's not a random failure.

rkorytkowski:
@mherman22 do you have any clues why it might be failing? Let me know if you run out of ideas. I can then look into that.

rkorytkowski:
@mherman22 all right! Getting back to this one. I've got a Windows VM so I'll try to reproduce.

web/src/test/java/org/openmrs/web/filter/initialization/InitializationFilterTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.spy;
+
+public class InitializationFilterTest extends BaseContextSensitiveTest {
```
kb019: Hi @rkorytkowski and @mherman22 , i am not sure what the root cause is but extending [BaseWebContextSensitiveTest](https://github.com/openmrs/openmrs-core/blob/2ffff14bac8763470a1cffbaf356cbadfcae4e45/web/src/test/java/org/openmrs/web/test/jupiter/BaseWebContextSensitiveTest.java#L26) instead of BaseContextSensitiveTest solved the issue when i checked locally on my windows machine.

rkorytkowski: Great catch! It should be BaseWebContextSensitiveTest indeed!

mherman22: Let me update that asap, thanks @kb019 !

rkorytkowski: Just did it! Tests are running. Fingers crossed.

mherman22: awesome!

rkorytkowski (approved):

PR #5040, opened 2025-05-28
TRUNK-6343: Add PostgreSQL support to the simple setup in the setup wizard

mherman22:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6343

  No? -> execute above command

### DIFF
--- a/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
+++ b/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
@@ -19,8 +19,6 @@ import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.sql.Connection;
 import java.sql.DriverManager;
-import java.sql.PreparedStatement;
-import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
@@ -1401,12 +1399,6 @@ public class InitializationFilter extends StartupFilter {
 							if (isCurrentDatabase(DATABASE_MYSQL)) {
 								sql = "create database if not exists `?` default character set utf8";
 							} else if (isCurrentDatabase(DATABASE_POSTGRESQL)) {
-								if (databaseExistsInPostgres(wizardModel.databaseName, wizardModel.createDatabaseUsername, 
-									wizardModel.createDatabasePassword)) {
-									reportError(String.format("Database '%s' already exists. Choose a different name or drop the existing database.",
-											wizardModel.databaseName), DEFAULT_PAGE);
-									return;
-								}
 								sql = "create database `?` encoding 'utf8'";
 							} else if (isCurrentDatabase(DATABASE_H2)) {
 								sql = null;
@@ -1472,11 +1464,6 @@ public class InitializationFilter extends StartupFilter {
 							if (isCurrentDatabase(DATABASE_MYSQL)) {
 								sql = "create user '?'@" + host + " identified by '?'";
 							} else if (isCurrentDatabase(DATABASE_POSTGRESQL)) {
-								if (userExistsInPostgres(connectionUsername, wizardModel.createDatabaseUsername, 
-									wizardModel.createDatabasePassword)) {
-									reportError(String.format("A database user named '%s' already exists.", connectionUsername), DEFAULT_PAGE);
-									return;
-								}
 								sql = "create user `?` with password '?'";
 							}
 							
@@ -2030,34 +2017,6 @@ public class InitializationFilter extends StartupFilter {
 			}
 		}
 		return prop;
-	}
-
-	private boolean databaseExistsInPostgres(String dbName, String user, String password) {
-		String query = "SELECT 1 FROM pg_database WHERE datname = ?";
-		try (Connection conn = DriverManager.getConnection(DEFAULT_POSTGRESQL_CONNECTION, user, password);
-			 PreparedStatement stmt = conn.prepareStatement(query)) {
-			stmt.setString(1, dbName);
-			try (ResultSet rs = stmt.executeQuery()) {
-				return rs.next();
-			}
-		} catch (SQLException e) {
-			log.warn("Failed to check if PostgreSQL DB exists: {}", dbName, e);
-			return false;
-		}
-	}
-
-	private boolean userExistsInPostgres(String usernameToCheck, String adminUser, String adminPassword) {
-		String query = "SELECT 1 FROM pg_roles WHERE rolname = ?";
-		try (Connection conn = DriverManager.getConnection(DEFAULT_POSTGRESQL_CONNECTION, adminUser, adminPassword);
-			 PreparedStatement stmt = conn.prepareStatement(query)) {
-			stmt.setString(1, usernameToCheck);
-			try (ResultSet rs = stmt.executeQuery()) {
-				return rs.next();
-			}
-		} catch (SQLException e) {
-			log.warn("Failed to check if PostgreSQL user exists: {}", usernameToCheck, e);
-			return false;
-		}
 	}
 
 }


mherman22:
> Can we break this down into smaller and simpler parts? For instance, if i used the advanced installation method, can't i install on postgresql? If not, can we start with only the changes required to have that work?

Actually these changes are for the simple setup only. 

(Except for the databaseuser.vm file which I am gonna remove)

mherman22:
> The last time i tried using the advanced installation for PostgreSQL database setup, it worked well. So if it is broken, i would expect very minimal changes to make it work. Simple setup involves a lot more work because it is a completely new feature for PostgreSQL.

Thanks for the heads up. Armed with that information, I think this should still be in draft.

dkayiwa:
@mherman22 do we need all these changes to just get the setup working on postgres? For instance i would not expect any changes to the user interface for using the advanced installation method, because all i would need to do is provide postgres specific parameters. So the only errors i would expect would be something like liquibase changesets that work on mysql but not postgres. I am not against any user interface improvements for postgres. But i would like us to start with the smallest possible changes to just have postgres working with the existing interface. We can do the improvements with a separate ticket and pull request. In a otherwords, smaller pull request focusing on a specific thing, are easier for me to review and quickly merge. 😊 

https://opensource.com/article/18/6/anatomy-perfect-pull-request

mherman22:
> @mherman22 do we need all these changes to just get the setup working on postgres? For instance i would not expect any changes to the user interface for using the advanced installation method, because all i would need to do is provide postgres specific parameters. So the only errors i would expect would be something like liquibase changesets that work on mysql but not postgres. I am not against any user interface improvements for postgres. But i would like us to start with the smallest possible changes to just have postgres working with the existing interface. We can do the improvements with a separate ticket and pull request. In a otherwords, smaller pull request focusing on a specific thing, are easier for me to review and quickly merge. 😊
> 
> https://opensource.com/article/18/6/anatomy-perfect-pull-request

Makes perfect sense and I completely agree. I will modifications to the pull request about advanced step. And only do changes to the initializationFilter to accept a db connection string from postgres.

dkayiwa:
It already accepts it without requiring any changes.

mherman22:
> It already accepts it without requiring any changes.

True dat!

Is this particular PR fine? or i have to break it down aswell? With only changes to the InitializationFilter and fixing postgres <-> mysql related errors? and then a separate PR for the UI?

dkayiwa:
The answer is here: https://opensource.com/article/18/6/anatomy-perfect-pull-request

mherman22:
> The answer is here: https://opensource.com/article/18/6/anatomy-perfect-pull-request

I have done the needful

dkayiwa:
@mherman22 let me simplify it this way, when you run openmrs and supply posgres parameters in the Advanced Installation wizard, what error do you get?

mherman22:
> @mherman22 let me simplify it this way, when you run openmrs and supply posgres parameters in the Advanced Installation wizard, what error do you get?

I get `org.postgresql.util.PSQLException: ERROR: role "openmrs_user" already exists` in the terminal and `Unable to create a database user` on the UI. which i fixed by dropping the user and tried again.

When i choose that i have an existing openmrs db, it goes ahead to completely successfully.

When i choose no and provide the username and password, `org.postgresql.util.PSQLException: FATAL: database "mac" does not exist`

dkayiwa:
For a brand new setup, if you specified the correct jdbc url for posgresql and driver, i would not expect you to get any errors apart from the one from some liquibase changesets. Can you cross check to confirm that you provided the correct input? an openmrs_user and database should be successfully created by postgres but only fail at the time of running some database changesets.

mherman22:
> For a brand new setup, if you specified the correct jdbc url for posgresql and driver, i would not expect you to get any errors apart from the one from some liquibase changesets. Can you cross check to confirm that you provided the correct input? an openmrs_user and database should be successfully created by postgres but only fail at the time of running some database changesets.

i have specified `jdbc:postgresql://localhost:5432/@DBNAME@?autoReconnect=true&useUnicode=true&characterEncoding=UTF-8`, driver as `org.postgresql.Driver` (no need to specify it anyway since `loadDatabaseDriver` will map the needed driver based on the jdbc url). But i still fail with `org.postgresql.util.PSQLException: FATAL: database "mac" does not exist`

I have uninstalled, reinstalled postgresql a couple of times now but it seems to fail at that point.

mherman22:
UPDATED: 

I have changed the jdbc url to `jdbc:postgresql://localhost:5432/postgres` and i now get the liquibase errors you talked about.

Let me modify this pull request to fix those errors because i faced them earlier and some of my changes here are the fix, let me get rid of the unnecessary stuff.

Thanks @dkayiwa

dkayiwa:
No user interface changes for postgresql simple setup?

mherman22:
Let me commit the UI changes

dkayiwa:
Shouldn't the default postgres user name be postgres?

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -482,16 +484,24 @@ else if (SIMPLE_SETUP.equals(page)) {
...
+				wizardModel.databaseType = databaseType;
+				if (DATABASE_MYSQL.equals(databaseType)) {
+					wizardModel.databaseConnection = DEFAULT_MYSQL_CONNECTION;
+					wizardModel.createDatabaseUsername = Context.getRuntimeProperties().getProperty("connection.username", wizardModel.createDatabaseUsername);
+					wizardModel.databaseRootPassword = httpRequest.getParameter("database_root_password");
+					checkForEmptyValue(wizardModel.databaseRootPassword, errors, ErrorMessageConstants.ERROR_DB_PSDW_REQ);
+				} else if (DATABASE_POSTGRESQL.equals(databaseType)) {
+					wizardModel.databaseConnection = DEFAULT_POSTGRESQL_CONNECTION;
+					wizardModel.databaseRootPassword = httpRequest.getParameter("database_root_password");
+					wizardModel.createDatabaseUsername = httpRequest.getParameter("postgres_superuser");
```
dkayiwa: Why are you calling it `postgres_superuser` and yet for mysql you call it `database_root_password`?

api/src/main/resources/messages.properties
```diff
@@ -2904,6 +2908,15 @@ install.simple.empty=(Empty MySQL passwords are not allowed here. If you don't h
 install.simple.passwd=Password
 install.simple.defpasswd=Default OpenMRS administrator username/password. Change this password after login in.<b>(Remember this password.)</b>
 install.simple.addDemoData=Do you want to also add demo data to your database
+install.simple.databaseType=Select Database Type
+install.simple.mysql.passwdinfo=Enter MySQL root password
```
dkayiwa: Can we use the full word in the above for password?

api/src/main/resources/messages_de.properties
```diff
@@ -3257,3 +3257,8 @@ uncaughtException.additionalInfo.modules=Namen und Versionen der installierten M
 
 OrderFrequency.concept.shouldBeClassFrequency=Konzept sollte von der Klasse 'Frequency' sein.
 OrderFrequency.concept.shouldNotBeShared=Nur eine Bestellfrequenz sollte einem bestimmten Begriff zugeordnet sein.
+
+install.simple.postgresql.auth.info=PostgreSQL-Authentifizierung
```
dkayiwa: Did you use google or AI to translate these? If yes, we do not normally recommend doing so. Just leave them out for those who speak the language to later on do the translations.

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -482,16 +484,24 @@ else if (SIMPLE_SETUP.equals(page)) {
 				renderTemplate(INSTALL_METHOD, referenceMap, httpResponse);
 				return;
 			}
-			wizardModel.databaseConnection = httpRequest.getParameter("database_connection");
-
-			wizardModel.createDatabaseUsername = Context.getRuntimeProperties().getProperty("connection.username",
-				wizardModel.createDatabaseUsername);
 			
+			String databaseType = httpRequest.getParameter("database_type");
```
dkayiwa, line 488: Why not directly assign it to wizardModel.databaseType?

api/src/main/resources/messages_de.properties
mherman22: makes sense

mherman22:
> Shouldn't the default postgres user name be postgres?

It is because for environments like macos, no default `postgres` user is created. The current macOS user is given superuser privileges. for my case, it is `mac`

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
mherman22: intention was to add a null check thereafter

api/src/main/resources/messages.properties
mherman22: true dat!

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
mherman22: mysql by default sets a username as `root` so there is no need to attempt to fetch it hence the reason why there is no username field for the mysql option. 

but i will rename `postgres_superuser` to `create_database_username`

dkayiwa:
> It is because for environments like macos, no default `postgres` user is created. The current macOS user is given superuser privileges. for my case, it is `mac`

Between root and postgres, which one has more chances of being the default?

mherman22:
> > It is because for environments like macos, no default `postgres` user is created. The current macOS user is given superuser privileges. for my case, it is `mac`
> 
> Between root and postgres, which one has more chances of being the default?

postgres

dkayiwa:
@mherman22 are you done with this?

mherman22:
> @mherman22 are you done with this?

Yes

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -482,16 +484,25 @@ else if (SIMPLE_SETUP.equals(page)) {
...
+			if (databaseType != null) {
+				wizardModel.databaseType = databaseType;
+				if (DATABASE_MYSQL.equals(databaseType)) {
+					wizardModel.databaseConnection = DEFAULT_MYSQL_CONNECTION;
+					wizardModel.createDatabaseUsername = Context.getRuntimeProperties().getProperty("connection.username", wizardModel.createDatabaseUsername);
+					wizardModel.databaseRootPassword = httpRequest.getParameter("database_root_password");
```
dkayiwa: Why are we doing the above line separately for mysql and postgresql?

mherman22:
One thing i have encountered is when i already have a database named `openmrs` in my postgres dbms and i attempt to run `mvn jetty:run` for whatever reason, it fails with `???Error executing sql: create database `?` encoding 'utf8' - ERROR: database "openmrs" already exists???`. 

To fix it, i added the clean process description at https://openmrs.atlassian.net/wiki/spaces/docs/pages/25471919/PostgreSQL+Support+in+the+OpenMRS+Setup+Wizard#Cleanup-Procedures, i was wondering if it makes sense to have it automated here?

cc: @dkayiwa

dkayiwa:
> i was wondering if it makes sense to have it automated here?

We need to have the same behaviour as for mysql.

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -482,16 +484,24 @@ else if (SIMPLE_SETUP.equals(page)) {
...
+			if (databaseType != null) {
+				wizardModel.databaseType = databaseType;
+				if (DATABASE_MYSQL.equals(databaseType)) {
+					wizardModel.databaseConnection = DEFAULT_MYSQL_CONNECTION;
+					wizardModel.createDatabaseUsername = Context.getRuntimeProperties().getProperty("connection.username", wizardModel.createDatabaseUsername);
+					checkForEmptyValue(wizardModel.databaseRootPassword, errors, ErrorMessageConstants.ERROR_DB_PSDW_REQ);
+				} else if (DATABASE_POSTGRESQL.equals(databaseType)) {
+					wizardModel.databaseConnection = DEFAULT_POSTGRESQL_CONNECTION;
+					String postgresUsername = httpRequest.getParameter("create_database_username");
+					wizardModel.createDatabaseUsername = StringUtils.hasText(postgresUsername) ?
```
dkayiwa, line 494: Is there a reason why you also do not do this for mysql?

mherman22: Just intended to leave mysql functionality not tampered with in this pull request that is all about postgres

dkayiwa: Fair enough!

web/src/main/java/org/openmrs/web/filter/initialization/InitializationWizardModel.java
```diff
@@ -81,6 +81,21 @@ public class InitializationWizardModel {
 	
 	public String installMethod = INSTALL_METHOD_SIMPLE;
 	
+	/**
+	 * The type of database being used (mysql or postgresql)
+	 */
+	public String databaseType = "mysql";
```
dkayiwa, line 87: Should this be a constant?

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
dkayiwa: looking at it again, if you did not want to touch mysql, then you would just do only postgresql changes for the database type.

web/src/main/java/org/openmrs/web/filter/initialization/InitializationWizardModel.java
mherman22: I don't this so, the value changes when a user selects PostgreSQL

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
mherman22: noted!

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1453,7 +1471,12 @@ public void run() {
 							if (isCurrentDatabase(DATABASE_MYSQL)) {
 								sql = "create user '?'@" + host + " identified by '?'";
 							} else if (isCurrentDatabase(DATABASE_POSTGRESQL)) {
-								sql = "create user `?` with password '?'";
+								if (userExistsInPostgres(connectionUsername, wizardModel.createDatabaseUsername,
```
dkayiwa: Why don't you just let this fail instead of silently changing the user's password which is very dangerous?

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1388,7 +1401,12 @@ public void run() {
 							if (isCurrentDatabase(DATABASE_MYSQL)) {
 								sql = "create database if not exists `?` default character set utf8";
 							} else if (isCurrentDatabase(DATABASE_POSTGRESQL)) {
-								sql = "create database `?` encoding 'utf8'";
+								if (databaseExistsInPostgres(wizardModel.databaseName, wizardModel.createDatabaseUsername,
```
dkayiwa: If the same database exists, we should just fail like we do for mysql.

dkayiwa: @mherman22 did you see the above comment?

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
dkayiwa: @mherman22 did you see the above comment?

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
mherman22: In my most recent commit i fixed it. see: https://github.com/mherman22/openmrs-core/blob/a24f70af9eed835fa340681c300261bfada7ff77/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java#L1404-L1407

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
mherman22: In my most recent commit i fixed it. see:https://github.com/mherman22/openmrs-core/blob/a24f70af9eed835fa340681c300261bfada7ff77/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java#L1475-L1478

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
dkayiwa: Why are we checking for user exists and database existing for postgresql and yet we did not have these methods for mysql?

mherman22: i thought we would catch this early and inform the user rather than breaking the program. But i can remove the checks for consistency

dkayiwa: To simplify this work, i want us to start with the exact same functionality that we have in mysql. After which you will be free to suggest improvements. But then they will need to be implemented in both mysql and postgresql

mherman22: Makes sense, let me remove that!

mherman22: just one thing to put out there is that mysql behaves differently from postgres. mysql allows something like `"create database if not exists `?` default character set utf8";` unlike postgres.

So in this case, the same behaviour with mysql would mean keeping this `databaseExistsInPostgres` check. 

They only share similar behaviour for an existing user.

PR #5049, opened 2025-05-31
TRUNK-5852: Provider Domain - Switching from Hibernate Mappings to Annotations

kb019:
TRUNK-5852: Provider Domain - Switching from Hibernate Mappings to Annotations
Add Jpa annotations to Provider class

## Issue I worked on
see https://openmrs.atlassian.net/browse/TRUNK-5852

- [] I have **added tests** to cover my changes. (If you refactored


api/src/main/java/org/openmrs/Provider.java
```diff
@@ -18,19 +32,40 @@
  *
  * @since 1.9
  */
+@Entity
+@Table(name = "provider")
 @Audited
-public class Provider extends BaseCustomizableMetadata<ProviderAttribute> {
+@AttributeOverride(name = "name", column = @Column(name = "name"))
+@AssociationOverride(name="attributes",
+	joinColumns=@JoinColumn(name="provider_id",insertable = false,updatable = false))
```
kb019, line 40: Made insertable and updatable false so that responsibility of creating/updating the referenced column's entity isn't in the current entity, but in another entity. This is because inverse="true" parameter was set on Provider.hb,.xml.
For more details see the following links
https://stackoverflow.com/questions/3805584/please-explain-about-insertable-false-and-updatable-false-in-reference-to-the-jp
https://docs.jboss.org/hibernate/envers/3.6/reference/en-US/html/exceptions.html

mherman22: How about this stackoverflow https://stackoverflow.com/questions/4865285/inverse-true-in-jpa-annotations?

kb019: Hi @mherman22 , i didnt find a way to override one to many annotation with mapped by attribute,hence took this approach.

api/src/main/java/org/openmrs/Provider.java
```diff
@@ -18,19 +32,40 @@
...
+public class Provider extends BaseCustomizableMetadata<ProviderAttribute>{
 	
 	private static final Logger log = LoggerFactory.getLogger(Provider.class);
 	
+	@Id
+	@GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "provider_id_seq")
```
dkayiwa, line 46: Is there any reason why you did not use GenerationType.IDENTITY?

dkayiwa:
Did you see the build failures?

api/src/main/java/org/openmrs/Provider.java
kb019: Hi @dkayiwa , In provider.hbm.xml sequence generator was used [here](https://github.com/openmrs/openmrs-core/blob/bcbbce41be3a516eea4bbb54cf29b0307e4653aa/api/src/main/resources/org/openmrs/api/db/hibernate/Provider.hbm.xml#L24) . Hence used GenerationType.SEQUENCE

kb019:
> Did you see the build failures?

Yes @dkayiwa , its builds successfully in my machine.Not sure whats the problem 
image

dkayiwa:
Are you running the same build command?

kb019:
Hi @dkayiwa , i dont find any methods to ignore the description field from the super class (which is why we are getting the error in build as provider table does not include the description column), other than rewriting all the fields in provider class. Would you recommend to do like this?

dkayiwa:
Wait a minute, but you said that it builds successfully on your machine, even without this field.

kb019:
> Wait a minute, but you said that it builds successfully on your machine, even without this field.

when i run mvn clean install, all test cases pass.But when i run the integration test, the SchemaValidation Fails.

api/src/main/java/org/openmrs/Provider.java
```diff
@@ -18,19 +32,40 @@
...
+	@GenericGenerator(
+		name = "provider_id_seq",
+		strategy = "native",
+		parameters = @Parameter(name = "sequence", value = "provider_provider_id_seq")
+	)
+	@Column(name = "provider_id", nullable = false,insertable = false)
```
dkayiwa: Can you put space after `false`?

api/src/main/java/org/openmrs/Provider.java
```diff
@@ -18,19 +32,40 @@
...
 	private Integer providerId;
 	
+	@ManyToOne
+	@JoinColumn(name="person_id")
+	@Cascade(CascadeType.SAVE_UPDATE)
 	private Person person;
 	
+	@Column(name="identifier")
```
dkayiwa, line 60: Did you take care of all the attributes in the xml mapping file?

kb019: Hi @dkayiwa , the default length is 255 hence did not put that value here

dkayiwa: Do you mind sharing where you got that from? 😊

kb019: Hi @dkayiwa , here is the link https://docs.jboss.org/hibernate/jpa/2.1/api/javax/persistence/Column.html

dkayiwa:
Did you address all review comments?

kb019:
> Did you address all review comments?

Hi @dkayiwa , added the space

api/src/main/java/org/openmrs/Provider.java
dkayiwa: How about the unsaved-value?

kb019: Hi @dkayiwa , made the changes to reflect unsaved value.
Reference -> https://stackoverflow.com/questions/35014889/how-to-prevent-hibernate-from-calling-default-constructor-when-resolving-the-uns

dkayiwa:
Can you take a look at the build failure?

dkayiwa:
@kb019 are you still working on this?

kb019:
> @kb019 are you still working on this?

No @dkayiwa , i am not working on this. I have unassigned myself from the ticket.

PR #5075, opened 2025-06-12
TRUNK-6356: Concept reference ranges should be taken into account

ibacher:

This changes most of the ObsValidator machinery to take account of the reference ranges defined on the concept itself as well as those that match person-specific criteria. The concept-defined reference range is only used in cases where no reference range is defined for a concept or no defined reference range matches for a specific patient.

I've slightly adjusted things so that we only try to valid concept reference ranges for numeric concepts. While there's probably something to be said for being able to define rule-based restrictions on non-numeric obs, ReferenceRanges remain a numeric-only concept.

This is targeted at 2.7.x, but will be forward-ported to 2.8.x.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6356

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/validator/ObsValidator.java
+++ b/api/src/main/java/org/openmrs/validator/ObsValidator.java
@@ -352,7 +352,9 @@ public class ObsValidator implements Validator {
 
 	/**
 	 * Finds the strictest {@link ConceptReferenceRange} from a list of valid ranges.
-	 * The strictest range is determined by having the highest lower bound and the lowest upper bound.
+	 * The strictest range is determined separately for each value, e.g., the lowAbsolute will
+	 * be the highest lowAbsolute of any matching range, the lowCritical value will be the
+	 * highest lowCritical value of any matching range.
 	 * e.g.
 	 * If ConceptReferenceRange-1 has a range of 80-150.
 	 * and ConceptReferenceRange-2 has a range of 60-140,
--- a/api/src/main/java/org/openmrs/validator/ObsValidator.java
+++ b/api/src/main/java/org/openmrs/validator/ObsValidator.java
@@ -307,14 +307,14 @@ public class ObsValidator implements Validator {
 				validRanges.add(referenceRange);
 			}
 		}
-		
-		ConceptReferenceRange defaultReferenceRange = getDefaultReferenceRange(conceptNumeric);
-		if (defaultReferenceRange != null) {
-			validRanges.add(defaultReferenceRange);
-		}
 
 		if (validRanges.isEmpty()) {
-			return null;
+			ConceptReferenceRange defaultReferenceRange = getDefaultReferenceRange(conceptNumeric);
+			if (defaultReferenceRange != null) {
+				validRanges.add(defaultReferenceRange);
+			} else {
+				return null;
+			}
 		}
 		
 		return findStrictestReferenceRange(validRanges);
--- a/api/src/main/java/org/openmrs/validator/ObsValidator.java
+++ b/api/src/main/java/org/openmrs/validator/ObsValidator.java
@@ -10,6 +10,7 @@
 package org.openmrs.validator;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


ibacher:
@dkayiwa I'd appreciate if you can look over whether or not the basic logic here makes sense. We can maybe discuss on the platform call, but I'd hope to get this in slightly sooner as its currently breaking a feature in the DRC project.

api/src/main/java/org/openmrs/validator/ObsValidator.java
```diff
@@ -319,11 +309,47 @@ public ConceptReferenceRange getReferenceRange(Obs obs) {
...
+		defaultReferenceRange.setHiNormal(conceptNumeric.getHiNormal());
+		defaultReferenceRange.setLowAbsolute(conceptNumeric.getLowAbsolute());
+		defaultReferenceRange.setLowCritical(conceptNumeric.getLowCritical());
+		defaultReferenceRange.setLowNormal(conceptNumeric.getLowNormal());
+		return defaultReferenceRange;
+	}
+
 	/**
 	 * Finds the strictest {@link ConceptReferenceRange} from a list of valid ranges.
 	 * The strictest range is determined by having the highest lower bound and the lowest upper bound.
```
dkayiwa: Since this is changing the definition of strictest range, should't the above comment also change to reflect it?

dkayiwa:
It would also be awesome if we had a test to reproduce this bug. 😊

ibacher:
Alright, I've added two tests for the two aspects of these changes:

1. Covers the case of running against a concept with no `ConceptReferenceRange` entities associated with it
2. Is more or less the same as an existing test but asserts the values of all elements of the range, which shows that they are individually derived from two `ConceptReferenceRange`s

Both of these tests would fail without these changes.

dkayiwa:
Did you notice that `shouldSetObsReferenceRangeValuesToConceptReferenceRangeValuesIfNoRuleBasedRangesArePresent` still passes even without these changes?

ibacher:
> Did you notice that shouldSetObsReferenceRangeValuesToConceptReferenceRangeValuesIfNoRuleBasedRangesArePresent still passes even without these changes?

Indeed it does. I think I'll rework the test to test what I actually want which is the call to `getReferenceRange(Observation)`.

api/src/test/java/org/openmrs/validator/ObsValidatorTest.java
```diff
@@ -641,6 +659,63 @@ public void shouldSetObsReferenceRangeValuesIfConceptReferenceRangeIsNullAndConc
...
+	 */
+	@Test
+	public void shouldSetObsReferenceRangeValuesToConceptReferenceRangeValuesIfNoRuleBasedRangesArePresent() {
+		Person person = new Person(1);
+		calendar.add(Calendar.YEAR, -600);
+		person.setBirthdate(calendar.getTime());
+
+		Obs obs = new Obs();
+		obs.setPerson(person);
+		obs.setConcept(Context.getConceptService().getConcept(5497));
```
dkayiwa, line 702: Did you rework this too? It still passes without the changes.

ibacher: Only to change the concept used.

So the previous implementation had a [couple of checks](https://github.com/openmrs/openmrs-core/blob/c3b3852cae8101fe00f73d3a7bb98a3e90962632/api/src/main/java/org/openmrs/validator/ObsValidator.java#L281) to set the Obs's reference range to the default if no matching reference range was found. The problem I was actually trying to fix is `getReferenceRange()` returning null if none of the rule-based ranges matched, since that's what we're returning via the API.

I'm leaving this test in because although it passes, it tests something the previous tests didn't, namely that we still get the concept's reference range even if there are no ConceptReferenceRange objects for that concept.

ibacher: I did, however, add more tests to the ConceptService and testing the `getReferenceRange()` function directly which should fail on the previous version.

dkayiwa: Makes perfect sense! 👍

dkayiwa (approved):

PR #5093, opened 2025-06-24
Changed the Person and Patient classes to hibernate annotations

900hacked:
Changing from XML to hibernate Annotations
TRUNK-5492, Replace Person and Patient hbm mapping file with annotations

I Changed the Patients and Persons classes from hbm mapping to hibernate Annotations


api/src/main/java/org/openmrs/Patient.java
```diff
@@ -9,13 +9,8 @@
...
 
-import javax.persistence.Cacheable;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Set;
-import java.util.TreeSet;
+import javax.persistence.*;
+import java.util.*;
```
mherman22: Also we don't recommend use of wildcards, take a look at https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#Do-not-use-wildcard-imports

api/src/main/java/org/openmrs/Person.java
```diff
@@ -9,8 +9,7 @@
  */
 package org.openmrs;
 
-import javax.persistence.Cacheable;
-import javax.persistence.Transient;
+import javax.persistence.*;
```
mherman22, line 12: We don't recommend use of wildcards, take a look at https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#Do-not-use-wildcard-imports

PR #5099, opened 2025-06-30
TRUNK-6351: Added Jackson Serializer/Deserializer

Ruhanga:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6351


api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
```diff
@@ -0,0 +1,184 @@
...
+	 *  @throws RuntimeException if serialization fails
+	 */
+	@Override
+	public String serialize(Object o) {
+		String className = o.getClass().getName();
+
+		try {
+			return getObjectMapper().writeValueAsString(o);
+		} catch (JsonProcessingException e) {
+			throw new RuntimeException("Unable to deserialize class: " + className, e);
```
suubi-joshua: @Ruhanga shouldn't this be `throw new RuntimeException("Unable to serialize class: " + className, e);`
 since this is the `serialize()` method.

api/src/test/java/org/openmrs/serialization/JacksonSerializerTest.java
```diff
@@ -0,0 +1,172 @@
...
+        String serializedEnc = serializer.serialize(Context.getEncounterService().getEncounter(3));
+
+        // verify
+        assertEquals(StringUtils.deleteWhitespace(serializedEnc),
+            StringUtils.deleteWhitespace("{\"uuid\":\"6519d653-393b-4118-9c83-a3715b82d4ac\",\"creator\":\"1010d442-e134-11de-babe-001e378eb67e\",\"dateCreated\":\"2008-08-18T14:09:05\",\"changedBy\":null,\"dateChanged\":null,\"voided\":false,\"dateVoided\":null,\"voidedBy\":null,\"voidReason\":null,\"encounterId\":3,\"encounterDatetime\":\"2008-08-01T00:00:00\",\"patient\":\"5946f880-b197-400b-9caa-a3c661d23041\",\"location\":\"8d6c993e-c2cc-11de-8d13-0010c6dffd0f\",\"form\":\"d9218f76-6c39-45f4-8efa-4c5c6c199f50\",\"encounterType\":\"07000be2-26b6-4cce-8b40-866d8435b613\",\"orders\":null,\"diagnoses\":null,\"conditions\":null,\"obs\":null,\"visit\":\"a2428fea-6b78-11e0-93c3-18a905e044dc\",\"encounterProviders\":null,\"allergies\":null}"));
+
+    }
+
+    @Test
+    public void deserialize_shouldDeserializeJsonStringToOpenmrsObjectClassInstance() throws Exception {
```
suubi-joshua, line 110: Correctly if am wrong @Ruhanga there might be test Data Inconsistency, the test shows mixed behavior - some collections are null, others are empty arrays. The UUID reference system might not handle collection initialization consistently.

In the first test, the expected output has "`conditions":null,"obs":null,"allergies":null`, but in the second test, the input string has "`conditions":[],"obs":[...],"allergies":[]`.
### Why This Indicates a Problem
1. Serialization produces `null` for collections that don't exist or are empty
2. Deserialization expects empty arrays `[]` for some of the same collections
3. This suggests the UUID reference system handles collection initialization differently depending on:
Whether the object is being serialized from the database
Whether the object is being reconstructed from JSON
The specific collection type or lazy-loading behavior
### Impact on UUID Reference System
This inconsistency could cause issues when:
Round-trip serialization/deserialization doesn't preserve the original state
Code expects either null or empty collections but gets the other
Collection initialization logic differs between persistence and JSON reconstruction contexts

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
Ruhanga: Good catch—thanks, @suubi-joshua! I’ll go ahead and fix that oversight.

api/src/test/java/org/openmrs/serialization/JacksonSerializerTest.java
Ruhanga: You're raising a valid concern, @suubi-joshua, and it's worth clarifying.

Jackson itself is quite resilient when it comes to handling null vs empty collections — it can safely (de)serialize both. However, what you see as inconsistency in test data (e.g., some collections being null and others [], may sound confusing) is actually helping prove capability of Jackson.

That said, this isn't necessarily a problem with the UUID reference system itself. The differences you're seeing is stemming from:

- How the test data is created: Manually built Java objects (possibly not initializing empty collections) vs deserialized JSON (which may default to []).
- Persistence behaviour: Hibernate might leave some collections as null when not initialized, whereas JSON input explicitly declares them as empty arrays.

Ruhanga: @suubi-joshua, so I've been prompted to reconfigure Jackson for finer handling of (set/get)-able collection properties.

api/src/main/java/org/openmrs/serialization/DomainService.java
```diff
@@ -0,0 +1,176 @@
...
+    private final Map<Class<?>, DomainFetcher> domainFetchers = new HashMap<>();
+
+    /**
+     * Constructs the {@code DomainService} and scans the provided OpenMRS services
+     * for eligible {@code getXByUuid(String)} methods to populate the internal lookup.
+     *
+     * @param services the list of all available {@code OpenmrsService} implementations
+     */
+    @Autowired
+    public DomainService(List<OpenmrsService> services) {
```
ibacher: I think this is actually going to do the wrong thing. The problem here is that we're heavily using the `TransactionProxyFactoryBean` to wrap all of the service classes so that the "real" `adminService` actually extends `TransactionProxyFactoryBean` and proxies and `OpenmrsService` implementation. This basically means that these calls to `getXXXByUuid()` will by-pass transaction management and that can lead to weird things.

Rather than creating a new service like this, is there a reason that we can't extend the existing `ServiceContext` to do this work?

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
```diff
@@ -0,0 +1,186 @@
...
+				String baseClassName = pattern.substring("hierarchyOf:".length());
+				Class<?> baseClass = Class.forName(baseClassName);
+				Class<?> actualClass = Class.forName(className);
+				return baseClass.isAssignableFrom(actualClass);
+			} else {
+				// Convert dot-style wildcard pattern to regex
+				String regex = pattern
+					.replace(".", "\\.")
+					.replace("**", ".+")
+					.replace("*", "[^.]+");
```
ibacher:
```suggestion
					.replace("*", "[^.]*");
```

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
```diff
@@ -0,0 +1,186 @@
...
+			if (pattern.startsWith("hierarchyOf:")) {
+				String baseClassName = pattern.substring("hierarchyOf:".length());
+				Class<?> baseClass = Class.forName(baseClassName);
+				Class<?> actualClass = Class.forName(className);
+				return baseClass.isAssignableFrom(actualClass);
+			} else {
+				// Convert dot-style wildcard pattern to regex
+				String regex = pattern
+					.replace(".", "\\.")
+					.replace("**", ".+")
```
ibacher:
```suggestion
					.replace("**", ".*")
```

api/src/test/java/org/openmrs/serialization/JacksonSerializerTest.java
```diff
@@ -0,0 +1,179 @@
...
+import org.openmrs.api.AdministrationService;
+import org.openmrs.api.context.Context;
+import org.openmrs.test.jupiter.BaseContextSensitiveTest;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class JacksonSerializerTest extends BaseContextSensitiveTest {
```
dkayiwa, line 37: What do you think of having an equivalent for each test in this class? https://github.com/openmrs/openmrs-core/blob/2.7.4/api/src/test/java/org/openmrs/serialization/SimpleXStreamSerializerTest.java

Ruhanga: I should be able to add but a few 1 or 2 relevant tests as others, not added already, are unrelated to the JSON data structure.

api/src/main/java/org/openmrs/api/LocationService.java
```diff
@@ -383,7 +383,7 @@ public List<Location> getLocations(String nameFragment, Location parent,
 	public List<String> getPossibleAddressValues(Address incomplete, String fieldName) throws APIException;
 	
 	/**
-	 * Returns the xml of default address template.
+	 * Returns the of default address template string.
```
mseaton: This doesn't read right.  Also, `xml` should not be changing as a result of this.

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -742,7 +750,7 @@ public static final List<GlobalProperty> CORE_GLOBAL_PROPERTIES() {
 		props.add(new GlobalProperty(ModuleConstants.REPOSITORY_FOLDER_PROPERTY,
 		        ModuleConstants.REPOSITORY_FOLDER_PROPERTY_DEFAULT, "Name of the folder in which to store the modules"));
 		props.add(new GlobalProperty(GLOBAL_PROPERTY_ADDRESS_TEMPLATE, DEFAULT_ADDRESS_TEMPLATE,
-		        "XML description of address formats"));
+		        "JSON description of address formats"));
```
mseaton: I don't think this is correct - we are not changing from xml to json, we are changing the library from xstream to jackson.  Jackson can handle xml.  This PR / change should _not_ introduce any backwards-incompatibilities, it should simply be swapping out the library and making the Jackson xml parsing do what the xstream parsing was previously doing.

mseaton (changes requested):

api/src/main/java/org/openmrs/api/LocationService.java
Ruhanga: Thanks for the review @mseaton. I've fixed this.

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
Ruhanga: Thanks for raising this, @mseaton. I completely understand the concerns around backward incompatibilities, and you're right—this change does introduce one. Jackson’s XML handling still differs quite significantly from XStream’s, meaning that existing XML would still need to be transformed to match Jackson’s expectations. Given that, adopting JSON—where Jackson is strongest—doesn't add any much extra incompatibility complexity and actually offers a more modern and intuitive format in the long run.

ibacher: At the very least, it seems we need some kind of step that would convert an existing implementations address template GP into the JSON format, right?

api/src/main/java/org/openmrs/serialization/DomainService.java
Ruhanga: Thanks for your review @ibacher. I'd intended that this lean towards a simple utility than a service, in anticipation of some form of a cyclic dependency if I auto-wired services the way I've done it so far. But then I see your point. I'll attempt your mentioned approach and see how it goes.

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
Ruhanga: Correct @ibacher. This was mentioned in the ticket description.

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
```diff
@@ -0,0 +1,185 @@
...
+ * </p>
+ *
+ * This is particularly useful for deserialization scenarios where a UUID string
+ * needs to be resolved into a fully initialized domain object (e.g., during JSON parsing).
+ *
+ * @see DomainService
+ * @see UuidReferenceModule
+ */
+@Transactional
+@Component("domainServiceTarget")
```
mseaton: We don't typically wire in services this way, rather we use xml for services - mainly because doing this and autowiring in properties seems to cause a massive performance issue at startup.  @ibacher / @dkayiwa - I assume this is still an issue?

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
```diff
@@ -0,0 +1,185 @@
...
+
+    private final Map<Class<?>, DomainFetcher> domainFetchers = new HashMap<>();
+
+    /**
+     * Constructs the {@code DomainService} and scans the dynamically provided OpenMRS services
+     * for eligible {@code getXByUuid(String)} methods to populate the internal lookup.
+     *
+     * @param services the list of all available {@code OpenmrsService} implementations
+     */
+    @Autowired
```
mseaton, line 59: See comment above - I worry that this will lead to performance issues at startup

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
```diff
@@ -0,0 +1,185 @@
...
+    public DomainServiceImpl(List<OpenmrsService> services) {
+        if (services != null) {
```
mseaton: Under what circumstances should services be allowed to be null?

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
```diff
@@ -0,0 +1,185 @@
...
+            for (OpenmrsService service : services) {
+                if (this.getClass().isAssignableFrom(service.getClass())) {
+                    continue;
+                }
+                for (Method method : service.getClass().getMethods()) {
+                    if (method.getName().startsWith("get") &&
+                        method.getName().endsWith("ByUuid") &&
```
mseaton: How comprehensive is this?  Are there any services that don't match this pattern that need adjusting?

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
```diff
@@ -0,0 +1,185 @@
...
+        if (fetcher != null) {
+            result = fetcher.fetch(uuid);
+        } else {
+            throw new RuntimeException("No suitable fetcher found for domain: " + type);
+        }
+
+        if (!type.isInstance(result)) {
+            throw new ClassCastException("Expected: " + type + ", but got: " + (result != null ? result.getClass() : "null"));
+        }
+        return type.cast(result);
```
mseaton, line 104: I'm not really sure why we would implement things this way.  I've seen similar functionality added to modules over the years, but wouldn't it be easiest to just re-implement this at the DAO level?  Pass in a class and a uuid and hibernate should be able to return you the object instance related to this via a simple, standard DAO method, assuming that `uuid` is a standard property name (which as far as I know it always is).  This would seem a lot more resilient and less dependent on service implementations.

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
```diff
@@ -0,0 +1,185 @@
...
+	 * Serializes an object to a JSON string if it passes the whitelist check.
+	 *
+	 * @param o the object to serialize
+	 * @return the serialized JSON string
+	 * @throws SecurityException if the class of the object is not whitelisted
+	 *  @throws RuntimeException if serialization fails
+	 */
+	@Override
+	public String serialize(Object o) {
+		String className = o.getClass().getName();
```
mseaton: What if the object is null?  Shouldn't we handle null?  Why get the className here if it is only used within the catch block?

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
```diff
@@ -0,0 +1,185 @@
...
+	 *
+	 * @param serializedObject the JSON string
+	 * @param clazz            the target class to deserialize to
+	 * @param <T>              the type parameter
+	 * @return the deserialized Java object
+	 * @throws SerializationException if deserialization fails
+	 */
+	@Override
+	public <T> T deserialize(String serializedObject, Class<? extends T> clazz) throws SerializationException {
+
```
mseaton, line 125: Should we handle the scenario where serializedObject is null or empty (and return null if so)?  Should we handle the scenario when clazz is null?

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	/**
+	 * Checks if a class is allowed for serialization based on the configured whitelist.
+	 *
+	 * @param className the fully qualified name of the class
+	 * @return true if whitelisted or if whitelist is empty, false otherwise
+	 */
+	private boolean isWhitelisted(String className) {
+		return adminService.getSerializerWhitelistTypes().isEmpty() ? true : adminService.getSerializerWhitelistTypes().stream().anyMatch(pattern -> matchPattern(pattern, className));
```
mseaton: This feels like it is a value we might want to cache.  If you use ConfigUtil.getGlobalProperty(), then this caching is handled for you.  Or if you use ConfigUtil.getProperty(), that would allow runtime properties and/or system properties to also provide this configuration value.

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
```diff
@@ -0,0 +1,185 @@
...
+	* @param className  the fully qualified name of the class to check
+	* @return {@code true} if the class matches the pattern; {@code false} otherwise or if class resolution fails
+	*
+	* @example
+	* <pre>
+	* matchPattern("org.openmrs.*", "org.openmrs.Patient") → true
+	* matchPattern("hierarchyOf:org.openmrs.OpenmrsObject", "org.openmrs.Patient") → true
+	* </pre>
+	*/
+	private boolean matchPattern(String pattern, String className) {
```
mseaton: This feels like something that is better located in a common utility class.

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
```diff
@@ -0,0 +1,185 @@
...
+		try {
+			if (pattern.startsWith("hierarchyOf:")) {
+				String baseClassName = pattern.substring("hierarchyOf:".length());
+				Class<?> baseClass = Class.forName(baseClassName);
+				Class<?> actualClass = Class.forName(className);
```
mseaton: I'm not sure about the use of Class.forName here?  Shouldn't this instead be `Context.loadClass(className)` , which delegates this to the OpenmrsClassloader?

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
dkayiwa: Yes it is still a problem.

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
Ruhanga: I think this should should go away. It's a left over of a previous idea before transforming this to a standard service.

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
Ruhanga: > How comprehensive is this? 

I think this is fairly comprehensive, as that's the precise and adopted way to query by UUID. 

> Are there any services that don't match this pattern that need adjusting?

I've not seen one. I'd be happy to consider such scenarios and adjust the approach given an example of such.

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
Ruhanga: While re-implementing the functionality directly at the DAO level delegating to OpenMRS core's `HibernateUtil.getUniqueEntityByUUID()` may appear simpler and more direct, I think it introduces concerns such as, violating the service-layer transactional contract and bypassing `@Authorized` checks which risks exposing unauthorized access.

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
Ruhanga: > Should we handle the scenario where serializedObject is null or empty (and return null if so)? 

Fair point. I should be able to the add necessary change. 

> Should we handle the scenario when clazz is null?

I would throw a `SerializationException` which is appropriate here I think.

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
Ruhanga: Ohh, yes. `ConfigUtil.getProperty()` is the way to go indeed. I simply didn't know this existed despite working with the api for a while... :)

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
Ruhanga: Agreed. I now realize this would be more appropriately placed in `OpenmrsUtil.java`. I'll add the relevant changes.

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
Ruhanga: Yes indeed. I just hadn't thought about this. Thanks for pointing it out @mseaton.

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
Ruhanga: > What if the object is null? Shouldn't we handle null? 

I would throw a SerializationException which, I think, is appropriate here.

> Why get the className here if it is only used within the catch block?

Makes sense to get rid of this, which I'll do in a followup commit.

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
rkorytkowski: It's a problem when you declare a service in xml and then use autowired inside it. See https://openmrs.atlassian.net/browse/TRUNK-6363  It's not yet possible to define services with annotations only as they won't have all interceptors (logging, authorization, etc.). Please see https://openmrs.atlassian.net/browse/TRUNK-6361

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
```diff
@@ -0,0 +1,185 @@
...
+     * Internal helper that holds metadata and invocation logic for a specific
+     * {@code getXByUuid(String)} method.
+     */
+    private class DomainFetcher {
+
+        private final Class<?> returnType;
+        private final OpenmrsService service;
+        private final Method method;
+
+        public DomainFetcher(Class<?> returnType, OpenmrsService service, Method method) {
```
rkorytkowski, line 128: I don't like using reflection due to performance reasons. It may hit us with files using hundreds of references. No exact figures for service calls, but simple tests show it's a few times slower than usual method calls. There's much faster LambdaMetafactory that we could use if we want to go that route. However, I would be in favour of introducing an interface to our services:
```
public interface RefByUuid {
	
	<T> T getRefByUuid(Class<T> type, String uuid);
	
	List<?> getRefTypes();
}
```
and **maybe** use reflection only if the interface is missing, but if I were to decide, I would not use it.

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
rkorytkowski: I don't like using reflection due to performance reasons. It may hit us with files using hundreds of references. No exact figures for service calls, but simple tests show it's a few times slower than usual method calls. There's much faster LambdaMetafactory that we could use if we want to go that route. However, I would be in favour of introducing an interface to our services:

public interface RefByUuid {
	
	<T> T getRefByUuid(Class<T> type, String uuid);
	
	List<?> getRefTypes();
}
and maybe use reflection only if the interface is missing, but if I were to decide, I would not use it.

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
rkorytkowski: We would probably benefit from caching this method with `@Cacheable`. Please see the note about using cached method within the same bean at https://openmrs.atlassian.net/wiki/spaces/Archives/pages/25504850/Caching#Applying-Cacheable-in-modules

Ruhanga: I haven’t had a chance to use `@Cacheable` yet, but it seems like a great feature in Spring. I trust it works well, though I’m not quite sure how to approach testing it. One thing I might worry about is how to update the cache in response to runtime changes, since the queried value is a global prop.

rkorytkowski: Runtime properties change requires restart and cache is not persisted between restarts. It also invalidates by default after being idle for 100s. You can also add `@CacheEvict` on saveGlobalProperty(ies).

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
Ruhanga: Sorry, I missed this one @rkorytkowski. I don't know what became of my response to this last week here.

In summary, I had considered the interface approach as well, but was concerned it might be harder to scale and more cumbersome for implementations to adopt or upgrade consistently. As for performance, I don’t think the impact is significant, even with multiple method invocations, since method lookups are cached, and the only real cost is at invocation. That’s similar to what Jackson does internally. In our case, the more substantial cost comes from hitting the database, not the reflective call itself IMO.

Ruhanga: Despite the above, I do appreciate having things explicitly coded. So, I’ve decided to go with the interface approach as the preferred solution, while keeping the current behaviour as a fallback.

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
Ruhanga: I’ve utilized `ServiceContext` in the direction @ibacher has suggested [above](https://github.com/openmrs/openmrs-core/pull/5099#discussion_r2180528175). Eventually, this service’s implementation should be integrated into the `ServiceContext` class.

api/src/main/java/org/openmrs/api/RefByUuid.java
```diff
@@ -0,0 +1,53 @@
...
+     * @return the resolved domain object instance, or {@code null} if no matching object is found
+     */
+    <T> T getRefByUuid(Class<T> type, String uuid);
+
+    /**
+     * Get the list of domain types that this service can resolve.
+     *
+     * @return a list of {@link Class} objects representing supported domain types
+     */
+    List<?> getRefTypes();
```
ibacher: Shouldn't the type here be `List<Class<?>>`

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
```diff
@@ -0,0 +1,157 @@
...
+
+	/**
+	 * Checks if a class is allowed for serialization based on the configured whitelist.
+	 *
+	 * @param className the fully qualified name of the class
+	 * @return true if whitelisted or if whitelist is empty, false otherwise
+	 */
+	private boolean isWhitelisted(String className) {
+		return adminService.getSerializerWhitelistTypes().isEmpty() ? true : adminService.getSerializerWhitelistTypes().stream().anyMatch(pattern -> OpenmrsUtil.matchPattern(pattern, className));
+	}
```
ibacher: FWIW, I think we can skip worrying about the whitelisting. The reason for the white-listing facility is, essentially, that the XStream team has claimed that the ability to run arbitrary code, including arbitrary remote code on deserialization is core functionality for XStream. Thus, whitelisting what classes XStream can serialize and deserialize is pretty much the sole control we have to avoid such vulnerabilities. Jackson, on the other hand, has taken a different attitude, routinely patching any discovered RCEs (and none have been disclosed since 2019, despite it being a higher-profile library than XStream).

api/src/main/java/org/openmrs/api/RefByUuid.java
Ruhanga: Yes indeed.

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
Ruhanga: Ok, this makes sense. I had thought of it more as an administrative rule rather than a security measure. I'll get rid of this whitelisting logic altogether. Thanks for your review @ibacher....

dkayiwa:
@Ruhanga have you tested this with any modules?

Ruhanga:
Not with the current codebase, given that it is using Java 21. Some of the affected modules still need to be made compatible with Java 21 and the latest Core 3.0.0-SNAPSHOT. Given this, testing will be unavoidable as I adopt modules to using the feature.

Ruhanga:
I'm happy to get this merged in. Are there any reservations about this @ibacher, @dkayiwa, @mseaton, @wikumChamith, @rkorytkowski, @mherman22?

api/src/main/java/org/openmrs/api/DomainService.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.openmrs.api;
+
+import java.util.List;
+
+/**
+ * Interface for Openmrs services supporting
+ * fetching domain objects by UUID.
+ */
+public interface DomainService extends OpenmrsService {
```
dkayiwa, line 20: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#New-Public-Methods-and-Classes

api/src/main/java/org/openmrs/api/RefByUuid.java
```diff
@@ -0,0 +1,54 @@
...
+ *     RefByUuid service = ...;
+ *
+ *     // Fetch a patient by UUID
+ *     Patient p = service.getRefByUuid(Patient.class, "some-uuid");
+ *
+ *     // Get all types this service can resolve
+ *     List<?> types = service.getRefTypes();
+ * </pre>
+ */
+public interface RefByUuid {
```
dkayiwa, line 37: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#New-Public-Methods-and-Classes

api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
```diff
@@ -0,0 +1,195 @@
...
+ * This is particularly useful for deserialization scenarios where a UUID string
+ * needs to be resolved into a fully initialized domain object (e.g., during JSON parsing).
+ *
+ * @see DomainService
+ * @see UuidReferenceModule
+ */
+
+@Service("domainService")
+@Transactional
+public class DomainServiceImpl extends BaseOpenmrsService implements DomainService {
```
dkayiwa, line 49: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#New-Public-Methods-and-Classes

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
```diff
@@ -0,0 +1,140 @@
...
+ * It supports UUID-based references for {@code OpenmrsObject}s using
+ * {@link UuidReferenceModule} and handles Hibernate proxies via {@link Hibernate5Module},
+ * </p>
+ *
+ * @see UuidReferenceModule
+ * @see OpenmrsSerializer
+ * @see SerializationException
+ */
+@Component("jacksonSerializer")
+public class JacksonSerializer implements OpenmrsSerializer {
```
dkayiwa, line 46: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#New-Public-Methods-and-Classes

api/src/main/java/org/openmrs/serialization/UuidReferenceModule.java
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * This approach helps to:
+ * <ul>
+ *   <li>Reduce object size by avoiding full object nesting</li>
+ *   <li>Enable reference-based object graph resolution</li>
+ *   <li>Support lazy loading of related entities based on UUIDs</li>
+ * </ul>
+ *
+ */
+public class UuidReferenceModule extends SimpleModule {
```
dkayiwa, line 56: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#New-Public-Methods-and-Classes

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
```diff
@@ -90,6 +91,8 @@ public class AdministrationServiceImpl extends BaseOpenmrsService implements Adm
 	@Autowired
 	@Qualifier("openmrsEventListeners")
 	private EventListeners eventListeners;
+
+	private static final String SERIALISER_WHITE_LIST_TYPES = "serializerWhiteListTypes";
```
dkayiwa: What is the use of declaring this in a constant that is private and used only once?

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
```diff
@@ -0,0 +1,142 @@
...
+ * @since 3.0.0
+ *
+ * @see UuidReferenceModule
+ * @see OpenmrsSerializer
+ * @see SerializationException
+ */
+@Component("jacksonSerializer")
+public class JacksonSerializer implements OpenmrsSerializer {
+
+	protected static final Logger log = LoggerFactory.getLogger(JacksonSerializer.class);
```
dkayiwa: Is there a reason why this is not private as advised at? https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#How-To-Log

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
```diff
@@ -0,0 +1,142 @@
...
+	 */
+	@Override
+	public String serialize(Object o) {
+		if (o == null) {
+			throw new RuntimeException("Cannot serialize null object");
+		}
+		try {
+			return getObjectMapper().writeValueAsString(o);
+		} catch (JsonProcessingException e) {
+			throw new RuntimeException("Unable to serialize class: " + o.getClass().getName(), e);
```
dkayiwa: Is there a reason why this is also not a SerializationException?

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
Ruhanga: No particular need in favour of this. I'll get rid of this declaration.

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
Ruhanga: `SerializationException` is a checked exception, but the implemented `serialize(Object)` method doesn’t declare it in a `throws` clause, so it can’t actually be thrown from that method.

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
Ruhanga: No special reason. I'll make it private as advised. Thanks for the pointer.

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
dkayiwa: Why doesn't it declare it in its throws clause?

dkayiwa: @Ruhanga did you see the above?

api/src/main/java/org/openmrs/serialization/UuidReferenceModule.java
```diff
@@ -0,0 +1,160 @@
...
+                    DeserializationConfig config,
+                    BeanDescription beanDesc,
+                    BeanDeserializerBuilder builder) {
+
+                Iterator<SettableBeanProperty> properties = builder.getProperties();
+                while (properties.hasNext()) {
+                    SettableBeanProperty prop = properties.next();
+                    Class<?> rawType = prop.getType().getRawClass();
+
+                    if (OpenmrsObject.class.isAssignableFrom(rawType)) {
```
samuelmale, line 133: This is likely not an issue, but I'll mention it anyway. Currently, the deserializer assumes all OpenMRS object fields are UUID strings. Should we support partial OpenMRS objects as well (object literals)?

We could apply the `uuidDeserializer` only for UUID-like values, and let Jackson handle literal objects normally. For example:

```ts
// UUID reference
{ "location": "8d6c993e-c2cc-11de-8d13-0010c6dffd0f" }

// Partial object
{ "location": { "name": "Unknown Location" } }

```

With the current setup, the partial object will be deserialized to a `null` value.

api/src/main/java/org/openmrs/serialization/UuidReferenceModule.java
```diff
@@ -0,0 +1,160 @@
...
+                    }
+                }
+                for (int i = 0; i < beanProperties.size(); i++) {
+                    BeanPropertyWriter original = beanProperties.get(i);
+
+                    beanProperties.set(i, new BeanPropertyWriter(original) {
+                        @Override
+                        public void serializeAsField(Object bean, JsonGenerator gen, SerializerProvider prov) throws Exception {
+                            Object value = get(bean);
+                            if (value instanceof OpenmrsObject) {
```
samuelmale, line 108: Would it be necessary to add a guardrail to confirm whether this object has a row in the DB?



```java
  if (value instanceof OpenmrsObject && ((OpenmrsObject) value).getId() != null) {
```

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -9,6 +9,8 @@
  */
 package org.openmrs.api.impl;
 
+
```
samuelmale: Nit: remove empty line

api/src/main/java/org/openmrs/serialization/UuidReferenceModule.java
```diff
@@ -0,0 +1,160 @@
...
+
+                Map<String, BeanPropertyDefinition> propertyDefMap = beanDesc.findProperties().stream().collect(Collectors.toMap(BeanPropertyDefinition::getName, def -> def));
+                Iterator<BeanPropertyWriter> propIt = beanProperties.iterator();
+
+                while (propIt.hasNext()) {
+                    BeanPropertyWriter writer = propIt.next();
+                    BeanPropertyDefinition def = propertyDefMap.get(writer.getName());
+
+                    if (def == null || !def.hasField()) {
+                        // Exclude from serialization if there's no backing field
```
dkayiwa, line 96: Is this something that we should log a warning for?

api/src/test/java/org/openmrs/serialization/JacksonSerializerTest.java
```diff
@@ -0,0 +1,131 @@
...
+    @Test
+    public void serialize_shouldSerializeOpenmrsObject() throws Exception{
+        // setup
+        ObjectMapper mapper = new ObjectMapper();
+
+        // replay
+        String serializedEnc = serializer.serialize(Context.getEncounterService().getEncounter(3));
+
+        // verify
+
```
samuelmale: Nit: remove empty line

api/src/test/java/org/openmrs/serialization/JacksonSerializerTest.java
```diff
@@ -0,0 +1,131 @@
...
+        JsonNode actaulEnc = mapper.readTree(serializedEnc);
```
samuelmale:
```suggestion
        JsonNode actualEnc = mapper.readTree(serializedEnc);
```

api/src/test/java/org/openmrs/serialization/JacksonSerializerTest.java
```diff
@@ -0,0 +1,131 @@
...
...tation\":null,\"status\":\"PRELIMINARY\",\"referenceRange\":null,\"complex\":false,\"obsGrouping\":false},{\"uuid\":\"39fb7f47-e80a-4056-9285-bd798be13c63\",\"creator\":\"1010d442-e134-11de-babe-001e378eb67e\",\"dateCreated\":\"2008-08-18T14:09:35\",\"changedBy\":null,\"dateChanged\":null,\"voided\":false,\"dateVoided\":null,\"voidedBy\":null,\"voidReason\":null,\"formNamespaceAndPath\":null,\"obsId\":7,\"concept\":\"c607c80f-1ea9-4da3-bb88-6276ce8868dd\",\"obsDatetime\":\"2008-07-01T00:00:00\",\"accessionNumber\":null,\"obsGroup\":null,\"groupMembers\":[],\"valueCoded\":null,\"valueCodedName\":null,\"valueDrug\":null,\"valueGroupId\":null,\"valueDatetime\":null,\"valueNumeric\":50.0,\"valueModifier\":null,\"valueText\":null,\"valueComplex\":null,\"comment\":\"\",\"person\":\"5946f880-b197-400b-9caa-a3c661d23041\",\"order\":null,\"location\":\"8d6c993e-c2cc-11de-8d13-0010c6dffd0f\",\"encounter\":\"6519d653-393b-4118-9c83-a3715b82d4ac\",\"previousVersion\":null,\"dirty\":false,\"interpretation\":null,\"status\":\"FINAL\",\"referenceRange\":null,\"complex\":false,\"obsGrouping\":false}],\"visit\":\"a2428fea-6b78-11e0-93c3-18a905e044dc\",\"encounterProviders\":null,\"allergies\":[]}";
```
samuelmale:
```suggestion
        String serializedEnc = "{\"uuid\":\"6519d653-393b-4118-9c83-a3715b82d4ac\",\"creator\":\"1010d442-e134-11de-babe-001e378eb67e\",\"dateCreated\":\"2008-08-18T14:09:05\",\"changedBy\":null,\"dateChanged\":null,\"voided\":false,\"dateVoided\":null,\"voidedBy\":null,\"voidReason\":null,\"encounterId\":3,\"encounterDatetime\":\"2008-08-01T00:00:00\",\"patient\":\"5946f880-b197-400b-9caa-a3c661d23041\",\"location\":\"8d6c993e-c2cc-11de-8d13-0010c6dffd0f\",\"form\":\"d9218f76-6c39-45f4-8efa-4c5c6c199f50\",\"encounterType\":\"07000be2-26b6-4cce-8b40-866d8435b613\",\"orders\":null,\"diagnoses\":null,\"conditions\":[],\"obs\":[{\"uuid\":\"be48cdcb-6a76-47e3-9f2e-2635032f3a9a\",\"creator\":\"1010d442-e134-11de-babe-001e378eb67e\",\"dateCreated\":\"2008-08-18T14:11:13\",\"changedBy\":null,\"dateChanged\":null,\"voided\":false,\"dateVoided\":null,\"voidedBy\":null,\"voidReason\":null,\"formNamespaceAndPath\":null,\"obsId\":9,\"concept\":\"a09ab2c5-878e-4905-b25d-5784167d0216\",\"obsDatetime\":\"2008-08-01T00:00:00\",\"accessionNumber\":null,\"obsGroup\":null,\"groupMembers\":[],\"valueCoded\":null,\"valueCodedName\":null,\"valueDrug\":null,\"valueGroupId\":null,\"valueDatetime\":null,\"valueNumeric\":150.0,\"valueModifier\":null,\"valueText\":null,\"valueComplex\":null,\"comment\":\"\",\"person\":\"5946f880-b197-400b-9caa-a3c661d23041\",\"order\":null,\"location\":\"8d6c993e-c2cc-11de-8d13-0010c6dffd0f\",\"encounter\":\"6519d653-393b-4118-9c83-a3715b82d4ac\",\"previousVersion\":null,\"dirty\":true,\"interpretation\":null,\"status\":\"PRELIMINARY\",\"referenceRange\":null,\"complex\":false,\"obsGrouping\":false},{\"uuid\":\"39fb7f47-e80a-4056-9285-bd798be13c63\",\"creator\":\"1010d442-e134-11de-babe-001e378eb67e\",\"dateCreated\":\"2008-08-18T14:09:35\",\"changedBy\":null,\"dateChanged\":null,\"voided\":false,\"dateVoided\":null,\"voidedBy\":null,\"voidReason\":null,\"formNamespaceAndPath\":null,\"obsId\":7,\"concept\":\"c607c80f-1ea9-4da3-bb88-6276ce8868dd\",\"obsDatetime\":\"2008-07-01T00:00:00\",\"accessionNumber\":null,\"obsGroup\":null,\"groupMembers\":[],\"valueCoded\":null,\"valueCodedName\":null,\"valueDrug\":null,\"valueGroupId\":null,\"valueDatetime\":null,\"valueNumeric\":50.0,\"valueModifier\":null,\"valueText\":null,\"valueComplex\":null,\"comment\":\"\",\"person\":\"5946f880-b197-400b-9caa-a3c661d23041\",\"order\":null,\"location\":\"8d6c993e-c2cc-11de-8d13-0010c6dffd0f\",\"encounter\":\"6519d653-393b-4118-9c83-a3715b82d4ac\",\"previousVersion\":null,\"dirty\":false,\"interpretation\":null,\"status\":\"FINAL\",\"referenceRange\":null,\"complex\":false,\"obsGrouping\":false}],\"visit\":\"a2428fea-6b78-11e0-93c3-18a905e044dc\",\"encounterProviders\":null,\"allergies\":[]}";
```

api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
```diff
@@ -1291,4 +1292,43 @@ public OrderAttributeType getOrderAttributeTypeByName(String orderAttributeTypeN
...
+        }
+        if (OrderAttributeType.class.equals(type)) {
+            return (T) getOrderAttributeTypeByUuid(uuid);
+        }
+        if (OrderAttribute.class.equals(type)) {
+            return (T) getOrderAttributeByUuid(uuid);
+        }
+        if (Order.class.equals(type)) {
+            return (T) getOrderByUuid(uuid);
+        }
```
ibacher: `Order` is intentionally subclassable. Is that properly handled?

api/src/main/java/org/openmrs/serialization/UuidReferenceModule.java
ibacher: @samuelmale The deserializer here only needs to be responsible for deserializing data in the format the serializer generates. Since it doesn't generate partial objects, there's no real reason for it to support reading them.

mseaton: Deserializing to a "null" value seems wrong, if that is indeed the cases @samuelmale .  I would expect an error to be thrown if there is a non-empty / non-null text to deserialize and it results in a null value.

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
Ruhanga: > Why doesn't it declare it in its throws clause?

I’ve seen it, thanks for the review, @dkayiwa. The `OpenmrsSerializer` interface originally defines `serialize(Object)` without a `throws <Checked Exception >` clause, which I assume was intentional, possibly because no critical exceptions needed to be handled at compile time. Given that, I’m not opposed to updating the method definition if you think it would be appropriate.

dkayiwa: Can you share a github link to the OpenmrsSerializer.serialize(Object) method that you are looking at?

api/src/main/java/org/openmrs/serialization/UuidReferenceModule.java
Ruhanga: @samuelmale, the partial object scenario you described should result in a `SerializationException`. Since the field is meant to reference a persisted OpenMRS object, I don’t see a strong case for allowing partial objects during deserialization, especially when the full object can be fetched from the database using its UUID.

That said, for non-referenceable objects (i.e., those without a corresponding service to fetch by UUID), full, partial, or nested objects can be deserialized as needed.

api/src/main/java/org/openmrs/serialization/UuidReferenceModule.java
Ruhanga: That would imply serialization is only allowed for `OpenmrsObject`s that have already been persisted, which feels overly restrictive to me. I'd think that any object, regardless of its persistence state, should be serializable.

api/src/main/java/org/openmrs/serialization/UuidReferenceModule.java
Ruhanga: Ideally, no. This simply instructs Jackson to ignore bean methods (getters/setters) that don’t have matching fields. For example:

```
public boolean getIsRetireable() {
    return org.apache.commons.collections4.CollectionUtils.isEmpty(someList);
}
```

api/src/main/java/org/openmrs/serialization/JacksonSerializer.java
Ruhanga: Ohh, sorry I coudn't find it @dkayiwa. My bad. I recall having trouble implementing `SerialzationExeption` and being blocked while working on modules being affected. I'll update the code base to throw this more contextually appropriate exception.

api/src/main/java/org/openmrs/serialization/UuidReferenceModule.java
samuelmale: What's the point of using a UUID serializer on a transient object? We may want to throw some kind of exception or use the default object serializer.

Ruhanga: This is because it's entirely valid to have transient objects with predefined UUIDs, or even objects that are never persisted but still have UUIDs, which I think is perfectly acceptable. I wouldn’t be concerned about whether the object is transient or not. If the object has a UUID, that’s the reference. If not, the reference is simply null.

PR #5135, opened 2025-07-17
TRUNK-6379: Fix broken database-specific tests for MySQL and PostgreSQL

mherman22:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6379

  No? -> execute above command


mherman22:
I have finally fixed these but i see these warnings when tests on `postgres` and the warnings seem to be coming from DbUnit, and it means that DefaultDataTypeFactory is being used for PostgreSQL, which is not ideal.

```
WARN - AbstractTableMetaData.getDataTypeFactory(166) |2025-07-17T17:43:52,287| Potential problem found: 
The configured data type factory 'class org.dbunit.dataset.datatype.DefaultDataTypeFactory' might cause 
problems with the current database 'PostgreSQL' (e.g. some datatypes may not be supported properly). 
In rare cases you might see this message because the list of supported database products is incomplete (list=[derby]). 
If so please request a java-class update via the forums.If you are using your own IDataTypeFactory extending 
DefaultDataTypeFactory, ensure that you override getValidDbProducts() to specify the supported database products.
```

test-module/omod/pom.xml
```diff
@@ -217,7 +241,7 @@
 						<goals>
 							<goal>unpack-dependencies</goal>
 						</goals>
-						<phase>generate-resources</phase>
```
dkayiwa, line 220: Was this change intentional?

mherman22: Yes, i was intending to fix `Artifact has not been packaged yet... unpack should be executed after packaging: see MDEP-98` as a way of ensuring that the unpack goal runs after the `test-module/api/` is packaged

api/pom.xml
```diff
@@ -221,6 +221,11 @@
 			<artifactId>mysql-connector-java</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.postgresql</groupId>
+			<artifactId>postgresql</artifactId>
```
dkayiwa: Did we get any errors because of this?

mherman22: yes, the postgres driver was not accessible here

rkorytkowski: Please get rid of it. I've added postgresql with runtime scope above.

api/src/main/java/org/openmrs/BaseOpenmrsMetadata.java
```diff
@@ -40,8 +41,9 @@ public abstract class BaseOpenmrsMetadata extends BaseOpenmrsObject implements O
 	@FullTextField(analyzer = SearchAnalysis.NAME_ANALYZER)
 	private String name;
 	
-	@Column(name = "description", length = 255)
 	@Lob
+	@Type(type = "large_text")
```
rkorytkowski, line 45: `@Type `is deprecated in Hibernate 6. Is` @Column(columnDefinition="text") `an option?

api/src/main/java/org/openmrs/hibernate/ClobOrTextType.java
```diff
@@ -0,0 +1,40 @@
...
+import org.hibernate.dialect.Dialect;
+import org.hibernate.dialect.MySQLDialect;
+import org.hibernate.type.AbstractSingleColumnStandardBasicType;
+import org.hibernate.type.descriptor.java.StringTypeDescriptor;
+import org.hibernate.type.descriptor.sql.ClobTypeDescriptor;
+import org.hibernate.type.descriptor.sql.SqlTypeDescriptor;
+import org.hibernate.type.descriptor.sql.VarcharTypeDescriptor;
+
+/**
+ * @since 2.8.0
```
rkorytkowski, line 22: Could you please add some description why it's needed?

api/src/main/java/org/openmrs/hibernate/ClobOrTextType.java
```diff
@@ -0,0 +1,40 @@
...
+	private static SqlTypeDescriptor resolveSqlTypeDescriptor(Dialect dialect) {
+		if (dialect instanceof MySQLDialect) {
+			return ClobTypeDescriptor.DEFAULT;
+		}
+		return VarcharTypeDescriptor.INSTANCE;
+	}
+
+    @Override
+    public String getName() {
+        return "large_text";
```
rkorytkowski, line 39: Is it "large_text" or "text" would be enough?

test-module/api/pom.xml
```diff
@@ -42,6 +42,26 @@
...
+		<dependency>
+			<groupId>mysql</groupId>
+			<artifactId>mysql-connector-java</artifactId>
+			<scope>runtime</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.postgresql</groupId>
+			<artifactId>postgresql</artifactId>
+			<scope>runtime</scope>
+		</dependency>
```
rkorytkowski: This should not be needed now. I've cleaned up the dependencies in https://github.com/openmrs/openmrs-core/commit/658a3df03c39450ab3acc7a1bbce9eb75285740b

test-module/omod/pom.xml
```diff
@@ -84,7 +84,31 @@
...
+			<groupId>mysql</groupId>
+			<artifactId>mysql-connector-java</artifactId>
+			<scope>runtime</scope>
+		</dependency>
+		
+		<dependency>
+			<groupId>org.postgresql</groupId>
+			<artifactId>postgresql</artifactId>
+			<scope>runtime</scope>
+		</dependency>
```
rkorytkowski: Should not be needed after https://github.com/openmrs/openmrs-core/commit/658a3df03c39450ab3acc7a1bbce9eb75285740b

test-module/omod/pom.xml
rkorytkowski: Is the resulting omod content exactly the same as without this change?

api/src/main/java/org/openmrs/hibernate/ClobOrTextType.java
mherman22: I thought large_text was more descriptive but yeah, I will use `text`

test-module/api/pom.xml
mherman22: Oh yeah, hadn't seen that commit

api/src/main/java/org/openmrs/hibernate/ClobOrTextType.java
mherman22: Before making this change, shouldn't we consider whether naming the custom type `text` might lead to confusion, since `text` is already a native SQL type?

api/src/main/java/org/openmrs/BaseOpenmrsMetadata.java
mherman22: We're still on Hibernate 5.6.15.Final, so `@Type(type = "large_text")` is valid. Shouldn't we handle Hibernate 6 when we actually upgrade or in another commit? I am happy to update it if you think it's feasible.

rkorytkowski: columnDefinition is available in 5.6 so please use it.

api/src/main/java/org/openmrs/hibernate/ClobOrTextType.java
rkorytkowski: Ahh I missed it's a custom name. Do we know why Hibernate cannot handle this? Are we setting PostgreSQL dialect correctly?

mherman22: In postgres, when hibernate is mapping a `@Lob`annotated `String`, it usually defaults to using the `oid` type, which breaks things when running on postgres for us. That’s why we avoid it by explicitly telling Hibernate to use the `TEXT` type.

rkorytkowski: Have you checked what's the actual type for this column in liquibase? Are we really defining description as longtext (max. 4GBs)? Maybe text type (<64 KBs in MySQL and <1GBs in PSQL) would just work in this case?

rkorytkowski: If we really need a custom type then I would call it clobOrText to make it more clear that it's custom and it matches the classname.

mherman22: Let me change the naming to `clobOrText`. 

> Are we really defining description as longtext (max. 4GBs)? Maybe text type (<64 KBs in MySQL and <1GBs in PSQL) would just work in this case?

Yes, i considered that. We don't actually need `longtext` or `oid`; we just want to avoid Hibernate defaulting to problematic types like `oid` in Postgres. The `text` type is sufficient, and our custom mapping ensures it's used consistently across both databases.

I better described the problem at https://openmrs.atlassian.net/browse/TRUNK-6358

mherman22:
closing this for individual fixes, i am beginning with mysql at https://issues.openmrs.org/browse/TRUNK-6379

PR #5169, opened 2025-07-25
TRUNK-6380: Use Daemon to authenticate task instead of scheduler.username and scheduler.password

IamMujuziMoses:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6380

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Daemon.java
+++ b/api/src/main/java/org/openmrs/api/context/Daemon.java
@@ -31,6 +31,8 @@ import org.openmrs.scheduler.Task;
 import org.openmrs.scheduler.timer.TimerSchedulerTask;
 import org.openmrs.util.OpenmrsSecurityManager;
 import org.openmrs.util.OpenmrsThreadPoolHolder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.context.support.AbstractRefreshableApplicationContext;
 
 /**
@@ -38,6 +40,8 @@ import org.springframework.context.support.AbstractRefreshableApplicationContext
  * module startup when there is no user to authenticate as.
  */
 public final class Daemon {
+
+	private static final Logger log = LoggerFactory.getLogger(Daemon.class);
 	
 	/**
 	 * The uuid defined for the daemon user object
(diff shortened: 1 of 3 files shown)


IamMujuziMoses:
@dkayiwa @ibacher @rkorytkowski could you please review this PR, your feedback will be appreciated.

api/src/main/java/org/openmrs/api/context/Daemon.java
```diff
@@ -184,6 +188,40 @@ public static void executeScheduledTask(final Task task) throws Exception {
...
+
+	/**
+	 * Authenticates a daemon user with privileges so the task can call service layers <br>
+	 * <br>
+	 * This can only be called from {@link TimerSchedulerTask} during actual task execution
+	 *
+	 * <strong>Should</strong> not be called from other methods other than TimerSchedulerTask
+	 * <strong>Should</strong> not throw error if called from a TimerSchedulerTask class
+	 */
+	public static void authenticateTask() throws Exception {
```
IamMujuziMoses, line 200: @dkayiwa This Daemon class is very layered, can I get more info on how it operates

ibacher:
If this only addresses something in the Address Hierarchy module, then we should just fix it there. Modules can mark their activator as "DaemonTokenAware" to get a valid token for running tasks using the Daemon. Honestly, though, it doesn't even look like the `authenticate()` method there is used, so we should just be able to remove that method. I don't think this requires any changes to the Daemon class itself.

mseaton:
> If this only addresses something in the Address Hierarchy module, then we should just fix it there. Modules can mark their activator as "DaemonTokenAware" to get a valid token for running tasks using the Daemon. Honestly, though, it doesn't even look like the `authenticate()` method there is used, so we should just be able to remove that method. I don't think this requires any changes to the Daemon class itself.

As I indicated on the ticket, this was originally something ticketed in addresshierarchy and it was moved into core.  I remain unclear about what core feature needs to change.  I don't really understand what this PR aims to accomplish.  @IamMujuziMoses - can you explain a bit more as to what issue this fixes?

There are many examples of modules that have successfully switched from using the scheduler.username/password properties to using Daemon to accomplish their tasks without requiring specific user authentication.  I would say almost all modules have been fixed to do this, most nearly a decade ago.  We just have a few straggler modules where this wasn't done.

Maybe I am missing the issue - if so let's please lay out what core currently does not do correctly that we are aiming to change.  Otherwise, let's close this PR and move the ticket back into the addresshierarchy module where (I think) it still belongs?

IamMujuziMoses:
@mseaton  This is about changing openmrs-core and modules to use Daemon to authenticate tasks, instead of scheduler.username and scheduler.password

PR #5206, opened 2025-08-03
TRUNK-6376: Fix ClassCastException when adding coded answers of type Drug

Abhxay:
## Issue
[TRUNK-6376](https://issues.openmrs.org/browse/TRUNK-6376)

## What this PR does
Prevents adding a `Drug` object as a coded answer to a `Concept`. Such an action previously caused a `ClassCastException` via the REST API when serializing.

## Fix
- Added a validation check in `Concept.addAnswer()` to throw an `IllegalArgumentException` if the answer is of type `Drug`.
- Ensures only valid `Concept` objects are added as answers.

## Testing
- Build successful (`mvn clean install`)
- All unit tests passing


dkayiwa:
Can you add some tests?

PR #5216, opened 2025-08-05
TRUNK-5899: ConceptAttributeType Domain - Switching from Hibernate Mappings to Annotations

wikumChamith:
## Issue I worked on

see https://openmrs.atlassian.net/browse/TRUNK-5899

  No? -> execute above command


api/src/main/java/org/openmrs/ConceptAttributeType.java
```diff
@@ -9,13 +9,24 @@
...
 import org.openmrs.attribute.BaseAttributeType;
 
+@Entity
+@Table(name = "concept_attribute_type")
 @Audited
 public class ConceptAttributeType extends BaseAttributeType<Concept> implements AttributeType<Concept> {
-	
+
+	@Id
+	@GeneratedValue(strategy = GenerationType.IDENTITY)
```
ManojLL, line 28: Isn't generation type should be `sequence`?

wikumChamith: Even though the HBM file defines the ID generation strategy as a sequence, in Liquibase, it's set to `autoIncrement`. So I believe using `IDENTITY` is valid in this context:
https://github.com/openmrs/openmrs-core/blob/f7cbc7197a3b03b0514727523f13dacb6f89b6fb/api/src/main/resources/org/openmrs/liquibase/snapshots/schema-only/liquibase-schema-only-2.8.x.xml#L261-L264

If we use `@SequenceGenerator` here, Hibernate will throw an error during schema validation because the sequence isn’t defined in the database:

`org.hibernate.tool.schema.spi.SchemaManagementException: Schema-validation: missing sequence [concept_attribute_type_concept_attribute_type_id_seq]`

We could use `@GenericGenerator` as a workaround, but it’s deprecated in Hibernate 6. Also, JPA expects a `@SequenceGenerator`, so if we use Hibernate `@GenericGenerator`, it’s likely ignored, and Hibernate probably falls back to using IDENTITY anyway.

PR #5243, opened 2025-08-16
TRUNK-5749: Migrate `Drug` from hibernate mappings to JPA annotations

JaiVora:

Migrate `Drug` from hibernate mappings to JPA annotations

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5749

  No? -> execute above command


ManojLL:
@JaiVora there is a build failure, can you have a look?

JaiVora:
### Problem

The `Drug` class extends `BaseChangeableOpenmrsMetadata`, which in turn extends `BaseOpenmrsMetadata`. The `BaseOpenmrsMetadata` class is annotated as a `@MappedSuperclass` and defines a `description` field. However, the `drug` table does not currently include the `description` column.

One possible solution is to explicitly add the `description` field to the `drug` table schema to align it with the class hierarchy. Would that approach be appropriate, or is there a specific reason this field was excluded?

@dkayiwa @ManojLL

dkayiwa:
@wikumChamith what did we agree on as the way forward for this?

wikumChamith:
> @wikumChamith what did we agree on as the way forward for this?

Yeah, we decided to find a way to update tests to ignore these errors.

wikumChamith:
> @wikumChamith what did we agree on as the way forward for this?

However, this feels like it goes against the fundamental use of the `ValidateHibernateMappingsDatabaseIT`

dkayiwa:
I do not feel comfortable adding useless fields just for the sake of making a test happy. I feel the test should be modelled in accordance to user requirements. Not the other way round of making the user requirements conform to the test. 😊

wikumChamith:
> > @wikumChamith what did we agree on as the way forward for this?
> 
> However, this feels like it goes against the fundamental use of the `ValidateHibernateMappingsDatabaseIT`



> I do not feel comfortable adding useless fields just for the sake of making a test happy. I feel the test should be modelled in accordance to user requirements. Not the other way round of making the user requirements conform to the test. 😊

To do this, we'll probably have to disable the whole ValidateHibernateMappingsDatabaseIT test, as we are relying on hb2ddl validate for this.

dkayiwa:
Don't we have a workaround? Because we still need the validation to happen when we expect the fields.

wikumChamith:
> Don't we have a workaround? Because we still need the validation to happen when we expect the fields.

How about removing the classes that we expect to get this error from this test?

dkayiwa:
Looks like the best option that we have for now. Remember to include a comment explaining why we have done so for now until we get a better solution.

JaiVora:
@dkayiwa @wikumChamith - Thank you for your guidance. I've implemented a custom schema filter to skip validation of specific tables. Please review it when you have a moment!

wikumChamith:
@JaiVora, your solution fixes the test, but it doesn’t address the errors we see when running the application. I think the ideal fix would be to add the missing columns to our Liquibase changeset, similar to [this example](https://github.com/openmrs/openmrs-core/blob/87a175e2bc721d1a2e369dcbedad063d82c774c3/api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml#L33-L41)

JaiVora:
@wikumChamith - I've made the suggested changes, can I request a re-review? Thanks!

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -39,5 +39,12 @@
 			<column name="retired_by" type="INT" />
 		</addColumn>
 	</changeSet>
+	
+	<!-- Temporary fix: added column to bypass Hibernate validation for inherited fields. Not functionally required — should be replaced with a proper solution. -->
```
wikumChamith: This comment is not needed, as this is the more permanent solution.

JaiVora: Addressed @wikumChamith

dkayiwa:
We have some merge conflicts to resolve.

JaiVora:
> We have some merge conflicts to resolve.

@dkayiwa - it's resolved now, can you please take another look? thanks!

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -39,5 +39,11 @@
 			<column name="retired_by" type="INT" />
 		</addColumn>
 	</changeSet>
+	
+	<changeSet id="TRUNK-5749-2025-08-20" author="Jai Vora">
+		<addColumn tableName="drug">
+			<column name="description" type="TEXT" />
```
dkayiwa: Do you have a reason for making the description so large?

JaiVora: A few tables have it set to `TEXT` and it's annotated as a `@Lob` in the base class. I agree that's excessive, especially since this column is only being added as a workaround for class hierarchy and Hibernate validation issues. I've set it to `varchar(255)` for now. I could reduce it further, but a description field on the drug table seems like it could have practical use. Since I’m still ramping up, I’m not sure if that’s a likely use case.

dkayiwa: 255 characters is a reasonable value. 👍

dkayiwa:
@wikumChamith your changed were addressed. Correct?

PR #5300, opened 2025-09-06
TRUNK-6418: Run liquibase checks and data imports only when core or modules version changes

Abhxay:
This PR addresses TRUNK-6418 by modifying the DatabaseUpdater to only run Liquibase checks and data imports when the OpenMRS core version or module versions have changed.

Key changes:
- Added `isCoreChanged()` method to detect if the recorded core version differs from the current version.
- Added `markVersionAs()` method to mark the current core version as applied in the database global properties.

These changes prevent unnecessary repeated database updates on startup if the versions have not changed.

No other files or functionality were modified.

Closes: https://openmrs.atlassian.net/browse/TRUNK-6418


Abhxay:
> I don't see where the new methods are used. You check core version only, whereas module versions need to be checked as well.

I’ve updated the PR to include checks for module versions as well, in addition to the core version. Please take a look and let me know if any further adjustments are needed.

rkorytkowski:
Again you must call the methods you introduce during startup... See InitializationFilter

Abhxay:
> Again you must call the methods you introduce during startup... See InitializationFilter

I have updated the PR accordingly ...please review this and let me know if any changes needed

rkorytkowski:
I'm sorry, but you don't seem to quite understand what needs to be done in the initialiser... As I said this issue is not an introductory ticket so you may want to look for a different one to get better understanding of the application and code. I will take it from here. Thanks!

PR #5319, opened 2025-09-14
TRUNK-6430:  Update incorrect link in README / Documentation

mo-farooq:
Replaced the outdated OpenMRS Maven link in the README with the official Maven documentation:
https://maven.apache.org/guides/index.html

This avoids a dead link and gives users an accurate resource for Maven commands.

Issue:https://openmrs.atlassian.net/browse/TRUNK-6430


README.md
```diff
@@ -110,8 +110,7 @@ mvn -Dcargo.servlet.port=8081 cargo:run
 
 If all goes well (check the console output) you can access the OpenMRS application at `localhost:8080/openmrs`.
 
-Refer to [Getting Started as a Developer - Maven](https://wiki.openmrs.org/display/docs/Maven) for some more information
-on useful Maven commands and build options.
+For additional Maven commands and guidance, refer to the [official Maven documentation](https://maven.apache.org/guides/index.html).
```
dkayiwa: Is there a reason why you did not link to this? https://openmrs.atlassian.net/wiki/spaces/Archives/pages/25520545/Maven

mo-farooq: That link is outdated, and it hasn't been updated, that's why I thought it would be better to add the official Maven documentation.
Should I keep it as it is , or change it to the link you have provided

dkayiwa: This one provides some openmrs specific context.

mo-farooq:
Thanks for the suggestion! I’ve updated the README file

README.md
mo-farooq: Thanks for the suggestion! I’ve updated the README file

dkayiwa:
Are you intentionally changing both the link and text around it?

mo-farooq:
Thanks for catching that — I didn’t mean to change the surrounding text, only the broken link. I’ll update the PR so that the text stays exactly the same and only the link is replaced with the OpenMRS Maven wiki.

PR #5337, opened 2025-09-17
TRUNK-6428: Move bean declarations from xml to Java config

tusharvidhani3:
- Replaced XML-defined beans in applicationContext-service.xml with @Configuration + @Bean methods in OpenmrsApplicationContextConfig
- Annotated classes with @Component where applicable
- Preserved component scanning for @Handler and OpenMRS profile filters
- Converted serviceContext, Context, event listeners, and Hibernate-related beans
- Preserved bean dependencies and initialization order using @DependsOn and constructor args
- Updated component scan to include/exclude filters previously defined in XML

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -513,7 +513,7 @@
 			<dependency>
 				<groupId>com.h2database</groupId>
 				<artifactId>h2</artifactId>
-				<version>2.4.240</version>
+				<version>2.3.232</version>
 			</dependency>
 			<dependency>
 				<groupId>org.dbunit</groupId>


api/src/main/java/org/openmrs/api/context/OpenmrsApplicationContextConfig.java
```diff
@@ -0,0 +1,208 @@
...
+    includeFilters = {
+        @ComponentScan.Filter(type = FilterType.ANNOTATION, classes = Handler.class),
+        @ComponentScan.Filter(type = FilterType.CUSTOM, classes = OpenmrsProfileIncludeFilter.class)
+    },
+    excludeFilters = {
+        @ComponentScan.Filter(type = FilterType.CUSTOM, classes = TestTypeFilter.class),
+        @ComponentScan.Filter(type = FilterType.CUSTOM, classes = OpenmrsProfileExcludeFilter.class)
+    }
+)
+public class OpenmrsApplicationContextConfig {
```
rkorytkowski: There's OpenmrsApplicationContextConfig under org.openmrs.api already. Please add it there instead of creating a new one.

api/src/main/resources/applicationContext-service.xml
```diff
@@ -242,52 +135,6 @@
...
-	<bean id="providerAttributeTypeValidator" class="org.openmrs.validator.ProviderAttributeTypeValidator"/>
-	<bean id="encounterValidator" class="org.openmrs.validator.EncounterValidator"/>
-	<bean id="locationAttributeTypeValidator" class="org.openmrs.validator.LocationAttributeTypeValidator"/>
-	<bean id="providerValidator" class="org.openmrs.validator.ProviderValidator"/>
-	<bean id="patientProgramValidator" class="org.openmrs.validator.PatientProgramValidator"/>
-	<bean id="conceptValidator" class="org.openmrs.validator.ConceptValidator"/>
-	<bean id="personMergeLogValidator" class="org.openmrs.validator.PersonMergeLogValidator"/>
-	<bean id="conceptReferenceTermValidator" class="org.openmrs.validator.ConceptReferenceTermValidator"/>
-	<bean id="conceptMapTypeValidator" class="org.openmrs.validator.ConceptMapTypeValidator"/>
-	<bean id="conceptAttributeTypeValidator" class="org.openmrs.validator.ConceptAttributeTypeValidator"/>
```
rkorytkowski, line 201: Did you make sure that all validators have `@Component` on them?

api/src/main/resources/applicationContext-service.xml
```diff
@@ -242,52 +135,6 @@
...
-	<bean id="implementationIdHttpClient" class="org.openmrs.util.HttpClient">
-		<constructor-arg name="url" value="https://implementation.openmrs.org"/>
-	</bean>
-
 	<!--  ************************ END VALIDATOR CONFIGURATION  ************************  -->
 
 	<context:component-scan base-package="org.openmrs">
```
rkorytkowski: It should not be needed if you added it to the configuration class. You just need to point Spring to the OpenmrsApplicationContextConfig in addition to xml files.

api/src/main/java/org/openmrs/api/context/OpenmrsApplicationContextConfig.java
```diff
@@ -0,0 +1,208 @@
...
+
+    @Bean
+    public PersonNameGlobalPropertyListener personNameGlobalPropertyListener() {
+        return new PersonNameGlobalPropertyListener();
+    }
+
+    @Bean
+    public LoggingConfigurationGlobalPropertyListener loggingConfigurationGlobalPropertyListener() {
+        return new LoggingConfigurationGlobalPropertyListener();
+    }
```
rkorytkowski: For simple beans, please just annotate classes with `@Component` instead of adding them to the config. It makes sense to add classes to the config if we want to specifically configure something or cannot annotate classes as they come from external libraries. Also if there was an id used in xml, please set the same id using annotation.

api/src/main/java/org/openmrs/api/context/OpenmrsApplicationContextConfig.java
```diff
@@ -0,0 +1,208 @@
...
+            PersonNameGlobalPropertyListener personNameGlobalPropertyListener,
+            LoggingConfigurationGlobalPropertyListener loggingConfigurationGlobalPropertyListener,
+            GlobalLocaleList globalLocaleList, AdministrationServiceImpl adminService,
+            OrderServiceImpl orderService) {
+        EventListeners eventListeners = new EventListeners();
+        List<GlobalPropertyListener> globalPropertyListeners = List.of(localeUtility, locationUtility,
+                configUtilGlobalPropertyListener, personNameGlobalPropertyListener,
+                loggingConfigurationGlobalPropertyListener, globalLocaleList, adminService, orderService);
+        eventListeners.setGlobalPropertyListeners(globalPropertyListeners);
+        return eventListeners;
```
rkorytkowski: It's a very strange construct, but I'm fine with this for now. We'll refactor in a separate issue.

api/src/main/java/org/openmrs/api/context/OpenmrsApplicationContextConfig.java
```diff
@@ -0,0 +1,208 @@
...
+        serviceContext.setCohortService(cohortService);
+        serviceContext.setMessageService(messageService);
+        serviceContext.setSerializationService(serializationService);
+        serviceContext.setSchedulerService(schedulerService);
+        serviceContext.setAlertService(alertService);
+        serviceContext.setHl7Service(hL7Service);
+        serviceContext.setMessageSourceService(messageSourceService);
+        serviceContext.setVisitService(visitService);
+        serviceContext.setProviderService(providerService);
+        return serviceContext;
```
rkorytkowski: We'll want to refactor serviceContext, but I'm fine with this for now...

api/src/main/java/org/openmrs/api/context/OpenmrsApplicationContextConfig.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    @Bean
+    public Context context(ContextDAO contextDAO, ServiceContext serviceContext) {
+        Context context = new Context();
+        context.setServiceContext(serviceContext);
+        context.setContextDAO(contextDAO);
+        return context;
```
rkorytkowski: Just annotate Context with `@Component`.

api/src/main/java/org/openmrs/api/context/OpenmrsApplicationContextConfig.java
```diff
@@ -0,0 +1,208 @@
...
+
+    @Bean
+    public List<Object> mappingJarResources() {
+        List<Object> mergedList = new ArrayList<>(moduleTestingMappingJarLocations());
+        return mergedList;
+    }
+
+    @Bean
+    public AuditableInterceptor auditableInterceptor() {
+        return new AuditableInterceptor();
```
rkorytkowski: Just annotate the class.

api/src/main/java/org/openmrs/api/context/OpenmrsApplicationContextConfig.java
```diff
@@ -0,0 +1,208 @@
...
+                new ClassPathResource("hibernate.cfg.xml")
+        });
+        sessionFactory.setMappingJarLocations(mappingJarResources);
+        sessionFactory.setPackagesToScan("org.openmrs");
+        return sessionFactory;
+    }
+
+    @Bean
+    public DbSessionFactory dbSessionFactory(SessionFactory sessionFactory) {
+        return new DbSessionFactory(sessionFactory);
```
rkorytkowski: Just annotate the class.

api/src/main/java/org/openmrs/api/context/OpenmrsApplicationContextConfig.java
```diff
@@ -0,0 +1,217 @@
...
+
+    @Bean
+    public HibernateSessionFactoryBean sessionFactory(Resource mappingJarResources) {
+        HibernateSessionFactoryBean sessionFactory = new HibernateSessionFactoryBean();
+        sessionFactory.setConfigLocations(new Resource[] {
+                new ClassPathResource("hibernate.cfg.xml")
+        });
+        sessionFactory.setMappingJarLocations(mappingJarResources);
+        sessionFactory.setPackagesToScan("org.openmrs");
+        return sessionFactory;
```
rkorytkowski: HibernateSessionFactoryBean is a strange mix of annotations and bean config, but we can refactor later on to annotations. I'm fine with this for now.

rkorytkowski (changes requested):

api/src/main/resources/applicationContext-service.xml
tusharvidhani3: These have @Handlers annotated instead of @Component that works similar

api/src/main/java/org/openmrs/api/OpenmrsApplicationContextConfig.java
```diff
@@ -56,12 +63,12 @@
  */
 @Configuration
```
rkorytkowski, line 57: Did you forget to add component scan config here?

api/src/main/java/org/openmrs/api/OpenmrsApplicationContextConfig.java
```diff
@@ -146,4 +155,42 @@ public Map<String, Application> hL7Handlers(ORUR01Handler orur01Handler, ADTA28H
 		map.put("ADT_A28", adta28Handler);
 		return map;
 	}
+
+	@Bean
+	@DependsOn("clearOpenmrsEventListeners")
+	public EventListeners openmrsEventListeners(List<GlobalPropertyListener> globalPropertyListeners) {
+		EventListeners eventListeners = new EventListeners();
+		eventListeners.setGlobalPropertyListeners(globalPropertyListeners);
+		return eventListeners;
```
rkorytkowski: Let's just get rid of clearOpenmrsEventListeners and have `@Component("openmrsEventListeners")` I don't quite understand the concept behind clearOpenmrsEventListeners as creating a new bean always starts with a fresh list of listeners.

rkorytkowski:
@tusharvidhani3 Could you please update to the latest master and resolve the conflict so the code builds?

tusharvidhani3:
@rkorytkowski Since DAO beans were already configured with Java config in latest commit, no beans are needed in applicationContext-service.xml & therefore I have removed the file. Is it ok?

rkorytkowski:
@tusharvidhani3, yes, let's delete the xml.

tusharvidhani3:
@rkorytkowski I have removed the xml file and the classpath references to it. I think all the errors are fixed now. I request you to please re-review the PR

PR #5379, opened 2025-10-07
TRUNK-6435: Password resets must go through service API

ibacher:

This adds guards to the `HibernateUserDao` for all calls that can change a user's password to ensure that they are called from the `UserService` class. I've also modified the `Daemon` class to use a slightly different mechanism to look-up the caller primarily because the `SecurityManager`, and hence the `OpenmrsSecurityManager` are deprecated.

N.B. There's a better for this available in Java 9+ using the `StackTraceWalker` machinery, but since in the 2.x branches we still support Java 8, I'm going to do this as two PRs, one here for 2.8.x and backporting if we opt to and a second PR for 3.x where we can use the Java 9 APIs. (The `StackTraceWalker` allows the lazy-loading of stack frames using the stream API; the technique used here—as with the technique of using a `SecurityManager`—materializes the full stack frame).

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6435

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Daemon.java
+++ b/api/src/main/java/org/openmrs/api/context/Daemon.java
@@ -23,13 +23,13 @@ import org.openmrs.api.APIAuthenticationException;
 import org.openmrs.api.APIException;
 import org.openmrs.api.OpenmrsService;
 import org.openmrs.api.db.ContextDAO;
+import org.openmrs.api.db.hibernate.HibernateContextDAO;
 import org.openmrs.module.DaemonToken;
 import org.openmrs.module.Module;
 import org.openmrs.module.ModuleException;
 import org.openmrs.module.ModuleFactory;
 import org.openmrs.scheduler.Task;
 import org.openmrs.scheduler.timer.TimerSchedulerTask;
-import org.openmrs.util.OpenmrsSecurityManager;
 import org.openmrs.util.OpenmrsThreadPoolHolder;
 import org.springframework.context.support.AbstractRefreshableApplicationContext;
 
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/openmrs/api/context/Daemon.java
```diff
@@ -76,9 +76,15 @@ public static Module startModule(Module module) throws ModuleException {
 	 */
 	public static Module startModule(final Module module, final boolean isOpenmrsStartup,
 	        final AbstractRefreshableApplicationContext applicationContext) throws ModuleException {
-		Class<?> callerClass = new OpenmrsSecurityManager().getCallerClass(0);
-		if (callerClass != Daemon.class && callerClass != ModuleFactory.class) {
-			throw new APIException("Module.factory.only", new Object[] { callerClass.getName() });
+		StackTraceElement[] stack = new Exception().getStackTrace();
```
dkayiwa, line 79: Is there a way of wrapping this in a method call and reuse it?

ibacher: Yes, but a few things lead me to not do that:

1. This is functionality we use a lot in this class (and now the `HibernateUserDAO`), but it's probably not something we need outside of those specialized areas. If we extend it to a third class, I'll refactor 😁. (Doing this is also something I don't _really_ want to encourage anyone to do if it's avoidable).
2. The JVM is _technically_ permitted to omit stack frames from this result, so calling into a new method could make the resulting stack trace less reliable, particularly in a couple of cases (like this method) where it's invoked from a lambda function and in cases where we need to look back more than 1 stack frame.
3. The API is much nicer for 3.x where I can use Java 9+ and we change this:

```java
StackTraceElement[] stack = new Exception().getStackTrace();
if (stack.length < 2) {
	throw new APIException("Could not determine where createUser() was called from");
}
StackTraceElement caller = stack[1];
String callerClass = caller.getClassName();
```

To this:

```java
var possibleFrame = STACK_WALKER.walk(s -> 
	s.skip(1).limit(1).map(StackWalker.StackFrame::getDeclaringClass).findFirst()
);
```

After which `possibleFrame` is a `Optional<Class<?>>` that contains the caller class, if there is any.

dkayiwa: Sounds fair enough and I love the Java 9+ replacement. 😊

PR #5392, opened 2025-10-10
TRUNK-6442: ObsValidator should allow user-supplied reference ranges

ibacher:

This changes the ObsValidator so that it will not overwrite an existing value for the `obsReferenceRange` or `interpretation` fields if these are already filled in.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6442

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/Obs.java
+++ b/api/src/main/java/org/openmrs/Obs.java
@@ -1279,5 +1279,8 @@ public class Obs extends BaseFormRecordableOpenmrsData {
 	 */
 	public void setReferenceRange(ObsReferenceRange referenceRange) {
 		this.referenceRange = referenceRange;
+		if (referenceRange != null) {
+			referenceRange.setObs(this);
+		}
 	}
 }
--- a/api/src/main/java/org/openmrs/validator/ObsValidator.java
+++ b/api/src/main/java/org/openmrs/validator/ObsValidator.java
@@ -10,12 +10,10 @@
 package org.openmrs.validator;
 
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
-import java.util.Objects;
 
 import org.apache.commons.text.StringEscapeUtils;
+import org.openmrs.BaseReferenceRange;
 import org.openmrs.Concept;
 import org.openmrs.ConceptDatatype;
 import org.openmrs.ConceptNumeric;
@@ -27,7 +25,6 @@ import org.openmrs.api.APIException;
 import org.openmrs.api.context.Context;
 import org.openmrs.api.db.hibernate.HibernateUtil;
 import org.openmrs.util.ConceptReferenceRangeUtility;
-import org.openmrs.util.OpenmrsUtil;
 import org.springframework.validation.Errors;
 import org.springframework.validation.Validator;
 
@@ -198,7 +195,7 @@ public class ObsValidator implements Validator {
 						}
 					}
 					
-					validateConceptReferenceRange(obs, errors, atRootNode);
+					validateReferenceRange(obs, errors, atRootNode);
 				} else if (dt.isText() && obs.getValueText() == null) {
 					if (atRootNode) {
 						errors.rejectValue("valueText", "error.null");
@@ -259,18 +256,22 @@ public class ObsValidator implements Validator {
 	 * @param obs Observation to validate
 	 * @param errors Errors to record validation issues
 	 */
-	private void validateConceptReferenceRange(Obs obs, Errors errors, boolean atRootNode) {
-		ConceptReferenceRange conceptReferenceRange = getReferenceRange(obs);
+	private void validateReferenceRange(Obs obs, Errors errors, boolean atRootNode) {
+		BaseReferenceRange referenceRange = obs.getReferenceRange();
+		if (referenceRange == null) {
+			referenceRange = getReferenceRange(obs);
+		}
 
-		if (conceptReferenceRange != null) {
-			validateAbsoluteRanges(obs, conceptReferenceRange, errors, atRootNode);
+		if (referenceRange != null) {
+			validateAbsoluteRanges(obs, referenceRange, errors, atRootNode);
 			
 			if (obs.getId() == null) {
-				setObsReferenceRange(obs, conceptReferenceRange);
+				setObsReferenceRange(obs, referenceRange);
 			}
 		} else if (obs.getId() == null) {
 			setObsReferenceRange(obs);
 		}
+		
 		setObsInterpretation(obs);
 	}
 
@@ -410,16 +411,16 @@ public class ObsValidator implements Validator {
 	 * Validates the high and low absolute values of the Obs.
 	 *
 	 * @param obs Observation to validate
-	 * @param conceptReferenceRange ConceptReferenceRange containing the range values
+	 * @param referenceRange ConceptReferenceRange containing the range values
 	 * @param errors Errors to record validation issues
 	 */
-	private void validateAbsoluteRanges(Obs obs, ConceptReferenceRange conceptReferenceRange, Errors errors, boolean atRootNode) {
-		if (conceptReferenceRange.getHiAbsolute() != null && conceptReferenceRange.getHiAbsolute() < obs.getValueNumeric()) {
+	private void validateAbsoluteRanges(Obs obs, BaseReferenceRange referenceRange, Errors errors, boolean atRootNode) {
+		if (referenceRange.getHiAbsolute() != null && referenceRange.getHiAbsolute() < obs.getValueNumeric()) {
 			if (atRootNode) {
 				errors.rejectValue(
 					"valueNumeric",
 					"error.value.outOfRange.high",
-					new Object[] { conceptReferenceRange.getHiAbsolute() },
+					new Object[] { referenceRange.getHiAbsolute() },
 					null
 				);
 			} else {
@@ -432,12 +433,12 @@ public class ObsValidator implements Validator {
 			}
 		}
 		
-		if (conceptReferenceRange.getLowAbsolute() != null && conceptReferenceRange.getLowAbsolute() > obs.getValueNumeric()) {
+		if (referenceRange.getLowAbsolute() != null && referenceRange.getLowAbsolute() > obs.getValueNumeric()) {
 			if (atRootNode) {
 				errors.rejectValue(
 					"valueNumeric",
 					"error.value.outOfRange.low",
-					new Object[] { conceptReferenceRange.getLowAbsolute() },
+					new Object[] { referenceRange.getLowAbsolute() },
 					null
 				);
 			} else {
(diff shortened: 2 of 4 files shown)


api/src/main/java/org/openmrs/validator/ObsValidator.java
```diff
@@ -455,20 +456,23 @@ private void validateAbsoluteRanges(Obs obs, ConceptReferenceRange conceptRefere
 	 * Builds and sets the ObsReferenceRange for the given Obs.
 	 *
 	 * @param obs Observation to set the reference range
-	 * @param conceptReferenceRange ConceptReferenceRange used to build the ObsReferenceRange
+	 * @param referenceRange ConceptReferenceRange used to build the ObsReferenceRange
```
samuelmale, line 459: Did you intend to replace `ConceptReferenceRange` with maybe `BaseReferenceRange`?

api/src/main/java/org/openmrs/validator/ObsValidator.java
```diff
@@ -410,16 +411,16 @@ private ConceptReferenceRange findStrictestReferenceRange(List<ConceptReferenceR
 	 * Validates the high and low absolute values of the Obs.
 	 *
 	 * @param obs Observation to validate
-	 * @param conceptReferenceRange ConceptReferenceRange containing the range values
+	 * @param referenceRange ConceptReferenceRange containing the range values
```
samuelmale, line 414: Did you intend to replace `ConceptReferenceRange` with maybe `BaseReferenceRange`?

Binayak490-cyber:
Hi @ibacher, @samuelmale, the PR is not touched since more than 4 months, are the updates done or the PR doesn't satisfy the fix for the issue...? If not so then i can go for a fix for the issue.

api/src/main/java/org/openmrs/Obs.java
```diff
@@ -1279,5 +1279,8 @@ public ObsReferenceRange getReferenceRange() {
 	 */
 	public void setReferenceRange(ObsReferenceRange referenceRange) {
 		this.referenceRange = referenceRange;
+		if (referenceRange != null) {
+			referenceRange.setObs(this);
+		}
```
Binayak490-cyber, line 1284: @ibacher, should the old reference range's obs back-pointer be cleared before setting the new one? e.g. if (this.referenceRange != null) this.referenceRange.setObs(null);

api/src/main/java/org/openmrs/validator/ObsValidator.java
```diff
@@ -455,20 +456,23 @@ private void validateAbsoluteRanges(Obs obs, ConceptReferenceRange conceptRefere
...
+		} else {
+			ObsReferenceRange obsRefRange = new ObsReferenceRange();
+			obsRefRange.setHiAbsolute(referenceRange.getHiAbsolute());
+			obsRefRange.setHiCritical(referenceRange.getHiCritical());
+			obsRefRange.setHiNormal(referenceRange.getHiNormal());
+			obsRefRange.setLowAbsolute(referenceRange.getLowAbsolute());
+			obsRefRange.setLowCritical(referenceRange.getLowCritical());
+			obsRefRange.setLowNormal(referenceRange.getLowNormal());
+			obs.setReferenceRange(obsRefRange);
+		}
```
Binayak490-cyber, line 475: @ibacher, what happens if a user supplies an ObsReferenceRange already linked to a different Obs? Should this case log a warning, throw an exception, or copy the values instead of silently skipping?

api/src/main/java/org/openmrs/validator/ObsValidator.java
```diff
@@ -479,33 +483,37 @@ private void setObsReferenceRange(Obs obs, ConceptReferenceRange conceptReferenc
 	private void setObsReferenceRange(Obs obs) {
 		if (obs.getConcept() == null) {
 			return;
+		} else if (obs.getReferenceRange() != null) {
+			return;
```
Binayak490-cyber, line 487: @ibacher, if a user sets a custom ObsReferenceRange but the concept has no ConceptReferenceRange, the absolute range validation (validateAbsoluteRanges) is skipped entirely. Is that intentional? The user-supplied range could still have meaningful absolute bounds to validate against.

dkayiwa (commented):
This one needs a real rebase rather than a conflict fix: all four files now conflict with master, and the code this PR edits has been restructured since (TRUNK-6512 moved the criteria evaluation behind ConceptService.getConceptReferenceRange(ConceptReferenceRangeContext), TRUNK-6610 reworked the validator internals, and the tree moved to Spotless formatting). The change is still worth carrying over, since master's validator still overwrites both the reference range and the interpretation, but it has to be re-applied onto the new structure. Two things to fold into that rebase: the ConceptReferenceRangeUtility.java hunks here are pure Javadoc re-wrapping and can be dropped, and TRUNK-6542 (Obs.newInstance now copies the reference range when an obs is edited) landed in the meantime, which fits the use-the-attached-range approach here but makes the interpretation issue in the inline comment on setObsInterpretation bite harder.

api/src/main/java/org/openmrs/validator/ObsValidator.java
```diff
@@ -479,33 +483,37 @@ private void setObsReferenceRange(Obs obs, ConceptReferenceRange conceptReferenc
...
 
 	/**
 	 * This method sets Obs interpretation based on the current obs' numeric value.
 	 *
 	 * @param obs Observation to set the interpretation
 	 */
 	private void setObsInterpretation(Obs obs) {
+		if (obs.getInterpretation() != null || obs.getValueNumeric() == null) {
+			return;
+		}
```
dkayiwa, line 513: This needs a rethink before merge: as written, editing a numeric obs keeps the interpretation that was computed for the old value. `Obs.newInstance` copies `interpretation` into the revision that `saveExistingObs` creates, and validation only runs against the original obs, so once an obs carries an interpretation (any numeric obs whose range set one at creation), no later edit ever re-interprets. I ran this on both branches with the standard test data: concept 4090, patient aged 60, value 100 saves as NORMAL; change the value to 131 (hiCritical is 130) and save again. On master the new revision comes back CRITICALLY_HIGH; on this branch it stays NORMAL. Nothing throws, so a corrected value silently keeps the old flag, and everything downstream that reads interpretation (FHIR, flowsheets) reports a normal result for a critical value.

The root problem is that the field carries no provenance, so this guard can't tell "the lab supplied this" from "I computed this on the last save". My preference is to scope the preserve rule to creation:

```suggestion
		if (obs.getValueNumeric() == null || (obs.getId() == null && obs.getInterpretation() != null)) {
			return;
		}
```

That keeps the TRUNK-6442 behavior for externally submitted new obs and restores re-interpretation when an existing obs is amended (the edit flow validates the original before `newInstance` copies it, so the new revision picks up the recomputed value). The trade-off is that an external system updating an existing obs gets its supplied interpretation recomputed; if that flow matters, clearing the interpretation in the edit path when `valueNumeric` changed would be the alternative. Either way, an `ObsServiceTest` that edits a numeric value and asserts the new revision's interpretation would have caught this; the added tests only cover creation.

api/src/main/java/org/openmrs/validator/ObsValidator.java
dkayiwa: Confirmed with the standard test data, and it goes further than the null case: an obs for concept 4090 (60-year-old, so concept_reference_range 30 applies, hiAbsolute 150) with value 200 fails validation on master, but passes on this branch as soon as a user range without absolute bounds is attached, because validateAbsoluteRanges now only ever sees the supplied range. The same happens when the supplied range simply has wider absolutes than the concept's. TRUNK-6442 talks about letting an assigned range overrule the automated selection of what gets stored; I don't read it as dropping the concept's own sanity bounds on the submitted value, and a bogus value that sails through validation is the kind of thing nobody notices until reports look wrong. My preference: keep validateAbsoluteRanges pinned to the concept/criteria-derived range regardless of what was supplied, and use the supplied range only for what gets stored on the obs and for interpretation. If the intent really is that a submitter's range also governs validation, that deserves a Javadoc note and a test.

api/src/main/java/org/openmrs/validator/ObsValidator.java
dkayiwa: With the setter change in this PR that state can't arise: Obs.setReferenceRange now re-points the range (referenceRange.setObs(this)), so by the time the validator sees a user-supplied ObsReferenceRange, getObs() is never null and never another obs. Which means the null check here can never fire, and the instanceof arm quietly does nothing, which happens to be the right outcome since the range is already attached. Re-attaching a range that is persisted for a different obs dies in Hibernate anyway, obs_id being the derived id (@MapsId). I'd simplify this to an early return for the ObsReferenceRange case with a comment saying it's already attached, and keep the copying for concept-derived ranges; as written the guard reads like it does something meaningful.

api/src/main/java/org/openmrs/validator/ObsValidator.java
ibacher: So, I'd like to agree with this, but realistically with our metadata changes we've adjusted CIEL so that it's "absolute" values do not always reflect bounds of possible measurement, but bounds of likely measurement. As long as that remains the case, enforcing absolute values would, in fact, lead to us rejecting perfectly valid readings. Consequently, those value-bounds cannot be used to validate a submitted obs and we must rely on the obs submitted with it's own reference range to correctly define the absolute bounds, just like we rely on it supplying the normal bounds.

api/src/main/java/org/openmrs/validator/ObsValidator.java
ibacher: I agree that we should be careful here. However, I think that the proposed change is _more_ harmful than the code in this PR, which implements something very intentional.

So, first, at a technical level, `obs.getId()` being `null` does not appropriately distinguish between a "new" obs and an updated obs. This is because updating an obs creates a new obs. It would be better to check that there is no _previous_ version.

However, the deeper problem is that this assume that our auto-calculated interpretations are the preferred interpretation. In fact, they are not. The preferred interpretation should always be the client-submitted interpretation, which is what this does. That is, if I run a test with a lab analyzer and the analyzer says the result is "CRITICAL", then the result should be regarded as "CRITICAL" regardless of what the reference range attached to the concept says. Concept reference ranges are, ultimately, not to be preferred over user-submitted values.

You are right that there is an issue where existing obs that have a new value will currently override the interpretation because we assume that the interpretation was auto-calculated. That assumption is also wrong. There's no way to resolve this without also tracking the provenance of the interpretation, but that's beyond the scope of this ticket.

ibacher: At best, we might propose if `previous_version` is not-null and the `value_numeric` has changed and the `interpretation` has not changed, then maybe we should re-interpret it, but this still puts us in a situation where the user-supplied interpretation (which should always be preferred if set) is overridden. The only workable solution is to somehow track the provenance of the interpretation.

PR #5431, opened 2025-10-27
TRUNK-6456 - Add inheritance strategy to ProviderRole jpa mappings

mseaton:
see https://issues.openmrs.org/browse/TRUNK-6456

See my comments on the ticket as to why I think this is needed (and why we may want to consider applying this strategy broadly across all JPA managed objects).  Interested in input from those that have worked with JPA a lot.

### DIFF
--- a/api/src/main/java/org/openmrs/ProviderRole.java
+++ b/api/src/main/java/org/openmrs/ProviderRole.java
@@ -16,6 +16,8 @@ import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
+import javax.persistence.Inheritance;
+import javax.persistence.InheritanceType;
 import javax.persistence.Table;
 import java.io.Serializable;
 
@@ -26,6 +28,7 @@ import java.io.Serializable;
  */
 @Entity
 @Table(name = "provider_role")
+@Inheritance(strategy = InheritanceType.JOINED)
 @Audited
 public class ProviderRole extends BaseOpenmrsMetadata implements Serializable {
 


dkayiwa:
I agree with applying this elsewhere because the default Single Table Inheritance strategy does not seem to apply for any of the openmrs use cases that i am aware of.

dkayiwa (approved):

ibacher:
@mseaton Does this still seem necessary? It look like we just missed merging it?

mseaton:
> @mseaton Does this still seem necessary? It look like we just missed merging it?

In the end, the providermanagement stuff was so tangled that I gave up on trying to support it at all in 2.8.x+ and so this didn't end up being something that I was blocked on, so I must have forgotten about it.  It still seems right to me in general (though I wouldn't necessarily limit it to provider, as I wrote on the ticket) if our general strategy for supporting subclassing of core objects via modules is by using the joined subclass approach (which I think it is, though this is what should be considered).

api/src/main/java/org/openmrs/ProviderRole.java
```diff
@@ -26,6 +28,7 @@
  */
 @Entity
 @Table(name = "provider_role")
+@Inheritance(strategy = InheritanceType.JOINED)
```
dkayiwa, line 31: Now that nobody is blocked on the 2.8.x line, the branch that needs these three lines most is master, and they are still missing there. I dropped a module style subclass of ProviderRole (its own table, `@PrimaryKeyJoinColumn` on `provider_role_id`) into master's mapping set, and the SessionFactory refuses to build at all: `org.hibernate.AnnotationException: Entity '...' is a subclass in a 'SINGLE_TABLE' hierarchy and may not be annotated '@Table'`. Hibernate 7 turns what 5.6 does quietly into a startup failure, so on 3.0 a module that subclasses ProviderRole would keep the platform from booting rather than just querying a DTYPE column that is not in the schema. Running the same subclass on 2.9.x, still Hibernate 5.6, gives the quiet 2.8.x behavior: a not-null DTYPE column and the subclass's own column pushed into `provider_role`.

Adding it on master and 2.9.x costs nothing, since the annotation does nothing until a subclass exists. With it applied to master's ProviderRole and no subclass on the classpath, the generated schema comes out identical to master today (Hibernate uses the single table persister for a root class with no subclasses whatever strategy is declared), and with the subclass present that subclass gets its own table keyed by `provider_role_id` while `provider_role` is left alone. Same three lines, with `jakarta.persistence` imports on master.

This is fine to merge for the 2.8.x line, it just should not be the only branch that gets it.

PR #5472, opened 2025-11-10
Trunk 5866 : OrderSet Domain - Switching from Hibernate Mappings to Annotations

sudhanshu-raj:

Converted OrderSet domain from Hibernate XML mappings to JPA annotations for better maintainability and alignment with modern Hibernate standards.

## Issue I worked on

see https://openmrs.atlassian.net/browse/TRUNK-5866

  No? -> execute above command


api/src/main/java/org/openmrs/OrderSet.java
```diff
@@ -36,12 +55,28 @@ public enum Operator {
...
 	}
 	
+	@Id
+	@GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "order_set_id_seq")
+	@GenericGenerator(
+		name = "order_set_id_seq",
+		strategy = "native",
+		parameters = @Parameter(name = "sequence", value = "order_set_order_set_id_seq")
+	)
+	@Column(name ="order_set_id" )
```
ManojLL: Use `GenerationType.IDENTITY` strategy primary key generation

sudhanshu-raj:
Hey , I did the change , it should build success now .

sudhanshu-raj:
Hey @dkayiwa  , can you please review this ?

dkayiwa:
Can you include a link to the ticket id as advised at https://openmrs.atlassian.net/wiki/display/docs/Pull+Request+Tips?

sudhanshu-raj:
Done !

api/src/main/java/org/openmrs/OrderSet.java
```diff
@@ -36,12 +60,28 @@ public enum Operator {
...
 	private Integer orderSetId;
 	
+	@Enumerated(EnumType.STRING)
+	@JdbcTypeCode(SqlTypes.VARCHAR)
+	@Column(name = "operator", nullable = false)
 	private Operator operator;
-	
+
+	@OneToMany(cascade = CascadeType.ALL, orphanRemoval = true)
```
dkayiwa, line 71: How did you deal with the lazy attribute?

sudhanshu-raj: I think  OneToMany has default fetch type lazy , ref : https://stackoverflow.com/questions/26601032/default-fetch-type-for-one-to-one-many-to-one-and-one-to-many-in-hibernate

api/src/main/java/org/openmrs/OrderSet.java
```diff
@@ -36,12 +60,28 @@ public enum Operator {
 		ALL, ONE, ANY
 	}
 	
+	@Id
+	@GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "order_set_id_seq")
```
dkayiwa: What do you think of using `GenerationType.IDENTITY`?

sudhanshu-raj: In orderset.hbm.xml sequence has been used , so that's why not using  identity  here .

dkayiwa: @wikumChamith what is our recommended pattern for this?

wikumChamith: We just use IDENTITY as the strategy

wikumChamith: https://github.com/openmrs/openmrs-core/pull/5216#discussion_r2257048387

sudhanshu-raj: > [#5216 (comment)](https://github.com/openmrs/openmrs-core/pull/5216#discussion_r2257048387)

Thanks , this is helpful

api/src/main/java/org/openmrs/OrderSet.java
```diff
@@ -21,7 +39,13 @@
  * 
  * @since 1.12
  */
+@Entity
+@Table(name = "order_set")
 @Audited
+@AssociationOverride(
+	name = "attributes",
```
dkayiwa, line 44: What is this override for?

sudhanshu-raj: This overrides the join column name for attributes field ,actually in the OrderSet's parent class which is BaseCustomizableMetadata it has join column name location_id but in orderset.hbm.xml it has order_set_id  join column name so it's overriding that join column name from it's parent.

dkayiwa: Did you evaluate using `AttributeOverride`?

sudhanshu-raj: AttributeOverride works  well for normal fields  not for relationship attributes so here AssociationOverride will work for relationships columns like OneToMany

dkayiwa: Even when you are not modifying a relationship but just renaming a filed like here?

sudhanshu-raj: yes , this just renaming the column name , it's just can change how it should be mapped to databased like joincolumn name or jointable name , but can't actually modify the relationship like onetomany to manytoone , fetch type etc

dkayiwa: Doesn't that then mean AttributeOverride is enough for this use case?

sudhanshu-raj: I don’t think `@AttributeOverride` works here. It’s used for simple attributes  for example, renaming a column on a basic field like `@Column(name = "idt") private String identifier;` or for fields inside an `@Embeddable.` It only changes how a single column is named or mapped.

But for relationships like `@OneToMany` or `@ManyToOne`, we use `@AssociationOverride`. That’s what lets us change the name of the join column or join table for an existing relation.
This is what I understand, please let me know if I’m missing something.

dkayiwa: What is the difference between this and https://github.com/openmrs/openmrs-core/blob/2.8.1/api/src/main/java/org/openmrs/Location.java#L52 ?

sudhanshu-raj: The @AttributeOverride(name = "attributes", column = @Column(name = "location_id")) annotation on the Location entity is not doing anything meaningful and can be safely removed. Why ?  because :@AttributeOverride only works for simple fields (String, Integer, Date, etc.), not relationships like @OneToMany

In BaseCustomizableMetadata, the attributes field is a relationship (@OneToMany), not a basic column:
```java
@OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, fetch = FetchType.LAZY)
@JoinColumn(name = "location_id")
private Set<A> attributes = new LinkedHashSet<>();
```

Since @AttributeOverride cannot override relationships, it gets silently ignored during entity mapping. The application runs fine without it which i also verified .

Ref : https://www.baeldung.com/jpa-attributeoverride
Recommendation: Remove the @AttributeOverride annotation from Location as it serves no purpose.

api/src/main/java/org/openmrs/OrderSet.java
dkayiwa: Do you think you can come up with a test in https://github.com/openmrs/openmrs-core/blob/master/api/src/test/java/org/openmrs/api/OrderSetServiceTest.java to catch this problem?

As in a test which would fail when using `AttributeOverride` but passes with `AssociationOverride`

jwnasambu:
@sudhanshu-raj kindly you have merge conflicts that requires your attention please!

sudhanshu-raj:
> @sudhanshu-raj kindly you have merge conflicts that requires your attention please!

Done !

api/src/main/java/org/openmrs/OrderSet.java
sudhanshu-raj: Hi @dkayiwa  , sorry I didn't chance to look into this . So I debugged it and found here both will work `AttributeOverride ` and `AssociationOverride` ,  even I not put one of these annotation, code will still work  because in schema we defined the correct column name but that's bad practice . So, if we use `AttributeOverride` it will create the two columns `order_set_id` and `location_id` but we need only one `order_set_id` which could be fixed by using  `AssociationOverride`.

api/src/test/java/org/openmrs/api/OrderServiceTest.java
```diff
@@ -2911,8 +2911,8 @@ public void saveOrder_shouldFailIfTheJavaTypeOfThePreviousOrderDoesNotMatch() th
 			    .addAnnotatedClass(ConceptReferenceTermMap.class)
 			    .addAnnotatedClass(ConceptReferenceTerm.class)
 				.addAnnotatedClass(ConceptAnswer.class)
-			    .addAnnotatedClass(ConceptDescription.class)
```
dkayiwa: Do you mind discarding these formatting changes?

sudhanshu-raj: Done !

api/src/main/java/org/openmrs/OrderSet.java
```diff
@@ -36,12 +58,23 @@ public enum Operator {
...
-	
+
+	@OneToMany(cascade = CascadeType.ALL, orphanRemoval = true)
+	@JoinColumn(name = "order_set_id", nullable = false)
+	@OrderColumn(name = "sequence_number")
 	private List<OrderSetMember> orderSetMembers;
-	
+
+	@ManyToOne
+	@JoinColumn(name = "category")
```
shivvani-r, line 77: `nullable` property is missing.

sudhanshu-raj: Isn't  default value will nullable only ? But for shake for readability I can add this.

PR #5546, opened 2025-12-01
Fix Path Traversal and XSS Security Vulnerabilities Detected by SpotBugs

DuoChen-317:
This PR fixes multiple security vulnerabilities detected by SpotBugs Security in the OpenMRS Core codebase.

Changes include:
- Added canonical path validation to prevent Path Traversal attacks in several file operations.
- Ensured path normalization and rejection of unsafe paths containing "..".
- Updated File constructors to use sanitized absolute paths.
- Added HTML escaping to all PrintWriter.write() calls that output request-derived content to prevent XSS.
- Ensured JSON responses use proper escaping or safe serialization.
- Removed unsafe string-based SQL concatenation and replaced with parameterized logic where applicable.

These changes eliminate all reported warnings under:
Security → Potential Path Traversal (file read)
Security → Potential XSS in Servlet
Security → Potential SQL Injection

### DIFF
--- a/web/src/main/java/org/openmrs/module/web/WebModuleUtil.java
+++ b/web/src/main/java/org/openmrs/module/web/WebModuleUtil.java
@@ -185,7 +185,13 @@ public class WebModuleUtil {
 						log.debug("Moving file from: {} to {}", name, absPath);
 						
 						// get the output file
-						File outFile = new File(absPath.toString().replace("/", File.separator));
+						String safePath = absPath.toString();
+
+						// basic path traversal protection
+						if (safePath.contains("..") || safePath.contains("\\") || safePath.contains("//")) {
+							throw new IllegalArgumentException("Unsafe file path detected: " + safePath);
+						}
+						File outFile = new File(safePath.replace("/", File.separator));
 						if (entry.isDirectory()) {
 							if (!outFile.exists()) {
 								outFile.mkdirs();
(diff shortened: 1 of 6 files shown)


web/src/main/java/org/openmrs/module/web/WebModuleUtil.java
```diff
@@ -782,7 +788,15 @@ public static void shutdownModules(ServletContext servletContext) {
 		
 		// clear the module messages
 		String messagesPath = realPath + "/WEB-INF/";
-		File folder = new File(messagesPath.replace("/", File.separator));
+		// Normalize separators
```
jwnasambu, line 791: Kindly the comments here don’t seem to add extra value so it would be better to remove them and rely on the code readability instead.

web/src/main/java/org/openmrs/module/web/WebModuleUtil.java
```diff
@@ -782,7 +788,15 @@ public static void shutdownModules(ServletContext servletContext) {
...
+		String safePath = messagesPath.replace("/", File.separator);
+
+		// Detect basic path traversal patterns
```
jwnasambu, line 794: Same here.

web/src/main/java/org/openmrs/module/web/WebModuleUtil.java
```diff
@@ -833,8 +847,22 @@ public static void stopModule(Module mod, ServletContext servletContext, boolean
 		String realPath = getRealPath(servletContext);
 		
 		// delete the web files from the webapp
-		String absPath = realPath + "/WEB-INF/view/module/" + moduleId;
-		File moduleWebFolder = new File(absPath.replace("/", File.separator));
+		String absPath = realPath + "/WEB-INF/web/module/" + moduleId;
+
+		// normalize path separator
```
jwnasambu, line 852: Same here. Kindly remove all the comments that starts with //.

web/src/main/java/org/openmrs/web/filter/update/UpdateFilter.java
```diff
@@ -274,8 +274,10 @@ else if (REVIEW_CHANGES.equals(page)) {
 				
 				addLogLinesToResponse(result);
 			}
-
```
jwnasambu: Is there away you can fix the format changes please?

jwnasambu (changes requested):

PR #5584, opened 2025-12-11
Trunk 6497 - Fix NPE for Integration Test Failures running on mysql or postgresql

sagarkhandagre998:
Problem - When running tests with -DuseInMemoryDatabase=false, many tests fail with NullPointerException at line 341 in getRuntimeProperties().

Error Pattern:
[ERROR] DrugOrderValidatorTest.<init>:53->BaseContextSensitiveTest.<init>:208 >BaseContextSensitiveTest.getRuntimeProperties:341 » NullPointer

Root Cause
In both BaseContextSensitiveTest classes (JUnit 4 and JUnit 5), but this error comes from Junit 5 when useInMemoryDatabase=false: The code reads system properties (databaseUrl, databaseUsername, databasePassword, databaseDriver, databaseDialect) that may be null. These null values are passed directly to  runtimeProperties.setProperty(). Later access to these properties causes NullPointerException.

The issue occurs because:
Containers.ensureDatabaseRunning() may not have set the system properties yet, or The properties weren't set manually, or There was a failure in the container initialization.

Solution
1. Added validation in both test base classes: Null check for runtimeProperties: ensure it's not null after loading from TestUtil.getRuntimeProperties().
2. Validation of system properties: check all required database properties before use.
3. Clear error handling: throw IllegalStateException with a helpful message instead of NPE.

Suggested Entire Fix 

1. The NPE is resolved by validation, but the root cause remains: missing system properties (databaseUrl, databaseUsername, databasePassword, databaseDriver, databaseDialect) when useInMemoryDatabase=false.

2. To fully fix the error, we must ensure those properties are set before tests start—either by letting Containers.ensureDatabaseRunning() set them, or by passing them explicitly on the Maven command line.

3. If using MySQL, swap the driver/dialect and URL accordingly.

4. If Containers.ensureDatabaseRunning() is supposed to set these automatically, we should verify why it isn’t running or isn’t able to set them in the current environment (e.g., missing -Ddatabase flag, Testcontainers disabled, or network/permission issues).

### DIFF
--- a/api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
+++ b/api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
@@ -311,10 +311,7 @@ public abstract class BaseContextSensitiveTest extends AbstractJUnit4SpringConte
 	 * @return Properties runtime
 	 */
 	public Properties getRuntimeProperties() {
-
-		System.err.println("DEBUG: getRuntimeProperties() ENTER; runtimeProperties=" + runtimeProperties);
-
-		
+	
 		// cache the properties for subsequent calls
 		if (runtimeProperties == null)
 			runtimeProperties = TestUtil.getRuntimeProperties(getWebappName());
--- a/api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
+++ b/api/src/test/java/org/openmrs/test/BaseContextSensitiveTest.java
@@ -311,10 +311,19 @@ public abstract class BaseContextSensitiveTest extends AbstractJUnit4SpringConte
 	 * @return Properties runtime
 	 */
 	public Properties getRuntimeProperties() {
+
+		System.err.println("DEBUG: getRuntimeProperties() ENTER; runtimeProperties=" + runtimeProperties);
+
 		
 		// cache the properties for subsequent calls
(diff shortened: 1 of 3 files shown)


sagarkhandagre998:
Need Clarification - We have find the breaking point and tackled that by adding validation but the main error is from System properties. We need to check why the database is not running  or  isn’t able to set the database props in the current environment (Containers.java) .

dkayiwa:
> We need to check why the database is not running or isn’t able to set the database props in the current environment (Containers.java) .

Can you investigate that too?

sagarkhandagre998:
Sure I will do that.

sagarkhandagre998:
> > We need to check why the database is not running or isn’t able to set the database props in the current environment (Containers.java) .
> 
> Can you investigate that too?

The actual error is comming from Containers.java where test container is not able to connect with docker, if the docker is not started or running so thats why it was not able to set the system props results in NPE but we have implements the validation for NPE and now i have also implmented the validation checks for Containers.java to handle the connection with docker. After starting the docker locally the test container is succesfully connected with docker.

sagarkhandagre998:
@dkayiwa Improved Exceptional handling in Containers.java to get the actual error while running the test.

PR #5588, opened 2025-12-11
TRUNK-5466: Add supportsVisits attribute to Location

Akshat-coder2106:
### Description
I have implemented the changes requested in TRUNK-5466 to support the "Visit Location" feature in the core domain model.

**Changes:**
1.  **Java Domain:** Added `private Boolean supportsVisits` field to `Location.java` with Getter/Setter methods.
2.  **Database:** Added a Liquibase changeset to create the `supports_visits` column in the `location` table (default: false).

### Ticket
https://issues.openmrs.org/browse/TRUNK-5466

### Testing
- Ran `mvn clean install -DskipTests` to verify build.
- Ran `mvn test -Dtest="Location*"` to verify unit tests for the modified class.
- Verified that the Liquibase changeset is valid xml.


README.md
```diff
@@ -309,3 +309,5 @@ Talk to us on [OpenMRS Talk](https://talk.openmrs.org/)
 
 [MPL 2.0 w/ HD](http://openmrs.org/license/) © [OpenMRS Inc.](http://www.openmrs.org/)
 
+'HELLO I AM AKSHAT AGRAWAL'
```
vimla01: there’s a small accidental change in **README.md** that should be removed.

adarshiiit0117 (commented):
Using Thread.sleep(10) in a unit test can lead to inconsistent results, especially in slower CI environments. 

Can we avoid timing-based checks here? For example, we could update a field that triggers dateChanged directly, mock the clock, or use a controllable time source.

adarshiiit0117 (commented):
Minor suggestion: instead of relying on after(initialDateChanged), we could assert that dateChanged is updated when calling savePatient, without depending on system time advancing.

PR #5595, opened 2025-12-14
TRUNK-6492: Add filtering methods to Encounter for voided orders and diagnoses.

sagarkhandagre998:
### **Description of what I changed -** 

This PR addresses an inconsistency in the Encounter API where getOrders() and getDiagnoses() return both active and voided items by default, unlike newer Encounter properties such as Conditions and Allergies.

To improve API flexibility while preserving backward compatibility, this change introduces overloaded methods that allow callers to explicitly control whether voided items are included:

**getOrders(boolean includeVoided)
getDiagnoses(boolean includeVoided)**

The existing getOrders() and getDiagnoses() methods were left unchanged to avoid breaking legacy code that depends on direct access to the underlying Hibernate-managed collections.

This allows client code and downstream consumers (including REST) to retrieve only non-voided orders and diagnoses without duplicating filtering logic.
### 
**Issue I worked on**
https://issues.openmrs.org/browse/TRUNK-6418

### Checklist: I completed these to help reviewers :)

- [x]  I have added tests to cover my changes.

- > New unit tests verify that:
-      getOrders(false) excludes voided orders
-      getDiagnoses(false) excludes voided diagnoses

- [x] - Existing behavior of getOrders() and getDiagnoses() remains unchanged.
- [x] - I ran mvn clean package right before creating this pull request and added all formatting changes to my commit.
- [x] - All new and existing tests passed.
- [x] - My pull request is based on the latest changes of the master branch.


api/src/main/java/org/openmrs/Encounter.java
```diff
@@ -487,6 +510,28 @@ public Set<Diagnosis> getDiagnoses() {
...
+	 * @return The set of diagnoses, or an empty set if there are no diagnoses to return.
+	 * @since 3.0.0
+	 */
+	public Set<Diagnosis> getDiagnoses(boolean includeVoided) {
+		if (diagnoses == null) {
+			return new LinkedHashSet<>();
+		}
+		
+		if (includeVoided) {
+			return getDiagnoses();
```
jwnasambu, line 512: Kindly using  getDiagnoses() while the rest of the method accesses diagnoses directly wont' it create inconsistent behavior? It will be better to use a single source of truth (getDiagnoses() assigned once at the top. Kindly @wikumChamith correct me please!

sagarkhandagre998: @jwnasambu did u check the requirement of this pull request ?

wikumChamith:
Can we stick to our existing PR template?

sagarkhandagre998:
> Can we stick to our existing PR template?

Sure @wikumChamith It's my bed.

api/src/main/java/org/openmrs/Encounter.java
```diff
@@ -422,7 +422,30 @@ public Set<Order> getOrders() {
...
+			return new LinkedHashSet<>();
+		}
+		
+		if (includeVoided) {
+			return getOrders();
+		}
+		
+		return orders.stream()
+			.filter(o -> !o.getVoided())
+			.collect(Collectors.toSet());
```
Binayak490-cyber: @sagarkhandagre998, includeVoided=false returns a plain HashSet (unordered) while includeVoided=true returns the original LinkedHashSet (ordered) — the return type is inconsistent between the two branches. Should use Collectors.toCollection(LinkedHashSet::new) to maintain consistent ordering.

api/src/main/java/org/openmrs/Encounter.java
Binayak490-cyber, line 512: The null check accesses the raw diagnoses field directly but the includeVoided=true branch delegates to getDiagnoses() — mixed access from two different sources. Should assign getDiagnoses() once at the top as a single source of truth (as flagged by @jwnasambu ).

api/src/test/java/org/openmrs/EncounterTest.java
```diff
@@ -1422,6 +1441,25 @@ public void hasDiagnosis_shouldReturnFalseIfEncounterDoesNotHaveDiagnosis(){
...
+		
+		Set<Diagnosis> diagnoses = new LinkedHashSet<>();
+		diagnoses.add(active);
+		diagnoses.add(voided);
+		encounter.setDiagnoses(diagnoses);
+		
+		assertEquals(1, encounter.getDiagnoses(false).size());
+		assertEquals(2, encounter.getDiagnoses(true).size());
+	}
+
```
Binayak490-cyber, line 1458: @sagarkhandagre998, No test covers the null collection edge case that the null guard in both methods handles — the guard exists but is completely untested.

PR #5630, opened 2025-12-28
Add getters and setters for closedLow and closedHigh in DoubleRange

airajena:
## Summary
Added the missing getters and setters for `closedLow` and `closedHigh` fields in `DoubleRange.java` as requested by the TODO comment.

## Changes
- Added `isClosedLow()` and `setClosedLow(boolean)` methods
- Added `isClosedHigh()` and `setClosedHigh(boolean)` methods
- Removed TODO comment

## Why
The `DoubleRange` class had a TODO requesting these accessors. Currently the `closedLow` and `closedHigh` fields could only be set via direct access or defaults, making it impossible to configure interval boundary behavior from external code.

## Testing
The existing tests continue to pass. The new methods follow the same pattern as existing `getLow/setLow` and `getHigh/setHigh` accessors.

### DIFF
--- a/api/src/main/java/org/openmrs/util/DoubleRange.java
+++ b/api/src/main/java/org/openmrs/util/DoubleRange.java
@@ -23,7 +23,7 @@ public class DoubleRange implements Comparable<DoubleRange> {
 	
 	private Double high;
 	
-	private boolean closedLow = true; //TODO: add setters and getters for these
+	private boolean closedLow = true;
 	
 	private boolean closedHigh = false;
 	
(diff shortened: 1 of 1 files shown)


adarshiiit0117 (commented):
Looks good, thanks for cleaning up the TODO and adding proper accessors.

I have a small question or suggestion. Would it make sense to use Lombok annotations, like @Getter and @ Setter, instead of manual getters and setters? This could help with consistency and reduce repetitive code.

It's totally fine as-is if Lombok isn’t preferred in this module. I just wanted to check.

PR #5662, opened 2026-01-12
TRUNK-6266: Order drug search results from simple to complex formulat...

akash8843:
…ions

## Description
This change updates the ordering of drug search results returned by the
existing search mechanism. As discussed in TRUNK-6266, the goal is not to
introduce a new search, but to improve how results are scored and ordered.

Currently, combination drugs can appear before simpler formulations.
This update adjusts the ordering so that simpler (single-ingredient)
formulations are prioritized ahead of more complex combination drugs,
which better matches expected clinical usage.

Alphabetical ordering alone is not sufficient in these cases (e.g.
Paracetamol should appear before Ibuprofen + Paracetamol), so this change
focuses on formulation complexity rather than name sorting.

## Changes
- Adjusted result ordering in `getDrugs(String phrase)`
- Simple (single-ingredient) drugs are listed before combination drugs

## Related ticket
https://issues.openmrs.org/browse/TRUNK-6266

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.Comparator;
 
 import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaQuery;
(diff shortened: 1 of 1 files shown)


sudhanshu-raj:
Hey @akash8843, are you sure about this change? Have you tested it?
This approach of sorting by ingredients won't work because the list of ingredients for a drug is zero. You can also see in the API response that it doesn't return any ingredients list, and if you check the database for drug_ingredient, there's nothing there at least not in the demo data. So most probably, we should apply the sorting logic on the drug name itself.

Please correct me if I’m mistaken, @ibacher, @dkayiwa.

akash8843:
> Hey @akash8843, are you sure about this change? Have you tested it? This approach of sorting by ingredients won't work because the list of ingredients for a drug is zero. You can also see in the API response that it doesn't return any ingredients list, and if you check the database for drug_ingredient, there's nothing there at least not in the demo data. So most probably, we should apply the sorting logic on the drug name itself.
> 
> Please correct me if I’m mistaken, @ibacher, @dkayiwa.

Thanks for pointing this out @sudhanshu-raj.

You’re right — I rechecked and the ingredients list isn’t available in the API response, and the `drug_ingredient` table also seems empty in the demo data. So sorting based on ingredients won’t work here.

A name-based approach (prioritizing single-drug names over combinations like "X + Y") sounds more reasonable. I’ll wait for confirmation or further input from @ibacher / @dkayiwa before updating the implementation.

Let me know if this makes sense.

PR #5683, opened 2026-01-19
TRUNK-5979: Guard web startup against late servlet context modification

akash8843:
## Description

While looking into TRUNK-5979, I noticed that during the first startup
OpenMRS can attempt to register module web components after the servlet
context has already been initialized.

Since the Servlet specification does not allow late registration of
filters or servlets, this results in an IllegalStateException being logged,
even though the application continues to start normally.

This change adds a small defensive check before starting the module web
components. If the servlet context is no longer modifiable, the startup
step is skipped and a clear warning is logged instead.

This avoids misleading startup errors while keeping the existing behavior
unchanged.

## Related Issue
https://issues.openmrs.org/browse/TRUNK-5979

## Checklist
- [x] Code follows the existing project style
- [x] Web module build was run locally
- [x] No functional behavior was changed

### DIFF
--- a/web/src/main/java/org/openmrs/web/Listener.java
+++ b/web/src/main/java/org/openmrs/web/Listener.java
@@ -83,6 +83,17 @@ public final class Listener extends ContextLoader implements ServletContextListe
 	
 	private static final org.slf4j.Logger log = LoggerFactory.getLogger(Listener.class);
 	
+	private boolean isServletContextModifiable(ServletContext servletContext) {
+	try {
+		servletContext.getFilterRegistrations();
+		return true;
+	}
+	catch (IllegalStateException e) {
+		return false;
+	}
+}
+
+
 	private static boolean runtimePropertiesFound = false;
 	
 	private static Throwable errorAtStartup = null;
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -83,6 +83,17 @@ public final class Listener extends ContextLoader implements ServletContextListe
 	
 	private static final org.slf4j.Logger log = LoggerFactory.getLogger(Listener.class);
 	
+	private boolean isServletContextModifiable(ServletContext servletContext) {
+	try {
+		servletContext.getFilterRegistrations();
```
siddhantappuni-ops, line 88: Interesting approach to check modifiability. I notice getFilterRegistrations() is being called purely as a side-effect-free probe its return value is discarded. Would it be clearer to add a brief inline comment like // probe call if context is frozen, this throws IllegalStateException so future readers immediately understand the intent? Otherwise someone might think the return value was accidentally ignored.

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -253,8 +264,15 @@ public void contextInitialized(ServletContextEvent event) {
 				servletContext.setAttribute(WebApplicationContext.ROOT_WEB_APPLICATION_CONTEXT_ATTRIBUTE, context);
 				log.debug("Done refreshing WAC");
 				
-				WebDaemon.startOpenmrs(event.getServletContext());
-			} else {
+				if (!isServletContextModifiable(servletContext)) {
+		        log.warn("Skipping module web component registration because the servlet context "
```
siddhantappuni-ops, line 268: The warning message is clear and helpful. One question since this is expected during first startup (as the message says), would log.info be more appropriate than log.warn? Warn typically signals something that might need attention, but if this is a known expected scenario, info-level might reduce noise in logs for ops teams monitoring for actual warnings.

jwnasambu (commented):
@akash8843 Thanks for the progress. Kindly do you mind fixing the merge conflicts on this PR please?

PR #5700, opened 2026-01-23
[TRUNK-6521] Add contract verification test for PersonService.getPersonByUuid

MonotonousHarsh:
### JIRA Ticket
https://issues.openmrs.org/browse/TRUNK-6521

### Description
**The Issue:**
The `PersonService` contract implies that querying `getPersonByUuid` with a non-existent UUID should return `null`. However, `PersonServiceTest` currently lacks a test case to enforce this "sad path" behavior.

**Risk:**
Without this coverage, a future regression could cause the service to throw an unhandled exception (e.g., `ObjectNotFoundException`) instead of returning `null`, potentially breaking downstream modules that expect a null check.

**My Solution:**
* Added `getPersonByUuid_shouldReturnNullIfNoPersonFoundWithUuid` to `PersonServiceTest.java`.
* Verifies that the API handles invalid UUIDs gracefully by returning `null` rather than crashing.

### Dev Checklist
- [x] Added unit tests for new functionality
- [x] Ran `mvn test -Dtest=PersonServiceTest` (147 tests passed)

### DIFF
--- a/api/src/test/java/org/openmrs/api/PersonServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/PersonServiceTest.java
@@ -2459,10 +2459,15 @@ public class PersonServiceTest extends BaseContextSensitiveTest {
 
 	@Test
 	public void getPersonByUuid_shouldReturnNullIfNoPersonFoundWithUuid() {
-	
+
+		Person p = new Person();
+   	    p.setGender("M");
+    	p.addName(new PersonName("Test", "User", "Person"));
+    	personService.savePerson(p);
+		
+		assertNotNull("Should find existing person", personService.getPersonByUuid(p.getUuid()));
 		String nonExistentUuid = "random-uuid-that-does-not-exist";
-		Person person = personService.getPersonByUuid(nonExistentUuid);
-		assertNull(person);
+    	assertNull("Should return null for non-existent UUID", personService.getPersonByUuid(nonExistentUuid));
 	}
 	
 }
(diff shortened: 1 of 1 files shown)


api/src/test/java/org/openmrs/api/PersonServiceTest.java
```diff
@@ -2451,5 +2451,18 @@ public void savePersonAddress_shouldNotBeNullOnReturningAsavedPersonAdress() thr
...
+
+	/**
+	 * @see PersonService#getPersonByUuid(String) 
+	 */
+
+	@Test
+	public void getPersonByUuid_shouldReturnNullIfNoPersonFoundWithUuid() {
+	
+		String nonExistentUuid = "random-uuid-that-does-not-exist";
+		Person person = personService.getPersonByUuid(nonExistentUuid);
```
saidworks: Even if the person exists in db the response will be null, probably mock certain data in repository then try to verify it with same call.

I hope this helps

saidworks (changes requested):

api/src/test/java/org/openmrs/api/PersonServiceTest.java
MonotonousHarsh: Thanks for the feedback! You are right the previous test didn't distinguish between 'empty DB' and 'correct null handling'. I have updated the test to first save/retrieve a real person (sanity check) and then verify the null return for the non-existent UUID.

PR #5711, opened 2026-01-27
DOC: Clarify validation responsibility in PatientValidator

akshat-s99:
PatientValidator relies on PersonValidator for validation of person-related
attributes such as names, but this responsibility was not clearly documented
and an outdated TODO suggested otherwise.

This PR clarifies the validation responsibility to avoid confusion for future
contributors.


adarshiiit0117 (commented):
Thanks for clarifying the JavaDoc here 👍

Just a small question ,the previous comment mentioned a TODO about validating all Patient fields. Has that been addressed elsewhere, or was it removed intentionally to avoid confusion?

Also, since this validator focuses only on patient-specific rules (like identifier validation) and delegates person-related attributes to PersonValidator, would it help to briefly clarify which responsibilities belong to each validator for future maintainers?

PR #5728, opened 2026-01-30
TRUNK-6529: Ensure patient has all required identifiers

Binayak490-cyber:
This PR enhances PatientValidator to ensure that a patient must have identifiers for all required and non-retired PatientIdentifierTypes before validation passes.

Changes included:

 Added validation logic to check that every required, non-retired identifier type has a corresponding non-voided patient identifier.

 Updated and added unit tests to cover the following scenarios:

 Patient missing a required identifier (validation fails)

 Patient with required identifier present (validation passes)

 Required identifier is voided (validation fails)

 Required identifier type is retired (validation ignored)

This change ensures data consistency and aligns patient validation with the configured identifier requirements.All tests pass locally.

### DIFF
--- a/api/src/main/java/org/openmrs/validator/PatientValidator.java
+++ b/api/src/main/java/org/openmrs/validator/PatientValidator.java
@@ -10,10 +10,15 @@
 package org.openmrs.validator;
 
 import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 
 import org.openmrs.Patient;
 import org.openmrs.PatientIdentifier;
+import org.openmrs.PatientIdentifierType;
 import org.openmrs.annotation.Handler;
+import org.openmrs.api.context.Context;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -57,7 +62,14 @@ public class PatientValidator extends PersonValidator {
 	 * voided patients<br/>
 	 * <strong>Should</strong> not fail when patient has only one identifier and its not preferred<br/>
 	 * <strong>Should</strong> pass validation if field lengths are correct<br/>
-	 * <strong>Should</strong> fail validation if field lengths are not correct
+	 * <strong>Should</strong> fail validation if field lengths are not correct<br/>
+	 * <strong>Should</strong> fail validation if a required patient identifier is missing<br/>
+	 * <strong>Should</strong> pass validation if a required patient identifier is present<br/>
+	 * <strong>Should</strong> fail validation if the only identifier for a required type is voided<br/>
+	 * <strong>Should</strong> pass validation if a voided required identifier is replaced by an active
+	 * one<br/>
+	 * <strong>Should</strong> ignore retired identifier types when checking required identifiers<br/>
+	 * <strong>Should</strong> not fail validation for voided patients missing required identifiers
 	 *
 	 * @param obj The patient to validate.
 	 * @param errors Errors
@@ -78,30 +90,71 @@ public class PatientValidator extends PersonValidator {
 
 		ValidationUtils.rejectIfEmptyOrWhitespace(errors, "gender", "Person.gender.required");
 
-		// Make sure they chose a preferred ID
-		Boolean preferredIdentifierChosen = false;
-		//Voided patients have only voided identifiers since they were voided with the patient,
-		//so get all otherwise get the active ones
-		Collection<PatientIdentifier> identifiers = patient.getVoided() ? patient.getIdentifiers()
+		validatePreferredIdentifier(patient, errors);
+
+		if (!Boolean.TRUE.equals(patient.getVoided())) {
+			validateRequiredIdentifiers(patient, errors);
+		}
+
+		validatePatientIdentifiers(patient, errors);
+
+		ValidateUtil.validateFieldLengths(errors, obj.getClass(), "voidReason");
+	}
+
+	private void validatePreferredIdentifier(Patient patient, Errors errors) {
+		boolean preferredIdentifierChosen = false;
+
+		Collection<PatientIdentifier> identifiers = Boolean.TRUE.equals(patient.getVoided()) ? patient.getIdentifiers()
 		        : patient.getActiveIdentifiers();
+
 		for (PatientIdentifier pi : identifiers) {
-			if (pi.getPreferred()) {
+			if (Boolean.TRUE.equals(pi.getPreferred())) {
 				preferredIdentifierChosen = true;
 			}
 		}
+
 		if (!preferredIdentifierChosen && identifiers.size() != 1) {
 			errors.reject("error.preferredIdentifier");
 		}
-		int index = 0;
-		if (!errors.hasErrors() && patient.getIdentifiers() != null) {
-			// Validate PatientIdentifers
-			for (PatientIdentifier identifier : patient.getIdentifiers()) {
-				errors.pushNestedPath("identifiers[" + index + "]");
-				patientIdentifierValidator.validate(identifier, errors);
-				errors.popNestedPath();
-				index++;
+	}
+
+	private void validateRequiredIdentifiers(Patient patient, Errors errors) {
+		Collection<PatientIdentifierType> identifierTypes = Context.getPatientService().getAllPatientIdentifierTypes(false);
+
+		Set<PatientIdentifierType> requiredTypes = new HashSet<>();
+
+		for (PatientIdentifierType type : identifierTypes) {
+			if (Boolean.TRUE.equals(type.getRequired())) {
+				requiredTypes.add(type);
 			}
 		}
-		ValidateUtil.validateFieldLengths(errors, obj.getClass(), "voidReason");
+
+		for (PatientIdentifier pi : patient.getActiveIdentifiers()) {
+			if (pi.getIdentifierType() != null) {
+				requiredTypes.remove(pi.getIdentifierType());
+			}
+		}
+
+		if (!requiredTypes.isEmpty()) {
+			List<String> missingRequiredIdentifiers = requiredTypes.stream().map(PatientIdentifierType::getName).toList();
+
+			errors.rejectValue("identifiers", "Patient.missingRequiredIdentifier",
+			    new Object[] { String.join(", ", missingRequiredIdentifiers) }, null);
+		}
+	}
+
+	private void validatePatientIdentifiers(Patient patient, Errors errors) {
+		if (errors.hasErrors() || patient.getIdentifiers() == null) {
+			return;
+		}
+
+		int index = 0;
+
+		for (PatientIdentifier identifier : patient.getIdentifiers()) {
+			errors.pushNestedPath("identifiers[" + index + "]");
+			patientIdentifierValidator.validate(identifier, errors);
+			errors.popNestedPath();
+			index++;
+		}
 	}
 }
--- a/api/src/main/java/org/openmrs/validator/PatientValidator.java
+++ b/api/src/main/java/org/openmrs/validator/PatientValidator.java
@@ -18,7 +18,7 @@ import org.openmrs.Patient;
 import org.openmrs.PatientIdentifier;
 import org.openmrs.PatientIdentifierType;
 import org.openmrs.annotation.Handler;
-import org.openmrs.api.context.Context;
+import org.openmrs.api.PatientService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -33,8 +33,15 @@ public class PatientValidator extends PersonValidator {
 
 	private static final Logger log = LoggerFactory.getLogger(PersonNameValidator.class);
 
+	private final PatientIdentifierValidator patientIdentifierValidator;
+
+	private final PatientService patientService;
+
 	@Autowired
-	private PatientIdentifierValidator patientIdentifierValidator;
+	public PatientValidator(PatientIdentifierValidator patientIdentifierValidator, PatientService patientService) {
+		this.patientIdentifierValidator = patientIdentifierValidator;
+		this.patientService = patientService;
+	}
 
 	/**
 	 * Returns whether or not this validator supports validating a given class.
@@ -90,54 +97,43 @@ public class PatientValidator extends PersonValidator {
 
 		ValidationUtils.rejectIfEmptyOrWhitespace(errors, "gender", "Person.gender.required");
 
-		validatePreferredIdentifier(patient, errors);
-
-		if (!Boolean.TRUE.equals(patient.getVoided())) {
-			validateRequiredIdentifiers(patient, errors);
-		}
+		validatePreferredAndRequiredIdentifiers(patient, errors);
 
 		validatePatientIdentifiers(patient, errors);
 
 		ValidateUtil.validateFieldLengths(errors, obj.getClass(), "voidReason");
 	}
 
-	private void validatePreferredIdentifier(Patient patient, Errors errors) {
-		boolean preferredIdentifierChosen = false;
-
-		Collection<PatientIdentifier> identifiers = Boolean.TRUE.equals(patient.getVoided()) ? patient.getIdentifiers()
+	private void validatePreferredAndRequiredIdentifiers(Patient patient, Errors errors) {
+		boolean isVoidedPatient = Boolean.TRUE.equals(patient.getVoided());
+		Collection<PatientIdentifier> identifiers = isVoidedPatient ? patient.getIdentifiers()
 		        : patient.getActiveIdentifiers();
 
+		Set<PatientIdentifierType> requiredTypes = new HashSet<>();
+		if (!isVoidedPatient) {
+			for (PatientIdentifierType type : patientService.getAllPatientIdentifierTypes(false)) {
+				if (Boolean.TRUE.equals(type.getRequired())) {
+					requiredTypes.add(type);
+				}
+			}
+		}
+
+		boolean preferredIdentifierChosen = false;
 		for (PatientIdentifier pi : identifiers) {
 			if (Boolean.TRUE.equals(pi.getPreferred())) {
 				preferredIdentifierChosen = true;
+			}
+			if (pi.getIdentifierType() != null) {
+				requiredTypes.remove(pi.getIdentifierType());
 			}
 		}
 
 		if (!preferredIdentifierChosen && identifiers.size() != 1) {
 			errors.reject("error.preferredIdentifier");
 		}
-	}
-
-	private void validateRequiredIdentifiers(Patient patient, Errors errors) {
-		Collection<PatientIdentifierType> identifierTypes = Context.getPatientService().getAllPatientIdentifierTypes(false);
-
-		Set<PatientIdentifierType> requiredTypes = new HashSet<>();
-
-		for (PatientIdentifierType type : identifierTypes) {
-			if (Boolean.TRUE.equals(type.getRequired())) {
-				requiredTypes.add(type);
-			}
-		}
-
-		for (PatientIdentifier pi : patient.getActiveIdentifiers()) {
-			if (pi.getIdentifierType() != null) {
-				requiredTypes.remove(pi.getIdentifierType());
-			}
-		}
 
 		if (!requiredTypes.isEmpty()) {
 			List<String> missingRequiredIdentifiers = requiredTypes.stream().map(PatientIdentifierType::getName).toList();
-
 			errors.rejectValue("identifiers", "Patient.missingRequiredIdentifier",
 			    new Object[] { String.join(", ", missingRequiredIdentifiers) }, null);
 		}
--- a/api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
+++ b/api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
@@ -10,6 +10,7 @@
 package org.openmrs.validator;
 
 import java.util.Date;
+import java.util.List;
 import java.util.Set;
 
 import org.junit.jupiter.api.Test;
@@ -26,6 +27,7 @@ import org.springframework.validation.BindException;
 import org.springframework.validation.Errors;
 import org.springframework.validation.Validator;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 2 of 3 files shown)


Binayak490-cyber:
Hi @ibacher,

This PR addresses TRUNK-6529 by ensuring that a patient must have identifiers for all required, non-retired PatientIdentifierTypes before validation passes.

I’ve added validation logic and updated unit tests to cover missing, present, voided, and retired required identifiers.

All tests pass locally (mvn clean test).

Please review when convenient.Thanks!

Binayak490-cyber:
Hi @ibacher,
Just a gentle reminder regarding this PR for TRUNK-6529: Ensure patient has all required identifiers.
The fix ensures validation for required patient identifiers and all tests are passing.
Please let me know if any changes or improvements are needed from my side.

kmvipin:
@Binayak490-cyber, kindly mention the ticket which is related to this PR!

Binayak490-cyber:
> @Binayak490-cyber, can you please mention the ticket which is related to this PR!

@kmvipin, it is the issue ticket TRUNK-6529 i.e; https://openmrs.atlassian.net/browse/TRUNK-6529?focusedCommentId=184800

Binayak490-cyber:
Hi @ibacher,
Quick follow-up on TRUNK-6529.
This PR introduces a new method along with tests that validate ensuring patients have all required identifiers.
Would appreciate a review.
Thanks!

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -90,6 +91,27 @@ public void validate(Object obj, Errors errors) {
 		if (!preferredIdentifierChosen && identifiers.size() != 1) {
 			errors.reject("error.preferredIdentifier");
 		}
+		Collection<PatientIdentifierType> identifierTypes = Context.getPatientService().getAllPatientIdentifierTypes(false);
+
+    	for (PatientIdentifierType type : identifierTypes) {
+        	if (Boolean.TRUE.equals(type.getRequired()) && !type.getRetired()) {
```
ibacher: We already know that these types are not retired, so the check here is redundant.

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -90,6 +91,27 @@ public void validate(Object obj, Errors errors) {
...
+                	if (!pi.getVoided() && pi.getIdentifierType() != null && pi.getIdentifierType().getId().equals(type.getId())) {
+                    	found = true;
+                    	break;
+                	}
+            	}
+
+            	if (!found) {
+                	errors.rejectValue(
+                        "identifiers",
+                        "Patient.missingRequiredIdentifier",
```
ibacher: You need to actually add this string to the message.properties file.

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -90,6 +91,27 @@ public void validate(Object obj, Errors errors) {
...
+                        "Missing required patient identifier: " + type.getName()
```
ibacher: Please lookup how you add parameters to a Spring errors object. Setting the default string isn't the right thing to do.

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -90,6 +91,27 @@ public void validate(Object obj, Errors errors) {
 		if (!preferredIdentifierChosen && identifiers.size() != 1) {
 			errors.reject("error.preferredIdentifier");
 		}
+		Collection<PatientIdentifierType> identifierTypes = Context.getPatientService().getAllPatientIdentifierTypes(false);
+
+    	for (PatientIdentifierType type : identifierTypes) {
+        	if (Boolean.TRUE.equals(type.getRequired()) && !type.getRetired()) {
+				boolean found = false;
```
ibacher: This isn't a case where we want to fail-fast. Instead we _want_ to collect _all_ missing required identifiers and send that back to the user as a single string.

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
```diff
@@ -12,6 +12,7 @@
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.Assert.*;
```
ibacher: We do not permit `*` imports.

ibacher (changes requested):
Thanks! This needs some work. Also, instead of the O(n^2) algorithm you might be able to do something clever with sets to just get the difference between the "set of required identifiers" and the "set of identifiers the patient has", but it's not a big deal as the set of identifiers is normally very small.

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: Okk @ibacher, I removed the !type.getRetired() check as getAllPatientIdentifierTypes(false) already guarantees that only non-retired types are returned, making the additional check redundant and misleading.

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: @ibacher, I’ve added Patient.missingRequiredIdentifier to messages.properties and pushed the update.

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: @ibacher, I have updated to use Spring Errors.rejectValue with message parameters and added the message to messages.properties.
Removed the default message as suggested.

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: okk @ibacher, for not getting fail-fast, i have updated the validator to collect all missing required identifier types and report them back as a single validation error.
The error now uses Spring’s Errors.rejectValue with parameters.

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
Binayak490-cyber: @ibacher, yepp i mistakenly added the full imported package which is not at all required for my added tests also and therefore, i have removed it.

Binayak490-cyber:
Hi @ibacher,
Any further changes required in this PR.
Waiting for further review!

Binayak490-cyber:
Hi @ibacher,
Any update for this PR...!

Binayak490-cyber:
Hi @ibacher,
Any update for this PR...!

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
```diff
@@ -195,4 +195,158 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+    	pAddress.setAddress1("123 My street");
+    	pAddress.setAddress2("Apt 402");
+    	pAddress.setCityVillage("Anywhere city");
+    	pAddress.setCountry("Some Country");
+   	 	patient.addAddress(pAddress);
+
+    	Errors errors = new BindException(patient, "patient");
+    	validator.validate(patient, errors);
+
+    	assertTrue(errors.hasErrors());
```
ibacher: Can we add more specific assertions? This just tells us that there was some validation error, but we're not verifying that its an error in anyway related to the patient identifier.

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
```diff
@@ -195,4 +195,158 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
 		
 		assertTrue(errors.hasFieldErrors("voidReason"));
 	}
+
+	/**
+	* @see PatientValidator#validate(Object,Errors)
+ 	*/
+	@Test
+	public void validate_shouldFailIfRequiredIdentifierIsMissing() {
```
ibacher, line 205: I think we can also add tests to cover:

1. A required identifier and a non-required identifier
2. A voided required identifier and a non-voided required identifier (simulating the case where a patients identifier was changed)
3. Ensuring that this _doesn't_ block edits to voided patients.

ibacher (changes requested):
Thanks @Binayak490-cyber! Looking much better! I think this just needs improved test coverage and it will be good to go.

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
Binayak490-cyber: okk @ibacher, i am adding it...!

Binayak490-cyber: Yes @ibacher! More specific assertions has been implemented in the last push.

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
Binayak490-cyber: Yup @ibacher! The demanded tests has been added...!

Binayak490-cyber:
@ibacher, any other changes to be made in this PR...!

Binayak490-cyber:
Hi @ibacher,
Any further updates for this PR...???

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -80,8 +84,7 @@ public void validate(Object obj, Errors errors) {
 		Boolean preferredIdentifierChosen = false;
 		//Voided patients have only voided identifiers since they were voided with the patient, 
 		//so get all otherwise get the active ones
-		Collection<PatientIdentifier> identifiers = patient.getVoided() ? patient.getIdentifiers() : patient
-		        .getActiveIdentifiers();
+		Collection<PatientIdentifier> identifiers = patient.getVoided() ? patient.getIdentifiers() : patient.getActiveIdentifiers();
```
ibacher: This looks to me like incorrect formatting that shouldn't be changed in this PR. Please undo this change and ensure your IDE is setup following our standards.

ibacher (changes requested):
@Binayak490-cyber I need you to fix your IDE formatting. That will make this easier to review, but you really need to follow the startup instructions or be very careful to only insert tabs and not autoformat in anyway.

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: Yeah @ibacher, sure and sorry for the incovenience from my side.

Binayak490-cyber: Thanks for this guide @ibacher of indentation to be fixed in my IDE formatting, I have restored the original indentation and cleaned up the formatting so the diff now reflects only the intended changes.

Binayak490-cyber:
@ibacher, any further changes to be addressed for this PR...?

Binayak490-cyber:
@ibacher, your changes has been addressed.
Any further updates for this PR...?

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -90,6 +94,38 @@ public void validate(Object obj, Errors errors) {
...
 			errors.reject("error.preferredIdentifier");
 		}
+
+		if (!patient.getVoided()) {
+			Collection<PatientIdentifierType> identifierTypes = Context.getPatientService().getAllPatientIdentifierTypes(false);
+
+			List<String> missingRequiredIdentifiers = new ArrayList<>();
+
+			for (PatientIdentifierType type : identifierTypes) {
+				if (Boolean.TRUE.equals(type.getRequired())) {
```
ibacher:
```suggestion
				if (type.getRequired()) {
```

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -90,6 +94,38 @@ public void validate(Object obj, Errors errors) {
...
+					boolean found = false;
+					for (PatientIdentifier pi : patient.getActiveIdentifiers()) {
+						if (!pi.getVoided() && pi.getIdentifierType() != null && pi.getIdentifierType().getId().equals(type.getId())) {
+							found = true;
+							break;
+						}
+					}
+
+					if (!found) {
+							missingRequiredIdentifiers.add(type.getName());
```
ibacher: This seems overindented?

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -90,6 +94,38 @@ public void validate(Object obj, Errors errors) {
...
+
+		if (!patient.getVoided()) {
+			Collection<PatientIdentifierType> identifierTypes = Context.getPatientService().getAllPatientIdentifierTypes(false);
+
+			List<String> missingRequiredIdentifiers = new ArrayList<>();
+
+			for (PatientIdentifierType type : identifierTypes) {
+				if (Boolean.TRUE.equals(type.getRequired())) {
+					boolean found = false;
+					for (PatientIdentifier pi : patient.getActiveIdentifiers()) {
```
ibacher: We already loop through the patients identifiers below. It seems like it would be cleaner to:

1. Gather all the required identifier types in one go.
2. When we look through the patient's identifiers, if the identifier is not voided, check to see if it's required. If it's a required type, pop the type out of the set of required identifiers.
3. Once we're through all the identifiers, if there are any remaining required types, then those are the ones that are missing for this patient.

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: Updated!

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: Good! Done!

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: Yeah actually @ibacher, it avoids nested loops and repeated checks — instead of comparing every identifier with every type (O(N×M)), we now do a single pass over types and identifiers (O(N+M)). It also makes the intent clearer: remaining items in the set = missing required identifiers.

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: Thanks @ibacher! I believe this suggestion was for the previous implementation. Since the logic is now refactored and `getRequired()` returns a nullable `Boolean`, I kept `Boolean.TRUE.equals(...)` to avoid a potential NPE during validation. Please let me know if you still prefer changing it.

Binayak490-cyber:
@ibacher, your changes has been addressed.
Any further updates for this PR...?

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -90,6 +96,41 @@ public void validate(Object obj, Errors errors) {
...
+			Set<PatientIdentifierType> requiredTypes = new HashSet<>();
+			for (PatientIdentifierType type : identifierTypes) {
+				if (type.getRequired()) {
+					requiredTypes.add(type);
+				}
+			}
+
+			// Step 2: remove the ones the patient already has
+			for (PatientIdentifier pi : patient.getActiveIdentifiers()) {
+				if (!pi.getVoided() && pi.getIdentifierType() != null) {
```
wikumChamith: Do we need `!pi.getVoided()` here?

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -9,11 +9,17 @@
...
+import java.util.Set;
+import java.util.HashSet;
+import org.openmrs.annotation.Handler;
 import org.openmrs.Patient;
 import org.openmrs.PatientIdentifier;
-import org.openmrs.annotation.Handler;
+import org.openmrs.PatientIdentifierType;
+import org.openmrs.api.context.Context;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
```
wikumChamith: This import order looks wrong. Imports need to be alphabetically ordered.

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -90,6 +96,41 @@ public void validate(Object obj, Errors errors) {
...
+
+			// Step 2: remove the ones the patient already has
+			for (PatientIdentifier pi : patient.getActiveIdentifiers()) {
+				if (!pi.getVoided() && pi.getIdentifierType() != null) {
+					requiredTypes.remove(pi.getIdentifierType());
+				}
+			}
+
+			// Step 3: remaining = missing
+			List<String> missingRequiredIdentifiers = new ArrayList<>();
```
wikumChamith: This should be declared inside the if block. The initial new ArrayList<>() is never used

wikumChamith (changes requested):

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: @wikumChamith, getActiveIdentifiers() should also exclude voided identifiers, but I kept the !pi.getVoided() check as a defensive safeguard in case identifiers from other contexts or future changes introduce voided entries. It ensures we only consider non-voided identifiers when removing required types.

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: Done!

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: @wikumChamith, I've moved the declaration inside the if block and removed the unused new ArrayList<>().

api/src/main/java/org/openmrs/validator/PatientValidator.java
wikumChamith: Let's remove this. We don't need it.

Binayak490-cyber: okk!

Binayak490-cyber: @wikumChamith, Removed the redundant !pi.getVoided() check as suggested.

Binayak490-cyber:
@ibacher, @wikumChamith, your changes has been addressed.
Could you go for another round of review...???

wikumChamith:
@Binayak490-cyber have you noticed the 4 new sonar issues?

Binayak490-cyber:
> @Binayak490-cyber have you noticed the 4 new sonar issues?

@wikumChamith, Addressed all the 4 new sonar issues and the last push shows no new issue.So, could you go for another round of review for this PR...?

Binayak490-cyber:
@wikumChamith, @ibacher, could you go for another round of review for this PR...?

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -9,11 +9,18 @@
  */
 package org.openmrs.validator;
 
+import java.util.ArrayList;
```
wikumChamith: Is this getting used anywhere?

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
```diff
@@ -188,11 +190,275 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+		PatientIdentifierType patientIdentifierType = Context.getPatientService().getAllPatientIdentifierTypes(false).get(0);
+		patientIdentifierType.setRequired(true);
+		patientIdentifierType.setRetired(false);
+		Context.getPatientService().savePatientIdentifierType(patientIdentifierType);
+
+		Patient patient = new Patient();
+
+		PersonName pName = new PersonName();
+   		pName.setGivenName("Tom");
```
wikumChamith: There are numerous instances like this where you are mixing tabs and spaces.

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
```diff
@@ -188,11 +190,275 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+		patient.addName(pName);
+		patient.setGender("male");
+		patient.setBirthdate(new Date());
+
+		PersonAddress pAddress = new PersonAddress();
+		pAddress.setAddress1("123 My street");
+		pAddress.setAddress2("Apt 402");
+		pAddress.setCityVillage("Anywhere city");
+		pAddress.setCountry("Some Country");
+   	 	patient.addAddress(pAddress);
```
wikumChamith: ditto

wikumChamith (commented):
Fix all the formatting changes (I only tagged a few of them). Otherwise  LGTM

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: Removed!

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
Binayak490-cyber: Done!

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
Binayak490-cyber: Done!

Binayak490-cyber:
@ibacher, could you take a look at it now as it has been approved by @wikumChamith and now needs your guidance and review over this...?

Binayak490-cyber:
@ibacher, could you take a look at it now as it has been approved by @wikumChamith and now needs your guidance and review over this...?

Binayak490-cyber:
@ibacher, could you take a look at it now as it has been approved by @wikumChamith and now needs your review over this...?

api/src/main/java/org/openmrs/validator/PatientValidator.java
wikumChamith, line 68: Could you update the Javadoc with new validation scenarios?

Binayak490-cyber: Yeah sure!

Binayak490-cyber: @wikumChamith, updated the Javadoc with the 6 new validation scenarios introduced by this PR. Can you check it out now...!

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -78,30 +90,71 @@ public void validate(Object obj, Errors errors) {
...
-			// Validate PatientIdentifers
-			for (PatientIdentifier identifier : patient.getIdentifiers()) {
-				errors.pushNestedPath("identifiers[" + index + "]");
-				patientIdentifierValidator.validate(identifier, errors);
-				errors.popNestedPath();
-				index++;
+	}
+
+	private void validateRequiredIdentifiers(Patient patient, Errors errors) {
+		Collection<PatientIdentifierType> identifierTypes = Context.getPatientService().getAllPatientIdentifierTypes(false);
```
ibacher: Since we can use autowiring in this class, we should in preference to calling `Context`. Please just ask Spring to autowire the patient service.

api/src/main/java/org/openmrs/validator/PatientValidator.java
```diff
@@ -78,30 +90,71 @@ public void validate(Object obj, Errors errors) {
 
 		ValidationUtils.rejectIfEmptyOrWhitespace(errors, "gender", "Person.gender.required");
 
-		// Make sure they chose a preferred ID
-		Boolean preferredIdentifierChosen = false;
-		//Voided patients have only voided identifiers since they were voided with the patient,
-		//so get all otherwise get the active ones
-		Collection<PatientIdentifier> identifiers = patient.getVoided() ? patient.getIdentifiers()
+		validatePreferredIdentifier(patient, errors);
```
ibacher: I'm a little worried that we've now expanded things into iterating over the patient's identifiers 3 times which seems wasteful.

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
```diff
@@ -195,4 +197,262 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+	@Test
+	public void validate_shouldNotFailForVoidedPatientWithoutRequiredIdentifier() {
+
+		PatientIdentifierType type = Context.getPatientService().getAllPatientIdentifierTypes(false).get(0);
+
+		type.setRequired(true);
+		type.setRetired(false);
+		Context.getPatientService().savePatientIdentifierType(type);
+
+		Patient patient = new Patient();
```
ibacher, line 483: Patient here doesn't seem to have any identifiers which makes this test... not terribly useful.

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
```diff
@@ -195,4 +197,262 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+
+		assertTrue(errors.hasFieldErrors("identifiers"));
+		assertEquals("Patient.missingRequiredIdentifier", errors.getFieldError("identifiers").getCode());
+	}
+
+	/**
+	 * @see PatientValidator#validate(Object,Errors)
+	 */
+	@Test
+	public void validate_shouldPassIfRequiredIdentifierIsPresent() {
```
ibacher, line 239: Seems like we should have a test with two required identifiers and only one assigned to a patient too.

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
```diff
@@ -195,4 +197,262 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
...
+		Errors errors = new BindException(patient, "patient");
+		validator.validate(patient, errors);
+
+		assertFalse(errors.hasFieldErrors("identifiers"));
+	}
+
+	/**
+	 * @see PatientValidator#validate(Object,Errors)
+	 */
+
```
ibacher: Unnecessary blank line.

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
Binayak490-cyber: okk...then I will modify the test to a more closer context which makes it valid.

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: Then I will merge validatePreferredIdentifier and validateRequiredIdentifiers into a single method validatePreferredAndRequiredIdentifiers, so both the preferred check and required type tracking now happen in a single pass over the active identifiers. This reduces the iterations from 3 to 2.

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: Done.

api/src/main/java/org/openmrs/validator/PatientValidator.java
Binayak490-cyber: Done.

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
Binayak490-cyber: Done.

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
Binayak490-cyber: Added.

api/src/test/java/org/openmrs/validator/PatientValidatorTest.java
Binayak490-cyber: Removed.

Binayak490-cyber:
@ibacher, @wikumChamith, can you check it out now as I have addressed the review comments and updated the PR accordingly...!

Binayak490-cyber:
@ibacher, @wikumChamith, can you check it out now as I have addressed the review comments and updated the PR accordingly...!

Binayak490-cyber:
@ibacher, @wikumChamith, can you check it out now as I have addressed the review comments and updated the PR accordingly...!

Binayak490-cyber:
@ibacher, @wikumChamith, can you check it out now as I have addressed the review comments and updated the PR accordingly...!

PR #5729, opened 2026-01-31
O3-4616: Add permission check for placing retrospective orders on behalf of another provider

akash8843:
## Description

While working on RDE-related order flows, I noticed that during retrospective order creation, there is currently no explicit permission check when an order is placed on behalf of another provider.

This PR adds a small but important authorization check to handle that case more safely.

### What this PR changes
- Introduces a new privilege: `PLACE_ORDERS_ON_BEHALF`
- Adds a backend check in `OrderServiceImpl` during retrospective order saving
- If the authenticated provider is different from the orderer, the user must have this privilege
- An `APIException` is thrown if the privilege is missing

The existing behavior remains unchanged for normal (non-retrospective) order flows.

### Why this is needed
In real-world RDE workflows, retrospective data entry is often done by data entry staff, not clinicians.  
Without an explicit privilege check, any authenticated user could place orders on behalf of another provider, which does not feel safe or intentional.

This change makes that behavior explicit and configurable through roles and privileges.

---

## Issue
- https://issues.openmrs.org/browse/O3-4616

---

## Notes
This PR focuses only on backend enforcement. A follow-up frontend PR can handle hiding or disabling RDE-related UI elements based on the same privilege.

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java
@@ -54,6 +54,7 @@ import org.openmrs.parameter.OrderSearchCriteria;
 import org.openmrs.util.ConfigUtil;
 import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.OpenmrsUtil;
+import org.openmrs.util.PrivilegeConstants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -71,9 +72,13 @@ import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
+import java.util.Collection;
 
 import static org.openmrs.Order.Action.DISCONTINUE;
 import static org.openmrs.Order.Action.REVISE;
+import org.openmrs.User;
+
+
 
 /**
  * Default implementation of the Order-related services class. This method should not be invoked by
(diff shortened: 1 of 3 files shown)


emphor11:
@akash8843 
I noticed the addition of isServletContextModifiable() and the conditional around WebDaemon.startOpenmrs. This appears to affect general webapp startup behavior and doesn’t seem directly related to the RDE permission changes in O3-4616. Could you clarify the motivation for this change or whether it should be handled in a separate PR?

web/src/main/java/org/openmrs/web/Listener.java
```diff
@@ -83,6 +83,17 @@ public final class Listener extends ContextLoader implements ServletContextListe
 	
 	private static final org.slf4j.Logger log = LoggerFactory.getLogger(Listener.class);
 	
+	private boolean isServletContextModifiable(ServletContext servletContext) {
+	try {
```
suubi-joshua: The changes appear unrelated to retrospective order authorization. If they are required, it would help to add a brief explanation in the PR description, or consider moving them into a separate PR to keep the scope focused.

suubi-joshua:
Could you also add test coverage for the changes you have made.

akash8843:
@emphor11 , @suubi-joshua Thanks for the review.

The backend permission enforcement for placing retrospective orders on behalf of another provider has been completed.

This PR includes:
- Privilege check implementation in OrderServiceImpl
- New PLACE_ORDERS_ON_BEHALF constant in PrivilegeConstants
- Unit test coverage in OrderServiceTest

All tests are passing locally.

Please let me know if any further refinements are needed.

PR #5859, opened 2026-02-26
TRUNK-6575: Fix Raw SQL Execution API

shivvani-r:
I have replaced the vulnerable string-based SQL validation logic in `DatabaseUtil` with a robust security framework powered by **JSqlParser 5.3** and **JDBC Read-Only enforcement**. This addresses a critical security flaw where malicious SQL could be bypassed using comments, whitespace, or stacked statements.

**Specific changes include:**
* **AST Integration:** Integrated **JSqlParser 5.3** to analyze SQL as an Abstract Syntax Tree (AST) rather than a raw string, enabling structural intent analysis.
* **Structural Validation:** Updated `checkQueryForManipulationCommands` to strictly enforce single-statement queries and block hazardous clauses like `SELECT ... INTO` and `FOR UPDATE`.
* **CTE Protection:** Added specific logic to detect and block data manipulation attempts smuggled inside PostgreSQL **Common Table Expressions (CTEs)** using the `WITH` keyword.
* **Defense in Depth:** Modified `populateResultsFromSQLQuery` to physically lock the JDBC connection to **Read-Only** mode for non-manipulation queries, providing a secondary layer of protection at the driver level.
* **Connection Safety:** Implemented state restoration in a `finally` block to ensure connection settings are returned to their original state, preserving the integrity of the **Connection Pool**.
* **Comprehensive Testing:** Created `DatabaseUtilTest.java` with 6 test cases covering:
    * Stacked query injection.
    * Data exfiltration via `INTO OUTFILE`.
    * Comment-based bypass attempts.
    * Unlisted destructive commands (e.g., `TRUNCATE`).
    * False-positive prevention for safe string literals.

## Issue I worked on
See: https://issues.openmrs.org/browse/TRUNK-6575


shivvani-r:
Hi @dkayiwa, please excuse the messy commit history I had while resolving merge conflicts. Could you please review this when you have a moment?

api/src/main/java/org/openmrs/util/DatabaseUtil.java
```diff
@@ -135,25 +142,52 @@ public static List<List<Object>> executeSQL(Connection conn, String sql, boolean
...
+			Select selectStmt = (Select) stmt;
+			if (selectStmt.getWithItemsList() != null) {
+				for (WithItem withItem : selectStmt.getWithItemsList()) {
+					if (!(withItem.getSelect() instanceof Select)) {
+						throw new IllegalArgumentException("Security violation: Data manipulation detected.");
+					}
+				}
+			}
+			if (selectStmt.getSelectBody() instanceof PlainSelect) {
+				PlainSelect plainSelect = (PlainSelect) selectStmt.getSelectBody();
```
sagarkhandagre998: I think it will throw compilation error as  Select.getSelectBody() was removed in JSQLParser 5.x. 

The correct 5.x pattern is to check and cast directly.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
```diff
@@ -135,25 +142,52 @@ public static List<List<Object>> executeSQL(Connection conn, String sql, boolean
...
-		        || sqlLower.startsWith("alter") || sqlLower.startsWith("drop") || sqlLower.startsWith("create")
-		        || sqlLower.startsWith("rename")) {
-			dataManipulation = true;
+		if (!selectOnly) {
+			return true;
 		}
-		
-		if (selectOnly && dataManipulation) {
-			throw new IllegalArgumentException("Illegal command(s) found in query string");
+		try {
```
sagarkhandagre998, line 152: JSQLParser block runs even when `selectOnly=false`, breaking legitimate DML.

This "try" block runs for all  SQL that doesn't early-return via the "if (!selectOnly) return true" path above. That early return only fires for commands already in the startsWith blocklist (INSERT, UPDATE, DELETE, ALTER, DROP, CREATE, RENAME). 

Any command not in that list — TRUNCATE, CALL, GRANT, REVOKE, LOCK, FLUSH — falls through to here and hits the !(stmt instanceof Select) check, which throws even when "selectOnly=false"  (i.e. when DML is explicitly allowed).

Fix: wrap this entire "try" block in "if (selectOnly) { ... }" so JSQLParser validation only runs when we actually want to restrict to SELECT-only queries.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
```diff
@@ -135,25 +142,52 @@ public static List<List<Object>> executeSQL(Connection conn, String sql, boolean
...
+				}
+				if (plainSelect.getForClause() != null) {
+					throw new IllegalArgumentException("Security violation: SELECT ... FOR UPDATE is not allowed.");
+				}
+			}
+			return false;
+		} catch (JSQLParserException e) {
+			throw new IllegalArgumentException("Invalid SQL syntax or illegal query structure detected.", e);
 		}
-		return dataManipulation;
```
sagarkhandagre998: return dataManipulation  is dead/unreachable/removed  code.

The method now always returns `false` for non-early-return paths, discarding the value of "dataManipulation = true" set for DML commands when "selectOnly=false". The intent of returning "true"  for DML (to route to executeUpdate vs executeQuery) is silently broken for the non-blocklisted DML commands.

sagarkhandagre998 (changes requested):

api/src/main/java/org/openmrs/util/DatabaseUtil.java
shivvani-r: @sagarkhandagre998  You were right about the logic trap. I've wrapped the strict JSqlParser validation inside if (selectOnly) so that valid DML commands (like TRUNCATE) are no longer accidentally blocked when selectOnly=false.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
shivvani-r: I've removed the deprecated getSelectBody() call and updated the code to use the modern 5.x pattern.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
shivvani-r: The method now correctly returns !isSelect at the end to ensure non-SELECT commands route properly to executeUpdate.

sagarkhandagre998:
@shivvani-r  Marked all the comment resolved for better tracking.

sagarkhandagre998:
@shivvani-r , you should not delete the previous commit, if you are fixing the reviewed changes then it should contain next commit for better tracking and understanding

shivvani-r:
> @shivvani-r , you should not delete the previous commit, if you are fixing the reviewed changes then it should contain next commit for better tracking and understanding

@sagarkhandagre998 Apologies, I'll commit normally. I marked the comments as resolved.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
```diff
@@ -135,25 +142,55 @@ public static List<List<Object>> executeSQL(Connection conn, String sql, boolean
...
+					if (plainSelect.getForClause() != null) {
+						throw new IllegalArgumentException("Security violation: SELECT ... FOR UPDATE is not allowed.");
+					}
+				}
+				return false;
+			}
+			
+			return !isSelect;
+			
+		} catch (JSQLParserException e) {
```
sagarkhandagre998, line 185: catch (JSQLParserException)` is too aggressive for selectOnly=false — blocks valid dialect-specific SQL

When `selectOnly=false`, the caller explicitly trusts the SQL (e.g. `ModuleFactory.runDiff()`, `dropNotNullConstraint()`). If JSQLParser fails to parse dialect-specific syntax it doesn't fully support (MySQL `ON DUPLICATE KEY UPDATE`, H2-specific syntax, Liquibase migration statements), this catch block throws `IllegalArgumentException` and **blocks the query entirely**, even though DML was explicitly allowed.


>  if (selectOnly) {
>         // selectOnly=true: unparseable SQL is a security risk, block it
>         throw new IllegalArgumentException(
>             "Invalid SQL syntax or illegal query structure detected.", e);
>     }
>     // selectOnly=false: caller trusts this SQL, treat as DML and let the
>     // database driver validate it
>     log.warn("Could not parse SQL for security validation, treating as DML: {}", 
>         e.getMessage());
>     return true;

sagarkhandagre998 (changes requested):

api/src/main/java/org/openmrs/util/DatabaseUtil.java
```diff
@@ -135,25 +142,55 @@ public static List<List<Object>> executeSQL(Connection conn, String sql, boolean
...
+			Statement stmt = statements.getStatements().get(0);
+			boolean isSelect = stmt instanceof Select;
+			if (selectOnly) {
+				if (!isSelect) {
+					throw new IllegalArgumentException("Security violation: Only SELECT statements are allowed.");
+				}
+				Select selectStmt = (Select) stmt;
+				if (selectStmt.getWithItemsList() != null) {
+					for (WithItem withItem : selectStmt.getWithItemsList()) {
+						if (!(withItem.getSelect() instanceof Select)) {
```
sagarkhandagre998, line 166: This `WithItem` security check is a logical no-op — it can never trigger.

In JSQLParser's grammar, `WithItem.getSelect()` always returns a `Select` by definition. A CTE body `WITH name AS (...)` is grammatically required to be a SELECT statement. So `withItem.getSelect() instanceof Select` is always `true`, meaning the `throw` inside this block **can never execute**.

The attack this is trying to prevent — a PostgreSQL writable CTE like:
> ```sql
> WITH deleted AS (DELETE FROM users RETURNING *)
> SELECT * FROM deleted
> ```
> ...is actually caught by the `JSQLParserException` in the `catch` block (because JSQLParser either fails to parse it or parses it in a way where the outer statement is not a `Select`, failing the `!isSelect` check above). So the protection exists, but **not through this check**.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
```diff
@@ -135,25 +142,55 @@ public static List<List<Object>> executeSQL(Connection conn, String sql, boolean
...
+							throw new IllegalArgumentException("Security violation: Data manipulation detected.");
+						}
+					}
+				}
+				if (selectStmt instanceof PlainSelect) {
+					PlainSelect plainSelect = (PlainSelect) selectStmt;
+					if (plainSelect.getIntoTables() != null && !plainSelect.getIntoTables().isEmpty()) {
+						throw new IllegalArgumentException("Security violation: SELECT ... INTO clauses are forbidden.");
+					}
+					if (plainSelect.getForClause() != null) {
```
sagarkhandagre998, line 176: Blocking `SELECT ... FOR UPDATE` is over-restrictive and causes false positives.

SELECT ... FOR UPDATE` is a standard SQL pessimistic locking construct, not a security threat. It does not write data, read files, or cause any harm in a `selectOnly` context. Blocking it here will cause false positives for any legitimate caller that needs row-level locking.
>
> The `selectOnly` parameter's purpose — as documented in `AdministrationService.java` — is to prevent *data manipulation* (INSERT, UPDATE, DELETE, DDL). `SELECT ... FOR UPDATE` does not manipulate data.

sagarkhandagre998 (changes requested):

api/pom.xml
```diff
@@ -298,6 +298,13 @@
 			<artifactId>js-scriptengine</artifactId>
 			<type>pom</type>
 		</dependency>
+
+		<dependency>
+			<groupId>com.github.jsqlparser</groupId>
+			<artifactId>jsqlparser</artifactId>
+			<version>5.3</version>
+			<scope>compile</scope>
```
ibacher:
```suggestion
```

api/src/main/java/org/openmrs/util/DatabaseUtil.java
ibacher: I'm not sure why you're so certain that this is a logical no-op. We can see [quite clearly](https://github.com/JSQLParser/JSqlParser/blob/a019aa01085f110d4f4c42cf49448289a077f9e5/src/main/jjtree/net/sf/jsqlparser/parser/JSqlParserCC.jjt#L979) that JSQLParser allows DML in `WITH` statements and, if we try it out:

```groovy
CCJSqlParserUtil.parse('with `delete` as (delete from my_table) select * from my_table;')
```

JSQLParser does have a tendency to support syntax no DBMS will actually execute.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
ibacher: In a general sense, this is correct. In a specific sense, this is not. This whole chain exists in pipelines that are intended to execute a single SQL statement at a time, each in it's own transaction. `SELECT ... FOR UPDATE` is thus semantically incorrect even if not per se a security issue.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
```diff
@@ -135,25 +142,55 @@ public static List<List<Object>> executeSQL(Connection conn, String sql, boolean
...
+						throw new IllegalArgumentException("Security violation: SELECT ... FOR UPDATE is not allowed.");
+					}
+				}
+				return false;
+			}
+			
+			return !isSelect;
+
```
ibacher:
```suggestion
```

api/src/main/java/org/openmrs/util/DatabaseUtil.java
sagarkhandagre998: Acknowledged! Blocking `FOR UPDATE` here is therefore the **right call** ,  not for security reasons, but for **semantic correctness**.

Should we improve the error message to explain the transactional reason?

api/pom.xml
shivvani-r: @ibacher Removed the `<scope>compile</scope>` tag from the `jsqlparser` dependency.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
shivvani-r: @ibacher @sagarkhandagre998 Updated the `FOR UPDATE` exception message to correctly explain that it is blocked because it's semantically incorrect for single-statement transactions, rather than just a security violation.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
shivvani-r: @sagarkhandagre998 Updated the catch block so it only throws an exception if selectOnly=true. If selectOnly=false, it now logs a warning and returns true (treating it as DML) to ensure valid, dialect-specific SQL isn't accidentally blocked.

sagarkhandagre998: you can also add comments to understand the reason better for new/other devs.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
sagarkhandagre998: That was my bed, missunderstood that.

api/src/main/java/org/openmrs/util/DatabaseUtil.java
shivvani-r: @sagarkhandagre998 sure, i will.

shivvani-r:
@ibacher could you please review this when you have a moment?

shivvani-r:
@ibacher could you please review this when you have a moment?

Binayak490-cyber:
Hi @shivvani-r, the build is failing...!

shivvani-r:
> Hi @shivvani-r, the build is failing...!

@Binayak490-cyber yes i know. im working on it as we speak.

PR #5867, opened 2026-02-27
TRUNK-2853: Fix cumulative progress reporting for update task

emphor11:
Improved the accuracy and perceived smoothness of the installation wizard progress bar, focusing on database updates and test installation tasks.

For the `UPDATE_TO_LATEST` task, progress reporting is now cumulative across all applicable Liquibase update files instead of resetting per file. The total number of unrun changesets is calculated once and a single callback is reused, ensuring that progress advances monotonically from 0% to 100% for the entire update operation.

Additionally, the polling interval in `progress.vm` was reduced to better reflect progress for fast-executing tasks (such as adding core data or applying a small number of update changesets during a fresh install). This addresses cases where progress previously appeared to jump directly from 0% to completion due to coarse polling, even though backend progress reporting was correct.

No changes were made to the overall wizard flow or task ordering.

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-2853

- [ ] I have **added tests** to cover my changes. (Not added — this change affects installation-time behavior and UI polling, which is not currently covered by automated tests.)

### DIFF
--- a/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
+++ b/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
@@ -1573,22 +1573,30 @@ public class InitializationFilter extends StartupFilter {
 							private int i = 1;
 							
 							private String message;
+                                                   
+                                                   private int totalToRun;
 							
 							public PrintingChangeSetExecutorCallback(String message) {
 								this.message = message;
 							}
+                                                   
+                                                   public PrintingChangeSetExecutorCallback(String message, int totalToRun) {
+                                                           this.message = message;
+                                                           this.totalToRun = totalToRun;
+                                                   }
 							
 							/**
 							 * @see ChangeSetExecutorCallback#executing(liquibase.changelog.ChangeSet, int)
 							 */
 							@Override
 							public void executing(ChangeSet changeSet, int numChangeSetsToRun) {
-								setMessage(message + " (" + i++ + "/" + numChangeSetsToRun + "): Author: "
+                                                           int effectiveTotal = totalToRun > 0 ? totalToRun : numChangeSetsToRun;
+                                                           setMessage(message + " (" + i++ + "/" + effectiveTotal + "): Author: "
 									+ changeSet.getAuthor() + " Comments: " + changeSet.getComments() + " Description: "
 									+ changeSet.getDescription());
-								float numChangeSetsToRunFloat = (float) numChangeSetsToRun;
+                                                           float effectiveTotalFloat = (float) effectiveTotal;
 								float j = (float) i;
-								setCompletedPercentage(Math.round(j * 100 / numChangeSetsToRunFloat));
+                                                           setCompletedPercentage(Math.round(j * 100 / effectiveTotalFloat));
 							}
 							
 						}
@@ -1703,11 +1711,17 @@ public class InitializationFilter extends StartupFilter {
 							List<String> changelogs = changeLogVersionFinder
 								.getUpdateFileNames(changeLogVersionFinder.getUpdateVersionsGreaterThan(version));
 							
+							int totalChangeSets = DatabaseUpdater
+							        .getUnrunDatabaseChanges(changelogs.toArray(new String[0])).size();
+							
+							PrintingChangeSetExecutorCallback updateCallback =
+							        new PrintingChangeSetExecutorCallback("Updating the database",
+							                totalChangeSets);
+							
 							for (String changelog : changelogs) {
 								log.debug("applying Liquibase changelog '{}'", changelog);
 								
-								DatabaseUpdater.executeChangelog(changelog,
-									new PrintingChangeSetExecutorCallback("executing Liquibase changelog " + changelog));
+								DatabaseUpdater.executeChangelog(changelog, updateCallback);
 							}
 							addExecutedTask(WizardTask.UPDATE_TO_LATEST);
 						}
--- a/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
+++ b/web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
@@ -1574,29 +1574,28 @@ public class InitializationFilter extends StartupFilter {
 							
 							private String message;
                                                    
-                                                   private int totalToRun;
+                            private int totalToRun;
 							
 							public PrintingChangeSetExecutorCallback(String message) {
 								this.message = message;
 							}
                                                    
-                                                   public PrintingChangeSetExecutorCallback(String message, int totalToRun) {
-                                                           this.message = message;
-                                                           this.totalToRun = totalToRun;
-                                                   }
+                            public PrintingChangeSetExecutorCallback(String message,int totalToRun) {
+                                this.message = message;
+                                this.totalToRun = totalToRun;
+                            }
 							
 							/**
 							 * @see ChangeSetExecutorCallback#executing(liquibase.changelog.ChangeSet, int)
 							 */
 							@Override
 							public void executing(ChangeSet changeSet, int numChangeSetsToRun) {
-                                                           int effectiveTotal = totalToRun > 0 ? totalToRun : numChangeSetsToRun;
-                                                           setMessage(message + " (" + i++ + "/" + effectiveTotal + "): Author: "
-									+ changeSet.getAuthor() + " Comments: " + changeSet.getComments() + " Description: "
-									+ changeSet.getDescription());
-                                                           float effectiveTotalFloat = (float) effectiveTotal;
+								int effectiveTotal = totalToRun > 0 ? totalToRun : numChangeSetsToRun;
+								setMessage(message + " (" + i++ + "/" + effectiveTotal + "): Author: " + changeSet.getAuthor() + " Comments: " + changeSet.getComments() + " Description: "
+								+ changeSet.getDescription());
+								float effectiveTotalFloat = (float) effectiveTotal;
 								float j = (float) i;
-                                                           setCompletedPercentage(Math.round(j * 100 / effectiveTotalFloat));
+								setCompletedPercentage(Math.round(j * 100 / effectiveTotalFloat));
 							}
 							
 						}
@@ -1714,9 +1713,8 @@ public class InitializationFilter extends StartupFilter {
 							int totalChangeSets = DatabaseUpdater
 							        .getUnrunDatabaseChanges(changelogs.toArray(new String[0])).size();
 							
-							PrintingChangeSetExecutorCallback updateCallback =
-							        new PrintingChangeSetExecutorCallback("Updating the database",
-							                totalChangeSets);
+							PrintingChangeSetExecutorCallback updateCallback = new PrintingChangeSetExecutorCallback("Updating the database",
+							totalChangeSets);
 							
 							for (String changelog : changelogs) {
 								log.debug("applying Liquibase changelog '{}'", changelog);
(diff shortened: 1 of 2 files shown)


emphor11:
Hi @suubi-joshua @dkayiwa 

Can u pls verify my pr and tell if anything further required.

Thanks

Binayak490-cyber:
Hi @emphor11, there is a mix of too much spaces and tabs which is leading to indentation errors and also hard to review for others. So, can you fix your IDE...before making further PRs.

emphor11:
@Binayak490-cyber 

I fixed the indentation thing. If anything else required let me know.

Thanks

web/src/main/resources/org/openmrs/web/filter/initialization/progress.vm
```diff
@@ -323,7 +323,7 @@
             
         }, "json");
         
-        setTimeout("showProgress()", 1000);
+        setTimeout("showProgress()", 200);
```
praisekavuma4: I think reducing this from 1 second to 200ms, it has been now made five times faster. Probably it can cause issues when it comes to slower connections and becoming more aggressive. i think you should consider an average middle ground of about 500ms. thank you

praisekavuma4 (commented):
Interval change risk.

Binayak490-cyber:
> @Binayak490-cyber
> 
> I fixed the indentation thing. If anything else required let me know.
> 
> Thanks

Yes @emphor11, go for the changes which has been stated by @praisekavuma4

emphor11:
@Binayak490-cyber @praisekavuma4 

I fixed them properly.

Thanks

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1573,22 +1573,29 @@ class PrintingChangeSetExecutorCallback implements ChangeSetExecutorCallback {
...
-									+ changeSet.getAuthor() + " Comments: " + changeSet.getComments() + " Description: "
-									+ changeSet.getDescription());
-								float numChangeSetsToRunFloat = (float) numChangeSetsToRun;
+								int effectiveTotal = totalToRun > 0 ? totalToRun : numChangeSetsToRun;
+								setMessage(message + " (" + i++ + "/" + effectiveTotal + "): Author: " + changeSet.getAuthor() + " Comments: " + changeSet.getComments() + " Description: "
+								+ changeSet.getDescription());
+								float effectiveTotalFloat = (float) effectiveTotal;
 								float j = (float) i;
-								setCompletedPercentage(Math.round(j * 100 / numChangeSetsToRunFloat));
+								setCompletedPercentage(Math.round(j * 100 / effectiveTotalFloat));
```
Binayak490-cyber, line 1598: @emphor11 
I think this improves the denominator by making it global, but the numerator (j) still resets for each executeChangelog invocation.
Wouldn’t the percentage drop back when the next changelog starts?
That would mean progress is still not truly cumulative — only the total is.

emphor11: Hi @Binayak490-cyber , thanks for the review!
That’s a good question to clarify.
In the updated implementation, the PrintingChangeSetExecutorCallback is instantiated once before the loop and the same instance is reused for all executeChangelog calls in the UPDATE_TO_LATEST task.
Because of that, the internal counter (i) — which is an instance field and incremented via i++ inside the executing method — does not reset between changelog files. Its state is preserved across all update files.
As a result, both the numerator and the denominator are now cumulative, and progress reflects the total number of executed changesets relative to the global totalChangeSets.

Binayak490-cyber: @emphor11, Since i++ is used inside setMessage(...), the value of i is incremented before calculating the percentage.
That means the percentage is effectively one step ahead of the displayed counter.
Should we compute the percentage before incrementing i, or use (i - 1) for the calculation to avoid an off-by-one issue?

emphor11: Hi @Binayak490-cyber , 
I’ve addressed the off-by-one issue in the PrintingChangeSetExecutorCallback by moving the i++ increment to the end of the executing() method. This ensures the displayed changeset index and the calculated percentage are always in sync and prevents any potential 101% progress scenario.

Binayak490-cyber: @emphor11, the updated logic looks good and keeps the counter and percentage properly aligned.

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1703,11 +1711,16 @@ public void executing(ChangeSet changeSet, int numChangeSetsToRun) {
 							List<String> changelogs = changeLogVersionFinder
 								.getUpdateFileNames(changeLogVersionFinder.getUpdateVersionsGreaterThan(version));
 							
+							int totalChangeSets = DatabaseUpdater
+							        .getUnrunDatabaseChanges(changelogs.toArray(new String[0])).size();
```
Binayak490-cyber, line 1724: @emphor11, One thought here: since getUnrunDatabaseChanges() precomputes the total before execution begins, is it guaranteed that Liquibase won’t skip any changesets at runtime (e.g., due to preconditions or context filtering)?
Just wondering whether there could be scenarios where the number of actual executing() calls differs slightly from this precomputed total.

emphor11: @Binayak490-cyber 
getUnrunDatabaseChanges() internally delegates to Liquibase’s listUnrunChangeSets, which already applies context filtering and evaluates preconditions against the current database state. Since we use the same context (DatabaseUpdater.CONTEXT) for both detection and execution, the precomputed total should normally match the number of executing() callbacks.
The only theoretical divergence would be if a precondition depends on database state that changes during the same update batch, which is uncommon in our update logs. In such a case, the progress percentage might slightly under-report (e.g., 98–99%) before completion, but it would not affect correctness.

Binayak490-cyber: @emphor11, thanks for clarifying how listUnrunChangeSets aligns with execution context. The explanation about the rare divergence scenario is helpful.

Binayak490-cyber:
I think this was it from my side which is all good for this PR and now it depends on @praisekavuma4 and his pov over this PR.

emphor11:
Thanks @Binayak490-cyber 
@praisekavuma4  all review comments from @Binayak490-cyber have been addressed and the discussions are resolved.
Please let me know if you’d like any further adjustments from my side

praisekavuma4:
There appear to be some inconsistent indentation in the added constructor (mixing tabs and spaces). This could be cleaned up for consistency. But otherwise, good work 👍👍

emphor11:
@praisekavuma4  Thanks for pointing that out! I've fixed the indentation in the constructor to align with the project's formatting style.

praisekavuma4:
Thank you for the fixes @emphor11. l.g.t.m

praisekavuma4:
@dkayiwa , @ibacher

suubi-joshua (commented):
Thanks for working on this @emphor11! 
The ticket mentions both the import test data and updates progress bars as broken. Looking at the `IMPORT_TEST_DATA` block, the progress still jumps between hardcoded values (0 → 40 → 0 → 90) rather than reporting incrementally. Could you also address this as well?

emphor11:
@suubi-joshua 
I’ve updated the progress reporting for the IMPORT_TEST_DATA and ADD_MODULES tasks by introducing intermediate progress updates and clearer status messages. This avoids the abrupt jumps previously observed and provides a smoother progression during these initialization steps. Please let me know if any further adjustments are needed.

emphor11:
@suubi-joshua 
I had updated the progress reporting for the IMPORT_TEST_DATA and ADD_MODULES tasks.
If there's anything else you want my attention ple let me know.
Thanks

emphor11:
@suubi-joshua @dkayiwa 
All things were fixed and waiting for your review.
Thanks

emphor11:
@suubi-joshua @dkayiwa @ibacher

emphor11:
@suubi-joshua @dkayiwa @ibacher

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1573,22 +1573,30 @@ class PrintingChangeSetExecutorCallback implements ChangeSetExecutorCallback {
...
 							/**
 							 * @see ChangeSetExecutorCallback#executing(liquibase.changelog.ChangeSet, int)
 							 */
 							@Override
 							public void executing(ChangeSet changeSet, int numChangeSetsToRun) {
-								setMessage(message + " (" + i++ + "/" + numChangeSetsToRun + "): Author: "
-									+ changeSet.getAuthor() + " Comments: " + changeSet.getComments() + " Description: "
-									+ changeSet.getDescription());
-								float numChangeSetsToRunFloat = (float) numChangeSetsToRun;
+								int effectiveTotal = totalToRun > 0 ? totalToRun : numChangeSetsToRun;
```
ibacher, line 1593: `totalToRun` is not initialized so we're sort of relying on Java initializing this to 0. All that said, I don't really understand why we need this construction or how that improves things. Is `numChangeSetsToRun` inaccurate or something? The reason it's not provided up-front here is that that way we can avoid the (very expensive and unnecessary) call below to `getUnrunDatabaseChanges()`

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1573,22 +1573,30 @@ class PrintingChangeSetExecutorCallback implements ChangeSetExecutorCallback {
 							private int i = 1;
 							
 							private String message;
+                                                   
+                        	private int totalToRun;
 							
 							public PrintingChangeSetExecutorCallback(String message) {
 								this.message = message;
 							}
+
```
ibacher, line 1582: Your IDE is not setup with our coding standards. Please fix that.

ibacher: Also, run spotless before committing. This is why the checkboxes on the PR exist.

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1657,8 +1671,11 @@ public void executing(ChangeSet changeSet, int numChangeSetsToRun) {
 										wizardModel.remoteUrl + RELEASE_TESTING_MODULE_PATH + "getModules.htm",
 										wizardModel.remoteUsername, wizardModel.remotePassword);
 									
-									setCompletedPercentage(90);
-									setMessage("Adding imported modules...");
+									setCompletedPercentage(50);
```
ibacher, line 1674: So our made-up sequence of percentage complete here instead of just jumping to 90 now goes 20 -> 40 -> 70 -> 50 -> 80? What are these numbers based on? How is this improving the UX?

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1703,11 +1720,16 @@ public void executing(ChangeSet changeSet, int numChangeSetsToRun) {
 							List<String> changelogs = changeLogVersionFinder
 								.getUpdateFileNames(changeLogVersionFinder.getUpdateVersionsGreaterThan(version));
 							
+							int totalChangeSets = DatabaseUpdater
```
ibacher, line 1723: See above. This is a call we want to avoid if we can (and the previous version did). It's unclear that this is a net improvement, but it definitely makes the process slower.

web/src/main/resources/org/openmrs/web/filter/initialization/progress.vm
```diff
@@ -323,7 +323,7 @@
             
         }, "json");
         
-        setTimeout("showProgress()", 1000);
+        setTimeout("showProgress()", 500);
```
ibacher, line 326: Why would doubling the number of requests we're making be an improvement?

ibacher (changes requested):

emphor11 (commented):
.

PR #5881, opened 2026-03-03
O3-5427: Validate condition onset date against patient DOB

BlessedAmrita:

Added DOB lower-bound validation for the Condition onset date in `ConditionValidator`.

When a `Condition` has an `onsetDate` that falls strictly before the associated patient's date of birth, the validator now rejects the value with a field error (`Condition.error.onsetDateBeforePatientBirthdate`). This prevents clinically impossible data entry; a patient cannot have a condition onset before they were born. Onset dates equal to the birthdate are allowed (a condition can be diagnosed at birth).

**Changes:**
- `ConditionValidator.java`: Added birthdate lower-bound check for `onsetDate` after existing required-field validations. Uses triple null guard (`onsetDate`, `patient`, `patient.getBirthdate()`) before comparing dates via `Date.before()`.
- `ConditionValidatorTest.java`: Added 5 JUnit 5 test cases covering all edge cases and a `date()` helper method for concise date construction.

**Test scenarios:**
- `onsetDate` before patient birthdate → validation error
- `onsetDate` equal to patient birthdate → no error (valid: condition at birth)
- `onsetDate` after patient birthdate → no error
- `onsetDate` is `null` → no error
- Patient birthdate is `null` → no error

Part of the Date Picker Improvements epic (O3-5424).

## Issue I worked on

see https://issues.openmrs.org/browse/O3-5427

### DIFF
--- a/api/src/main/java/org/openmrs/validator/ConditionValidator.java
+++ b/api/src/main/java/org/openmrs/validator/ConditionValidator.java
@@ -10,6 +10,7 @@
 package org.openmrs.validator;
 
 import org.openmrs.Condition;
+import org.openmrs.Patient;
 import org.openmrs.annotation.Handler;
 import org.springframework.validation.Errors;
 import org.springframework.validation.Validator;
@@ -52,6 +53,12 @@ public class ConditionValidator implements Validator {
 		if (condition.getClinicalStatus() == null) {
 			errors.rejectValue("clinicalStatus", "Condition.clinicalStatusShouldNotBeNull",
 			    "The clinical status is required");
+		}
+		Patient patient = condition.getPatient();
+		if (condition.getOnsetDate() != null && patient != null && patient.getBirthdate() != null
+		        && condition.getOnsetDate().before(patient.getBirthdate())) {
+			errors.rejectValue("onsetDate", "Condition.error.onsetDateBeforePatientBirthdate",
+			    "Onset date cannot be before patient's date of birth");
 		}
 	}
 }
--- a/api/src/main/resources/messages.properties
+++ b/api/src/main/resources/messages.properties
@@ -3335,6 +3335,7 @@ Module.error.cyclicDependencies=Could not start some modules because they depend
 
 Condition.conditionShouldNotBeNull=Condition is required
 Condition.clinicalStatusShouldNotBeNull=Clinical status is required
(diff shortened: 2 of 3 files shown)


api/src/main/java/org/openmrs/validator/ConditionValidator.java
```diff
@@ -50,5 +51,10 @@ public void validate(Object object, Errors errors) {
 		if(condition.getClinicalStatus() == null){
 			errors.rejectValue("clinicalStatus", "Condition.clinicalStatusShouldNotBeNull", "The clinical status is required");
 		}
+		Patient patient = condition.getPatient();
+		if (condition.getOnsetDate() != null && patient != null && patient.getBirthdate() != null
+		        && condition.getOnsetDate().before(patient.getBirthdate())) {
+			errors.rejectValue("onsetDate", "Condition.error.onsetDateBeforePatientBirthdate");
```
jwnasambu:
```suggestion
			errors.rejectValue("onsetDate", "Condition.error.onsetDateBeforePatientBirthdate", "Onset date cannot be before patient's date of birth");
```

jwnasambu:
Kindly add the message key `Condition.error.onsetDateBeforePatientBirthdate` on this file `api/src/main/resources/messages.properties`.

BlessedAmrita:
@denniskigen @dkayiwa please take a look at this

omeriinnocent (commented):
welldone @BlessedAmrita ,considering line 54-59,Should the onset date on the same calendar day (but a few milliseconds/seconds before the birthdate time) be considered "before" the birthdate for clinical purposes? Or should we only care about the calendar day (ignoring time of day)?

BlessedAmrita:
> welldone @BlessedAmrita ,considering line 54-59,Should the onset date on the same calendar day (but a few milliseconds/seconds before the birthdate time) be considered "before" the birthdate for clinical purposes? Or should we only care about the calendar day (ignoring time of day)?

Great question! The birthdate column in the DB is mapped as DATE (no time component), so when read from the database it comes back as midnight (00:00:00). This means any date on the same calendar day will be equal to or after the birthdate, so before() returns false and it passes correctly.

This is also consistent with how the existing birthdate check in VisitValidator works (it uses the same Date.before() comparison without time normalization).

That said, if maintainers prefer comparing dates without the time component for extra safety, I'm happy to adjust.

PR #5887, opened 2026-03-04
O3-5425: Validate order dates against patient DOB

BlessedAmrita:

Added DOB lower-bound validation for `dateActivated` and `scheduledDate` in `OrderValidator`.

Since these date fields live on the base `Order` class, placing the validation in `OrderValidator` means it automatically applies to all order types: `TestOrder`, `DrugOrder`, `ReferralOrder`, etc. When either date falls before the patient's date of birth, the validator rejects it with a field error. Dates equal to the birthdate are allowed.

**Changes:**
- `OrderValidator.java`: Added birthdate lower-bound checks in `validateDateActivated()` and `validateScheduledDate()` methods, with null guards for patient and birthdate
- `OrderValidatorTest.java`: Added 10 test cases (5 per date field) covering before/equal/after birthdate, null date, and null birthdate scenarios
- `messages.properties`: Added `Order.error.dateActivatedBeforePatientBirthdate` and `Order.error.scheduledDateBeforePatientBirthdate` keys

**Test scenarios (for each date field):**
- Date before patient birthdate → validation error
- Date equal to patient birthdate → no error
- Date after patient birthdate → no error
- Date is null → no error
- Patient birthdate is null → no error

Part of the Date Picker Improvements epic (O3-5424).

## Issue I worked on

see https://issues.openmrs.org/browse/O3-5425


api/src/main/java/org/openmrs/validator/OrderValidator.java
```diff
@@ -123,6 +123,11 @@ private void validateDateActivated(Order order, Errors errors) {
 			        && encounter.getEncounterDatetime().after(dateActivated)) {
 				errors.rejectValue("dateActivated", "Order.error.encounterDatetimeAfterDateActivated");
 			}
+			if (order.getPatient() != null && order.getPatient().getBirthdate() != null
```
sagarkhandagre998, line 126: Hi @BlessedAmrita, same issue as the previous PR — this is solving the wrong problem again.

The issue is about  rejecting dates with incomplete/single-digit years  (e.g., `"02/01/1"`), not about comparing order dates against the patient's date of birth.

The fix should validate that the year component of `scheduledDate` has **4 digits** (i.e., `Calendar.get(Calendar.YEAR) >= 1000`) and reject it if not — using the error key `Order.error.scheduledDateInvalidYear` with the message `"Invalid value - scheduled date must use a 4-digit year"`.

api/src/main/java/org/openmrs/validator/OrderValidator.java
```diff
@@ -142,6 +147,12 @@ private void validateScheduledDate(Order order, Errors errors) {
 		if (isUrgencyOnScheduledDate && order.getScheduledDate() == null) {
 			errors.rejectValue("scheduledDate", "Order.error.scheduledDateNullForOnScheduledDateUrgency");
 		}
+		Date scheduledDate = order.getScheduledDate();
```
sagarkhandagre998, line 150: Please update the PR accordingly. Thanks!

sagarkhandagre998 (changes requested):

api/src/main/java/org/openmrs/validator/OrderValidator.java
BlessedAmrita: Thanks for trying to review this sagar but please do check jira discussions before commenting.
https://openmrs.atlassian.net/browse/O3-5430?focusedCommentId=188818

The approach has been already discussed with the maintainers.

sagarkhandagre998: Sure I'll check.

sagarkhandagre998:
@BlessedAmrita, apologies for the earlier incorrect feedback. After reviewing the maintainer's guidance more carefully, your approach of **validating dates against the patient's date of birth** is actually the **right direction**. The maintainer confirmed that the patient's DOB is the natural lower bound for these dates, which inherently catches single-digit year entries like `"02/01/1"` since year 1 will always be before any real patient's birthdate.

api/src/main/java/org/openmrs/validator/OrderValidator.java
```diff
@@ -123,6 +123,11 @@ private void validateDateActivated(Order order, Errors errors) {
 			        && encounter.getEncounterDatetime().after(dateActivated)) {
 				errors.rejectValue("dateActivated", "Order.error.encounterDatetimeAfterDateActivated");
 			}
+			if (order.getPatient() != null && order.getPatient().getBirthdate() != null
+			        && dateActivated.before(order.getPatient().getBirthdate())) {
```
sagarkhandagre998, line 127: One edge case to consider —> in OpenMRS, a patient's birthdate can be **estimated** (`birthdateEstimated = true`), When this is the case, the stored `birthdate` is not an exact date —> it is a **synthetic value calculated from the patient's approximate age**.

 Would it could cause **false rejections**? 

Reference - `visitValidator`

api/src/main/java/org/openmrs/validator/OrderValidator.java
```diff
@@ -142,6 +147,12 @@ private void validateScheduledDate(Order order, Errors errors) {
 		if (isUrgencyOnScheduledDate && order.getScheduledDate() == null) {
 			errors.rejectValue("scheduledDate", "Order.error.scheduledDateNullForOnScheduledDateUrgency");
 		}
+		Date scheduledDate = order.getScheduledDate();
+		if (scheduledDate != null && order.getPatient() != null && order.getPatient().getBirthdate() != null
+		        && scheduledDate.before(order.getPatient().getBirthdate())) {
```
sagarkhandagre998, line 152: Same here.

api/src/main/java/org/openmrs/validator/OrderValidator.java
```diff
@@ -142,6 +147,12 @@ private void validateScheduledDate(Order order, Errors errors) {
...
+			errors.rejectValue("scheduledDate", "Order.error.scheduledDateBeforePatientBirthdate",
+			    "Scheduled date cannot be before patient's birthdate");
```
sagarkhandagre998, line 154: You can remove the `hard coded` message.

api/src/test/java/org/openmrs/validator/OrderValidatorTest.java
```diff
@@ -486,6 +486,212 @@ public void validate_shouldFailValidationIfFieldLengthsAreNotCorrect() {
 		assertTrue(errors.hasFieldErrors("voidReason"));
 	}
 	
+	/**
+	 * @see OrderValidator#validate(Object, org.springframework.validation.Errors)
+	 */
+	@Test
+	void validate_shouldFailIfDateActivatedIsBeforePatientBirthdate() {
```
sagarkhandagre998, line 493: Should be `public`.

sagarkhandagre998 (changes requested):

api/src/test/java/org/openmrs/validator/OrderValidatorTest.java
BlessedAmrita: JUnit 5 doesn't require public on test methods package-private is the recommended convention. SonarQube flags unnecessary public modifiers on @Test methods. Keeping them package-private avoids those warnings.

sagarkhandagre998, line 493: You are right that JUnit 5 doesn't technically require `public` on test methods. However, if you look at **every existing test in this file and across the entire `openmrs-core` test suite**, they all consistently use `public`.

When contributing to an existing project, **consistency with the project's own conventions takes priority over general JUnit 5 best practices**.

Thanks!

sagarkhandagre998 (changes requested):

PR #5899, opened 2026-03-06
RESTWS-1005: Filter Null Patients from Search Endpoints

youssefgamal123:
Filtered null patients returned from DAO queries in `PatientServiceImpl`.

DAO queries may occasionally return null elements in the resulting list. 
This change filters null patients objects before returning the list in the service layer.

The fix was applied to the relevant `getPatients` methods in `PatientServiceImpl`. 
If there is a preferred approach for handling this in a different layer or another recommended solution, I would appreciate feedback from the maintainers.

## Issue I worked on
see https://openmrs.atlassian.net/browse/RESTWS-1005

- [ ] I have **added tests** to cover my changes.  
  (Not required since this is a filtering fix and existing behavior is preserved.)

- [x] I ran `mvn clean package` right before creating this pull request.

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -10,15 +10,17 @@
 package org.openmrs.api.impl;
 
 
-import java.util.Arrays;
 import java.util.ArrayList;
-import java.util.Collection;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.Collection;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
@@ -1580,12 +1582,15 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 	@Override
 	@Transactional(readOnly = true)
 	public List<Patient> getPatients(String query, Integer start, Integer length) throws APIException {
-		List<Patient> patients = new ArrayList<>();
 		if (StringUtils.isBlank(query)) {
-			return patients;
+			return Collections.emptyList();
 		}
-		
-		return dao.getPatients(query, start, length);
+
+		return Optional.ofNullable(dao.getPatients(query, start, length))
+			.orElse(Collections.emptyList())
+			.stream()
+			.filter(Objects::nonNull)
+			.toList();
 	}
 	
 	/**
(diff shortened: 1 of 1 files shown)


sagarkhandagre998:
@youssefgamal123 Thanks for the contribution! 

The direction is right, but the fix is applied at the wrong layer.

 Filtering nulls in `PatientServiceImpl` is the same kind of partial fix as doing it in the REST module — it only protects callers going through the service layer. Internal paths like `getCountOfPatients()` still call `SearchQueryUnique` directly and will still produce nulls.

The root cause is in `SearchQueryUnique.search()`, where the mapper (e.g. `pN -> getPatient(pN.getPerson().getId())`) returns `null` when the Lucene index is stale and the patient no longer exists in the DB. 

Please check the flow there.

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -1580,12 +1582,15 @@ public Integer getCountOfPatients(String query, boolean includeVoided) {
...
 	public List<Patient> getPatients(String query, Integer start, Integer length) throws APIException {
-		List<Patient> patients = new ArrayList<>();
 		if (StringUtils.isBlank(query)) {
-			return patients;
+			return Collections.emptyList();
 		}
-		
-		return dao.getPatients(query, start, length);
+
+		return Optional.ofNullable(dao.getPatients(query, start, length))
```
sagarkhandagre998, line 1589: `dao.getPatients()`  never returns `null`  —> it always returns a `List` (either populated or empty). Wrapping it in `Optional.ofNullable()` adds noise with no benefit.

Binayak490-cyber:
Hi @youssefgamal123, I think the title of the PR should also be changed.

suubi-joshua (commented):
I feel like your changes filter out nulls after the DAO returns them, but doesn't address why the DAO returns nulls in the first place.

youssefgamal123:
> Hi @youssefgamal123, I think the title of the PR should also be changed.

done , Thank you

youssefgamal123:
> @youssefgamal123 Thanks for the contribution!
> 
> The direction is right, but the fix is applied at the wrong layer.
> 
> Filtering nulls in `PatientServiceImpl` is the same kind of partial fix as doing it in the REST module — it only protects callers going through the service layer. Internal paths like `getCountOfPatients()` still call `SearchQueryUnique` directly and will still produce nulls.
> 
> The root cause is in `SearchQueryUnique.search()`, where the mapper (e.g. `pN -> getPatient(pN.getPerson().getId())`) returns `null` when the Lucene index is stale and the patient no longer exists in the DB.
> 
> Please check the flow there.
@sagarkhandagre998 
Thank you for the constructive feedback and clarification, I agree with you the fix is in the wrong layer, I have noticed that `SearchQueryUnique` is used across the DAO layer in all the searching methods, So fixing it will resolve everything and we will not even need to filter nulls from the service layer,I will check flows from there.

PR #5910, opened 2026-03-10
TRUNK-4988: Use Spring bean lookup for MessageSender and MessagePreparator

Chinmay7070:

- Replaced direct instantiation of MessageSender and MessagePreparator in Context with Spring bean lookup using ApplicationContext.

- Added @Component annotations to MailMessageSender and VelocityMessagePreparator so they can be managed by Spring.

- Implemented a fallback mechanism to use the default implementations when no Spring bean is available, preserving backward compatibility.

## Issue I worked on

see https://openmrs.atlassian.net/browse/TRUNK-4988

  No? -> execute above command


omeriinnocent (commented):
In Lines 658-666 & 677-685, If Context.getServiceContext() returns null, will catch(Exception e) actually catch the NullPointerException? How would you distinguish 
between "no bean configured" vs "context not available" from the 
debug-level logs?

omeriinnocent (commented):
Why don,t you use @PostConstruct for Proper Bean Initialization?

Chinmay7070:
> In Lines 658-666 & 677-685, If Context.getServiceContext() returns null, will catch(Exception e) actually catch the NullPointerException? How would you distinguish between "no bean configured" vs "context not available" from the debug-level logs?

Thanks for pointing this out. I’ll update the implementation to check for the `ServiceContext` before performing the bean lookup and improve the debug logging to clearly distinguish between a missing context and a missing bean.

Chinmay7070:
> Why don,t you use @PostConstruct for Proper Bean Initialization?

Thanks for the review. The `VelocityEngine` is currently initialized in the constructor to keep the behavior consistent with the previous implementation. Since the bean does not depend on injected fields, constructor initialization works correctly here. However, I’m happy to refactor it to use `@PostConstruct` if that would better align with the preferred pattern in this module.

Chinmay7070:
hi @dkayiwa ,
please review when you get chance.

Chinmay7070:
hi @dkayiwa , @ibacher 
please review when you get chance.

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -649,7 +649,15 @@ public PasswordAuthentication getPasswordAuthentication() {
...
-		return new MailMessageSender(getMailSession());
+		MessageSender sender = new MailMessageSender(getMailSession());
+		try {
+			sender = Context.getServiceContext().getApplicationContext().getBean(MessageSender.class);
+		} catch (NoSuchBeanDefinitionException e) {
+			log.debug("No MessageSender bean configured, using default MailMessageSender");
+		} catch (Exception e) {
+			log.debug("Service context not available while retrieving MessageSender, using default MailMessageSender");
+		}
+		return sender;
```
ibacher: Isn't this exactly backwards? I.e., basically we should be loading the bean first and then falling back. But, in any case, it's never a great idea to imperatively fetch the bean from the application context if we can avoid it. So, how could we refactor this to work that way?

Chinmay7070:
Thanks for the feedback!
Updated to use Context.getRegisteredComponents() to resolve beans first and fallback to defaults, avoiding direct getBean() usage.

ibacher:
So I think this approach is better, but it's still missing the fundamental point. We want the `MessageSender` and `MailSession` and all the rest of it done through Spring. Just getting a component like this does is making OpenMRS do extra work for no real benefit. The benefit lies in defining all the things that make up the `MessageSender` and `MessagePreparator` in the Spring context rather than via these static methods.

ibacher:
Note, that this should be a medium-sized refactor, that is, I'd expect it to be more than 200 lines of code changes.

Chinmay7070:
I realize my implementation is still missing the key point — `MailSession` should be managed by Spring.

I’m not sure of the best approach here, especially since `Context.getMailSession()` is private. Would you recommend using a `FactoryBean` or another pattern?

I’d appreciate your guidance.

Thank you!

Chinmay7070:
Hi @ibacher, @dkayiwa 

I tried using MailSessionFactoryBean to make Session a Spring bean, but it fails because Context.getMailProperties() 
requires AdministrationService which is not available during Spring context initialization.

PR #5924, opened 2026-03-12
TRUNK-0: Fix test setup bug and improve assertions in PersonAttributeTest

siddhi2005-beep:

Fixed a test setup issue in PersonAttributeTest where values were
accidentally set on the wrong object during equalsContent testing.

The test previously updated the same PersonAttribute instance instead
of the comparison object, which could lead to misleading results.

Additionally improved assertion messages and readability in several tests.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-0

### DIFF
--- a/api/src/test/java/org/openmrs/PersonAttributeTest.java
+++ b/api/src/test/java/org/openmrs/PersonAttributeTest.java
@@ -41,10 +41,10 @@ public class PersonAttributeTest extends BaseContextSensitiveTest {
 		pa.setAttributeType(new PersonAttributeType(1));
 		pa.setValue("1");
 		pa.setVoided(false);
-		PersonAttribute other = new PersonAttribute(1); // a different personAttributeid than above
-		pa.setAttributeType(new PersonAttributeType(1));
-		pa.setValue("1");
-		pa.setVoided(false);
+		PersonAttribute other = new PersonAttribute(1);
+        other.setAttributeType(new PersonAttributeType(1));
+        other.setValue("1");
+        other.setVoided(false);
 
 		assertTrue(pa.equalsContent(other));
 	}


adarshiiit0117 (commented):
Thanks for fixing the test setup.

The change correctly configures the `other` PersonAttribute instead of
modifying `pa` again, which makes the equality test more accurate.

One small suggestion: the previous comment mentioning that this
represents a different `personAttributeId` was useful for understanding
the intent of the test. It might be worth keeping a brief comment
there for clarity.

Overall the fix looks good and improves the correctness of the test.

api/src/test/java/org/openmrs/PersonAttributeTest.java
```diff
@@ -41,10 +41,10 @@ public void equalsContent_shouldReturnTrueIfAttributeTypeValueAndVoidStatusAreTh
...
-		PersonAttribute other = new PersonAttribute(1); // a different personAttributeid than above
-		pa.setAttributeType(new PersonAttributeType(1));
-		pa.setValue("1");
-		pa.setVoided(false);
+		PersonAttribute other = new PersonAttribute(1);
+        other.setAttributeType(new PersonAttributeType(1));
+        other.setValue("1");
+        other.setVoided(false);
 
 		assertTrue(pa.equalsContent(other));
```
0-YuvrajSingh, line 49: Nice fix, this makes the equality check more accurate.

Since equalsContent() compares multiple fields, it might be useful to add
a negative test case where attributeType or value is different,
to ensure the comparison fails as expected.

tendomart:
@siddhi2005-beep  please learn hot to give the right link to the issue you're working on, this applies to the Title as well

tendomart (commented):
Remove this extra space

PR #5933, opened 2026-03-14
AUDIT-28: Backfill pre-existing rows into Envers audit tables to fix "Unable to read" in audit UI

Binayak490-cyber:
### Summary
Fixes AUDIT-28 where important fields appear as "Unable to read" in the audit UI.

### Root Cause
When Envers auditing is enabled after data already exists, the *_audit tables are created empty. New audit records referencing those pre-existing entities cannot be resolved by Envers.

### Fix
Added backfillAuditTables() in EnversAuditTableInitializer to populate audit tables with existing rows during initialization. Missing rows are inserted with REVTYPE=0 (ADD) under a synthetic revision.

### Changes
- Updated EnversAuditTableInitializer.java to perform audit table backfill
- Added EnversAuditTableInitializerTest.java with unit tests for the backfill logic

### Testing
Verified using H2 and unit tests that:
- Existing rows are copied to *_audit tables
- A single revision is created
- Duplicate rows are not inserted on subsequent runs

## Release Notes

* **New Features**
  * Envers audit tables now automatically backfill with existing database records when auditing is enabled, ensuring complete audit histories for all entities.

* **Documentation**
  * Updated documentation to clarify audit table initialization behavior and the automatic data backfill process.

* **Tests**
  * Added test coverage for audit table initialization and backfill scenarios.


Binayak490-cyber:
Hi @wikumChamith, @druchniewicz, @nsalifu, SonarCloud flagged a few SQL Injection security hotspots due to dynamically constructed queries in `EnversAuditTableInitializer`.

These queries only use table and column names sourced from Hibernate/JDBC metadata (not user input). To make this explicit, I added a `requireSafeIdentifier(String)` check that validates identifiers against `[a-zA-Z_][a-zA-Z0-9_]*` before they are used in SQL.

If an identifier contains any unsafe characters, an `IllegalArgumentException` is thrown and the query is not executed.

Since SonarCloud hotspots require manual review, could a maintainer please review and mark them as **Safe** if this approach looks correct?

Binayak490-cyber:
Hi @wikumChamith, @druchniewicz, @nsalifu, SonarCloud flagged a few SQL Injection security hotspots due to dynamically constructed queries in EnversAuditTableInitializer.

These queries only use table and column names sourced from Hibernate/JDBC metadata (not user input). To make this explicit, I added a requireSafeIdentifier(String) check that validates identifiers against [a-zA-Z_][a-zA-Z0-9_]* before they are used in SQL.

If an identifier contains any unsafe characters, an IllegalArgumentException is thrown and the query is not executed.

Since SonarCloud hotspots require manual review, could a maintainer please review and mark them as Safe if this approach looks correct?

druchniewicz:
@Binayak490-cyber did you run your code on real instance of OpenMRS? (real I mean e.g. on your localhost), because when I ran it on my local machine I got multiple errors like
```
 backend-1   
| WARN - EnversAuditTableInitializer.tryBackfillEntity(210) |2026-03-17T11:27:02,208| Failed to backfill audit relationship_type_audit: Unknown column 'REVTSTMP' in 'field list'  and 
```
It is just an example, it occurs for all audit tables.

Binayak490-cyber:
> @Binayak490-cyber did you run your code on real instance of OpenMRS? (real I mean e.g. on your localhost), because when I ran it on my local machine I got multiple errors like
> 
> ```
>  backend-1   
> | WARN - EnversAuditTableInitializer.tryBackfillEntity(210) |2026-03-17T11:27:02,208| Failed to backfill audit relationship_type_audit: Unknown column 'REVTSTMP' in 'field list'  and 
> ```
> 
> It is just an example, it occurs for all audit tables.

Thanks for testing @druchniewicz ! The fix dynamically discovers the timestamp column from JDBC metadata instead of hardcoding REVTSTMP. Could you test the updated code on your local instance to confirm it resolves the errors?

druchniewicz:
> > @Binayak490-cyber did you run your code on real instance of OpenMRS? (real I mean e.g. on your localhost), because when I ran it on my local machine I got multiple errors like
> > ```
> >  backend-1   
> > | WARN - EnversAuditTableInitializer.tryBackfillEntity(210) |2026-03-17T11:27:02,208| Failed to backfill audit relationship_type_audit: Unknown column 'REVTSTMP' in 'field list'  and 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It is just an example, it occurs for all audit tables.
> 
> Thanks for testing @druchniewicz ! The fix dynamically discovers the timestamp column from JDBC metadata instead of hardcoding REVTSTMP. Could you test the updated code on your local instance to confirm it resolves the errors?

Did you test your changes? I think the best will be running clean [distro](https://github.com/openmrs/openmrs-distro-referenceapplication) on your local machine and check if your changes work.

wikumChamith:
@Binayak490-cyber, as @druchniewicz pointed out earlier, your changes don't seem to be working. I can confirm the same. I'm still seeing empty audit tables. Make sure you actually run and manually verify your changes before making a PR.

Binayak490-cyber:
Thanks @druchniewicz and @wikumChamith for testing. In the last push, I’ve verified the fixes locally on a fresh OpenMRS distro using Docker + MariaDB.

Fixed the `revision_entity.id` issue by computing `MAX(id)+1` and inserting id + timestamp explicitly. Also resolved the column mismatch by reading columns directly from the audit table (excluding `REV` and `REVTYPE`).

On a clean DB, multiple audit tables were successfully backfilled (20+ tables), and no errors were observed in logs.

Could you please re-test on your end? 
Also, SonarCloud flagged a few SQL hotspots — identifiers are validated via `requireSafeIdentifier()`, but they require manual marking as **Safe** in SonarCloud.

wikumChamith:
What do you think about using Liquibase [CustomTaskChange](https://docs.liquibase.com/secure/reference-guide-5-1/change-types/customchange) rather than doing a JDBC backfill?

cc: @dkayiwa, @ibacher

ibacher:
> What do you think about using Liquibase [CustomTaskChange](https://docs.liquibase.com/secure/reference-guide-5-1/change-types/customchange) rather than doing a JDBC backfill?

Yeah, this.

Binayak490-cyber:
So, @ibacher, @wikumChamith, @dkayiwa, does it means that I will go for a rework, but before that want to clarify that:

1. Should only the backfill logic move into a CustomTaskChange (wired into a Liquibase changelog entry), while the audit table creation part of EnversAuditTableInitializer stays as the Hibernate Integrator?

2. Or should the entire EnversAuditTableInitializer (both table creation and backfill) be replaced by Liquibase?

3. Are there any existing CustomTaskChange implementations in the codebase I can reference for the pattern?

What do you want...?

wikumChamith:
> So, @ibacher, @wikumChamith, @dkayiwa, does it means that I will go for a rework, but before that want to clarify that:
> 
>     1. Should only the backfill logic move into a CustomTaskChange (wired into a Liquibase changelog entry), while the audit table creation part of EnversAuditTableInitializer stays as the Hibernate Integrator?
> 
>     2. Or should the entire EnversAuditTableInitializer (both table creation and backfill) be replaced by Liquibase?
> 
>     3. Are there any existing CustomTaskChange implementations in the codebase I can reference for the pattern?
> 
> 
> What do you want...?

We should keep EnversAuditTableInitializer as it is because it needs to run with any schema changes. However, the data backfill is only going to happen once.

wikumChamith:
> So, @ibacher, @wikumChamith, @dkayiwa, does it means that I will go for a rework, but before that want to clarify that:
> 
>     1. Should only the backfill logic move into a CustomTaskChange (wired into a Liquibase changelog entry), while the audit table creation part of EnversAuditTableInitializer stays as the Hibernate Integrator?
> 
>     2. Or should the entire EnversAuditTableInitializer (both table creation and backfill) be replaced by Liquibase?
> 
>     3. Are there any existing CustomTaskChange implementations in the codebase I can reference for the pattern?
> 
> 
> What do you want...?

https://github.com/search?q=org%3Aopenmrs%20CustomTaskChange&type=code

Binayak490-cyber:
> > So, @ibacher, @wikumChamith, @dkayiwa, does it means that I will go for a rework, but before that want to clarify that:
> > ```
> > 1. Should only the backfill logic move into a CustomTaskChange (wired into a Liquibase changelog entry), while the audit table creation part of EnversAuditTableInitializer stays as the Hibernate Integrator?
> > 
> > 2. Or should the entire EnversAuditTableInitializer (both table creation and backfill) be replaced by Liquibase?
> > 
> > 3. Are there any existing CustomTaskChange implementations in the codebase I can reference for the pattern?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What do you want...?
> 
> We should keep EnversAuditTableInitializer as it is because it needs to run with any schema changes. However, the data backfill is only going to happen once.

Okk, thanks @wikumChamith!👍

Binayak490-cyber:
@wikumChamith, @druchniewicz, @ibacher, @dkayiwa, can you take a look at it now, I have now moved the Envers audit table backfill from a JDBC Hibernate Integrator (runs every startup) into a Liquibase CustomTaskChange (runs exactly once per database), verified locally also and pushed it to the PR...?

druchniewicz:
> @wikumChamith, @druchniewicz, @ibacher, @dkayiwa, can you take a look at it now, I have now moved the Envers audit table backfill from a JDBC Hibernate Integrator (runs every startup) into a Liquibase CustomTaskChange (runs exactly once per database), verified locally also and pushed it to the PR...?

@Binayak490-cyber I'm trying to deploy your changes on my local instance but seems something does not work. I'm using openmrs-core 2.9.x version - from [this](https://github.com/openmrs/openmrs-core/tree/2.9.x) branch 

I copied BackfillEnversAuditTablesChangeset.java, EnversAuditTableInitializer.java and your change set to liquibase-update-to-latest-2.9.x (3.0.x xml does not exist in omrs-core 2.9.x) built and deployed it to my local instance. Seems that this migration didn't run at all because I don't see the entry in liquibasechangelog table in database. And audit tables are also empty. 

Could you verify it? Maybe I miss something? Can you try to run these changes on omrs-core 2.9.x as and check if it works for you?

Binayak490-cyber:
> > @wikumChamith, @druchniewicz, @ibacher, @dkayiwa, can you take a look at it now, I have now moved the Envers audit table backfill from a JDBC Hibernate Integrator (runs every startup) into a Liquibase CustomTaskChange (runs exactly once per database), verified locally also and pushed it to the PR...?
> 
> @Binayak490-cyber I'm trying to deploy your changes on my local instance but seems something does not work. I'm using openmrs-core 2.9.x version - from [this](https://github.com/openmrs/openmrs-core/tree/2.9.x) branch
> 
> I copied BackfillEnversAuditTablesChangeset.java, EnversAuditTableInitializer.java and your change set to liquibase-update-to-latest-2.9.x (3.0.x xml does not exist in omrs-core 2.9.x) built and deployed it to my local instance. Seems that this migration didn't run at all because I don't see the entry in liquibasechangelog table in database. And audit tables are also empty.
> 
> Could you verify it? Maybe I miss something? Can you try to run these changes on omrs-core 2.9.x as and check if it works for you?

@druchniewicz, thanks for testing! I wanted to clarify that this PR targets openmrs-core 3.0.x (master/3.0.0-SNAPSHOT), not 2.9.x. Our changeset is in liquibase-update-to-latest-3.0.x.xml, which only exists in the 3.0.x codebase.

The reason the migration didn't run on your 2.9.x instance is that:

1. The liquibase-update-to-latest-2.9.x.xml file is a different changelog — simply copying the changeset there doesn't guarantee Liquibase will pick it up the same way
2. The BackfillEnversAuditTablesChangeset class needs to be compiled into the deployed JAR — just copying the .java file is not enough
3. Hibernate Envers version differences between 2.9.x and 3.0.x may also affect compatibility

Could you test by building and running the full 3.0.x branch with our changes? That is the intended target for this PR. If a 2.9.x backport is needed, that would be a separate effort.

druchniewicz:
> > > @wikumChamith, @druchniewicz, @ibacher, @dkayiwa, can you take a look at it now, I have now moved the Envers audit table backfill from a JDBC Hibernate Integrator (runs every startup) into a Liquibase CustomTaskChange (runs exactly once per database), verified locally also and pushed it to the PR...?
> > 
> > 
> > @Binayak490-cyber I'm trying to deploy your changes on my local instance but seems something does not work. I'm using openmrs-core 2.9.x version - from [this](https://github.com/openmrs/openmrs-core/tree/2.9.x) branch
> > I copied BackfillEnversAuditTablesChangeset.java, EnversAuditTableInitializer.java and your change set to liquibase-update-to-latest-2.9.x (3.0.x xml does not exist in omrs-core 2.9.x) built and deployed it to my local instance. Seems that this migration didn't run at all because I don't see the entry in liquibasechangelog table in database. And audit tables are also empty.
> > Could you verify it? Maybe I miss something? Can you try to run these changes on omrs-core 2.9.x as and check if it works for you?
> 
> @druchniewicz, thanks for testing! I wanted to clarify that this PR targets openmrs-core 3.0.x (master/3.0.0-SNAPSHOT), not 2.9.x. Our changeset is in liquibase-update-to-latest-3.0.x.xml, which only exists in the 3.0.x codebase.
> 
> The reason the migration didn't run on your 2.9.x instance is that:
> 
> 1. The liquibase-update-to-latest-2.9.x.xml file is a different changelog — simply copying the changeset there doesn't guarantee Liquibase will pick it up the same way
> 2. The BackfillEnversAuditTablesChangeset class needs to be compiled into the deployed JAR — just copying the .java file is not enough
> 3. Hibernate Envers version differences between 2.9.x and 3.0.x may also affect compatibility
> 
> Could you test by building and running the full 3.0.x branch with our changes? That is the intended target for this PR. If a 2.9.x backport is needed, that would be a separate effort.

I'm using 2.9.x because there were some issues with running the latest omrs core snapshot version (3.0.0-snapshot). I used OpenMRS 2.8.3 version and I wanted to test changes related to hibernate envers that were merged recently (in February). When I deployed the latest version of omrs-core from master branch (3.0.0-snapshot) then the run app does not start correctly because of some failing migrations. I needed to back to stable 2.9.x branch (it also contains hibernate envers commit) to run the app successfully.

Binayak490-cyber:
> > > > @wikumChamith, @druchniewicz, @ibacher, @dkayiwa, can you take a look at it now, I have now moved the Envers audit table backfill from a JDBC Hibernate Integrator (runs every startup) into a Liquibase CustomTaskChange (runs exactly once per database), verified locally also and pushed it to the PR...?
> > > 
> > > 
> > > @Binayak490-cyber I'm trying to deploy your changes on my local instance but seems something does not work. I'm using openmrs-core 2.9.x version - from [this](https://github.com/openmrs/openmrs-core/tree/2.9.x) branch
> > > I copied BackfillEnversAuditTablesChangeset.java, EnversAuditTableInitializer.java and your change set to liquibase-update-to-latest-2.9.x (3.0.x xml does not exist in omrs-core 2.9.x) built and deployed it to my local instance. Seems that this migration didn't run at all because I don't see the entry in liquibasechangelog table in database. And audit tables are also empty.
> > > Could you verify it? Maybe I miss something? Can you try to run these changes on omrs-core 2.9.x as and check if it works for you?
> > 
> > 
> > @druchniewicz, thanks for testing! I wanted to clarify that this PR targets openmrs-core 3.0.x (master/3.0.0-SNAPSHOT), not 2.9.x. Our changeset is in liquibase-update-to-latest-3.0.x.xml, which only exists in the 3.0.x codebase.
> > The reason the migration didn't run on your 2.9.x instance is that:
> > 
> > 1. The liquibase-update-to-latest-2.9.x.xml file is a different changelog — simply copying the changeset there doesn't guarantee Liquibase will pick it up the same way
> > 2. The BackfillEnversAuditTablesChangeset class needs to be compiled into the deployed JAR — just copying the .java file is not enough
> > 3. Hibernate Envers version differences between 2.9.x and 3.0.x may also affect compatibility
> > 
> > Could you test by building and running the full 3.0.x branch with our changes? That is the intended target for this PR. If a 2.9.x backport is needed, that would be a separate effort.
> 
> I'm using 2.9.x because there were some issues with running the latest omrs core snapshot version (3.0.0-snapshot). I used OpenMRS 2.8.3 version and I wanted to test changes related to hibernate envers that were merged recently (in February). When I deployed the latest version of omrs-core from master branch (3.0.0-snapshot) then the run app does not start correctly because of some failing migrations. I needed to back to stable 2.9.x branch (it also contains hibernate envers commit) to run the app successfully.

Thank you for understanding @druchniewicz. Actually, since the Envers commit is also in 2.9.x, the backfill should work there too.

Since this PR is scoping for 3.0.0-SNAPSHOT only, I think it would be cleaner to keep it focused here. Once this gets accepted and merged, I can open a separate PR to add the AUDIT-28-2026-03-19 changeset to liquibase-update-to-latest-2.9.x.xml, rather than mixing both version changes in the same PR which will be very big for a single PR and also there are many tickets and issues in jira related around this, so separately doing all, I think will be good.

Regarding the failing migrations on 3.0.0-SNAPSHOT — is that a known issue being tracked separately? Just want to make sure our changes aren't contributing to it.

druchniewicz:
> > > > > @wikumChamith, @druchniewicz, @ibacher, @dkayiwa, can you take a look at it now, I have now moved the Envers audit table backfill from a JDBC Hibernate Integrator (runs every startup) into a Liquibase CustomTaskChange (runs exactly once per database), verified locally also and pushed it to the PR...?
> > > > 
> > > > 
> > > > @Binayak490-cyber I'm trying to deploy your changes on my local instance but seems something does not work. I'm using openmrs-core 2.9.x version - from [this](https://github.com/openmrs/openmrs-core/tree/2.9.x) branch
> > > > I copied BackfillEnversAuditTablesChangeset.java, EnversAuditTableInitializer.java and your change set to liquibase-update-to-latest-2.9.x (3.0.x xml does not exist in omrs-core 2.9.x) built and deployed it to my local instance. Seems that this migration didn't run at all because I don't see the entry in liquibasechangelog table in database. And audit tables are also empty.
> > > > Could you verify it? Maybe I miss something? Can you try to run these changes on omrs-core 2.9.x as and check if it works for you?
> > > 
> > > 
> > > @druchniewicz, thanks for testing! I wanted to clarify that this PR targets openmrs-core 3.0.x (master/3.0.0-SNAPSHOT), not 2.9.x. Our changeset is in liquibase-update-to-latest-3.0.x.xml, which only exists in the 3.0.x codebase.
> > > The reason the migration didn't run on your 2.9.x instance is that:
> > > 
> > > 1. The liquibase-update-to-latest-2.9.x.xml file is a different changelog — simply copying the changeset there doesn't guarantee Liquibase will pick it up the same way
> > > 2. The BackfillEnversAuditTablesChangeset class needs to be compiled into the deployed JAR — just copying the .java file is not enough
> > > 3. Hibernate Envers version differences between 2.9.x and 3.0.x may also affect compatibility
> > > 
> > > Could you test by building and running the full 3.0.x branch with our changes? That is the intended target for this PR. If a 2.9.x backport is needed, that would be a separate effort.
> > 
> > 
> > I'm using 2.9.x because there were some issues with running the latest omrs core snapshot version (3.0.0-snapshot). I used OpenMRS 2.8.3 version and I wanted to test changes related to hibernate envers that were merged recently (in February). When I deployed the latest version of omrs-core from master branch (3.0.0-snapshot) then the run app does not start correctly because of some failing migrations. I needed to back to stable 2.9.x branch (it also contains hibernate envers commit) to run the app successfully.
> 
> Thank you for understanding @druchniewicz. Actually, since the Envers commit is also in 2.9.x, the backfill should work there too.
> 
> Since this PR is scoping for 3.0.0-SNAPSHOT only, I think it would be cleaner to keep it focused here. Once this gets accepted and merged, I can open a separate PR to add the AUDIT-28-2026-03-19 changeset to liquibase-update-to-latest-2.9.x.xml, rather than mixing both version changes in the same PR which will be very big for a single PR.
> 
> Regarding the failing migrations on 3.0.0-SNAPSHOT — is that a known issue being tracked separately? Just want to make sure our changes aren't contributing to it.

@Binayak490-cyber if you have time you can just use the same changes for 2.9.x and check if they works. Just run the refdistro app (use e.g. dev3 tag for docker image), build the omrs-core module from 2.9.x branch and check if audit tables will be backfilled properly. 
I also tested your previous version of this and in general it worked fine but I saw some discrepancies e.g. patient_audit table was totally empty and patient references were not read properly(because of empty patient_audit table) in audit log page. I don't know why patient_audit was empty because other _audit tables looked like the they were well filled with data

Binayak490-cyber:
> > > > > > @wikumChamith, @druchniewicz, @ibacher, @dkayiwa, can you take a look at it now, I have now moved the Envers audit table backfill from a JDBC Hibernate Integrator (runs every startup) into a Liquibase CustomTaskChange (runs exactly once per database), verified locally also and pushed it to the PR...?
> > > > > 
> > > > > 
> > > > > @Binayak490-cyber I'm trying to deploy your changes on my local instance but seems something does not work. I'm using openmrs-core 2.9.x version - from [this](https://github.com/openmrs/openmrs-core/tree/2.9.x) branch
> > > > > I copied BackfillEnversAuditTablesChangeset.java, EnversAuditTableInitializer.java and your change set to liquibase-update-to-latest-2.9.x (3.0.x xml does not exist in omrs-core 2.9.x) built and deployed it to my local instance. Seems that this migration didn't run at all because I don't see the entry in liquibasechangelog table in database. And audit tables are also empty.
> > > > > Could you verify it? Maybe I miss something? Can you try to run these changes on omrs-core 2.9.x as and check if it works for you?
> > > > 
> > > > 
> > > > @druchniewicz, thanks for testing! I wanted to clarify that this PR targets openmrs-core 3.0.x (master/3.0.0-SNAPSHOT), not 2.9.x. Our changeset is in liquibase-update-to-latest-3.0.x.xml, which only exists in the 3.0.x codebase.
> > > > The reason the migration didn't run on your 2.9.x instance is that:
> > > > 
> > > > 1. The liquibase-update-to-latest-2.9.x.xml file is a different changelog — simply copying the changeset there doesn't guarantee Liquibase will pick it up the same way
> > > > 2. The BackfillEnversAuditTablesChangeset class needs to be compiled into the deployed JAR — just copying the .java file is not enough
> > > > 3. Hibernate Envers version differences between 2.9.x and 3.0.x may also affect compatibility
> > > > 
> > > > Could you test by building and running the full 3.0.x branch with our changes? That is the intended target for this PR. If a 2.9.x backport is needed, that would be a separate effort.
> > > 
> > > 
> > > I'm using 2.9.x because there were some issues with running the latest omrs core snapshot version (3.0.0-snapshot). I used OpenMRS 2.8.3 version and I wanted to test changes related to hibernate envers that were merged recently (in February). When I deployed the latest version of omrs-core from master branch (3.0.0-snapshot) then the run app does not start correctly because of some failing migrations. I needed to back to stable 2.9.x branch (it also contains hibernate envers commit) to run the app successfully.
> > 
> > 
> > Thank you for understanding @druchniewicz. Actually, since the Envers commit is also in 2.9.x, the backfill should work there too.
> > Since this PR is scoping for 3.0.0-SNAPSHOT only, I think it would be cleaner to keep it focused here. Once this gets accepted and merged, I can open a separate PR to add the AUDIT-28-2026-03-19 changeset to liquibase-update-to-latest-2.9.x.xml, rather than mixing both version changes in the same PR which will be very big for a single PR.
> > Regarding the failing migrations on 3.0.0-SNAPSHOT — is that a known issue being tracked separately? Just want to make sure our changes aren't contributing to it.
> 
> @Binayak490-cyber if you have time you can just use the same changes for 2.9.x and check if they works. Just run the refdistro app (use e.g. dev3 tag for docker image), build the omrs-core module from 2.9.x branch and check if audit tables will be backfilled properly. I also tested your previous version of this and in general it worked fine but I saw some discrepancies e.g. patient_audit table was totally empty and patient references were not read properly(because of empty patient_audit table) in audit log page. I don't know why patient_audit was empty because other _audit tables looked like the they were well filled with data

@druchniewicz, I tested the changes on 2.9.x using the refdistro app with the dev3 Docker image as you suggested. The backfill changeset works correctly on 2.9.x as well — AUDIT-28-2026-03-19 was recorded in liquibasechangelog and audit tables like concept_name_audit, concept_audit, concept_reference_map_audit were all populated with data.

I have ported the changeset to liquibase-update-to-latest-2.9.x.xml and pushed it to my fork. Should I open a separate PR targeting the 2.9.x branch for this as you can also check and test it in that branch...???

druchniewicz:
> > > > > > > @wikumChamith, @druchniewicz, @ibacher, @dkayiwa, can you take a look at it now, I have now moved the Envers audit table backfill from a JDBC Hibernate Integrator (runs every startup) into a Liquibase CustomTaskChange (runs exactly once per database), verified locally also and pushed it to the PR...?
> > > > > > 
> > > > > > 
> > > > > > @Binayak490-cyber I'm trying to deploy your changes on my local instance but seems something does not work. I'm using openmrs-core 2.9.x version - from [this](https://github.com/openmrs/openmrs-core/tree/2.9.x) branch
> > > > > > I copied BackfillEnversAuditTablesChangeset.java, EnversAuditTableInitializer.java and your change set to liquibase-update-to-latest-2.9.x (3.0.x xml does not exist in omrs-core 2.9.x) built and deployed it to my local instance. Seems that this migration didn't run at all because I don't see the entry in liquibasechangelog table in database. And audit tables are also empty.
> > > > > > Could you verify it? Maybe I miss something? Can you try to run these changes on omrs-core 2.9.x as and check if it works for you?
> > > > > 
> > > > > 
> > > > > @druchniewicz, thanks for testing! I wanted to clarify that this PR targets openmrs-core 3.0.x (master/3.0.0-SNAPSHOT), not 2.9.x. Our changeset is in liquibase-update-to-latest-3.0.x.xml, which only exists in the 3.0.x codebase.
> > > > > The reason the migration didn't run on your 2.9.x instance is that:
> > > > > 
> > > > > 1. The liquibase-update-to-latest-2.9.x.xml file is a different changelog — simply copying the changeset there doesn't guarantee Liquibase will pick it up the same way
> > > > > 2. The BackfillEnversAuditTablesChangeset class needs to be compiled into the deployed JAR — just copying the .java file is not enough
> > > > > 3. Hibernate Envers version differences between 2.9.x and 3.0.x may also affect compatibility
> > > > > 
> > > > > Could you test by building and running the full 3.0.x branch with our changes? That is the intended target for this PR. If a 2.9.x backport is needed, that would be a separate effort.
> > > > 
> > > > 
> > > > I'm using 2.9.x because there were some issues with running the latest omrs core snapshot version (3.0.0-snapshot). I used OpenMRS 2.8.3 version and I wanted to test changes related to hibernate envers that were merged recently (in February). When I deployed the latest version of omrs-core from master branch (3.0.0-snapshot) then the run app does not start correctly because of some failing migrations. I needed to back to stable 2.9.x branch (it also contains hibernate envers commit) to run the app successfully.
> > > 
> > > 
> > > Thank you for understanding @druchniewicz. Actually, since the Envers commit is also in 2.9.x, the backfill should work there too.
> > > Since this PR is scoping for 3.0.0-SNAPSHOT only, I think it would be cleaner to keep it focused here. Once this gets accepted and merged, I can open a separate PR to add the AUDIT-28-2026-03-19 changeset to liquibase-update-to-latest-2.9.x.xml, rather than mixing both version changes in the same PR which will be very big for a single PR.
> > > Regarding the failing migrations on 3.0.0-SNAPSHOT — is that a known issue being tracked separately? Just want to make sure our changes aren't contributing to it.
> > 
> > 
> > @Binayak490-cyber if you have time you can just use the same changes for 2.9.x and check if they works. Just run the refdistro app (use e.g. dev3 tag for docker image), build the omrs-core module from 2.9.x branch and check if audit tables will be backfilled properly. I also tested your previous version of this and in general it worked fine but I saw some discrepancies e.g. patient_audit table was totally empty and patient references were not read properly(because of empty patient_audit table) in audit log page. I don't know why patient_audit was empty because other _audit tables looked like the they were well filled with data
> 
> @druchniewicz, I tested the changes on 2.9.x using the refdistro app with the dev3 Docker image as you suggested. The backfill changeset works correctly on 2.9.x as well — AUDIT-28-2026-03-19 was recorded in liquibasechangelog and audit tables like concept_name_audit, concept_audit, concept_reference_map_audit were all populated with data.
> 
> I have ported the changeset to liquibase-update-to-latest-2.9.x.xml and pushed it to my fork. Should I open a separate PR targeting the 2.9.x branch for this as you can also check and test it in that branch...???

Yes, please open a separate PR for 2.9.x branch. Other basic tables like Patient, Encounters, Obs etc. have also been populated with all data?

Binayak490-cyber:
> > > > > > > > @wikumChamith, @druchniewicz, @ibacher, @dkayiwa, can you take a look at it now, I have now moved the Envers audit table backfill from a JDBC Hibernate Integrator (runs every startup) into a Liquibase CustomTaskChange (runs exactly once per database), verified locally also and pushed it to the PR...?
> > > > > > > 
> > > > > > > 
> > > > > > > @Binayak490-cyber I'm trying to deploy your changes on my local instance but seems something does not work. I'm using openmrs-core 2.9.x version - from [this](https://github.com/openmrs/openmrs-core/tree/2.9.x) branch
> > > > > > > I copied BackfillEnversAuditTablesChangeset.java, EnversAuditTableInitializer.java and your change set to liquibase-update-to-latest-2.9.x (3.0.x xml does not exist in omrs-core 2.9.x) built and deployed it to my local instance. Seems that this migration didn't run at all because I don't see the entry in liquibasechangelog table in database. And audit tables are also empty.
> > > > > > > Could you verify it? Maybe I miss something? Can you try to run these changes on omrs-core 2.9.x as and check if it works for you?
> > > > > > 
> > > > > > 
> > > > > > @druchniewicz, thanks for testing! I wanted to clarify that this PR targets openmrs-core 3.0.x (master/3.0.0-SNAPSHOT), not 2.9.x. Our changeset is in liquibase-update-to-latest-3.0.x.xml, which only exists in the 3.0.x codebase.
> > > > > > The reason the migration didn't run on your 2.9.x instance is that:
> > > > > > 
> > > > > > 1. The liquibase-update-to-latest-2.9.x.xml file is a different changelog — simply copying the changeset there doesn't guarantee Liquibase will pick it up the same way
> > > > > > 2. The BackfillEnversAuditTablesChangeset class needs to be compiled into the deployed JAR — just copying the .java file is not enough
> > > > > > 3. Hibernate Envers version differences between 2.9.x and 3.0.x may also affect compatibility
> > > > > > 
> > > > > > Could you test by building and running the full 3.0.x branch with our changes? That is the intended target for this PR. If a 2.9.x backport is needed, that would be a separate effort.
> > > > > 
> > > > > 
> > > > > I'm using 2.9.x because there were some issues with running the latest omrs core snapshot version (3.0.0-snapshot). I used OpenMRS 2.8.3 version and I wanted to test changes related to hibernate envers that were merged recently (in February). When I deployed the latest version of omrs-core from master branch (3.0.0-snapshot) then the run app does not start correctly because of some failing migrations. I needed to back to stable 2.9.x branch (it also contains hibernate envers commit) to run the app successfully.
> > > > 
> > > > 
> > > > Thank you for understanding @druchniewicz. Actually, since the Envers commit is also in 2.9.x, the backfill should work there too.
> > > > Since this PR is scoping for 3.0.0-SNAPSHOT only, I think it would be cleaner to keep it focused here. Once this gets accepted and merged, I can open a separate PR to add the AUDIT-28-2026-03-19 changeset to liquibase-update-to-latest-2.9.x.xml, rather than mixing both version changes in the same PR which will be very big for a single PR.
> > > > Regarding the failing migrations on 3.0.0-SNAPSHOT — is that a known issue being tracked separately? Just want to make sure our changes aren't contributing to it.
> > > 
> > > 
> > > @Binayak490-cyber if you have time you can just use the same changes for 2.9.x and check if they works. Just run the refdistro app (use e.g. dev3 tag for docker image), build the omrs-core module from 2.9.x branch and check if audit tables will be backfilled properly. I also tested your previous version of this and in general it worked fine but I saw some discrepancies e.g. patient_audit table was totally empty and patient references were not read properly(because of empty patient_audit table) in audit log page. I don't know why patient_audit was empty because other _audit tables looked like the they were well filled with data
> > 
> > 
> > @druchniewicz, I tested the changes on 2.9.x using the refdistro app with the dev3 Docker image as you suggested. The backfill changeset works correctly on 2.9.x as well — AUDIT-28-2026-03-19 was recorded in liquibasechangelog and audit tables like concept_name_audit, concept_audit, concept_reference_map_audit were all populated with data.
> > I have ported the changeset to liquibase-update-to-latest-2.9.x.xml and pushed it to my fork. Should I open a separate PR targeting the 2.9.x branch for this as you can also check and test it in that branch...???
> 
> Yes, please open a separate PR for 2.9.x branch. Other basic tables like Patient, Encounters, Obs etc. have also been populated with all data?

Thank you @druchniewicz then I will open a separate PR for the 2.9.x branch shortly.

Regarding patient_audit, encounter_audit, obs_audit — in my local test I used a fresh database, so those tables were empty simply because there were no existing patients, encounters or observations to backfill. The backfill logic ran correctly for all tables that had data (global_property_audit, privilege_audit, concept_name_audit etc.).

To properly verify patient_audit, encounter_audit and obs_audit I would need to test against a database with existing clinical data, which is what you have in your 2.8.3 instance. Would you be able to test the 2.9.x PR once I open it?

druchniewicz:
> Regarding patient_audit, encounter_audit, obs_audit — in my local test I used a fresh database, so those tables were empty simply because there were no existing patients, encounters or observations to backfill. The backfill logic ran correctly for all tables that had data (global_property_audit, privilege_audit, concept_name_audit etc.).
> 
> To properly verify patient_audit, encounter_audit and obs_audit I would need to test against a database with existing clinical data, which is what you have in your 2.8.3 instance. Would you be able to test the 2.9.x PR once I open it?

You can simply test it by yourself. Just register a patient, fill in any visit form and you should have data in patient/encounter/obs audit tables. Then you can just remove these audit tables, restart the server and these tables should be created with expected data.

Binayak490-cyber:
> > > > > > > > @wikumChamith, @druchniewicz, @ibacher, @dkayiwa, can you take a look at it now, I have now moved the Envers audit table backfill from a JDBC Hibernate Integrator (runs every startup) into a Liquibase CustomTaskChange (runs exactly once per database), verified locally also and pushed it to the PR...?
> > > > > > > 
> > > > > > > 
> > > > > > > @Binayak490-cyber I'm trying to deploy your changes on my local instance but seems something does not work. I'm using openmrs-core 2.9.x version - from [this](https://github.com/openmrs/openmrs-core/tree/2.9.x) branch
> > > > > > > I copied BackfillEnversAuditTablesChangeset.java, EnversAuditTableInitializer.java and your change set to liquibase-update-to-latest-2.9.x (3.0.x xml does not exist in omrs-core 2.9.x) built and deployed it to my local instance. Seems that this migration didn't run at all because I don't see the entry in liquibasechangelog table in database. And audit tables are also empty.
> > > > > > > Could you verify it? Maybe I miss something? Can you try to run these changes on omrs-core 2.9.x as and check if it works for you?
> > > > > > 
> > > > > > 
> > > > > > @druchniewicz, thanks for testing! I wanted to clarify that this PR targets openmrs-core 3.0.x (master/3.0.0-SNAPSHOT), not 2.9.x. Our changeset is in liquibase-update-to-latest-3.0.x.xml, which only exists in the 3.0.x codebase.
> > > > > > The reason the migration didn't run on your 2.9.x instance is that:
> > > > > > 
> > > > > > 1. The liquibase-update-to-latest-2.9.x.xml file is a different changelog — simply copying the changeset there doesn't guarantee Liquibase will pick it up the same way
> > > > > > 2. The BackfillEnversAuditTablesChangeset class needs to be compiled into the deployed JAR — just copying the .java file is not enough
> > > > > > 3. Hibernate Envers version differences between 2.9.x and 3.0.x may also affect compatibility
> > > > > > 
> > > > > > Could you test by building and running the full 3.0.x branch with our changes? That is the intended target for this PR. If a 2.9.x backport is needed, that would be a separate effort.
> > > > > 
> > > > > 
> > > > > I'm using 2.9.x because there were some issues with running the latest omrs core snapshot version (3.0.0-snapshot). I used OpenMRS 2.8.3 version and I wanted to test changes related to hibernate envers that were merged recently (in February). When I deployed the latest version of omrs-core from master branch (3.0.0-snapshot) then the run app does not start correctly because of some failing migrations. I needed to back to stable 2.9.x branch (it also contains hibernate envers commit) to run the app successfully.
> > > > 
> > > > 
> > > > Thank you for understanding @druchniewicz. Actually, since the Envers commit is also in 2.9.x, the backfill should work there too.
> > > > Since this PR is scoping for 3.0.0-SNAPSHOT only, I think it would be cleaner to keep it focused here. Once this gets accepted and merged, I can open a separate PR to add the AUDIT-28-2026-03-19 changeset to liquibase-update-to-latest-2.9.x.xml, rather than mixing both version changes in the same PR which will be very big for a single PR.
> > > > Regarding the failing migrations on 3.0.0-SNAPSHOT — is that a known issue being tracked separately? Just want to make sure our changes aren't contributing to it.
> > > 
> > > 
> > > @Binayak490-cyber if you have time you can just use the same changes for 2.9.x and check if they works. Just run the refdistro app (use e.g. dev3 tag for docker image), build the omrs-core module from 2.9.x branch and check if audit tables will be backfilled properly. I also tested your previous version of this and in general it worked fine but I saw some discrepancies e.g. patient_audit table was totally empty and patient references were not read properly(because of empty patient_audit table) in audit log page. I don't know why patient_audit was empty because other _audit tables looked like the they were well filled with data
> > 
> > 
> > @druchniewicz, I tested the changes on 2.9.x using the refdistro app with the dev3 Docker image as you suggested. The backfill changeset works correctly on 2.9.x as well — AUDIT-28-2026-03-19 was recorded in liquibasechangelog and audit tables like concept_name_audit, concept_audit, concept_reference_map_audit were all populated with data.
> > I have ported the changeset to liquibase-update-to-latest-2.9.x.xml and pushed it to my fork. Should I open a separate PR targeting the 2.9.x branch for this as you can also check and test it in that branch...???
> 
> Yes, please open a separate PR for 2.9.x branch. Other basic tables like Patient, Encounters, Obs etc. have also been populated with all data?

@druchniewicz, I have now opened PR #5952 targeting the 2.9.x branch with the same backfill changeset ported over with the similar changes as I have done in this PR but this PR is for master 3.0.0-SNAPSHOT only, so you can check it!

sudhanshu-raj:
> > > > > @wikumChamith, @druchniewicz, @ibacher, @dkayiwa, can you take a look at it now, I have now moved the Envers audit table backfill from a JDBC Hibernate Integrator (runs every startup) into a Liquibase CustomTaskChange (runs exactly once per database), verified locally also and pushed it to the PR...?
> > > > 
> > > > 
> > > > @Binayak490-cyber I'm trying to deploy your changes on my local instance but seems something does not work. I'm using openmrs-core 2.9.x version - from [this](https://github.com/openmrs/openmrs-core/tree/2.9.x) branch
> > > > I copied BackfillEnversAuditTablesChangeset.java, EnversAuditTableInitializer.java and your change set to liquibase-update-to-latest-2.9.x (3.0.x xml does not exist in omrs-core 2.9.x) built and deployed it to my local instance. Seems that this migration didn't run at all because I don't see the entry in liquibasechangelog table in database. And audit tables are also empty.
> > > > Could you verify it? Maybe I miss something? Can you try to run these changes on omrs-core 2.9.x as and check if it works for you?
> > > 
> > > 
> > > @druchniewicz, thanks for testing! I wanted to clarify that this PR targets openmrs-core 3.0.x (master/3.0.0-SNAPSHOT), not 2.9.x. Our changeset is in liquibase-update-to-latest-3.0.x.xml, which only exists in the 3.0.x codebase.
> > > The reason the migration didn't run on your 2.9.x instance is that:
> > > 
> > > 1. The liquibase-update-to-latest-2.9.x.xml file is a different changelog — simply copying the changeset there doesn't guarantee Liquibase will pick it up the same way
> > > 2. The BackfillEnversAuditTablesChangeset class needs to be compiled into the deployed JAR — just copying the .java file is not enough
> > > 3. Hibernate Envers version differences between 2.9.x and 3.0.x may also affect compatibility
> > > 
> > > Could you test by building and running the full 3.0.x branch with our changes? That is the intended target for this PR. If a 2.9.x backport is needed, that would be a separate effort.
> > 
> > 
> > I'm using 2.9.x because there were some issues with running the latest omrs core snapshot version (3.0.0-snapshot). I used OpenMRS 2.8.3 version and I wanted to test changes related to hibernate envers that were merged recently (in February). When I deployed the latest version of omrs-core from master branch (3.0.0-snapshot) then the run app does not start correctly because of some failing migrations. I needed to back to stable 2.9.x branch (it also contains hibernate envers commit) to run the app successfully.
> 
> Thank you for understanding @druchniewicz. Actually, since the Envers commit is also in 2.9.x, the backfill should work there too.
> 
> Since this PR is scoping for 3.0.0-SNAPSHOT only, I think it would be cleaner to keep it focused here. Once this gets accepted and merged, I can open a separate PR to add the AUDIT-28-2026-03-19 changeset to liquibase-update-to-latest-2.9.x.xml, rather than mixing both version changes in the same PR which will be very big for a single PR and also there are many tickets and issues in jira related around this, so separately doing all, I think will be good.
> 
> Regarding the failing migrations on 3.0.0-SNAPSHOT — is that a known issue being tracked separately? Just want to make sure our changes aren't contributing to it.

I might be mistaken, but regarding the snapshot version, it seems that we cannot run the OpenMRS Core 3.0.0 snapshot with the Docker-based Reference Application. From what I understand, deployments work only up to Core 2.9.x.

The likely reason is the Tomcat version mismatch. The snapshot version uses Tomcat 10 or above, while the Reference Application Docker setup appears to be using a Tomcat version below 10. Because of this, it probably only supports Core up to 2.9.x.

Also, even when I tried running the snapshot version manually on a local Tomcat setup, the audit web module did not run smoothly.

Just to confirm, have you tested all these changes in a real application environment using the snapshot version along with the existing audit web module? I’m asking because I tried running the audit web module with the current snapshot, but it did not work.

Binayak490-cyber:
> > > > > > @wikumChamith, @druchniewicz, @ibacher, @dkayiwa, can you take a look at it now, I have now moved the Envers audit table backfill from a JDBC Hibernate Integrator (runs every startup) into a Liquibase CustomTaskChange (runs exactly once per database), verified locally also and pushed it to the PR...?
> > > > > 
> > > > > 
> > > > > @Binayak490-cyber I'm trying to deploy your changes on my local instance but seems something does not work. I'm using openmrs-core 2.9.x version - from [this](https://github.com/openmrs/openmrs-core/tree/2.9.x) branch
> > > > > I copied BackfillEnversAuditTablesChangeset.java, EnversAuditTableInitializer.java and your change set to liquibase-update-to-latest-2.9.x (3.0.x xml does not exist in omrs-core 2.9.x) built and deployed it to my local instance. Seems that this migration didn't run at all because I don't see the entry in liquibasechangelog table in database. And audit tables are also empty.
> > > > > Could you verify it? Maybe I miss something? Can you try to run these changes on omrs-core 2.9.x as and check if it works for you?
> > > > 
> > > > 
> > > > @druchniewicz, thanks for testing! I wanted to clarify that this PR targets openmrs-core 3.0.x (master/3.0.0-SNAPSHOT), not 2.9.x. Our changeset is in liquibase-update-to-latest-3.0.x.xml, which only exists in the 3.0.x codebase.
> > > > The reason the migration didn't run on your 2.9.x instance is that:
> > > > 
> > > > 1. The liquibase-update-to-latest-2.9.x.xml file is a different changelog — simply copying the changeset there doesn't guarantee Liquibase will pick it up the same way
> > > > 2. The BackfillEnversAuditTablesChangeset class needs to be compiled into the deployed JAR — just copying the .java file is not enough
> > > > 3. Hibernate Envers version differences between 2.9.x and 3.0.x may also affect compatibility
> > > > 
> > > > Could you test by building and running the full 3.0.x branch with our changes? That is the intended target for this PR. If a 2.9.x backport is needed, that would be a separate effort.
> > > 
> > > 
> > > I'm using 2.9.x because there were some issues with running the latest omrs core snapshot version (3.0.0-snapshot). I used OpenMRS 2.8.3 version and I wanted to test changes related to hibernate envers that were merged recently (in February). When I deployed the latest version of omrs-core from master branch (3.0.0-snapshot) then the run app does not start correctly because of some failing migrations. I needed to back to stable 2.9.x branch (it also contains hibernate envers commit) to run the app successfully.
> > 
> > 
> > Thank you for understanding @druchniewicz. Actually, since the Envers commit is also in 2.9.x, the backfill should work there too.
> > Since this PR is scoping for 3.0.0-SNAPSHOT only, I think it would be cleaner to keep it focused here. Once this gets accepted and merged, I can open a separate PR to add the AUDIT-28-2026-03-19 changeset to liquibase-update-to-latest-2.9.x.xml, rather than mixing both version changes in the same PR which will be very big for a single PR and also there are many tickets and issues in jira related around this, so separately doing all, I think will be good.
> > Regarding the failing migrations on 3.0.0-SNAPSHOT — is that a known issue being tracked separately? Just want to make sure our changes aren't contributing to it.
> 
> I might be mistaken, but regarding the snapshot version, it seems that we cannot run the OpenMRS Core 3.0.0 snapshot with the Docker-based Reference Application. From what I understand, deployments work only up to Core 2.9.x.
> 
> The likely reason is the Tomcat version mismatch. The snapshot version uses Tomcat 10 or above, while the Reference Application Docker setup appears to be using a Tomcat version below 10. Because of this, it probably only supports Core up to 2.9.x.
> 
> Also, even when I tried running the snapshot version manually on a local Tomcat setup, the audit web module did not run smoothly.
> 
> Just to confirm, have you tested all these changes in a real application environment using the snapshot version along with the existing audit web module? I’m asking because I tried running the audit web module with the current snapshot, but it did not work.

@sudhanshu-raj, you are correct but not fully about that the 3.0.0-SNAPSHOT uses Tomcat 10+ (Jakarta EE) while the refdistro Docker image uses Tomcat 9 — this is a known compatibility issue.

For local testing of the 3.0.x changes I used a custom Docker setup with tomcat:10.1-jdk21 as the base image and replaced the WAR directly, which allowed the backfill to be verified successfully — audit tables like privilege_audit, global_property_audit, concept_name_audit were all populated correctly.

Regarding the audit web module not working on the snapshot — that is a separate issue from what this PR addresses. This PR only fixes the backfill of pre-existing data into audit tables at the core level. The audit web module compatibility with 3.0.x is outside the scope of this PR.

For 2.9.x (PR #5952), the refdistro dev3 Docker image works correctly since it runs Tomcat 9, and the backfill was verified there as well.

druchniewicz:
> @druchniewicz, I have now opened PR #5952 targeting the 2.9.x branch with the same backfill changeset ported over with the similar changes as I have done in this PR but this PR is for master 3.0.0-SNAPSHOT only, so you can check it!

@Binayak490-cyber I fetched your PR and deployed it on my local machine and seems that this migration is not executed at all. I'm 100% sure that openmrs.war with your changes is deployed on my local machine properly. I don't see the appropriate entry in liquibasechangelog table `select * from liquibasechangelog where ID = 'AUDIT-28-2026-03-19';` gives me empty result

Binayak490-cyber:
> > @druchniewicz, I have now opened PR #5952 targeting the 2.9.x branch with the same backfill changeset ported over with the similar changes as I have done in this PR but this PR is for master 3.0.0-SNAPSHOT only, so you can check it!
> 
> @Binayak490-cyber I fetched your PR and deployed it on my local machine and seems that this migration is not executed at all. I'm 100% sure that openmrs.war with your changes is deployed on my local machine properly. I don't see the appropriate entry in liquibasechangelog table `select * from liquibasechangelog where ID = 'AUDIT-28-2026-03-19';` gives me empty result

@druchniewicz, have you tried this PR #5952, this is the real PR for 2.9.x branch...!

Binayak490-cyber:
> > > @druchniewicz, I have now opened PR #5952 targeting the 2.9.x branch with the same backfill changeset ported over with the similar changes as I have done in this PR but this PR is for master 3.0.0-SNAPSHOT only, so you can check it!
> > 
> > 
> > @Binayak490-cyber I fetched your PR and deployed it on my local machine and seems that this migration is not executed at all. I'm 100% sure that openmrs.war with your changes is deployed on my local machine properly. I don't see the appropriate entry in liquibasechangelog table `select * from liquibasechangelog where ID = 'AUDIT-28-2026-03-19';` gives me empty result
> 
> @druchniewicz, have you tried this PR #5952, this is the real PR for 2.9.x branch...!

https://github.com/openmrs/openmrs-core/pull/5952#issuecomment-4104672983, @druchniewicz, please go for my this comment....!

api/src/test/java/org/openmrs/api/context/DaemonTest.java
```diff
@@ -238,7 +237,7 @@ public void execute() throws InterruptedException {
 
 			// another.join(10000); doesn't actually work as runInNewDaemonThread doesn't use the Thread object rather it
 			// only executes the run method. The only way to determine it completed is to wait for wasRun to return true.
-			await().atMost(10, TimeUnit.SECONDS).untilTrue(new AtomicBoolean(wasRun));
+			await().atMost(10, TimeUnit.SECONDS).until(() -> wasRun);
```
wikumChamith, line 165: What is the reason for this change?

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
```diff
@@ -0,0 +1,294 @@
...
+		}
+		return identifier;
+	}
+
+	/**
+	 * Returns true if the given audit table exists but contains no rows.
+	 */
+	static boolean isAuditTableEmpty(Connection connection, String tableName) {
+		try (Statement stmt = connection.createStatement();
+		        ResultSet rs = stmt.executeQuery("SELECT COUNT(*) FROM " + requireSafeIdentifier(tableName))) {
```
wikumChamith: Can't we use `SELECT 1 FROM <table> LIMIT 1`?

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
```diff
@@ -0,0 +1,294 @@
...
+			return false;
+		}
+	}
+
+	/**
+	 * Returns true if the given source table has no rows.
+	 */
+	static boolean isTableEmpty(Connection connection, String tableName) throws SQLException {
+		try (Statement stmt = connection.createStatement();
+		        ResultSet rs = stmt.executeQuery("SELECT COUNT(*) FROM " + requireSafeIdentifier(tableName))) {
```
wikumChamith: Can't we use SELECT 1 FROM <table> LIMIT 1?

api/src/test/java/org/openmrs/api/context/DaemonTest.java
Binayak490-cyber: @wikumChamith, untilTrue(new AtomicBoolean(wasRun)) I think is actually broken - it wraps the current boolean value of wasRun into a new AtomicBoolean at the time of the call, so it never reflects future changes to wasRun. It would either pass immediately or never pass.

until(() -> wasRun) is correct as the lambda captures the variable by reference and is re-evaluated on each poll, so Awaitility actually observes when wasRun becomes true.

So the old code was functionally incorrect and the fix makes the condition actually work as intended.

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
Binayak490-cyber: @wikumChamith, SELECT COUNT(*), I used here because it's standard SQL and works across all databases OpenMRS supports (MySQL, MariaDB, H2) without any syntax differences. However, your one is more efficient of that SELECT 1 FROM LIMIT 1 - it short-circuits on the first row found instead of scanning the entire table to count all rows, which can be costly on large audit tables. 
Will update it to use SELECT 1 FROM LIMIT 1 with !rs.next() as the empty check.

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
Binayak490-cyber: Same goes for here too.

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
Binayak490-cyber: @wikumChamith, updated.

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
Binayak490-cyber: Updated.

api/src/test/java/org/openmrs/api/context/DaemonTest.java
wikumChamith: Is this related to the current ticket/PR?

Binayak490-cyber: Actually @wikumChamith, This is not directly related to AUDIT-28, it was a minor fix noticed while working on the PR. If you want then i can remove it.

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
```diff
@@ -0,0 +1,294 @@
...
+			Connection connection = ((JdbcConnection) database.getConnection()).getUnderlyingConnection();
+
+			String revisionTableName = findRevisionEntityTable(connection);
+			Integer revId = null;
+
+			DatabaseMetaData metaData = connection.getMetaData();
+			try (ResultSet tables = metaData.getTables(null, null, "%", new String[] { "TABLE" })) {
+				while (tables.next()) {
+					String tableName = tables.getString("TABLE_NAME");
+					if (!tableName.endsWith(AUDIT_SUFFIX)) {
```
wikumChamith: What happens if someone is using a custom SUFFIX?

`org.hibernate.envers.audit_table_suffix=_AUD`

druchniewicz: I guess @Binayak490-cyber you handled it in another [PR](https://github.com/openmrs/openmrs-core/pull/5952) but not in this PR yet?

Binayak490-cyber: Yes I also got this and made change in the PR for 2.9.x branch and I think you are correct that the fix works for the default OpenMRS setup but breaks for any deployment using a custom Envers suffix - the changeset would skip all their audit tables and backfill would never run. The fix would be to make the suffix configurable via a Liquibase <param> with _audit as the default fallback using a setAuditSuffix(String) setter. Want me to make this change?

Binayak490-cyber: > I guess @Binayak490-cyber you handled it in another [PR](https://github.com/openmrs/openmrs-core/pull/5952) but not in this PR yet?

yes exactly @druchniewicz, I said that only, I addressed in the other PR but not in this, will update it now...!

Binayak490-cyber: This were edge cases I missed in this PR but added in the other PR.

Binayak490-cyber: @druchniewicz, @wikumChamith, updated....!

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
```diff
@@ -0,0 +1,298 @@
...
+	static String getRevisionPrimaryKeyColumn(Connection connection, String revisionTableName) throws SQLException {
+		DatabaseMetaData metaData = connection.getMetaData();
+		for (String name : new String[] { revisionTableName, revisionTableName.toUpperCase() }) {
+			try (ResultSet rs = metaData.getPrimaryKeys(null, null, name)) {
+				if (rs.next()) {
+					return rs.getString("COLUMN_NAME");
+				}
+			}
+		}
+		return "id";
```
ibacher: This seems like a very bad assumption

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
```diff
@@ -0,0 +1,298 @@
...
+	 * @throws SQLException if the revision entry cannot be created
+	 */
+	static int createBackfillRevision(Connection connection, String revisionTableName) throws SQLException {
+		String pkColumn = getRevisionPrimaryKeyColumn(connection, revisionTableName);
+		String timestampColumn = getRevisionTimestampColumn(connection, revisionTableName);
+		int nextId;
+		try (Statement stmt = connection.createStatement();
+		        ResultSet rs = stmt.executeQuery("SELECT COALESCE(MAX(" + requireSafeIdentifier(pkColumn) + "), 0) + 1 FROM "
+		                + requireSafeIdentifier(revisionTableName))) {
+			nextId = rs.next() ? rs.getInt(1) : 1;
```
ibacher: `1` seems like a bad default assumption here.

api/src/test/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangesetTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * Unit tests for {@link BackfillEnversAuditTablesChangeset}, focused on the backfill behaviour
+ * introduced to fix AUDIT-28: "Unable to read" values in audit tables for entities that existed
+ * before auditing was enabled.
+ * <p>
+ * These tests call the package-private helper methods directly against a real H2 in-memory
+ * database.
+ */
+class BackfillEnversAuditTablesChangesetTest {
```
ibacher, line 34: Why are we running any tests against a completely dummy database instead of using the existing `BaseModuleContextSensitiveTest` class that includes a database that at least _resembles_ the database we need to support.

Binayak490-cyber: @ibacher, the raw H2 approach was intentional - these tests are focused on validating the helper methods (createBackfillRevision, isAuditTableEmpty, isTableEmpty, backfillTable) in complete isolation, without the overhead of loading the full OpenMRS Spring context. Since these methods only depend on a JDBC Connection and basic SQL operations, a minimal H2 schema is sufficient to verify their correctness. Using BaseModuleContextSensitiveTest would make the tests significantly heavier and slower for what are essentially unit-level checks on SQL logic. If you say to migrate to BaseModuleContextSensitiveTest or if you feel the tests need to run against a more realistic schema then I can refactor the tests to that but it will much slower the tests - could you confirm if that's the direction you'd prefer?

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
Binayak490-cyber: yes @ibacher, If the primary key column cannot be determined from JDBC metadata, silently falling back to "id" could cause incorrect inserts or silent data corruption. Fixed it to throw a SQLException instead so the failure is explicit and visible.

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
Binayak490-cyber: Agreed @ibacher, as the SELECT COALESCE(MAX(...), 0) + 1 is an aggregate query that always returns exactly one row, so defaulting to 1 when rs.next() returns false was masking a serious underlying problem. Fixed it to throw a SQLException instead.

Binayak490-cyber:
@ibacher, @wikumChamith, @druchniewicz, then can you now check the new changes after @ibacher 's review comments are addressed....???

tendomart:
@Binayak490-cyber  link to the issue missing

Binayak490-cyber:
> @Binayak490-cyber link to the issue missing

@tendomart, the jira ticket issue number is already given in the PR title....!

Binayak490-cyber:
@ibacher, @wikumChamith, @druchniewicz, then can you now check the new changes after @ibacher 's review comments are addressed....???

Binayak490-cyber:
@ibacher, @wikumChamith, @druchniewicz, then can you now check the new changes after @ibacher 's review comments are addressed....???

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
```diff
@@ -0,0 +1,301 @@
...
+				return revId;
+			}
+			if (revId == null) {
+				revId = createBackfillRevision(connection, revisionTableName);
+			}
+			List<String> columns = getAuditTableDataColumns(connection, auditTable);
+			if (!columns.isEmpty()) {
+				backfillTable(connection, sourceTable, auditTable, columns, revId);
+			}
+		} catch (SQLException e) {
```
druchniewicz: seems that these errors will be swallowed and there can be situation with inconsistent database state. I mean some operation can pass, some can fail but migration will be marked as executed and will never re-run in the future. Is my understanding correct?

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
```diff
@@ -0,0 +1,301 @@
...
+	/**
+	 * Returns true if the given audit table exists but contains no rows.
+	 */
+	static boolean isAuditTableEmpty(Connection connection, String tableName) {
+		try (Statement stmt = connection.createStatement();
+		        ResultSet rs = stmt.executeQuery("SELECT 1 FROM " + requireSafeIdentifier(tableName) + " LIMIT 1")) {
+			return !rs.next();
+		} catch (SQLException e) {
+			log.debug("Audit table {} not accessible, skipping backfill: {}", tableName, e.getMessage());
+			return false;
```
druchniewicz: not sure if it's correct assumption to correct false if something went wrong

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
```diff
@@ -0,0 +1,301 @@
...
+ * Because this is a Liquibase changeset it is tracked in {@code databasechangelog} and runs exactly
+ * once per database, never on subsequent startups.
+ */
+public class BackfillEnversAuditTablesChangeset implements CustomTaskChange {
+
+	private static final Logger log = LoggerFactory.getLogger(BackfillEnversAuditTablesChangeset.class);
+
+	private static final Pattern SAFE_SQL_IDENTIFIER = Pattern.compile("[a-zA-Z_]\\w*");
+
+	private String auditSuffix = "_audit";
```
druchniewicz, line 53: does it mean that we still support only _audit prefix? If so, this is wrong because this suffix is configurable and can be different than '_audit'

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
```diff
@@ -0,0 +1,301 @@
...
+			Connection connection = ((JdbcConnection) database.getConnection()).getUnderlyingConnection();
+
+			String revisionTableName = findRevisionEntityTable(connection);
+			Integer revId = null;
+
+			DatabaseMetaData metaData = connection.getMetaData();
+			try (ResultSet tables = metaData.getTables(null, null, "%", new String[] { "TABLE" })) {
+				while (tables.next()) {
+					String tableName = tables.getString("TABLE_NAME");
+					if (!tableName.endsWith(auditSuffix)) {
```
druchniewicz: not sure but can be situation that different sql providers treat these suffixes in different way, e.g. one provider creates PATIENT_AUDIT table, second provider patient_audit table. It may be so, but it does not have to be. But you can ignore case sensitivity and protect it easily

druchniewicz:
@Binayak490-cyber I left some more comments.

@wikumChamith @ibacher @dkayiwa can you take a look at this PR and check if it's going in right direction? Would be good to have this before next June release but as it is change in openmrs-core module we should be doubly careful about the changes we introduce

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
```diff
@@ -0,0 +1,301 @@
...
+	 * @param revisionTableName name of the revision entity table
+	 * @return the primary key column name, falling back to "id" if not found
+	 * @throws SQLException if metadata cannot be read
+	 */
+	static String getRevisionPrimaryKeyColumn(Connection connection, String revisionTableName) throws SQLException {
+		DatabaseMetaData metaData = connection.getMetaData();
+		for (String name : new String[] { revisionTableName, revisionTableName.toUpperCase() }) {
+			try (ResultSet rs = metaData.getPrimaryKeys(null, null, name)) {
+				if (rs.next()) {
+					return rs.getString("COLUMN_NAME");
```
jwnasambu: Nice progress @Binayak490-cyber. Kindly it seems you missed out to capture this build failure which is caused by the string literal "COLUMN_NAME" is used 4 times in this class, which triggers SonarQube rule  to solve this you need to define a constant instead of duplicating this literal".
```suggestion
					private static final String COLUMN_NAME = "COLUMN_NAME";
```

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
Binayak490-cyber: Yes @druchniewicz! You're partially correct - the class is flexible via `setAuditSuffix()`, but the concern about consistency is valid.

The suffix `_audit` matches what's already configured in `hibernate.default.properties`:

```
org.hibernate.envers.audit_table_suffix=_audit
```

The real issue was the same value existing in both the Java field default and the XML `<param>`, creating a risk of them drifting out of sync. Fixed by:

* Removing the redundant `<param name="auditSuffix" value="_audit"/>` from the XML
* Adding a comment on the field pointing to `hibernate.default.properties` as the source of truth

```java
// Must match org.hibernate.envers.audit_table_suffix in hibernate.default.properties
private String auditSuffix = "_audit";
```

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
Binayak490-cyber: Yes, corrected.

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
Binayak490-cyber: Corrected.

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
Binayak490-cyber: Corrected.

api/src/main/java/org/openmrs/util/databasechange/BackfillEnversAuditTablesChangeset.java
Binayak490-cyber: Corrected.

Binayak490-cyber:
@ibacher, @dkayiwa, @druchniewicz, @jwnasambu, @wikumChamith, I have addressed the review comments, can you please go for checking out the latest changes...!

Binayak490-cyber:
@ibacher, @dkayiwa, @druchniewicz, @jwnasambu, @wikumChamith, I have addressed the review comments, can you please go for checking out the latest changes...!

druchniewicz:
@dkayiwa @wikumChamith can you take a look on this if it fits your expectations? If not, can you provide your feedback?

wikumChamith:
@Binayak490-cyber what do you think about these comments from the AI agent? Do you think they're valid?

druchniewicz:
> @Binayak490-cyber what do you think about these comments from the AI agent? Do you think they're valid?

@Binayak490-cyber reminder

PR #5934, opened 2026-03-15
TRUNK-4988: Convenience method to allow us to change the configuration more easily

shivvani-r:
I have implemented dynamic Spring dependency injection for `MessageService` in `Context.java` to allow easier configuration of message senders and preparators.

**Why the change was needed:**
Previously, the `MessageService` dependencies (`MessageSender` and `MessagePreparator`) were strictly hardcoded to instantiate `MailMessageSender` and `VelocityMessagePreparator`. This rigid coupling prevented hospitals or developers from easily swapping in custom modules (such as an SMS sender) without modifying and recompiling the core Java code.

**Specific changes include:**
* Replaced the manual instantiations in `Context.getMessageService()` with dynamic Spring bean lookups using `getRegisteredComponents()`.
* Added fallback defensive logic to preserve backward compatibility: if no custom beans are registered in the Spring context, it safely defaults to the original `MailMessageSender` and `VelocityMessagePreparator`.
* Added a warning log if multiple beans of the same type are registered, ensuring developers know which bean was selected.
* Removed the outdated private static helper methods for creating the default senders.
* Added comprehensive JUnit integration tests in `ContextTest.java` to verify that dynamically registered `MessageSender` and `MessagePreparator` beans are successfully discovered and loaded from the `ApplicationContext`.

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-4988

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Context.java
+++ b/api/src/main/java/org/openmrs/api/context/Context.java
@@ -64,7 +64,6 @@ import org.openmrs.messagesource.MessageSourceService;
 import org.openmrs.module.ModuleMustStartException;
 import org.openmrs.module.ModuleUtil;
 import org.openmrs.notification.AlertService;
-import org.openmrs.notification.MessageException;
 import org.openmrs.notification.MessagePreparator;
 import org.openmrs.notification.MessageSender;
 import org.openmrs.notification.MessageService;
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -565,13 +564,31 @@ public static ProgramWorkflowService getProgramWorkflowService() {
...
+				List<MessageSender> senders = getRegisteredComponents(MessageSender.class);
+				if (senders != null && !senders.isEmpty()) {
+					if (senders.size() > 1) {
+						log.warn("Multiple MessageSenders found. Using the first one: {}",
+						    senders.get(0).getClass().getName());
+					}
+					ms.setMessageSender(senders.get(0));
+				} else {
+					ms.setMessageSender(new MailMessageSender(getMailSession()));
+				}
```
jwnasambu: Kindly refactor the duplicate logic in MessagePreparator and MessageSender into a private helper method to make the flow much easier to maintain.

jwnasambu (changes requested):

jwnasambu (commented):
Kindly we need to verify that the system correctly falls back to the default implementations MailMessageSender, VelocityMessagePreparator when no beans are registered.

In addition, MessageService is a singleton, modifying it in a test can break other tests if the state isn't perfectly restored. I would recommend adding this test case using a try-finally block to guarantee cleanup. Otherwise, Great work!

shivvani-r:
> Kindly we need to verify that the system correctly falls back to the default implementations MailMessageSender, VelocityMessagePreparator when no beans are registered.
> 
> In addition, MessageService is a singleton, modifying it in a test can break other tests if the state isn't perfectly restored. I would recommend adding this test case using a try-finally block to guarantee cleanup. Otherwise, Great work!

Thanks for the feedback, @jwnasambu! I've extracted the duplicate logic into a `getSingleRegisteredComponent` helper method. I also added the `getMessageService_shouldFallbackToDefaultSenderAndPreparatorWhenNoBeansRegisteredtest` (with a try-finally cleanup block) to explicitly verify the fallback behavior. Let me know if everything looks good to go!

cc: @dkayiwa @ibacher

PR #5950, opened 2026-03-19
TRUNK-6588: OWASP check fails due to arg parsing and missing NVD key

jwnasambu:

I updated the OWASP Dependency Check GitHub Actions workflow to fix the dependency scan and improve reliability:

1. **Fix arg parsing** — Pinned the action to a current SHA (`1e54355a`) that supports the `args` input. The old pin (`75ba02d6`, v1.1.0 from 2021) silently discarded all `args`, so `--failOnCVSS 6.2` never reached the scanner and the check always passed.
2. **Consolidate steps** — Removed the redundant PR/master split (20 duplicated arg lines). The NVD API key expression already handles the empty-secret case correctly on both triggers.
3. **Add `--disableCentral`** — Avoids flaky Central-search timeouts (the analyzer can stall for 30+ min). This narrows scan coverage, which is documented inline.
4. **Add `--disableOssIndex`** — No-op; the analyzer already auto-disables without credentials.
5. **Fix `args` format** — Changed from `>` (folded scalar) to `|` (literal scalar) so each flag is passed as a separate argument to the scanner.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6588


jwnasambu:
@dkayiwa, @ibacher I have tried fixing this issue but OWASP Dependency Check is still failing because it detected vulnerable dependencies above the configured CVSS threshold and I am torn apart from these three options [Update the dependency](https://mvnrepository.com/artifact/com.mchange/mchange-commons-java/versions),  Suppress the finding or  Relax the threshold which may become less secure. Kindly I need your guide on this.

dependency-check-suppressions.xml
```diff
@@ -26,4 +26,15 @@
 	   <packageUrl regex="true">^pkg:javascript/jquery@.*$</packageUrl>
 	   <vulnerabilityName regex="true">^jquery issue:.*$</vulnerabilityName>
 	</suppress>
+	 <suppress until="2026-09-22Z">
```
ibacher: This doesn't seem to belong with this PR?

ibacher (commented):
Thanks @jwnasambu! However I noticed the OWASP check is still failing for the same reason....

jwnasambu:
@dkayiwa Kindly free to review my PR at your convenient time please!

jwnasambu:
@claude review

jwnasambu:
@ibacher, @dkayiwa, @rkorytkowski  Kindly feel free to review my PR at your convenient time please!

.github/workflows/dependency-check.yml
```diff
@@ -16,32 +16,69 @@ jobs:
...
+        with:
+          path: /github/home/.dependency-check/data
+          key: nvd-${{ runner.os }}-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            nvd-${{ runner.os }}-
+
+      - name: Run OWASP Dependency-Check (PR)
+        if: ${{ github.event_name == 'pull_request' }}
+        timeout-minutes: 15
+        uses: dependency-check/Dependency-Check_Action@75ba02d6183445fe0761d26e836bde58b1560600
```
dkayiwa: This pin needs to change before the PR can merge. `75ba02d6` is tag 1.1.0, from April 2021, and that version of `action.yml` has no `args` input at all: it declares `project`, `path`, `format`, `others`, and passes `'${{ inputs.others }}'` as the final argument. Everything under `args:` below is silently discarded. Compare the two docker invocations. This PR's run:

```
... owasp/dependency-check-action:latest "--project" "openmrs-core" "--scan" "." "--format" "ALL" "--out" "/github/workspace/reports" "--noupdate" ""
```

and master's push run from the 14th:

```
... "--noupdate" "--failOnCVSS 6.2 --enableRetired --suppression dependency-check-suppressions.xml --nvdApiKey ***"
```

The step also logs `Unexpected input(s) 'out', valid inputs are ['entryPoint', 'args', 'project', 'path', 'format', 'others']`, which is the same version mismatch surfacing.

So `--failOnCVSS 6.2` never reaches the scanner, and the CLI default is to never fail on severity. That is the only reason this check is green: the report this very run uploaded lists seven unsuppressed CVEs between 6.5 and 8.5 in `mysql-connector-j-9.7.0.jar`, and the job passed anyway. Your TRUNK-5731 PR ran a few minutes earlier this afternoon on master's workflow and failed on exactly those CVEs. I reproduced it locally against the same image with just that one jar: with the args replaced by `""` the tool exits 0, with them present it exits 15 and prints the failOnCVSS error.

Merged as is, every PR and every master push reports a passing dependency scan while enforcing no threshold at all, and a green check is not something anyone goes back and double-checks.

Keeping a SHA pin is the right instinct, it just needs to be a current one:

```suggestion
        uses: dependency-check/Dependency-Check_Action@1e54355a8b4c8abaa8cc7d0b70aa655a3bb15a6c
```

That is the action's `main` today, and its `action.yml` does take `args`. The second step on line 61 needs the same change. One caveat worth knowing either way: 1.1.0 runs `docker://owasp/dependency-check-action:latest` directly, so the old pin was never pinning the scanner version.

dkayiwa (commented):
Picking up the question you asked back in March that never got an answer: of the three options I would update the dependency, and I would not relax the threshold or reach for a blanket suppression.

What is tripping the gate on master today is `mysql-connector-j` 9.7.0, seven CVEs scoring between 6.5 and 8.5, and NVD scopes every one of them to versions 9.7.0-9.7.1. Three of the seven (CVE-2026-60193, CVE-2026-60192, CVE-2026-60317) are Connector/Net rather than Connector/J, so they are false positives for us, but they reach us through the shared `oracle:mysql_connectors` CPE, which means suppressing your way out costs seven entries and goes stale the next time Oracle publishes a batch. Central has 26.7.0, which sits outside the affected range; whether it drops in cleanly here is worth checking. That is its own ticket though, not something to fold into this one.

.github/workflows/dependency-check.yml
```diff
@@ -16,32 +16,69 @@ jobs:
...
-        id: Depcheck
+        run: mvn clean install -DskipTests -Dspotbugs.skip=true
+
+      - name: Cache NVD database
+        uses: actions/cache@v4
+        with:
+          path: /github/home/.dependency-check/data
+          key: nvd-${{ runner.os }}-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            nvd-${{ runner.os }}-
```
dkayiwa: This cache step cannot work, and it takes the `--data` flag down with it. `actions/cache` runs on the runner host, whereas `/github/home` only exists inside the container GitHub spins up for a Docker action (it is where `/home/runner/work/_temp/_github_home` gets mounted). There is no such path on the host, so this run logged `Cache not found for input keys` on restore and `Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved` on save. It will do that on every run.

It turns into a hard failure the moment the pin on line 41 is fixed, because then `--data /github/home/.dependency-check/data` (lines 56 and 76) does reach the scanner, pointed at a directory nothing ever populates. The action always passes `--noupdate`, so there is no download to fall back on. Running that combination against the same image:

```
[ERROR] One or more fatal errors occurred
[ERROR] Autoupdate is disabled and the database does not exist
```

exit 13. Fixing line 41 on its own would therefore take the job from wrongly green to red for a brand new reason, so the two need to land together.

I would drop this step and both `--data` lines. The image ships the database itself at `/usr/share/dependency-check/data`, freshly built (the report this run uploaded records `NVD API Last Checked: 2026-08-16T11:44:14Z`, which is the image build, not the run), and that is what master's runs and this run are both already scanning against. There is nothing here worth caching.

.github/workflows/dependency-check.yml
```diff
@@ -16,32 +16,69 @@ jobs:
...
+          out: 'reports'
+          args: |
+            --failOnCVSS 6.2
+            --enableRetired
+            --disableCentral
+            --disableOssIndex
+            --suppression dependency-check-suppressions.xml
+            --data /github/home/.dependency-check/data
+
+      - name: Run OWASP Dependency-Check (master push)
```
dkayiwa: I do not think this second step needs to exist. The parse error TRUNK-6588 describes was fixed on master back in June by #6210, which added the very expression you have copied down to line 77: when the secret is empty it expands to nothing, so no bare `--nvdApiKey` is emitted and there is nothing left to fail on. Your own TRUNK-5731 run this afternoon shows it working from a fork, its args line reads `--failOnCVSS 6.2 --enableRetired --suppression dependency-check-suppressions.xml` with no key and no error.

So the split is not buying behaviour the single step lacks, and it costs twenty duplicated lines of arguments that now have to be kept in sync; whoever next changes `--failOnCVSS` has to remember there are two of them. It also leaves a gap: the two `if:` conditions cover exactly `pull_request` and `push` to master, so the day someone adds `schedule:` or `workflow_dispatch:` to the triggers, neither step runs and the job goes green having scanned nothing.

Worth knowing before you decide: the API key is not doing anything in this workflow either way. Both versions of the action hardcode `--noupdate`, so dependency-check never calls the NVD API and the database comes from the image (`Finished NVD CVE Analyzer (0 seconds)` in the master run, with no download).

What is actually red on master right now is real findings over the threshold plus intermittent `Could not connect to Central search`, and both are different from the problem in the ticket. It would help to update the PR description, and the ticket, to say which of them this PR is now fixing.

.github/workflows/dependency-check.yml
```diff
@@ -16,32 +16,69 @@ jobs:
...
+        with:
+          project: 'openmrs-core'
+          path: '.'
+          format: 'ALL'
+          out: 'reports'
+          args: |
+            --failOnCVSS 6.2
+            --enableRetired
+            --disableCentral
+            --disableOssIndex
```
dkayiwa: Was `--disableCentral` aimed at the `Could not connect to Central search` failures? If so the motivation makes sense, the master push on the 14th sat in the Central Analyzer for 1916 seconds and then died on exactly that. Upstream is blunt about the price, though: "Disabling this analyzer is not recommended as it could lead to false negatives (e.g. libraries that have vulnerabilities may not be reported correctly)." That is a real narrowing of what the gate catches, and nothing in the PR description mentions it, so I would rather it were a decision we wrote down than a side effect of a workflow tidy-up.

`--disableOssIndex` is harmless, that analyzer already switches itself off: `Sonatype OSS Index Analyzer disabled due to missing credentials`.

.github/workflows/dependency-check.yml
```diff
@@ -16,32 +16,69 @@ jobs:
...
+        uses: actions/cache@v4
+        with:
+          path: /github/home/.dependency-check/data
+          key: nvd-${{ runner.os }}-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            nvd-${{ runner.os }}-
+
+      - name: Run OWASP Dependency-Check (PR)
+        if: ${{ github.event_name == 'pull_request' }}
+        timeout-minutes: 15
```
dkayiwa: Is the 15 minute cap deliberate, here and on line 60? The job-level `timeout-minutes: 60` a few lines above was set in #6277 with a comment recording that runs occasionally tail out to around 35 minutes, and the master push on 2026-08-14 spent 32m23s in this step. A 15 minute step cap would have killed that run rather than letting it finish. If the thinking is that `--disableCentral` removes the long tail, that is worth saying in the comment; otherwise I would leave the 60 minute job budget to do the bounding.

bom/pom.xml
ibacher, line 1: The stuff around protobuf seems unrelated to the rest of the PR?

jwnasambu:
Closed by mistake

.github/workflows/dependency-check.yml
```diff
@@ -32,9 +32,14 @@ jobs:
           path: '.'
           format: 'ALL'
           out: 'reports'
-          args: >
+          # --disableCentral avoids flaky Central-search timeouts (the analyzer
+          # can stall for 30+ min). --disableOssIndex is a no-op here (the
+          # analyzer auto-disables without credentials).
+          args: |
             --failOnCVSS 6.2
```
dkayiwa, line 39: This threshold is what fails the job, and nothing here moves it, so `dependencycheck_test` stays red on every master push after this merges. The seven master runs between 26 and 31 August all ended on `One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '6.2'`, always the same two artifacts: `mysql-connector-j-9.7.0.jar` with seven CVEs between 6.5 and 8.5, and `spring-security-crypto-7.1.1.jar` with CVE-2018-1258 at 8.8. I pulled the report artifact off the master-tip run (33430985248) to check whether `--disableCentral` would take those out, and it will not. Neither is identified through the Central analyzer, their evidence is Manifest, jar and hint only, and none of the eight entries carrying vulnerabilities in that report has any `central` evidence on it.

The symptom TRUNK-6588 describes cannot happen any more either, for reasons that predate this PR. #6210 replaced `--nvdApiKey ${{ secrets.NVD_API_KEY }}` with the guarded expression now on line 44 back in June, and be87167 dropped the `pull_request` trigger on 31 August, so the workflow does not run on pull requests at all. That is why this PR has no `dependencycheck_test` check of its own.

So I would take this as what it is, pinning the action off `main` and dropping the Central analyzer, and not as the fix for TRUNK-6588. If it merges under that ticket, the ticket closes while the gate is still failing on every push, and the two vulnerable dependencies stop being anyone's item.

Could you redo the description before it goes in? Points 1 and 2 describe an earlier state of this branch rather than the change against master: master is on `@main`, not `75ba02d6`, and it has carried a single dependency-check step throughout, the PR/master split having been introduced and then removed inside this branch. Point 5 is not a change either. `>` and `|` produce the same argv here, because the entrypoint does `all_args=$@` and then expands `${all_args}` unquoted, and newline is in the default IFS alongside space. I ran both forms through that entrypoint and got the same 16 arguments, which matches the master logs, where the folded form reaches the scanner correctly split.

.github/workflows/dependency-check.yml
```diff
@@ -21,9 +21,9 @@ jobs:
           distribution: 'temurin'
           cache: 'maven'
       - name: Build project with Maven
-        run: mvn clean install -DskipTests
+        run: mvn clean install -DskipTests -Dspotbugs.skip=true
```
dkayiwa, line 24: This is not in the ticket, and it is not in the description's list of five changes either, so anyone reading it has to work out for themselves whether skipping static analysis inside the security-scan job weakens the scan. It does not, as far as I can see: spotbugs is bound to `verify` with `failOnError=false`, nothing in this job reads `target/spotbugsXml.xml` (only build.yaml uploads it), and on the master-tip run it accounted for about 1m29s of a 4m30s build step. That is a fair saving, but it is work a reviewer has to do rather than read. I would take it out and let it go in on its own, or failing that say in the description what it is doing here.

.github/workflows/dependency-check.yml
```diff
@@ -32,9 +32,14 @@ jobs:
...
           format: 'ALL'
           out: 'reports'
-          args: >
+          # --disableCentral avoids flaky Central-search timeouts (the analyzer
+          # can stall for 30+ min). --disableOssIndex is a no-op here (the
+          # analyzer auto-disables without credentials).
+          args: |
             --failOnCVSS 6.2
             --enableRetired
+            --disableCentral
```
dkayiwa, line 41: While you are in this file, the timeout rationale on line 12 is worth fixing, since this flag is what makes it wrong. It blames the long runs on "a stalled NVD feed download" and "NVD-update tails up to ~35" minutes, but the action hardcodes `--noupdate` and the NVD CVE Analyzer finished in 0 seconds on every run I looked at. The tail it is describing is this Central analyzer: 1971s on 26 August, 1640s on the 27th, 847s on the master tip, against 5s and 8s on the quick runs. Once `--disableCentral` is in, that tail goes away and the comment is left justifying a 60 minute budget by a mechanism that never fires.

PR #5953, opened 2026-03-20
Refactor DoubleRange.contains() to simplify logic and remove dead code

VarshithReddy2006:
## Description
The existing implementation of `DoubleRange.contains()` includes redundant conditional checks and unreachable branches due to the fixed values of `closedLow` and `closedHigh`.

This PR simplifies the method while preserving its original behavior.

## Changes
- Simplified `contains()` logic to directly evaluate the range condition `[low, high)`
- Removed conditional branches that were never executed
- Added unit tests to verify boundary behavior (low inclusive, high exclusive)

## Reasoning
The previous implementation relied on flags (`closedLow`, `closedHigh`) that are never modified, resulting in dead code. Simplifying the method improves readability and maintainability without altering functionality.

## Verification
- Added `DoubleRangeTest` covering:
  - value inside range
  - value equal to low (inclusive)
  - value equal to high (exclusive)
- All tests pass successfully


api/src/main/resources/org/openmrs/liquibase/snapshots/schema-only/liquibase-schema-only-2.8.x.xml
```diff
@@ -2407,8 +2407,8 @@
       <column defaultValueNumeric="0" name="state" type="INT"> 
         <constraints nullable="false"/> 
       </column>  
-      <column name="start_date" type="date"/>  
-      <column name="end_date" type="date"/>  
+      <column name="start_date" type="datetime(0)"/>
+      <column name="end_date" type="datetime(0)"/>
```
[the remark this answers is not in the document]
VarshithReddy2006: Thanks for the feedback and review!

I realized that this PR included unrelated changes and mixed commit history, which made it harder to review. To keep things clean and focused, I have created a new PR that contains only the DoubleRange refactor and associated tests.

New PR: https://github.com/openmrs/openmrs-core/pull/5954

Closing this PR to avoid confusion. Thanks again for the review!

api/src/test/java/org/openmrs/util/DoubleRangeTest.java
```diff
@@ -3,270 +3,91 @@
...
-
-	/**
-	 * @see DoubleRange#toString()
-	 */
-	@Test
-	public void toString_shouldPrintEmptyLowIfLowIsNull() {
-		DoubleRange r = new DoubleRange(null, 1.0);
-		assertEquals("< 1.0", r.toString());
-	}
-}
```
RujataGaikwad: This refactor simplifies the contains() logic 👍

However, I noticed that many existing tests were removed. 
Would it be better to keep those tests to maintain overall test coverage, 
and only update/add tests related to contains()?

Removing them might reduce regression safety.

VarshithReddy2006: Thanks for the feedback!

I’ve addressed this by restoring the existing tests and keeping only the additional boundary tests for contains() in the new PR (#5954).

Please feel free to review it there. Thanks again!

PR #5954, opened 2026-03-20
Refactor DoubleRange.contains() to simplify logic and remove dead code

VarshithReddy2006:
Jira Ticket  
https://openmrs.atlassian.net/browse/TRUNK-6589  

Description  
The current implementation of DoubleRange.contains() has redundant and unreachable conditions due to fixed values of closedLow and closedHigh.  

This PR simplifies the method while preserving the existing behavior and aligning with the class design.  

Changes  
- Removed redundant and unreachable conditions  
- Simplified the contains() logic  
- Removed unnecessary null checks based on the assumption that bounds represent infinite values when not explicitly set  
- Restored existing test cases  

Behavior  
No functional changes:  

- Lower bound is inclusive  
- Upper bound is exclusive  
- Existing behavior is preserved  

Verification  
All existing tests pass


api/src/test/java/org/openmrs/util/DoubleRangeTest.java
```diff
@@ -3,270 +3,97 @@
...
+    public void toStringShouldReturnRangeRepresentation() {
+        DoubleRange range = new DoubleRange(1.0, 10.0);
+        String text = range.toString();
+
+        assertNotNull(text);
+        assertTrue(text.contains("1.0"));
+        assertTrue(text.contains("10.0"));
+    }
+
+    @Test
```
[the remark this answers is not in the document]
VarshithReddy2006: Thanks for the detailed review!

- Added explicit handling for NaN values in contains() so they are treated as out of range.
- Restored the full OpenMRS license header in the test file.

For the remaining suggestions, I focused on preserving existing behavior and keeping the scope of this PR limited to the refactor and necessary test coverage improvements.

Chinmay7070:
hi @VarshithReddy2006
Could you please add the related Jira ticket (e.g., TRUNK-XXXX) to the PR description? It would help with better traceability.
https://devmanual.openmrs.org/technology/developmentprocess/

VarshithReddy2006:
Hi @Chinmay7070,
Thanks for the suggestion! I've added the Jira ticket and updated the PR description accordingly.

Please let me know if anything else is needed.

api/src/test/java/org/openmrs/util/DoubleRangeTest.java
```diff
@@ -3,270 +3,91 @@
  * v. 2.0. If a copy of the MPL was not distributed with this file, You can
  * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
  * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
- *
```
sagarkhandagre998: Why did you remove the comments ?

api/src/test/java/org/openmrs/util/DoubleRangeTest.java
```diff
@@ -3,270 +3,91 @@
...
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
-/**
```
sagarkhandagre998: same here

sagarkhandagre998 (changes requested):

api/src/main/java/org/openmrs/util/DoubleRange.java
```diff
@@ -123,26 +123,21 @@ public int compareTo(DoubleRange other) {
...
+			return false;
 		}
-		if (high != null) {
-			if (closedHigh) {
-				//unreachable code as closedHigh is never set to true anywhere
-				return d <= high;
-			} else {
-				return d < high;
-			}
+		if (low == null && high == null) {
```
Kapil-kundu: no need to check null because we already set high and low as Positve and Negative Infinity. 
so low and high never be null.

api/src/test/java/org/openmrs/util/DoubleRangeTest.java
```diff
@@ -11,262 +11,92 @@
...
-		Object objTest3 = new Object();
-
-		assertTrue(obj.equals(objTest1));
-		assertFalse(obj.equals(objTest2));
-		assertFalse(obj.equals(objTest3));
-	}
-
-	@Test
-	public void DoubleRange_shouldReturnNullLowAndHighIfAccessorsAreNotCalled() {
-		DoubleRange dr = new DoubleRange();
```
Kapil-kundu: you should not remove all the test cases.
you can add or update an existing test case according to output we want.
But deletion a test case may create problem in future

api/src/test/java/org/openmrs/util/DoubleRangeTest.java
```diff
@@ -11,262 +11,92 @@
...
+        DoubleRange range = new DoubleRange(1.0, 10.0);
+        String text = range.toString();
+
+        assertNotNull(text);
+        assertTrue(text.contains("1.0"));
+        assertTrue(text.contains("10.0"));
+    }
+
+    @Test
+    void shouldHandleNullBoundsSafely() {
```
Kapil-kundu: No need to add null test case because high and low never be null

Kapil-kundu (commented):
Do not remove necessary test cases

And do not use chatgpt or any AI tool everywhere. Take your time and try yourself

api/src/test/java/org/openmrs/util/DoubleRangeTest.java
VarshithReddy2006: Hi @Kapil-kundu,

Just a quick correction — the latest update removes the null-bound handling based on reviewer feedback, since low and high are effectively treated as infinite bounds and are not expected to be null in practice.

The current implementation focuses on simplifying the logic while preserving the original behavior.

Thanks!

VarshithReddy2006: Hi everyone,

I’ve updated the PR based on the feedback.

- Simplified the contains() logic and removed redundant conditions
- Restored all existing test cases
- Added handling for the case when both bounds are null

All tests are passing now.

Please let me know if any further changes are needed. Thanks!

api/src/main/java/org/openmrs/util/DoubleRange.java
```diff
@@ -123,26 +123,18 @@ public int compareTo(DoubleRange other) {
...
-				return d < high;
-			}
+
+		if (low != null && d < low) {
+			return false;
+		}
+
+		if (high != null && d >= high) {
+			return false;
 		}
```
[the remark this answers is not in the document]
0-YuvrajSingh: Good point about the NaN behavior change.

Just to confirm, the intention is that NaN should always be treated as out of range, right?
If so, maybe it would help to add an explicit test for NaN to make sure future refactors don’t accidentally change this behavior.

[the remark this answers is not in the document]
VarshithReddy2006: Hi, just a gentle follow-up.

@0-YuvrajSingh Thanks for pointing this out!

The intention of this refactor is to preserve the existing behavior of contains(). Based on the previous implementation, NaN is treated as out of range, and I’ve kept the behavior consistent.

@Kapil-kundu I’ve also addressed the review comments by restoring the original test cases and simplifying the contains() logic without introducing additional checks.

All tests are passing locally. Please let me know if any further changes are needed.

Thanks!

api/src/test/java/org/openmrs/util/DoubleRangeTest.java
VarshithReddy2006: Thanks for pointing that out!

The comments were unintentionally removed during refactoring of the test file. I’ve restored the necessary comments now to maintain consistency and readability with the existing test structure.

Please let me know if anything else needs to be adjusted.

api/src/test/java/org/openmrs/util/DoubleRangeTest.java
VarshithReddy2006: I’ve restored the removed comments including the license header and test documentation. 

Please let me know if anything is still missing.

VarshithReddy2006: Thanks for the review. I have restored the removed comments including the license header and test documentation.

Could you please review again when you get a chance?

PR #5959, opened 2026-03-21
TRUNK-5362: Fix extreme slowness in getEncountersByPatientIdentifier …

0-YuvrajSingh:
## Summary
Fixes extreme slowness (~20 seconds) in EncounterService.getEncountersByPatientIdentifier() 
reported after upgrading from OpenMRS 1.12 to 2.0.6.

## Root Cause
The method was calling `PatientService.getPatients()` which triggers a full Lucene text 
search. This is extremely slow compared to a direct database lookup.

## Fix
Replaced `getPatients(identifier, ...)` with `getPatientIdentifiers(identifier, ...)` 
which performs a direct indexed lookup on the `patient_identifier` table via Hibernate.

## Testing
- Build passes: `mvn clean package -DskipTests` ✅
- Voided identifiers are explicitly excluded from results

**Jira:** https://issues.openmrs.org/browse/TRUNK-5362
Fixes: TRUNK-5362


omeriinnocent (commented):
@0-YuvrajSingh ,i think is better to include a jira ticket link for your work here.

0-YuvrajSingh:
> @0-YuvrajSingh ,i think is better to include a jira ticket link for your work here.

Thanks! Added the Jira link to the PR description.

api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
```diff
@@ -69,961 +70,1005 @@
 @Transactional
 public class EncounterServiceImpl extends BaseOpenmrsService implements EncounterService, RefByUuid {
 
-	@Autowired
```
sagarkhandagre998: @0-YuvrajSingh 

There is a lot of unnecessary removals , How we'll gonna review this.

Please do implementation in a clean manner

sagarkhandagre998 (changes requested):

api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
0-YuvrajSingh: Cleaned up the diff — pushed a new commit with only the import and getEncountersByPatientIdentifier change. Sorry for the noise earlier!

arti-nova:
This looks like a solid performance improvement !
Switching from getPatients() to getPatientIdentifiers() makes sense to avoid the expensive Lucene search.

A couple of suggestions:

-It might be helpful to add a brief comment in the code explaining why getPatientIdentifiers() is preferred here (for future maintainability).
-Do we need any additional tests or benchmarks to demonstrate the performance improvement?

api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
```diff
@@ -375,8 +376,10 @@ public List<Encounter> getEncountersByPatientIdentifier(String identifier) throw
 		}
 
 		List<Encounter> encs = new ArrayList<>();
-		for (Patient p : Context.getPatientService().getPatients(identifier, null, null, false)) {
-			encs.addAll(Context.getEncounterService().getEncountersByPatientId(p.getPatientId()));
+		for (PatientIdentifier pi : Context.getPatientService().getPatientIdentifiers(identifier, null, null, null, null)) {
+			if (!pi.getVoided()) {
```
sagarkhandagre998: Check sonar warning.

sagarkhandagre998 (changes requested):

0-YuvrajSingh:
> This looks like a solid performance improvement ! Switching from getPatients() to getPatientIdentifiers() makes sense to avoid the expensive Lucene search.
> 
> A couple of suggestions:
> 
> -It might be helpful to add a brief comment in the code explaining why getPatientIdentifiers() is preferred here (for future maintainability). -Do we need any additional tests or benchmarks to demonstrate the performance improvement?

Thanks for the review! Added an inline comment explaining why getPatientIdentifiers() is preferred over getPatients() to avoid the Lucene-based full-text search. Regarding benchmarks — the existing tests cover the functional behavior. I can add a benchmark if the team feels it's necessary, but happy to defer to your judgment on that.

api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
0-YuvrajSingh: Fixed! Sonar flagged a boxed Boolean unboxing risk on !pi.getVoided() — replaced with !Boolean.TRUE.equals(pi.getVoided()) which is null-safe. Quality Gate passes with 0 new issues now.

0-YuvrajSingh:
Addressed the requested changes, cleaned the diff, and synced with latest upstream/master.
Quality gate passes and branch is up to date.
Waiting for workflow approval and re-review.

api/src/main/java/org/openmrs/api/impl/EncounterServiceImpl.java
jwnasambu, line 388: Note: A single patient can have multiple identifiers that all match the same query. In other words, two identifier types sharing the same value. Based on what you have fixed the loop will hit that patient twice and call getEncountersByPatientId() twice, adding the same encounters to encs twice.

I will suggest we add a Set<Integer> to track which patient IDs have already been processed so that the encounters are only fetched once per patient.

ibacher:
@0-YuvrajSingh Do you have any benchmarking to show that this makes a performance difference?

PR #5967, opened 2026-03-25
Improve deployment instructions in README

Kartheekmurala:
## Description
Improved the deployment section in README.

## Changes
- Added explanation for Jetty run command
- Added access URL
- Added troubleshooting steps

## Why
Helps new contributors understand how to run the project locally

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,18 @@ cd openmrs-core/webapp
 ../mvnw jetty:run
 ```
 
+This command starts a local development server using Jetty.
+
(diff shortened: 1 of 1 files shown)


0-YuvrajSingh:
Nice improvement to the README, this makes the local setup much clearer.

One suggestion:
should we also mention the required Java and Maven versions here?
New contributors sometimes run into startup issues due to version mismatches.

It might also help to suggest running `mvn clean install`
before `mvn jetty:run`, since missing dependencies can cause startup failures.

Kartheekmurala:
Got it, thanks for pointing that out! I’ll make the changes and update the PR shortly.

PR #5974, opened 2026-03-27
TRUNK-6594: Correct Logger class initialization in PatientValidator

shivvani-r:
I corrected the class reference in the logger initialization for `PatientValidator.java`.

**Why the change was needed:**
The logger was previously initialized with `PersonNameValidator.class` likely due to a copy-paste error. This caused all log messages (debug, info, error) generated by the `PatientValidator` to be incorrectly attributed to the `PersonNameValidator` in the server logs, making production debugging and log filtering difficult.

**Specific changes include:**
* Updated the `LoggerFactory.getLogger()` call to use `PatientValidator.class` instead of `PersonNameValidator.class`.

## Issue I worked on
See [TRUNK-6594](https://issues.openmrs.org/browse/TRUNK-6594)

- [ ] I have added tests to cover my changes. (Verified locally with a reflection-based unit test to confirm the logger name matches the class).

[TRUNK-6594]: https://openmrs.atlassian.net/browse/TRUNK-6594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/src/main/java/org/openmrs/validator/PatientValidator.java
+++ b/api/src/main/java/org/openmrs/validator/PatientValidator.java
@@ -26,7 +26,7 @@ import org.springframework.validation.ValidationUtils;
 @Handler(supports = { Patient.class }, order = 25)
 public class PatientValidator extends PersonValidator {
 
-	private static final Logger log = LoggerFactory.getLogger(PersonNameValidator.class);
+	private static final Logger log = LoggerFactory.getLogger(PatientValidator.class);
 
 	@Autowired
 	private PatientIdentifierValidator patientIdentifierValidator;


IvanescuChristian:
Good catch — the fix is correct and consistent with all other validators in the codebase (PersonValidator, UserValidator, ConceptValidator all use their own class).
Now the check list mentions you have locally added tests to check your changes. Wouldn`t adding the tests also make a great addition?
I also noticed a similar inconsistency inside ShortPatientFormValidator . Will do some further investigation then check if a ticket has already been made on the issue or report it shortly. (edit) It is part of the legacy UI so of no consequence.

shivvani-r:
> Good catch — the fix is correct and consistent with all other validators in the codebase (PersonValidator, UserValidator, ConceptValidator all use their own class). Now the check list mentions you have locally added tests to check your changes. Wouldn`t adding the tests also make a great addition? I also noticed a similar inconsistency inside ShortPatientFormValidator . Will do some further investigation then check if a ticket has already been made on the issue or report it shortly. (edit) It is part of the legacy UI so of no consequence.

Thanks for the review! 

I did write a quick test locally just to double-check my sanity before pushing, but I opted not to commit it because it requires using Java Reflection to access the `private static final` logger field. 

Since verifying the logger's internal string name isn't really testing the public behavior of the validator, I felt that adding a reflection-based test might be overkill and clutter the test suite for a one-line typo fix. I wanted to keep this PR as atomic as possible.

For context, here is the snippet I ran locally to verify it:

```java
@Test
public void shouldHaveCorrectLoggerName() throws Exception {
    java.lang.reflect.Field field = PatientValidator.class.getDeclaredField("log");
    field.setAccessible(true);
    org.slf4j.Logger logger = (org.slf4j.Logger) field.get(null);
    assertEquals("org.openmrs.validator.PatientValidator", logger.getName());
}

File: api/src/main/java/org/openmrs/validator/PatientValidator.java [hunk 1]
```diff
@@ -26,7 +26,7 @@
 @Handler(supports = { Patient.class }, order = 25)
 public class PatientValidator extends PersonValidator {
 
-	private static final Logger log = LoggerFactory.getLogger(PersonNameValidator.class);
+	private static final Logger log = LoggerFactory.getLogger(PatientValidator.class);
```
CodeWithJatinSh, line 29: Good catch! Using PersonNameValidator.class inside PatientValidator means all log output would appear under the wrong class name in production logs, making it very hard to trace issues back to PatientValidator. This one-line fix significantly improves debuggability. LGTM!

CodeWithJatinSh (commented):
Simple but important logger fix  LGTM

chaiwat123431 (commented):
Good catch. Using the wrong class reference in LoggerFactory.getLogger() is a classic copy-paste issue that makes log filtering and debugging misleading in production. The fix is minimal and correct. It might be worth adding a unit test that verifies the logger name matches PatientValidator.class to prevent regression.

PR #5979, opened 2026-03-28
TRUNK-6593: Use try-with-resources in getFileAsBytes

siddhantappuni-ops:
Replaced manual FileInputStream handling with try-with-resources for better resource management and cleaner code.

This prevents potential resource leaks and follows modern Java best practices.

### DIFF
--- a/api/src/main/java/org/openmrs/util/OpenmrsUtil.java
+++ b/api/src/main/java/org/openmrs/util/OpenmrsUtil.java
@@ -238,24 +238,13 @@ public class OpenmrsUtil {
 	 * @throws IOException
 	 */
 	public static byte[] getFileAsBytes(File file) throws IOException {
-		FileInputStream fileInputStream = null;
-		try {
-			fileInputStream = new FileInputStream(file);
+		try (FileInputStream fileInputStream = new FileInputStream(file)) {
 			byte[] b = new byte[fileInputStream.available()];
 			fileInputStream.read(b);
 			return b;
 		} catch (Exception e) {
 			log.error("Unable to get file as byte array", e);
-		} finally {
-			if (fileInputStream != null) {
-				try {
-					fileInputStream.close();
-				} catch (IOException io) {
-					log.warn("Couldn't close fileInputStream: " + io);
-				}
-			}
 		}
-
 		return null;
 	}
 


NitinKumar1-1 (commented):
This is a nice improvement using try-with-resources @siddhantappuni-ops It definitely makes the code cleaner and avoids manual resource handling.

A couple of small points I noticed:

• Use of available() for buffer size
The byte array is initialized using fileInputStream.available(). As far as I know, available() doesn’t always guarantee the full size of the file and can sometimes be unreliable.
Would it be safer to use file.length() or read using a loop instead?

• Exception handling
Currently catching a generic Exception. Maybe we can catch a more specific exception like IOException since the method already throws it, which would make the error handling clearer.

• Return null
Previously there was a return null at the end which is now removed (good 👍). Just confirming — in case of an exception, do we want to propagate it fully instead of returning null? That seems consistent with the method signature.

Overall, this looks like a good refactor and improves readability and safety. Nice work!

siddhantappuni-ops:
Thanks for the feedback! 
1. Good catch on available() you're right that it only estimates bytes available without blocking, not the actual file size.
I looked into using file.length() but it returns long, and casting to int for the byte array size could cause data loss for large files. A cleaner approach might be replacing the manual read entirely with Files.readAllBytes(file.toPath()), which handles sizing internally. Would that work here, or is there a reason we need to stick with FileInputStream?

2. Good point — I'll narrow the catch to IOException instead of generic Exception.

3. Yes, since the method signature already declares throws IOException, letting it propagate gives the caller control over how to handle it. Returning null would silently hide the failure, which is worse for debugging.

chaiwat123431 (commented):
Good refactor. Replacing manual FileInputStream handling with try-with-resources eliminates the risk of resource leaks if an exception is thrown before the close() call. The code reduction from 12 lines to 1 also improves readability significantly. All checks passing and SonarQube reports no new issues.

kamakhyadhruv5 (commented):
Hi @siddhantappuni-ops

This is a good improvement. Using try-with-resources makes the code cleaner and ensures proper resource handling without needing a finally block.

One suggestion: instead of catching a generic `Exception`, it would be better to catch `IOException`. Also, using `fileInputStream.available()` may not always be reliable for reading full content; a buffered approach could be safer.

Overall, nice refactor that improves code quality.

PR #5980, opened 2026-03-28
Trunk 6575: Restrict executeSQL to safe SELECT queries when selectOnly is true

siddhantappuni-ops:
Added validation to the `executeSQL` method in `AdministrationServiceImpl` to restrict execution of unsafe SQL queries when `selectOnly` is set to true.

The changes include:
- Ensuring that only queries starting with `SELECT` are allowed
- Blocking potentially dangerous SQL keywords such as INSERT, UPDATE, DELETE, DROP, ALTER, and TRUNCATE
- Preventing file-based data exfiltration by blocking clauses like INTO OUTFILE, INTO DUMPFILE, and LOAD DATA

This improves security by preventing misuse of the API to execute harmful or non-read-only SQL queries.

---

## Issue I worked on
     https://openmrs.atlassian.net/browse/TRUNK-6575

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
@@ -408,11 +408,35 @@ public class AdministrationServiceImpl extends BaseOpenmrsService implements Adm
 	 */
 	@Override
 	public List<List<Object>> executeSQL(String sql, boolean selectOnly) throws APIException {
-		if (sql == null || "".equals(sql.trim())) {
+		if (sql == null || sql.trim().isEmpty()) {
 			return null;
 		}
 
-		return dao.executeSQL(sql, selectOnly);
+		String trimmedSql = sql.trim();
+		String normalized = trimmedSql.toLowerCase();
+
+		if (selectOnly) {
+			// Must start with SELECT
+			if (!normalized.startsWith("select")) {
+				throw new APIException("Only SELECT queries are allowed");
+			}
+
+			// Block dangerous keywords
+			if (normalized.contains(" insert ") ||
+				normalized.contains(" update ") ||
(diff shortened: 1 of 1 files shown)


yousung1020:
Great initiative on improving the security of the executeSQL method! Restricting dangerous operations when selectOnly is true is a very important step.

However, using simple string matching like .contains(" insert ") might introduce security vulnerabilities and false positives:

1. Bypass with different characters: Attackers could use tabs, newlines, or parenthesis instead of spaces.
2. Bypass with multiple queries: A query like SELECT * FROM users; DROP TABLE data has no space before DROP, so it will completely bypass the check.
3. False Positives: A perfectly safe query containing these words in a literal string will be blocked (e.g., SELECT 'how to insert data' AS instruction).
4. CTE Restrictions: Valid SELECT queries using Common Table Expressions start with WITH, not SELECT, so they will fail the startsWith("select") check.

I recommend using regular expressions with word boundaries to catch keywords regardless of spacing, or ideally, integrating a lightweight SQL parser like JSqlParser to safely analyze the query structure instead of relying on naive string matching.

siddhantappuni-ops:
yeah fair enough, these are good points. honestly i was already a bit unsure about the string matching approach but figured it was better than nothing as a first pass.

do you think this is worth reworking with JSqlParser before merging, or should i just tighten it up with regex for now and we can revisit later? don't want to hold this up if regex is good enough for the time being.

yousung1020:
I think it depends on the project's policy regarding new dependencies.

If adding a lightweight library like JSqlParser is acceptable and does not require a long discussion, it is definitely the most robust long-term solution.

However, if adding a new dependency might block or delay this PR, tightening it up with regex using word boundaries is a perfectly fine approach for now. We can always create a follow-up issue to implement the AST parser later.

Either way, thank you for being so open to feedback!

PR #5981, opened 2026-03-28
TRUNK-6541: Fix UnexpectedRollbackException by disabling transaction in validation

siddhantappuni-ops:
**Description of what I changed**

Updated AdministrationServiceImpl.validate() to run outside of a transactional context by changing the annotation from @Transactional(readOnly = true) to @Transactional(propagation = Propagation.NOT_SUPPORTED).

Previously, validation was executed within a transaction, which could lead to the transaction being marked as rollback-only due to internal validator behavior or Hibernate session interactions. This resulted in an UnexpectedRollbackException instead of properly returning validation errors.

With this change, validation executes without participating in a transaction, ensuring that validation errors are correctly captured in the Errors object and no unintended rollback occurs.

**Issue I worked on**

see https://issues.openmrs.org/browse/TRUNK-6541

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
@@ -72,6 +72,7 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.cache.annotation.CacheEvict;
 import org.springframework.cache.annotation.Cacheable;
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.validation.Errors;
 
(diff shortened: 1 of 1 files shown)


yousung1020:
Great catch on fixing the UnexpectedRollbackException. Changing the propagation to NOT_SUPPORTED correctly suspends the active transaction, which prevents any internal validation errors from marking the global transaction as rollback-only. However, there is one quick thing to verify for safety. Since this validate method now runs completely outside of a transaction context, we should ensure that it does not access any uninitialized lazy-loaded collections of the object. If it does, it might throw a LazyInitializationException instead. If the validation logic only uses already initialized data or simple direct queries, this refactoring looks great.

jwnasambu (commented):
Thanks @siddhantappuni-ops for your work on this PR! Just a friendly reminder that we should only start working on tickets that are in the “Ready to Work” state. Currently, your ticket is still marked as “Needs Assessment.” Could you please leave a comment on the ticket to confirm if these changes are needed and request that its status be updated to “Ready to Work” before continuing with the PR? Thank you for your understanding!

siddhantappuni-ops:
@jwnasambu I don't have the comment access to any ticket. I have raised the request but according to help desk I have got all the permissions required but I cannot comment on any ticket!
Screenshot 2026-04-01 at 1 22 44 PM

shivvani-r:
Appreciate the PR. It turns out the ticket [TRUNK-6541](https://openmrs.atlassian.net/browse/TRUNK-6541) has already been addressed and merged.

[TRUNK-6541]: https://openmrs.atlassian.net/browse/TRUNK-6541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5975, opened 2026-03-27
TRUNK-6623: fix: add null guard to purgePatient consistent with voidPatient and unvoidPatient

manoj-1407:
## What

PatientServiceImpl has three patient lifecycle methods:

- voidPatient → handles null
- unvoidPatient → handles null
- purgePatient → did not handle null

This PR adds a null guard to purgePatient.

## Changes

- Added null check in purgePatient
- Added test for null input case

## Testing

mvn test -pl api -Dtest=PatientServiceTest

All tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Patient deletion now safely ignores null inputs, preventing failures when deletion requests are empty or missing; normal deletion behavior for valid patients is unchanged.

* **Tests**
  * Added test coverage to verify deletion remains safe and is a no-op when given a null input.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openmrs/openmrs-core/pull/5975)

[Review Change Stack](https://app.coderabbit.ai/change-stack/openmrs/openmrs-core/pull/5975)

PR #5982, opened 2026-03-28
TRUNK-6622: fix: add null guard to purgeObs consistent with voidObs

manoj-1407:
## Problem

`ObsServiceImpl.voidObs()` guards against null input (line 104) before 
proceeding. `purgeObs(Obs obs)` does not — it delegates directly to 
`purgeObs(obs, false)` which reaches `dao.deleteObs(obs)` without any 
null check, creating a behavioral inconsistency between sibling methods 
in the same class.

This mirrors the same inconsistency previously fixed in `PatientServiceImpl` 
(PR #5975).

## Fix

Added a null guard to `purgeObs(Obs obs)` consistent with the pattern 
already established by `voidObs`.

## Test

Added `purgeObs_shouldDoNothingWhenObsIsNull` to `ObsServiceTest`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deletion operations for observations and patient records now safely return immediately when given null inputs.

* **Tests**
  * Added unit tests ensuring purge methods for observations and patients are null-safe.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openmrs/openmrs-core/pull/5982)

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
@@ -313,7 +313,10 @@ public class ObsServiceImpl extends BaseOpenmrsService implements ObsService, Re
 	 * @see org.openmrs.api.ObsService#purgeObs(org.openmrs.Obs, boolean)
 	 */
 	@Override
-	public void purgeObs(Obs obs, boolean cascade) throws APIException {
+    public void purgeObs(Obs obs, boolean cascade) throws APIException {
+		if (obs == null) {
+			return;
+		}
 		if (!purgeComplexData(obs)) {
 			// Log a warning instead of throwing an error.
 			// This allows purging the obs row even if the associated file is missing,
--- a/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
@@ -336,6 +336,9 @@ public class ObsServiceImpl extends BaseOpenmrsService implements ObsService, Re
 	 */
 	@Override
 	public void purgeObs(Obs obs) throws APIException {
+		if (obs == null) {
+			return;
+		}
 		Context.getObsService().purgeObs(obs, false);
 	}
 
--- a/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
@@ -313,7 +313,7 @@ public class ObsServiceImpl extends BaseOpenmrsService implements ObsService, Re
 	 * @see org.openmrs.api.ObsService#purgeObs(org.openmrs.Obs, boolean)
 	 */
 	@Override
-    public void purgeObs(Obs obs, boolean cascade) throws APIException {
+	public void purgeObs(Obs obs, boolean cascade) throws APIException {
 		if (obs == null) {
 			return;
 		}
--- a/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
@@ -315,8 +315,9 @@ public class ObsServiceImpl extends BaseOpenmrsService implements ObsService, Re
 	@Override
 	public void purgeObs(Obs obs, boolean cascade) throws APIException {
 		if (obs == null) {
-			return;
+			throw new IllegalArgumentException("obs cannot be null");
 		}
+
 		if (!purgeComplexData(obs)) {
 			// Log a warning instead of throwing an error.
 			// This allows purging the obs row even if the associated file is missing,
@@ -340,7 +341,7 @@ public class ObsServiceImpl extends BaseOpenmrsService implements ObsService, Re
 	@Override
 	public void purgeObs(Obs obs) throws APIException {
 		if (obs == null) {
-			return;
+			throw new IllegalArgumentException("obs cannot be null");
 		}
 		Context.getObsService().purgeObs(obs, false);
 	}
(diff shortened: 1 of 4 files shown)


manoj-1407:
JIRA account access is still broken for new accounts — 
logged in but getting "not authorized to create issues" 
(known issue, also raised in OpenMRS Talk). 
Will add ticket number to title as soon as permissions are granted.

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -336,6 +336,9 @@ public void purgeObs(Obs obs, boolean cascade) throws APIException {
 	 */
 	@Override
 	public void purgeObs(Obs obs) throws APIException {
+		if (obs == null) {
```
siddhantappuni-ops, line 343: Nice fix this makes purgeObs consistent with voidObs. One thought: I notice the other overload purgeObs(Obs obs, boolean cascade) at line 336 doesn't have this same null guard. Should we add it there too for consistency, or is that handled differently?

api/src/test/java/org/openmrs/api/ObsServiceTest.java
```diff
@@ -1464,6 +1464,11 @@ public boolean supportsView(String view) {
 	public void purgeObs_shouldThrowAPIExceptionIfGivenTrueCascade() {
 		assertThrows(APIException.class, () -> Context.getObsService().purgeObs(new Obs(1), true));
 	}
+	public void purgeObs_shouldDoNothingWhenObsIsNull() {
+		// voidObs guards against null input (line 104 of ObsServiceImpl).
+		// purgeObs must be consistent to prevent null reaching the DAO layer.
+		assertDoesNotThrow(() -> Context.getObsService().purgeObs(null));
```
siddhantappuni-ops: Good test. The comments explaining the rationale (lines 1468-1469) are really helpful. Minor thought would it also be worth verifying that no interaction with the DAO layer happens when null is passed, to confirm it truly short-circuits? Though that might be overkill for this scope.

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
manoj-1407: Good catch. The single-argument purgeObs delegates to the cascade 
overload — so the null guard at the entry point prevents null from 
ever reaching purgeObs(obs, false). Adding a guard to the cascade 
overload as well would be defensive but redundant given this delegation 
chain. Happy to add it there too if that's the preferred pattern.

api/src/test/java/org/openmrs/api/ObsServiceTest.java
manoj-1407: Fair point. The current test verifies no exception is thrown, which 
confirms the null guard works. Adding a Mockito spy to verify 
dao.deleteObs is never called would make it more explicit — I can 
add that if it's preferred, though it would require mocking the 
context which adds some test complexity.

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
manoj-1407: Added the null guard to the cascade overload in the latest commit. Both purgeObs(Obs) and purgeObs(Obs, boolean) now guard against null at the entry point. Thanks for catching it.

api/src/test/java/org/openmrs/api/ObsServiceTest.java
manoj-1407: Agreed it would add test complexity for limited gain given the guard is already verified by the no-exception assertion. Happy to add it if you'd like but won't push it unless you think it's worth it.

VarshithReddy2006 (commented):
Should we ensure consistency with other purge methods by centralizing null checks in a common utility method?

manoj-1407:
A shared utility for a two-line null check would add indirection without real benefit — each service method is independently readable as-is, and the existing OpenMRS codebase follows the pattern of inline guards rather than centralizing them. The consistency here is in the pattern applied, not in a shared abstraction.

manoj-1407:
Rebased on trunk, CI passing — ready for review

dkayiwa:
@coderabbitai fullreview

manoj-1407:
@dkayiwa the build check has been stuck on "Waiting for status to be reported" — could you help re-trigger CI?

manoj-1407:
Hi @dkayiwa — just checking in on this PR as well. Happy to address any feedback if needed

dkayiwa:
@claude review

manoj-1407:
@claude review

manoj-1407:
hi @dkayiwa  sorry for the back and forth. updated in latest commit — addressed all formatting and logic issues flagged in review. the remaining CI failures (OWASP CVE fetch, Java 25 build) appear to be infrastructure-level and affect other PRs in the repo as well

dkayiwa:
@claude review

jwnasambu (commented):
Thanks for the progress @manoj-1407. Kindly run `mvn spotless:apply` to fix the error you are facing.

manoj-1407:
thanks @jwnasambu, ran mvn spotless:apply and pushed the formatting fixes.

api/src/main/java/org/openmrs/api/impl/ObsServiceImpl.java
```diff
@@ -314,6 +314,9 @@ public Obs unvoidObs(Obs obs) throws APIException {
 	 */
 	@Override
 	public void purgeObs(Obs obs, boolean cascade) throws APIException {
+		if (obs == null) {
+			return;
+		}
 		if (!purgeComplexData(obs)) {
```
ibacher, line 321:
```suggestion
		}
		
		if (!purgeComplexData(obs)) {
```

PR #5984, opened 2026-03-28
TRUNK-6621: fix: add null guard to purgeVisit to prevent NullPointerException on null input

manoj-1407:
JIRA: https://openmrs.atlassian.net/browse/TRUNK-6621

## Problem

`VisitServiceImpl.purgeVisit(Visit visit)` called `visit.getVisitId()` before checking if `visit` itself was null. Passing null threw a `NullPointerException` before reaching the existing unsaved-visit guard.

Beyond the NPE, all three sibling purge methods (`purgeVisit`, `purgePatient`, `purgeObs`) had inconsistent null contracts — two silently returned on null while this one crashed. Purge is irreversible, so silently accepting null is a worse failure mode than rejecting it loudly.

## Fix

- Added a null guard at the entry point of `purgeVisit` in `VisitServiceImpl` — throws `IllegalArgumentException` before touching the visit object
- Updated `purgePatient` (PR #5975) and `purgeObs` (PR #5982) to throw `IllegalArgumentException` as well, making all three consistent
- Removed the now-redundant null check in `HibernateVisitDAO.deleteVisit` — the service-layer throw already prevents null from reaching the DAO, so that check was dead code with no reachable test path (which was causing the Codecov 0% flag)

## Testing
mvn test -pl api -Dtest=VisitServiceTest

All tests pass.

Added `purgeVisit_shouldThrowIllegalArgumentExceptionWhenVisitIsNull` to `VisitServiceTest`.

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateVisitDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateVisitDAO.java
@@ -171,14 +171,15 @@ public class HibernateVisitDAO implements VisitDAO {
 	/**
 	 * @see org.openmrs.api.db.VisitDAO#deleteVisit(org.openmrs.Visit)
 	 */
-@Override
-@Transactional
-public void deleteVisit(Visit visit) throws DAOException {
-    if (visit == null) {
-        throw new DAOException("Visit must not be null");
-    }
-    getCurrentSession().remove(visit);
-}
+	@Override
+	@Transactional
+	public void deleteVisit(Visit visit) throws DAOException {
+		if (visit == null) {
+			throw new DAOException("Visit must not be null");
+		}
+		getCurrentSession().remove(visit);
+	}
+
 	/**
 	 * @see org.openmrs.api.db.VisitDAO#getVisits(java.util.Collection, java.util.Collection,
 	 *      java.util.Collection, java.util.Collection, java.util.Date, java.util.Date, java.util.Date,
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateVisitDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateVisitDAO.java
@@ -171,12 +171,14 @@ public class HibernateVisitDAO implements VisitDAO {
 	/**
 	 * @see org.openmrs.api.db.VisitDAO#deleteVisit(org.openmrs.Visit)
 	 */
-	@Override
-	@Transactional
-	public void deleteVisit(Visit visit) throws DAOException {
-		getCurrentSession().remove(visit);
-	}
-
+@Override
+@Transactional
+public void deleteVisit(Visit visit) throws DAOException {
+    if (visit == null) {
+        throw new DAOException("Visit must not be null");
+    }
+    getCurrentSession().remove(visit);
+}
 	/**
 	 * @see org.openmrs.api.db.VisitDAO#getVisits(java.util.Collection, java.util.Collection,
 	 *      java.util.Collection, java.util.Collection, java.util.Date, java.util.Date, java.util.Date,
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateVisitDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateVisitDAO.java
@@ -174,9 +174,6 @@ public class HibernateVisitDAO implements VisitDAO {
 	@Override
 	@Transactional
 	public void deleteVisit(Visit visit) throws DAOException {
-		if (visit == null) {
-			throw new DAOException("Visit must not be null");
-		}
 		getCurrentSession().remove(visit);
 	}
 
--- a/api/src/main/java/org/openmrs/api/impl/VisitServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/VisitServiceImpl.java
@@ -222,7 +222,7 @@ public class VisitServiceImpl extends BaseOpenmrsService implements VisitService
 	@Override
 	public void purgeVisit(Visit visit) throws APIException {
     if (visit == null) {
-        return;
+        throw new IllegalArgumentException("visit must not be null");
     }
     if (visit.getVisitId() == null) {
 			return;
--- a/api/src/main/java/org/openmrs/api/impl/VisitServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/VisitServiceImpl.java
@@ -221,7 +221,10 @@ public class VisitServiceImpl extends BaseOpenmrsService implements VisitService
 	 */
 	@Override
 	public void purgeVisit(Visit visit) throws APIException {
-		if (visit.getVisitId() == null) {
+    if (visit == null) {
+        return;
+    }
+    if (visit.getVisitId() == null) {
 			return;
 		}
 		if (!Context.getEncounterService().getEncountersByVisit(visit, true).isEmpty()) {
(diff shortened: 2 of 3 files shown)


IvanescuChristian:
Salute! While the enthusiasm is highly appreciated, adding the null guard is not actually of help or proper coding practice.
When we create functions for use it is not always our job to prevent eronated input. There is no reason for this function to be called with null and if it is the exception thrown is what we want, similar to a !panic in rust.
Other parts of the OpenMRS API throw explicitly in similar cases (e.g. PatientServiceImpl.savePatientIdentifier throws APIException on null input) rather than silently returning.
But if you truly want to make a PR u could consider the following proposition: Would it make more sense to throw an IllegalArgumentException("visit must not be null") instead of silently returning?

manoj-1407:
Good point — updated to throw IllegalArgumentException("visit must not be null") instead of silently returning, consistent with the explicit failure pattern used elsewhere in the API. Test updated accordingly.

api/src/main/java/org/openmrs/api/impl/VisitServiceImpl.java
```diff
@@ -221,7 +221,10 @@ public Visit unvoidVisit(Visit visit) throws APIException {
 	 */
 	@Override
 	public void purgeVisit(Visit visit) throws APIException {
-		if (visit.getVisitId() == null) {
+    if (visit == null) {
+        throw new IllegalArgumentException("visit must not be null");
+    }
```
VarshithReddy2006: Should we also log a warning when a null input is received to help with debugging unexpected cases?

manoj-1407: Since we're throwing IllegalArgumentException, the exception itself signals the problem to the caller. A log warning before a throw would be redundant — the caller receives the exception and handles or logs it as needed. If this were a silent no-op I'd agree a warning makes sense, but with an explicit throw the intent is already clear.

CodeWithJatinSh: That's a fair point. Since IllegalArgumentException is unchecked, it makes sense not to double-log. One thing worth noting: the new unit test on line 845 uses assertThrows(IllegalArgumentException.class) which is good but the comment on line 843-844 mentions the old NPE behavior. Since the behavior is now changed to IllegalArgumentException, those inline comments could be updated to reflect the new contract rather than describing the old bug.

CodeWithJatinSh (commented):
Reviewed the null guard fix and unit test. The inline comments on lines 843-844 still reference the old NPE behavior but the code now throws IllegalArgumentException worth updating those comments to reflect the new contract.

chaiwat123431 (commented):
Good fix! The null guard is consistent with the pattern used in purgePatient (#5975) and purgeObs (#5982). One thing worth noting: the inline comments on lines 843-844 still reference the old NPE behavior, but since the code now throws IllegalArgumentException, those comments should be updated to reflect the new contract. Otherwise the logic is clean and the test covers the null case well.

manoj-1407:
Updated the inline comments in the latest commit to reflect the IllegalArgumentException contract rather than the old NPE description. Thanks for catching that.

shivvani-r:
Kindly refer to the [Pull Request Tips](https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477199/Pull+Request+Tips) and make the required changes.

manoj-1407:
Thanks for the review and engagement on this!

tendomart (commented):
Add  issue name to your title, also include a link to your issue to make navigation easy for reviewers

manoj-1407:
Rebased on trunk, CI passing — ready for review

dkayiwa:
@coderabbitai fullreview

manoj-1407:
@dkayiwa the build check has been stuck on "Waiting for status to be reported" — could you help re-trigger CI?

manoj-1407:
Hi @dkayiwa — just checking in on this PR. Happy to address any feedback if needed

dkayiwa:
@claude review

manoj-1407:
@claude review

manoj-1407:
hi @dkayiwa  sorry for the back and forth. updated in latest commit — addressed all formatting and logic issues flagged in review. the remaining CI failures (OWASP CVE fetch, Java 25 build) appear to be infrastructure-level and affect other PRs in the repo as well

dkayiwa:
@claude review

manoj-1407:
Hi @dkayiwa — these three PRs (https://github.com/openmrs/openmrs-core/pull/5982, https://github.com/openmrs/openmrs-core/pull/5975, https://github.com/openmrs/openmrs-core/pull/5984) add null guards consistent with existing void methods. All CI passing. Would appreciate a review when you have a moment.

api/src/main/java/org/openmrs/api/impl/VisitServiceImpl.java
```diff
@@ -221,6 +221,9 @@ public Visit unvoidVisit(Visit visit) throws APIException {
 	 */
 	@Override
 	public void purgeVisit(Visit visit) throws APIException {
+		if (visit == null) {
+			throw new IllegalArgumentException("visit cannot be null");
+		}
```
dkayiwa, line 226: Is the divergence from the other two PRs deliberate? The description says this follows purgePatient (#5975) and purgeObs (#5982), but both of those return silently on null while this one throws, and TRUNK-6621 asks for an early exit as well. All three are up for review together, so as it stands core would get two different null contracts across three sibling purge methods.

I'd keep the throw here and change the other two to match, rather than the other way round. Purge is irreversible, so quietly treating a null argument as success is worse than an exception. The unsaved-visit return just below is a different case: there the postcondition (this visit is not in the database) already holds, so there genuinely is nothing to do.

Whichever way it goes, the description needs a pass; it still describes a `purgeVisit_shouldDoNothingWhenVisitIsNull` test that isn't in the diff.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateVisitDAO.java
```diff
@@ -171,12 +171,14 @@ public Visit saveVisit(Visit visit) throws DAOException {
...
-	@Override
-	@Transactional
-	public void deleteVisit(Visit visit) throws DAOException {
-		getCurrentSession().remove(visit);
-	}
-
+@Override
+@Transactional
+public void deleteVisit(Visit visit) throws DAOException {
+    if (visit == null) {
```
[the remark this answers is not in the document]
dkayiwa: This is still worth dropping. `deleteVisit` has exactly one caller in this repo, `VisitServiceImpl.purgeVisit`, and that method now throws two lines before it reaches `dao.deleteVisit(visit)`, so a null visit cannot get here. I checked by deleting the service guard locally and rerunning `purgeVisit_shouldThrowIllegalArgumentExceptionWhenVisitIsNull`: it fails with `NullPointerException: Cannot invoke "org.openmrs.Visit.getVisitId()" because "visit" is null` at VisitServiceImpl:224, so even with the service guard gone the call never gets as far as the DAO. That also means there is no test that can cover these three lines. I'd delete them and keep the guard in the service, where the contract lives and where the test can reach it.

api/src/main/java/org/openmrs/api/impl/VisitServiceImpl.java
manoj-1407: Not deliberate — updated all three to throw. purgePatient (#5975) and purgeObs (#5982) now throw IllegalArgumentException on null instead of silently returning, matching purgeVisit. Also removed the redundant null check in HibernateVisitDAO.deleteVisit — the service-layer throw already prevents null reaching the DAO, so that check was dead code (which was the Codecov 0% flag). Description updated — stale test name removed

manoj-1407:
hey @dkayiwa — updated all three PRs (#5975, #5982, #5984) to throw 
IllegalArgumentException on null, consistent with purgeVisit. 
Also updated the PR descriptions to reflect the actual behavior. 
Could you take another look when you get a chance?

PR #5986, opened 2026-03-29
TRUNK-6599: Refactor containsAllergen() using Stream API

VarshithReddy2006:
### Jira Ticket

https://openmrs.atlassian.net/browse/TRUNK-6599

### Description

The current implementation of `containsAllergen()` in `Allergies.java` uses a manual loop to check if a given allergy exists in a collection.

This PR refactors the method to use the Java Stream API, improving readability and reducing boilerplate code while preserving the existing behavior.

### Changes

* Replaced manual loop with `stream().anyMatch()` for a more declarative and concise implementation
* Added a unit test to document the current behavior when the allergies collection is `null` (throws `NullPointerException`)

### Behavior

No functional changes. The method retains its original behavior, including throwing a `NullPointerException` when the allergies collection is `null`.

### Verification

* All existing tests pass
* Added test ensures current null-handling behavior is explicitly documented

### DIFF
--- a/api/src/main/java/org/openmrs/Allergies.java
+++ b/api/src/main/java/org/openmrs/Allergies.java
@@ -327,12 +327,7 @@ public class Allergies implements List<Allergy> {
 	 * @return true if the same allergen exists, else false
 	 */
 	public boolean containsAllergen(Allergy allergy, Collection<? extends Allergy> allergies) {
-		for (Allergy alg : allergies) {
-			if (alg.hasSameAllergen(allergy)) {
-				return true;
-			}
-		}
-		return false;
+		return allergies.stream().anyMatch(a -> a.hasSameAllergen(allergy));
 	}
 
 	/**
--- a/api/src/test/java/org/openmrs/AllergiesTest.java
+++ b/api/src/test/java/org/openmrs/AllergiesTest.java
@@ -512,4 +512,16 @@ public class AllergiesTest extends BaseContextSensitiveTest {
 
 		assertThrows(APIException.class, () -> new Allergies().addAll(0, allergies));
 	}
+
+	/**
+	 * @see {@link Allergies#containsAllergen(Allergy, java.util.Collection)}
+	 */
+	@Test
+	public void containsAllergen_shouldThrowExceptionWhenAllergiesIsNull() {
+		Allergy allergy = new Allergy();
+
+		assertThrows(NullPointerException.class, () -> {
+			allergies.containsAllergen(allergy, null);
+		});
+	}
 }


VarshithReddy2006:
Hi, this PR addresses TRUNK-6599.

Please let me know if any improvements are needed. Thanks!

api/src/main/java/org/openmrs/Allergies.java
```diff
@@ -327,12 +327,7 @@ private void throwExceptionIfHasAllergen(Allergy allergy, Collection<? extends A
 	 * @return true if the same allergen exists, else false
 	 */
 	public boolean containsAllergen(Allergy allergy, Collection<? extends Allergy> allergies) {
-		for (Allergy alg : allergies) {
-			if (alg.hasSameAllergen(allergy)) {
-				return true;
-			}
-		}
-		return false;
+		return allergies.stream().anyMatch(a -> a.hasSameAllergen(allergy));
```
yousung1020, line 330: This is a nice and clean refactoring! Using the Stream API makes the code much more readable and declarative while perfectly maintaining the short-circuiting behavior with anyMatch().

One quick suggestion for robustness: if there is any chance that the allergies collection could be passed as null, calling allergies.stream() will immediately throw a NullPointerException.

Since we are refactoring this method, it might be a great opportunity to add a quick safety check before opening the stream, such as:

```suggestion
        if (allergies == null || allergies.isEmpty()) {
            return false;
        }
	return allergies.stream().anyMatch(a -> a.hasSameAllergen(allergy));
```

VarshithReddy2006: Thanks for the review and suggestion!

That's a good point about null safety. The original implementation would also throw a NullPointerException if the allergies collection was null.

To preserve existing behavior, I kept the implementation unchanged. However, I’m happy to add a null check if making the method more defensive is preferred.

Please let me know your preference.

CodeWithJatinSh: I agree with preserving existing behavior as the primary goal. One middle ground worth considering: since this is already a refactoring PR, adding a unit test that explicitly documents the current behavior when allergies is null (i.e., that NPE is thrown) would make the contract clear to future developers. That way, if someone later wants to add defensive null handling, the test would guide them.

CodeWithJatinSh (commented):
Left a suggestion on documenting the null contract via a unit test, so the behavior is explicit for future developers.

api/src/main/java/org/openmrs/Allergies.java
VarshithReddy2006: Thanks for the suggestion! I’ve added a unit test to explicitly document the current behavior when the allergies collection is null (i.e., a NullPointerException is thrown). This should help clarify the contract for future changes.

0abhi007:
Stream refactor looks clean and all checks pass. One suggestion: rather than documenting the `NullPointerException` behavior via a test, consider adding an explicit null guard at the top of `containsAllergen()`. This would make the method safer for callers who might not expect a silent NPE.

VarshithReddy2006:
> Stream refactor looks clean and all checks pass. One suggestion: rather than documenting the `NullPointerException` behavior via a test, consider adding an explicit null guard at the top of `containsAllergen()`. This would make the method safer for callers who might not expect a silent NPE.

Thanks for the suggestion!

I considered adding a null guard, but I intentionally preserved the existing behavior since this PR is focused on refactoring only.

The previous implementation also threw a NullPointerException when the allergies collection is null, so I kept that behavior unchanged and added a test to document it.

If needed, I’d be happy to open a separate PR to improve null-safety.

VarshithReddy2006:
Hi, I’ve updated the branch with the latest master.

I’ve kept the refactor limited to improving readability while preserving the existing behavior, including the current null handling. 

Happy to update the implementation if introducing a null guard is preferred, or handle that in a separate PR.

Please let me know your thoughts. Thanks!

PR #5989, opened 2026-03-30
Add null guard to purgeEncounter to prevent NullPointerException on null input

chaiwat123431:
## Problem
purgeEncounter(Encounter encounter) calls encounter methods before checking if encounter itself is null. Passing null throws a NullPointerException.

This follows the same pattern fixed in purgePatient (PR #5975), purgeObs (PR #5982), and purgeVisit (PR #5984).

## Fix
Added a null guard at the entry point of purgeEncounter, consistent with the pattern established across other purge methods in the codebase.

## Test
Added purgeEncounter_shouldDoNothingWhenEncounterIsNull to EncounterServiceTest.

## Testing
All tests pass.

PR #5990, opened 2026-03-30
Add null guard to purgeCondition to prevent NullPointerException on null input

chaiwat123431:
## Problem
purgeCondition(Condition condition) calls conditionDAO.deleteCondition(condition) without checking if condition is null first. Passing null can cause unexpected behavior.

This follows the same pattern fixed in purgePatient (PR #5975), purgeObs (PR #5982), purgeVisit (PR #5984), and purgeEncounter (PR #5989).

## Fix
Added a null guard at the entry point of purgeCondition, consistent with the pattern established across other purge methods in the codebase.

## Test
Added ConditionServiceTest with purgeCondition_shouldDoNothingWhenConditionIsNull test.

## Testing
All tests pass.

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
@@ -182,6 +182,9 @@ public class ConditionServiceImpl extends BaseOpenmrsService implements Conditio
 	 */
 	@Override
 	public void purgeCondition(Condition condition) {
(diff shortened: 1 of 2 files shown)


fierce-adventurer:
chore: nice contribution checking for null before deleting it from the database and can also simulate deleting junit for simulate condition being deleted successfully.

fierce-adventurer (commented):
this is a nice and clean code just apply jira ticket number in the pull request title.

PR #5992, opened 2026-03-30
fix: add null guard to purgePatient consistent with voidPatient

CodeWithJatinSh:
This PR adds a null guard to purgePatient() in PatientServiceImpl.java. Previously, passing null would cause a NullPointerException. The fix throws IllegalArgumentException with a clear message, consistent with how voidPatient() and unvoidPatient() handle null input.

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -355,9 +355,11 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 	 */
 	@Override
 	public void purgePatient(Patient patient) throws APIException {
+
     if (patient == null) {
         throw new IllegalArgumentException("patient must not be null");
       }
+		
     dao.deletePatient(patient);
     } 
 
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -355,8 +355,11 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 	 */
 	@Override
 	public void purgePatient(Patient patient) throws APIException {
-		dao.deletePatient(patient);
-	}
+    if (patient == null) {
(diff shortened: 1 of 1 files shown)


fierce-adventurer (changes requested):
chore: great observation now it won't fail silently great contribution.

Add Junit testing simulating similar conditions and describe about the test case in the PR

Also add Jira ticket to the title of the pull request

CodeWithJatinSh:
Thank you for the feedback! I tried creating a JIRA ticket but the OpenMRS ID login is currently having issues. I'll add the ticket number as soon as it's resolved. Regarding the unit test — I'll add assertThrows(IllegalArgumentException.class, () -> patientService.purgePatient(null)) to the existing PatientServiceTest to cover the null input scenario. Will push the update shortly!

Kanishka680:
Simple and effective fix. Adding null guards to maintenance methods like purgePatient is a great safety measure.

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -355,8 +355,13 @@ public Patient unvoidPatient(Patient patient) throws APIException {
 	 */
 	@Override
 	public void purgePatient(Patient patient) throws APIException {
-		dao.deletePatient(patient);
-	}
+
+    if (patient == null) {
+        throw new IllegalArgumentException("patient must not be null");
```
tendomart, line 360: Have you added a Unit test for this ?

Also is it attached to any particular JIRA Ticket?

CodeWithJatinSh: Yes! I've added a unit test purgePatient_shouldThrowIllegalArgumentExceptionWhenPatientIsNull() in PatientServiceTest.java it's included in this PR (2 commits now). Regarding the JIRA ticket  the OpenMRS ID login is currently having technical issues on my end, but I'll add the ticket number to the PR title as soon as it's resolved. Thanks for the review!

omeriinnocent:
@CodeWithJatinSh have you reached https://help.openmrs.org (Jira service management) regarding your issues

PR #6004, opened 2026-03-31
test: add null guard test for purgePatient with null input

CodeWithJatinSh:
Added a unit test to verify that purgePatient() throws IllegalArgumentException when null is passed. This test documents and protects the null guard behavior added in the related fix PR #5992.


Kanishka680 (commented):
Simple and effective fix. Adding null guards to maintenance methods like purgePatient is a great safety measure.

stefanachiorean:
Good addition of a null guard test. It would be helpful to also consider adding a test case that verifies the behavior when a partially null input is provided, to ensure edge cases are covered.

PR #6006, opened 2026-03-31
TRUNK-6605: Implement globalPropertyDeleted in AdministrationServiceImpl

ShrutiSocrates:
Implements globalPropertyDeleted() to reset presentationLocales when LOCALE_ALLOWED_LIST 
property is deleted, consistent with globalPropertyChanged() behavior. Added unit test to verify the behavior.

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-6605

- [x] My IDE is configured to follow the code style
- [x] I have added tests to cover my changes
- [x] I ran mvn clean package
- [x] All new and existing tests passed
- [x] My pull request is based on latest master

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
@@ -697,8 +697,10 @@ public class AdministrationServiceImpl extends BaseOpenmrsService implements Adm
 	 */
 	@Override
 	public void globalPropertyDeleted(String propertyName) {
-		// TODO Auto-generated method stub
-
+		if (propertyName.equals(
+			OpenmrsConstants.GLOBAL_PROPERTY_LOCALE_ALLOWED_LIST)) {
+			presentationLocales = null;
+		}
 	}
 
 	/**
--- a/api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
@@ -1180,4 +1180,28 @@ public class AdministrationServiceTest extends BaseContextSensitiveTest {
 		// verify hook methods must be called
 		verify(activator).setupOnVersionChange(previousCoreVersion, previousModuleVersion);
 	}
+
+	@Test
+	public void globalPropertyDeleted_shouldResetPresentationLocalesWhenLocaleAllowedListDeleted() {
+		// First get presentation locales to initialize them
+		AdministrationService as = Context.getAdministrationService();
+
+		// Set allowed locales so presentationLocales gets initialized
+		GlobalProperty gp = new GlobalProperty(
+			OpenmrsConstants.GLOBAL_PROPERTY_LOCALE_ALLOWED_LIST, "en_GB");
+		as.saveGlobalProperty(gp);
+
+		// Initialize presentationLocales by calling getPresentationLocales
+		as.getPresentationLocales();
+
+		// Now delete the global property
+		((GlobalPropertyListener) as).globalPropertyDeleted(
+			OpenmrsConstants.GLOBAL_PROPERTY_LOCALE_ALLOWED_LIST);
+
+		// presentationLocales should be reset to null
+		// Calling getPresentationLocales again should 
+		// reinitialize it from scratch
+		assertNotNull(as.getPresentationLocales());
+	}
+
 }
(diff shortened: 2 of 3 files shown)


0abhi007:
Good implementation! SonarCloud flagged 2 new issues on this PR, worth looking into before merge to keep the quality gate fully clean. Also, consider adding a test for when the deleted property key is NOT `LOCALE_ALLOWED_LIST`, to confirm there are no unintended side effects on other properties.

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
```diff
@@ -697,8 +697,10 @@ public void globalPropertyChanged(GlobalProperty newValue) {
 	 */
 	@Override
 	public void globalPropertyDeleted(String propertyName) {
-		// TODO Auto-generated method stub
-
```
jwnasambu: Kindly why did you remove line?

jwnasambu (commented):
Kindly is your ticket Id supposed to read something like this `TRUNK-6605: Implement globalPropertyDeleted in AdministrationServiceImpl` and not `TRUNK-6605: Implement globalPropertyDeleted to reset presentationLocales`

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
ShrutiSocrates: Those lines were auto-generated placeholder stubs with no actual implementation. Since I've now provided the real implementation for globalPropertyDeleted, the TODO stub was no longer needed and I removed it to keep the code clean.

ShrutiSocrates:
> Kindly is your ticket Id supposed to read something like this `TRUNK-6605: Implement globalPropertyDeleted in AdministrationServiceImpl` and not `TRUNK-6605: Implement globalPropertyDeleted to reset presentationLocales`

You're right, thank you for pointing that out! The commit message should reflect the broader implementation scope. I'll update it to TRUNK-6605: Implement globalPropertyDeleted in AdministrationServiceImpl.

ShrutiSocrates:
> Good implementation! SonarCloud flagged 2 new issues on this PR, worth looking into before merge to keep the quality gate fully clean. Also, consider adding a test for when the deleted property key is NOT `LOCALE_ALLOWED_LIST`, to confirm there are no unintended side effects on other properties.

Thanks for the feedback! I'll look into the 2 SonarCloud issues and address them before merge. I'll also add a test case to verify that deleting a property key other than LOCALE_ALLOWED_LIST doesn't cause any unintended side effects on other properties.

jwnasambu (commented):
Kindly these changes are unrelated to the ticket’s scope and would be better reviewed and tracked separately.

ShrutiSocrates:
> Kindly these changes are unrelated to the ticket’s scope and would be better reviewed and tracked separately.

Thank you for the feedback! You're right, the HibernateFormDAOTest changes were unintentionally included due to a merge 
conflict. I will clean up the PR to only include changes related to TRUNK-6605.

ShrutiSocrates:
> Kindly these changes are unrelated to the ticket’s scope and would be better reviewed and tracked separately.

 @jwnasambu I have removed the unrelated HibernateFormDAOTest changes from this PR. The PR now only contains changes related to TRUNK-6605.

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
jwnasambu: You are right. Thanks for correcting me I really appreciate it.

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
```diff
@@ -697,8 +697,10 @@ public void globalPropertyChanged(GlobalProperty newValue) {
...
+		if (propertyName.equals(
+			OpenmrsConstants.GLOBAL_PROPERTY_LOCALE_ALLOWED_LIST)) {
```
jwnasambu, line 701: If propertyName is null, propertyName.equals(...) will crash.
```suggestion
			if (OpenmrsConstants.GLOBAL_PROPERTY_LOCALE_ALLOWED_LIST.equals(propertyName)) {
```

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -1180,4 +1180,28 @@ public void runModuleSetupOnVersionChange_shouldExecuteLiquibaseAndStoreNewVersi
...
+		as.getPresentationLocales();
+
+		// Now delete the global property
+		((GlobalPropertyListener) as).globalPropertyDeleted(
+			OpenmrsConstants.GLOBAL_PROPERTY_LOCALE_ALLOWED_LIST);
+
+		// presentationLocales should be reset to null
+		// Calling getPresentationLocales again should 
+		// reinitialize it from scratch
+		assertNotNull(as.getPresentationLocales());
```
jwnasambu, line 1204: These changes do not actually verify that presentationLocales was reset by globalPropertyDeleted(). Do you mind replacing  the not null assertion with a check that the cached set was actually rebuilt after deletion.

PR #6028, opened 2026-04-09
TRUNK-6612 Add missing future end date validation to PersonAddressValidator

Kapil-kundu:
Currently, `PersonAddressValidator` does not reject cases where `endDate` is set in the future, creating a validation gap despite having an existing message key. This change adds validation to disallow future `endDate` values and includes a unit test to verify the behavior, ensuring consistency with existing date validation rules.

**Jira ticket**: https://openmrs.atlassian.net/browse/TRUNK-6612

### DIFF
--- a/api/src/main/java/org/openmrs/validator/PersonAddressValidator.java
+++ b/api/src/main/java/org/openmrs/validator/PersonAddressValidator.java
@@ -89,9 +89,17 @@ public class PersonAddressValidator implements Validator {
 			    "The Start Date for address '" + addressString + "' shouldn't be in the future");
 		}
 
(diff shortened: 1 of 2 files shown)


Kapil-kundu:
Hey @jwnasambu, Is there any improvement i can do in this PR?

omeriinnocent (commented):
welldone @Kapil-kundu ,This change makes PersonAddressValidator reject any future endDate (see the new check around line ~92). Do we know whether any existing OpenMRS implementations or integrations use endDate to represent a planned/expected move-out date (i.e., a future “valid until” date), and if so, should we make this rule configurable (e.g., via a global property to allow future end dates) to avoid a backwards-incompatible core validator change?

Kapil-kundu (commented):
I have made the future endDate validation configurable via a global property to avoid breaking existing implementations.

PR #6037, opened 2026-04-13
TRUNK-6510: Add priority field to ConceptReferenceRange for range selection

Binayak490-cyber:

This PR adds priority-based selection to the concept reference range functionality:

* Added an `Integer priority` field to `ConceptReferenceRange` (nullable, with getter/setter). A `null` priority preserves full backward compatibility with the existing "strictest bounds" merging behaviour.
* Updated `ConceptServiceImpl.selectReferenceRange()` to implement the agreed selection logic (higher number = higher precedence, descending order):

  * If all matching ranges have `priority = null` → legacy "strictest bounds" merging is used (backward compatible).
  * If any matching range has a priority set → the range(s) with the highest priority value are selected; null-priority ranges are ignored in this case.
  * If multiple ranges share the same highest priority → strictest bounds are applied among them as a tie-breaker.
* Added a Liquibase migration (`liquibase-update-to-latest-3.0.x.xml`) to add a nullable `INT priority` column to the `concept_reference_range` table.
* Added 3 new tests covering: highest priority wins, prioritized range beats null-priority range, and strictest-among-tied behaviour.

One thing is to be **Noted on the mixed-priority edge case:** The ticket marks as TBD what should happen when some matching ranges have a priority and some don't. In this PR I treated null-priority ranges as lower than any explicitly set priority, which aligns with the spirit of the design discussed in the ticket.

## Issue I worked on

https://openmrs.atlassian.net/browse/TRUNK-6510

* [x] My IDE is configured to follow the code style of this project
* [x] I have added tests to cover my changes
* [x] I ran `mvn clean package` right before creating this pull request and added all formatting changes to my commit
* [x] All new and existing tests passed
* [x] My pull request is based on the latest changes of the master branch

### DIFF
--- a/api/src/main/java/org/openmrs/ConceptReferenceRange.java
+++ b/api/src/main/java/org/openmrs/ConceptReferenceRange.java
@@ -54,7 +54,7 @@ public class ConceptReferenceRange extends BaseReferenceRange implements Openmrs
 	 * priority (descending order) is selected. A null priority falls back to the legacy "strictest
 	 * bounds" merging behaviour for backward compatibility.
 	 *
-	 * @since 2.9.0
+	 * @since 3.0.0
 	 */
 	@Column(name = "priority")
 	private Integer priority;
--- a/api/src/main/java/org/openmrs/ConceptReferenceRange.java
+++ b/api/src/main/java/org/openmrs/ConceptReferenceRange.java
@@ -49,6 +49,16 @@ public class ConceptReferenceRange extends BaseReferenceRange implements Openmrs
 	@Column(name = "criteria", length = 65535)
 	private String criteria;
 
+	/**
+	 * Priority for this reference range. When multiple ranges match a patient, the one with the highest
+	 * priority (descending order) is selected. A null priority falls back to the legacy "strictest
+	 * bounds" merging behaviour for backward compatibility.
+	 *
+	 * @since 2.9.0
+	 */
+	@Column(name = "priority")
+	private Integer priority;
+
 	@ManyToOne(fetch = FetchType.LAZY)
 	@JoinColumn(name = "concept_id", nullable = false)
 	private ConceptNumeric conceptNumeric;
@@ -90,6 +100,25 @@ public class ConceptReferenceRange extends BaseReferenceRange implements Openmrs
 	 */
 	public void setCriteria(String criteria) {
 		this.criteria = criteria;
+	}
+
+	/**
+	 * Gets the priority of this conceptReferenceRange. Higher values take precedence when multiple
+	 * ranges match a patient. A null priority uses legacy "strictest bounds" logic.
+	 *
+	 * @return the priority, or null if not set
+	 */
+	public Integer getPriority() {
+		return priority;
+	}
+
+	/**
+	 * Sets the priority of this conceptReferenceRange.
+	 *
+	 * @param priority the priority to set
+	 */
+	public void setPriority(Integer priority) {
+		this.priority = priority;
 	}
 
 	/**
--- a/api/src/main/java/org/openmrs/ConceptReferenceRange.java
+++ b/api/src/main/java/org/openmrs/ConceptReferenceRange.java
@@ -54,7 +54,7 @@ public class ConceptReferenceRange extends BaseReferenceRange implements Openmrs
 	 * priority (descending order) is selected. A null priority falls back to the legacy "strictest
 	 * bounds" merging behaviour for backward compatibility.
 	 *
-	 * @since 3.0.0
+	 * @since 3.0.0, 2.9.0
 	 */
 	@Column(name = "priority")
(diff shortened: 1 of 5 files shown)


Binayak490-cyber:
Hi @ibacher, @mseaton, @dkayiwa, @mogoodrich - raised this PR for TRUNK-6510. Adds a nullable priority field to ConceptReferenceRange so the highest-priority range wins when multiple ranges match a patient, with full backward compatibility when priority is null. All tests pass. Please review as it now needs your feedback over if any changes needs to be done over the fix!

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -2139,7 +2140,36 @@ public ConceptReferenceRange getConceptReferenceRange(ConceptReferenceRangeConte
...
+
+		if (prioritizedRanges.isEmpty()) {
+			// All priorities are null — use legacy strictest-bounds logic
+			return findStrictestReferenceRange(conceptReferenceRanges);
+		}
+
+		int maxPriority = prioritizedRanges.stream().mapToInt(ConceptReferenceRange::getPriority).max().getAsInt();
+
+		List<ConceptReferenceRange> highestPriorityRanges = prioritizedRanges.stream()
+		        .filter(r -> r.getPriority() == maxPriority).collect(Collectors.toList());
```
jwnasambu: Kindly even though maxPriority is a init primitive, comparing it with a wrapper via == can be brittle. Let's make the unboxing explicit or add a null guard to prevent a NullPointerException and avoid any confusion regarding the Integer cache limits.

api/src/main/java/org/openmrs/ConceptReferenceRange.java
```diff
@@ -49,6 +49,16 @@ public class ConceptReferenceRange extends BaseReferenceRange implements Openmrs
 	@Column(name = "criteria", length = 65535)
 	private String criteria;
 
+	/**
+	 * Priority for this reference range. When multiple ranges match a patient, the one with the highest
+	 * priority (descending order) is selected. A null priority falls back to the legacy "strictest
+	 * bounds" merging behaviour for backward compatibility.
+	 *
+	 * @since 2.9.0
```
jwnasambu: The current project version in pom.xml is 3.0.0-SNAPSHOT and not 2.9.0.
```suggestion
	 @since 3.0.0
```
to match the current version.

Binayak490-cyber: Done.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
Binayak490-cyber: @jwnasambu, Fixed! Replaced `r.getPriority() == maxPriority` with `r.getPriority().equals(maxPriority)`. This is safe since `maxPriority` is an `int` primitive so `getPriority()` is never `null` at that point, but `.equals()` is the correct way to compare `Integer` objects and avoids any Integer cache boundary issues above 127.

Binayak490-cyber:
@jwnasambu, I have updated the PR according to your review, can you please again take a look at it now...!

Binayak490-cyber:
Hi @ibacher, @mseaton, @dkayiwa, @mogoodrich, @jwnasambu - raised this PR for TRUNK-6510. Adds a nullable priority field to ConceptReferenceRange so the highest-priority range wins when multiple ranges match a patient, with full backward compatibility when priority is null. All tests pass. Please review as it now needs your feedback over if any changes needs to be done over the fix!

IamMujuziMoses:
Thank you for this contribution @Binayak490-cyber though try getting reference here: https://wiki.openmrs.org/display/docs/Pull+Request+Tips while making a Pull request, especially about claiming an issue. When ready for review, remember to always hit the Request Code Review button as advised.

ibacher (approved):

api/src/main/java/org/openmrs/ConceptReferenceRange.java
```diff
@@ -49,6 +49,16 @@ public class ConceptReferenceRange extends BaseReferenceRange implements Openmrs
 	@Column(name = "criteria", length = 65535)
 	private String criteria;
 
+	/**
+	 * Priority for this reference range. When multiple ranges match a patient, the one with the highest
+	 * priority (descending order) is selected. A null priority falls back to the legacy "strictest
+	 * bounds" merging behaviour for backward compatibility.
+	 *
+	 * @since 3.0.0
```
ibacher:
```suggestion
	 * @since 3.0.0, 2.9.0
```

jwnasambu:
@Binayak490-cyber Kindly do you mind fixing the merge conflicts on this PR please!

PR #6042, opened 2026-04-17
TRUNK-6616: Create function in ObsService to efficiently return all previous versions of obs

jwnasambu:

I have updated the ObsService to optimize the retrieval of historical obs data. The previous implementation utilized an iterative approach that triggered a separate database query for every preceding version of an observation, leading to significant performance lag. I have replaced  it with a single recursive Common Table Expression (CTE) query allowing the system to fetch the entire version history in one efficient database operation. This change eliminates the "N+1 query" bottleneck and ensures the REST API remains responsive even when handling observations with extensive edit histories.

## Issue I worked on

https://openmrs.atlassian.net/browse/TRUNK-6616

  No? -> execute above command

## Summary by CodeRabbit

* **New Features**
  * Retrieve an observation's complete version history, ordered newest→oldest.
  * Enhanced observation filtering with support for filtering by visits.

* **Behavior Changes**
  * Input validation: calling the history retrieval with a null or unsaved observation now raises an error.

* **Tests**
  * Added tests and fixture validating multi-step version-history behavior and edge cases (null/unsaved obs, partial histories).


jwnasambu:
@chibongho Kindly feel free to review by PR at your convenient time please!

chibongho:
Defer to @ibacher and @dkayiwa, but LGTM. Thanks!

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -52,6 +52,10 @@
 @Repository("obsDAO")
 public class HibernateObsDAO implements ObsDAO {
 
+	private static final String OBS_ID_QUERY_PARAM = "obsId";
+
+	private static final String GET_SAVED_STATUS_SQL = "select status from obs where obs_id = :" + OBS_ID_QUERY_PARAM;
+
```
ibacher: I wouldn't define these as constants like this. It's inconsistent with everything else.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -67,6 +71,34 @@ public void deleteObs(Obs obs) throws DAOException {
...
+		        + "  SELECT obs_id, previous_version, date_created FROM obs WHERE obs_id = :startId" + "  UNION ALL"
+		        + "  SELECT o.obs_id, o.previous_version, o.date_created" + "  FROM obs o"
+		        + "  INNER JOIN version_chain vc ON o.obs_id = vc.previous_version" + ")"
+		        + "SELECT obs_id FROM version_chain ORDER BY date_created DESC";
+
+		Session session = sessionFactory.getCurrentSession();
+		List<Integer> ids = session.createNativeQuery(cteSql, Integer.class).setParameter("startId", obs.getObsId()).list();
+
+		List<Obs> history = new ArrayList<>(ids.size());
+		for (Integer id : ids) {
```
ibacher: We shouldn't loop through the ids like this. Just run "FROM Obs where Obs.id in (:obsIds)" (or whatever the appropriate JPAQL is here. This isn't that efficient because it requires N+1 selects for N obs.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -345,9 +377,8 @@ public Obs.Status getSavedStatus(Obs obs) {
 		FlushMode flushMode = session.getHibernateFlushMode();
 		session.setHibernateFlushMode(FlushMode.MANUAL);
 		try {
-			NativeQuery<String> sql = session.createNativeQuery("select status from obs where obs_id = :obsId",
-			    String.class);
-			sql.setParameter("obsId", obs.getObsId());
+			NativeQuery<String> sql = session.createNativeQuery(GET_SAVED_STATUS_SQL, String.class);
+			sql.setParameter(OBS_ID_QUERY_PARAM, obs.getObsId());
 			return Obs.Status.valueOf(sql.uniqueResult());
```
ibacher, line 351: This refactoring seems unrelated and should be done in a different PR.

api/src/main/java/org/openmrs/api/ObsService.java
```diff
@@ -40,6 +40,27 @@
...
+	 *
+	 * @param obs the <strong>current/latest</strong> Obs whose version history is to be retrieved; must
+	 *            not be null and must have a non-null obsId. The traversal walks backwards from this
+	 *            Obs via the {@code previousVersion} association: OpenMRS stores
+	 *            {@code previousVersion} on the <em>newer</em> Obs pointing to the <em>older</em> one
+	 *            it superseded. Passing an intermediate version will silently return only a partial
+	 *            history (from that version back to the original, excluding any newer versions that
+	 *            superseded it).
+	 * @return a List of Obs representing the full version history, including the given Obs itself
+	 * @since 2.7.0
```
ibacher: This `@since` isn't accurate.

api/src/main/java/org/openmrs/api/ObsService.java
```diff
@@ -40,6 +40,27 @@
...
 
+	/**
+	 * Returns the complete version history of the given Obs, including the Obs itself and all previous
+	 * versions it was revised from. Results are ordered from the most recently created version to the
+	 * oldest.
+	 * <p>
+	 * Voided obs are intentionally included because previous versions are always voided when a new
+	 * revision is saved via {@link #saveObs(Obs, String)}.
+	 *
+	 * @param obs the <strong>current/latest</strong> Obs whose version history is to be retrieved; must
```
ibacher:
```suggestion
	 * @param obs the Obs whose version history is to be retrieved; must
```

api/src/main/java/org/openmrs/api/ObsService.java
```diff
@@ -40,6 +40,27 @@
...
+	 *            not be null and must have a non-null obsId. The traversal walks backwards from this
+	 *            Obs via the {@code previousVersion} association: OpenMRS stores
+	 *            {@code previousVersion} on the <em>newer</em> Obs pointing to the <em>older</em> one
+	 *            it superseded. Passing an intermediate version will silently return only a partial
+	 *            history (from that version back to the original, excluding any newer versions that
+	 *            superseded it).
```
ibacher:
```suggestion
	 *            Obs via the {@code previousVersion} association.
```

api/src/main/java/org/openmrs/api/ObsService.java
```diff
@@ -40,6 +40,27 @@
...
+	 * @return a List of Obs representing the full version history, including the given Obs itself
```
ibacher:
```suggestion
	 * @return a List of Obs representing the full version history up to and including the supplied obs
```

api/src/main/java/org/openmrs/api/db/ObsDAO.java
```diff
@@ -89,6 +89,14 @@ public Long getObservationCount(List<Person> whom, List<Encounter> encounters, L
 	 */
 	public Obs.Status getSavedStatus(Obs obs);
 
+	/**
+	 * @see org.openmrs.api.ObsService#getObsVersionHistory(Obs)
+	 * @param obs the Obs to retrieve version history for; must not be null and must have a non-null
+	 *            obsId (precondition enforced by the service layer)
+	 * @since 2.7.0
```
ibacher: `@since` here isn't correct.

api/src/main/java/org/openmrs/api/db/ObsDAO.java
```diff
@@ -89,6 +89,14 @@ public Long getObservationCount(List<Person> whom, List<Encounter> encounters, L
 	 */
 	public Obs.Status getSavedStatus(Obs obs);
 
+	/**
+	 * @see org.openmrs.api.ObsService#getObsVersionHistory(Obs)
+	 * @param obs the Obs to retrieve version history for; must not be null and must have a non-null
+	 *            obsId (precondition enforced by the service layer)
```
ibacher:
```suggestion
	 * @param obs the Obs to retrieve version history for; must not be null and must have a non-null
	 *            obsId
```

ibacher: It's a bad idea for the DAO to _rely_ on the service layer being called. It should still check this itself.

ibacher (changes requested):
Thanks @jwnasambu!

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -67,6 +67,32 @@ public void deleteObs(Obs obs) throws DAOException {
...
+	public List<Obs> getObsVersionHistory(Obs obs) {
+		// Use a single recursive CTE to collect all obs_ids in the previousVersion chain,
+		// then load the Obs entities in one IN-clause query — O(1) round-trips instead of O(N).
+		// Voided obs are included intentionally — previous versions are voided by saveObs().
+		// Results are ordered newest→oldest (descending date_created) to match the traversal order.
+		String cteSql = "WITH RECURSIVE version_chain (obs_id, previous_version, date_created) AS ("
+		        + "  SELECT obs_id, previous_version, date_created FROM obs WHERE obs_id = :startId" + "  UNION ALL"
+		        + "  SELECT o.obs_id, o.previous_version, o.date_created" + "  FROM obs o"
+		        + "  INNER JOIN version_chain vc ON o.obs_id = vc.previous_version" + ")"
+		        + "SELECT obs_id FROM version_chain ORDER BY date_created DESC";
```
ibacher:
```suggestion
		String cteSql = "WITH RECURSIVE version_chain (obs_id, previous_version, date_created) AS ("
		        + "  SELECT obs_id, previous_version, date_created FROM obs WHERE obs_id = :startId   UNION ALL"
		        + "  SELECT o.obs_id, o.previous_version, o.date_created  FROM obs o"
		        + "  INNER JOIN version_chain vc ON o.obs_id = vc.previous_version)"
		        + "SELECT obs_id FROM version_chain ORDER BY date_created DESC";
```

api/src/main/java/org/openmrs/api/ObsService.java
```diff
@@ -40,6 +40,23 @@
  */
 public interface ObsService extends OpenmrsService {
 
+	/**
+	 * Returns the complete version history of the given Obs, including the Obs itself and all previous
+	 * versions it was revised from. Results are ordered from the most recently created version to the
+	 * oldest.
+	 * <p>
+	 * Voided obs are intentionally included because previous versions are always voided when a new
+	 * revision is saved via {@link #saveObs(Obs, String)}.
```
ibacher:
```suggestion
```

api/src/main/java/org/openmrs/api/ObsService.java
```diff
@@ -40,6 +40,23 @@
...
+	 *
+	 * @param obs the Obs whose version history is to be retrieved; must not be null and must have a
+	 *            non-null obsId. The traversal walks backwards from this Obs via the
+	 *            {@code previousVersion} association.
+	 * @return a List of Obs representing the full version history up to and including the supplied obs
+	 * @since 3.0.0
```
ibacher: We'll likely need to backport this, so `@since 3.0.0, 2.9.0` and whatever the next version of 2.8 will be, I think.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -67,6 +67,32 @@ public void deleteObs(Obs obs) throws DAOException {
...
+
+		Session session = sessionFactory.getCurrentSession();
+		List<Integer> ids = session.createNativeQuery(cteSql, Integer.class).setParameter("startId", obs.getObsId()).list();
+
+		if (ids.isEmpty()) {
+			return new ArrayList<>();
+		}
```
ibacher, line 100: It seems like in this case, we should return a list of exactly the passed in obs, e.g., `Arrays.asList(obs);`

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -67,6 +67,32 @@ public void deleteObs(Obs obs) throws DAOException {
 		sessionFactory.getCurrentSession().remove(obs);
 	}
 
+	/**
+	 * @see org.openmrs.api.ObsService#getObsVersionHistory(Obs)
+	 */
+	@Override
+	public List<Obs> getObsVersionHistory(Obs obs) {
+		// Use a single recursive CTE to collect all obs_ids in the previousVersion chain,
+		// then load the Obs entities in one IN-clause query — O(1) round-trips instead of O(N).
```
ibacher:
```suggestion
		// then load the Obs entities in one IN-clause query.
```

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
ibacher: There's actually no point to having the `ORDER BY` here if we have it in the second query.

api/src/test/java/org/openmrs/api/ObsServiceTest.java
```diff
@@ -2138,6 +2140,75 @@ public void saveObs_shouldVoidTheGivenObsAndSetReferenceRangeForTheNewObs() {
 		assertEquals(originalRange.getLowNormal(), newRange.getLowNormal());
 	}
 
+	/**
```
ibacher, line 2145: We shouldn't just have one happy-path chain of data, but also unrelated observations and a couple of different scenarios (e.g., a 15-obs chain and an obs with no previous members)

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -67,6 +68,35 @@ public void deleteObs(Obs obs) throws DAOException {
...
+
+		Session session = sessionFactory.getCurrentSession();
+		List<Integer> ids = session.createNativeQuery(cteSql, Integer.class).setParameter("startId", obs.getObsId()).list();
+
+		if (ids.isEmpty()) {
+			return Arrays.asList(obs);
+		}
+
+		return session.createQuery("FROM Obs o WHERE o.obsId IN (:obsIds) ORDER BY o.dateCreated DESC", Obs.class)
+		        .setParameter("obsIds", ids).getResultList();
```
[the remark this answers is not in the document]
ibacher: Won't let me submit a review, but this suggestion is correct. See the proposed fix for a possible implementation. Better to order this by the "depth" than trying to do this based on `dateCreated`.

jwnasambu:
@ibacher kindly feel free to review my PR please!

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -67,6 +69,43 @@ public void deleteObs(Obs obs) throws DAOException {
...
+	@Override
+	public List<Obs> getObsVersionHistory(Obs obs) {
+		if (obs == null) {
+			throw new IllegalArgumentException("Obs must not be null");
+		}
+		if (obs.getObsId() == null) {
+			throw new IllegalArgumentException("Obs must have an obsId");
+		}
+		// Use a single recursive CTE to collect all obs_ids in the previousVersion chain,
+		// then load the Obs entities in one IN-clause query.
```
ibacher, line 88:
```suggestion
		
		// Use a single recursive CTE to collect all obs_ids in the previousVersion chain,
		// then load the Obs entities in one IN-clause query.
		String cteSql = "WITH RECURSIVE version_chain (obs_id, previous_version, date_created) AS ("
```

api/src/main/java/org/openmrs/api/db/ObsDAO.java
```diff
@@ -89,6 +89,14 @@ public Long getObservationCount(List<Person> whom, List<Encounter> encounters, L
 	 */
 	public Obs.Status getSavedStatus(Obs obs);
 
+	/**
+	 * @see org.openmrs.api.ObsService#getObsVersionHistory(Obs)
+	 * @param obs the Obs to retrieve version history for; must not be null and must have a non-null
+	 *            obsId
+	 * @since 3.0.0, 2.9.0, 2.8.1
```
ibacher:
```suggestion
	 * @since 3.0.0, 2.9.0, 2.8.7
```

ibacher (approved):
Two small things, but LGTM otherwise.

jwnasambu (commented):
@ibacher Kindly help me understand how to fix the build failure above. Initially, it appeared to be caused by an incorrect context configuration. After fixing the issue locally, the build succeeds, but I am still getting a build failure on GitHub. What beats my understanding even before fixing the build failure, locally the build is successful.

ibacher:
@jwnasambu The output I think is pretty clear. You need to run `mvn spotless:apply` to fix the formatting. We added a check in CI that's likely skipped locally to ensure that things are formatted correctly.

jwnasambu:
@ibacher Thanks for pointing that out. Although I fixed the issue, I’m still encountering the original error. This is the same error I reported on slack channel https://pastebin.com/0D2WJNbi

jwnasambu:
@wikumChamith, @ibacher, @dkayiwa Kindly feel free to review my PR at your convenient time please!

jwnasambu:
@chibongho , @ibacher , @wikumChamith , @dkayiwa kindly help me review this PR at your convenient time please.

api/src/main/java/org/openmrs/api/ObsService.java
dkayiwa: Following up on this: the interface still says `and the next 2.8.1 release`, but `ObsDAO` got set to `2.8.7`, so the two do not match. And 2.8.7 has since been released (it is the current `<release>` in the openmrs-api maven-metadata, with 2.8.8-SNAPSHOT as the active 2.8 line now), so a backport merging today would first appear in 2.8.8. Could you set both this and `ObsDAO` to `@since 3.0.0, 2.9.0, 2.8.8`? The 2.9.0 and 3.0.0 parts are still accurate since neither has shipped.

jwnasambu: @dkayiwa, thanks for the review. I have addressed the proposed changes, and I learned a lot from your feedback. Kindly feel free to review my PR again at your convenience. Thank you!

jwnasambu:
@chibongho , @ibacher , @wikumChamith , @dkayiwa kindly help me review this PR at your convenient time please.

jwnasambu:
@claude review

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -67,6 +69,42 @@ public void deleteObs(Obs obs) throws DAOException {
...
+			throw new IllegalArgumentException("Obs must have an obsId");
+		}
+
+		// Use a single recursive CTE to collect all obs_ids in the previousVersion chain,
+		// then load the Obs entities in one IN-clause query.
+		String cteSql = "WITH RECURSIVE version_chain (obs_id, previous_version, depth) AS ("
+		        + "  SELECT obs_id, previous_version, 0 FROM obs WHERE obs_id = :startId UNION ALL"
+		        + "  SELECT o.obs_id, o.previous_version, vc.depth + 1 FROM obs o"
+		        + "  INNER JOIN version_chain vc ON o.obs_id = vc.previous_version)"
+		        + " SELECT obs_id FROM version_chain ORDER BY depth";
```
dkayiwa, line 93: Recursive CTEs need MySQL 8. On 5.7 `WITH RECURSIVE` is a plain syntax error, which I confirmed against the image this project's own `Containers.newMySQLContainer()` pins:

```
mysql:5.7.39> WITH RECURSIVE vc (n) AS (SELECT 1 UNION ALL SELECT n+1 FROM vc WHERE n < 3) SELECT * FROM vc;
ERROR 1064 (42000): You have an error in your SQL syntax ... near 'RECURSIVE vc (n) AS ...'
```

The query as written is fine everywhere current. I ran it against MySQL 8.0.46, PostgreSQL 16 and the H2 the suite actually uses, and it walks the chain in the right order on all three. Master is not new ground for this either: `HibernateLocationDAO.getLocations` already runs a recursive CTE, and JobRunr's `v016__alter_jobs_stats_add_awaiting_jobs.sql` migration uses one too, so on 5.7 the Spring context does not even start (that is what I hit when I pointed the suite at the 5.7 container instead of H2).

2.8.x is the one I would check before backporting. It has no JobRunr, no recursive CTE anywhere, and still pins `mysql:5.7.39`, so this would be the first hard MySQL 8 requirement in that line, and it would surface as a 1064 on a live install rather than in CI, since the tests that cover this method run against in-memory H2. Given the `@since` names 2.8.x, is that backport still on, or does 2.8.x need to keep walking the chain iteratively? @ibacher likely has the call there.

api/src/main/java/org/openmrs/api/ObsService.java
```diff
@@ -40,6 +40,21 @@
  */
 public interface ObsService extends OpenmrsService {
 
+	/**
+	 * Returns the complete version history of the given Obs, including the Obs itself and all previous
+	 * versions it was revised from. Results are ordered from the most recently created version to the
+	 * oldest.
```
dkayiwa: Both halves of this sentence promise more than the method delivers. Starting from an intermediate revision returns the history only up to that obs, not the complete chain, which is what `getObsVersionHistory_shouldReturnPartialHistoryForIntermediateVersion` asserts (90102 is the most recently created version in that chain and it is deliberately left out). And the ordering key is position in the `previousVersion` chain rather than creation time, which is the whole point of the CTE carrying `depth` instead of `date_created`. Since this is the first thing a caller reads, it is worth stating what the method actually guarantees:

```suggestion
	 * Returns the given Obs together with every earlier version it was revised from, walking backwards
	 * along the {@code previousVersion} chain. The supplied Obs comes first and each following element
	 * is the version immediately before it, so passing an intermediate revision returns the history up
	 * to that revision rather than the whole chain.
```

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -67,6 +69,42 @@ public void deleteObs(Obs obs) throws DAOException {
...
+
+		Session session = sessionFactory.getCurrentSession();
+		List<Integer> ids = session.createNativeQuery(cteSql, Integer.class).setParameter("startId", obs.getObsId()).list();
+
+		if (ids.isEmpty()) {
+			return Arrays.asList(obs);
+		}
```
dkayiwa, line 100: With the null-obsId guard above, an empty `ids` can now only mean there is no `obs` row with that id, say a caller holding a stale `new Obs(id)` or an id that has since been purged. Returning the passed obs then reports that stub as a one-entry version history, concept and person null, with nothing to tell the caller it was never found. `Arrays.asList(obs)` fitted the earlier shape of this method, where this branch was the unsaved-obs case; for "no such row" I would hand back an empty list.

```suggestion
		if (ids.isEmpty()) {
			return Collections.emptyList();
		}
```
That needs `java.util.Collections` imported, and it leaves `java.util.Arrays` unused so spotless will drop that import.

api/src/test/java/org/openmrs/api/ObsServiceTest.java
```diff
@@ -2138,6 +2140,94 @@ public void saveObs_shouldVoidTheGivenObsAndSetReferenceRangeForTheNewObs() {
...
+			o.setPerson(latestObs.getPerson());
+			o.setObsDatetime(new java.util.Date());
+			o.setValueNumeric((double) i);
+			if (prev != null) {
+				o.setPreviousVersion(prev);
+			} else {
+				first = o;
+			}
+			prev = Context.getObsService().saveObs(o, "chain test");
+		}
```
dkayiwa, line 2177: This chain is assembled by calling `setPreviousVersion` on brand new obs, which is not how one comes to exist in a running system. Real revisions go through `saveObs(existingObs, changeMessage)`, which copies the obs, sets `previousVersion` on the copy and voids the original, and no test here covers that path: the three-version case hard-codes `previous_version` in the fixture and this loop sets it by hand. The CTE reads the database rather than the in-memory object graph the REST loop used to walk, so one case that revises an obs twice through `saveObs` and asserts the resulting three-entry history would be worth having. I tried it locally and it passes, so this is about keeping it that way rather than a bug you need to chase.

api/src/main/java/org/openmrs/api/ObsService.java
dkayiwa: 2.8.8 has shipped since I last looked: the 2.8.8 tag is in the repo and the 2.8.x branch is now on 2.8.9-SNAPSHOT, so a backport merging today would first appear in 2.8.9, here and in the `ObsDAO` copy. 2.9.0 and 3.0.0 are still unreleased, so those parts stay as they are. One caveat before you change the number: see the comment I left on the CTE in `HibernateObsDAO`. If 2.8.x cannot take a recursive CTE, the 2.8 part should come off this tag altogether rather than move to 2.8.9.

api/src/test/java/org/openmrs/api/ObsServiceTest.java
```diff
@@ -2138,6 +2140,117 @@ public void saveObs_shouldVoidTheGivenObsAndSetReferenceRangeForTheNewObs() {
...
+		List<Obs> longChain = obsService.getObsVersionHistory(prev);
+		assertNotNull(longChain);
+		assertEquals(15, longChain.size());
```
dkayiwa, line 2180: This suite would be worth one more test, on the property the change exists for. Every assertion here is about the contents and order of the returned list, and a plain loop following `getPreviousVersion()` one row at a time satisfies all of them, so a refactor that quietly put back a select per version would leave every one of these tests green.

I measured it on this branch: with `sessionFactory.getStatistics()` cleared right before the call, a 3-obs chain and a 40-obs chain each come to exactly 2 prepared statements, the CTE and the IN-clause load. `PatientIdentifierBatchLoadTest`, added by the TRUNK-6734 commit this branch sits on, is the shape to copy, with `getPrepareStatementCount()` in place of the collection fetch count.

api/src/main/java/org/openmrs/api/ObsService.java
dkayiwa: Still `2.8.8` here and in the `ObsDAO` copy. 2.8.8 was released on 2026-07-09 and shipped without this method, and the 2.8.x branch is already on 2.8.9-SNAPSHOT, so as it stands the tag names a release the API is not in. Worth fixing before this merges, since a wrong number is much harder to walk back once 3.0.0 ships.

One more input for choosing between 2.8.9 and dropping the 2.8 part: both 2.8.x and 2.9.x compile at Java 8 (`javaCompilerVersion` is 1.8 in each pom, with no module or profile overriding it), and `ids.stream()...toList()` needs Java 16. So neither line takes this as a cherry-pick, quite apart from the MySQL question on the CTE.

api/src/main/java/org/openmrs/api/ObsService.java
```diff
@@ -40,6 +40,22 @@
...
+	/**
+	 * Returns the given Obs together with every earlier version it was revised from, walking backwards
+	 * along the {@code previousVersion} chain. The supplied Obs comes first and each following element
+	 * is the version immediately before it, so passing an intermediate revision returns the history up
+	 * to that revision rather than the whole chain.
+	 *
+	 * @param obs the Obs whose version history is to be retrieved; must not be null and must have a
+	 *            non-null obsId. The traversal walks backwards from this Obs via the
+	 *            {@code previousVersion} association.
+	 * @return a List of Obs representing the full version history up to and including the supplied obs
```
dkayiwa: This `@return` line needs a third case. If the obs carries an id with no matching row, the DAO now hands back an empty list, so a caller who took "the supplied Obs comes first" at face value and wrote `history.get(0)` gets an `IndexOutOfBoundsException` instead. That is reachable with the stub-by-id idiom this codebase leans on freely (`buildObservation` further down `ObsServiceTest` builds `new Location(1)` and `new Concept(3)` the same way); I ran `getObsVersionHistory(new Obs(999999))` on this branch and it comes back empty rather than throwing. Nothing in the suite reaches that branch either, so it is worth a test next to the doc line.

```suggestion
	 * @return a List of Obs representing the full version history up to and including the supplied obs, or an
	 *         empty List if no obs with the given obsId exists
```

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -67,6 +69,42 @@ public void deleteObs(Obs obs) throws DAOException {
...
+			return Collections.emptyList();
+		}
+
+		List<Obs> loaded = session.createQuery("FROM Obs o WHERE o.obsId IN (:obsIds)", Obs.class)
+		        .setParameter("obsIds", ids).getResultList();
+		Map<Integer, Obs> byId = new java.util.HashMap<>();
+		for (Obs o : loaded) {
+			byId.put(o.getObsId(), o);
+		}
+		return ids.stream().map(byId::get).filter(java.util.Objects::nonNull).toList();
```
dkayiwa: `toList()` hands back an unmodifiable list, which leaves this the only list-returning method on `ObsService` whose result a caller cannot reorder. I checked on this branch: what comes back is an `ImmutableCollections$ListN`, and both `Collections.reverse(history)` and `history.sort(...)` throw `UnsupportedOperationException`. Reversing is the obvious thing to want, since the list is newest-first and a timeline usually renders oldest-first, and everything else here (`getObservations` and the methods that delegate to it) returns Hibernate's own result list, which is mutable. I would switch to `Collectors.toList()` rather than document the immutability, partly for that consistency and partly because it is the form that can be cherry-picked at all, `Stream.toList()` needing Java 16.

```suggestion
		return ids.stream().map(byId::get).filter(Objects::nonNull).collect(Collectors.toList());
```
That needs `java.util.Objects` and `java.util.stream.Collectors` imported, which also drops the inline qualification this line currently carries.

api/src/main/java/org/openmrs/api/ObsService.java
dkayiwa: Still `2.8.8` here and in the `ObsDAO` copy. The 2.8.8 tag landed on 9 July and shipped without this method, and 2.8.x is on 2.8.9-SNAPSHOT, so as it stands the 3.0.0 javadoc will tell readers the method is in 2.8.8, and anyone who pins 2.8.8 and writes against it gets a `NoSuchMethodError`. A wrong number is much harder to walk back once 3.0.0 ships, so I would settle this before merging.

On the choice between 2.8.9 and dropping the 2.8 part: `toCollection(ArrayList::new)` compiles at Java 8, so the cherry-pick objection I raised above about `toList()` no longer applies and 2.9.0 can stay as it is. What is still open is the MySQL question on the CTE in `HibernateObsDAO`. If 2.8.x can take a recursive CTE this becomes `3.0.0, 2.9.0, 2.8.9`; if not, drop the 2.8 piece and leave `3.0.0, 2.9.0`.

api/src/main/java/org/openmrs/api/ObsService.java
dkayiwa: Nothing in the suite reaches the branch this line documents. I ran `getObsVersionHistory(new Obs(999999))` against the current head and it comes back empty, so a short test next to `getObsVersionHistory_shouldReturnOnlyTheObsItselfWhenNoPreviousVersionExists` would pin it. The `ids.isEmpty()` return is the only path through the method with nothing exercising it, and it is easy to reach by accident given how freely this codebase builds stubs by id.

jwnasambu:
@claude review

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
ibacher: Yeah, this is non-backportable, but we ran into a similar issue with Locations. As I recall, we agreed there to the CTE in 3.x / `master` but the 2.x backport version was substantially less efficient. I think we're going to need to pay that price here as well.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -67,6 +71,42 @@ public void deleteObs(Obs obs) throws DAOException {
 		sessionFactory.getCurrentSession().remove(obs);
 	}
 
+	/**
+	 * @see org.openmrs.api.ObsService#getObsVersionHistory(Obs)
+	 */
+	@Override
+	public List<Obs> getObsVersionHistory(Obs obs) {
+		if (obs == null) {
+			throw new IllegalArgumentException("Obs must not be null");
```
ibacher: We have an `APIException` class for pretty much exactly this use case and the next one. I think it's probably better to throw from there.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -67,6 +71,42 @@ public void deleteObs(Obs obs) throws DAOException {
...
+		}
+		if (obs.getObsId() == null) {
+			throw new IllegalArgumentException("Obs must have an obsId");
```
ibacher: Maybe this could be reworded? The real thing where checking here (or trying to) is whether or not the obs is _committed_ or if it's just _in memory_. The message here, as written, might encourage callers to assign an arbitrary `obsId`, which would return meaningless data, but that might not be immediately noticed.

jwnasambu:

api/src/main/java/org/openmrs/api/db/ObsDAO.java
```diff
@@ -89,6 +89,14 @@ public Long getObservationCount(List<Person> whom, List<Encounter> encounters, L
 	 */
 	public Obs.Status getSavedStatus(Obs obs);
 
+	/**
+	 * @see org.openmrs.api.ObsService#getObsVersionHistory(Obs)
+	 * @param obs the Obs to retrieve version history for; must not be null and must have a non-null
+	 *            obsId
+	 * @since 3.0.0, 2.9.0
```
ibacher:
```suggestion
	 * @since 2.9.0, 3.0.0
```

api/src/main/java/org/openmrs/api/ObsService.java
```diff
@@ -40,6 +40,23 @@
...
+	 * is the version immediately before it, so passing an intermediate revision returns the history up
+	 * to that revision rather than the whole chain.
+	 *
+	 * @param obs the Obs whose version history is to be retrieved; must not be null and must be
+	 *            persisted (i.e. have a database-assigned obsId). The traversal walks backwards from
+	 *            this Obs via the {@code previousVersion} association.
+	 * @return a List of Obs representing the full version history up to and including the supplied obs,
+	 *         or an empty List if no obs with the given obsId exists
+	 * @throws APIException if obs is null or is not persisted (has no database-assigned obsId)
+	 * @since 3.0.0, 2.9.0
```
ibacher:
```suggestion
	 * @since 2.9.0, 3.0.0
```

jwnasambu:
@ibacher Kindly feel free to review my PR at your convenient time please!

jwnasambu:
@ibacher Kindly feel free to review my PR at your convenient time please!

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -67,6 +72,42 @@ public void deleteObs(Obs obs) throws DAOException {
...
+		        + " SELECT obs_id FROM version_chain ORDER BY depth";
+
+		Session session = sessionFactory.getCurrentSession();
+		List<Integer> ids = session.createNativeQuery(cteSql, Integer.class).setParameter("startId", obs.getObsId()).list();
+
+		if (ids.isEmpty()) {
+			return Collections.emptyList();
+		}
+
+		List<Obs> loaded = session.createQuery("FROM Obs o WHERE o.obsId IN (:obsIds)", Obs.class)
```
dkayiwa: These obs come back from an HQL query rather than by id, and `Obs.hbm.xml:101` maps `referenceRange` as an unowned `<one-to-one>`, which Hibernate cannot proxy. So it resolves that association with its own `select ... from obs_reference_range where obs_id = ?` for every row of this result, and the per-version query TRUNK-6616 asks us to remove is still here, just against a different table.

I measured it on this branch with the session cleared before the call, so nothing is served out of the first level cache:

| chain length | this method | the `getPreviousVersion()` loop it replaces |
| --- | --- | --- |
| 3 versions | 5 statements | 3 |
| 40 versions | 42 statements | 40 |

The loop comes out lower because loading an obs by id join fetches the reference range in the same statement, which an HQL query does not do. If this merges as is, a REST read of an obs with an N version history makes N+2 round trips where today it makes N, so the N+1 is not gone and that endpoint ends up slightly slower than before rather than faster.

One join fetch closes it. `referenceRange` is the only `<one-to-one>` on `Obs`, and it is a to-one, so it cannot multiply rows:

```suggestion
		List<Obs> loaded = session
		        .createQuery("FROM Obs o LEFT JOIN FETCH o.referenceRange WHERE o.obsId IN (:obsIds)", Obs.class)
```

With that in, the 40 version chain comes to exactly 2 statements, and all 94 tests in `ObsServiceTest` pass.

api/src/test/java/org/openmrs/api/ObsServiceTest.java
```diff
@@ -2138,6 +2142,171 @@ public void saveObs_shouldVoidTheGivenObsAndSetReferenceRangeForTheNewObs() {
...
+			if (prev != null) {
+				o.setPreviousVersion(prev);
+			}
+			prev = Context.getObsService().saveObs(o, "chain test");
+		}
+
+		// flush pending inserts and reset the counters so that only the version history
+		// lookup is measured
+		SessionFactory sessionFactory = (SessionFactory) applicationContext.getBean("sessionFactory");
+		sessionFactory.getCurrentSession().flush();
```
dkayiwa, line 2790: The 40 obs are still managed when the count is taken, so this measures a warm session and cannot see the load it is meant to be pinning. Swapping a `getPreviousVersion()` loop into the DAO makes this test report 0 statements rather than 40, because the walk never leaves the first level cache, and the per-obs reference range selects I flagged in `HibernateObsDAO` do not happen here either, so the count reads 2 whether or not they are there.

Clearing the session after the flush makes it measure what the method actually costs a caller. I ran the combinations on this branch: with the clear, the loop reports 40, the code as it stands reports 42, and 2 once the reference range is join fetched. Without it, the loop reports 0 and both versions of the current code report 2.

```suggestion
		sessionFactory.getCurrentSession().flush();
		sessionFactory.getCurrentSession().clear();
```

jwnasambu:
@ibacher, @dkayiwa kindly help me review this PR at your convenient time please.

ibacher:
@jwnasambu I still see several comments here that aren't addressed yet?

jwnasambu:
@dkayiwa, @ibacher Kindly feel free to review my PR at your convenient time please

api/src/test/java/org/openmrs/api/ObsServiceTest.java
ibacher: @jwnasambu This hasn't been addressed yet, right?

api/src/test/java/org/openmrs/api/ObsServiceTest.java
ibacher: @jwnasambu Or this

api/src/test/java/org/openmrs/api/ObsServiceTest.java
ibacher: @jwnasambu Or this

jwnasambu:
@ibacher Kindly I have pushed some proposed changes feel free to review my PR at your convenient time please!

jwnasambu:
@ibacher Kindly I have pushed some proposed changes feel free to review my PR at your convenient time please!

jwnasambu:
@ibacher Kindly I have pushed some proposed changes feel free to review my PR at your convenient time please!

jwnasambu:
@dkayiwa, @ibacher, @chibongho Kindly feel free to review my PR at your convenient time please

api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java
```diff
@@ -80,6 +85,43 @@ public void deleteObs(Obs obs) throws DAOException {
...
+		// Use a single recursive CTE to collect all obs_ids in the previousVersion chain,
+		// then load the Obs entities in one IN-clause query.
+		String cteSql = "WITH RECURSIVE version_chain (obs_id, previous_version, depth) AS ("
+		        + "  SELECT obs_id, previous_version, 0 FROM obs WHERE obs_id = :startId UNION ALL"
+		        + "  SELECT o.obs_id, o.previous_version, vc.depth + 1 FROM obs o"
+		        + "  INNER JOIN version_chain vc ON o.obs_id = vc.previous_version)"
+		        + " SELECT obs_id FROM version_chain ORDER BY depth";
+
+		Session session = sessionFactory.getCurrentSession();
+		List<Integer> ids = session.createNativeQuery(cteSql, Integer.class).setParameter("startId", obs.getObsId()).list();
```
dkayiwa: On MySQL-family databases this history stops at about a thousand versions, and on MariaDB it stops without any error. MySQL 8.0 and MariaDB 10.11 both cap a recursive CTE at 1000 iterations by default (`cte_max_recursion_depth` and `max_recursive_iterations`). I ran this method against MariaDB 10.11, which is what the RefApp ships, on chains built through `saveObs`: up to 1001 versions it returns everything, but a 1002-version chain comes back as 1001 entries with the original missing, and a 1500-version chain comes back as 1001 as well, with no exception either time (MariaDB only raises warning 1931 on the statement). On MySQL 8.0 the same query fails with `ERROR 3636 (HY000): Recursive query aborted after 1001 iterations` as soon as a chain passes 1000 versions. The suite cannot see either, because H2 has no such cap (it returns a 1500-version chain whole) and the longest chain the tests build is 40.

An obs needs over a thousand versions to get there, so I would not hold the PR on this, but the MariaDB half hands the caller a short history that it cannot tell apart from a complete one. Capping the recursive member below the limit and resuming from the last row's `previous_version` removes the ceiling without touching the common case. With the change below, chains of up to 1000 versions still take two statements (`getObsVersionHistory_shouldLoadTheWholeChainInTwoQueries` and the rest of `ObsServiceTest` pass as they are), each further 1000 versions adds one more CTE, and 2500-version chains come back whole and in order on both MariaDB and H2:

```suggestion
		// Use a recursive CTE to collect the obs_ids in the previousVersion chain, then load the Obs
		// entities in one IN-clause query. MySQL (cte_max_recursion_depth) and MariaDB
		// (max_recursive_iterations) stop a recursive CTE after 1000 iterations by default, so each query
		// walks at most 1000 versions and the next one resumes from where it stopped, until the chain ends
		// or loops back on itself.
		String cteSql = "WITH RECURSIVE version_chain (obs_id, previous_version, depth) AS ("
		        + "  SELECT obs_id, previous_version, 0 FROM obs WHERE obs_id = :startId UNION ALL"
		        + "  SELECT o.obs_id, o.previous_version, vc.depth + 1 FROM obs o"
		        + "  INNER JOIN version_chain vc ON o.obs_id = vc.previous_version WHERE vc.depth < 999)"
		        + " SELECT obs_id, previous_version FROM version_chain ORDER BY depth";

		Session session = sessionFactory.getCurrentSession();
		List<Integer> ids = new ArrayList<>();
		Integer startId = obs.getObsId();
		while (startId != null && !ids.contains(startId)) {
			List<Object[]> rows = session.createNativeQuery(cteSql, Object[].class).setParameter("startId", startId).list();
			startId = null;
			for (Object[] row : rows) {
				ids.add(((Number) row[0]).intValue());
				startId = row[1] == null ? null : ((Number) row[1]).intValue();
			}
		}
```

If you take it, a test on a 1001-version chain would pin it, and that works on H2 as well since the cap is now in the SQL: I measured three statements for it there (two CTE batches and the load).

PR #6056, opened 2026-04-22
TRUNK-6624 : Add since and forRemoval arguments to deprecated annotations in Allergy

imandeepduhan:
Update deprecated methods in the Allergy class to include since and forRemoval arguments in the @Deprecated annotation.

This change improves compatibility with newer Java standards and resolves static analysis warnings.

**Updated methods:**

● getId()
● setId(Integer allergyId)
● getComment()
● setComment(String comment)

**All deprecated methods now use:**

@Deprecated(since = "2.3.0", forRemoval = false)

https://openmrs.atlassian.net/browse/TRUNK-6624

### DIFF
--- a/api/src/main/java/org/openmrs/Allergy.java
+++ b/api/src/main/java/org/openmrs/Allergy.java
@@ -111,6 +111,7 @@ public class Allergy extends BaseFormRecordableOpenmrsData {
 	 * @see org.openmrs.OpenmrsObject#getId()
 	 */
 	@Override
+	@Deprecated(since = "2.3.0", forRemoval = false)
 	public Integer getId() {
 		return allergyId;
 	}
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/Allergy.java
```diff
@@ -85,198 +86,198 @@ public Allergy(Patient patient, Allergen allergen, Concept severity, String comm
...
 		return getComments();
 	}
-
+	
 	/**
 	 * @param comment the comment to set
 	 * @deprecated as of 2.3.0, replaced by {@link #setComments(String)}
 	 */
-	@Deprecated
+	@Deprecated(since = "2.3.0", forRemoval = false)
```
Binayak490-cyber: @imandeepduhan, why you have done this, what was the problem with the older code...?

imandeepduhan: Added "since" to clearly indicate the version when it was deprecated, and "forRemoval = false" to show it’s not planned for removal yet—just aligning with modern Java deprecation standards.

PR #6061, opened 2026-04-24
TRUNK-6623: Add null guard to purgePatient for consistency

Gaurav-E3508:
## Description
Added a null check (guard clause) in the purgePatient method of PatientServiceImpl to ensure consistent handling of null inputs.

## Problem
Previously, passing a null patient object could lead to unexpected behavior or potential NullPointerException.

## Solution
Introduced a null guard to safely handle null inputs before proceeding with the purge operation.

## Impact
- Improves code robustness
- Prevents runtime errors
- Ensures consistency with similar service methods

## Testing
- Ran existing unit tests
- No test failures observed

## Related Issue
https://issues.openmrs.org/browse/TRUNK-6623

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -345,6 +345,9 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 	 */
 	@Override
 	public void purgePatient(Patient patient) throws APIException {
+		if (patient == null) {
+			return;
+		}
 		dao.deletePatient(patient);
 	}
 


api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -345,6 +345,9 @@ public Patient unvoidPatient(Patient patient) throws APIException {
 	 */
 	@Override
 	public void purgePatient(Patient patient) throws APIException {
+		if (patient == null) {
```
omeriinnocent, line 348: Hello @Gaurav-E3508,Should `purgePatient(null)` instead throw a clear exception (e.g., `IllegalArgumentException("patient must not be null")` or an `APIException`) to match OpenMRS service-layer expectations, rather than silently returning?

omeriinnocent: silently returning may mask upstream bugs.

Gaurav-E3508: Thanks for the feedback. Updated the implementation to throw an `APIException` when `purgePatient(null)` is called instead of silently returning, to better align with OpenMRS service-layer expectations and avoid masking upstream issues.

PR #6065, opened 2026-04-29
TRUNK-6590: Both /openmrs and /openmrs/ return 404

shivvani-r:
Fixed the 404 when navigating to `/openmrs` or `/openmrs/` in the Docker/Tomcat 11 environment. 

The `/openmrs` context is configured correctly, and the deployed WAR contains a physical `/index.htm` welcome file. However, because `web.xml` maps `*.htm` requests to the OpenMRS `DispatcherServlet`, `/openmrs/index.htm` was routed to Spring MVC instead of being served as the welcome file, resulting in:

```text
No endpoint GET /openmrs/index.htm
```

Since `/openmrs/` uses `index.htm` as its welcome file, the context root also returned 404.

Specific changes include:

This PR adds an exact `/index.htm` mapping to the container default servlet, allowing only the physical welcome file to be served statically while leaving the existing OpenMRS mappings for `*.htm`, `*.form`, `*.list`, `*.portlet`, `*.field`, and `/ws/*` unchanged.

## Issue I worked on
see https://openmrs.atlassian.net/browse/TRUNK-6590

- [ ] I have **added tests** to cover my changes. (The fix was verified via manual integration testing/curl in the Docker environment as it addresses deployment-level routing).

## Summary by CodeRabbit

* **Chores**
  * Updated routing so requests to /ws/* are handled by the designated API handler.
  * Adjusted routing so /index.htm is served by the default handler, improving request resolution and routing clarity.

### DIFF
--- a/webapp/src/main/webapp/WEB-INF/web.xml
+++ b/webapp/src/main/webapp/WEB-INF/web.xml
@@ -323,6 +323,11 @@
  		<servlet-name>openmrs_static_content</servlet-name>
  		<url-pattern>/scripts/*</url-pattern>
 	</servlet-mapping>
+
+	<servlet-mapping>
+		<servlet-name>openmrs</servlet-name>
+		<url-pattern>/</url-pattern>
+	</servlet-mapping>
 	
 	<servlet-mapping>
  		<servlet-name>openmrs</servlet-name>
--- a/webapp/src/main/webapp/WEB-INF/web.xml
+++ b/webapp/src/main/webapp/WEB-INF/web.xml
@@ -323,15 +323,15 @@
  		<servlet-name>openmrs_static_content</servlet-name>
  		<url-pattern>/scripts/*</url-pattern>
 	</servlet-mapping>
-
-	<servlet-mapping>
-		<servlet-name>openmrs</servlet-name>
-		<url-pattern>/</url-pattern>
-	</servlet-mapping>
 	
 	<servlet-mapping>
  		<servlet-name>openmrs</servlet-name>
  		<url-pattern>/ws/*</url-pattern>
+	</servlet-mapping>
+
+	<servlet-mapping>
+		<servlet-name>default</servlet-name>
+		<url-pattern>/index.htm</url-pattern>
 	</servlet-mapping>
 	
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -323,6 +323,11 @@
  		<servlet-name>openmrs_static_content</servlet-name>
  		<url-pattern>/scripts/*</url-pattern>
 	</servlet-mapping>
+
+	<servlet-mapping>
+		<servlet-name>openmrs</servlet-name>
+		<url-pattern>/</url-pattern>
+	</servlet-mapping>
```
Muta-Jonathan: I understand this was meant to fix the `/openmrs` 404 issue, but the existing servlet mappings are already correct for OpenMRS routing. The problem likely lies in context/path configuration rather than `web.xml`. Changing these mappings may interfere with request handling instead of resolving the root cause.

shivvani-r: Thanks, I understand that my original approach of mapping the DispatcherServlet to "/" was broader than necessary.

The context path itself is configured correctly: /openmrs redirects to /openmrs/. I also confirmed that the deployed WAR contains a physical /index.htm file.

The issue is that, under Tomcat 11, the existing *.htm mapping routes /index.htm to the OpenMRS DispatcherServlet instead of allowing the physical welcome file to be served. The DispatcherServlet then returns "No endpoint GET /openmrs/index.htm." Since /openmrs/ uses index.htm as its welcome file, the context root also returns 404.

I revised the PR to avoid global DispatcherServlet changes. The new approach adds an exact /index.htm mapping to the container default servlet, so only the physical welcome file is served statically while the existing *.htm, *.form, *.list, etc. mappings remain unchanged.

Lmk if this sounds good.

Muta-Jonathan (approved):
LGTM thanks @shivvani-r 
cc @dkayiwa @rkorytkowski

dkayiwa:
@coderabbitai full review

Muta-Jonathan:
@coderabbitai review

PR #6079, opened 2026-05-05
TRUNK-6633 Add a way to search Concepts by ConceptSearchCriteria,

chibongho:

TRUNK-6633 Add a way to search Concepts by ConceptSearchCriteria

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6633

  No? -> execute above command

## Summary by CodeRabbit

* **New Features**
  * Flexible concept search: query by UUID, numeric ID, mapping (source:code) or name, with option to include/exclude retired concepts; identifier matches are combined and global filters applied.
  * Fluent builder to compose search criteria and a single read-only search entry to execute queries.

* **Tests**
  * Added tests for UUID/ID/mapping/name searches, retired filtering, unions of criteria, empty-criteria behavior, and builder reference resolution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openmrs/openmrs-core/pull/6079)

### DIFF
--- a/api/src/main/java/org/openmrs/api/ConceptService.java
+++ b/api/src/main/java/org/openmrs/api/ConceptService.java
@@ -99,13 +99,13 @@ public interface ConceptService extends OpenmrsService {
 
 	/**
 	 * Search for concepts matching the given criteria. The identifier fields ({@code uuids},
-	 * {@code conceptIds}, and {@code mappingCode}/{@code mappingSourceName}) are combined with OR
-	 * semantics — a concept matching any one of them is included. Global filter fields such as
-	 * {@code includeRetired} are then applied with AND semantics on top of those results.
+	 * {@code conceptIds}, and {@code mappings}) are combined with OR semantics — a concept matching any
+	 * one of them is included. Global filter fields such as {@code includeRetired} are then applied
+	 * with AND semantics on top of those results.
 	 *
 	 * @param conceptSearchCriteria the search criteria
 	 * @return list of matching concepts
-	 * @since 2.8.0
+	 * @since 2.9.0
 	 */
 	@Authorized(PrivilegeConstants.GET_CONCEPTS)
 	public List<Concept> getConcepts(ConceptSearchCriteria conceptSearchCriteria);
--- a/api/src/main/java/org/openmrs/api/ConceptService.java
+++ b/api/src/main/java/org/openmrs/api/ConceptService.java
@@ -42,6 +42,7 @@ import org.openmrs.DrugIngredient;
 import org.openmrs.Person;
 import org.openmrs.annotation.Authorized;
 import org.openmrs.api.db.ConceptDAO;
+import org.openmrs.parameter.ConceptSearchCriteria;
 import org.openmrs.util.PrivilegeConstants;
 
 /**
@@ -95,6 +96,19 @@ public interface ConceptService extends OpenmrsService {
 	 */
 	@Authorized(PrivilegeConstants.GET_CONCEPTS)
 	public Concept getConceptByUuid(String uuid);
+
+	/**
+	 * Search for concepts matching the given criteria. The identifier fields ({@code uuids},
+	 * {@code conceptIds}, and {@code mappingCode}/{@code mappingSourceName}) are combined with OR
+	 * semantics — a concept matching any one of them is included. Global filter fields such as
+	 * {@code includeRetired} are then applied with AND semantics on top of those results.
+	 *
+	 * @param conceptSearchCriteria the search criteria
+	 * @return list of matching concepts
+	 * @since 2.8.0
+	 */
+	@Authorized(PrivilegeConstants.GET_CONCEPTS)
+	public List<Concept> getConcepts(ConceptSearchCriteria conceptSearchCriteria);
 
 	/**
 	 * Save or update the given <code>Concept</code> or <code>ConceptNumeric</code> in the database
--- a/api/src/main/java/org/openmrs/api/ConceptService.java
+++ b/api/src/main/java/org/openmrs/api/ConceptService.java
@@ -105,7 +105,7 @@ public interface ConceptService extends OpenmrsService {
 	 *
 	 * @param conceptSearchCriteria the search criteria
 	 * @return list of matching concepts
-	 * @since 2.9.0
+	 * @since 2.8.7
 	 */
 	@Authorized(PrivilegeConstants.GET_CONCEPTS)
 	public List<Concept> getConcepts(ConceptSearchCriteria conceptSearchCriteria);
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
@@ -1165,30 +1165,36 @@ public class HibernateConceptDAO implements ConceptDAO {
 			orPredicates.add(root.get("conceptId").in(criteria.getConceptIds()));
 		}
 
-		if (criteria.getMappingCode() != null && criteria.getMappingSourceName() != null) {
-			Subquery<ConceptMap> sub = cq.subquery(ConceptMap.class);
-			Root<ConceptMap> mapRoot = sub.from(ConceptMap.class);
-			Join<ConceptMap, ConceptReferenceTerm> termJoin = mapRoot.join("conceptReferenceTerm");
-			Join<ConceptReferenceTerm, ConceptSource> sourceJoin = termJoin.join("conceptSource");
+		if (criteria.getMappings() != null && !criteria.getMappings().isEmpty()) {
+			boolean caseSensitive = Context.getAdministrationService().isDatabaseStringComparisonCaseSensitive();
+			for (String mapping : criteria.getMappings()) {
+				// mapping is assumed to be in the form "source:code"
+				int idx = mapping.indexOf(":");
+				if (idx < 0 || idx >= mapping.length() - 1) {
+					continue;
+				}
+				String sourceName = mapping.substring(0, idx);
+				String code = mapping.substring(idx + 1);
 
-			List<Predicate> mappingPredicates = new ArrayList<>();
-			mappingPredicates.add(cb.equal(mapRoot.get("concept"), root));
+				Subquery<ConceptMap> sub = cq.subquery(ConceptMap.class);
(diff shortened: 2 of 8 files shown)


api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -96,6 +97,19 @@ public interface ConceptService extends OpenmrsService {
...
 
+	/**
+	 * Search for concepts matching the given criteria. The identifier fields ({@code uuids},
+	 * {@code conceptIds}, and {@code mappingCode}/{@code mappingSourceName}) are combined with OR
+	 * semantics — a concept matching any one of them is included. Global filter fields such as
+	 * {@code includeRetired} are then applied with AND semantics on top of those results.
+	 *
+	 * @param conceptSearchCriteria the search criteria
+	 * @return list of matching concepts
+	 * @since 2.8.0
```
ibacher: This isn't accurate, since where on 2.8. something higher than 0. Should be something like `2.8.x, 2.9.0, 3.0.0`

api/src/main/java/org/openmrs/parameter/ConceptSearchCriteria.java
```diff
@@ -0,0 +1,135 @@
...
+package org.openmrs.parameter;
+
+import java.util.Collection;
+
+/**
+ * The ConceptSearchCriteria class encapsulates the criteria used for searching concepts in the
+ * system. It supports searching by UUIDs, integer IDs, and concept mapping (code + source name).
+ * Multiple criteria are combined with OR semantics — a concept matching any one criterion is
+ * returned. The {@code includeRetired} flag is applied as a global AND filter on top of those
+ * results.
```
ibacher: I might not specify what the search semantics are on this class, since that's implemented elsewhere and there's no reason we couldn't reuse these for a scenario where ANDing things made sense.

api/src/main/java/org/openmrs/parameter/ConceptSearchCriteria.java
```diff
@@ -0,0 +1,135 @@
...
+ * <p>
+ * Instead of calling the constructor directly, use {@link ConceptSearchCriteriaBuilder}.
+ *
+ * @since 2.8.0
```
ibacher: Same note about since

api/src/main/java/org/openmrs/parameter/ConceptSearchCriteria.java
```diff
@@ -0,0 +1,135 @@
...
+	 * @param mappingSourceName the mapping source name to set.
+	 */
+	public void setMappingSourceName(String mappingSourceName) {
+		this.mappingSourceName = mappingSourceName;
+	}
+
+	/**
+	 * @return true if retired concepts are included in the search results, false otherwise.
+	 */
+	public boolean isIncludeRetired() {
```
ibacher: I thought as a rule we were moving away from the `is...` convention? Maybe I'm misremembering.

api/src/main/java/org/openmrs/parameter/ConceptSearchCriteriaBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+public class ConceptSearchCriteriaBuilder {
+
+	private Collection<String> uuids;
+
+	private Collection<Integer> conceptIds;
+
+	private String mappingCode;
+
+	private String mappingSourceName;
```
ibacher: These should be collections as well I think since we need to submit multiple of them in one go. It might make sense to make a tiny class to hold these two files so we can have a collection of those?

ibacher (changes requested):
Thanks @chibongho! A few small points to address, but basically looks ok.

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -96,6 +97,19 @@ public interface ConceptService extends OpenmrsService {
...
+	 */
+	@Authorized(PrivilegeConstants.GET_CONCEPTS)
+	public List<Concept> getConcepts(ConceptSearchCriteria conceptSearchCriteria);
```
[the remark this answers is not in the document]
ibacher: We don't expect implementations to supply their own services.

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -1067,6 +1068,15 @@ public Concept getConceptByUuid(String uuid) {
...
 	}
 
+	/**
+	 * @see org.openmrs.api.ConceptService#getConcepts(ConceptSearchCriteria)
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	public List<Concept> getConcepts(ConceptSearchCriteria conceptSearchCriteria) {
+		return dao.getConcepts(conceptSearchCriteria);
+	}
```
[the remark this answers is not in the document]
ibacher: Conventionally, we handle these kinds of issues at the DAO layer, but I think the point about the NPE and mapping filter should be addressed.

api/src/main/java/org/openmrs/parameter/ConceptSearchCriteriaBuilder.java
mseaton: I think the general OpenMRS convention has been to combine these together into a single string in the form of "source:term", so maybe we can just support a `Collection<String> references` (which could refer to either uuids or source:term), or if we want to limit this to mappings, have a property named `Collection<String> mappings`?

chibongho: I think having the mappings in the form "source:term" makes sense.

I also just noticed that [`ConceptServiceImpl.getConceptByReference(String ref)`](https://github.com/openmrs/openmrs-core/blob/9e65f5f3e819c02a712ad4a241b922607420452e/api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java#L360) is a utility-knife function that can take multiple formats for `ref` (id, uuid, mapping in the form "source:term", or a fully qualified name of a java String constant whose value is either one of these 4). I think I'll lift that logic to the Builder class here, so that it normalizes the `ref` into one of these Collections.

api/src/main/java/org/openmrs/api/ConceptService.java
chibongho: I'll change it to `2.9.0`

api/src/main/java/org/openmrs/parameter/ConceptSearchCriteria.java
chibongho: I'll move the ORing logic explanation to the `ConceptService.getConcepts()` function instead

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
```diff
@@ -1143,6 +1152,110 @@ public Concept getConceptByUuid(String uuid) {
...
+		predicates.add(cb.equal(mapRoot.get("concept"), root));
+		if (caseSensitive) {
+			predicates.add(cb.equal(cb.lower(termJoin.get("code")), code.toLowerCase()));
+			predicates.add(cb.or(cb.equal(cb.lower(sourceJoin.get("name")), sourceName.toLowerCase()),
+			    cb.equal(cb.lower(sourceJoin.get(HL7_CODE)), sourceName.toLowerCase())));
+		} else {
+			predicates.add(cb.equal(termJoin.get("code"), code));
+			predicates.add(
+			    cb.or(cb.equal(sourceJoin.get("name"), sourceName), cb.equal(sourceJoin.get(HL7_CODE), sourceName)));
+		}
```
mseaton, line 1235: One question here is whether or not we should require these to only be same-as mappings?  Whatever the existing code does should be matched.

api/src/main/java/org/openmrs/parameter/ConceptSearchCriteriaBuilder.java
```diff
@@ -0,0 +1,241 @@
...
+	/**
+	 * Adds multiple concept UUIDs to the search criteria.
+	 *
+	 * @param uuids the collection of UUIDs to include.
+	 * @return the current builder instance for method chaining.
+	 */
+	public ConceptSearchCriteriaBuilder addUuids(Collection<String> uuids) {
+		this.uuids.addAll(uuids);
+		return this;
+	}
```
mseaton: Having both of these methods seems unnecessary?  Why not just have one that can take in a String... uuid?

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
```diff
@@ -360,47 +360,10 @@ public Concept getConceptByReference(String conceptRef) {
...
-		if (conceptRef.contains(".")) {
-			try {
-				return getConceptByReference(evaluateStaticConstant(conceptRef));
-			} catch (APIException e) {
-				log.warn("Unable to translate '{}' into a concept", conceptRef, e);
-			}
-		}
-		return cpt == null ? null : cpt;
+		ConceptSearchCriteria criteria = new ConceptSearchCriteriaBuilder().addConceptReference(conceptRef)
+		        .includeRetired(true).build();
```
mseaton, line 364: Not really 100% sure about including retired here.  If that is what the existing code did, then fine, but please double check

mseaton (approved):
Generally looks good, though I do have a few comments

api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java
chibongho: The existing `getConceptsByMapping` and `createSearchConceptMapCriteria` functions match on any ConceptMap, not just 'same-as'

api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java
chibongho: The old code calls either `getConceptByUuid()`, `getConceptByMapping(code, source)`, `getConcept()` or `getConceptByName()`; none of them filter retired concepts

api/src/main/java/org/openmrs/parameter/ConceptSearchCriteriaBuilder.java
chibongho: switched to `addUuids(String... uuids)` and `addConceptId(Integer... conceptIds)`

PR #6084, opened 2026-05-07
TRUNK-6429: Create application events for service method calls and en…

rkorytkowski:
…tity changes

## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/aop/OpenmrsServiceEventAdvice.java
+++ b/api/src/main/java/org/openmrs/aop/OpenmrsServiceEventAdvice.java
@@ -1,0 +1,144 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.aop;
+
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import org.openmrs.OpenmrsObject;
+import org.openmrs.aop.event.PurgeServiceEvent;
+import org.openmrs.aop.event.RetireServiceEvent;
+import org.openmrs.aop.event.SaveServiceEvent;
+import org.openmrs.aop.event.UnretireServiceEvent;
+import org.openmrs.aop.event.UnvoidServiceEvent;
+import org.openmrs.aop.event.VoidServiceEvent;
+import org.openmrs.event.EventPublisher;
+import org.springframework.aop.support.AopUtils;
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+import java.util.Optional;
+
+/**
+ * Publishes {@link SaveServiceEvent}, {@link VoidServiceEvent}, {@link UnvoidServiceEvent},
+ * {@link RetireServiceEvent}, and {@link UnretireServiceEvent}.
+ * <p>
+ * It is registered after transaction advice so that {@link org.openmrs.event.outbox.OutboxEventListener},
+ * {@link org.springframework.transaction.event.TransactionalEventListener} and 
+ * {@link org.springframework.context.event.EventListener}
+ * can run in the same transaction.
+ * 
+ * @since 2.9.x
+ */
+@Aspect
+@Order(50) // Guarantees execution after transaction advice
+@Component
+public class OpenmrsServiceEventAdvice {
+
+	private final EventPublisher eventPublisher;
+	
+	public OpenmrsServiceEventAdvice(EventPublisher eventPublisher) {
+		this.eventPublisher = eventPublisher;
+	}
+	
+    /**
+     * This intercepts method executions starting with "save" on OpenmrsService.
+     */
+    @Around("(execution(* *.save*(..)) || execution(* *.create*(..))) && target(org.openmrs.api.OpenmrsService)")
+    public Object interceptSaveOrCreate(ProceedingJoinPoint pjp) throws Throwable {
+		// Do not apply to proxies created by TransactionProxyFactoryBean (legacy) in addition to transactional advice so
+		// that events are not emitted twice.
+		if (AopUtils.isAopProxy(pjp.getTarget())) {
+			return pjp.proceed();
+		}
+
+		getOpenmrsObjectArgument(pjp)
+			.ifPresent(object -> eventPublisher.publishEvent(new SaveServiceEvent<>(object)));
+        return pjp.proceed();
+    }
+
+	/**
+	 * This intercepts method executions starting with "void" on OpenmrsService.
+	 */
+	@Around("execution(* *.void*(..)) && target(org.openmrs.api.OpenmrsService)")
+	public Object interceptVoid(ProceedingJoinPoint pjp) throws Throwable {
+		if (AopUtils.isAopProxy(pjp.getTarget())) {
+			return pjp.proceed();
+		}
+
+		getOpenmrsObjectArgument(pjp)
+			.ifPresent(object -> eventPublisher.publishEvent(new VoidServiceEvent<>(object)));
+		return pjp.proceed();
+	}
+
+	/**
+	 * This intercepts method executions starting with "unvoid" on OpenmrsService.
+	 */
+	@Around("execution(* *.unvoid*(..)) && target(org.openmrs.api.OpenmrsService)")
+	public Object interceptUnvoid(ProceedingJoinPoint pjp) throws Throwable {
+		if (AopUtils.isAopProxy(pjp.getTarget())) {
+			return pjp.proceed();
+		}
+
+		getOpenmrsObjectArgument(pjp)
+			.ifPresent(object -> eventPublisher.publishEvent(new UnvoidServiceEvent<>(object)));
+		return pjp.proceed();
+	}
+
+	/**
+	 * This intercepts method executions starting with "retire" on OpenmrsService.
+	 */
+	@Around("execution(* *.retire*(..)) && target(org.openmrs.api.OpenmrsService)")
+	public Object interceptRetire(ProceedingJoinPoint pjp) throws Throwable {
+		if (AopUtils.isAopProxy(pjp.getTarget())) {
+			return pjp.proceed();
+		}
+
+		getOpenmrsObjectArgument(pjp)
+			.ifPresent(object -> eventPublisher.publishEvent(new RetireServiceEvent<>(object)));
+		return pjp.proceed();
+	}
+
+	/**
+	 * This intercepts method executions starting with "unretire" on OpenmrsService.
+	 */
+	@Around("execution(* *.unretire*(..)) && target(org.openmrs.api.OpenmrsService)")
+	public Object interceptUnretire(ProceedingJoinPoint pjp) throws Throwable {
+		if (AopUtils.isAopProxy(pjp.getTarget())) {
+			return pjp.proceed();
+		}
+
+		getOpenmrsObjectArgument(pjp)
+			.ifPresent(object -> eventPublisher.publishEvent(new UnretireServiceEvent<>(object)));
+		return pjp.proceed();
+	}
+
+	/**
+	 * This intercepts method executions starting with "purge" on OpenmrsService.
+	 */
+	@Around("execution(* *.purge*(..)) && target(org.openmrs.api.OpenmrsService)")
+	public Object interceptPurge(ProceedingJoinPoint pjp) throws Throwable {
+		if (AopUtils.isAopProxy(pjp.getTarget())) {
+			return pjp.proceed();
+		}
+
+		getOpenmrsObjectArgument(pjp)
+			.ifPresent(object -> eventPublisher.publishEvent(new PurgeServiceEvent<>(object)));
+		return pjp.proceed();
+	}
+
+	private Optional<OpenmrsObject> getOpenmrsObjectArgument(ProceedingJoinPoint pjp) {
+		if (pjp.getArgs() != null && pjp.getArgs().length > 0 && pjp.getArgs()[0] instanceof OpenmrsObject) {
+			return Optional.of((OpenmrsObject) pjp.getArgs()[0]);
+		} else {
+			return Optional.empty();
+		}
+	}
+}
--- a/api/src/main/java/org/openmrs/aop/event/RetireServiceEvent.java
+++ b/api/src/main/java/org/openmrs/aop/event/RetireServiceEvent.java
@@ -1,0 +1,36 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.aop.event;
+
+import org.openmrs.event.EntityEvent;
+
+import java.util.Set;
+
+/**
+ * Published by {@link org.openmrs.aop.OpenmrsServiceEventAdvice}
+ * 
+ * @param <T> the changed entity
+ *     
+ * @since 2.9.x
+ */
+public class RetireServiceEvent<T> extends EntityEvent<T> {
+	private static final long serialVersionUID = 1L;
+	
+	public RetireServiceEvent() {
+	}
+	
+	public RetireServiceEvent(T entity) {
+		super(entity);
+	}
+
+	public RetireServiceEvent(T entity, Set<String> tags) {
+		super(entity, tags);
+	}
+}
--- a/api/src/main/java/org/openmrs/aop/event/SaveServiceEvent.java
+++ b/api/src/main/java/org/openmrs/aop/event/SaveServiceEvent.java
@@ -1,0 +1,36 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.aop.event;
+
+import org.openmrs.event.EntityEvent;
+
+import java.util.Set;
+
+/**
+ * Published by {@link org.openmrs.aop.OpenmrsServiceEventAdvice}
+ *
+ * @param <T> the changed entity
+ *
+ * @since 2.9.x
+ */
+public class SaveServiceEvent<T> extends EntityEvent<T> {
+	private static final long serialVersionUID = 1L;
+	
+	public SaveServiceEvent() {
+	}
+	
+	public SaveServiceEvent(T entity) {
+		super(entity);
+	}
+
+	public SaveServiceEvent(T entity, Set<String> tags) {
+		super(entity, tags);
+	}
+}
--- a/api/src/main/java/org/openmrs/api/db/event/DeleteDbEvent.java
+++ b/api/src/main/java/org/openmrs/api/db/event/DeleteDbEvent.java
@@ -1,0 +1,54 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.db.event;
+
+import org.openmrs.event.EntityEvent;
+
+import java.io.Serializable;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * DB level event emitted upon deletion of an entity.
+ *
+ * @param <T> type of entity
+ */
+public class DeleteDbEvent<T> extends EntityEvent<T> {
+	
+	private Serializable id;
+	private Object[] state;
+	private String[] propertyNames;
+	
+	public DeleteDbEvent() {
+	}
+	
+	public DeleteDbEvent(T entity, Serializable id, Object[] state, String[] propertyNames) {
+		this(entity, id, state, propertyNames, new HashSet<>());
+	}
+
+	public DeleteDbEvent(T entity, Serializable id, Object[] state, String[] propertyNames, Set<String> tags) {
+		super(entity, tags);
+		this.state = state;
+		this.propertyNames = propertyNames;
+		this.id = id;
+	}
+
+	public Serializable getId() {
+		return id;
+	}
+
+	public Object[] getState() {
+		return state;
+	}
+
+	public String[] getPropertyNames() {
+		return propertyNames;
+	}
+}
--- a/api/src/main/java/org/openmrs/event/EntityEvent.java
+++ b/api/src/main/java/org/openmrs/event/EntityEvent.java
@@ -1,0 +1,70 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.event;
+
+import org.jspecify.annotations.Nullable;
+import org.springframework.core.ResolvableType;
+import org.springframework.core.ResolvableTypeProvider;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * It can be implemented to publish application events related to an entity.
+ * <p>
+ * Entity events may be put in a transactional outbox. If there is a listener annotated with
+ * {@link org.openmrs.event.outbox.OutboxEventListener} for an event, the event needs to be 
+ * serializable to JSON with Jackson {@link com.fasterxml.jackson.databind.ObjectMapper} or
+ * implement {@link org.openmrs.event.outbox.OutboxEventPayload}.
+ * <p>
+ * Entity events contain autopopulated sessionId to group events happening in the same session.
+ * <p>
+ * They may also contain tags to use them e.g. as listener conditions 
+ * or include additional information for listeners.
+ * 
+ * @param <T> the related entity
+ *     
+ * @since 2.9.x
+ */
+	public abstract class EntityEvent<T> extends BaseEvent implements ResolvableTypeProvider {
+
+	protected T entity;
+	
+	public EntityEvent() {
+	}
+	
+	public EntityEvent(T entity) {
+		this(entity, new HashSet<>());
+	}
+ 	
+	public EntityEvent(T entity, Set<String> tags) {
+		super(entity, tags);
+		this.entity = entity;
+	}
+
+	public T getEntity() {
+		return entity;
+	}
+	
+	public void setEntity(T entity) {
+		this.entity = entity;
+		this.source = entity;
+	}
+
+	@Override
+	public @Nullable ResolvableType getResolvableType() {
+		if (entity != null && getClass().getTypeParameters().length == 1) {
+			return ResolvableType.forClassWithGenerics(getClass(), ResolvableType.forInstance(entity));
+		} else {
+			return ResolvableType.forClass(getClass());
+		}
+	}
+	
+}
--- a/api/src/main/java/org/openmrs/event/EventPublisher.java
+++ b/api/src/main/java/org/openmrs/event/EventPublisher.java
@@ -1,0 +1,50 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.event;
+
+import org.hibernate.SessionFactory;
+import org.hibernate.engine.spi.SessionImplementor;
+import org.jspecify.annotations.NonNull;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.context.annotation.Primary;
+import org.springframework.stereotype.Component;
+
+/**
+ * OpenMRS custom {@link ApplicationEventPublisher}, which populates
+ * sessionId in order to be able to group events happening in the same session.
+ * <p>
+ * It can be autowired as {@link EventPublisher} or Spring's {@link ApplicationEventPublisher}.
+ * 
+ * @since 2.9.x
+ */
+@Primary
+@Component
+public class EventPublisher implements ApplicationEventPublisher {
+	private final ApplicationEventPublisher delegate;
+	private final SessionFactory sessionFactory;
+	
+	public EventPublisher(ApplicationEventPublisher delegate, SessionFactory sessionFactory) {
+		this.delegate = delegate;
+		this.sessionFactory = sessionFactory;
+	}
+
+	@Override
+	public void publishEvent(@NonNull Object event) {
+		if (event instanceof BaseEvent) {
+			String sessionId = ((SessionImplementor) sessionFactory.getCurrentSession()).getSessionIdentifier().toString();
+			BaseEvent baseEvent = (BaseEvent) event;
+			baseEvent.setSessionId(sessionId);
+		}
+		delegate.publishEvent(event);
+	}
+
+	
+}
+
--- a/api/src/main/java/org/openmrs/event/TransactionalEventAggregator.java
+++ b/api/src/main/java/org/openmrs/event/TransactionalEventAggregator.java
@@ -1,0 +1,115 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.event;
+
+import org.openmrs.api.context.Context;
+import org.springframework.context.event.EventListener;
+import org.springframework.core.Ordered;
+import org.springframework.transaction.event.TransactionalEventListener;
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Extend to aggregate events for specific event types that happen in a single transaction.
+ * Please note that if transaction is not active, events are not aggregated.
+ * <p>
+ * It may be used to create a single event for a number of entity changes.
+ * <p>
+ * It is recommended for each aggregator to produce a custom type extending {@link AggregatedEntityEvent}.
+ * <p>
+ * The event can be listened to with {@link org.springframework.context.event.EventListener}, 
+ * {@link TransactionalEventListener} (<b>except for <i>TransactionPhase.BEFORE_COMMIT</i></b>) 
+ * or {@link org.openmrs.event.outbox.OutboxEventListener}.
+ * <p>
+ * Since this event is emitted in the BEFORE_COMMIT phase, any <i>@EventListener</i> for this event is executed in 
+ * the BEFORE_COMMIT phase as well. 
+ * <p>
+ * <b>It is not supported to use <i>@TransactionalEventListener(phase = TransactionPhase.BEFORE_COMMIT)</i> 
+ * to listen for an aggregated event, use <i>@EventListener</i> instead.</b>
+ * 
+ * @since 2.9.x
+ */
+public abstract class TransactionalEventAggregator {
+
+	private final EventPublisher eventPublisher;
+	
+	public TransactionalEventAggregator(EventPublisher eventPublisher) {
+		this.eventPublisher = eventPublisher;
+	}
+
+	/**
+	 * It must be very fast as it is passed all EntityEvents.
+	 * 
+	 * @param event the event
+	 * @return true if event should be aggregated
+	 */
+	public abstract boolean supportsEvent(EntityEvent<?> event);
+
+	/**
+	 * Returns a concrete instance of {@link AggregatedEntityEvent} to be published for listeners.
+	 * 
+	 * @param events the aggregated events
+	 * @return a new aggregated event
+	 */
+	public abstract AggregatedEntityEvent newAggregatedEntityEvent(List<EntityEvent<?>> events);
+	
+	/**
+	 * Adds event to a group associated with a current transaction.
+	 * 
+	 * @param event the event
+	 */
+	@EventListener
+	public void aggregateEvent(EntityEvent<?> event) {
+		if (!TransactionSynchronizationManager.isSynchronizationActive() || !supportsEvent(event)) {
+			return;
+		}
+		
+		// Get or create the list of events for this transaction.
+		@SuppressWarnings("unchecked") 
+		List<EntityEvent<?>> events = (List<EntityEvent<?>>) TransactionSynchronizationManager.getResource(this);
+		if (events == null) {
+			events = new ArrayList<>();
+			TransactionSynchronizationManager.bindResource(this, events);
+			registerTransactionSynchronization();
+		}
+		events.add(event);
+	}
+
+	private void registerTransactionSynchronization() {
+		// Register a one-time synchronization to flush the events at the end of the commit phase.
+		TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
+
+			@Override
+			public int getOrder() {
+				return Ordered.HIGHEST_PRECEDENCE; // Run before any other TransactionalEventListeners
+			}
+
+			@Override
+			public void beforeCommit(boolean readOnly) {
+				Context.flushSession(); // Make sure that all DB events are published before this event.
+				@SuppressWarnings("unchecked") 
+				List<EntityEvent<?>> events = (List<EntityEvent<?>>) TransactionSynchronizationManager.getResource(TransactionalEventAggregator.this);
+				if (events != null && !events.isEmpty()) {
+					eventPublisher.publishEvent(newAggregatedEntityEvent(new ArrayList<>(events)));
+					events.clear();
+				}
+			}
+
+			@Override
+			public void afterCompletion(int status) {
+				// Clean up the resource regardless of commit or rollback.
+				TransactionSynchronizationManager.unbindResourceIfPossible(TransactionalEventAggregator.this);
+			}
+		});
+	}
+}
--- a/api/src/main/java/org/openmrs/event/outbox/OutboxEvent.java
+++ b/api/src/main/java/org/openmrs/event/outbox/OutboxEvent.java
@@ -1,0 +1,184 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.event.outbox;
+
+import org.openmrs.Auditable;
+import org.openmrs.BaseOpenmrsObject;
+import org.openmrs.User;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.EnumType;
+import javax.persistence.Enumerated;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.Lob;
+import javax.persistence.ManyToOne;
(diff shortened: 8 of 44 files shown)


api/src/main/java/org/openmrs/event/outbox/OutboxEventInterceptor.java
```diff
@@ -0,0 +1,50 @@
...
+								  ObjectMapper objectMapper) {
+		this.sessionFactory = sessionFactory;
+		this.objectMapper = objectMapper; 
+	}
+
+	@TransactionalEventListener(
+			phase = TransactionPhase.BEFORE_COMMIT,
+			condition = "@outboxEventRegistry.hasOutboxListeners(#event.getClass())"
+	)
+	public void interceptAndSaveToOutbox(OutboxableEvent event) {
```
rkorytkowski: Even though we make it listen for possibly a lot of events in the system we have the condition in place so that it's not even fired by Spring if there's no `@OutboxEventListener` registered for a particular event.

api/src/main/java/org/openmrs/event/outbox/OutboxEventInterceptor.java
```diff
@@ -0,0 +1,50 @@
...
+@Component
+public class OutboxEventInterceptor {
+
+	private static final Logger log = LoggerFactory.getLogger(OutboxEventInterceptor.class);
+	
+	private final DbSessionFactory sessionFactory;
+	private final ObjectMapper objectMapper;
+
+	public OutboxEventInterceptor(DbSessionFactory sessionFactory, 
+								  ObjectMapper objectMapper) {
```
rkorytkowski: We use system defined objectMapper to serialize events to json with Jackson.

rkorytkowski:
It's only for API review and business logic. Not ready for a proper code-review.

rkorytkowski:
See https://talk.openmrs.org/t/revamping-bahmnis-event-driven-architecture/48581/26

api/src/main/java/org/openmrs/event/outbox/OutboxEventRegistry.java
```diff
@@ -0,0 +1,101 @@
...
+					}
+				});
+			}
+		}
+
+		// Sort all listeners once at application startup
+		Collections.sort(registry);
+	}
+
+	public boolean hasOutboxListeners(Class<?> eventType) {
```
rkorytkowski: This method needs to be super fast as it will be called for almost all events in the system (implementing OutboxableEvent including EntityEvent).

api/src/main/java/org/openmrs/event/EventPublisher.java
```diff
@@ -0,0 +1,32 @@
...
+import org.hibernate.engine.spi.SessionImplementor;
+import org.jspecify.annotations.NonNull;
+import org.openmrs.api.db.hibernate.DbSessionFactory;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.context.annotation.Primary;
+import org.springframework.stereotype.Component;
+
+@Primary
+@Component
+public class EventPublisher implements ApplicationEventPublisher {
```
rkorytkowski, line 29: It will be injected instead of a default Spring Event publisher to include sessionId for EntityEvents.

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxPollingTaskHandler.java
```diff
@@ -0,0 +1,142 @@
...
+				Query revertQuery = session.createQuery(
+						"update OutboxEvent set status = 'PENDING', errorCount = :errorCount, errorMessage = :errorMessage, dateUpdated = :now where id = :id");
+				revertQuery.setParameter("errorCount", errorCount);
+				revertQuery.setParameter("errorMessage", errorMessage);
+				revertQuery.setParameter("id", item.getId());
+				revertQuery.setParameter("now", new Date());
+				revertQuery.executeUpdate();
+
+				// Stop processing subsequent events to ensure strict ordering
+				throw new OutboxException("Failed to process outbox item ID: " + item.getId(), e);
```
rkorytkowski: In case of a failure, the task will be re-rerun by the scheduler 3 times. The error will be persisted in scheduler as well and can be inspected though its UI. The recurrent task will continue to be scheduled every 15 seconds. The outbox will continue to grow, but it will not be processed until the issue is fixed. For now a manual action needs to be taken by changing the status of the event to ERROR to completely ignore the failing event or by deleting it entirely.

api/src/main/java/org/openmrs/event/outbox/OutboxEventInterceptor.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.transaction.event.TransactionalEventListener;
+
+import java.util.Date;
+
+@Component
+public class OutboxEventInterceptor {
+
+	private static final Logger log = LoggerFactory.getLogger(OutboxEventInterceptor.class);
+	
+	private final DbSessionFactory sessionFactory;
```
ibacher: Small nit: shouldn't we just use the SessionFactory here? `DbSessionFactory`, as I understand it, was added for backwards compatibility with 1.x.

api/src/main/java/org/openmrs/event/outbox/OutboxEventRegistry.java
```diff
@@ -0,0 +1,101 @@
...
+		// Make sure to ignore outbox processing events so we do not end up in infinite loop
+		if (ignoredEvents.stream().anyMatch(ignoredEvent -> ignoredEvent.isAssignableFrom(eventType))) {
+			return false;
+		}
+		// Checks if the published event type (or its subclass) is registered for the outbox
+		return registry.stream()
```
ibacher: Since, as you say, this needs to be super-fast, it might make sense to, e.g., memoize the results with a `WeakHashMap`. That way we only need to run the ignored event check and listener check once per eventType.

api/src/main/java/org/openmrs/event/outbox/OutboxEventRegistry.java
```diff
@@ -0,0 +1,101 @@
...
+	private final List<ListenerMethod> registry = new ArrayList<>();
+	private final List<Class<?>> ignoredEvents = Arrays.asList(OutboxDispatchEvent.class, OutboxExceptionEvent.class);
+
+	public OutboxEventRegistry(ApplicationContext applicationContext) {
+		this.applicationContext = applicationContext;
+	}
+
+	@Override
+	public void afterSingletonsInstantiated() {
+		for (String beanName : applicationContext.getBeanDefinitionNames()) {
```
ibacher, line 62: For context refreshes, it probably makes sense for this method to clear the registry?

api/src/main/java/org/openmrs/event/outbox/OutboxEventInterceptor.java
```diff
@@ -0,0 +1,50 @@
...
+
+	@TransactionalEventListener(
+			phase = TransactionPhase.BEFORE_COMMIT,
+			condition = "@outboxEventRegistry.hasOutboxListeners(#event.getClass())"
+	)
+	public void interceptAndSaveToOutbox(OutboxableEvent event) {
+		try {
+			OutboxEvent item = new OutboxEvent();
+			item.setEventType(event.getClass().getName());
+			item.setPayload(objectMapper.writeValueAsString(event));
```
ibacher: Does this mean we need to ensure that every element saved is JSON serializable?

ibacher (commented):
Couple of remarks

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxTaskSchedulerInitializer.java
```diff
@@ -0,0 +1,51 @@
...
+				schedulerService.scheduleRecurrently(
+					OUTBOX_CLEANUP_TASK_NAME,
+					new OutboxCleanupTaskData(),
+					Duration.ofDays(1)
+				);
+			}
+		} finally {
+			Context.removeProxyPrivilege(PrivilegeConstants.MANAGE_SCHEDULER);
+			Context.closeSession();
+		}
```
rkorytkowski, line 72: Maybe we should add a utility method to the SchedulerService to be able to easily schedule tasks on startup as daemon user from core and modules without having to deal with session and privileges and checking for existence with optional overwrite.

api/src/main/java/org/openmrs/event/outbox/OutboxEventInterceptor.java
rkorytkowski: Right!

api/src/main/java/org/openmrs/event/outbox/OutboxEventInterceptor.java
rkorytkowski: If you want to put the event in the outbox, then yes, it must be JSON serializable. If we want to provide a way to use other serialization mechanism then we can add OutboxEventPayload interface with String toPayload() and void fromPayload(String) to optionally provide a custom mechanism for an event. An event would then implement OutboxEventPayload.

api/src/main/java/org/openmrs/event/outbox/OutboxEventRegistry.java
rkorytkowski: It's a spring bean singleton itself so it will be recreated upon context refresh.

api/src/main/java/org/openmrs/event/outbox/OutboxEventRegistry.java
rkorytkowski: Yes, probably caching is needed.

api/src/main/java/org/openmrs/event/EntityEvent.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.openmrs.event.outbox.OutboxableEvent;
+import org.springframework.context.ApplicationEvent;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public abstract class EntityEvent<T> extends ApplicationEvent implements OutboxableEvent {
+	
+	private final T entity;
```
rkorytkowski: Confirmed that generics are preserved by Spring Events and events are handled efficiently.

api/src/main/java/org/openmrs/event/EventPublisher.java
```diff
@@ -0,0 +1,32 @@
...
+		this.delegate = delegate;
+		this.sessionFactory = sessionFactory;
+	}
+
+	@Override
+	public void publishEvent(@NonNull Object event) {
+		if (event instanceof EntityEvent) {
+			String sessionId = ((SessionImplementor) sessionFactory.getHibernateSessionFactory().getCurrentSession()).getSessionIdentifier().toString();
+			EntityEvent<?> entityEvent = (EntityEvent<?>) event;
+			entityEvent.setSessionId(sessionId);
```
rkorytkowski: @mseaton We can't get a hold of a transaction id in a universal way (different for each rdbms), thus using session id, which is usually scoped to a single http request in OpenMRS. I think it's a good enough identifier for grouping related operations.

api/src/test/java/org/openmrs/event/outbox/TestEventPublisherService.java
```diff
@@ -0,0 +1,41 @@
...
+		this.sessionFactory = sessionFactory;
+	}
+	
+	@Transactional
+	public void clearOutbox() {
+		sessionFactory.getCurrentSession().createQuery("delete from OutboxEvent").executeUpdate();
+	}
+
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
+	public void publishEvent(OutboxableEvent event) {
```
rkorytkowski: propagation = Propagation.REQUIRES_NEW is not really needed here.

api/src/main/java/org/openmrs/aop/RequiredDataAdvice.java
```diff
@@ -127,6 +135,7 @@ public void before(Method method, Object[] args, Object target) throws Throwable
 				
 				recursivelyHandle(SaveHandler.class, (OpenmrsObject) mainArgument, other);
 				ValidateUtil.validate(mainArgument);
+				eventPublisher.publishEvent(new SaveServiceEvent<>(mainArgument));
```
rkorytkowski: Event though this is resolved at compile time to `new SaveServiceEvent<Object>(mainArgument)` it is correctly handled by Spring Events framework that at runtime asks for a specific class of mainArgument and preserves it so that you can register a specific listener for `SaveServiceEvent<Concept>` for example.

api/src/main/java/org/openmrs/event/outbox/OutboxEvent.java
```diff
@@ -0,0 +1,147 @@
...
+import javax.persistence.Table;
+import java.util.Date;
+
+import org.openmrs.Auditable;
+import org.openmrs.BaseOpenmrsObject;
+import org.openmrs.User;
+
+@Entity
+@Table(name = "event_outbox")
+public class OutboxEvent extends BaseOpenmrsObject implements Auditable {
```
rkorytkowski, line 38: I haven't added a regular OpenMRS service for the OutboxEvent as I think it should remain internal. There's Spring Modulith that provides a similar functionality that we may want to adopt at some point once we upgrade to Spring Boot. Any changes will be easier if not relied on by modules.

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxPollingTaskHandler.java
```diff
@@ -0,0 +1,145 @@
...
+				if (OutboxEventPayload.class.isAssignableFrom(eventClass)) {
+					event = eventClass.getDeclaredConstructor().newInstance();
+					((OutboxEventPayload) event).fromPayload(item.getPayload());
+				} else {
+					event = objectMapper.readValue(item.getPayload(), eventClass);
+				}
+				
+				// Dispatch ONLY to registered outbox listeners, bypassing the ApplicationEventPublisher
+				// If one of listeners fails, the event will be re-tried for all listeners
+				// Each listener runs in a new transaction so it doesn't affect the current one
```
rkorytkowski: I think I will change this so that we do not re-try listeners that succeeded. It will be easier to handle from here than from a listener itself, but I'd appreciate your thoughts.

The only thing is that without re-try you can not rely entirely on some listener to run before yours (e.g. to modify the event in some way). I think it should be discouraged anyway even though we support ordering of listeners as Spring event does.

rkorytkowski (commented):
Some general comments.

api/src/main/java/org/openmrs/event/outbox/OutboxEventRegistry.java
ibacher: 🤦  I accidentally made up a `static` in the registry declaration.

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxPollingTaskHandler.java
ibacher: I don't think we need to support chained event listeners. If you need that functionality, it should be easy enough to wire-up a listener that publishes to Kafka / Flink / Camel / w/e and do the streaming processing that way.

api/src/main/java/org/openmrs/event/EventPublisher.java
mseaton: Wouldn't it be more accurate to use the same sort of approach we are using in the event module to keep a dequeue of session ids for transactions - just populated with new uuids on transaction begin and removed on transaction completion (successful or unsuccessful)?

api/src/main/java/org/openmrs/event/outbox/OutboxEventInterceptor.java
rkorytkowski: I changed the listener from TransactionalEventListener(phase = BEFORE_COMMIT) to a simple EventListener. The events are now persisted in the same transaction in the very moment they happen and not at the end of transaction. It's more efficient as we don't have to store them in memory. If a transaction is rolled back, events are rolled back as well. 
After this change the condition is faster to be executed inside the method as no additional SpEL processing is involved.

api/src/main/java/org/openmrs/scheduler/Mutex.java
```diff
@@ -0,0 +1,47 @@
...
+ * <p>
+ * Remember to always use a unique name to avoid conflicts with other executions in the system. 
+ * Use `yourmoduleid.` prefix for executions defined in modules.
+ *
+ * @since 2.9.x
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@SchedulerLock
+public @interface Mutex {
```
rkorytkowski, line 37: Eventually I haven't used it for outbox events, because I wanted to have more control over the lock, but left it to be used by others if needed instead of `@SchedulerLock`.

api/src/test/java/org/openmrs/event/outbox/TestTransactionalEventAggregator.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.stereotype.Component;
+
+import java.util.List;
+
+/**
+ * Aggregates events <i>SaveServiceEvent&lt;GlobalProperty&gt;</i> and <i>SaveDbEvent&lt;GlobalProperty&gt;</i>
+ * into {@link TestAggregatedEntityEvent}.
+ */
+@Component
+public class TestTransactionalEventAggregator extends TransactionalEventAggregator {
```
rkorytkowski, line 29: @mseaton this is how you can implement an event aggregator to gather specific events that happen in a single transaction and have them published as a single event.

I this case I'm interested in `SaveServiceEvent<GlobalProperty> and SaveDbEvent<GlobalProperty>`.

api/src/test/java/org/openmrs/event/outbox/TestOutboxEventListener.java
```diff
@@ -0,0 +1,98 @@
...
+	@Order(1)
+	@OutboxEventListener
+	public void onAggregatedGeneric(AggregatedEntityEvent event) {
+		capturedEvents.add(new TestEvent("onAggregatedGeneric", event));
+	}
+	
+	@Order(3)
+	@TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)
+	public void onAfterCommitAggregated(TestAggregatedEntityEvent event) {
+		capturedEvents.add(new TestEvent("onAfterCommitAggregated", event));
```
rkorytkowski, line 88: Test if an aggregated event can be listened to with AFTER_COMMIT.

api/src/test/java/org/openmrs/event/outbox/TestOutboxEventListener.java
```diff
@@ -0,0 +1,98 @@
...
+			capturedEvents.add(new TestEvent("onPatientCreatedFailingFailed", event));
+			throw new RuntimeException("Failing listener");
+		}
+		capturedEvents.add(new TestEvent("onPatientCreatedFailingSucceeded", event));
+	}
+
+	@Order(2)
+	@OutboxEventListener
+	public void onAggregatedExact(TestAggregatedEntityEvent event) {
+		capturedEvents.add(new TestEvent("onAggregatedExact", event));
```
rkorytkowski, line 76: Test if an aggregated event can be listened to with OutboxEventListener

api/src/main/java/org/openmrs/aop/OpenmrsServiceEventAdvice.java
```diff
@@ -0,0 +1,129 @@
...
+ * {@link org.springframework.transaction.event.TransactionalEventListener} and 
+ * {@link org.springframework.context.event.EventListener}
+ * can run in the same transaction.
+ * 
+ * @since 2.9.x
+ */
+@Aspect
+@Order(50) // Guarantees execution after transaction advice
+@Component
+public class OpenmrsServiceEventAdvice {
```
rkorytkowski, line 43: Originally I wanted to add this to our RequiredDataAdvice interceptor, but it turns out our service interceptors are registered as preInterceptors in TransactionProxyFactoryBean, which means they are executed before transaction is started. Not sure if it was intended or not, but we want service events to participate in transaction thus this new aspect. It is the modern way to define aspects. We may eventually migrate our interceptors to use aspect as well in core 3.0.0.

api/src/main/java/org/openmrs/event/EventPublisher.java
rkorytkowski: I included TransactionalEventAggregator to be able to aggregate events happening in a single transaction. You may decide, which events you are interested in when implementing the aggregator. It does keep a list of events that is attached to a transaction and emitted as a single event in the BEFORE_COMMIT phase.

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxPollingTaskHandler.java
rkorytkowski: I finally decided to add completedListeners to the event to keep track of successful listeners and only re-try a failed one. Listener's `@Order` is respected.

api/src/main/java/org/openmrs/api/db/event/DeleteDbEvent.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * DB level event emitted upon deletion of an entity.
+ *
+ * @param <T> type of entity
+ */
+public class DeleteDbEvent<T> extends EntityEvent<T> {
+	
+	private Serializable id;
+	private Object[] state;
```
rkorytkowski, line 26: State contains primitive values, but it may also contain references to other entities. All our entities are json serializable in case we want to put an event in the outbox. We are safe here.

dkayiwa:
@rkorytkowski did you get a chance to look at the failing tests?

api/src/main/java/org/openmrs/aop/OpenmrsServiceEventAdvice.java
ibacher: Most likely this is just a result of the limitations of the `TransactionProxyFactoryBean`.

rkorytkowski:
> @rkorytkowski did you get a chance to look at the failing tests?

I've re-run them. I'll have an eye on them. Not sure what happened with the failure.

api/src/main/java/org/openmrs/aop/OpenmrsServiceEventAdvice.java
rkorytkowski: TransactionProxyFactoryBean can also have postInterceptors, which are executed in transaction, but still before method execution (not very intuitive), but it's not a good idea to change to postInterceptors now. Better to migrate away from TransactionProxyFactoryBean...

api/src/main/java/org/openmrs/aop/event/RetireServiceEvent.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Set;
+
+/**
+ * Published by {@link org.openmrs.aop.RequiredDataAdvice}
+ * 
+ * @param <T> the changed entity
+ *     
+ * @since 2.9.x
+ */
+public class RetireServiceEvent<T> extends EntityEvent<T> {
```
dkayiwa, line 23: Could there be a reason why you are adding `Service` to all these event types? For instance, why not just `RetireEvent` instead of `RetireServiceEvent`?

rkorytkowski: To indicate that it comes from OpenmrsService AOP as opposed to db level hibernate interceptors SaveDbEvent or even CDCEvent that will be added for Debezium.

dkayiwa: Ok. Should i also expect to see a PurgeServiceEvent?

rkorytkowski: Yes, missed that one. I'll add it.

dkayiwa:
@rkorytkowski i would love to test these changes from the new module that iam developing, querystore. Do you think you can come up with some sort of documentation/guide that i can follow? Something like `Publishing and consuming events from a module`?

rkorytkowski:
> @rkorytkowski i would love to test these changes from the new module that iam developing, querystore. Do you think you can come up with some sort of documentation/guide that i can follow? Something like `Publishing and consuming events from a module`?

Absolutely.

Hold on with using it though until I add CDC debezium events for querystore... I'm working on it right now.

dkayiwa:
Does that mean the non CDC debezium events are not to be used? 😊

rkorytkowski:
> Does that mean the non CDC debezium events are not to be used? 😊

They may be used for other purposes... like if you want to react on events within the same application transaction (synchronously) or issue custom business logic events. For getting data out of the system asynchronously regardless of application transaction it's best to use CDC events as you won't miss any changes happening outside of application nor within application, but via direct SQL queries.

dkayiwa:
How about using this for now, just for the sake of testing its completeness? 😊

rkorytkowski:
Sounds fine. It's considerably easy switch to just catching a different event and adjusting to a new model.

dkayiwa:
Any module documentation that i can get started with? 😊

rkorytkowski:
https://openmrs.atlassian.net/wiki/spaces/docs/pages/1079803925/Application+Events

dkayiwa:
Awesome! 👍

api/src/main/java/org/openmrs/aop/event/RetireServiceEvent.java
rkorytkowski: Added.

rkorytkowski:
@dkayiwa here's a proposed API for CDCEvent: https://github.com/rkorytkowski/openmrs-core/commit/5b6686968a33aae97b85ef673420dccdec820583

rkorytkowski:
@dkayiwa I'm ready to create a PR for TRUNK-6516, but it depends on this one being merged.

dkayiwa:
@rkorytkowski i am not yet done with testing this out. But in order to unblock you, i am going to merge this for now, and i will continue doing a post commit review. If there are any things to address, you will open a new pull request to address the review comments.

rkorytkowski:
@dkayiwa sounds great! I'll not port it to 3.x until your tests and post-commit code reviews are done.

api/src/main/java/org/openmrs/aop/OpenmrsServiceEventAdvice.java
```diff
@@ -0,0 +1,144 @@
...
+	}
+	
+    /**
+     * This intercepts method executions starting with "save" on OpenmrsService.
+     */
+    @Around("(execution(* *.save*(..)) || execution(* *.create*(..))) && target(org.openmrs.api.OpenmrsService)")
+    public Object interceptSaveOrCreate(ProceedingJoinPoint pjp) throws Throwable {
+		// Do not apply to proxies created by TransactionProxyFactoryBean (legacy) in addition to transactional advice so
+		// that events are not emitted twice.
+		if (AopUtils.isAopProxy(pjp.getTarget())) {
```
dkayiwa, line 58: **Critical: this guard makes the entire aspect a no-op in production.** `AopUtils.isAopProxy` returns `true` for **any** Spring AOP proxy, not just `TransactionProxyFactoryBean` ones. Every core service in `applicationContext-service.xml:333+` (and conventional module services) is TPFB-wrapped, so `pjp.getTarget()` is itself an AOP proxy and we bail out before publishing. The same logic repeats at lines 72/86/100/114/128.

The intent ("skip when the legacy TPFB is in front of us so we don't double-emit") needs a TPFB-specific check, e.g. inspect `Advised.getAdvisors()` for a `TransactionInterceptor`, or detect via the bean post-processor. Please add a runtime test that asserts `SaveServiceEvent<Patient>` actually fires for `PatientService#savePatient` before merging — without that, the whole service-event layer is dead code.

api/src/main/java/org/openmrs/aop/OpenmrsServiceEventAdvice.java
```diff
@@ -0,0 +1,144 @@
...
+	
+	public OpenmrsServiceEventAdvice(EventPublisher eventPublisher) {
+		this.eventPublisher = eventPublisher;
+	}
+	
+    /**
+     * This intercepts method executions starting with "save" on OpenmrsService.
+     */
+    @Around("(execution(* *.save*(..)) || execution(* *.create*(..))) && target(org.openmrs.api.OpenmrsService)")
+    public Object interceptSaveOrCreate(ProceedingJoinPoint pjp) throws Throwable {
```
dkayiwa, line 55: **Simplify:** the 5 advice methods (`interceptSaveOrCreate`, `interceptVoid`, `interceptUnvoid`, `interceptRetire`, `interceptUnretire`) are structurally identical — same proxy check, same `getOpenmrsObjectArgument(...).ifPresent(publish)` body, only the event constructor differs. A single private helper parameterized by `Function<OpenmrsObject, EntityEvent<?>>` and 5 thin pointcut delegates would cut ~80 lines and remove the risk of fixing the bug above in 4 of 5 places.

api/src/main/java/org/openmrs/event/outbox/OutboxEventInterceptor.java
```diff
@@ -0,0 +1,70 @@
...
+				item.setPayload(((OutboxEventPayload) event).toPayload());
+			} else {
+				item.setPayload(objectMapper.writeValueAsString(event));
+			}
+			item.setDateCreated(new Date());
+			item.setDateChanged(item.getDateCreated());
+			item.setStatus(OutboxEvent.Status.PENDING);
+
+			// Guaranteed to run within the existing OpenMRS wrapping transaction (if exists)
+			outboxEventService.saveOutboxEvent(item);
```
dkayiwa, line 62: **Critical: the outbox pattern loses its atomicity guarantee when there is no ambient transaction.** `OutboxEventService#saveOutboxEvent` is `@Transactional` with default propagation `REQUIRED`, so when the originating call site has no tx, the outbox row gets committed in its own private tx. The test `interceptAndSaveToOutbox_shouldNotSaveEventWhenNoTransaction` asserts this is desired — but it defeats the point of the transactional outbox: the row is durable even if the originating change never happens.

Suggested fix: change propagation on `saveOutboxEvent` to `MANDATORY`, *or* check `TransactionSynchronizationManager.isActualTransactionActive()` here and log-and-skip (with a metric) when false. Either way, this needs to be a deliberate decision called out in the PR description.

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxPollingTaskHandler.java
```diff
@@ -0,0 +1,148 @@
...
+				
+				if (!outboxEventService.lockEventForProcessing(item)) {
+					// Another scheduled task handler just grabbed it and is processing it right now.
+					// Stop processing subsequent events to maintain strict ordering.
+					return;
+				}
+
+				Set<String> completedListeners = new LinkedHashSet<>();
+				try {
+					Class<?> eventClass = Class.forName(item.getEventType());
```
dkayiwa, line 97: **Critical for modules: `Class.forName(String)` uses the caller's classloader (core), which cannot see module classes.** Any `OutboxableEvent` defined in a module that is still PENDING when the server restarts (or even immediately, if the poller thread's classloader differs from the publishing thread's) will fail with `ClassNotFoundException`, and per the poison-pill issue below will block the queue indefinitely.

Use `OpenmrsClassLoader.getInstance().loadClass(item.getEventType())`. Same fix needed for the `OutboxEventPayload.class.isAssignableFrom(eventClass)` path that follows. Worth a dedicated test: register a module-style listener for an event class loaded via OpenmrsClassLoader, restart the context, verify the poller can still deserialize.

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxPollingTaskHandler.java
```diff
@@ -0,0 +1,148 @@
...
+					}
+
+					// Revert to PENDING (to retry)
+					item.setStatus(OutboxEvent.Status.PENDING);
+					item.setErrorCount(errorCount);
+					item.setErrorMessage(errorMessage);
+					outboxEventService.saveOutboxEvent(item);
+
+					// Stop processing subsequent events to ensure strict ordering
+					throw outboxException;
```
dkayiwa, line 142: **Critical: one bad event blocks the entire queue, forever.** Re-throwing here stops processing of *all* subsequent events to preserve ordering, but the broken event stays PENDING with no max-retry, no exponential backoff, no dead-letter, and no terminal `ERROR` state in the `Status` enum (despite the offline discussion mentioning "manually change status to ERROR"). A single deserialization failure or a buggy listener will quietly halt every consumer in the system.

Minimum acceptable: after N retries (configurable, e.g. 10) move the event to a new `Status.FAILED` state, publish a `OutboxExceptionEvent`, and skip past it so the rest of the queue drains. Add an admin endpoint (or at least a documented SQL update) to requeue manually.

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxPollingTaskHandler.java
```diff
@@ -0,0 +1,148 @@
...
+	 * 
+	 * @param taskData task data
+	 * @param taskContext task context
+	 * @throws Exception
+	 */
+	@Override
+	public void execute(OutboxPollingTaskData taskData, TaskContext taskContext) throws Exception {
+		outboxEventService.resetStuckEvent();
+
+		while (true) {
```
dkayiwa, line 71: **Critical: `while (true)` can starve the scheduler.** Under sustained load (or a write storm) this method only returns when the queue empties, which may never happen. The author's own javadoc on line 67 ("can run indefinitely regardless of the schedule") acknowledges this — but "indefinitely" inside a JobRunr worker means the 15s recurring schedule is meaningless and other tasks behind this worker are blocked.

Add a per-invocation budget: cap at e.g. 500 events or 30 seconds wall-clock, then return so the scheduler can reclaim the worker and re-fire on schedule.

api/src/main/java/org/openmrs/event/outbox/OutboxEventRegistry.java
```diff
@@ -0,0 +1,156 @@
...
+
+	private final ApplicationContext applicationContext;
+	private final List<ListenerMethod> registry;
+	private final ConcurrentMap<ResolvableType, Boolean> hasOutboxListenersCache;
+	private final boolean enabled;
+	
+	public OutboxEventRegistry(ApplicationContext applicationContext, @Value("${outboxevent.enabled:true}") boolean enabled) {
+		this.applicationContext = applicationContext;
+		this.registry = new ArrayList<>();
+		this.hasOutboxListenersCache = new MapMaker().weakKeys().weakValues().makeMap();
```
dkayiwa, line 52: **Critical: this cache is effectively disabled.** `MapMaker().weakKeys().weakValues()` will GC entries as soon as no other strong reference exists to the `ResolvableType` key — and callers in `hasOutboxListeners(Object event)` build a fresh `ResolvableType.forInstance(event)` on each call and don't retain it. Net result: cache miss on every publish, defeating the whole reason the inline comment at line 38 says "needs to be super fast."

Use a strong-referenced bounded cache (`Caffeine.newBuilder().maximumSize(2048).build()`) keyed by `event.getClass()` (not `ResolvableType` — class identity is enough since `targetEventType.isAssignableFrom` only needs raw type compatibility for the common case).

api/src/main/java/org/openmrs/serialization/JacksonConfig.java
```diff
@@ -26,7 +26,7 @@
 public class JacksonConfig {
 	
 	@Bean
-	public ObjectMapper schedulerObjectMapper() {
+	public ObjectMapper objectMapper() {
```
dkayiwa, line 29: **Critical: bean rename is a breaking change for modules.** Renaming `schedulerObjectMapper()` → `objectMapper()` changes the Spring bean name. Any module that has `@Qualifier("schedulerObjectMapper")` (a published name that has been part of this branch) will fail at startup with `NoSuchBeanDefinitionException` after upgrading. Worse, modules that have an `@Autowired ObjectMapper` without a qualifier will silently bind to *this* bean and may get different serialization behaviour than they expect.

Keep the old bean name as the primary and add an `@Bean(name = {"objectMapper", "schedulerObjectMapper"})` alias for one release cycle, with a deprecation note in the upgrade guide.

api/src/main/java/org/openmrs/event/outbox/OutboxEventService.java
```diff
@@ -0,0 +1,99 @@
...
+	public OutboxEventService(SessionFactory sessionFactory, @Value("${outboxevent.listener.timeout:120}") int listenerTimeout) {
+		this.sessionFactory = sessionFactory;
+		this.listenerTimeout = Duration.ofSeconds(listenerTimeout);
+	}
+
+	/**
+	 * Revert tasks stuck in PROCESSING for more than {@code outboxevent.listener.timeout} .
+	 */
+	@Transactional(noRollbackFor = OutboxException.class)
+	public void resetStuckEvent() throws OutboxException {
```
dkayiwa, line 47: **Cluster race:** in a multi-node deployment every node runs this poller; nothing prevents two nodes from executing `resetStuckEvent` simultaneously. The UPDATE filters on `status = 'PROCESSING'` so it's safe-ish from a correctness standpoint, but `errorCount` will be double-incremented and an `OutboxException` will be thrown twice. The PR ships a `@Mutex` annotation explicitly for this use case but doesn't apply it here (or to the polling task at large). At minimum, annotate this method (and `OutboxPollingTaskHandler#execute`) with `@Mutex(name = "outbox.poller", lockAtMostFor = "5m")`.

api/src/main/java/org/openmrs/event/EventPublisher.java
```diff
@@ -0,0 +1,50 @@
...
+	
+	public EventPublisher(ApplicationEventPublisher delegate, SessionFactory sessionFactory) {
+		this.delegate = delegate;
+		this.sessionFactory = sessionFactory;
+	}
+
+	@Override
+	public void publishEvent(@NonNull Object event) {
+		if (event instanceof BaseEvent) {
+			String sessionId = ((SessionImplementor) sessionFactory.getCurrentSession()).getSessionIdentifier().toString();
```
dkayiwa, line 41: **`sessionFactory.getCurrentSession()` throws when no Hibernate session is bound to the thread.** Any code path that publishes a `BaseEvent` from outside a Hibernate session — scheduled tasks before `Context.openSession()`, async workers, startup `SmartInitializingSingleton`s, anything reachable from a module's bootstrap — will get `HibernateException: No Session found for current thread` instead of an event.

Wrap in `try { ... } catch (HibernateException ignored) { /* sessionId stays null */ }` or check `TransactionSynchronizationManager.hasResource(sessionFactory)` first. Also worth a test that publishes a `BaseEvent` from a non-Hibernate thread.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.9.x.xml
```diff
@@ -48,5 +45,56 @@
...
+			<column name="payload" type="TEXT">
+				<constraints nullable="false"/>
+			</column>
+			<column defaultValue="PENDING" name="status" type="VARCHAR(50)">
+				<constraints nullable="false"/>
+			</column>
+			<column defaultValueNumeric="0" name="error_count" type="INT">
+				<constraints nullable="false"/>
+			</column>
+			<column name="error_message" type="VARCHAR(1024)"/>
```
dkayiwa, line 69: **`VARCHAR(1024)` is too small for real Java stack traces** (commonly 10–50 KB). Combined with `errorMessage.substring(0, 1024)` in `OutboxPollingTaskHandler.java:132`, the stored message will be a truncated head with no terminal frame — usually useless for debugging. Also note `substring(0, 1024)` counts chars not bytes; if MySQL's column is byte-length-limited under utf8mb4, multibyte chars near the boundary may still overflow.

Change the column type to `TEXT` (or `CLOB` for portability) and remove the substring truncation.

api/src/main/java/org/openmrs/event/outbox/OutboxEvent.java
```diff
@@ -0,0 +1,184 @@
...
+
+	@Column(name = "error_count", nullable = false)
+	private Integer errorCount = 0;
+
+	@Column(name = "error_message", length = 1024)
+	private String errorMessage;
+	
+	@Lob
+	@Column(name = "completed_listeners")
+	private String completedListeners;
```
dkayiwa, line 71: **`completedListeners` as a comma-separated string of `beanName + "." + methodName` is fragile.** Two overloaded `@OutboxEventListener` methods on the same bean collapse to the same id and one will be skipped on retry. Bean names containing commas (rare but legal) corrupt the list. Splitting/joining on every listener invocation also allocates needlessly.

Options: (a) one row per (event, listener) in a join table, (b) a JSON array column, (c) include the parameter type signature in the listener id. (a) also lets you index it for the cleanup task.

api/src/main/java/org/openmrs/event/outbox/OutboxEvent.java
```diff
@@ -0,0 +1,184 @@
...
+		return this.dateChanged;
+	}
+
+	@Override
+	public void setDateChanged(Date dateChanged) {
+		this.dateChanged = dateChanged;
+	}
+	
+	public enum Status {
+		PENDING, PROCESSING, COMPLETED
```
dkayiwa, line 182: **Status enum is missing `ERROR` / `FAILED`.** The author's own discussion above says operators are expected to "manually change the status to ERROR" to skip a poison-pill event — but no such value exists in this enum, so manual SQL would have to insert an out-of-band string. Add a terminal failure state and surface it from the scheduler UI.

api/src/main/java/org/openmrs/event/TransactionalEventAggregator.java
```diff
@@ -0,0 +1,115 @@
...
+		TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
+
+			@Override
+			public int getOrder() {
+				return Ordered.HIGHEST_PRECEDENCE; // Run before any other TransactionalEventListeners
+			}
+
+			@Override
+			public void beforeCommit(boolean readOnly) {
+				Context.flushSession(); // Make sure that all DB events are published before this event.
```
dkayiwa, line 99: **Calling `Context.flushSession()` inside `beforeCommit` is historically fragile.** Hibernate flushes in `beforeCommit` can trigger re-entrant flushes if a listener writes back to the session, and can mask `LazyInitializationException`s that only surface in production. Please add an explicit test that exercises a `@TransactionalEventListener(BEFORE_COMMIT)` chain on top of the aggregated event, and another that has a listener performing an additional save (verifying the second save also dirties the session correctly).

api/src/main/java/org/openmrs/event/outbox/OutboxEventService.java
```diff
@@ -0,0 +1,99 @@
...
+import org.springframework.transaction.annotation.Transactional;
+
+import javax.persistence.Query;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Date;
+import java.util.List;
+
+/**
+ * It's internal and not for use in modules.
```
dkayiwa, line 26: **Module gap:** marking this `@Service` as "internal and not for use in modules" closes off the only management surface modules have over their own outbox events. Combined with the poison-pill issue, a module-defined event that fails to deserialize has *no* public API a module-side admin page could use to inspect, retry, or expire it. Please either expose a stable read-only API (e.g. `getPendingEvents(eventTypePrefix)`, `markFailed(id)`) or add a separate `OutboxAdminService` that modules can use.

api/src/main/java/org/openmrs/aop/event/SaveServiceEvent.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Set;
+
+/**
+ * Published by {@link org.openmrs.aop.OpenmrsServiceEventAdvice}
+ *
+ * @param <T> the changed entity
+ *
+ * @since 2.9.x
+ */
+public class SaveServiceEvent<T> extends EntityEvent<T> {
```
dkayiwa, line 23: **Simplify:** `SaveServiceEvent`, `VoidServiceEvent`, `UnvoidServiceEvent`, `RetireServiceEvent`, `UnretireServiceEvent` are byte-for-byte identical except the class name. Listeners that want type-based dispatch do still need distinct classes (Spring's `ResolvableType` keys on the concrete event class), so deleting them isn't an option — but the bodies should be generated from a single abstract base with a no-op subclass per action. As written, a future change has to be made in 5 places. Consider also documenting which is the canonical (and recommended) shape for module-defined service-level events.

api/src/main/java/org/openmrs/scheduler/SchedulerService.java
```diff
@@ -279,14 +291,45 @@ public interface SchedulerService extends OpenmrsService {
...
 	 * See {@link #schedule(TaskData)} for more details.
-	 *  
+	 *
+	 * @param taskData
 	 * @param name
+	 * @return TaskDetails
+	 * @since 2.9.x
+	 */
+	@Authorized
+	TaskDetails schedule(TaskData taskData, String name);
```
dkayiwa, line 301: **Simplify:** this PR adds 16 new overloads of `schedule(...)` / `scheduleRecurrently(...)`. The combinatorial pattern of `(uuid?, taskData, when?, name?)` is hard to read, encourages positional confusion (note that this PR itself swaps the `(String, TaskData)` ↔ `(TaskData, String)` order, breaking existing call sites like `SchedulerServiceIT.java:223`), and grows quadratically as new parameters are added.

Replace with one `schedule(ScheduleRequest)` / `scheduleRecurrently(ScheduleRequest)` taking a builder: `ScheduleRequest.builder().uuid(...).taskData(...).runAt(...).name(...).build()`. The 16 overloads can be deprecated rather than removed, but new callers (including the `OutboxTaskSchedulerInitializer` added in this PR) should use the builder.

api/src/main/java/org/openmrs/event/outbox/OutboxEventInterceptor.java
rkorytkowski: @dkayiwa I will fix the test method name to interceptAndSaveToOutbox_shouldSaveEventWhenNoTransaction. If an outer transaction exist, the event should be persisted in the same transaction if it is not rolled-back. If an outer transaction does not exist, the event should still be persisted in its own transaction. We don't want to limit the use of outbox for transactional events only.

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxPollingTaskHandler.java
rkorytkowski: AI is not correct. Core's classloader sees all module classes.

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxPollingTaskHandler.java
rkorytkowski: This is intended. There's only a single instance of this task running across all replicas and jobrunr has at minimum 8 workers on a single core CPU node so we can safely use one worker to run indefinitely. Recurring schedule is there to guarantee the task is re-scheduled on a different node if the current one fails.

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxPollingTaskHandler.java
rkorytkowski: Agree, it would be good to have some dead letter queue, yet I wanted to give implementers the choice on what to do in case of failing event thus publishing OutboxExceptionEvent.

I'll go ahead with the suggestion to move to failed state after N retries by default, but I let it be driven by the OutboxExceptionEvent.

api/src/main/java/org/openmrs/aop/OpenmrsServiceEventAdvice.java
rkorytkowski: AI is not correct. It doesn't take into account that we have a Transactional advice in addition to TransactionProxyFactoryBean (due to legacy code), which is not the usual setup. We have tests in OutboxEventIT confirming the advice works for SaveServiceEvent<GlobalProperty>. It's the same as what AI suggested with SaveServiceEvent<Patient>.

rkorytkowski (commented):
I'll add a basic dead letter queue.

api/src/main/java/org/openmrs/aop/event/SaveServiceEvent.java
rkorytkowski: I don't quite understand what's being suggested here... ;)

api/src/main/java/org/openmrs/scheduler/SchedulerService.java
rkorytkowski: Agree a builder would be nice, but 4 args methods are not that much for the ease of use. JobRunr API does the same thing... The order swap is fine now since API has not been released yet.

api/src/main/java/org/openmrs/event/outbox/OutboxEventService.java
rkorytkowski: Fair point.

api/src/main/java/org/openmrs/event/TransactionalEventAggregator.java
rkorytkowski: We explicitly say in docs that @TransactionalEventListener(BEFORE_COMMIT) is not supported on AggregatedEntityEvent as it already runs in the BEFORE_COMMIT phase.

PR #6088, opened 2026-05-11
TRUNK-6635 Add search by LocationSearchCriteria, with option to find …

chibongho:
…descendent Locations efficiently

## Description of what I changed

Add `LocationSearchCriteria` class for search Location. Supports searching by tags, name fragments and descendant of another location. Queries efficiently via CTE query for descendant search.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6635

## Checklist: I completed these to help reviewers :)

- [x] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

  No? Unsure? -> [configure your IDE](https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE), format the code and add the changes with `git add . && git commit --amend`

- [x] I have **added tests** to cover my changes. (If you refactored
  existing code that was well tested you do not have to add tests)

  No? -> write tests and add them to this commit `git add . && git commit --amend`

- [x] I ran `mvn clean package` right before creating this pull request and
  added all formatting changes to my commit.

  No? -> execute above command

- [x] All new and existing **tests passed**.

  No? -> figure out why and add the fix to your commit. It is your responsibility to make sure your code works.

- [x] My pull request is **based on the latest changes** of the master branch.

  No? Unsure? -> execute command `git pull --rebase upstream master`

## Summary by CodeRabbit

* **New Features**
  * New flexible location search API: descendant/ancestor filtering, tag matching (ALL/ANY), name-fragment filtering, and include-retired option; combinable filters.

* **Refactor**
  * Descendant-location resolution delegated to the new search-based mechanism for consistent, scalable queries.

* **Tests**
  * Added comprehensive service-level tests for search behaviors; simplified some unit tests to JUnit 5.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openmrs/openmrs-core/pull/6088)

PR #6089, opened 2026-05-11
TRUNK-6635 Add search by LocationSearchCriteria, with option to find …

chibongho:
…descendent Locations efficiently

Add `LocationSearchCriteria` class for search Location. Supports searching by tags, name fragments and descendant of another location. Similar to https://github.com/openmrs/openmrs-core/pull/6088, but for the 2.9.x branch; **however**, we do not use CTE query here as that's not supported in MySQL 5.7. Instead, we get all `Locations` and filter in memory. (This is probably good enough as Location should be < 10000 for most implementations.) 

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6635

  No? -> execute above command


api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
```diff
@@ -0,0 +1,97 @@
...
+	 * Controls how multiple tags are matched against a location.
+	 */
+	public enum TagMatchMode {
+		/** Location must have every tag in the collection. */
+		ALL,
+		/** Location must have at least one tag in the collection. */
+		ANY
+	}
+
+	private final String descendantOfLocationUuid;
```
mseaton: Any particular reason why you are using uuid here vs. an actual Location?  I don't love using untyped uuids when we can use a properly typed Location

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
```diff
@@ -0,0 +1,97 @@
...
+
+	private final Collection<String> locationTagUuids;
```
mseaton: Any particular reason for the uuids here?  I would personally prefer that we either:
a) make this a `Collection<LocationTag>`, or
b) make this a `Collection<String> locationTags`, which isn't tied to uuid (eg. using tag name is more typical and more descriptive, and equally unique)

mogoodrich: Agree!

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
```diff
@@ -436,4 +441,64 @@ private List<Integer> getLocationTagIds(List<LocationTag> tags) {
...
+
+	/**
+	 * @see LocationDAO#getLocations(LocationSearchCriteria)
+	 */
+	@Override
+	public List<Location> getLocations(LocationSearchCriteria criteria) {
+		List<String> tagUuids = criteria.getLocationTagUuids() == null ? null
+		        : criteria.getLocationTagUuids().stream().filter(StringUtils::isNotBlank).collect(Collectors.toList());
+
+		List<Location> result = getAllLocations(true).stream().filter(loc -> matchesCriteria(loc, criteria, tagUuids))
```
mogoodrich: Is it intentional to do all the searching/filtering programmatically (and not via the DB?).  I find the programmatic method easier to follow, but generally DAO stuff operate on the DB layer for performance reasons--though admittedly the # of locations in any system should be small (compared to data like patients, obs, etc)

mogoodrich (commented):
Thanks @chibongho , see my question.

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
chibongho: I was thinking if we expose this as a SearchHandler in the REST module, like `/rest/v1/location?descendantOfLocation=<uuid>&tag=...`, we can pass the uuid directly into the `LocationSearchCriteria` instead of having to query for the `Location` object by  uuid first.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
chibongho: This PR is for the 2.9.x branch. The [corresponding PR](https://github.com/openmrs/openmrs-core/pull/6088) to the master branch does it via a more efficient CTE query, which we can't do here because 2.9.x, presumably like 2.8, still supports MySQL 5.7.

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
chibongho: Yeah, I can change it to tag names. Or if we feel strongly that it should be typed more strictly, `LocationTag`.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
mogoodrich: Ah, I see, works for me... any concerns @mseaton  @ibacher ?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
```diff
@@ -436,4 +442,61 @@ private List<Integer> getLocationTagIds(List<LocationTag> tags) {
...
+		}
+
+		if (StringUtils.isNotBlank(criteria.getNameFragment())
+		        && (loc.getName() == null
+		                || !loc.getName().toLowerCase().startsWith(criteria.getNameFragment().toLowerCase()))) {
+			return false;
+		}
+
+		if (tags != null && !tags.isEmpty()) {
+			Set<Integer> locTagIds = loc.getTags().stream().map(LocationTag::getLocationTagId).collect(Collectors.toSet());
```
mseaton: Why do we need to convert to ids for comparison.  I would hope we can trust the equals method of all of our openmrs objects.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
```diff
@@ -436,4 +442,61 @@ private List<Integer> getLocationTagIds(List<LocationTag> tags) {
...
+		}
+
+		if (criteria.getDescendantOfLocation() != null
+		        && !isDescendantOf(loc, criteria.getDescendantOfLocation(), criteria.getIncludeRetired())) {
+			return false;
+		}
+
+		if (StringUtils.isNotBlank(criteria.getNameFragment())
+		        && (loc.getName() == null
+		                || !loc.getName().toLowerCase().startsWith(criteria.getNameFragment().toLowerCase()))) {
```
mseaton, line 469: "nameFragment" doesn't necessarily imply "startWith".  If this is existing behavior, then fine.  But if we are adding this new, I would better match the name and the behavior.  Not sure why this isn't just being done at the DB layer?

api/src/main/java/org/openmrs/parameter/LocationSearchCriteriaBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+
+import org.openmrs.Location;
+import org.openmrs.LocationTag;
+
+/**
+ * A builder for constructing {@link LocationSearchCriteria} instances.
+ *
+ * @since 2.9.0
```
mseaton: This class is fine, but not sure it is strictly needed.  I'd love to move all our our services to use a criteria search pattern, but not sure I want to create and mantain builder classes for all of them.  But defer to @ibacher  and @dkayiwa as to what they think generally.

api/src/main/java/org/openmrs/Location.java
```diff
@@ -147,6 +148,7 @@ public class Location extends BaseCustomizableMetadata<LocationAttribute> implem
 		name = "location_tag_map",
 		joinColumns = @JoinColumn(name = "location_id"),
 		inverseJoinColumns = @JoinColumn(name = "location_tag_id"))
+	@BatchSize(size = 100)
```
mseaton, line 150: What is the implication of this and why is it needed?

api/src/main/java/org/openmrs/Location.java
```diff
@@ -463,15 +465,8 @@ public Set<Location> getChildLocations(boolean includeRetired) {
...
-		
-		for (Location childLocation : getChildLocations()) {
-			if (!childLocation.getRetired() || includeRetired) {
-				result.add(childLocation);
-				result.addAll(childLocation.getDescendantLocations(includeRetired));
-			}
-		}
-		return result;
+		return new HashSet<>(Context.getLocationService().getLocations(
+		    new LocationSearchCriteriaBuilder().setDescendantOfLocation(this).includeRetired(includeRetired).build()));
```
mseaton: I don't really love this.  If we want to share this functionality, let's move it to a utility method that both the criteria builder and this use.  But using the criteria builder directly here feels wrong.

api/src/main/java/org/openmrs/Location.java
chibongho: Yeah, I should have commented on this. I added this line after doing some DB profiling yesterday. 

When we get a `List<Location>` resulting from a Hibernate query, by default the corresponding locations tags are not queried (because of `FetchType.LAZY`). The `@BatchSize` annotation makes it so that, if we call location.getTags() on any of the `Location`s in the list, we request tags for not just that one `Location`, but batch request for up to 100 Locations. So if we have a loop like this:

```
List<Location> locations;
for(Location l : locations) {
  l.getTags();
}
```
instead of making n queries for tags for n locations, we make `ceil(n/100)` queries. This affects the `getLocations(LocationSearchCriteria criteria)` function most because it loads all locations in memory first and then filters on them, but it also improves performance on a few O2 pages. (for zl-ci, it shaves about ~0.3s for each page.)

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
chibongho: I changed my mind on this given that: a few `*Criteria` classes also use the native types instead of uuids, and one extra query to get the Location is trivial.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
chibongho: This matches the existing behavior: https://github.com/openmrs/openmrs-core/blob/19a2bfd5aee7961682117bc436b46b3a748d8ab6/api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java#L272

We're already doing in-memory comparison for `isDescendantOf` and tags. There is marginal improvement for putting this in the DB layer, but I was hoping we don't need to maintain / optimize this too much given that:
- the 3.0 (master) branch does this in DB layer
- for most distros, number of locations is < 1000.

api/src/main/java/org/openmrs/Location.java
chibongho: The builder is just to specify searching for descendants. I can make a `getDescendantOf(Location location)` function in LocationService.

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
```diff
@@ -0,0 +1,100 @@
...
+
+import org.openmrs.Location;
+import org.openmrs.LocationTag;
+
+/**
+ * Encapsulates the criteria used for searching locations. All provided criteria are ANDed together.
+ * <p>
+ * Instead of calling the constructor directly, use {@link LocationSearchCriteriaBuilder}.
+ *
+ * @since 2.9.0
```
mseaton: If you are planning to backport this to 2.8.x, then this needs to be since whatever 2.8.x release is next.  Doing this PR against the 2.8.x branch would have made this easier / more obvious.
You have this correct in the builder class.  Needs to be applied here too

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
```diff
@@ -0,0 +1,100 @@
...
+	/**
+	 * Constructs a LocationSearchCriteria. Use {@link LocationSearchCriteriaBuilder} instead.
+	 *
+	 * @param descendantOfLocation the ancestor location; only its descendants are returned (root excluded)
+	 * @param locationTags tags to filter by
+	 * @param tagMatchMode whether all or any of the tags must match
+	 * @param nameFragment only return locations whose name starts with this fragment
+	 * @param includeRetired whether to include retired locations
+	 */
+	public LocationSearchCriteria(Location descendantOfLocation, Collection<LocationTag> locationTags,
```
mseaton: I do not like having this constructor.  One of the big goals of creating this class is to avoid having methods that take in lots of parameters.  We should just have a default constructor in this class.

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
```diff
@@ -0,0 +1,100 @@
...
+	    TagMatchMode tagMatchMode, String nameFragment, boolean includeRetired) {
+		this.descendantOfLocation = descendantOfLocation;
+		this.locationTags = locationTags == null ? Collections.emptyList()
+		        : Collections.unmodifiableCollection(locationTags);
```
mseaton: Why should it be unmodifiable?  And shouldn't it be a set vs a list?  Just set it to an empty set when the variable is declared.

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Collection;
+import java.util.Collections;
+
+import org.openmrs.Location;
+import org.openmrs.LocationTag;
+
+/**
+ * Encapsulates the criteria used for searching locations. All provided criteria are ANDed together.
+ * <p>
+ * Instead of calling the constructor directly, use {@link LocationSearchCriteriaBuilder}.
```
mseaton: You can indicate that LocationSearchCriteriaBuilder exists, but no reason someone needs to use it.

api/src/main/java/org/openmrs/parameter/LocationSearchCriteriaBuilder.java
```diff
@@ -0,0 +1,95 @@
+/**
```
mseaton: I would personally get rid of this class.  Just keep things simple for now.  Consumers can construct a LocationSearchCriteria without needing a builder

api/src/main/java/org/openmrs/Location.java
mseaton: Sounds good, just want to make sure there are no side effects to this that would have an unexpected impact.

api/src/main/java/org/openmrs/Location.java
```diff
@@ -463,15 +465,7 @@ public Set<Location> getChildLocations(boolean includeRetired) {
...
-		Set<Location> result = new HashSet<>();
-		
-		for (Location childLocation : getChildLocations()) {
-			if (!childLocation.getRetired() || includeRetired) {
-				result.add(childLocation);
-				result.addAll(childLocation.getDescendantLocations(includeRetired));
-			}
-		}
-		return result;
+		return new HashSet<>(Context.getLocationService().getDescendantLocations(this, includeRetired));
```
mseaton: This is not correct.  A domain object should be be making calls to the service.  This needs to be removed / fixed.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
```diff
@@ -436,4 +442,61 @@ private List<Integer> getLocationTagIds(List<LocationTag> tags) {
...
+	public List<Location> getLocations(LocationSearchCriteria criteria) {
+		Collection<LocationTag> tags = criteria.getLocationTags();
+
+		return getAllLocations(true).stream().filter(loc -> matchesCriteria(loc, criteria, tags))
+		        .collect(Collectors.toList());
+	}
+
+	private boolean matchesCriteria(Location loc, LocationSearchCriteria criteria, Collection<LocationTag> tags) {
+		if (!criteria.getIncludeRetired() && loc.getRetired()) {
+			return false;
```
mseaton, line 459: Can you fix the above sonarcloud warnings?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
mseaton: You didn't address this comment @chibongho

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
```diff
@@ -436,4 +442,61 @@ private List<Integer> getLocationTagIds(List<LocationTag> tags) {
...
+		}
+
+		return true;
+	}
+
+	private boolean isDescendantOf(Location loc, Location ancestor, boolean includeRetired) {
+		Location current = loc.getParentLocation();
+		while (current != null) {
+			if (!includeRetired && current.getRetired()) {
+				return false;
```
mseaton, line 490: Address this sonarcloud comment

mseaton (changes requested):

api/src/main/java/org/openmrs/Location.java
chibongho: I'm just going to deprecate this in favor of `LocationService#getDescendantLocations(Location, boolean)`

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
chibongho: Sorry... my change got lost. Re-adding.

mseaton (approved):
Looks good @chibongho - thanks for working through those reviews

chibongho:
Thanks. Merging this in. I will also:
- back-port this PR wholesale to the 2.8.x branch
- forward-port the commits addressing PR comments to the master (3.x) branch PR https://github.com/openmrs/openmrs-core/pull/6088

PR #6128, opened 2026-05-26
TRUNK-6550: Logging should not be vulnerable to injection attacks

jwnasambu:
I updated logging statements to prevent log injection vulnerabilities by sanitizing untrusted input before writing it to logs. Specifically, carriage return (\r) and newline (\n) characters are replaced to prevent attackers from forging or manipulating log entries. This ensures log integrity and aligns with secure logging practices recommended by SonarQube and OWASP.

## Issue I worked on

https://openmrs.atlassian.net/browse/TRUNK-6550

  No? -> execute above command


jwnasambu:
@dkayiwa, @chibongho  kindly feel free to review my PR at your convenient time please!

jwnasambu:
@wikumChamith, @ibacher, @dkayiwa Kindly feel free to review my PR at your convenient time please!

jwnasambu:
@wikumChamith, @ibacher, @dkayiwa Kindly feel free to review my PR at your convenient time please!

jwnasambu:
@claude review

dkayiwa:
Did you get a chance to look at the build failure?

api/src/main/java/org/openmrs/logging/OpenmrsLoggingUtil.java
```diff
@@ -320,4 +320,21 @@ private static LoggerContext getLog4j2Context() {
...
 		return ((Logger) rootLogger).getContext();
 	}
+
+	/**
+	 * Sanitizes a string for logging by replacing newline and carriage return characters with
+	 * underscores.
+	 *
+	 * @param value the string to sanitize
+	 * @return the sanitized string
+	 * @since 2.4.4, 2.5.1, 2.6.0
```
ibacher: This `@since` makes literally no sense.

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -374,7 +375,7 @@ public static void refreshAuthenticatedUser() {
 	 * @throws ContextAuthenticationException
 	 */
 	public static void becomeUser(String systemId) throws ContextAuthenticationException {
-		log.info("systemId: {}", systemId);
+		log.info("systemId: {}", OpenmrsLoggingUtil.sanitize(systemId));
```
ibacher: @jwnasambu As Sonar says, we should only invoke this method conditionally. This applies everywhere. But also, this is a very brute-force solution that isn't going to be scalable because we need this to apply to literally every logging message the system makes. I think that means that we have to apply the sanitization _lower down_ in the logging chain than the example from the ticket.

ibacher: FWIW https://logging.apache.org/log4j/2.x/manual/pattern-layout.html#converter-encode

ibacher (changes requested):

jwnasambu:
@wikumChamith, @ibacher, @dkayiwa Kindly feel free to review my PR at your convenient time please!

api/src/main/resources/log4j2.xml
```diff
@@ -15,7 +15,7 @@
 	<Properties>
 		<!-- The default pattern is stored as a property so that it's only defined once.
 		     It's also quite challenging to escape using Log4j2's variable substitution. -->
-		<Property name="defaultPattern">%p - %C{1}.%M(%L) |%d{ISO8601}| %m%n</Property>
+		<Property name="defaultPattern">%p - %C{1}.%M(%L) |%d{ISO8601}| %encode{%m}{CRLF}%n</Property>
```
dkayiwa, line 18: This encode never reaches an appender, so as written the layout change does nothing. I loaded this file at the PR head into a LoggerContext and dumped what each appender actually ends up with:

```
APPENDER CONSOLE -> pattern=[%p - %C{1}.%M(%L) |%d{ISO8601}| %m%n]
APPENDER OPENMRS FILE APPENDER -> pattern=[%p - %C{1}.%M(%L) |%d{ISO8601}| %m%n]
APPENDER MEMORY_APPENDER -> pattern=[%p - %C{1}.%M(%L) |%d{ISO8601}| %m%n]
```

All three read `${openmrs:logLayout:-${defaultPattern}}`, and `OpenmrsPropertyLookup.lookup(..., "logLayout")` never returns null: with no system property, no runtime property and no `log.layout` global property it falls through to `getPropertyDefault()`, which hands back `OpenmrsConstants.DEFAULT_LOG_LAYOUT_PATTERN`, still `%m%n` on this branch. So `${defaultPattern}` is only consulted if the lookup itself throws, and the value edited here is dead. Rendering a forged message through the pattern the appenders do get still splits it into two lines:

```
INFO - ?.() |2026-09-11T15:15:28,498| user=bob
ERROR - Forged.entry(1) |2026-01-01 00:00:00,000| attacker owned the box
```

Merged as is, anyone who can get a newline into a logged string still forges entries exactly as before, while the PR reads as though the layout now stops it.

To make it real the encode also has to go into `OpenmrsConstants.DEFAULT_LOG_LAYOUT_PATTERN`, and `log.layout` needs a liquibase update, because it is seeded from that constant at install time so every existing database already holds the unencoded value. `webapp/src/main/resources/log4j2.xml` is a near duplicate of this file and still carries the old `defaultPattern` too.

#6165 already changes this same line plus the constant plus the global property, and adds a second sanitizing helper next to the one added here. Those two will collide, so it is worth deciding which PR owns the layout fix instead of landing both.

A test that renders a CR/LF message through the pattern the appenders are configured with would have caught this. The tests added here only exercise `sanitize()` on plain strings, which is why the build stayed green.

api/src/main/java/org/openmrs/api/context/Context.java
dkayiwa: The call site sanitizing is still here, 13 calls against roughly 1100 `log.*` statements in `src/main`, and none of the 13 sits at a spot this rule has ever been raised on. In the repo's whole code scanning history `javasecurity:S5145` has only ever fired on three files, and the four alerts open today are `InitializationFilter.java:994` and `:1910` plus `DatabaseUtil.java:79` and `:111`. TRUNK-6550's own link is alert 188, and at the commit that alert was filed against, line 979 is this:

```java
log.info("Stored locale parameter to session " + localeParameter);
```

That is `checkLocaleAttributes`, where `localeParameter` is `httpRequest.getParameter(...)` straight off the setup wizard request, so it is the one entry in the list where the logged string genuinely comes from outside. It has since drifted to line 994 and is still open as alert 347, and this PR does not touch it, so merging closes TRUNK-6550 without changing the line TRUNK-6550 points at.

Doing it the way you suggested fixes that for free: once the layout does the encoding, all 13 of these calls become redundant and can come out, and the flagged sites are covered without anyone having to remember to wrap a new one. The layout half of this PR is not doing the encoding yet though, see my note on `log4j2.xml`.

ibacher:
I'm cancelling this because this ticket is just a duplicate of TRUNK-6549, which has a separate PR.

PR #6151, opened 2026-06-02
TRUNK-6516: Provide CDC mechanism with Debezium

rkorytkowski:

Please see https://github.com/openmrs/openmrs-module-debezium/commit/6fdf039bc36236ca611fc924ecf26765b9415b29 for a module that produces CDC events.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6516

  No? -> execute above command


api/src/main/java/org/openmrs/event/BaseEvent.java
```diff
@@ -22,34 +21,20 @@
  * 
  * @since 2.9.x
  */
-class BaseEvent extends ApplicationEvent implements OutboxableEvent {
+class BaseEvent implements OutboxableEvent {
```
rkorytkowski, line 24: According to Spring docs extending ApplicationEvent is no longer recommended or needed for events.

api/src/main/java/org/openmrs/event/BaseSessionEvent.java
```diff
@@ -0,0 +1,47 @@
...
+package org.openmrs.event;
+
+import java.util.Set;
+
+/**
+ * See {@link EntityEvent} or {@link AggregatedEntityEvent}.
+ * 
+ * @since 2.9.x
+ */
+public abstract class BaseSessionEvent extends BaseEvent {
```
rkorytkowski, line 19: Moving sessionId to BaseSessionEvent since BaseEvent is now extended by CDCEvent, for which sessionId does not make any sense as it is published by an asynchronous task.

api/src/main/java/org/openmrs/event/CDCEvent.java
```diff
@@ -0,0 +1,162 @@
...
+ * 
+ * @param <T> entityType associated with this event
+ * 
+ * @since 2.9.x
+ */
+public class CDCEvent<T> extends BaseEvent implements ResolvableTypeProvider {
+	
+	private boolean shapshot = false;
+	
+	private Class<T> entityType;
```
rkorytkowski, line 49: It makes it easier to listen to events for certain entity types without a need for conditions in code.

api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveTest.java
```diff
@@ -9,102 +9,8 @@
...
-import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.context.ApplicationContext;
-import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.test.annotation.Rollback;
-import org.springframework.test.context.ContextConfiguration;
-import org.springframework.test.context.TestExecutionListeners;
-import org.springframework.test.context.junit.jupiter.SpringExtension;
 import org.springframework.transaction.annotation.Transactional;
-import org.xml.sax.InputSource;
```
rkorytkowski, line 107: Cleanup.

api/src/test/resources/TestingApplicationContext.xml
```diff
@@ -65,4 +67,8 @@
 		<property name="preInterceptors" ref="serviceInterceptors"/>
 		<property name="transactionAttributeSource" ref="transactionAttributeSource"/>
 	</bean>
+
+	<jdbc:initialize-database data-source="dataSource" ignore-failures="DROPS">
+		<jdbc:script location="classpath:create_shedlock_table.sql"/>
+	</jdbc:initialize-database>
```
rkorytkowski, line 73: Fix for tests in modules with custom DB instead of in-memory one.

api/src/main/java/org/openmrs/event/CDCEvent.java
```diff
@@ -0,0 +1,162 @@
...
+ * {@link org.springframework.context.event.EventListener} to a message broker for asynchronous processing.
+ * The latter is more performant, and it can also be used to exchange messages with external systems.
+ * 
+ * @param <T> entityType associated with this event
+ * 
+ * @since 2.9.0
+ */
+public class CDCEvent<T> extends BaseEvent implements ResolvableTypeProvider {
+	
+	private boolean shapshot = false;
```
dkayiwa: **Typo (the one thing I'd block on): `shapshot` → `snapshot`.** This ships in 2.9.0 as public API (`isShapshot()`/`setShapshot()`), and since outbox serialization is field-based, the misspelling also lands in persisted payloads (`{"shapshot":false,...}` — verified locally). Renaming after release would be breaking; renaming now is free — code search shows openmrs-module-debezium doesn't reference it yet.

Related design question: the debezium module currently never calls `setShapshot(...)` (it maps Debezium `op="r"` to `Operation.READ` instead), so as merged the flag can never be `true`. Either wire it to `payload.source.snapshot` in the module or drop it from core.

api/src/main/java/org/openmrs/event/CDCEvent.java
```diff
@@ -0,0 +1,162 @@
...
+	public Map<String, Object> getPreviousState() {
+		return previousState;
+	}
+
+	public void setPreviousState(Map<String, Object> previousState) {
+		this.previousState = previousState;
+	}
+	
+	public enum Operation {
+		READ, CREATE, UPDATE, DELETE
```
dkayiwa: Debezium also emits `op="t"` (truncate) — the module's `getOperation()` currently maps that to a `null` operation. Worth deciding explicitly whether to add `TRUNCATE` here or document that truncates are ignored.

api/src/main/java/org/openmrs/event/CDCEvent.java
```diff
@@ -0,0 +1,162 @@
...
+	public void setPreviousState(Map<String, Object> previousState) {
+		this.previousState = previousState;
+	}
+	
+	public enum Operation {
+		READ, CREATE, UPDATE, DELETE 
+	}
+
+	@Override
+	public @Nullable ResolvableType getResolvableType() {
```
dkayiwa, line 157: Suggest `@JsonIgnore` here (and on the pre-existing `EntityEvent.getResolvableType()`): serializing this class with a plain `ObjectMapper` fails with `InvalidDefinitionException: Direct self-reference leading to cycle (...resolvableType->ResolvableType["componentType"]->...)` — reproduced locally against this branch. It only works through the outbox today because `JobRunrConfig`'s `JacksonJsonMapper` mutates the shared `ObjectMapper` bean to field-based visibility, so getters are never serialized. Since the class javadoc recommends forwarding events to message brokers, the first consumer to use a standard mapper will hit this.

api/src/main/java/org/openmrs/event/CDCEvent.java
```diff
@@ -0,0 +1,162 @@
...
+ * are fetched from DB by a separate process e.g. Debezium. For this reason it does not make sense
+ * to use {@link org.springframework.transaction.event.TransactionalEventListener} to process them.
+ * <p>
+ * If you want to guarantee delivery, please use {@link org.springframework.context.event.EventListener} 
+ * so that an event is processed in the same thread as the publisher and if the processing fails, it will be
+ * re-tried by the publisher. It is recommended only for fast enough listeners so that publisher is not
+ * blocked for too long, and it keeps up with the events queue.
+ * <p>
+ * If you combine {@link org.springframework.context.event.EventListener} with {@link org.springframework.scheduling.annotation.Async}
+ * you may run longer processing, but you loose the ability to re-try if anything fails, and in worst case you may lose the event.
```
dkayiwa: Javadoc typos: "you **loose** the ability" → "lose"; and above: "not happening inside an application transaction **rather are** fetched" → "rather **they are** fetched".

api/src/main/java/org/openmrs/event/BaseSessionEvent.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Set;
+
+/**
+ * See {@link EntityEvent} or {@link AggregatedEntityEvent}.
+ * 
+ * @since 2.9.0
+ */
+public abstract class BaseSessionEvent extends BaseEvent {
+	private static final long serialVersionUID = 1L;
+
```
dkayiwa, line 21: Consistency nit: `BaseEvent` is `Serializable` (via `OutboxableEvent`), but only this class declares `serialVersionUID` — `BaseEvent`, `CDCEvent` and the two `CDCTransaction*Event`s don't. Add it everywhere or drop it here.

Also: double blank line before `getSessionId()`, and the `(String sessionId, Set<String> tags)` constructor has no callers (fine if intended as API).

api/src/main/java/org/openmrs/event/AggregatedEntityEvent.java
```diff
@@ -9,6 +9,7 @@
  */
 package org.openmrs.event;
 
+import java.util.ArrayList;
 import java.util.HashSet;
```
dkayiwa: `java.util.HashSet` is no longer used after this change.

api/src/main/java/org/openmrs/event/EventPublisher.java
```diff
@@ -37,14 +42,17 @@ public EventPublisher(ApplicationEventPublisher delegate, SessionFactory session
...
-			String sessionId = ((SessionImplementor) sessionFactory.getCurrentSession()).getSessionIdentifier().toString();
-			BaseEvent baseEvent = (BaseEvent) event;
-			baseEvent.setSessionId(sessionId);
+		if (event instanceof BaseSessionEvent) {
+			try {
+				String sessionId = ((SessionImplementor) sessionFactory.getCurrentSession()).getSessionIdentifier().toString();
+				BaseSessionEvent sessionEvent = (BaseSessionEvent) event;
+				sessionEvent.setSessionId(sessionId);
+			} catch (HibernateException e) {
+				log.warn("No session bound to the current thread", e);
```
dkayiwa: Minor: this warns with a full stack trace on every session-less publish — if a scheduled task routinely publishes entity events, it gets noisy. Consider `log.debug`, or `warn` without the throwable. Otherwise the catch itself is the right call — it's what lets session events be published from background threads.

api/src/test/java/org/openmrs/test/jupiter/BaseContextSensitiveNonTransactionalTest.java
```diff
@@ -211,6 +212,13 @@ public BaseContextSensitiveNonTransactionalTest() {
 		
 		loadCount++;
 	}
+
+	/**
+	 * @since 2.9.0
+	 */
+	public void ensureDatabaseRunning() {
```
dkayiwa: This overridable method is invoked from the constructor, so a subclass override runs **before subclass instance fields are initialized** — it can only safely reference static state. That's the usual testcontainers idiom, but easy to trip over; suggest documenting the constraint in the javadoc. Also `protected` seems more fitting than `public` for an override hook.

api/src/test/java/org/openmrs/test/Containers.java
```diff
@@ -43,6 +44,31 @@ public static void ensureDatabaseRunning() {
 		}
 	}
 	
+	public static void ensureDatabaseRunning(JdbcDatabaseContainer<?> dbContainer, String database) {
```
dkayiwa, line 47: Unlike the legacy paths, the custom container isn't recorded in this class's static state — so if a default-container test later runs in the same JVM, `ensureDatabaseRunning()` will start a second container and repoint the `database*` system properties mid-suite. Probably fine for homogeneous module suites, but worth a comment. (Also a couple of trailing-whitespace lines crept into this method.)

api/src/test/java/org/openmrs/test/Containers.java
```diff
@@ -43,6 +44,31 @@ public static void ensureDatabaseRunning() {
...
+		if (dbContainer == null) {
+			ensureDatabaseRunning();
+		} else {
+			if (!dbContainer.isRunning()) {
+				dbContainer.start();
+				System.setProperty("databaseUrl", dbContainer.getJdbcUrl());
+				System.setProperty("databaseName", dbContainer.getDatabaseName());
+				System.setProperty("databaseUsername", dbContainer.getUsername());
+				System.setProperty("databasePassword", dbContainer.getPassword());
+				if (database.equals("mariadb")) {
```
dkayiwa: `database.equals("mariadb")` NPEs if `database` is null — prefer `"mariadb".equals(database)`. And `com.mysql.jdbc.Driver` below is the deprecated alias (works on mysql-connector-java 8.0.30 but logs a deprecation notice) — `com.mysql.cj.jdbc.Driver` is the proper name.

api/src/test/java/org/openmrs/test/jupiter/BaseModuleContextSensitiveTest.java
```diff
@@ -23,8 +24,8 @@
  * 
  * @since 2.4.0
  */
-@ContextConfiguration(locations = { "classpath:applicationContext-service.xml", "classpath*:TestingApplicationContext.xml",
-        "classpath*:moduleApplicationContext.xml" }, inheritLocations = false)
-public abstract class BaseModuleContextSensitiveTest extends BaseContextSensitiveTest {
+@Transactional
+@Rollback
+public abstract class BaseModuleContextSensitiveTest extends BaseModuleContextSensitiveNonTransactionalTest {
```
dkayiwa, line 29: Compatibility note: this class no longer extends `BaseContextSensitiveTest`. Moving the implementation up to `BaseContextSensitiveNonTransactionalTest` is binary-compatible and `@Transactional`/`@Rollback` are re-applied, so behavior is preserved — but module test utilities doing `instanceof BaseContextSensitiveTest` (or typed against it) will need adjusting. Worth a release note.

dkayiwa (commented):
Reviewed the full diff plus the surrounding event/outbox/test infrastructure, and verified behavior locally:

- `OutboxEventIT` passes 4/4 on both the base branch and this PR's branch (run locally) — covering POJO events → `PayloadApplicationEvent` wrapping → generic `@EventListener`/`@TransactionalEventListener(AFTER_COMMIT)` matching → aggregator → outbox JSON round-trip → shedlock-backed polling. The new `create_shedlock_table.sql` initializer works, and is strictly more robust than the old inline creation (it also covers `@SkipBaseSetup` and JUnit-4 paths that previously missed the table).
- Round-tripped the new `CDCEvent` through the outbox-configured `ObjectMapper` (works) and a plain `ObjectMapper` (fails — see inline comment on `getResolvableType()`).
- Checked the referenced debezium module commit against this API (`entityClass` may be null for unmapped tables — `getResolvableType()` handles that correctly).

Overall the direction looks right: POJO events per current Spring guidance, `sessionId` separated into `BaseSessionEvent` so CDC events don't carry a meaningless session id, and the `EventPublisher` change is what makes publishing from background threads possible at all. The test-infra changes unblock module-supplied containers.

Beyond the inline comments:

1. **Tests**: the diff adds no core tests for the new API. Two cheap, high-value ones: typed-listener matching for `CDCEvent<T>` (the `ResolvableTypeProvider` dispatch is its main feature), and `EventPublisher` publishing a `BaseSessionEvent` with no session bound (the new catch branch is untested).
2. **Follow-up ticket worth filing (pre-existing, surfaced while verifying):** outbox serialization silently depends on `JobRunrConfig` mutating the shared `ObjectMapper` bean (JobRunr's `JacksonJsonMapper` constructor switches it to field visibility and activates `@class` default typing — which is also the only reason abstract `List<EntityEvent<?>>` elements deserialize at all). If that coupling ever changes, outbox serialization of `ResolvableTypeProvider` events breaks outright; and a context-wide mapper with default typing enabled is a deserialization-gadget hazard if any module reuses the bean on untrusted input. A dedicated, explicitly-configured outbox mapper would remove both risks.
3. 45 files still carry `@since 2.9.x` after this PR fixes 5 — worth a one-shot sweep before 2.9.0 ships.
4. Trivia: `OpenmrsPropertyConfig` is a whitespace-only hunk; `CDCTransactionStartedEvent`/`CDCTransactionCompletedEvent` are identical and could share a small base.

I'd treat the `shapshot` typo as the one blocker while this API is still unreleased; everything else is should-fix or at your discretion.

🤖 Review prepared with [Claude Code](https://claude.com/claude-code)

rkorytkowski:
@dkayiwa addressed all review comments

api/src/main/java/org/openmrs/event/CDCEvent.java
```diff
@@ -0,0 +1,164 @@
...
+	public Map<String, Object> getPreviousState() {
+		return previousState;
+	}
+
+	public void setPreviousState(Map<String, Object> previousState) {
+		this.previousState = previousState;
+	}
+	
+	public enum Operation {
+		READ, CREATE, UPDATE, DELETE, TRUNCATE
```
dkayiwa, line 152: What is TRUNCATE used for?

api/src/main/java/org/openmrs/event/CDCTransactionCompletedEvent.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.openmrs.event;
+
+/**
+ * Issued by the CDC engine upon completed transaction. It can be used to gather
+ * events belonging to the same transaction.
+ * 
+ * @since 2.9.0
+ */
+public class CDCTransactionCompletedEvent extends BaseEvent {
```
dkayiwa, line 18: Do we also need one for transaction rolled back?

api/src/main/java/org/openmrs/event/CDCEvent.java
rkorytkowski: See https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-truncate-events not useful to us at the moment, but added as suggested by AI for completeness.

api/src/main/java/org/openmrs/event/CDCTransactionCompletedEvent.java
rkorytkowski: No CDC for roll back. The DB log contains only committed data.

dkayiwa: Oh i see! 😊

api/src/main/java/org/openmrs/event/CDCEvent.java
dkayiwa: ok 👍

rkorytkowski:
@dkayiwa thanks for reviewing! As soon as this is merged I'll be able to issue a PR with post-commit review fixes for TRUNK-6429. Also have a pending PR for TRUNK-6515 that depends on this one :)

api/src/test/java/org/openmrs/test/Containers.java
rkorytkowski: Addressed.

api/src/main/java/org/openmrs/event/EntityEvent.java
```diff
@@ -45,7 +46,7 @@ public EntityEvent(T entity) {
 	}
  	
 	public EntityEvent(T entity, Set<String> tags) {
-		super(entity, tags);
+		super(tags);
```
dkayiwa, line 49: What do you think about the above sonar suggestions? 😊

api/src/main/java/org/openmrs/event/CDCEvent.java
rkorytkowski: Added `@JsonIgnore` to getResolvableType

api/src/main/java/org/openmrs/event/EntityEvent.java
rkorytkowski: I don't think we ever followed this, but fine with me to start doing that. Technically no difference.

dkayiwa:
As per my initial review, this looks good. So i have merged it to unblock you. In the meantime, i will continue with testing it out by using it for events for the querystore/chartsearchai module. If i experience any issues, i will raise them here.

PR #6157, opened 2026-06-04
TRUNK-5731: Ensure the Lucene index is upgraded

jwnasambu:

I moved OpenmrsUtil.startup(props) and getContextDAO().setupSearchIndex() to execute before the version change
check and Liquibase update (runCoreSetupOnVersionChange()). This ensures that the Lucene search index is properly initialized and upgraded to the current version's format before any Liquibase changes run, which might otherwise trigger Hibernate Search with an outdated index.
     

## Issue I worked on

https://openmrs.atlassian.net/issues?filter=10641&selectedIssue=TRUNK-5731

  No? -> execute above command


jwnasambu:
@claude review

jwnasambu:
@dkayiwa , @ibacher Kindly feel free to review my PR at your convenient time please!

jwnasambu:
@dkayiwa , @ibacher Kindly feel free to review my PR at your convenient time please!

jwnasambu:
@dkayiwa, @ibacher, @wikumChamith , @rkorytkowski Kindly feel free to review my PR at your convenient time please!

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -1040,24 +1040,32 @@ public static synchronized void startup(Properties props)
...
+
+		// opening a session before running the index setup as it adds a proxy privilege, which resolves
+		// through the user context that openSession installs; without it, a fresh boot fails on
+		// Context.getUserContext()
+		openSession();
+		try {
+			getContextDAO().setupSearchIndex();
+		} finally {
+			closeSession();
+		}
```
dkayiwa: This needs to stay after `runCoreSetupOnVersionChange()`. When `search.indexVersion` doesn't match `SEARCH_INDEX_VERSION`, `setupSearchIndex()` mass-indexes through the current mappings, and before Liquibase has run, those mappings select columns the database doesn't have yet. A 2.7.x database stores 7 (master is at 8), and indexing any `ConceptName` loads its concept's `ConceptDatatype`, which now maps `date_changed` and `changed_by`. Only `TRUNK-5836-2025-10-09` in the 3.0.x changelog adds those (`drug.description` from `TRUNK-5749-2025-08-20` does the same to drugs). So if this is merged as-is, upgrading any 2.7.x instance with `auto_update_database=true` (the default in our Docker image) fails in `Context.startup` with "Failed to update the search index" before Liquibase ever runs. I reproduced that step on H2 by dropping those two columns and calling `setupSearchIndex()` with the property at 7:

```
HSEARCH700101: 47 failure(s) occurred during mass indexing. See the logs for details. First failure on entity 'ConceptName#2457': HSEARCH700083: Exception while building document for entity 'ConceptName#2457': Could not prepare statement [Column "CD1_0.CHANGED_BY" not found; ...
```

With the columns in place, the same call rebuilds the index and stores 8. 2.8.x and 2.9.x don't have these columns either, so the next `SEARCH_INDEX_VERSION` bump would break their upgrades the same way.

Running the rebuild first doesn't protect anything on master. None of core's changesets go through Hibernate (the custom ones in `util/databasechange` use the Liquibase JDBC connection, or the mysql client), so Hibernate Search never sees them. Could you put the call back where it was? The extra `openSession()`/`closeSession()` and `ContextStartupTest`, which pins the new order, can go with it.

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -336,7 +336,9 @@ protected void initializeWizardFromResolvedPropertiesIfPresent() {
 		Properties script = new Properties();
 
 		Properties installScript = getInstallationScript();
-		script.putAll(installScript);
+		if (installScript != null) {
+			script.putAll(installScript);
+		}
```
dkayiwa: Could the `InitializationFilter` changes and the two new tests in `InitializationFilterTest` move to their own ticket and PR? None of it is about the search index. I went through it, and apart from this null check, which does fix an NPE when `OPENMRS_INSTALLATION_SCRIPT` points at a classpath resource that doesn't exist, it doesn't change behavior. But turning the `if` into an early return re-indents the whole method, so it shows up here as 122 changed lines for what `git diff --ignore-all-space` shows is 18. In that PR it's probably simpler to have `getInstallationScript()` return the empty `Properties` there, as it already does when a file path doesn't exist, instead of `null`.

dkayiwa (commented):
I don't think reordering can fix TRUNK-5731 on master, because the ticket's exact error still happens there, just with a newer codec. 2.8.x and 2.9.x write Lucene 8 indexes (codec `Lucene87`), master's Hibernate Search 8.3.2 runs on lucene-core 9.12.3, and `lucene-backward-codecs` was dropped in TRUNK-6316. When I pointed master at the `Drug` index from a 2.8.4 server, searching it failed with:

```
An SPI class of type org.apache.lucene.codecs.Codec with name 'Lucene87' does not exist.  You need to add the corresponding JAR file supporting this SPI to your classpath.  The current classpath supports the following names: [Lucene912]
```

`SEARCH_INDEX_VERSION` is 8 on 2.8.x, 2.9.x and master, so `setupSearchIndex()` never rebuilds that index during the upgrade, whichever side of Liquibase it runs on. Bumping the version on its own would make it worse: the default mass indexer fails on the old index with the same codec error, and `updateSearchIndex()` rethrows that, so the upgrade would stop at startup instead. Rebuilding with `massIndexer().dropAndCreateSchemaOnStart(true)` worked against the same index, and with the test dataset loaded it came back with every drug. So I'd suggest this PR bump `SEARCH_INDEX_VERSION` and have the rebuild that `setupSearchIndex()` triggers drop and recreate the index instead of purging it.

PR #6165, opened 2026-06-05
TRUNK-6549: Logging should not be vulnerable to injection attacks

jwnasambu:

I have fixed the log injection vulnerability across several core files by sanitizing user controlled input before it is logged. This prevents attackers from forging log entries by injecting newline characters.

 These is the overview of the changes I made:
   - Created OpenmrsUtil.sanitizeForLogging which replaces carriage return (\r) and line feed (\n) characters with underscores (_).
   - Sanitized systemId in becomeUser and the authenticated user in the logout log statement.
   - Sanitized multiple log statements involving client names, system IDs, usernames, and privileges.
   -  Sanitized the username variable in the getUserByUsername method before it is written to the logs.

## Issue I worked on

https://openmrs.atlassian.net/issues?filter=10641&selectedIssue=TRUNK-6549

  No? -> execute above command


api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -154,7 +155,7 @@ public User getUserByUsername(String username) {
 		List<User> users = query.getResultList();
 
 		if (users == null || users.isEmpty()) {
-			log.warn("request for username '" + username + "' not found");
+			log.warn("request for username '{}' not found", OpenmrsUtil.sanitizeForLogging(username));
```
dkayiwa: **[medium] The fix is incomplete within this same file** — three sibling user-influenced log sinks remain unsanitized:

- [L462](https://github.com/openmrs/openmrs-core/blob/499b9b6ccd26da4ec1cfedf9eb8b027d212b713a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java#L462): `log.info("updating password for {}", u.getUsername());`
- [L492](https://github.com/openmrs/openmrs-core/blob/499b9b6ccd26da4ec1cfedf9eb8b027d212b713a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java#L492): `log.info("Updating secret question and answer for " + u.getUsername());` (string concatenation — worth converting to `{}` while here, like this line was)
- [L716](https://github.com/openmrs/openmrs-core/blob/499b9b6ccd26da4ec1cfedf9eb8b027d212b713a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java#L716): `log.debug("name: " + name);` — the raw search string from the public `UserService.getUsers(String name, …)` API, the most directly user-typed value in this class.

`username` is settable via the public `User.setUsername` and `name` comes straight from the search API, so the same data sanitized here at L158 still forges log lines through those paths.

dkayiwa (commented):
## Code review — 9 findings

Reviewed the full diff at max effort (multi-angle finder pass, each finding independently verified against the PR head and current master). Ranked most-severe first; details are in the inline comments.

| # | Severity | Location | Finding |
|---|----------|----------|---------|
| 1 | medium | `HibernateUserDAO:462/492/716` | Three sibling user-influenced log sinks in this same file remain unsanitized |
| 2 | medium | `OpenmrsUtil:2172` | Blacklist misses other injection-capable chars (VT, FF, NEL, U+2028/9, ESC/ANSI) |
| 3 | low | `UserContext:388` | `proxies` list rendered raw in the same statement whose other arg was sanitized |
| 4 | low | `UserContext:113` | `credentials.getAuthenticationScheme()` arg left unsanitized |
| 5 | low | `UserContext:171` | Raw `systemId` concatenated into exception message that can reach logs |
| 6 | medium (design) | `OpenmrsUtil:2168` | A one-line `%encode{%m}{CRLF}` in the central log4j2 layout would sanitize every log site, past and future |
| 7 | low (perf) | `UserContext:388` | Eager sanitize defeats slf4j lazy `{}` on the AuthorizationAdvice hot path |
| 8 | low | `OpenmrsUtil:2172` | `replaceAll` recompiles a regex `Pattern` per call; literal `replace` is equivalent (suggestion attached) |
| 9 | low | `OpenmrsUtil:2168` | No unit test pins the new security primitive |

Findings 3–5 are mechanism-real but currently need a module/custom-SPI caller to be exploitable; surfaced because the touched statements are exactly the ones being hardened.

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -2157,4 +2157,19 @@ public static String isValidNumericValue(Float value, Obs obs) {
...
+	 *
+	 * @param value the value to sanitize
+	 * @return the sanitized string
+	 * @since 3.0.0
+	 */
+	public static String sanitizeForLogging(Object value) {
+		if (value == null) {
+			return null;
+		}
+		return value.toString().replaceAll("[\n\r]", "_");
```
dkayiwa: **[medium] The blacklist only covers `\n`/`\r` — other injection-capable characters pass through:** vertical tab (U+000B), form feed (U+000C), NEL (U+0085), Unicode line/paragraph separators (U+2028/U+2029), backspace (U+0008), and ESC (U+001B). The CONSOLE appender writes `%m` to `SYSTEM_OUT`, so a terminal tailing the log will honor injected ANSI escape sequences (cursor movement, screen clearing, recoloring adjacent lines) — a residual log-injection vector that CR/LF stripping alone doesn't close.

Consider widening the strip, e.g. `replaceAll("[\\p{Cntrl}\\u2028\\u2029&&[^\\t]]", "_")`, or handling it once at the layout level with log4j2's purpose-built `%encode{%m}{CRLF}` (see the comment on the method signature).

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -384,7 +385,7 @@ public Set<Role> getAllRoles(User user) throws Exception {
 	 * @return true if authenticated user has given privilege
 	 */
 	public boolean hasPrivilege(String privilege) {
-		log.debug("Checking '{}' against proxies: {}", privilege, proxies);
+		log.debug("Checking '{}' against proxies: {}", OpenmrsUtil.sanitizeForLogging(privilege), proxies);
```
dkayiwa: **[low] This statement still logs unsanitized data through its second argument.** `proxies` is rendered via `List.toString()`, and it holds the *raw* strings stored at L286 (`proxies.add(privilege)`) — the same values that get sanitized only for the log line at L285. Note the list can't be sanitized at insertion (equality in `hasPrivilege`/`remove` needs the real names), so wrap the rendering instead:

```java
log.debug("Checking '{}' against proxies: {}", OpenmrsUtil.sanitizeForLogging(privilege), OpenmrsUtil.sanitizeForLogging(proxies));
```

(All in-tree `addProxyPrivilege` callers pass `PrivilegeConstants.*` constants today, so this needs a module passing tainted input to trigger — but since this exact statement is being hardened, the sibling arg should get the same treatment.)

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -109,7 +110,7 @@ public UserContext(AuthenticationScheme authenticationScheme) {
 	 */
 	public Authenticated authenticate(Credentials credentials) throws ContextAuthenticationException {
 
-		log.debug("Authenticating client '{}' with scheme '{}'", credentials.getClientName(),
+		log.debug("Authenticating client '{}' with scheme '{}'", OpenmrsUtil.sanitizeForLogging(credentials.getClientName()),
```
dkayiwa: **[low] The second argument is left unsanitized.** `credentials.getAuthenticationScheme()` is rendered raw in the same statement whose first arg was just wrapped. In-repo implementations return constants (`UsernamePasswordCredentials.SCHEME`), but `Credentials` is a public SPI (since 2.3.0) and module implementations control that string — wrapping both args keeps the statement consistently hardened.

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -163,7 +164,7 @@ public User becomeUser(String systemId) throws ContextAuthenticationException {
 			throw new APIAuthenticationException("You must be a superuser to assume another user's identity");
 		}
 
-		log.debug("Turning the authenticated user into user with systemId: {}", systemId);
+		log.debug("Turning the authenticated user into user with systemId: {}", OpenmrsUtil.sanitizeForLogging(systemId));
```
dkayiwa: **[low] Four lines below, the same `systemId` goes raw into an exception message** (L171):

```java
throw new ContextAuthenticationException("User not found with systemId: " + systemId);
```

Exception messages routinely end up in logs via upstream catch-and-log handlers, so the value sanitized here at L167 can still reach the log unsanitized through `getMessage()`. Worth sanitizing there too while this method is being hardened.

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -2157,4 +2157,19 @@ public static String isValidNumericValue(Float value, Obs obs) {
...
 
+	/**
+	 * Sanitizes a string for logging by replacing carriage return and line feed characters with
+	 * underscores to prevent log injection attacks.
+	 *
+	 * @param value the value to sanitize
+	 * @return the sanitized string
+	 * @since 3.0.0
+	 */
+	public static String sanitizeForLogging(Object value) {
```
dkayiwa, line 2169: **[medium, design] Consider fixing this once at the logging layer instead of per call site.** Every OpenMRS appender already routes through a single pattern: `defaultPattern` in `api/src/main/resources/log4j2.xml` (L18, plus the webapp copy), mirrored by `OpenmrsConstants.DEFAULT_LOG_LAYOUT_PATTERN` and the `log.layout` global property. log4j 2.26.0 (current bom) ships both `%replace{%m}{[\\r\\n]}{_}` and the purpose-built `%encode{%m}{CRLF}` converters — one change there sanitizes **every** log statement, past and future, instead of 11 hand-picked arguments. Per-call-site wrapping is already incomplete in this PR (see the `HibernateUserDAO` comment) and nothing gates future `log.*` calls, so the vulnerability regrows over time.

Notes: stack traces are unaffected (emitted by the implicit throwable converter, not `%m`); installs with a persisted `log.layout` GP value would need a companion liquibase changeset; intentionally multi-line messages get flattened — which is the point of the defense. If the layout route is taken, this helper (and the 11 wraps) can be dropped entirely.

api/src/main/java/org/openmrs/api/context/UserContext.java
dkayiwa: **[low, perf] Wrapping `{}` args defeats slf4j's lazy evaluation on a hot path.** Previously this argument cost nothing when DEBUG was off (production default — `org.openmrs.api` is at INFO); now `sanitizeForLogging(privilege)` runs unconditionally, and `String.replaceAll` compiles a fresh regex `Pattern` on every call. `hasPrivilege` is invoked via `AuthorizationAdvice` for every `@Authorized` service method (AOPConfig binds the advice to every `@Service` class), and `addProxyPrivilege`/`removeProxyPrivilege` (L285/L308) run around those same calls — so this is pure waste on every authorization check, several times per service call.

Microseconds each, but avoidable: guard the hot sites with `if (log.isDebugEnabled())`, use the non-regex `replace(char, char)` form (see the comment on `sanitizeForLogging`), or move sanitization to the log4j2 layout where laziness is preserved automatically.

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
dkayiwa: **[low] `replaceAll` takes a regex and compiles a fresh `Pattern` on every invocation** — for a 2-character literal strip. `String.replace(char, char)` produces byte-identical output with no regex machinery, and returns `this` (zero allocation) when the char is absent, which is the overwhelmingly common case:

```suggestion
		return value.toString().replace('\n', '_').replace('\r', '_');
```

Two smaller notes: the javadoc says `@return the sanitized string` but null input returns `null` (fine for `{}` args — slf4j renders both as `null` — but `String.valueOf(value)` would make the method total and collapse the null branch); `StringUtils.replaceChars(value.toString(), "\n\r", "__")` is the commons-lang3 alternative already imported in this class.

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
dkayiwa, line 2169: **[low] No test pins the new security primitive.** `OpenmrsUtilTest` already exists with the project's naming convention (e.g. `url2file_shouldReturnNullGivenNullParameter`). Worth adding e.g. `sanitizeForLogging_shouldReplaceCarriageReturnAndLineFeedWithUnderscores` and `sanitizeForLogging_shouldReturnNullGivenNullParameter` so a future refactor can't silently weaken the sanitizer — especially since the whole PR rests on this one-liner. (The PR checklist's tests box is unchecked.)

jwnasambu:
@chibongho , @ibacher , @wikumChamith , @dkayiwa kindly help me review this PR at your convenient time please.

dkayiwa:
@jwnasambu did you take a look at the above failing build?

jwnasambu:
@dkayiwa Thanks for the review. I have fixed the build failure.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -141,6 +141,19 @@
...
+
+	<changeSet id="TRUNK-6549-2026-06-07" author="dkayiwa">
+		<preConditions onFail="MARK_RAN">
+			<sqlCheck expectedResult="1">
+				SELECT count(*) FROM global_property WHERE property = 'log.layout' AND property_value = '%p - %C{1}.%M(%L) |%d{ISO8601}| %m%n'
+			</sqlCheck>
+		</preConditions>
+		<comment>Updating log.layout to include centralized log sanitization</comment>
+		<update tableName="global_property">
+			<column name="property_value" value="%p - %C{1}.%M(%L) |%d{ISO8601}| %replace{%m}{[\\p{Cntrl}\\u2028\\u2029&amp;&amp;[^\\t]]}{_}%n"/>
```
dkayiwa: This needs fixing before it merges, because as written it disables the very defense the PR is adding. The `%replace` regex uses doubled backslashes, but log4j2's pattern parser does not unescape them: it reads each `\\` as a literal backslash followed by the literal characters `p{Cntrl}`, `u2028`, and so on, so the class ends up matching the literal characters `\ p { C n t r l } u 2 0 8 9` and never matches a real control character.

I verified this end to end on log4j2 2.26.1: I ran this exact changeset against H2, read the stored `log.layout` value back (liquibase persists it verbatim, doubled backslashes and all), and rendered messages through a `PatternLayout` built from that stored value. Two things go wrong at once. A carriage return or line feed that the correct pattern replaces with an underscore passes straight through here, so the injection this ticket is about is not actually blocked for anything relying on the layout. And ordinary text is corrupted: the string `pCntrl url` in a message comes out as `___t__ ___`, because `p`, `C`, `n`, `r`, `l`, `u` and digits are exactly the literal characters this broken class matches.

And it lands on every install, not a corner case. The core-data snapshot seeds `log.layout` with the old `%p - %C{1}.%M(%L) |%d{ISO8601}| %m%n`, so this precondition matches, the update runs, and `OpenmrsPropertyLookup` then feeds the broken value to the CONSOLE, FILE and MEMORY appenders.

The single-backslash form is the one that works, and `OpenmrsConstants.DEFAULT_LOG_LAYOUT_PATTERN` is already correct: its Java source needs the doubled `\\p{Cntrl}` because a string literal requires it, and the compiler turns that into a single-backslash regex token at runtime. The XML files and this SQL value have no Java string layer, so they must contain single backslashes directly. The simplest fix is to use the same class as the constant's runtime value: `\p{Cntrl}`, the two Unicode line separators U+2028 and U+2029, intersected with `[^\t]`, each metacharacter written with one backslash rather than two.

The same single-backslash fix is needed on the `defaultPattern` in `api/src/main/resources/log4j2.xml` and `webapp/src/main/resources/log4j2.xml`; leave `OpenmrsConstants` as it is. Those two copies also disagree on U+0085 (NEL): the api one includes it, the webapp one and this value do not. Since the constant omits it and `isUnsafe()` already covers the whole U+0080 to U+009F range, I would drop U+0085 everywhere so there is a single definition of what gets stripped.

This slipped through because nothing renders a message through the configured layout in tests. `OpenmrsPropertyLookupTest` only asserts the returned string equals the constant, and the new sanitizer tests only exercise the Java helper. A test that builds a `PatternLayout` from the GP value (and from `DEFAULT_LOG_LAYOUT_PATTERN`) and asserts that a message containing a carriage return and line feed comes out with underscores would have caught this and would keep it from regressing.

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -128,6 +128,8 @@ private OpenmrsUtil() {
 
 	private static Map<Locale, SimpleDateFormat> timeFormatCache = new HashMap<>();
 
+	private static final Pattern LOG_SANITIZATION_PATTERN = Pattern.compile("[\\p{Cntrl}\\u0085\\u2028\\u2029&&[^\\t]]");
```
dkayiwa: This one is not blocking. The `LOG_SANITIZATION_PATTERN` field added here is never referenced anywhere: `sanitizeForLogging` does its own scan through `isUnsafe`, so the compiled `Pattern` is left over from the earlier `replaceAll` implementation. I would drop it rather than leave a second, unused definition of the unsafe set sitting next to `isUnsafe`.

jwnasambu:
@claude review

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
dkayiwa: The regression-test part of this is still open. The new layoutPattern tests build the layout from OpenmrsConstants.DEFAULT_LOG_LAYOUT_PATTERN, which is the one copy of the pattern that was never broken (its doubled backslashes are Java string escapes the compiler already collapses to single ones). The copies that actually carried the bug, the defaultPattern property in the two log4j2.xml files and the value in the liquibase changeset, still have no test, so a reintroduced double backslash there, or any drift from the constant, would sail through CI exactly as it did before. A cheap pin is a test asserting that the pattern extracted from those files equals the constant. If it lives in the api module, two mechanics to watch: api/src/test/resources/log4j2.xml shadows the main config on the test classpath, so the main file needs to be read by path (src/main/resources/log4j2.xml resolves from the module directory under surefire), while the liquibase file can be read as the classpath resource org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml. The webapp copy would need a tiny twin in that module, and it is the copy that ships in WEB-INF/classes of the war, so it is worth including. I am not holding the merge on this: I compared all four copies at the current head and they are byte identical, and rendering them through PatternLayout on log4j 2.26.1 turns CR and LF into underscores and leaves ordinary text untouched. Without the pin, though, the exact bug this thread was about can come back with nothing to catch it.

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -2157,4 +2157,71 @@ public static String isValidNumericValue(Float value, Obs obs) {
 		}
 	}
 
+	/**
+	 * Sanitizes a string for logging by replacing carriage return and line feed characters with
+	 * underscores to prevent log injection attacks.
```
dkayiwa: This javadoc still describes the first cut of the method, which only replaced carriage return and line feed. The method now strips every C0 and C1 control character except tab, plus U+2028 and U+2029. Since this doc is what a future reader will trust when deciding whether something like ESC or NEL is already covered, worth making it state the real contract:

```suggestion
	 * Sanitizes a value for logging by replacing characters that can be used to forge log entries
	 * (all C0 and C1 control characters except tab, plus the U+2028 and U+2029 line separators)
	 * with underscores to prevent log injection attacks.
```

No impact on behavior, so fine to leave if you'd rather not touch it again.

jwnasambu:
@claude review

jwnasambu:
@claude review

jwnasambu:
@chibongho , @ibacher , @wikumChamith , @dkayiwa kindly help me review this PR at your convenient time please.

ibacher:
I closed #6128 because the ticket was a duplicate, but the implemenation was actually more correct than this one. We need to:

* Change the default logging as I indicated
* Update the constant for the default log pattern
* Write a liquibase migration that changes the GP if it's set to the old default log pattern

ibacher:
@jwnasambu I'll leave it to your discretion if you want to continue on this PR or migrate to a new version of the other branch pointing at TRUNK-6549 instead of TRUNK-6550.

jwnasambu:

api/src/main/java/org/openmrs/util/OpenmrsUtil.java
```diff
@@ -2157,4 +2157,72 @@ public static String isValidNumericValue(Float value, Obs obs) {
...
+	/**
+	 * Sanitizes a value for logging by replacing characters that can be used to forge log entries (all
+	 * C0 and C1 control characters except tab, plus the U+2028 and U+2029 line separators) with
+	 * underscores to prevent log injection attacks.
+	 *
+	 * @param value the value to sanitize
+	 * @return the sanitized string
+	 * @since 3.0.0
+	 */
+	public static String sanitizeForLogging(Object value) {
```
ibacher: Can we remove this santizeForLogging helper? I think the updates to the pattern address the changes and that will greatly reduce the churn from this PR.

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -1056,4 +1056,76 @@ public void isValidNumericValue_shouldNotReturnErrorMessageWhenNoReferenceRangeI
...
+	/**
+	 * @see OpenmrsUtil#sanitizeForLogging(Object)
+	 */
+	@Test
+	public void sanitizeForLogging_shouldSanitizeInjectionCapableCharacters() {
+		assertEquals("a_b", OpenmrsUtil.sanitizeForLogging("a\rb"));
+		assertEquals("a_b", OpenmrsUtil.sanitizeForLogging("a\nb"));
+		assertEquals("a_b", OpenmrsUtil.sanitizeForLogging("a\u000Bb")); // VT
+		assertEquals("a_b", OpenmrsUtil.sanitizeForLogging("a\u000Cb")); // FF
+		assertEquals("a_b", OpenmrsUtil.sanitizeForLogging("a\u0085b")); // NEL
```
ibacher: If we're meaning to exclude NEL, we need to change the default pattern we're using to:

`\p{Cntrl}\u0085\u2028\u2029&&[^\t]`

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -173,12 +181,16 @@ public User becomeUser(String systemId) throws ContextAuthenticationException {
...
+			log.debug("Turning the authenticated user into user with systemId: {}",
+			    OpenmrsUtil.applyLogSanitization(systemId));
+		}
 
 		User userToBecome = Context.getUserService().getUserByUsername(systemId);
 
 		if (userToBecome == null) {
-			throw new ContextAuthenticationException("User not found with systemId: " + systemId);
+			throw new ContextAuthenticationException(
+			        "User not found with systemId: " + OpenmrsUtil.sanitizeForLogging(systemId));
```
ibacher: Not sure why we're sanitizing this for logging?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -144,7 +144,7 @@ public void collectionContains_shouldUseEqualsMethodForComparisonInsteadOfCompar
 
 		// sanity check
 		identifiers.add(pi);
-		assertTrue(identifiers.size() == 1, "There should still be only 1 identifier in the patient object now");
+		assertEquals(1, identifiers.size(), "There should still be only 1 identifier in the patient object now");
```
ibacher, line 148: While this and the change on line 533 are correct, they are extraneous to what this PR is doing. Can we remove them?

api/src/test/java/org/openmrs/util/OpenmrsUtilTest.java
```diff
@@ -530,11 +530,11 @@ public void openmrsDateFormat_shouldNotAllowDatesWithInvalidDaysOrMonths() {
 	}
 
 	@Test
-	public void openmrsDateFormat_shouldAllowSingleDigitDatesAndMonths() throws ParseException {
```
ibacher, line 533: This change should also be undone.

api/src/test/java/org/openmrs/logging/OpenmrsPropertyLookupTest.java
```diff
@@ -207,4 +223,137 @@ void lookup_shouldAddAndRemoveProxyPrivilegeWhenFetchingGlobalProperty() {
...
+	@Test
+	void layoutPattern_shouldReplaceControlCharactersWithUnderscores() {
+		String message = "before\nafter";
+		PatternLayout layout = PatternLayout.newBuilder().withPattern(OpenmrsConstants.DEFAULT_LOG_LAYOUT_PATTERN).build();
+		LogEvent event = Log4jLogEvent.newBuilder().setLoggerName("test").setLevel(Level.INFO)
+		        .setMessage(new SimpleMessage(message)).build();
+		String result = layout.toSerializable(event);
+		assertThat(result, containsString("before_after"));
+		assertThat(result, not(containsString("before\n")));
+	}
```
ibacher: So the convention here for naming tests is `<method_name>_<description>`. `lookup_shouldAddAndRemoveProxyPrivilegeWhenFetchingGlobalProperty` runs the `lookup()` method with relevant asserts. None of these methods fit that pattern because they're pinning things that have nothing to do with the `OpenmrsPropertyLookup` class at all. These test belong somewhere else.

api/src/test/java/org/openmrs/logging/OpenmrsPropertyLookupTest.java
```diff
@@ -207,4 +223,137 @@ void lookup_shouldAddAndRemoveProxyPrivilegeWhenFetchingGlobalProperty() {
...
+	 * to the module directory rather than as a classpath resource.
+	 */
+	@Test
+	void log4j2xml_shouldCarryTheSameSanitizingLayoutAsTheDefaultConstant() throws Exception {
+		String xml = Files.readString(Path.of("src/main/resources/log4j2.xml"));
+		String pattern = extractDefaultPattern(xml);
+
+		assertThat(pattern, equalTo(OpenmrsConstants.DEFAULT_LOG_LAYOUT_PATTERN));
+		assertHeaderSanitizes(pattern);
+	}
```
ibacher: These tests that ping the state of files are the kind of slop that Claude likes to suggest. They are not helpful. Please remove this test.

api/src/test/java/org/openmrs/logging/OpenmrsPropertyLookupTest.java
```diff
@@ -207,4 +223,137 @@ void lookup_shouldAddAndRemoveProxyPrivilegeWhenFetchingGlobalProperty() {
...
+
+	/**
+	 * The value written into {@code log.layout} for existing installs must behave like the constant: it
+	 * is persisted verbatim by liquibase, so the same doubled-backslash mistake would be shipped to
+	 * every upgraded database. Read as a classpath resource since nothing on the test classpath shadows
+	 * it.
+	 */
+	@Test
+	void liquibaseChangelog_shouldCarryTheSameSanitizingLayoutAsTheDefaultConstant() throws Exception {
```
ibacher: This test pins things way too specifically and way to narrowly. Ultimately, this is, at best, a test that goes with some part of the Liquibase changelog, not here.

This test ultimately pins conditions on the `liquibase-update-to-latest-3.0.x.xml` class, which happens to be the relevant file at the time. But because of the way our snapshotting system works, that files ceases to be relevant as soon as 3.0.0 is out and we're making changes in 3.1.x.xml. This is why this kinds of pinning tests are not actually helping anyone.

ibacher (commented):
Thanks @jwnasambu. There's still a lot of cruft in here and we should move the tests to a place that makes sense for what they test, i.e., the test of the default logging pattern should be in a class dedicated to that, not in the tests for a loosely-related class. I also think the tests pinning the contents of specific files, while understandable, are not useful tests for us to have.

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -1170,7 +1170,7 @@ public static final Collection<String> CONCEPT_PROPOSAL_STATES() {
 	 *
 	 * @since 2.9.0, 2.8.7
 	 */
-	public static final String DEFAULT_LOG_LAYOUT_PATTERN = "%p - %C{1}.%M(%L) |%d{ISO8601}| %m%n";
+	public static final String DEFAULT_LOG_LAYOUT_PATTERN = "%p - %C{1}.%M(%L) |%d{ISO8601}| %replace{%m}{[\\p{Cntrl}\\u0085\\u2028\\u2029&&[^\\t]]}{_}%n";
```
ibacher: Shouldn't the substitution here just be a space?

```suggestion
	public static final String DEFAULT_LOG_LAYOUT_PATTERN = "%p - %C{1}.%M(%L) |%d{ISO8601}| %replace{%m}{[\\p{Cntrl}\\u0085\\u2028\\u2029&&[^\\t]]}{ }%n";
```

ibacher: Other than this, I think this is good to go.

jwnasambu:
@dkayiwa, @ibacher, @wikumChamith , @rkorytkowski  Kindly feel free to review my PR at your convenient time please!

api/src/test/java/org/openmrs/logging/OpenmrsPropertyLookupTest.java
```diff
@@ -207,4 +207,5 @@ void lookup_shouldAddAndRemoveProxyPrivilegeWhenFetchingGlobalProperty() {
 		contextMock.verify(() -> Context.addProxyPrivilege(PrivilegeConstants.GET_GLOBAL_PROPERTIES));
 		contextMock.verify(() -> Context.removeProxyPrivilege(PrivilegeConstants.GET_GLOBAL_PROPERTIES));
 	}
+
```
dkayiwa: Could you drop this blank line? It's left over from moving the layout tests to `DefaultLogLayoutPatternTest`, and without it the diff in this file is only the three assertions that have to change.

```suggestion
```

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -141,6 +141,19 @@
 			<column name="description" type="varchar(255)"/>
 		</addColumn>
 	</changeSet>
-	
+
```
dkayiwa: Could you put back the tab this blank line has on master? Stripping it is the only deletion in this file, and in an append-only changelog any deletion sends the next reviewer off to check that no shipped changeset was touched. With the tab back, the diff here is additions only.

```suggestion
	
```

PR #6190, opened 2026-06-13
TRUNK-6429: Addressing post commit review

rkorytkowski:

Addresses review comments on https://github.com/openmrs/openmrs-core/pull/6084

Adds a basic dead letter queue for OutboxEvents failing a configurable number of times (default: 16) with API methods to force retry.

## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/event/outbox/OutboxEvent.java
+++ b/api/src/main/java/org/openmrs/event/outbox/OutboxEvent.java
@@ -31,7 +31,7 @@ import java.util.Date;
 /**
  * OutboxEvent persisted in DB.
  * 
- * @since 2.9.x
+ * @since 2.9.0
  */
 @Entity
 @Table(name = "outbox_event")
@@ -40,7 +40,7 @@ public class OutboxEvent extends BaseOpenmrsObject implements Auditable {
 	@Id
 	@GeneratedValue(strategy = GenerationType.IDENTITY)
 	private Integer id;
-
+	
 	@Column(name = "event_type", nullable = false)
 	private String eventType;
 
@@ -179,6 +179,6 @@ public class OutboxEvent extends BaseOpenmrsObject implements Auditable {
 	}
 	
 	public enum Status {
-		PENDING, PROCESSING, COMPLETED
+		PENDING, PROCESSING, COMPLETED, FAILED
 	}
 }
--- a/api/src/main/java/org/openmrs/event/outbox/OutboxEventService.java
+++ b/api/src/main/java/org/openmrs/event/outbox/OutboxEventService.java
@@ -23,9 +23,9 @@ import java.util.Date;
 import java.util.List;
 
 /**
- * It's internal and not for use in modules.
+ * API for OutboxEvent.
  * 
- * @since 2.9.x
+ * @since 2.9.0
  */
 @Service
 public class OutboxEventService {
@@ -34,10 +34,32 @@ public class OutboxEventService {
 	
 	private final SessionFactory sessionFactory;
 	private final Duration listenerTimeout;
+	private final int retryLimit;
 	
-	public OutboxEventService(SessionFactory sessionFactory, @Value("${outboxevent.listener.timeout:120}") int listenerTimeout) {
+	public OutboxEventService(SessionFactory sessionFactory, @Value("${outboxevent.listener.timeout:120}") int listenerTimeout,
+			@Value("${outboxevent.retry.limit:16}") int retryLimit) {
 		this.sessionFactory = sessionFactory;
 		this.listenerTimeout = Duration.ofSeconds(listenerTimeout);
+		this.retryLimit = retryLimit;
+	}
+
+	/**
+	 * Reset retry count for outbox event to continue retrying.
+	 *
+	 * @param uuid
+	 */
+	@Transactional
+	public void retryFailedOutboxEvent(String uuid) {
+		org.hibernate.query.Query<OutboxEvent> query = sessionFactory.getCurrentSession().createQuery("from OutboxEvent where uuid = :uuid and status = 'FAILED'", OutboxEvent.class);
+		query.setParameter("uuid", uuid);
+		OutboxEvent event = query.uniqueResult();
+
+		if (event != null) {
+			event.setStatus(OutboxEvent.Status.PENDING);
+			event.setErrorCount(0);
+			event.setErrorMessage(null);
+			sessionFactory.getCurrentSession().save(event);
+		}
 	}
 
 	/**
@@ -47,15 +69,17 @@ public class OutboxEventService {
 	public void resetStuckEvent() throws OutboxException {
 		Date threshold = Date.from(Instant.now().minus(listenerTimeout));
 		Query resetStuckQuery = sessionFactory.getCurrentSession().createQuery(
-			"update OutboxEvent set status = 'PENDING', errorCount = coalesce(errorCount, 0) + 1, " +
-				"errorMessage = :errorMessage, dateChanged = :now where status = 'PROCESSING' and dateChanged < :threshold");
+			"update OutboxEvent set status = case when coalesce(errorCount, 0) + 1 >= :retryLimit then 'FAILED' else 'PENDING' end, " +
+				"errorCount = coalesce(errorCount, 0) + 1, errorMessage = :errorMessage, dateChanged = :now " +
+				"where status = 'PROCESSING' and dateChanged < :threshold");
 		resetStuckQuery.setParameter("now", new Date());
 		resetStuckQuery.setParameter("threshold", threshold);
+		resetStuckQuery.setParameter("retryLimit", retryLimit);
 		resetStuckQuery.setParameter("errorMessage", "Stuck in PROCESSING state for more than " + listenerTimeout.getSeconds() + " seconds");
 		int resetCount = resetStuckQuery.executeUpdate();
 		if (resetCount > 0) {
 			// Throw an error to increase visibility of the event
-			throw new OutboxException("Reset stuck outbox item from PROCESSING back to PENDING");
+			throw new OutboxException("Reset stuck outbox item(s) from PROCESSING back to PENDING or FAILED");
 		}
 	}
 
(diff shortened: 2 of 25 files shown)


api/src/main/java/org/openmrs/event/outbox/OutboxEvent.java
```diff
@@ -88,6 +88,12 @@ public void setId(Integer id) {
 		this.id = id;
 	}
 
+	@Column(name = "uuid", nullable = false)
+	@Override
+	public String getUuid() {
+		return super.getUuid();
+	}
```
dkayiwa: suggestion (non-blocking): This `getUuid()` override looks redundant. `uuid` is already mapped by `BaseOpenmrsObject`, a `@MappedSuperclass` whose `uuid` field carries `@Column(name = "uuid", unique = true, nullable = false, length = 38, updatable = false)`. Since `@Id` is on the `id` field, this entity uses FIELD access, so Hibernate ignores a `@Column` placed on a getter — the green build (no duplicate-mapping error) and the passing `where uuid = :uuid` query both confirm uuid is mapped and queryable without this override.

So it adds nothing today, and it's slightly risky: if the entity ever switched to property access, this definition would override the superclass's and silently drop `unique = true`, `length = 38`, and `updatable = false`. I'd remove the five lines unless they're working around something I'm not seeing.

dkayiwa (commented):
**Merge** — no blocking issues. Clean cleanup/enhancement of the unreleased 2.9.0 outbox feature. Two non-blocking items, both inline:

- `OutboxEvent.getUuid()` adds a redundant `@Column(name = "uuid")` — already mapped by the `BaseOpenmrsObject` `@MappedSuperclass` and ignored under field access. Recommend removing.
- The persisted payload format gained a `@class` type id — `question` to confirm no SNAPSHOT data depends on the old format and that `Id.CLASS` was deliberate.

Minor (take or leave): `@since` was normalized to `2.9.0` across the outbox package, but the six sibling event classes (`SaveServiceEvent`, `VoidServiceEvent`, `UnvoidServiceEvent`, `RetireServiceEvent`, `UnretireServiceEvent`, `PurgeServiceEvent`) still read `2.9.x`; the new public `OutboxEventService` methods lack `@since` (and `getFailingEvents` has no Javadoc) now that the class is documented as public API.

Needs action before merge: none.

api/src/main/java/org/openmrs/event/outbox/OutboxableEvent.java
```diff
@@ -9,11 +9,14 @@
...
+
 import java.io.Serializable;
 
 /**
  * Marker interface for events that could be persisted in the transactional outbox.
  * 
- * @since 2.9.x
+ * @since 2.9.0
  */
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "@class", defaultImpl = OutboxableEvent.class)
```
dkayiwa: question: This changes the persisted payload format — events now serialize with a `@class` type id (`{"@class":"…","uuid":"…"}` instead of `{"uuid":"…"}`, per the updated `OutboxEventIT`). On replay, `readValue` expects `@class`; a row written before this change has no `@class`, so it falls back to `defaultImpl = OutboxableEvent.class` — an interface Jackson can't instantiate, so deserialization fails hard and the event exhausts its retries into the new `FAILED` state.

For unreleased 2.9.0 with no persisted data this is moot — is that the case, or could SNAPSHOT deployments have existing `outbox_event` rows needing a migration? Also, was `Id.CLASS` (vs `Id.NAME` + registered subtypes) deliberate? Low risk since payloads are system-generated, but it bakes FQCNs into stored data and enables class-name-driven deserialization, so worth a conscious call.

rkorytkowski: Using class is desired as it simplifies subclassing without a need to define subtypes. The events are public API similar to domain model thus should not change. I'll remove the defaultImpl.

api/src/main/java/org/openmrs/event/outbox/OutboxEvent.java
rkorytkowski: I'll remove getUuid() override.

api/src/main/java/org/openmrs/event/outbox/OutboxEventService.java
```diff
@@ -40,6 +40,44 @@ public OutboxEventService(SessionFactory sessionFactory, @Value("${outboxevent.l
...
+		}
+	}
+
+	/**
+	 * Reset retry count for outbox event to continue retrying.
+	 *
+	 * @param uuid
+	 */
+	@Transactional
+	public void resetOutboxEventRetryCount(String uuid) {
```
dkayiwa: question: `resetOutboxEventRetryCount` zeroes `errorCount`/`errorMessage` but leaves `status` untouched. For an event that's already `FAILED`, the counter is reset but the status stays `FAILED`, so the poller never re-selects it — `getProcessingAndPendingEvents()` only matches `PENDING`/`PROCESSING` — meaning no actual retry happens, even though the Javadoc says "to continue retrying." Recovering a dead-lettered event currently also requires `retryFailedOutboxEvent(uuid)` to flip it back to `PENDING`, and the cross-reference between the two methods hints that pairing is the intent.

Is that the intended contract? If so, I'd reword this Javadoc to say it only clears the counter and the event must be moved to `PENDING` (via `retryFailedOutboxEvent`) to actually retry. Cleaner for callers would be one-call recovery — e.g. have `retryFailedOutboxEvent` also reset the count, leaving this method for the rarer "clear the counter without changing status" case. Recommended: pick one and document it, since modules are the intended consumers and have only the Javadoc to go on.

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxPollingTaskHandler.java
```diff
@@ -121,23 +125,34 @@ public void execute(OutboxPollingTaskData taskData, TaskContext taskContext) thr
...
+					OutboxExceptionEvent outboxExceptionEvent = new OutboxExceptionEvent(outboxException, item.getUuid());
+					outboxExceptionEvent.setRetryCount(errorCount);
+					
+					if (errorCount < retryLimit) {
+						// Revert to PENDING to retry
+						item.setStatus(OutboxEvent.Status.PENDING);
+						outboxExceptionEvent.setPendingRetry(true);
+					} else {
+						// Stop retries to process the next item
+						item.setStatus(OutboxEvent.Status.FAILED);
```
dkayiwa, line 148: suggestion (non-blocking): This `errorCount >= retryLimit` -> `FAILED` transition is the headline behavior of the PR, but nothing tests it. The IT's `onPatientCreatedFailing` listener throws only 4 times (< the default limit of 16) and then succeeds, so the event reaches `COMPLETED` and this `else` branch is never taken; `OutboxEventServiceTest` exercises retry/reset/`getFailingEvents` only on events it pre-sets to `Status.FAILED`. So neither the boundary (`< retryLimit` vs `>=`) nor the `pendingRetry=false`/`retryCount` carried on the published `OutboxExceptionEvent` is covered.

Worth a test that sets `outboxevent.retry.limit` low (e.g. 1-2) with an always-failing listener and asserts the event lands in `FAILED` and the emitted `OutboxExceptionEvent` has `pendingRetry=false` and the expected `retryCount`.

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxPollingTaskHandler.java
```diff
@@ -67,7 +71,7 @@ public OutboxPollingTaskHandler(OutboxEventRegistry registry, ObjectMapper objec
 	@Override
 	public void execute(OutboxPollingTaskData taskData, TaskContext taskContext) throws Exception {
 		outboxEventService.resetStuckEvent();
```
dkayiwa, line 73: question: `resetStuckEvent()` reverts any event stuck in `PROCESSING` back to `PENDING` and bumps `errorCount` (`coalesce(errorCount,0)+1`), but it never consults `retryLimit`. So a listener that consistently *times out* (hangs rather than throws) keeps getting reset to `PENDING` forever — `errorCount` climbs without bound and the event is never dead-lettered, so under the strict-ordering guarantee it can block the queue indefinitely. Only listeners that *throw* (the `catch` below) ever reach `FAILED`.

Is dead-lettering intentionally limited to throwing listeners, treating timeouts as always-retryable? If a perpetually-stuck event should also be capped, the `retryLimit` check needs to apply on this stuck-reset path too.

dkayiwa (commented):
**Merge.** No blockers — the three findings below are a contract/doc question, a missing test, and a design question, all non-blocking. The one worth settling before modules build on this newly-public API is finding 1 (`resetOutboxEventRetryCount` doesn't re-poll a `FAILED` event). Scope: unreleased 2.9.0 outbox feature, judged as production API.

1. **`resetOutboxEventRetryCount(uuid)` leaves status `FAILED`.** It zeroes the counter but the poller only selects `PENDING`/`PROCESSING`, so for a dead-lettered event no actual retry happens — despite the "to continue retrying" Javadoc. Recovery currently also requires `retryFailedOutboxEvent`. (inline)
2. **The `errorCount >= retryLimit` -> `FAILED` transition has no test** — the headline DLQ behavior. The IT's failing listener throws 4x (< the default 16) then succeeds (-> `COMPLETED`); `OutboxEventServiceTest` only pre-sets `Status.FAILED`. (inline)
3. **`resetStuckEvent()` bypasses `retryLimit`** — it bumps `errorCount` on every stuck revert but never dead-letters, so a listener that consistently times out (rather than throws) is retried forever. Question on intent. (inline)

Needs action before merge: none.

rkorytkowski:
Addressed 1. and 3.

api/src/main/java/org/openmrs/event/outbox/tasks/OutboxPollingTaskHandler.java
rkorytkowski: retryLimit is tested in OutboxEventServiceTest. I don't want to extend already long running OutboxEventIT to reach the retryLimit, which is not easy to adjust in tests.

dkayiwa:
@rkorytkowski this is not a blocker. But just wanted to get your thoughts on it: https://github.com/openmrs/openmrs-core/issues/6195

dkayiwa:
@rkorytkowski what do you also think about this? https://github.com/openmrs/openmrs-core/issues/6197

dkayiwa:
And finally, @rkorytkowski what do you think about this? https://github.com/openmrs/openmrs-core/issues/6199

rkorytkowski:
@dkayiwa are we ready to merge this? I'd like to port changes to master.

dkayiwa:
@rkorytkowski do you have any response to the `resetStuckEvent() reverts any event stuck in PROCESSING back to PENDING and bumps errorCount (coalesce(errorCount,0)+1)` comment above?

rkorytkowski:
It's point 3. Addressed it. See above or line https://github.com/openmrs/openmrs-core/pull/6190/changes#diff-570e86def7c773b1533e5e8cfd22fd10ae82de015292338e31484ac83cc2bbe2R77

PR #6193, opened 2026-06-16
TRUNK-6650: Verify database authentication in startup.sh before starting Tomcat

suubi-joshua:
The `startup.sh` script currently waits for the database TCP port to be available via `wait-for-it.sh`, but does not verify that MariaDB is actually ready to accept authenticated connections. On Galera clusters and container orchestrators, the TCP port can open before InnoDB recovery completes, causing OpenMRS to start against a not-yet-ready database.

This adds a credential-verified check (`mariadb -e "SELECT 1"` with retry) after the TCP port check and before Tomcat starts. The
`mariadb-client` package is installed in the Docker image so the `mariadb` command is available at runtime.

30 retries at 2-second intervals (60s total) are sufficient since `wait-for-it.sh` already confirms the port is open, only the brief InnoDB recovery window (~12s on Galera) needs to pass.

This PR is as a result of this PR https://github.com/openmrs/openmrs-contrib-cluster/pull/14

## Issue I worked on

see https://openmrs.atlassian.net/browse/TRUNK-6650

  No? -> execute above command

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,7 +63,7 @@ RUN mvn $MVN_SETTINGS $MVN_ARGS
 ### Development Stage
 FROM maven:3.9-$DEV_JDK AS dev
 
-RUN apt-get update && apt-get install -y tar gzip git curl && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y tar gzip git curl mariadb-client && rm -rf /var/lib/apt/lists/*
 
 # Setup Tini
 ARG TARGETARCH
@@ -112,7 +112,7 @@ CMD ["/openmrs/startup-dev.sh"]
 ### Production Stage
 FROM tomcat:11-$RUNTIME_JDK
 
-RUN apt-get update && apt-get install -y curl && rm -rf /var/lib/apt/lists/* && rm -rf /usr/local/tomcat/webapps/*
+RUN apt-get update && apt-get install -y curl mariadb-client && rm -rf /var/lib/apt/lists/* && rm -rf /usr/local/tomcat/webapps/*
 
 # Setup Tini
 ARG TARGETARCH
--- a/startup.sh
+++ b/startup.sh
@@ -13,6 +13,27 @@ echo "Waiting for database to initialize..."
 
 /openmrs/wait-for-it.sh -t 3600 -h "${OMRS_DB_HOSTNAME}" -p "${OMRS_DB_PORT}"
 
+echo "Verifying database authentication..."
+if [ "${OMRS_DB}" = "mysql" ] || [ "${OMRS_DB}" = "mariadb" ]; then
+  DB_READY=false
+  for i in $(seq 1 30); do
+    if mariadb -h "${OMRS_DB_HOSTNAME}" -P "${OMRS_DB_PORT}" \
+        -u "${OMRS_DB_USERNAME}" -p"${OMRS_DB_PASSWORD}" \
+        --ssl=0 -e "SELECT 1" > /dev/null 2>&1; then
+      DB_READY=true
+      echo "Database authentication successful."
+      break
+    fi
+    echo "  Database not accepting credentials yet (attempt ${i}/30) — retrying in 2s..."
+    sleep 2
+  done
+
+  if [ "$DB_READY" = false ]; then
+    echo "ERROR: Database did not accept credentials after 30 attempts."
+    exit 1
+  fi
+fi
+
 wait_for_es()
 {
 	IFS=',' read -a es_uris <<< "${OMRS_SEARCH_ES_URIS}"


suubi-joshua:
@dkayiwa

Dockerfile
```diff
@@ -112,7 +112,7 @@ CMD ["/openmrs/startup-dev.sh"]
 ### Production Stage
 FROM tomcat:11-$RUNTIME_JDK
 
-RUN apt-get update && apt-get install -y curl && rm -rf /var/lib/apt/lists/* && rm -rf /usr/local/tomcat/webapps/*
+RUN apt-get update && apt-get install -y curl mariadb-client && rm -rf /var/lib/apt/lists/* && rm -rf /usr/local/tomcat/webapps/*
```
dkayiwa, line 115: issue (blocking): The dev image runs this same check but has no `mariadb` client, so it can't start.

`startup-dev.sh` ends with `source /openmrs/startup.sh` (line 21), so the dev stage executes the new credential check too — but the dev stage's apt-get (line 66) installs only `tar gzip git curl`, and the `maven:3.9-eclipse-temurin-*` base ships no mariadb/mysql client. I reproduced it against a healthy MariaDB: every attempt returns `127` (command-not-found, swallowed by `2>/dev/null`), so after the 30 retries the script prints `ERROR: Database did not accept credentials after 30 attempts.` and exits 1 — Tomcat never starts. This is the default local workflow (`docker-compose.override.yml` sets `target: dev`) and the CI image target (`bamboo.yml --target dev`).

Fix: add `mariadb-client` to the dev stage at line 66:

`RUN apt-get update && apt-get install -y tar gzip git curl mariadb-client && rm -rf /var/lib/apt/lists/*`

(The db-type guard I suggested on `startup.sh` does not cover this — the dev compose uses `OMRS_DB=mariadb`, so the check still runs and still needs the binary present.)

startup.sh
```diff
@@ -13,6 +13,25 @@ echo "Waiting for database to initialize..."
...
+    break
+  fi
+  echo "  Database not accepting credentials yet (attempt ${i}/30) — retrying in 2s..."
+  sleep 2
+done
+
+if [ "$DB_READY" = false ]; then
+  echo "ERROR: Database did not accept credentials after 30 attempts."
+  exit 1
+fi
```
dkayiwa, line 35: issue (blocking): This check hardcodes the `mariadb` client, but `startup.sh` is DB-agnostic — `OMRS_DB=postgresql` deployments can't start.

PostgreSQL is a supported runtime path: runtime-scope `org.postgresql:postgresql` in `api/pom.xml`, the `dbms="postgresql"` Liquibase changesets, and the `postgresql` branch in `startup-init.sh` (which defaults the port to 5432). `wait-for-it.sh` confirms the Postgres port is open, then this check runs the MariaDB client against it. I reproduced it against a healthy Postgres 16: `ERROR 2013 (HY000): Lost connection to server at 'handshake: reading initial communication packet'`, exit 1 — so all 30 attempts fail and the script exits 1 after ~60s even though Postgres is healthy.

Guard the check by DB type so non-MySQL/MariaDB deployments skip it:

```suggestion
echo "Verifying database authentication..."
if [ "${OMRS_DB}" = "mysql" ] || [ "${OMRS_DB}" = "mariadb" ]; then
  DB_READY=false
  for i in $(seq 1 30); do
    if mariadb -h "${OMRS_DB_HOSTNAME}" -P "${OMRS_DB_PORT}" \
        -u "${OMRS_DB_USERNAME}" -p"${OMRS_DB_PASSWORD}" \
        --ssl=0 -e "SELECT 1" > /dev/null 2>&1; then
      DB_READY=true
      echo "Database authentication successful."
      break
    fi
    echo "  Database not accepting credentials yet (attempt ${i}/30) — retrying in 2s..."
    sleep 2
  done

  if [ "$DB_READY" = false ]; then
    echo "ERROR: Database did not accept credentials after 30 attempts."
    exit 1
  fi
fi
```

dkayiwa (commented):
**Needs work** — the new check breaks two paths that run `startup.sh` unmodified, each failing startup after a ~60s delay against a *healthy* database with a misleading "did not accept credentials" error (Findings 1–2). The PR targets the production MySQL/MariaDB image, but `startup.sh` is also sourced by the dev image and runs for PostgreSQL deployments, and neither was accounted for.

**Finding 1 — dev image fails to start (blocking).** `startup-dev.sh` ends with `source /openmrs/startup.sh`, so the dev stage runs this same check, but its apt-get (`Dockerfile:66`) installs only `tar gzip git curl` and the `maven:3.9-eclipse-temurin-*` base ships no mariadb client. Reproduced against a healthy MariaDB: every attempt is `127` (command-not-found, swallowed by `2>/dev/null`), so after 30 retries the script prints `ERROR: Database did not accept credentials after 30 attempts.` and exits 1 — Tomcat never starts. This is the default local dev workflow (`docker-compose.override.yml` → `target: dev`) and the CI image target (`bamboo.yml --target dev`). Fix: add `mariadb-client` to `Dockerfile:66`.

**Finding 2 — PostgreSQL deployments fail to start (blocking).** The check hardcodes the `mariadb` client, but `OMRS_DB=postgresql` is a supported runtime path (runtime-scope `org.postgresql:postgresql` in `api/pom.xml`, the `dbms="postgresql"` Liquibase changesets, and the `postgresql` branch in `startup-init.sh`, which defaults the port to 5432). Reproduced against a healthy Postgres 16: the MariaDB client fails the protocol handshake (`ERROR 2013 (HY000): Lost connection to server at 'handshake: reading initial communication packet'`, exit 1), so all 30 attempts fail and startup exits 1 after ~60s. Fix: guard the check to run only for `mysql`/`mariadb`.

CI is green because the build jobs compile and run the Java suite but never boot a container through `startup.sh` against a DB, so neither failure is exercised by existing checks.

**Needs action before merge:**
1. Install `mariadb-client` in the dev stage — `Dockerfile:66` → `apt-get install -y tar gzip git curl mariadb-client`.
2. Guard the credential check so it runs only for `mysql`/`mariadb` (skip `postgresql` and unknown types).

.github/workflows/dependency-check.yml
```diff
@@ -35,7 +35,7 @@ jobs:
             --failOnCVSS 6.2
             --enableRetired
             --suppression dependency-check-suppressions.xml
-            --nvdApiKey ${{ secrets.NVD_API_KEY }}
+            ${{ secrets.NVD_API_KEY != '' && format('--nvdApiKey {0}', secrets.NVD_API_KEY) || '' }}
```
suubi-joshua, line 38: PRs don't have access to the `NVD_API_KEY` secret, so `${{ secrets.NVD_API_KEY }}` evaluates to an empty string and the CLI crashes with a missing argument error. This makes the CI fail on every fork PR regardless of the change.

This change makes `--nvdApiKey` conditional on the secret being present, so the check runs without an API key (falling back to rate-limited NVD access) when the secret isn't available..

dkayiwa (commented):
**Merge** — nothing blocking remains; one non-blocking item below.

The `.github/workflows/dependency-check.yml` change added since the last round is unrelated to this PR's DB-auth scope (it's a separate fork-CI fix for the empty `--nvdApiKey` argument) and isn't tracked by its own ticket. I've suggested in-thread that it move to its own PR/ticket so this PR stays a single concern.

**Needs action before merge:** none. (Non-blocking: extract the `dependency-check.yml` fix to its own PR/ticket.)

.github/workflows/dependency-check.yml
dkayiwa: suggestion (non-blocking): This `--nvdApiKey` change is unrelated to TRUNK-6650's DB-auth scope and has no ticket of its own. I'd lift it into a separate small PR/ticket and merge that first — it keeps this PR to a single concern (so a DB-auth revert won't drag the CI change along) and gets the fork-CI fix in independently of this review. Non-blocking; if you'd rather keep it here, please note it in the PR description for traceability.

suubi-joshua: Here is the PR https://github.com/openmrs/openmrs-core/pull/6210

rkorytkowski (approved):

rkorytkowski:
@suubi-joshua looks good to me. Please clean up the PR to have just one commit with message matching the PR title: 
TRUNK-6650: Verify database authentication in startup.sh before starting Tomcat

PR #6198, opened 2026-06-16
mergePatients does not reassign Conditions, Allergies, or MedicationDispenses to the surviving patient

### LINKED ISSUE #6194: mergePatients does not reassign Conditions, Allergies, or MedicationDispenses to the surviving patient
## Summary

`PatientService.mergePatients(preferred, notPreferred)` does **not** reassign **Conditions**, **Allergies**, or **MedicationDispenses** from the non-preferred (loser) patient to the preferred (winner) patient. After a merge these records remain attached to the now-**voided** loser, so they silently disappear from the surviving patient's active record. This is a clinical-data-loss bug.

## Environment

- Verified on OpenMRS Platform **2.9.0-SNAPSHOT** (referenceapplication-standalone 3.7.0-SNAPSHOT, Version `2.9.0 SNAPSHOT Build 329a0d`).
- Discovered while building a CQRS read store that consumes the `*ServiceEvent`s from #6084; the issue is in core and is independent of that module.

## Root cause

`PatientServiceImpl.mergePatients` has per-type reassignment helpers for visits, encounters, program enrolments, observations (not in encounters), relationships, identifiers, names, addresses, person attributes, gender, DOB and DOD — but **no helper for Conditions, Allergies, or MedicationDispenses**, which are patient-scoped clinical tables added to core after the merge logic was written.

Encounter-linked clinical data (Diagnoses, Orders, encounter-contained Obs) is reassigned correctly because it rides along when `mergeEncounters` moves the encounter. The gap is specific to **patient-level records that are not encounter-linked and have no merge helper**.

## Reproduction

1. Create two patients, A (preferred) and B (non-preferred).
2. On patient B, add a Condition, an Allergy, and a MedicationDispense (REST: `POST /condition`, `POST /patient/{B}/allergy`, `POST /medicationdispense`).
3. Merge B into A (Administration → Manage Patients → Merge Patients, preferred = A).
4. Observe: B is voided, but the Condition, Allergy and MedicationDispense are **still owned by B**; A has none of them.

## Empirically verified result (each record's owning patient after merge)

| Type | After merge | Why |
|---|---|---|
| Obs | reassigned to winner ✅ | `mergeObservationsNotContainedInEncounters` |
| Encounter / Visit / Program | reassigned ✅ | dedicated helpers |
| Diagnosis | reassigned ✅ | rides with its encounter |
| Order (drug/test/referral) | reassigned ✅ | rides with its encounter |
| **Condition** | **stranded on voided loser ❌** | patient-level, no helper |
| **Allergy** | **stranded on voided loser ❌** | patient-level, no helper |
| **MedicationDispense** | **stranded on voided loser ❌** | patient-level, no helper |

## Impact

A clinician viewing the merged (surviving) patient would not see, for example, a penicillin allergy or an active condition that belonged to the merged-away record — a patient-safety concern. Any downstream consumer (analytics, FHIR export, event-driven projections) also loses these records, because they are never re-saved during the merge and so emit no events.

## Suggested fix

Add `mergeConditions`, `mergeAllergies`, and `mergeMedicationDispenses` helpers to `PatientServiceImpl.mergePatients` that reassign the loser's records to the winner via the corresponding services (so the changes are audited in `PersonMergeLog` and emit the normal save events), mirroring the existing `mergeObservationsNotContainedInEncounters` / `mergeEncounters` pattern.

dkayiwa:
## Description

`PatientService.mergePatients(preferred, notPreferred)` did **not** reassign **Conditions**, **Allergies**, or **MedicationDispenses** from the non-preferred (loser) patient to the preferred (winner) patient. After a merge these patient-scoped clinical records stayed attached to the now-voided loser and silently disappeared from the surviving patient's active record — a clinical-data-loss / patient-safety concern (e.g. a hidden penicillin allergy or active condition).

These three tables were added to core *after* the merge logic was written, and unlike encounter-linked clinical data (Diagnoses, Orders, encounter-contained Obs) they do **not** ride along when an encounter is moved — `EncounterServiceImpl.saveEncounter` propagates the patient to Obs/Orders/Diagnoses, but not to MedicationDispenses, and it saves encounter-linked Conditions/Allergies without changing their patient.

## Changes

- Add `mergeConditions`, `mergeAllergies` and `mergeMedicationDispenses` helpers to `PatientServiceImpl.mergePatients`. Each moves every **non-voided** record from the non-preferred to the preferred patient via the corresponding service (`ConditionService.saveCondition`, `PatientService.saveAllergy`, `MedicationDispenseService.saveMedicationDispense`), mirroring the existing `mergeObservationsNotContainedInEncounters` pattern. Saving through the services keeps the changes audited and emits the normal save events.
- **Allergies are deduplicated by allergen.** A patient cannot hold two allergies for the same allergen (`Allergies` enforces this on load), so an allergy whose allergen the preferred patient already records is left behind rather than moved — otherwise the survivor's allergy list would throw `APIException("Duplicate allergens not allowed")` on every subsequent load. (`AllergyValidator`'s duplicate check only runs for *new* allergies, so a moved allergy bypasses it — the explicit dedup guard is what protects the survivor.) This mirrors the dedup the existing merge already does for identifiers/names/addresses/relationships. Conditions and MedicationDispenses have no such uniqueness constraint and are moved unconditionally.
- Add `movedConditions`, `movedAllergies` and `movedMedicationDispenses` collections (plus accessors) to `PersonMergeLogData`, and include them in `computeHashValue`, so the moves are recorded in the `PersonMergeLog`.
- Document the new behaviors as `<strong>Should</strong>` doclet tags on the `PatientService.mergePatients` javadoc.

## Result after merge

| Type | Before | After |
|---|---|---|
| Condition | stranded on voided loser ❌ | reassigned to winner ✅ |
| Allergy | stranded on voided loser ❌ | reassigned to winner (deduped by allergen) ✅ |
| MedicationDispense | stranded on voided loser ❌ | reassigned to winner ✅ |

(Obs, Encounter/Visit/Program, Diagnosis and Order were already handled correctly and are unchanged.)

## Testing

Nine new `PatientServiceTest` cases. Each creates the record(s) on the non-preferred patient, merges, and asserts the runtime effect on the survivor **and** the `PersonMergeLog` audit:

Conditions:
- `mergePatients_shouldMoveConditionsFromNonPreferredToPreferredPatient`
- `mergePatients_shouldMoveEncounterLinkedConditionFromNonPreferredToPreferredPatient` — verifies the condition and its encounter both end up on the winner (encounter move + condition copy-on-write composition).
- `mergePatients_shouldNotMoveVoidedConditions`

Allergies:
- `mergePatients_shouldMoveAllergiesFromNonPreferredToPreferredPatient`
- `mergePatients_shouldNotMoveAllergyIfPreferredPatientAlreadyHasSameAllergen` — regression for the duplicate-allergen crash; fails (throws) without the dedup.
- `mergePatients_shouldMoveOnlyAllergensNotAlreadyOnPreferredPatient` — loser shares one allergen with the winner and has a second unique one; the unique one moves and the shared one is skipped in the same merge.
- `mergePatients_shouldNotMoveVoidedAllergies`

MedicationDispenses:
- `mergePatients_shouldMoveMedicationDispensesFromNonPreferredToPreferredPatient`
- `mergePatients_shouldNotMoveVoidedMedicationDispense`

The voided-record tests give the loser both a live and a voided record and assert only the live one moves, so they cannot pass vacuously. All `mergePatients_*` tests (46 total, including the 9 new) plus the `PersonMergeLogData` serialization round-trip test (`PersonServiceTest`) pass.

Fixes #6194

### DIFF
--- a/api/src/main/java/org/openmrs/api/PatientService.java
+++ b/api/src/main/java/org/openmrs/api/PatientService.java
@@ -525,6 +525,15 @@ public interface PatientService extends OpenmrsService {
 	 * <strong>Should</strong> merge non voided person attributes from non preferred to preferred patient
 	 * <strong>Should</strong> merge other non voided observations from non preferred to preferred patient
 	 * <strong>Should</strong> merge other non voided orders from non preferred to preferred patient
+	 * <strong>Should</strong> move conditions from non preferred to preferred patient
+	 * <strong>Should</strong> move encounter linked condition from non preferred to preferred patient
+	 * <strong>Should</strong> not move voided conditions
+	 * <strong>Should</strong> move allergies from non preferred to preferred patient
+	 * <strong>Should</strong> not move allergy if preferred patient already has same allergen
+	 * <strong>Should</strong> move only allergens not already on preferred patient
+	 * <strong>Should</strong> not move voided allergies
+	 * <strong>Should</strong> move medication dispenses from non preferred to preferred patient
+	 * <strong>Should</strong> not move voided medication dispense
 	 * <strong>Should</strong> merge non preferred death date when preferred death date is not null or empty
 	 * <strong>Should</strong> merge non preferred death cause when preferred death cause is not null or empty
 	 * <strong>Should</strong> void non preferred person object
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -27,8 +27,10 @@ import org.openmrs.Allergies;
 import org.openmrs.Allergy;
 import org.openmrs.BaseOpenmrsMetadata;
 import org.openmrs.Concept;
+import org.openmrs.Condition;
 import org.openmrs.Encounter;
 import org.openmrs.Location;
+import org.openmrs.MedicationDispense;
 import org.openmrs.Obs;
 import org.openmrs.Order;
 import org.openmrs.Patient;
@@ -44,9 +46,11 @@ import org.openmrs.User;
 import org.openmrs.Visit;
 import org.openmrs.api.APIException;
 import org.openmrs.api.BlankIdentifierException;
+import org.openmrs.api.ConditionService;
 import org.openmrs.api.DuplicateIdentifierException;
 import org.openmrs.api.EncounterService;
 import org.openmrs.api.InsufficientIdentifiersException;
+import org.openmrs.api.MedicationDispenseService;
 import org.openmrs.api.MissingRequiredIdentifierException;
 import org.openmrs.api.ObsService;
 import org.openmrs.api.PatientIdentifierException;
@@ -61,6 +65,8 @@ import org.openmrs.api.db.PatientDAO;
 import org.openmrs.api.db.hibernate.HibernateUtil;
 import org.openmrs.parameter.EncounterSearchCriteria;
 import org.openmrs.parameter.EncounterSearchCriteriaBuilder;
+import org.openmrs.parameter.MedicationDispenseCriteria;
+import org.openmrs.parameter.MedicationDispenseCriteriaBuilder;
 import org.openmrs.patient.IdentifierValidator;
 import org.openmrs.patient.impl.LuhnIdentifierValidator;
 import org.openmrs.person.PersonMergeLog;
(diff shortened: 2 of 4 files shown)


dkayiwa (commented):
**Merge** — the fix does what #6194 asks: it reassigns the three patient-scoped clinical record types to the surviving patient, and the PR's own changes are sound. The one thing below is a pre-existing latent bug in the method this PR extends — non-blocking.

**(non-blocking, pre-existing) `PersonMergeLogData.computeHashValue()` line 454 guards the wrong field.**

```java
str += (getPriorDateOfBirth() != null) ? getPriorDateOfDeath().toString() : getPriorDateOfDeath();
```

The ternary tests `getPriorDateOfBirth()` but dereferences `getPriorDateOfDeath()`, so it throws an NPE whenever a merge records a prior birthdate but no prior death date — the common case, a living patient whose birthdate was filled in from the loser (`mergeDateOfBirth` always sets `priorDateOfBirth = preferred.getBirthdate()`, while `mergeDateOfDeath` leaves `priorDateOfDeath` null for the living). It doesn't bite today because `computeHashValue()` has no production callers (only `PersonServiceTest` invokes it) and that test's data avoids the combination — but you're already editing this method, and the fix is one line: guard on `getPriorDateOfDeath() != null`, mirroring line 453. It's identical on `2.9.x`, so it isn't from this PR and doesn't block.

**Needs action before merge: none.**

Issue #6196, opened 2026-06-16
purgePatient fails with FK violations when patient has Conditions (purge cascade incomplete for newer clinical tables)


dkayiwa:
## Summary

`purgePatient` (and some related purge paths) fail with foreign-key constraint violations when the patient/record has data in newer clinical tables (Conditions, etc.). The purge cascade was not extended to cover these tables.

## Environment

- OpenMRS Platform **2.9.0-SNAPSHOT** (referenceapplication-standalone 3.7.0-SNAPSHOT, MySQL 8 backend).

## Observed failures

1. **Purge a patient that has a Condition** → 500 with:
   ```
   ConstraintViolationException: Cannot delete or update a parent row: a foreign key constraint fails
   (`conditions`, CONSTRAINT `condition_patient_fk` FOREIGN KEY (`patient_id`) REFERENCES `patient`)
   ```
   `purgePatient` does not cascade-delete the patient's `conditions` rows before deleting the `patient`/`person` row.

2. **Purge a Condition that was previously voided then unvoided** → 500 with:
   ```
   ConstraintViolationException: ... (`conditions`, CONSTRAINT `condition_previous_version_fk` ...)
   ```
   The void/unvoid created a `previous_version` self-reference that blocks the delete.

3. **Purge an Order** → 500 with a FK violation on `delete from orders` (even for a freshly created, never-voided order).

## Reproduction (case 1)

1. Create a patient; add a Condition to them.
2. `DELETE /ws/rest/v1/patient/{uuid}?purge=true` (or call `PatientService.purgePatient`).
3. 500 with the `condition_patient_fk` violation above; the patient is not purged.

## Impact

Patients or records that carry newer-table clinical data cannot be hard-purged. Lower severity than a data-loss bug (purge is an uncommon admin/maintenance operation, and void is the normal path), but it reflects the same underlying gap: clinical tables added after the original lifecycle logic (Conditions, Allergies, MedicationDispenses) are not fully integrated into core's patient-lifecycle cascades (merge — see #6194 — and purge).

## Suggested fix

Extend the purge cascade so `purgePatient` removes dependent rows in the newer clinical tables (conditions, allergies, etc.) before deleting the patient, and have the per-record purge paths handle their self/parent FKs (e.g. `condition.previous_version`, order chains).

dkayiwa:
[...]

PR #6202, opened 2026-06-17
TRUNK-6196: purgePatient fails with FK violations on newer clinical tables

dkayiwa:
## Description

`purgePatient` deleted only the person/patient rows (plus names, addresses and attributes), so a patient carrying data in clinical tables added to core **after** the original lifecycle logic — **conditions, allergies and medication dispenses** — could not be hard-purged: the delete aborted with a foreign-key violation (e.g. `condition_patient_fk`). `purgeCondition` had the same gap on its `condition_previous_version_fk` self reference.

Fixes the purge half of the patient-lifecycle gap described in #6196 (the merge half was #6194 / #6198).

## Changes

- **`HibernatePatientDAO.deletePatient`** now removes the patient's medication dispenses, allergies and conditions before deleting the person/patient rows.
  - Rows are removed through the session (load + `delete`) rather than a bulk HQL delete, so any already-managed instance is marked deleted and cannot be re-flushed against the now-removed row at commit, and so the allergy cascade removes its `allergy_reaction` children.
  - The reads include **voided** rows (unlike the void-filtered service fetch methods) because a purge must remove everything.
  - Condition `previous_version` self references are cleared at the DB level first, so the order the conditions are deleted among themselves cannot trip the self foreign key.
  - All of it runs under `FlushMode.MANUAL` so it cannot auto-flush unrelated transient session state.
- **`HibernateConditionDAO.deleteCondition`** now clears any newer version's `previous_version` reference (created when an edited condition is re-saved) before deleting — on the loaded entities rather than via a bulk update, so a managed referrer cannot re-flush the stale reference.

## Out of scope

- `Diagnosis` (`encounter_diagnosis`) also carries a `patient_id` FK, but its `encounter_id` is **NOT NULL** — a diagnosis can only exist on a patient who also has an encounter, and encounters already block purge (long-standing, unchanged behaviour). So it cannot block purge for any patient that was otherwise purgeable.
- Case 3 from the issue (purging an `Order`) needs a separate reproduction and is not addressed here.

## Tests

- `PatientServiceTest.purgePatient_shouldDeletePatientWithConditionsAllergiesAndMedicationDispenses` — purges a patient with an active condition, a voided condition, an edited condition (previous_version chain), an allergy + reaction, a voided allergy, and a medication dispense; uses an encounter/order-free patient so the delete is actually flushed to the DB.
- `PatientServiceTest.purgePatient_shouldDeletePatientWhenModifiedClinicalDataIsHeldInTheSession` — regression test: a dirty managed condition held in the session across the purge (would throw `StaleStateException` under a bulk delete).
- `ConditionServiceImplTest.purgeCondition_shouldPurgeConditionReferencedByAnotherConditionsPreviousVersion`.

All green: `PatientServiceTest` (183), `ConditionServiceImplTest` (18), `HibernateConditionDAOTest` (6).

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConditionDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConditionDAO.java
@@ -13,6 +13,7 @@ import javax.persistence.TypedQuery;
 import java.util.Arrays;
 import java.util.List;
 
+import org.hibernate.FlushMode;
 import org.hibernate.SessionFactory;
 import org.openmrs.Condition;
 import org.openmrs.Encounter;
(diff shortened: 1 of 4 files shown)


dkayiwa:
Closing without merging. Per discussion, conditions/allergies/medication dispenses should behave **exactly like the existing clinical tables** (encounters, observations, orders, visits), which are not cascade-deleted by `purgePatient` — a patient carrying any of them simply cannot be hard-purged (the void workflow is used instead). The newer tables already behave that way; cascading them would have made them diverge from the established behavior. No code change is warranted.

PR #6205, opened 2026-06-18
TRUNK-6515: Include message broker for publishing / subscribing to external events

rkorytkowski:

It is used by https://github.com/openmrs/openmrs-module-artemis

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6515

  No? -> execute above command

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -91,6 +91,14 @@
 			<artifactId>spring-context-support</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-jms</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>javax.jms</groupId>
+			<artifactId>javax.jms-api</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>org.infinispan</groupId>
 			<artifactId>infinispan-spring5-embedded</artifactId>
 		</dependency>
--- a/api/src/main/java/org/openmrs/serialization/JacksonConfig.java
+++ b/api/src/main/java/org/openmrs/serialization/JacksonConfig.java
@@ -9,8 +9,21 @@
  */
 package org.openmrs.serialization;
 
+import com.fasterxml.jackson.annotation.JsonIdentityInfo;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.annotation.ObjectIdGenerators;
+import com.fasterxml.jackson.databind.BeanDescription;
+import com.fasterxml.jackson.databind.DeserializationConfig;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.KeyDeserializer;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.module.SimpleKeyDeserializers;
+import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.datatype.hibernate5.Hibernate5Module;
+import org.jspecify.annotations.NonNull;
+import org.openmrs.OpenmrsObject;
 import org.openmrs.util.OpenmrsJacksonLocaleModule;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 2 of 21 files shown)


rkorytkowski:
Contains one commit from unmerged PR: https://github.com/openmrs/openmrs-core/pull/6190

rkorytkowski:
Updated with the already merged commit. Now just one commit in the PR.

dkayiwa (commented):
Two of these changes looked like API breaks at first: renaming `OutboxEventPayload` to `EventPayload` (with the package move), and making `OutboxException` unchecked. But since 2.9.0 hasn't shipped yet (we're still on 2.9.0-SNAPSHOT, latest release is 2.8.7), neither affects downstream code.

I also left a couple of minor, non-blocking notes inline.

api/pom.xml
```diff
@@ -90,6 +90,14 @@
...
 			<artifactId>spring-context-support</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-jms</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>javax.jms</groupId>
+			<artifactId>javax.jms-api</artifactId>
+		</dependency>
```
dkayiwa, line 100: Are `spring-jms` and `javax.jms-api` used by core anywhere? I couldn't find any reference to them under `api/src` (no imports, nothing in the XML or resources), so as compile-scope dependencies they'd get pulled into everything that depends on openmrs-api without being needed. Unless core is going to grow its own JMS-based broker, it might be better to declare them in the artemis module instead, or at least mark them `provided` here.

api/src/test/java/org/openmrs/event/broker/BrokerEventListenerTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class BrokerEventListenerTest extends BaseContextSensitiveTest {
+	
+	@Autowired
+	BrokerEventListenerFactory listenerFactory;
+	
+	@Test
+	public void shouldBeRegisteredWithListenerFactory() {
+		assertThat(listenerFactory.getListeners(), contains(allOf(
+			hasProperty("source", equalTo("my-test-source")))));
```
dkayiwa, line 51: This test only checks that the listener gets registered. Could we also publish a `BrokerIncomingEvent` here and assert that the `@BrokerEventListener` method actually receives it (and that a non-matching source or broker is filtered out)? The adapter test calls `processEvent(...)` directly, so right now nothing covers the full path of an event published through Spring and dispatched to the listener.

api/pom.xml
rkorytkowski: I wanted to put it in core so that jms can be used in other modules, e.g. camel that I'm working on right now. Implementers are also free to choose their implementation, it can be Artemis or any other broker supporting jms.

api/src/test/java/org/openmrs/event/broker/BrokerEventListenerTest.java
rkorytkowski: Added tests.

rkorytkowski:
@dkayiwa addressed all review comments. Shall we merge and port it to master?

api/src/main/java/org/openmrs/serialization/JacksonConfig.java
```diff
@@ -9,28 +9,95 @@
...
-	
+
 	@Bean
 	public ObjectMapper objectMapper() {
 		ObjectMapper mapper = new ObjectMapper();
-		mapper.registerModule(new Hibernate5Module());
+		mapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+		
+		Hibernate5Module hibernate5Module = new Hibernate5Module();
+		hibernate5Module.enable(Hibernate5Module.Feature.FORCE_LAZY_LOADING);
```
dkayiwa, line 48: Was the scheduler-side effect of this intended? This retunes the shared `objectMapper()` bean (`FORCE_LAZY_LOADING` here, plus the `@JsonIdentityInfo` mixin and disabling `FAIL_ON_UNKNOWN_PROPERTIES` below), and that bean is what `JobRunrConfig` passes to the scheduler via `new JacksonJsonMapper(objectMapper)`. Since `TaskDefinition` is itself an `OpenmrsObject` and `JobRunrSchedulerService` enqueues `new JobRequestAdapter(taskDefinition, …)`, scheduler job serialization now force-loads the lazy graph off each TaskDefinition (creator/changedBy `User` → roles, person, …) where it didn't before.

`SchedulerServiceIT` covers the enqueue path and stays green, so this isn't crashing — I just want to be sure the cross-cutting change is deliberate. If the richer graph semantics are really only needed for the broker and outbox payloads, I'd lean toward giving those their own mapper (or a `@Qualifier`-ed one) instead of changing the one the scheduler shares. If the scheduler benefits from it too, no change needed.

api/src/main/java/org/openmrs/serialization/JacksonConfig.java
```diff
@@ -9,28 +9,95 @@
...
+				if (OpenmrsObject.class.isAssignableFrom(type.getRawClass())) {
+					return new KeyDeserializer() {
+						@Override
+						public Object deserializeKey(String key, DeserializationContext ctxt) {
+							try {
+								OpenmrsObject obj = (OpenmrsObject) type.getRawClass().getDeclaredConstructor().newInstance();
+								obj.setUuid(key);
+								return obj;
+							} catch (Exception e) {
+								return null;
```
dkayiwa, line 85: Minor, and not a blocker: this catch swallows every exception and hands back a `null` map key. If an `OpenmrsObject`-keyed map ever fails to instantiate a key (say, no accessible no-arg constructor), the entry silently gets a null key — and a second failure would collide on null and overwrite the first, with nothing in the logs to show why. Might be worth at least logging the exception here instead of dropping it silently.

dkayiwa (commented):
Good to merge and port to master once you've confirmed the inline question on the `JacksonConfig` change — the other note is optional.

api/src/main/java/org/openmrs/serialization/JacksonConfig.java
rkorytkowski: This is intended. The initial version may have produced TaskDetails/TaskDefinition that would be serialized without some fields (nulls), which may have been unexpected.

dkayiwa (approved):

api/src/main/java/org/openmrs/serialization/JacksonConfig.java
```diff
@@ -9,28 +9,95 @@
...
+							}
```
dkayiwa, line 86: This key deserializer treats the JSON map-key string as the object's uuid (`obj.setUuid(key)`), but nothing registers a matching key serializer, so on the way out an OpenmrsObject key is written via `toString()`. I confirmed the round-trip against this mapper: a `Map<Patient, String>` serializes to `{"Patient#1":"hello"}` and comes back with the key's uuid set to `"Patient#1"` instead of the original. The value survives; the key is silently corrupted, and since the outbox serializes and later deserializes with this same mapper, an event payload carrying an OpenmrsObject-keyed map would hit it.

Either register a key serializer that emits `getUuid()` to mirror the deserializer (and add a round-trip test, since nothing exercises the map-key path today), or drop the deserializer if such maps aren't a payload shape you need. Also, the `catch (Exception e) { return null; }` here turns any instantiation failure into a null key with no log, which hides problems rather than surfacing them.

dkayiwa (commented):
Took another pass over the new commit (0c367353). The end-to-end listener tests and the raw-type factory change look good. Two notes on the JacksonConfig rework, inline: the first is a confirmed round-trip bug for OpenmrsObject-keyed maps that I'd resolve if such payloads can occur, the second a non-blocking question about the shared mapper's reach into JobRunr.

api/src/main/java/org/openmrs/serialization/JacksonConfig.java
dkayiwa, line 48: `FORCE_LAZY_LOADING` here changes behavior for every consumer of this shared `objectMapper` bean, not just the outbox. It's also injected into `JobRunrConfig`, so job-argument serialization is affected too. With it on, serializing an entity eagerly pulls its lazy associations; the `@JsonIdentityInfo` mixin avoids infinite recursion, but a full graph can still be large and add queries. Was the JobRunr impact intended, and is forcing full-graph loading what you want for event and job payloads?

PR #6207, opened 2026-06-18
Make OpenmrsPropertyLookup resilient when the service layer is unavailable

dkayiwa:
## What

Makes `OpenmrsPropertyLookup` resilient when the service layer is not (yet) available, fixing a re-entrant logging-initialization crash introduced by #6104 (TRUNK-6392).

## Root cause

When `ServiceContext.<clinit>` is the first code to initialize log4j2, configuring the logger evaluates `${openmrs:logLayout}` / `${openmrs:logLocation}`. `OpenmrsPropertyLookup` resolves those by reaching into the service layer:

```
ServiceContext.<clinit>  (initializes the static logger)
  → log4j2 config → ${openmrs:logLayout}
  → OpenmrsPropertyLookup.lookup → getProperty
  → ConfigUtil.getGlobalProperty → Context.getAdministrationService()
  → ServiceContext.getInstance() → new ServiceContext()
  → log.debug(...)   ← ServiceContext.log is still null (clinit hasn't finished) ⇒ NPE
```

The lookup only caught `ServiceNotFoundException`, so the `NullPointerException` escaped, aborted log4j configuration **and** `ServiceContext` class initialization. Every subsequent access then failed with `NoClassDefFoundError: Could not initialize class org.openmrs.api.context.ServiceContext`.

This does **not** surface in core's own test run (logging is initialized before that path), but it breaks downstream modules whose unit tests touch `ServiceContext` first — e.g. **openmrs/openmrs-module-webservices.rest**, whose `master` build has been red since the post-#6104 `3.0.0-SNAPSHOT` was deployed (green 2026-06-03, red from 2026-06-12).

## Fix

Broaden the catch in `getProperty` to any `RuntimeException` (covers both `ServiceNotFoundException` and the re-entrant init error — both are unchecked) and fall back to the hardcoded default instead of returning `null`. Logging configuration must tolerate the service layer being unavailable or mid-initialization.

## Tests

`OpenmrsPropertyLookupTest`: added regression tests where `ConfigUtil.getGlobalProperty` throws — `logLayout` falls back to the default pattern, `logLocation` returns `null`, and the proxy privilege is still removed. These fail on the current code (the exception propagates) and pass with the fix. `mvn -pl api test -Dtest=OpenmrsPropertyLookupTest` → 17/17.

Final confirmation that the rest module goes green will come from re-running its CI against the redeployed `3.0.0-SNAPSHOT`.

## Notes
- Please link the corresponding `TRUNK-` ticket (this is a follow-up regression fix for TRUNK-6392 / #6104).

https://claude.ai/code/session_01GswaapaA8WAbd7V7dv3yxW

### DIFF
--- a/api/src/main/java/org/openmrs/logging/OpenmrsPropertyLookup.java
+++ b/api/src/main/java/org/openmrs/logging/OpenmrsPropertyLookup.java
@@ -14,7 +14,6 @@ import org.apache.logging.log4j.core.config.plugins.Plugin;
 import org.apache.logging.log4j.core.lookup.AbstractLookup;
 import org.apache.logging.log4j.core.lookup.StrLookup;
 import org.apache.logging.log4j.status.StatusLogger;
-import org.openmrs.api.ServiceNotFoundException;
 import org.openmrs.api.context.Context;
 import org.openmrs.util.ConfigUtil;
 import org.openmrs.util.OpenmrsConstants;
@@ -77,10 +76,14 @@ public class OpenmrsPropertyLookup extends AbstractLookup {
 			Context.addProxyPrivilege(PrivilegeConstants.GET_GLOBAL_PROPERTIES);
 			try {
 				value = ConfigUtil.getGlobalProperty(propertyName);
-			} catch (ServiceNotFoundException e) {
-				StatusLogger.getLogger().error("An exception was thrown while trying to get the \"{}\" global property",
+			} catch (RuntimeException e) {
+				// Resolving a global property requires the service layer, which may be unavailable while
+				// the logging system is being configured - in particular before, or re-entrantly during,
+				// ServiceContext initialization. Logging configuration must be resilient, so swallow the
+				// failure (e.g. ServiceNotFoundException, or a re-entrant initialization error) and fall
+				// back to the default below.
+				StatusLogger.getLogger().warn("Could not read the \"{}\" global property; falling back to the default",
 				    propertyName, e);
-				return null;
 			} finally {
 				Context.removeProxyPrivilege(PrivilegeConstants.GET_GLOBAL_PROPERTIES);
 			}
--- a/api/src/main/java/org/openmrs/logging/OpenmrsConfigurationFactory.java
+++ b/api/src/main/java/org/openmrs/logging/OpenmrsConfigurationFactory.java
@@ -174,6 +174,11 @@ public class OpenmrsConfigurationFactory extends ConfigurationFactory {
 					StatusLogger.getLogger()
 					        .debug("AdministrationService is not yet available; skipping log-level overrides");
 				}
+			} catch (RuntimeException e) {
+				// The service layer may be unavailable or mid-initialization (e.g. a re-entrant
+				// ServiceContext initialization) while logging is being configured. Logging configuration
+				// must be resilient, so skip the log-level overrides rather than aborting configuration.
+				StatusLogger.getLogger().warn("Could not apply log-level overrides; the service layer is unavailable", e);
 			}
 		}
 	}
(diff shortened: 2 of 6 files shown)


api/src/main/java/org/openmrs/logging/OpenmrsPropertyLookup.java
```diff
@@ -77,10 +76,14 @@ private String getProperty(String propertyName) {
 			Context.addProxyPrivilege(PrivilegeConstants.GET_GLOBAL_PROPERTIES);
 			try {
 				value = ConfigUtil.getGlobalProperty(propertyName);
-			} catch (ServiceNotFoundException e) {
-				StatusLogger.getLogger().error("An exception was thrown while trying to get the \"{}\" global property",
+			} catch (RuntimeException e) {
```
dkayiwa, line 79: **issue (blocking):** Broadening the catch here is correct, but the same re-entrant `NullPointerException` can still escape one step later in the *same* config pass — so this likely relocates the crash rather than fixing the downstream breakage.

`OpenmrsXmlConfiguration.doConfigure()` runs `super.doConfigure()` (where this lookup fires — now handled) and then `doOpenmrsCustomisations(this)`, still inside `ServiceContext.<clinit>` with `ServiceContext.log == null`. When `log.level` is not a system/runtime property — the default, since it is a global property (seeded in core-data) — `doOpenmrsCustomisations` takes the `Context.isSessionOpen()` branch → `applyLogLevels(configuration)` → `ConfigUtil.getGlobalProperty("log.level")` → `getAdministrationService()` → `new ServiceContext()` → `log.debug(...)` → the identical NPE. That site catches only `ServiceNotFoundException` (`OpenmrsConfigurationFactory#doOpenmrsCustomisations`, line 169), so the NPE escapes `doConfigure()` and aborts configuration + `ServiceContext` init exactly as before. The `isSessionOpen()` guard is the same one this lookup just passed, so it holds there too.

Could you confirm the webservices.rest module actually goes green end-to-end, rather than just failing a few lines later at `applyLogLevels`? If it does not, that path needs the same resilience. Note a naive broadening of line 169 won't compile (the body calls `e.getServiceClass()`); add a second clause:

```java
} catch (ServiceNotFoundException e) {
    if (!AdministrationService.class.isAssignableFrom(e.getServiceClass())) {
        throw e;
    }
    StatusLogger.getLogger().debug("AdministrationService is not yet available; skipping log-level overrides");
} catch (RuntimeException e) {
    StatusLogger.getLogger().warn("Could not apply log-level overrides; service layer unavailable", e);
}
```

The same `ServiceNotFoundException`-only pattern is also in `OpenmrsLoggingUtil#applyLogLevels` (line 144) — worth the same treatment for consistency, though that path isn't on the init chain.

dkayiwa (commented):
**Verdict: merge after closing the `log.level` gap (see inline).** The re-entrancy fix is correct, but it hardens only one of the service-layer touches in the log4j2 configuration pass. The `log.level` path — `OpenmrsConfigurationFactory#doOpenmrsCustomisations` → `applyLogLevels` → `getGlobalProperty` — runs in the *same* pass, still inside `ServiceContext.<clinit>` (so `ServiceContext.log` is still null), and still catches only `ServiceNotFoundException`. So the identical NPE escapes there and aborts configuration as before — unless `log.level` is set as a system/runtime property, which by default it isn't (it's a global property). Judged against the PR's goal of un-breaking the webservices.rest module, this likely relocates the crash rather than removing it. Core's own CI stays green because logging is initialized before `ServiceContext` there, so this re-entrant path isn't exercised — the green checks don't cover it.

**Needs action before merge:**
1. Confirm the webservices.rest module actually goes green end-to-end — or apply the same broadened catch to `OpenmrsConfigurationFactory#doOpenmrsCustomisations` (line 169), and ideally `OpenmrsLoggingUtil#applyLogLevels` (line 144), so the re-entrant NPE can't escape via the log-level path. Mechanism and a suggested patch are in the inline comment.

jwnasambu:
@claude review

PR #6224, opened 2026-06-24
Harden Velocity message templating against server-side template injection

dkayiwa:
## Summary

`VelocityMessagePreparator` initialised its `VelocityEngine` with the default `UberspectImpl`, which lets a template reflectively navigate from any context object to arbitrary classes — the classic Velocity SSTI primitive `$obj.class.forName('java.lang.Runtime').getRuntime().exec(...)`. I verified empirically that with the default engine a template can obtain a live `java.lang.Runtime`, and that `SecureUberspector` blocks it.

This PR hardens that path as **defence in depth**. To be clear about scope: **no current OpenMRS Core entry point feeds untrusted input into `prepareMessage()`.** `prepareMessage(...)` has no callers outside the service itself, `Template#setTemplate(...)` is never called with request data, and the only real `MessageService` consumers (`UserServiceImpl#setUserActivationKey`, `AlertReminderTask`) use server-side strings. So this guards against a future regression rather than fixing a reachable vulnerability.

## Changes

1. **`VelocityMessagePreparator`** — configure the engine with `SecureUberspector` via `RuntimeConstants.UBERSPECT_CLASSNAME`. This blocks the reflective method calls used to load classes / obtain `Runtime`/`ProcessBuilder`, while leaving ordinary `$variable` substitution intact. (Using the constant rather than a string literal avoids a silently-ineffective config if the property key ever changes.)
2. **`MessageService`** — gate the two `prepareMessage(...)` overloads (the only paths that drive the template engine) behind `@Authorized(PrivilegeConstants.MANAGE_ALERTS)`, consistent with the sibling `AlertService`. `sendMessage(...)` is intentionally left unannotated because the unauthenticated password-reset flow (`UserServiceImpl#setUserActivationKey`) depends on it.
3. **`VelocityMessagePreparatorTest`** — new test asserting (a) a template cannot reflect its way to `java.lang.Runtime`, and (b) normal variable substitution still works.

## Testing

- `VelocityMessagePreparatorTest` — both tests pass. The reflection test renders `java.lang.Runtime@…` against the old (default) engine and is blocked with `SecureUberspector`, so it is a genuine regression guard, not a tautology.
- `mvn -pl api spotless:check` passes.
- The pre-existing `MessageServiceTest#sendMessage_shouldSendMessage` failure is unrelated (no SMTP server in the build environment) and reproduces on `master` without this change.

### DIFF
--- a/api/src/main/java/org/openmrs/notification/MessageService.java
+++ b/api/src/main/java/org/openmrs/notification/MessageService.java
@@ -15,6 +15,8 @@ import java.util.Map;
 
 import org.openmrs.Role;
 import org.openmrs.User;
+import org.openmrs.annotation.Authorized;
+import org.openmrs.util.PrivilegeConstants;
 
 public interface MessageService {
 
@@ -79,8 +81,22 @@ public interface MessageService {
 	public Message createMessage(String recipients, String sender, String subject, String message, String attachment,
 	        String attachmentContentType, String attachmentFileName) throws MessageException;
 
+	/**
+	 * Prepares a message from a stored template, rendering it with the Velocity engine. Because the
+	 * template body is treated as executable Velocity source, this is gated behind
+	 * {@link PrivilegeConstants#MANAGE_ALERTS} so that untrusted callers cannot drive the template
+	 * engine.
+	 */
+	@Authorized(PrivilegeConstants.MANAGE_ALERTS)
 	public Message prepareMessage(String templateName, Map data) throws MessageException;
 
+	/**
+	 * Prepares a message from the given template, rendering it with the Velocity engine. Because the
+	 * template body is treated as executable Velocity source, this is gated behind
+	 * {@link PrivilegeConstants#MANAGE_ALERTS} so that untrusted callers cannot drive the template
+	 * engine.
+	 */
+	@Authorized(PrivilegeConstants.MANAGE_ALERTS)
 	public Message prepareMessage(Template template) throws MessageException;
 
 	// Template methods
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/openmrs/notification/MessageService.java
```diff
@@ -79,8 +81,22 @@ public interface MessageService {
 	public Message createMessage(String recipients, String sender, String subject, String message, String attachment,
 	        String attachmentContentType, String attachmentFileName) throws MessageException;
 
+	/**
+	 * Prepares a message from a stored template, rendering it with the Velocity engine. Because the
+	 * template body is treated as executable Velocity source, this is gated behind
+	 * {@link PrivilegeConstants#MANAGE_ALERTS} so that untrusted callers cannot drive the template
+	 * engine.
+	 */
+	@Authorized(PrivilegeConstants.MANAGE_ALERTS)
```
dkayiwa, line 90: `prepareMessage` is a public `MessageService` method, so beyond the core callers you traced, any module that calls it through `Context.getMessageService()` as a user without Manage Alerts will now get an `APIAuthenticationException`. Nothing in core reaches that path, and these tests run as a super user (which passes the check regardless), so a broken module consumer wouldn't surface in this build. Is gating message preparation behind the alerts privilege the contract you want for external callers, or would a dedicated privilege fit better? Fine to keep it as-is if the compatibility impact is acceptable; flagging mainly because the reachability analysis in the description is scoped to core and this is a public API.

dkayiwa: Confirming this is intended rather than an accidental break. `prepareMessage` hands the template body to the Velocity engine as executable source, so driving it is a privileged operation by nature, and requiring Manage Alerts of every caller (modules included) is exactly the contract I want here. There are no core callers today, so nothing in-tree changes behavior, and any external consumer that was rendering templates as an unprivileged user is precisely the case this is meant to stop. The SecureUberspector change closes the RCE primitive on its own regardless of caller, so this annotation is the belt-and-suspenders layer, consistent with how the sibling AlertService gates its operations. Keeping it as-is.

PR #6234, opened 2026-06-30
Publish MergePatientServiceEvent after patient merge

### LINKED ISSUE #6195: No dedicated patient-merge event; consumers must rely on the incidental SaveServiceEvent<PersonMergeLog>
## Summary

There is no dedicated **patient-merge** domain event. Consumers that need to react to a merge (read-store/CDC projections, cache invalidation, analytics) must depend on an incidental side effect rather than a first-class signal.

## Environment

- OpenMRS Platform **2.9.0-SNAPSHOT**.
- Context: the `OpenmrsServiceEventAdvice` added in #6084 emits typed `*ServiceEvent`s for `save* / create* / void* / unvoid* / retire* / unretire* / purge*` methods of `OpenmrsService`s.

## Problem

`PatientService.mergePatients` matches none of those method-name patterns, so **no merge event is published**. A merge currently surfaces only as a collection of incidental events:

- `VoidServiceEvent` for the loser patient (`voidPatient`),
- per-record `SaveServiceEvent`s for the types that are reassigned through the proxy (encounters, visits, programs, non-encounter obs),
- a `SaveServiceEvent<PersonMergeLog>` from the final `savePersonMergeLog(...)` call.

A consumer wanting to handle "a merge happened" has to key off `SaveServiceEvent<PersonMergeLog>` and read `getWinner()` / `getLoser()`. That works, but it is a dependency on an implementation detail (that the merge log is persisted through the service proxy) rather than a documented contract — a future refactor of how the merge log is saved would silently break every such consumer, with no event and no error.

## Suggested fix

Publish a first-class merge event — e.g. a `MergeServiceEvent` carrying the winner and loser `Person`/`Patient`, fired from `mergePatients` after the transaction's work is complete — or extend `OpenmrsServiceEventAdvice` to recognise `mergePatients`. This gives consumers a stable, documented signal instead of relying on the `PersonMergeLog` save.

## Related

Closely related to the data-completeness bug where `mergePatients` does not reassign Conditions/Allergies/MedicationDispenses (#6194) — fixing that also makes per-record save events fire for those types during a merge.

amitmishra11:
This PR was written primarily by Claude Code; I reviewed the change and ran the mergePatients test suite before submitting.

## Problem

`PatientService.mergePatients` doesn't match the `save*/create*/void*/unvoid*/retire*/unretire*/purge*` method-name patterns recognised by `OpenmrsServiceEventAdvice` (added in #6084 / #6190), so no domain event is published for a patient merge. Consumers have to key off the incidental `SaveServiceEvent<PersonMergeLog>` from the final `savePersonMergeLog` call instead of a documented, stable signal - a future refactor of how the merge log is persisted would silently break those consumers.

## Fix

As suggested in the issue and confirmed by @rkorytkowski ("For methods that do not follow our save/create/void/... convention we should not extend our AOP rather publish events directly from such methods. The suggested MergePatientServiceEvent sounds right"):

- Added `MergePatientServiceEvent` (in `org.openmrs.aop.event`, alongside the existing `SaveServiceEvent`/`VoidServiceEvent`/etc.), carrying the winner and loser `Patient`.
- `PatientServiceImpl` now has an `EventPublisher` property (wired in `applicationContext-service.xml`, matching the existing `patientDAO` wiring) and publishes the event manually at the end of `mergePatients`, after `savePersonMergeLog` - i.e. after the transaction's work is complete, as suggested in the issue.
- The list-based `mergePatients(Patient, List<Patient>)` overload already delegates to the single-pair method in a loop, so it gets the event for free.

## Testing

Added `mergePatients_shouldPublishMergePatientServiceEvent` to `PatientServiceTest`, using a small test `@Component`/`@EventListener` (`TestMergePatientServiceEventListener`) to capture and assert the published event's winner/loser, following the existing `TestTransactionalEventAggregator` pattern for test-scoped listener beans.

Ran the full `mergePatients_*` test suite in `PatientServiceTest` - all pass, no regressions.

Fixes #6195

### DIFF
--- a/api/src/main/java/org/openmrs/aop/event/MergePatientServiceEvent.java
+++ b/api/src/main/java/org/openmrs/aop/event/MergePatientServiceEvent.java
@@ -1,0 +1,60 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.aop.event;
+
+import org.openmrs.Patient;
+import org.openmrs.event.BaseSessionEvent;
+
+import java.util.Set;
+
+/**
+ * Published manually by {@link org.openmrs.api.impl.PatientServiceImpl#mergePatients(Patient, Patient)}
+ * after a patient merge completes, since {@code mergePatients} does not match the
+ * save/create/void/unvoid/retire/unretire/purge naming convention recognised by
+ * {@link org.openmrs.aop.OpenmrsServiceEventAdvice}.
+ *
+ * @since 2.9.0
+ */
+public class MergePatientServiceEvent extends BaseSessionEvent {
+	private static final long serialVersionUID = 1L;
+
+	private Patient winner;
+
+	private Patient loser;
+
+	public MergePatientServiceEvent() {
+	}
+
+	public MergePatientServiceEvent(Patient winner, Patient loser) {
+		this(winner, loser, Set.of());
+	}
+
+	public MergePatientServiceEvent(Patient winner, Patient loser, Set<String> tags) {
+		super(tags);
+		this.winner = winner;
+		this.loser = loser;
+	}
+
+	public Patient getWinner() {
+		return winner;
+	}
+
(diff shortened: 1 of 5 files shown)


jwnasambu (commented):
Hi @amitmishra11  Thanks for the contribution Per our [PR Tips](https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477199/Pull+Request+Tips), please include a link to the related Jira issue. This helps us track and review your work more effectively. Thanks!

amitmishra11:
Thank you for the review. This fix originated as GitHub issue #6195 (where @rkorytkowski confirmed the approach). I don't have access to create tickets on the OpenMRS Jira — could you advise whether linking to the GitHub issue is acceptable, or if a Jira ticket is needed? If a Jira ticket is required I'm happy to provide the details so one can be created.

jwnasambu:
@claude review

PR #6233, opened 2026-06-30
Porting TRUNK-6429 and TRUNK-6515 to master

rkorytkowski:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6429
https://issues.openmrs.org/browse/TRUNK-6515

## Checklist: I completed these to help reviewers :)

- [x] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

  No? Unsure? -> [configure your IDE](https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE), format the code and add the changes with `git add . && git commit --amend`

- [x] I have **added tests** to cover my changes. (If you refactored
  existing code that was well tested you do not have to add tests)

  No? -> write tests and add them to this commit `git add . && git commit --amend`

- [x] I ran `mvn clean package` right before creating this pull request and
  added all formatting changes to my commit.

  No? -> execute above command

- [x] All new and existing **tests passed**.

  No? -> figure out why and add the fix to your commit. It is your responsibility to make sure your code works.

- [x] My pull request is **based on the latest changes** of the master branch.

  No? Unsure? -> execute command `git pull --rebase upstream master`

PR #6240, opened 2026-07-01
TRUNK-6429: Fixing authorization check

rkorytkowski:
## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/Context.java
+++ b/api/src/main/java/org/openmrs/api/context/Context.java
@@ -694,6 +694,18 @@ public class Context {
 	}
 
 	/**
+	 * Used by {@link org.openmrs.aop.AuthorizationAdvice} to give access for unauthenticated users when
+	 * running initialization code with proxy privileges. See e.g.
+	 * {@link org.openmrs.event.outbox.tasks.OutboxTaskSchedulerInitializer#afterSingletonsInstantiated()}.
+	 *
+	 * @return true if there are any proxy privileges
+	 * @since 2.9.0
+	 */
+	public static boolean hasProxyPrivileges() {
+		return getUserContext().hasProxyPrivileges();
+	}
+
+	/**
 	 * logs out the "active" (authenticated) user within context
 	 *
 	 * @see #authenticate
--- a/api/src/main/resources/applicationContext-service.xml
+++ b/api/src/main/resources/applicationContext-service.xml
@@ -292,6 +292,8 @@
 	</bean>
 	<!-- /SerializationService setup -->
 
+	<bean id="schedulerServiceTarget" class="org.openmrs.scheduler.jobrunr.JobRunrSchedulerService" autowire="constructor" />
+
 	<bean id="hL7ServiceTarget" class="org.openmrs.hl7.impl.HL7ServiceImpl">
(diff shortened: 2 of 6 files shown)


rkorytkowski:
It fixes the lack of authorization check discovered when porting to master. See https://github.com/openmrs/openmrs-core/pull/6233#discussion_r3498467662

api/src/main/resources/applicationContext-service.xml
```diff
@@ -478,6 +480,12 @@
 		<property name="preInterceptors" ref="serviceInterceptors"/>
 		<property name="transactionAttributeSource" ref="transactionAttributeSource"/>
 	</bean>
+	<bean id="schedulerService" class="org.springframework.transaction.interceptor.TransactionProxyFactoryBean">
+		<property name="transactionManager" ref="transactionManager"/>
+		<property name="target" ref="schedulerServiceTarget"/>
+		<property name="preInterceptors" ref="serviceInterceptors"/>
+		<property name="transactionAttributeSource" ref="transactionAttributeSource"/>
+	</bean>
```
dkayiwa, line 488: This is where authorization enforcement actually gets turned on for the scheduler service, moving it off the unadvised `@Service` bean and onto the interceptor chain. It would be worth pinning this down with a test.

The existing authorization tests in `SchedulerServiceIT` authenticate before every call and only assert the in-method visibility filtering (`hasPrivileges` / `isScheduledBy`), so they pass whether or not the advice is applied. They would not have failed against the old `@Service` wiring, so they do not actually lock in this fix.

A test that logs out and asserts an unauthenticated scheduler call throws `APIAuthenticationException` (or that a user without Manage Scheduler is rejected by one of the legacy `@Authorized({"Manage Scheduler"})` methods) would fail on the previous wiring and pass with this change. Without it, re-introducing `@Service` on `JobRunrSchedulerService` later (for example when reconciling 2.9.x with master) would silently drop authorization again with the whole suite still green.

rkorytkowski: Added a test.

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -693,6 +693,18 @@ public static boolean isAuthenticated() {
 		}
 	}
 
+	/**
+	 * Used by {@link org.openmrs.aop.AuthorizationAdvice} to give access for unauthenticated users when
+	 * running initialization code with proxy privileges. See e.g.
+	 * {@link org.openmrs.event.outbox.tasks.OutboxTaskSchedulerInitializer#afterSingletonsInstantiated()}.
+	 *
+	 * @return true if there are any proxy privileges
+	 * @since 2.9.0
```
dkayiwa, line 702: These two new `hasProxyPrivileges()` methods (this one and the one on `UserContext`) are tagged `@since 2.9.0`, but the rest of the new code on this branch uses `@since 2.9.x`. Can we make these `2.9.x` too for consistency, unless you intend this branch to cut 2.9.0?

rkorytkowski: I think our convention is to use 2.9.0 and not 2.9.x.

PR #6245, opened 2026-07-02
TRUNK-6674: Accept new SNOMED CT code for Minutes duration unit (2.8.x backport)

dkayiwa:
Backport of #6242 to the 2.8.x branch, shipping in 2.8.8.

dev3, where the failure was reported, runs platform 2.8.7, so the 2.8 line is the one that actually delivers this fix to that environment.

## Description of what I changed

**Before:** Saving a drug order with duration unit "Minutes" returned HTTP 500 (`Unknown code 1156209001 for SNOMED CT duration units`) because core only recognized the legacy SNOMED CT code `258701004`.

**After:** Core also recognizes `1156209001`, the active SNOMED CT minute concept. SNOMED CT inactivated the legacy code `258701004` in its 2021-07-31 release, and dictionaries such as CIEL now map Minutes concepts (CIEL 1733) to the new code, which is why current dictionaries trigger the failure. The legacy code is still accepted for backward compatibility.

Same change as #6242, applied to this branch's formatting conventions (this branch predates the Spotless migration).

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6674

## Checklist: I completed these to help reviewers :)

- [x] My IDE is configured to follow the code style of this project.
- [x] I have added tests to cover my changes.
- [ ] I ran `mvn clean package` right before creating this pull request and added all formatting changes to my commit.
- [x] All new and existing `DurationTest` tests pass locally (`mvn -pl api -am test -Dtest=DurationTest`); full suite left to CI.
- [x] My pull request is based on the latest changes of the 2.8.x branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6246, opened 2026-07-02
TRUNK-6674: Resolve drug order duration units across all known SNOMED CT and UCUM codes

dkayiwa:

Saving a drug order whose duration units concept carries the current SNOMED CT minute code fails with HTTP 500: `Unknown code 1156209001 for SNOMED CT duration units` (`org.openmrs.Duration:104`). SNOMED CT inactivated the legacy minute code `258701004` ("min", ambiguous) in its 2021-07-31 release, and dictionaries that track SNOMED CT releases now map their minutes concepts to the active code `1156209001`. CIEL's Minutes (1733) today carries SAME-AS mappings to *both* SNOMED codes, plus UCUM `min`.

The 500 is a symptom of three related defects in how core resolves duration units:

1. **Validation and execution disagree.** `SimpleDosingInstructions.validate()` and `DrugOrderValidator` accept a units concept if it has *any* SNOMED CT SAME-AS code, while `Duration.addToDate()` requires the code to be on a hardcoded list. A concept mapped to a real-but-unrecognized code passes validation and then throws at save time, surfacing as an HTTP 500 instead of a validation error.
2. **Resolution is nondeterministic.** `Duration.getCode()` returns the first SNOMED SAME-AS code in mapping-set iteration order. With CIEL's Minutes now carrying two SNOMED SAME-AS codes, whether an order saves depends on which mapping happens to be returned first.
3. **The hardcoded code list ages.** Core only knew the pre-2021 SNOMED codes, so every vocabulary change breaks order saving platform-wide until core chases the new code.

**The fix** introduces one deterministic resolution, `Duration.getDuration(Integer duration, Concept durationUnits)`, shared by auto-expiry derivation and both validators. Callers go straight from the order fields to a `Duration`; reference term codes stay entirely under the hood.

- It considers **all** SAME-AS mappings of the concept to the SNOMED CT and UCUM sources (each matched by its HL7 code or its source name), not just whichever mapping iterates first.
- Codes resolve to a unit through source-scoped registries (one per vocabulary, so a code can never resolve through the wrong source) that also know the current SNOMED CT minute code `1156209001` and the UCUM time codes `s`, `min`, `h`, `d`, `wk`, `mo`, `a`. UCUM codes are stable by design, which makes resolution robust against future SNOMED code churn instead of chasing codes one release at a time.
- SNOMED CT mappings take priority: the first SNOMED CT mapping carrying a known code resolves the concept, and UCUM is consulted only when no SNOMED CT mapping does. Per review, refusing to resolve when SNOMED CT and UCUM mappings disagree would have imposed a consistency rule on data that was valid while UCUM was ignored; with source priority, every concept that resolves today keeps resolving identically however its UCUM mappings are curated, and UCUM strictly extends resolution to concepts that previously resolved to nothing.
- `BaseDosingInstructions.getAutoExpireDate()`, `SimpleDosingInstructions.validate()` and `DrugOrderValidator` all use it, so an unresolvable unit is rejected at validation time with the existing `durationUnits` field error, and auto-expiry derivation can no longer throw for dictionary data. `Duration.getCode()` is deprecated with behavior unchanged (fhir2 still calls it), and `addToDate(...)` keeps accepting bare codes for API compatibility.

**Behavior compatibility:**

- Concepts mapped to the legacy SNOMED codes (existing working deployments): unchanged.
- CIEL-current dictionaries (the reported failure): orders save, deterministically.
- Concepts mapped only to UCUM: previously rejected by validation, now work.
- Concepts whose SNOMED SAME-AS code is unknown to core: previously passed validation and then failed with a 500 at save; now rejected at validation with a meaningful message.

## Verification

- `258701004` inactive since 2021-07-31 and `1156209001` active, confirmed against a SNOMED terminology server (tx.fhir.org `$lookup`).
- dev3 (platform 2.8.7, current CIEL): Minutes is the only duration unit whose SNOMED mapping changed; it carries both SNOMED codes as SAME-AS plus UCUM `min`. Most units in the "Duration units" set (CIEL 1732) also carry their UCUM code; Hours (1822) carries a malformed UCUM mapping (`h57.1`, reported upstream), which resolution skips as designed, so Hours resolves through its still-active SNOMED code and has no UCUM fallback until the CIEL mapping is fixed.
- `DurationTest`, `SimpleDosingInstructionsTest` and `DrugOrderValidatorTest` pass locally, with cases for the current minute code, the UCUM codes, multi-mapping concepts, unknown codes, the SNOMED CT preference when a UCUM mapping disagrees, the UCUM fallback when the only SNOMED CT code is unknown, case-insensitive UCUM source names and source filtering.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6674

- [ ] I ran `mvn clean package` right before creating this pull request and added all formatting changes to my commit.
- [x] All new and existing tests in the touched test classes pass locally; full suite left to CI.

Opened as an independently derived fix at the maintainer's request, without consulting the other open PRs for this report, so the approaches can be compared.

## Alternatives considered

- **Runtime-configurable code registry** (global property or metadata mapping): would let admins add codes without a platform release, but it moves a semantic-correctness decision ("this code means days") from compile-time review into runtime config, where a typo silently changes when medications auto-expire. For codes that have changed once in roughly a decade, that trades a rare, loud, reviewed failure for a rare, silent, config-shaped one. The existing split is kept deliberately: implementations configure *which concepts* are orderable duration units; the platform owns *what the codes mean*.
- **Terminology-server lookup** (ask SNOMED CT whether a code denotes a unit of time): adds a runtime dependency and offline failure modes, and still would not supply the calendar arithmetic each unit needs.
- **UCUM as the only vocabulary**: closest to durable, since UCUM codes are stable by design and FHIR's `Timing.UnitsOfTime` hardcodes the same seven, but every existing dictionary carries SNOMED mappings, so SNOMED stays first-class with UCUM alongside as the stable rail.
- **Repurposing `SNOMED_CT_MINUTES_CODE` to the new value** instead of adding a constant: public String constants inline at compile time, so modules compiled against older core keep the old value until rebuilt and then silently flip behavior; fhir2's translator map is the concrete casualty.

What makes the compiled-in registry safe, and must be preserved when extending it: entries are additive and never removed or repurposed; concepts that resolve to nothing fail as validation errors, never a save-time 500; and two vocabulary rails, with SNOMED CT taking priority and UCUM filling gaps, mean no single SNOMED inactivation is load-bearing. This rationale is also summarized in the `Duration` class Javadoc.

### DIFF
--- a/api/src/main/java/org/openmrs/BaseDosingInstructions.java
+++ b/api/src/main/java/org/openmrs/BaseDosingInstructions.java
@@ -26,7 +26,7 @@ public abstract class BaseDosingInstructions implements DosingInstructions {
 		if (drugOrder.getDuration() == null || drugOrder.getDurationUnits() == null) {
 			return null;
 		}
-		String durationCode = Duration.getCode(drugOrder.getDurationUnits());
+		String durationCode = Duration.getKnownCode(drugOrder.getDurationUnits());
 		if (durationCode == null) {
 			return null;
 		}
--- a/api/src/main/java/org/openmrs/BaseDosingInstructions.java
+++ b/api/src/main/java/org/openmrs/BaseDosingInstructions.java
@@ -26,11 +26,10 @@ public abstract class BaseDosingInstructions implements DosingInstructions {
 		if (drugOrder.getDuration() == null || drugOrder.getDurationUnits() == null) {
 			return null;
 		}
-		String durationCode = Duration.getKnownCode(drugOrder.getDurationUnits());
-		if (durationCode == null) {
+		Duration duration = Duration.getDuration(drugOrder.getDuration(), drugOrder.getDurationUnits());
+		if (duration == null) {
 			return null;
 		}
-		Duration duration = new Duration(drugOrder.getDuration(), durationCode);
 		return aMomentBefore(duration.addToDate(drugOrder.getEffectiveStartDate(), drugOrder.getFrequency()));
 	}
 
--- a/api/src/main/java/org/openmrs/Duration.java
+++ b/api/src/main/java/org/openmrs/Duration.java
@@ -61,6 +61,15 @@ public class Duration {
 	public static final String SNOMED_CT_CONCEPT_SOURCE_HL7_CODE = "SCT";
 
 	/**
+	 * Name identifying the SNOMED CT concept source, used as a fallback for dictionaries that register
+	 * the source by name without setting its HL7 code to {@link #SNOMED_CT_CONCEPT_SOURCE_HL7_CODE},
+	 * mirroring how {@link #UCUM_CONCEPT_SOURCE_NAME} is matched.
+	 *
+	 * @since 3.0.0
+	 */
+	public static final String SNOMED_CT_CONCEPT_SOURCE_NAME = "SNOMED CT";
+
+	/**
 	 * Name identifying the concept source for The Unified Code for Units of Measure
 	 * (<a href="https://ucum.org">ucum.org</a>), whose codes for the units of time are stable by
 	 * design. Dictionaries commonly map duration unit concepts to UCUM in addition to SNOMED CT.
@@ -204,11 +213,25 @@ public class Duration {
 
 	private final Integer duration;
 
+	private final Unit unit;
+
+	/**
+	 * The raw code this duration was constructed from, retained only so {@link #addToDate} can name the
+	 * offending code when it is not one this class recognises. Null when the duration was built from an
+	 * already-resolved {@link Unit}.
+	 */
 	private final String code;
 
 	public Duration(Integer duration, String code) {
 		this.duration = duration;
 		this.code = code;
+		this.unit = findUnit(code);
+	}
+
+	private Duration(Integer duration, Unit unit) {
+		this.duration = duration;
+		this.unit = unit;
+		this.code = null;
 	}
 
 	/**
@@ -220,14 +243,20 @@ public class Duration {
 	 * @return date which is startDate plus duration
 	 */
 	public Date addToDate(Date startDate, OrderFrequency frequency) {
-		Unit unit = SNOMED_CT_UNITS_BY_CODE.get(code);
-		if (unit == null) {
-			unit = UCUM_UNITS_BY_CODE.get(code);
-		}
 		if (unit == null) {
 			throw new APIException("Duration.unknown.code", new Object[] { code });
 		}
 		return unit.addToDate(startDate, this.duration, frequency);
+	}
+
+	/**
+	 * Resolves a bare code to a {@link Unit} the way the legacy {@link #Duration(Integer, String)}
+	 * constructor must: by trying the SNOMED CT then the UCUM code tables, since the constructor
+	 * carries no concept source. Returns null when the code is not recognised.
+	 */
+	private static Unit findUnit(String code) {
+		Unit unit = SNOMED_CT_UNITS_BY_CODE.get(code);
+		return unit != null ? unit : UCUM_UNITS_BY_CODE.get(code);
 	}
 
 	/**
@@ -278,7 +307,6 @@ public class Duration {
 	 * @since 3.0.0
 	 */
 	public static Duration getDuration(Integer duration, Concept durationUnits) {
-		String knownCode = null;
 		Unit knownUnit = null;
 		for (ConceptMap conceptMapping : durationUnits.getConceptMappings()) {
 			if (!ConceptMapType.SAME_AS_MAP_TYPE_UUID.equals(conceptMapping.getConceptMapType().getUuid())) {
@@ -290,19 +318,18 @@ public class Duration {
 				continue;
 			}
 			if (knownUnit == null) {
-				knownCode = conceptReferenceTerm.getCode();
 				knownUnit = unit;
 			} else if (knownUnit != unit) {
 				log.warn("Not resolving duration units concept {} because its SAME-AS mappings denote different units,"
-				        + " e.g. codes {} and {}",
-				    durationUnits.getUuid(), knownCode, conceptReferenceTerm.getCode());
+				        + " e.g. {} and {}",
+				    durationUnits.getUuid(), knownUnit, unit);
 				return null;
 			}
 		}
-		if (knownCode == null) {
+		if (knownUnit == null) {
 			return null;
 		}
-		return new Duration(duration, knownCode);
+		return new Duration(duration, knownUnit);
 	}
 
 	private static Unit findUnit(ConceptSource conceptSource, String code) {
@@ -316,7 +343,8 @@ public class Duration {
 	}
 
 	private static boolean isSnomedCtSource(ConceptSource conceptSource) {
-		return SNOMED_CT_CONCEPT_SOURCE_HL7_CODE.equals(conceptSource.getHl7Code());
+		return SNOMED_CT_CONCEPT_SOURCE_HL7_CODE.equals(conceptSource.getHl7Code())
+		        || SNOMED_CT_CONCEPT_SOURCE_NAME.equalsIgnoreCase(conceptSource.getName());
 	}
 
 	private static boolean isUcumSource(ConceptSource conceptSource) {
--- a/api/src/main/java/org/openmrs/Duration.java
+++ b/api/src/main/java/org/openmrs/Duration.java
@@ -271,7 +271,7 @@ public class Duration {
 	 *             concept's mappings and is not necessarily a code this class can interpret; use
 	 *             {@link #getDuration(Integer, Concept)} instead
 	 */
-	@Deprecated
+	@Deprecated(since = "3.0.0")
 	public static String getCode(Concept durationUnits) {
 		for (ConceptMap conceptMapping : durationUnits.getConceptMappings()) {
 			ConceptReferenceTerm conceptReferenceTerm = conceptMapping.getConceptReferenceTerm();
@@ -309,27 +309,30 @@ public class Duration {
 	public static Duration getDuration(Integer duration, Concept durationUnits) {
 		Unit knownUnit = null;
 		for (ConceptMap conceptMapping : durationUnits.getConceptMappings()) {
-			if (!ConceptMapType.SAME_AS_MAP_TYPE_UUID.equals(conceptMapping.getConceptMapType().getUuid())) {
-				continue;
-			}
-			ConceptReferenceTerm conceptReferenceTerm = conceptMapping.getConceptReferenceTerm();
-			Unit unit = findUnit(conceptReferenceTerm.getConceptSource(), conceptReferenceTerm.getCode());
-			if (unit == null) {
-				continue;
-			}
-			if (knownUnit == null) {
-				knownUnit = unit;
-			} else if (knownUnit != unit) {
-				log.warn("Not resolving duration units concept {} because its SAME-AS mappings denote different units,"
-				        + " e.g. {} and {}",
-				    durationUnits.getUuid(), knownUnit, unit);
-				return null;
+			Unit unit = findUnit(conceptMapping);
+			if (unit != null) {
+				if (knownUnit == null) {
+					knownUnit = unit;
+				} else if (knownUnit != unit) {
+					log.warn("Not resolving duration units concept {} because its SAME-AS mappings denote different"
+					        + " units, e.g. {} and {}",
+					    durationUnits.getUuid(), knownUnit, unit);
+					return null;
+				}
 			}
 		}
 		if (knownUnit == null) {
 			return null;
 		}
 		return new Duration(duration, knownUnit);
+	}
+
+	private static Unit findUnit(ConceptMap conceptMapping) {
+		if (!ConceptMapType.SAME_AS_MAP_TYPE_UUID.equals(conceptMapping.getConceptMapType().getUuid())) {
+			return null;
+		}
+		ConceptReferenceTerm conceptReferenceTerm = conceptMapping.getConceptReferenceTerm();
+		return findUnit(conceptReferenceTerm.getConceptSource(), conceptReferenceTerm.getCode());
 	}
 
 	private static Unit findUnit(ConceptSource conceptSource, String code) {
--- a/api/src/main/java/org/openmrs/Duration.java
+++ b/api/src/main/java/org/openmrs/Duration.java
@@ -9,7 +9,10 @@
  */
 package org.openmrs;
 
+import java.util.Collections;
 import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.openmrs.api.APIException;
 
@@ -22,7 +25,7 @@ import static org.apache.commons.lang3.time.DateUtils.addWeeks;
 import static org.apache.commons.lang3.time.DateUtils.addYears;
 
 /**
- * Duration represented using SNOMED CT duration codes
+ * Duration represented using SNOMED CT or UCUM duration codes
  *
  * @since 1.10
  */
@@ -31,6 +34,15 @@ public class Duration {
 	public static final String SNOMED_CT_SECONDS_CODE = "257997001";
 
 	public static final String SNOMED_CT_MINUTES_CODE = "258701004";
+
+	/**
+	 * The SNOMED CT code for minute that is active since the 2021-07-31 release, in which the legacy
+	 * code {@link #SNOMED_CT_MINUTES_CODE} was inactivated as ambiguous. Dictionaries that track
+	 * current SNOMED CT releases, such as CIEL, map their minutes concepts to this code.
+	 *
+	 * @since 3.0.0
+	 */
+	public static final String SNOMED_CT_MINUTES_CODE_2021 = "1156209001";
 
 	public static final String SNOMED_CT_HOURS_CODE = "258702006";
 
@@ -46,6 +58,50 @@ public class Duration {
 
 	public static final String SNOMED_CT_CONCEPT_SOURCE_HL7_CODE = "SCT";
 
+	/**
+	 * Name identifying the concept source for The Unified Code for Units of Measure
+	 * (<a href="https://ucum.org">ucum.org</a>), whose codes for the units of time are stable by
+	 * design. Dictionaries commonly map duration unit concepts to UCUM in addition to SNOMED CT.
+	 *
+	 * @since 3.0.0
+	 */
+	public static final String UCUM_CONCEPT_SOURCE_NAME = "UCUM";
+
+	/**
+	 * @since 3.0.0
+	 */
+	public static final String UCUM_SECONDS_CODE = "s";
+
+	/**
+	 * @since 3.0.0
+	 */
+	public static final String UCUM_MINUTES_CODE = "min";
+
+	/**
+	 * @since 3.0.0
+	 */
+	public static final String UCUM_HOURS_CODE = "h";
+
+	/**
+	 * @since 3.0.0
+	 */
+	public static final String UCUM_DAYS_CODE = "d";
+
+	/**
+	 * @since 3.0.0
+	 */
+	public static final String UCUM_WEEKS_CODE = "wk";
+
+	/**
+	 * @since 3.0.0
+	 */
+	public static final String UCUM_MONTHS_CODE = "mo";
+
+	/**
+	 * @since 3.0.0
+	 */
+	public static final String UCUM_YEARS_CODE = "a";
+
 	private static final int SECONDS_PER_MINUTE = 60;
 
 	private static final int MINUTES_PER_HOUR = 60;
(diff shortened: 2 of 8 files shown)


denniskigen:
Thanks for working on this, @dkayiwa. One non-blocking release coordination note: I checked `openmrs-module-fhir2`, and [`DurationUnitTranslatorImpl`](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DurationUnitTranslatorImpl.java#L41-L54) still calls `Duration.getCode(...)` and maps only the legacy minutes code, `258701004`. If the Minutes concept resolves to the current SNOMED CT code `1156209001`, FHIR2 will translate the duration unit to `Timing.UnitsOfTime.NULL`.

I do not think this blocks this core PR, since saving drug orders, validation, and auto-expiry are handled here.

denniskigen (dismissed):

denniskigen:
It looks like the [Hours concept](https://app.openconceptlab.org/#/orgs/CIEL/sources/CIEL/concepts/1822/) (CIEL 1822) carries the UCUM SAME-AS code `h57.1` rather than `h`. This is confirmed in the CIEL source itself on OCL (current in v2026-06-27, unchanged since at least 2026-05-30), so it ships with CIEL content rather than being a dev3 artifact. `getKnownCode` handles it exactly as designed: the unknown code is skipped and Hours resolves through SNOMED 258702006, so nothing breaks. It does mean Hours has no working UCUM fallback until the mapping is fixed upstream, which we should report to CIEL.

api/src/main/java/org/openmrs/Duration.java
```diff
@@ -56,6 +112,105 @@ public class Duration {
...
+				return addSeconds(startDate, (int) (duration * SECONDS_PER_DAY / frequency.getFrequencyPerDay()));
+			}
+		};
+
+		abstract Date addToDate(Date startDate, Integer duration, OrderFrequency frequency);
+	}
+
+	private static final Map<String, Unit> UNITS_BY_CODE;
+
+	static {
```
mseaton: I wonder if this would be better if it had both the source and code as the key to the map.  Right now, this works because the UCUM and SNOMED are non-conflicting, which seems like a reasonable assumption to make, but making the key something like `SOURCE:CODE` instead of just the raw code might be better.  This is also more consistently in line with what we do when looking up concepts by reference generally.

api/src/main/java/org/openmrs/Duration.java
```diff
@@ -118,11 +253,68 @@ public static String getCode(Concept durationUnits) {
...
+	 * current SNOMED CT minute code<br/>
+	 * <strong>Should</strong> return a UCUM code when the concept has only a UCUM mapping<br/>
+	 * <strong>Should</strong> return null if no SAME-AS mapping carries a known code<br/>
+	 * <strong>Should</strong> return null if known codes of the concept denote different units
+	 *
+	 * @param durationUnits the duration units concept of a drug order
+	 * @return a known reference term code, or null if the concept cannot be resolved to a single unit
+	 * @since 3.0.0
+	 */
+	public static String getKnownCode(Concept durationUnits) {
```
mseaton: It's a bit unclear to me why we need to expose the underlying codes in the API like this.  From what I gather, the only real purpose of the codes is to then pass them back into the API to get a `Duration` from them.  So really all we care about is going from `Concept` -> `Duration`.  What actual concept reference term (source + code) indicates this is a bit irrelevant.  So instead of creating this `getKnownCode` method that has no real value, let's rename this to `getDuration(Concept)`, that returns the Duration that maps to this concept (through any number of possible supported mappings), and leave those details under the hood.

api/src/main/java/org/openmrs/validator/DrugOrderValidator.java
```diff
@@ -190,7 +190,7 @@ private void validateUnitsAreAmongAllowedConcepts(Errors errors, DrugOrder order
 			if (!drugDurationUnits.contains(order.getDurationUnits())) {
 				errors.rejectValue("durationUnits", "DrugOrder.error.notAmongAllowedConcepts");
 			}
-			if (Duration.getCode(order.getDurationUnits()) == null) {
+			if (Duration.getKnownCode(order.getDurationUnits()) == null) {
```
mseaton: The only real goal here is to validate that the order durationUnits is a concept that maps to a valid Duration.  We don't actually care about the code.  We only care about whether it can be unambiguously mapped to a Duration.  Instead of adding this new `getKnownCode` method which seems to have limited value, can't we just add a `getDuration(Concept)` method?  Wouldn't that be better and more useful?

api/src/main/java/org/openmrs/BaseDosingInstructions.java
```diff
@@ -26,7 +26,7 @@ public Date getAutoExpireDate(DrugOrder drugOrder) {
 		if (drugOrder.getDuration() == null || drugOrder.getDurationUnits() == null) {
 			return null;
 		}
-		String durationCode = Duration.getCode(drugOrder.getDurationUnits());
+		String durationCode = Duration.getKnownCode(drugOrder.getDurationUnits());
```
mseaton: Same comment here - let's just go straight to a Duration from a Concept by renaming and reimplementing `getKnownCode(Concept)` to `getDuration(Concept)`.  Why go from Concept -> String -> Duration here?

api/src/main/java/org/openmrs/Duration.java
```diff
@@ -118,11 +253,68 @@ public static String getCode(Concept durationUnits) {
 		for (ConceptMap conceptMapping : durationUnits.getConceptMappings()) {
 			ConceptReferenceTerm conceptReferenceTerm = conceptMapping.getConceptReferenceTerm();
 			if (ConceptMapType.SAME_AS_MAP_TYPE_UUID.equals(conceptMapping.getConceptMapType().getUuid())
-			        && Duration.SNOMED_CT_CONCEPT_SOURCE_HL7_CODE
-			                .equals(conceptReferenceTerm.getConceptSource().getHl7Code())) {
```
mseaton, line 122: I don't really see the need to change this method around.  Just deprecate it, since per my other comments, I also don't really see any value in this method or the new `getKnownCode` method anyway.  Just return the first code returned that is valid from the unitsByCode map.  If we are concerned about backwards compatibility, then check legacy codes first

mseaton (changes requested):

api/src/main/java/org/openmrs/Duration.java
```diff
@@ -56,6 +112,105 @@ public class Duration {
...
+		unitsByCode.put(UCUM_DAYS_CODE, Unit.DAYS);
+		unitsByCode.put(SNOMED_CT_WEEKS_CODE, Unit.WEEKS);
+		unitsByCode.put(UCUM_WEEKS_CODE, Unit.WEEKS);
+		unitsByCode.put(SNOMED_CT_MONTHS_CODE, Unit.MONTHS);
+		unitsByCode.put(UCUM_MONTHS_CODE, Unit.MONTHS);
+		unitsByCode.put(SNOMED_CT_YEARS_CODE, Unit.YEARS);
+		unitsByCode.put(UCUM_YEARS_CODE, Unit.YEARS);
+		unitsByCode.put(SNOMED_CT_RECURRING_INTERVAL_CODE, Unit.RECURRING_INTERVAL);
+		UNITS_BY_CODE = Collections.unmodifiableMap(unitsByCode);
+	}
```
wikumChamith:
```suggestion
		UNITS_BY_CODE = Map.ofEntries(Map.entry(SNOMED_CT_SECONDS_CODE, Unit.SECONDS), Map.entry(UCUM_SECONDS_CODE, Unit.SECONDS), Map.entry(SNOMED_CT_MINUTES_CODE, Unit.MINUTES), Map.entry(SNOMED_CT_MINUTES_CODE_2021, Unit.MINUTES), Map.entry(UCUM_MINUTES_CODE, Unit.MINUTES), Map.entry(SNOMED_CT_HOURS_CODE, Unit.HOURS), Map.entry(UCUM_HOURS_CODE, Unit.HOURS), Map.entry(SNOMED_CT_DAYS_CODE, Unit.DAYS), Map.entry(UCUM_DAYS_CODE, Unit.DAYS), Map.entry(SNOMED_CT_WEEKS_CODE, Unit.WEEKS), Map.entry(UCUM_WEEKS_CODE, Unit.WEEKS), Map.entry(SNOMED_CT_MONTHS_CODE, Unit.MONTHS), Map.entry(UCUM_MONTHS_CODE, Unit.MONTHS), Map.entry(SNOMED_CT_YEARS_CODE, Unit.YEARS), Map.entry(UCUM_YEARS_CODE, Unit.YEARS), Map.entry(SNOMED_CT_RECURRING_INTERVAL_CODE, Unit.RECURRING_INTERVAL));

	}
```

api/src/main/java/org/openmrs/Duration.java
```diff
@@ -118,11 +253,68 @@ public static String getCode(Concept durationUnits) {
...
+			}
+			Unit unit = UNITS_BY_CODE.get(conceptReferenceTerm.getCode());
+			if (unit == null) {
+				continue;
+			}
+			if (knownUnit == null) {
+				knownCode = conceptReferenceTerm.getCode();
+				knownUnit = unit;
+			} else if (knownUnit != unit) {
+				return null;
```
wikumChamith: I think we should log here before returning null.

api/src/main/java/org/openmrs/Duration.java
dkayiwa: Applied in 096b1ed. `getKnownCode` is gone and the API is now `Duration.getDuration(Integer duration, Concept durationUnits)`, returning null when the concept does not resolve to exactly one known unit, so callers go straight from the order fields to a `Duration` and never see a code. It takes the length as well because a `Duration` carries the value. With the code fully internal, choosing among several equivalent codes no longer needs an ordering preference, since any known code the concept carries denotes the same unit or resolution refuses outright.

api/src/main/java/org/openmrs/Duration.java
dkayiwa: Done in 096b1ed. The registry is now one map per concept source and resolution consults only the map matching the mapping's source, so a code cannot resolve through the wrong vocabulary. The one entry point that still accepts a bare code is `addToDate`, because `new Duration(value, code)` predates this change and receives codes without a source; it checks the two maps in sequence, which stays unambiguous while the vocabularies do not collide, and no resolution path goes through it anymore.

api/src/main/java/org/openmrs/validator/DrugOrderValidator.java
dkayiwa: Done in 096b1ed. The validator now checks `Duration.getDuration(order.getDuration(), order.getDurationUnits()) == null`.

api/src/main/java/org/openmrs/BaseDosingInstructions.java
dkayiwa: Done in 096b1ed. It now goes straight from the order fields to a `Duration` with no intermediate code.

api/src/main/java/org/openmrs/Duration.java
dkayiwa: Restored byte identical to master and deprecated in 096b1ed. I left its behavior alone rather than filtering to known codes, because fhir2's `DurationUnitTranslatorImpl` still calls it expecting the raw SNOMED SAME-AS code, and changing semantics at the moment of deprecation would move behavior under existing consumers. The legacy-first ordering also stopped mattering once code selection moved inside `getDuration`, since equivalent codes all denote the same unit.

api/src/main/java/org/openmrs/Duration.java
dkayiwa: Applied in 096b1ed as two `Map.ofEntries` registries, one per concept source, folding in Mike's suggestion to scope lookups by source. `Map.ofEntries` also fails fast at class load if a code is ever registered twice.

api/src/main/java/org/openmrs/Duration.java
dkayiwa: Done in 096b1ed. It now logs a warning naming the concept and the two conflicting codes before returning null.

dkayiwa:
Good catch @denniskigen. That `h57.1` string was in the dev3 data I collected and I read past it, so the PR description overclaimed that every unit in the set carries its UCUM code. I have corrected the description: Hours resolves through its still-active SNOMED code and simply has no UCUM fallback until the CIEL mapping is fixed. Happy to file the CIEL report for 1822 unless you already have.

denniskigen:
@dkayiwa The PR description still describes `getKnownCode(Concept)`; mind updating it to reflect the reworked `getDuration(Integer duration, Concept durationUnits)` API so the description matches what merges?

dkayiwa:
Thanks @denniskigen. The description was refreshed when the rework landed, but it still named getKnownCode in a parenthetical about the history; I have removed that too, so it now reads purely in terms of the getDuration(Integer duration, Concept durationUnits) API that merges. While closing loops, the fhir2 half you flagged is now open as openmrs/openmrs-module-fhir2#615, teaching DurationUnitTranslatorImpl the current minute code and the UCUM mappings in both directions.

api/src/main/java/org/openmrs/Duration.java
```diff
@@ -125,4 +254,73 @@ public static String getCode(Concept durationUnits) {
...
+		for (ConceptMap conceptMapping : durationUnits.getConceptMappings()) {
+			if (!ConceptMapType.SAME_AS_MAP_TYPE_UUID.equals(conceptMapping.getConceptMapType().getUuid())) {
+				continue;
+			}
+			ConceptReferenceTerm conceptReferenceTerm = conceptMapping.getConceptReferenceTerm();
+			Unit unit = findUnit(conceptReferenceTerm.getConceptSource(), conceptReferenceTerm.getCode());
+			if (unit == null) {
+				continue;
+			}
+			if (knownUnit == null) {
```
mseaton: I really don't understand all of this unit and knownUnit stuff.  I guess things don't work as cleanly as I was anticipating given the Duration constructor takes in a String code and then this is used throughout.  Ideally we wouldn't need to carry this String `code` as a member variable and continuously use it to convert to a `Unit`.

api/src/main/java/org/openmrs/Duration.java
```diff
@@ -125,4 +254,73 @@ public static String getCode(Concept durationUnits) {
...
+	private static Unit findUnit(ConceptSource conceptSource, String code) {
+		if (isSnomedCtSource(conceptSource)) {
+			return SNOMED_CT_UNITS_BY_CODE.get(code);
+		}
+		if (isUcumSource(conceptSource)) {
+			return UCUM_UNITS_BY_CODE.get(code);
+		}
+		return null;
+	}
+
```
mseaton, line 354: Kind of strange that these two methods have different logic.  Why don't we check the name for the snomed_ct in addition to checking the hl7 code, like we do for ucum?

api/src/main/java/org/openmrs/Duration.java
dkayiwa: The `knownUnit` comparison is doing real work — it's what rejects a concept whose SAME-AS mappings *disagree* (e.g. one maps to minutes, another to hours) instead of silently resolving to whichever mapping happens to be iterated first; the legacy + current SNOMED minute codes resolve fine precisely because they map to the same `Unit`.

What's gratuitous is carrying the result as a **String** `code`, and that's only there because the long-standing public `Duration(Integer, String)` constructor takes one. I'll switch the member to the `Unit` enum: `getDuration()` builds the `Duration` from the resolved `Unit` directly (no String round-trip), and the public String constructor stays for backward compatibility — resolving its code to a `Unit` once, and still deferring the unknown-code `APIException` to `addToDate()` exactly as today. That drops `knownCode` and the repeated map lookups in `addToDate()`.

api/src/main/java/org/openmrs/Duration.java
dkayiwa: Fair — the asymmetry is empirical, not principled. In the reference metadata SNOMED CT reliably carries the HL7 code `SCT`, whereas UCUM is frequently registered by source *name* only (no HL7 code), so only UCUM needed the name fallback. But nothing stops a dictionary from registering SNOMED by name too, so I'll make them symmetric — add a SNOMED source-name check (`SNOMED CT` / `SNOMED_CT`) alongside the `SCT` HL7 code, behind a named constant, mirroring `isUcumSource`.

mseaton (approved):
My feedback was addressed, though would be happy for someone else to have a look and review that perspective

wikumChamith:
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=openmrs_openmrs-core&pullRequest=6246) **Quality Gate passed**
> 
> Issues  [2 New issues](https://sonarcloud.io/project/issues?id=openmrs_openmrs-core&pullRequest=6246&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=openmrs_openmrs-core&pullRequest=6246&issueStatuses=ACCEPTED)
> 
> Measures  [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=openmrs_openmrs-core&pullRequest=6246&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0.0% Coverage on New Code](https://sonarcloud.io/component_measures?id=openmrs_openmrs-core&pullRequest=6246&metric=new_coverage&view=list)  [0.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=openmrs_openmrs-core&pullRequest=6246&metric=new_duplicated_lines_density&view=list)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=openmrs_openmrs-core&pullRequest=6246)

@dkayiwa, what do you think about these sonar issues?

api/src/main/java/org/openmrs/Duration.java
dkayiwa: Done in 40dfc30. Duration now stores the resolved Unit as its member; getDuration builds it straight from the resolved unit with no String round trip, the public String constructor resolves its code once, and the raw code is retained only to name the offending value in the deferred unknown-code exception.

api/src/main/java/org/openmrs/Duration.java
dkayiwa: Done in 40dfc30. isSnomedCtSource now matches the source name as well, behind Duration.SNOMED_CT_CONCEPT_SOURCE_NAME, mirroring isUcumSource, and the fhir2 translator got the same treatment on both directions.

api/src/main/java/org/openmrs/Duration.java
dkayiwa: Small correction to the above after 40dfc30: since the resolution now carries units rather than codes, the warning names the concept and the two conflicting units instead of the raw codes. The concept uuid in the message is what identifies the offending mappings.

dkayiwa:
@wikumChamith both are fixed in 546029391 rather than argued with. The bare @Deprecated on getCode now carries since = "3.0.0" (no forRemoval, since the method stays for module compatibility; fhir2 still calls it), and the getDuration loop lost both continue statements by folding the SAME-AS check into a findUnit(ConceptMap) overload, which reads better anyway. The gate had passed either way, but there was no reason to carry the smells into master.

dkayiwa:
@claude review

api/src/main/java/org/openmrs/Duration.java
```diff
@@ -125,4 +283,75 @@ public static String getCode(Concept durationUnits) {
...
+	private static boolean isSnomedCtSource(ConceptSource conceptSource) {
+		return SNOMED_CT_CONCEPT_SOURCE_HL7_CODE.equals(conceptSource.getHl7Code())
+		        || SNOMED_CT_CONCEPT_SOURCE_NAME.equalsIgnoreCase(conceptSource.getName());
+	}
+
+	private static boolean isUcumSource(ConceptSource conceptSource) {
+		return UCUM_CONCEPT_SOURCE_NAME.equalsIgnoreCase(conceptSource.getName())
+		        || UCUM_CONCEPT_SOURCE_NAME.equals(conceptSource.getHl7Code());
+	}
```
ibacher, line 363: Two things:

1. It seems like we should have a different constant for `UCUM_CONCEPT_SOURCE_HL7_CODE` and `UCUM_CONCEPT_SOURCE_NAME` even if these have the same value as they are not the same thing semantically.
2. It's probably better to do the simple `equals()` check first.

api/src/main/java/org/openmrs/validator/DrugOrderValidator.java
```diff
@@ -190,7 +190,7 @@ private void validateUnitsAreAmongAllowedConcepts(Errors errors, DrugOrder order
 			if (!drugDurationUnits.contains(order.getDurationUnits())) {
 				errors.rejectValue("durationUnits", "DrugOrder.error.notAmongAllowedConcepts");
 			}
-			if (Duration.getCode(order.getDurationUnits()) == null) {
+			if (Duration.getDuration(order.getDuration(), order.getDurationUnits()) == null) {
 				errors.rejectValue("durationUnits", "DrugOrder.error.durationUnitsNotMappedToSnomedCtDurationCode");
```
ibacher: It feels like maybe we should be using a different translation string here. It needs to be updated in all languages anyway.

api/src/main/resources/messages.properties
```diff
@@ -1007,15 +1007,15 @@ DrugOrder.error.instructionIsNullForDosingTypeFreeText=Instructions is required
 DrugOrder.error.dosingInstructionsIsNullForDosingTypeFreeText=Dosing instructions is required if dosing type is FreeText
 DrugOrder.error.routeNotAmongAllowedConcepts=The route concept must be among allowed concepts as specified by the order.drugRoutesConceptUuid global property
 DrugOrder.error.drugIsRequired=Drug is required
-DrugOrder.error.durationUnitsNotMappedToSnomedCtDurationCode=Duration units must be mapped to SNOMED CT duration code with the map type set to SAME-AS
+DrugOrder.error.durationUnitsNotMappedToSnomedCtDurationCode=Duration units must be mapped, with the map type set to SAME-AS, to a known SNOMED CT or UCUM duration code
```
ibacher: Nit: "Duration units must be mapped to a known SNOMED CT or UCUM duration code with the map type SAME-AS" in my mind reads cleaner and emphases the most likely cause of the error (missing a SNOMED CT or UCUM mapping) over the SAME-AS part.

ibacher (commented):
A couple of minor points. My main concern here is that this PR fixes the issue but in a way that we can't really cleanly backport to the 2.x line (do we backport the UCUM stuff? What do we do with the various `Map.ofEntries(...)` or `Map.entry(...)` stuff that doesn't exist on JDK 8?) where we also need this fixed.

api/src/main/java/org/openmrs/Duration.java
```diff
@@ -22,16 +25,27 @@
...
  *
  * @since 1.10
  */
 public class Duration {
 
+	private static final Logger log = LoggerFactory.getLogger(Duration.class);
+
 	public static final String SNOMED_CT_SECONDS_CODE = "257997001";
 
 	public static final String SNOMED_CT_MINUTES_CODE = "258701004";
```
ibacher, line 51: Small correctness thing: should we be updating `SNOMED_CT_MINUTES_CODE` instead of adding the new value as a new constant? `258701004` is marked as `inactive` in SNOMED CT itself, so arguably that would be the wrong code to use places. Plus, this way we force every downstream caller to update to take advantage of the updated code.

api/src/main/java/org/openmrs/Duration.java
dkayiwa: Done in 197172b. UCUM now has its own UCUM_CONCEPT_SOURCE_HL7_CODE constant, distinct from the name constant even though the values coincide, and the exact hl7Code equality runs before the case-insensitive name comparison, matching isSnomedCtSource's order.

api/src/main/java/org/openmrs/validator/DrugOrderValidator.java
dkayiwa: Done in 197172b. Both validators now use a new key, DrugOrder.error.durationUnitsNotMappedToKnownDurationCode, so every locale falls back to the accurate English text instead of a stale SNOMED-only translation. The old key keeps its original text and translations for anything external that still references it.

api/src/main/resources/messages.properties
dkayiwa: Took this wording verbatim for the new key in 197172b.

api/src/main/java/org/openmrs/Duration.java
dkayiwa: I kept the value where it is and documented the constant as legacy instead, in 197172b. Repurposing the value of a public constant is riskier than it looks because String constants are inlined at compile time: every module compiled against an older core keeps 258701004 baked into its bytecode until rebuilt, and then silently flips behavior on rebuild. fhir2 is a concrete example, since its translator builds a code map with Duration.SNOMED_CT_MINUTES_CODE as the legacy entry, and a value swap would silently turn that entry into a duplicate of the new code and drop legacy coverage on the next compile. Core has to keep recognising both codes either way, because existing dictionaries carry the legacy mapping, so a swap would not change the registry contents, only the meaning of the name underneath binaries that cannot see it. The Javadoc on the legacy constant now says it is inactive in SNOMED CT and points new mappings at SNOMED_CT_MINUTES_CODE_2021.

dkayiwa:
@ibacher on the backport concern: the plan Mike and I converged on is to backport only after this merges, and not to backport this shape. The 2.x lines would get the minimal fix in each branch's own idiom: accept 1156209001 alongside the legacy code, static-block map instead of Map.ofEntries, no UCUM, no new API, JDK 8 clean. That patch effectively already exists; it was written and green against 2.8.x and 2.9.x as #6244 and #6245 before we parked backports pending this PR, and dev3 runs 2.8.7, so the 2.8.x one is what actually fixes the reported environment. If we later want UCUM resolution on 2.x it is expressible with JDK 8 constructs, but my inclination is to keep maintenance lines to the minimal accept-the-new-code change.

api/src/main/java/org/openmrs/Duration.java
ibacher: This actually seems like a strong argument that we should expose these constants via a function or some form of indirection that would stop compile-time inlining, but I suppose it's too late to make that decision.

ibacher:
Thanks @dkayiwa. I think that addresses my concerns here.

dkayiwa:
Heads up on a semantics change in fac549035, following @ibacher's review of the fhir2 companion (openmrs/openmrs-module-fhir2#615, https://github.com/openmrs/openmrs-module-fhir2/pull/615#discussion_r3544573457): refusing to resolve when a concept's SNOMED CT and UCUM mappings denote different units imposed a consistency rule that nothing required while UCUM was ignored, so previously valid orders could start failing validation. getDuration now gives SNOMED CT priority instead, stopping at the first known code, and consults UCUM only when no SNOMED CT mapping resolves. Every concept that resolves today keeps resolving identically however its UCUM mappings are curated (CIEL's malformed `h57.1` on Hours being a live example of why that matters), and UCUM strictly extends coverage to concepts that resolved to nothing. The conflict warning went away with the conflict branch, and DurationTest now pins the preference plus the UCUM fallback for a concept whose only SNOMED CT code is unknown.

@mseaton flagging you since the disagreement-rejection behavior was part of what you reviewed; the API shape is unchanged.

api/src/main/java/org/openmrs/Duration.java
dkayiwa: Closing the loop here once more: the warning went away together with the conflict branch in fac54903. Per Ian's review on the fhir2 companion PR, SNOMED CT mappings now take priority and UCUM only fills gaps, so the disagreeing-mappings case this thread was about can no longer influence resolution and there is nothing left to log.

api/src/main/java/org/openmrs/Duration.java
dkayiwa: Agreed, and it is a lesson worth carrying into future API additions: accessors would have kept the values swappable. For these particular constants the values are already inlined into downstream binaries either way, so they have to stay stable regardless, and the class Javadoc now records that invariant.

PR #6251, opened 2026-07-04
TRUNK-6634: Run mysql-only changesets on MariaDB

VivekVardhanArrabelli:
MariaDB 11 dropped the "5.5.5-" version-string prefix, so Liquibase now detects it as MariaDB rather than MySQL and silently skips changesets declared with dbms="mysql". This left schema and core data missing on MariaDB 11 deployments: cohort_member lost its primary-key/uuid migration in the 2.1.x update log, scheduler_task_config uuids were never generated before their not-null constraint, the 'System Developer' role was never re-created, and 248 further 2.0.x changesets plus the 1.9.x snapshot FOREIGN_KEY_CHECKS toggles never ran.

Checksum-safety was verified empirically against Liquibase 4.32.0 (v8 and v9 checksum versions): editing the changeSet-level dbms attribute or a dbms precondition does NOT change a changeset's checksum, but adding a modifySql block DOES. The fix therefore:

- extends dbms="mysql" to dbms="mysql,mariadb" in place for 248 changesets in liquibase-update-to-latest-2.0.x.xml, 3 in liquibase-update-to-latest-2.1.x.xml and the 2 FOREIGN_KEY_CHECKS toggles in liquibase-core-data-1.9.x.xml. All raw-SQL changesets were audited individually for MariaDB compatibility (multi-table UPDATE/DELETE joins, backslash escapes, double-quoted literals, boolean literals - all identical semantics on MariaDB), and the three customChange classes involved contain only database-agnostic JDBC.

- leaves 201610042145-2.1 (cohort_member uuid backfill), 20100128-1 (System Developer role) and 20090831-1040-scheduler_task_config (scheduler uuid backfill) untouched, since extending them would require new modifySql blocks and change their checksums; dedicated mariadb-only changesets (TRUNK-6634-2026-07-03-1000/-1001/-1002) perform the equivalent work with valueComputed="UUID()", each ordered before its dependent constraint. These also repair databases already stuck at a failed constraint: the backfill fills the missing uuids on the next update run.

- does NOT touch the 57 uuid backfills gated by mysql/oracle dbms preconditions: on MariaDB those are covered by the existing java fallback in changeset 20090402-1517 (GenerateUuid), which already special-cases mariadb.

Tests:

- ChangeSetChecksumBaselineTest pins the v8 checksums of all 718 changesets in the three touched changelogs to a baseline generated from the changelogs before this change, i.e. the values stored in the DATABASECHANGELOG of existing production databases. Any future edit that would break "liquibase validate" for existing deployments now fails CI. It also enforces that mysql changesets include mariadb unless documented as mysql-only by design, and asserts repair changeset ordering.

- MariaDbUpdateChangeSetsTest covers the original four 2.1.x changesets from the ticket in detail.

## Issue I worked on

  No? -> execute above command


api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.0.x.xml
```diff
@@ -2479,6 +2496,23 @@
...
 		<modifySql dbms="oracle"><replace replace="name-of-uuid-function" with="SYS_GUID()"/></modifySql>
 	</changeSet>
+	<changeSet id="TRUNK-6634-2026-07-03-1001" author="arrabelli" dbms="mariadb">
+		<preConditions onFail="MARK_RAN">
+			<columnExists tableName="scheduler_task_config" columnName="uuid"/>
+		</preConditions>
+		<comment>
+			Generating UUIDs for rows in scheduler_task_config on MariaDB. Changeset
+			20090831-1040-scheduler_task_config only runs for MySQL and Oracle, and the table is not
+			covered by the java fallback in 20090402-1517, so on MariaDB the uuid column stayed null
```
dkayiwa: Is this changeset actually needed? Unlike cohort_member, scheduler_task_config already has a Java UUID fallback. 20090831-1041-scheduler_task_config calls GenerateUuid for this table with a not(mysql/oracle/mssql) precondition, so on MariaDB 11 it runs (mariadb is no longer detected as "mysql", which is the premise of this whole PR) and backfills via `update scheduler_task_config set uuid = uuid() where uuid is null`, because GenerateUuid special-cases mariadb. That fallback sits between this changeset and the not-null constraint 20090831-1042, so the uuids are filled before the constraint regardless of this changeset.

So the rationale in the comment doesn't look right for this table: 20090402-1517 indeed doesn't list scheduler_task_config, but 20090831-1041 does, so the column doesn't stay null on MariaDB and 20090831-1042 wouldn't fail. cohort_member is genuinely different because it appears in no GenerateUuid list anywhere, which is why its dedicated backfill (1000) is needed; scheduler_task_config is covered, so this one looks removable, or at least the comment needs correcting.

If you did observe the scheduler constraint actually failing on MariaDB 11, I'd like to understand how 1041 was bypassed. Not blocking either way.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.0.x.xml
```diff
@@ -1700,23 +1700,23 @@
...
+	<changeSet id="20090414-0805" author="bwolfe" dbms="mysql,mariadb">
 		<preConditions onFail="MARK_RAN">
 			<not><columnExists tableName="concept_set" columnName="concept_set_id"/></not>
 		</preConditions>
 		<comment>Dropping primary key on concept set table</comment>
 		<dropPrimaryKey tableName="concept_set" />
 	</changeSet>
 	
-	<changeSet id="20090414-0806" author="bwolfe" dbms="mysql">
+	<changeSet id="20090414-0806" author="bwolfe" dbms="mysql,mariadb">
```
dkayiwa, line 1719: One side effect of extending these three changesets to mariadb: the `<modifySql dbms="mysql"><append value=" FIRST"/></modifySql>` block only applies on mysql, so on MariaDB the new primary-key column is added at the end of the table instead of first. Same pattern in 20090414-0809 (patient_identifier) and 20090414-0812 (concept_word). MariaDB does support ADD COLUMN ... FIRST, but extending the modifySql to cover it would change the checksum, so the column-ordering difference between mysql and mariadb installs can't easily be avoided here.

It's cosmetic (the auto-increment primary key and unique constraint are still created correctly, and OpenMRS addresses columns by name), so my suggestion is to just accept it rather than add more changesets. Flagging it only because the description mentions auditing the raw SQL for MariaDB but not this modifySql case. Not blocking.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.0.x.xml
VivekVardhanArrabelli: yep, 20090831-1041 already covers MariaDB via GenerateUuid
since its precondition excludes only mysql/oracle/mssql. Removed the redundant
changeset in ff0b880. The ordering test now pins the existing 1041 fallback
between the backfill and the not-null constraint instead. (The 'System Developer'
changeset is the one dedicated changeset still needed — 20100128-2, now extended
to mariadb, reassigns users to that role, so it must exist first; also asserted
in the test now.)

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.0.x.xml
VivekVardhanArrabelli: Agreed — that's the trade-off this PR takes: extending the modifySql blocks would
change the checksums, and the FIRST placement is cosmetic (the PK/auto-increment
constraints are in the addColumn itself), so column position simply differs on
MariaDB. Happy to add a comment in the changelog noting it if you'd like.

PR #6257, opened 2026-07-07
TRUNK-6675: Module is not updated when omod changes

rkorytkowski:
## Summary

- `File.delete()` only removes empty directories, so when a SNAPSHOT omod was rebuilt between restarts the lib cache directory (e.g. `legacyui/`) could not be deleted and its stale expanded resources persisted.
- `OpenmrsClassLoader.expandURL` has a `file.exists()` short-circuit that returns the cached (stale) resource without re-expanding, so any resource loaded through the lazy-expansion path would remain stale indefinitely after the first failed delete.
- The new timestamp was still written to `.moduleLastModified` after the failed delete, causing every subsequent restart to see matching timestamps and skip the deletion entirely.

Both `tmpModuleDir.delete()` calls in `getLibCacheFolderForModule` are replaced with `FileUtils.deleteDirectory()`, which recursively removes the directory tree. Two regression tests are added to `ModuleClassLoaderTest` covering the module-changed and optimized-startup-disabled paths.

Fixes https://issues.openmrs.org/browse/TRUNK-6675
Related to https://issues.openmrs.org/browse/TRUNK-6509

## Test plan

- [x] `ModuleClassLoaderTest#getLibCacheFolderForModule_shouldRecreateNonEmptyDirWhenModuleChanged`
- [x] `ModuleClassLoaderTest#getLibCacheFolderForModule_shouldRecreateNonEmptyDirWhenOptimizedStartupDisabled`
- [x] Manual: rebuild a SNAPSHOT omod, restart the server, and verify the updated code is picked up

### DIFF
--- a/api/src/main/java/org/openmrs/module/ModuleClassLoader.java
+++ b/api/src/main/java/org/openmrs/module/ModuleClassLoader.java
@@ -472,6 +472,7 @@ public class ModuleClassLoader extends URLClassLoader {
 			long moduleLastModified = module.getFile().lastModified();
 			File moduleLastModifiedFile = new File(tmpModuleDir, ".moduleLastModified");
 			
+			boolean deleted = true;
 			if (Context.isOptimizedStartup() && moduleLastModifiedFile.exists()) {
 				// Re-create tmpModuleDir if module changed
 				try {
@@ -479,11 +480,7 @@ public class ModuleClassLoader extends URLClassLoader {
 						Charset.defaultCharset());
 					if (!Long.valueOf(savedLastModified).equals(moduleLastModified)) {
 						log.debug("Deleting {} since the module was modified", tmpModuleDir.getAbsolutePath());
-						try {
-							FileUtils.deleteDirectory(tmpModuleDir);
-						} catch (IOException e) {
-							log.warn("Failed to delete lib cache dir for module {}", module.getModuleId(), e);
-						}
+						deleted = deleteLibCacheDir(tmpModuleDir, module.getModuleId());
 					}
 				} catch (IOException | NumberFormatException e) {
 					log.warn("Error while reading module last modified file: {}", moduleLastModifiedFile, e);
@@ -491,17 +488,13 @@ public class ModuleClassLoader extends URLClassLoader {
 			} else {
 				log.debug("Optimized startup disabled or {} does not exist, deleting {}", moduleLastModifiedFile,
 					tmpModuleDir);
-				try {
-					FileUtils.deleteDirectory(tmpModuleDir);
-				} catch (IOException e) {
-					log.warn("Failed to delete lib cache dir for module {}", module.getModuleId(), e);
-				}
+				deleted = deleteLibCacheDir(tmpModuleDir, module.getModuleId());
 			}
 
 			tmpModuleDir.mkdirs();
 			libCacheFolders.put(module.getModuleId(), tmpModuleDir);
 
-			if (moduleLastModified != 0L) {
+			if (deleted && moduleLastModified != 0L) {
 				try {
 					FileUtils.writeStringToFile(moduleLastModifiedFile, moduleLastModified + "", Charset.defaultCharset());
 				} catch (IOException e) {
@@ -513,6 +506,16 @@ public class ModuleClassLoader extends URLClassLoader {
 		}
 	}
 	
+	private static boolean deleteLibCacheDir(File dir, String moduleId) {
+		try {
+			FileUtils.deleteDirectory(dir);
+			return true;
+		} catch (IOException e) {
+			log.warn("Failed to delete lib cache dir {} for module {}", dir, moduleId, e);
+			return false;
+		}
+	}
+
 	/**
 	 * Get all urls for the given <code>module</code> that are not already in the
 	 * <code>existingUrls</code>
--- a/api/src/main/java/org/openmrs/module/ModuleClassLoader.java
+++ b/api/src/main/java/org/openmrs/module/ModuleClassLoader.java
@@ -484,6 +484,7 @@ public class ModuleClassLoader extends URLClassLoader {
 					}
 				} catch (IOException | NumberFormatException e) {
 					log.warn("Error while reading module last modified file: {}", moduleLastModifiedFile, e);
+					deleted = deleteLibCacheDir(tmpModuleDir, module.getModuleId());
 				}
 			} else {
 				log.debug("Optimized startup disabled or {} does not exist, deleting {}", moduleLastModifiedFile,
--- a/api/src/main/java/org/openmrs/module/ModuleClassLoader.java
+++ b/api/src/main/java/org/openmrs/module/ModuleClassLoader.java
@@ -479,15 +479,19 @@ public class ModuleClassLoader extends URLClassLoader {
 						Charset.defaultCharset());
 					if (!Long.valueOf(savedLastModified).equals(moduleLastModified)) {
 						log.debug("Deleting {} since the module was modified", tmpModuleDir.getAbsolutePath());
-						tmpModuleDir.delete();
+						FileUtils.deleteDirectory(tmpModuleDir);
 					}
 				} catch (IOException | NumberFormatException e) {
 					log.warn("Error while reading module last modified file: {}", moduleLastModifiedFile, e);
 				}
 			} else {
-				log.debug("Optimized startup disabled or {} does not exist, deleting {}", moduleLastModifiedFile, 
+				log.debug("Optimized startup disabled or {} does not exist, deleting {}", moduleLastModifiedFile,
 					tmpModuleDir);
-				tmpModuleDir.delete();
+				try {
+					FileUtils.deleteDirectory(tmpModuleDir);
+				} catch (IOException e) {
+					log.warn("Failed to delete lib cache dir for module {}", module.getModuleId(), e);
+				}
 			}
 
 			tmpModuleDir.mkdirs();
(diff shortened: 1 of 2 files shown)


dkayiwa:
@claude review

api/src/test/java/org/openmrs/module/ModuleClassLoaderTest.java
```diff
@@ -475,6 +482,86 @@ void testGlobPatternDoesNotMatch(String filePath) throws MalformedURLException {
...
+			module.setFile(omodFile);
+
+			File moduleDir = new File(tempLibCache, "testmodule");
+			moduleDir.mkdirs();
+			FileUtils.writeStringToFile(new File(moduleDir, ".moduleLastModified"),
+				String.valueOf(currentTimestamp - 1000), Charset.defaultCharset());
+			File staleFile = new File(moduleDir, "old-resource.txt");
+			FileUtils.writeStringToFile(staleFile, "stale content", Charset.defaultCharset());
+
+			ModuleClassLoader.getLibCacheFolderForModule(module);
```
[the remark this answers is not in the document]
rkorytkowski: Addressed.

api/src/main/java/org/openmrs/module/ModuleClassLoader.java
```diff
@@ -479,15 +479,23 @@ public static File getLibCacheFolderForModule(Module module) {
 						Charset.defaultCharset());
 					if (!Long.valueOf(savedLastModified).equals(moduleLastModified)) {
 						log.debug("Deleting {} since the module was modified", tmpModuleDir.getAbsolutePath());
-						tmpModuleDir.delete();
+						try {
+							FileUtils.deleteDirectory(tmpModuleDir);
+						} catch (IOException e) {
+							log.warn("Failed to delete lib cache dir for module {}", module.getModuleId(), e);
+						}
```
dkayiwa: I'd fix this before merging: when this delete fails, the code below still writes the fresh timestamp into `.moduleLastModified`, which is the masking called out in the summary. The next startup sees matching timestamps, never retries the cleanup, and whatever couldn't be deleted keeps being served through the `expandURL` `exists()` short-circuit until the omod's timestamp changes again (for a released module, effectively never). I reproduced it on this branch with a file made undeletable: after the failed delete the marker carried the new timestamp, and a second call skipped deletion and left the stale file in place. The window is narrow but real. `dispose()` evicts the `libCacheFolders` entry, so a hot module upgrade re-enters this code in the same JVM while the old classloader can still hold jars open on Windows, and commons-io only self-heals read-only failures (`forceDelete` passes `OVERRIDE_READ_ONLY`), not held locks.

I'd pull the delete into a small helper that reports success, and skip the marker write when it failed:

```java
private static boolean deleteLibCacheDir(File dir, String moduleId) {
	try {
		FileUtils.deleteDirectory(dir);
		return true;
	} catch (IOException e) {
		log.warn("Failed to delete lib cache dir {} for module {}", dir, moduleId, e);
		return false;
	}
}
```

with both branches assigning to a local and the timestamp write at the bottom guarded by it. If the old marker survives a partial delete, timestamps still differ next boot and it retries; if the marker got deleted, the missing-marker branch retries. Either way it recovers instead of masking. Extracting the helper also removes the nested try and drops the method's cognitive complexity, which are the two things SonarCloud is currently failing the quality gate on, so it should turn that green as well.

api/src/test/java/org/openmrs/module/ModuleClassLoaderTest.java
```diff
@@ -475,6 +482,88 @@ void testGlobPatternDoesNotMatch(String filePath) throws MalformedURLException {
...
+			Module module = new Module("testmodule2", "testmodule2", "org.openmrs.module.testmodule2",
+				"author", "description", "1.0", "1.0");
+			module.setFile(omodFile);
+
+			File moduleDir = new File(tempLibCache, "testmodule2");
+			moduleDir.mkdirs();
+			File staleFile = new File(moduleDir, "old-resource.txt");
+			FileUtils.writeStringToFile(staleFile, "stale content", Charset.defaultCharset());
+
+			Context.getRuntimeProperties().setProperty("optimized.startup", "false");
```
dkayiwa: This line sets the property on a throwaway copy. `Context.getRuntimeProperties()` returns a defensive copy (its Javadoc says "copy of the runtime properties"), so `optimized.startup` never actually becomes false here and `isOptimizedStartup()` keeps returning true. The test then reaches the deleting branch only because `.moduleLastModified` doesn't exist, not because optimized startup is disabled. I checked on this branch: keep this line but add a marker file matching the omod timestamp, and the stale file survives; switch to `Context.setRuntimeProperties` and it gets deleted. Not blocking, since I also ran both new tests against the old `File.delete()` code and they fail there, so the regression is genuinely covered either way. But to make the test cover the branch it's named for, mutate the copy and push it back the way `UserServiceTest` and the logging tests do:

```java
Properties props = Context.getRuntimeProperties();
props.setProperty("optimized.startup", "false");
Context.setRuntimeProperties(props);
```

restoring the saved `Properties` in the `finally` (the `remove()` there is a no-op for the same reason).

dkayiwa (commented):
One thing with no line to hang it on: master carries the identical `tmpModuleDir.delete()` code and TRUNK-6675 lists Core 3.0.0 as a fix version, but 6255 was closed. Is the plan to open a fresh master PR once this one settles? Otherwise 3.0 keeps the staleness bug this fixes for 2.9.

api/src/test/java/org/openmrs/module/ModuleClassLoaderTest.java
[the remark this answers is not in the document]
dkayiwa: The added setProperty doesn't pin the branch, unfortunately. `Context.getRuntimeProperties()` hands back a defensive copy, so the property never reaches `isOptimizedStartup()` and the test still passes only because the default is true. I verified on this branch: setting `optimized.startup=false` through this same call leaves `isOptimizedStartup()` returning true. To actually pin it, mutate the copy and push it back:

```java
Properties props = Context.getRuntimeProperties();
props.setProperty("optimized.startup", "true");
Context.setRuntimeProperties(props);
```

saving the original before the try and restoring it in the finally (the `remove()` is a no-op on another copy). Same applies to the disabled test, where I left a separate note.

rkorytkowski:
> One thing with no line to hang it on: master carries the identical `tmpModuleDir.delete()` code and [TRUNK-6675](https://issues.openmrs.org/browse/TRUNK-6675) lists Core 3.0.0 as a fix version, but 6255 was closed. Is the plan to open a fresh master PR once this one settles? Otherwise 3.0 keeps the staleness bug this fixes for 2.9.

I'll port to master once merged.

[TRUNK-6675]: https://openmrs.atlassian.net/browse/TRUNK-6675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

rkorytkowski:
Addressed review comments.

api/src/main/java/org/openmrs/module/ModuleClassLoader.java
```diff
@@ -472,28 +472,29 @@ public static File getLibCacheFolderForModule(Module module) {
...
 				try {
 					String savedLastModified = FileUtils.readFileToString(moduleLastModifiedFile,
 						Charset.defaultCharset());
 					if (!Long.valueOf(savedLastModified).equals(moduleLastModified)) {
 						log.debug("Deleting {} since the module was modified", tmpModuleDir.getAbsolutePath());
-						tmpModuleDir.delete();
+						deleted = deleteLibCacheDir(tmpModuleDir, module.getModuleId());
 					}
 				} catch (IOException | NumberFormatException e) {
 					log.warn("Error while reading module last modified file: {}", moduleLastModifiedFile, e);
```
dkayiwa, line 486: One more hole you could close while in this method, not blocking: when this read fails, `deleted` stays true, so the write at the bottom still stamps the new timestamp onto a dir that couldn't be verified. An empty marker is all it takes to land here (a crash during the marker rewrite leaves one behind, the rewrite happens on every optimized boot, and `Long.valueOf("")` then throws into this catch). I reproduced it on this branch: empty marker plus a changed omod keeps the stale dir and overwrites the marker with the fresh timestamp, so every later boot sees matching values and never rebuilds. Same terminal state as the failed-delete case this PR fixes. Treating unreadable like missing closes it with the helper you already have:

```suggestion
					log.warn("Error while reading module last modified file: {}", moduleLastModifiedFile, e);
					deleted = deleteLibCacheDir(tmpModuleDir, module.getModuleId());
```

api/src/test/java/org/openmrs/module/ModuleClassLoaderTest.java
dkayiwa: One gap remains here: the test still passes even if `optimized.startup` never reaches the context, since with no `.moduleLastModified` in the dir the else branch runs whatever the flag says. I checked on this branch, dropping the `setRuntimeProperties` push-back doesn't fail it. Seeding a marker that matches the omod timestamp ties the deletion to the flag itself: if the property plumbing ever regresses to true, the dir is kept and this test fails instead of silently passing.

```java
FileUtils.writeStringToFile(new File(moduleDir, ".moduleLastModified"),
	String.valueOf(omodFile.lastModified()), Charset.defaultCharset());
```

before the property flip. Not blocking either way.

PR #6262, opened 2026-07-07
TRUNK-6665: Detect Legacy Passwords and Map to Modern Encoder

ELVIS-KATO:

Added a LegacyPasswordEncoder to encapsulate validation of legacy password hashes.
Added legacy hash detection logic to identify hashes generated using historical OpenMRS password hashing algorithms.
Introduced centralized password matching through Security.matchesPassword() while preserving backward compatibility with existing authentication flows.
Updated authentication logic to delegate validation of legacy hashes to the legacy encoder when a legacy hash is detected.
Added support for flagging accounts authenticated with legacy password hashes using the legacyPassword user property.
Ensured that legacy password hashes are not modified during authentication and continue to work without service disruption.
Updated password change handling to clear the legacy password flag after a successful password update.
Updated all relevant password verification paths to use the centralized password matching mechanism.
Added unit tests covering legacy password detection and validation behavior.
Added integration tests covering legacy authentication, legacy account flagging, and clearing the legacy flag after password changes.

## Issue I worked on
[TRUNK-6665](https://openmrs.atlassian.net/browse/TRUNK-6665)

##Ref:
@dkayiwa  

  No? -> execute above command

[TRUNK-6665]: https://openmrs.atlassian.net/browse/TRUNK-6665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
@@ -165,9 +165,14 @@ public class HibernateContextDAO implements ContextDAO {
 
 			String passwordOnRecord = (String) passwordAndSalt[0];
 			String saltOnRecord = (String) passwordAndSalt[1];
+			String passwordToHash = password + saltOnRecord;
 
 			// if the username and password match, hydrate the user and return it
-			if (passwordOnRecord != null && Security.hashMatches(passwordOnRecord, password + saltOnRecord)) {
+			if (passwordOnRecord != null && Security.matchesPassword(passwordOnRecord, passwordToHash)) {
+				if (Security.isLegacyPasswordHash(passwordOnRecord)) {
+					candidateUser.setUserProperty(OpenmrsConstants.USER_PROPERTY_LEGACY_PASSWORD,
+					    String.valueOf(System.currentTimeMillis()));
+				}
 				// hydrate the user object
 				candidateUser.getAllRoles().size();
 				candidateUser.getUserProperties().size();
@@ -349,7 +354,7 @@ public class HibernateContextDAO implements ContextDAO {
 	public void closeSession() {
 		log.debug("HibernateContext: closing Hibernate Session");
 		if (!participate) {
-			log.debug("Unbinding session from synchronization manager (" + sessionFactory.hashCode() + ")");
+			log.debug("Unbinding session from synchronization manager ({})", sessionFactory.hashCode());
 			
 			if (TransactionSynchronizationManager.hasResource(sessionFactory)) {
 				Object value = TransactionSynchronizationManager.unbindResource(sessionFactory);
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
@@ -360,7 +360,7 @@ public class HibernateContextDAO implements ContextDAO {
 	public void closeSession() {
 		log.debug("HibernateContext: closing Hibernate Session");
 		if (!participate) {
-			log.debug("Unbinding session from synchronization manager ({})", sessionFactory.hashCode());
+			log.debug("Unbinding session from synchronization manager (" + sessionFactory.hashCode() + ")");
 			
 			if (TransactionSynchronizationManager.hasResource(sessionFactory)) {
 				Object value = TransactionSynchronizationManager.unbindResource(sessionFactory);
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
@@ -171,9 +171,14 @@ public class HibernateContextDAO implements ContextDAO {
 
 			String passwordOnRecord = (String) passwordAndSalt[0];
 			String saltOnRecord = (String) passwordAndSalt[1];
+			String passwordToHash = password + saltOnRecord;
 
 			// if the username and password match, hydrate the user and return it
-			if (passwordOnRecord != null && Security.checkPassword(passwordOnRecord, password + saltOnRecord)) {
+			if (passwordOnRecord != null && Security.matchesPassword(passwordOnRecord, passwordToHash)) {
+				if (Security.isLegacyPasswordHash(passwordOnRecord)) {
+					candidateUser.setUserProperty(OpenmrsConstants.USER_PROPERTY_LEGACY_PASSWORD,
+					    String.valueOf(System.currentTimeMillis()));
+				}
 				// hydrate the user object
 				candidateUser.getAllRoles().size();
 				candidateUser.getUserProperties().size();
@@ -355,7 +360,7 @@ public class HibernateContextDAO implements ContextDAO {
 	public void closeSession() {
 		log.debug("HibernateContext: closing Hibernate Session");
 		if (!participate) {
-			log.debug("Unbinding session from synchronization manager (" + sessionFactory.hashCode() + ")");
+			log.debug("Unbinding session from synchronization manager ({})", sessionFactory.hashCode());
 			
 			if (TransactionSynchronizationManager.hasResource(sessionFactory)) {
 				Object value = TransactionSynchronizationManager.unbindResource(sessionFactory);
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
@@ -192,8 +192,9 @@ public class HibernateUserDAO implements UserDAO {
 			//Hardcoding in Luhn since past user IDs used this validator.
 			usernameWithCheckDigit = new LuhnIdentifierValidator().getValidIdentifier(username);
 		}
-		catch (Exception e) {}
-		
+		catch (Exception e) {
+			// Ignore exception, use original username if check digit generation fails
+		}		
 		Query query = sessionFactory
 		        .getCurrentSession()
 		        .createQuery(
@@ -403,6 +404,7 @@ public class HibernateUserDAO implements UserDAO {
 		// reset lockout 
 		changeForUser.setUserProperty(OpenmrsConstants.USER_PROPERTY_LOCKOUT_TIMESTAMP, "");
 		changeForUser.setUserProperty(OpenmrsConstants.USER_PROPERTY_LOGIN_ATTEMPTS, OpenmrsConstants.ZERO_LOGIN_ATTEMPTS_VALUE);
+		changeForUser.removeUserProperty(OpenmrsConstants.USER_PROPERTY_LEGACY_PASSWORD);
 		saveUser(changeForUser, null);
 	}
 	
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
@@ -192,9 +192,8 @@ public class HibernateUserDAO implements UserDAO {
 			//Hardcoding in Luhn since past user IDs used this validator.
 			usernameWithCheckDigit = new LuhnIdentifierValidator().getValidIdentifier(username);
 		}
-		catch (Exception e) {
-			// Ignore exception, use original username if check digit generation fails
-		}		
+		catch (Exception e) {}
+		
 		Query query = sessionFactory
 		        .getCurrentSession()
 		        .createQuery(
--- a/api/src/main/java/org/openmrs/util/LegacyPasswordEncoder.java
+++ b/api/src/main/java/org/openmrs/util/LegacyPasswordEncoder.java
@@ -1,0 +1,84 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.util;
+
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+import org.openmrs.api.APIException;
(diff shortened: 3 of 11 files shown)


jwnasambu (commented):
@ELVIS-KATO Thanks for this changes. Kindly this change should be made on the 2.8.x branch rather than master. Could you please change the base branch of this PR to 2.8.x?

jwnasambu:
@ELVIS-KATO Kindly follow these steps to change the base branch on gitHub

- Open your pull request on GitHub.
- Near the top of the PR, click Edit.
- You will see something like base: master and compare: your-branch.
- Change the base dropdown from master to 2.8.x.
- Review any changes GitHub shows, then click Save or Update pull request.

ELVIS-KATO:
> @ELVIS-KATO Kindly follow these steps to change the base branch on gitHub
> 
> * Open your pull request on GitHub.
> * Near the top of the PR, click Edit.
> * You will see something like base: master and compare: your-branch.
> * Change the base dropdown from master to 2.8.x.
> * Review any changes GitHub shows, then click Save or Update pull request.

I have worked upon it

jwnasambu:
@claude review

jwnasambu (commented):
The PR is currently failing the SonarCloud Quality Gate. Could you please address the SonarCloud issues so the checks pass?

jwnasambu:
@ELVIS-KATO Kindly since the workflow isn't triggered by a force push, you can create an empty commit to trigger CI:

```
git commit --allow-empty -m "Trigger SonarQube analysis"
git push
```

web/src/main/java/org/openmrs/web/filter/update/UpdateFilter.java
```diff
@@ -201,10 +201,11 @@ protected synchronized void doPost(HttpServletRequest httpRequest, HttpServletRe
 					log.debug("Sleeping for 3 seconds because of a bad username/password");
 					Thread.sleep(3000);
 				}
-				catch (InterruptedException e) {
-					log.error("Unable to sleep", e);
-					throw new ServletException("Got interrupted while trying to sleep thread", e);
-				}
+			catch (InterruptedException e) {
```
jwnasambu: Kindly the catch block is indented at 3 tabs when it should be at 5 tabs to match its corresponding try at line 200. The catch body is also under-indented. This makes the control flow structure misleading at a glance. 
Well, the Java compiler ignores indentation, so this doesn't change runtime behavior. But in security-sensitive authentication code, readability is correctness  a reviewer or future contributor scanning this block could misread which scope the catch belongs to.

jwnasambu (commented):
SonarCloud flagged the manual hex encoding in LegacyPasswordEncoder.hexString(). Could you replace the loop with String.format("%02X", byte) or, since the codebase uses lowercase hex elsewhere, String.format("%02x", byte)? The same applies to incorrectHexString  though that one uses Integer.toHexString which doesn't zero-pad, so you'd need %02x there too.

web/src/main/java/org/openmrs/web/filter/update/UpdateFilter.java
ELVIS-KATO: this has been worked on

jwnasambu:
@claude review

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -67,8 +67,17 @@ public final class OpenmrsConstants {
...
+	static {
+		String version = THIS_PACKAGE.getSpecificationVendor();
+		if (version == null) {
+			version = getBuildVersion();
+			if (version == null) {
+				version = getVersion();
+			}
+		}
+		OPENMRS_VERSION = version;
+	}
```
jwnasambu, line 80: Kindly OPENMRS_VERSION refactoring from nested ternary to if-else. Clean improvement but unrelated to legacy password detection. These should ideally be separate commits .

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -349,7 +354,7 @@ public void openSession() {
 	public void closeSession() {
 		log.debug("HibernateContext: closing Hibernate Session");
 		if (!participate) {
-			log.debug("Unbinding session from synchronization manager (" + sessionFactory.hashCode() + ")");
+			log.debug("Unbinding session from synchronization manager ({})", sessionFactory.hashCode());
```
jwnasambu, line 357: Kindly log format string changed from string concatenation ("... " + sessionFactory.hashCode() + ")") to SLF4J parameterized format ("... {}", sessionFactory.hashCode()). Good practice but unrelated. These should ideally be separate commits or a separate PR to keep

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -192,8 +192,9 @@ public boolean hasDuplicateUsername(String username, String systemId, Integer us
 			//Hardcoding in Luhn since past user IDs used this validator.
 			usernameWithCheckDigit = new LuhnIdentifierValidator().getValidIdentifier(username);
 		}
-		catch (Exception e) {}
-		
+		catch (Exception e) {
+			// Ignore exception, use original username if check digit generation fails
+		}
```
jwnasambu, line 197: Kindly remove these unrelated changes.

web/src/main/java/org/openmrs/web/filter/update/UpdateFilter.java
```diff
@@ -202,6 +202,7 @@ protected synchronized void doPost(HttpServletRequest httpRequest, HttpServletRe
 					Thread.sleep(3000);
 				}
 				catch (InterruptedException e) {
+					Thread.currentThread().interrupt();
```
jwnasambu, line 205: Good fix but unrelated.

api/src/main/java/org/openmrs/util/LegacyPasswordEncoder.java
```diff
@@ -0,0 +1,82 @@
...
+		StringBuilder buf = new StringBuilder(block.length * 2);
+		for (byte aBlock : block) {
+			buf.append(String.format("%02x", aBlock));
+		}
+		return buf.toString();
+	}
+	
+	private static String incorrectHexString(byte[] b) {
+		if (b == null || b.length < 1) {
+			return "";
```
jwnasambu, line 74: Kindly remove the null/empty guard from incorrectHexString since digest() never returns null, matching the style of hexString alternatively add the same guard to hexString for consistency. Either way is fine.
```suggestion
	private static String hexString(byte[] block) {
    StringBuilder buf = new StringBuilder(block.length * 2);
```

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -34,16 +34,55 @@
...
+	 */
+	public static boolean matchesPassword(String storedHash, String passwordToHash) {
+		if (storedHash == null || passwordToHash == null) {
+			throw new APIException("password.cannot.be.null", (Object[]) null);
+		}
+		if (isLegacyPasswordHash(storedHash)) {
+			return LegacyPasswordEncoder.matches(storedHash, passwordToHash);
+		}
+		return storedHash.equals(encodeString(passwordToHash));
+	}
```
jwnasambu, line 84: Kindly when isLegacyPasswordHash returns false (e.g., for a 128-char SHA-512 hash), only encodeString (SHA-512) is tried. The old hashMatches tried all three: SHA-512, SHA-1, and incorrect SHA-1. This is correct behavior for the new design since a non-legacy hash should only be verified against the current encoder. However, if somehow a non-SHA-512, non-SHA-1 hash format were stored (hypothetically), it would silently fail to authenticate.

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -550,6 +550,67 @@ public void changePassword_shouldMatchOnCorrectlyHashedSha1StoredPassword() {
 		Context.logout(); // so that the next test reauthenticates
 	}
 
+	/**
+	 * @see UserService#changePassword(String,String)
+	 */
```
jwnasambu, line 555: Kindly change the @see to point at `ContextDAO#authenticate` since those tests exercise authentication.

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -550,6 +550,67 @@ public void changePassword_shouldMatchOnCorrectlyHashedSha1StoredPassword() {
...
+		User reloadedUser = userService.getUser(user.getUserId());
+		assertNotNull(reloadedUser.getUserProperty(OpenmrsConstants.USER_PROPERTY_LEGACY_PASSWORD));
+
+		Context.removeProxyPrivilege(PrivilegeConstants.GET_USERS);
+		Context.logout();
+	}
+
+	/**
+	 * @see UserService#changePassword(String,String)
+	 */
```
jwnasambu, line 575: Same here!

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -550,6 +550,67 @@ public void changePassword_shouldMatchOnCorrectlyHashedSha1StoredPassword() {
...
+
+		User user = Context.getAuthenticatedUser();
+		assertThat(user.getUserProperty(OpenmrsConstants.USER_PROPERTY_LEGACY_PASSWORD), is(emptyString()));
+
+		Context.logout();
+	}
+
+	/**
+	 * @see UserService#changePassword(String,String)
+	 */
```
jwnasambu, line 590: Kindly change the @see to point at `ContextDAO#authenticate` since those tests exercise authentication.

jwnasambu:
@dkayiwa, @ibacher Kindly feel free to review this PR at your convenient time please!

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -89,7 +88,51 @@ public static boolean checkPassword(String storedEncodedPassword, String rawPass
...
+		if (!StringUtils.hasText(storedHash)) {
+			return false;
+		}
+		if (storedHash.startsWith(LegacyPasswordEncoder.LEGACY_HASH_PREFIX)) {
+			return true;
+		}
+		if (storedHash.length() == SHA512_HEX_LENGTH && isLowerHex(storedHash)) {
+			return false;
+		}
+		return isLowerHex(storedHash);
```
dkayiwa, line 108: `isLegacyPasswordHash` reports a 128 character lowercase hex string as not legacy, and that is what core stores today, since `Security.encodePassword` goes through `openmrsPasswordEncoder`, which is currently the legacy SHA-512 encoder. So `legacyPassword` only ever gets set on the old SHA-1 rows, and every account the epic actually needs moved onto Argon2 is reported as fine. The ticket points the same way: "Legacy algorithm: DigestUtils.sha512Hex(password)? Confirm by reading OpenMRS 1.9.x Security.hash(String). Use MessageDigest.getInstance(\"SHA-512\") without salt/iterations".

The branch already answers this question. `OpenmrsDelegatingPasswordEncoder.upgradeEncoding(hash)` is true for any row with no `{id}` prefix, which covers the SHA-1 rows and the SHA-512 rows alike; I checked it returns true for both once an `idForEncode` is configured, and false under the empty one that ships. Taking it from there also leaves one definition of legacy instead of two, since `org.openmrs.spring.LegacyOpenmrsPasswordEncoder` means SHA-512 plus the older SHA-1 variants while the new `org.openmrs.util.LegacyPasswordEncoder` means SHA-1 only, and nothing in either name warns you that they disagree.

Worth settling who consumes the flag before keeping it: nothing on this branch reads `legacyPassword`, and #6502, the lazy rehash this detection is meant to feed, asks `upgradeEncoding` rather than a user property. If you do go through the encoder, note that an unprefixed row currently gets handed to the new encoder rather than the fallback once `idForEncode` is set, and #6502 already carries that one line fix.

api/src/main/java/org/openmrs/util/Security.java
dkayiwa: It is not hypothetical any more. This branch has been rebased onto a 2.8.x that now has TRUNK-6662 (0240a505f), so writes go through `openmrsPasswordEncoder` via `Security.encodePassword` (all three `HibernateUserDAO` write paths) while `matchesPassword` reads with a hard-coded SHA-512. The day the bean is given a non-empty `idForEncode`, which is the point of TRUNK-6661, every password set after that is stored as `{id}...`, this returns false, and the account cannot log in. Nothing throws, it just reports "Invalid username and/or password", and seven of those lock the account out.

Measured rather than read off the diff: registering a stand-in modern encoder as `openmrsPasswordEncoder` in a context-sensitive test and writing the hash it produced (`{modern}MODbfd94285`) into `users.password`, `dao.authenticate("correct", "test")` returns the user on 2.8.x and throws `ContextAuthenticationException` on this branch. For that same pair `Security.checkPassword` is true on both branches and only `matchesPassword` is false. `LoginCredential.checkPassword` and both `UpdateFilter.authenticateAsSuperUser` paths moved over too, so they go the same way.

`Security.checkPassword` came in with TRUNK-6662 to be the one encoder-aware entry point, and after this change no production code calls it. I would put those four call sites back on `checkPassword` and drop `matchesPassword`. Nothing in the suite configures a non-SHA-512 encoder, so CI cannot see any of this.

ibacher:
Isn't this PR redundant with TRUNK-6662? That gives us a legacy password encoder, etc. I think what we really need is TRUNK-6666 which uses the decoder stack (see my comments on #6301) to upgrade passwords. The PR for that was close to right, it was just lacking the right implementation to build on. Basically, Spring already has the logic we need to detect legacy passwords.

dkayiwa:
Redundant, yes, and that is what the inline comment on `isLegacyPasswordHash` argues.

On Spring already having the logic, the part that does it did not survive the port. Spring's `DelegatingPasswordEncoder` sends an unprefixed row to `setDefaultPasswordEncoderForMatches`, which is how a legacy hash still verifies once you have an `idForEncode`. `OpenmrsDelegatingPasswordEncoder` drops the `fallbackEncoder` argument whenever `idForEncode` is non-empty (`defaultEncoder = idToPasswordEncoder.get(idForEncode)`) and `matches` falls back to that instead, so the row goes to the new encoder. With `idForEncode="argon2"` I measured `matches(<existing SHA-512 row>)` false whether or not `openmrsLegacyPasswordEncoder` is also in the map, and `dao.authenticate` throwing `ContextAuthenticationException` against the real context. So as merged, `upgradeEncoding` can only return true in a configuration where the account can no longer log in, and TRUNK-6666 needs that sorted before it has anything safe to act on. #6502 carries that change.

PR #6266, opened 2026-07-07
TRUNK-5466: Visit Locations in core domain model

IamMujuziMoses:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5466

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/Location.java
+++ b/api/src/main/java/org/openmrs/Location.java
@@ -95,7 +95,7 @@ public class Location extends BaseCustomizableMetadata<LocationAttribute> implem
 	private String countyDistrict;
 
 	@Column(name = "supports_visits")
-	private Boolean supportsVisits;
+	private Boolean supportsVisits = Boolean.FALSE;
 
 	@Column(name = "address3")
 	private String address3;
@@ -862,14 +862,20 @@ public class Location extends BaseCustomizableMetadata<LocationAttribute> implem
 	}
 
 	/**
-	 * {@inheritDoc}
+	 * Gets whether this location supports visits.
+	 *
+	 * @return {@code true} if this location supports visits, otherwise {@code false}
+	 * @since 3.0.0
 	 */
 	public Boolean getSupportsVisits() {
 		return supportsVisits;
 	}
 
 	/**
-	 * {@inheritDoc}
+	 * Sets supports visits for this location.
+	 *
+	 * @param supportsVisits {@code true} if this location supports visits, otherwise {@code false}
+	 * @since 3.0.0
 	 */
 	public void setSupportsVisits(Boolean supportsVisits) {
 		this.supportsVisits = supportsVisits;
--- a/api/src/main/java/org/openmrs/Location.java
+++ b/api/src/main/java/org/openmrs/Location.java
@@ -94,6 +94,9 @@ public class Location extends BaseCustomizableMetadata<LocationAttribute> implem
 	@Column(name = "county_district")
 	private String countyDistrict;
 
+	@Column(name = "supports_visits")
+	private Boolean supportsVisits;
+
 	@Column(name = "address3")
 	private String address3;
 
@@ -856,5 +859,19 @@ public class Location extends BaseCustomizableMetadata<LocationAttribute> implem
 	@Override
 	public void setAddress15(String address15) {
 		this.address15 = address15;
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public Boolean getSupportsVisits() {
+		return supportsVisits;
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public void setSupportsVisits(Boolean supportsVisits) {
+		this.supportsVisits = supportsVisits;
 	}
 }
--- a/api/src/main/java/org/openmrs/api/VisitService.java
+++ b/api/src/main/java/org/openmrs/api/VisitService.java
@@ -425,4 +425,61 @@ public interface VisitService extends OpenmrsService {
 	 */
 	@Authorized(PrivilegeConstants.EDIT_VISITS)
 	public void stopVisits(Date maximumStartDate);
+
+	/**
+	 * Gets an existing suitable visit for the patient at the specified location and time, or creates
+	 * one if none exists.
+	 * <p>
+	 * <strong>Should</strong> return an existing suitable visit when one is available, otherwise create
+	 * and save a new visit.
+	 *
+	 * @param patient the patient
+	 * @param visitTime the date and time of the visit
+	 * @param location the location where the visit occurred
+	 * @return an existing or newly created {@link Visit}
+	 */
+	@Authorized(PrivilegeConstants.GET_VISITS)
+	Visit ensureVisit(Patient patient, Date visitTime, Location location);
+
+	/**
+	 * Gets an existing suitable visit for the patient at the specified location and time, or creates
+	 * one if none exists using the specified visit type.
+	 * <p>
+	 * <strong>Should</strong> return an existing suitable visit when one is available, otherwise create
+	 * and save a new visit with the specified visit type.
+	 *
+	 * @param patient the patient
+	 * @param visitTime the date and time of the visit
+	 * @param location the location where the visit occurred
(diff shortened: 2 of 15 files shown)


IamMujuziMoses:
@djazayeri @dkayiwa @ibacher could you please review this PR, your feedback will be appreciated.

dkayiwa:
@claude review

IamMujuziMoses:
I’m currently using a temporary placeholder to determine the default VisitType, but before I finalize it I’d like to understand the intended design.

In the EMR API, the default VisitType is obtained via Metadata Mapping `EmrApiProperties#getAtFacilityVisitType()`, which isn’t available in `openmrs-core`. Should core introduce a configurable default VisitType (e.g via a global property)

api/src/main/java/org/openmrs/Location.java
```diff
@@ -94,6 +94,9 @@ public class Location extends BaseCustomizableMetadata<LocationAttribute> implem
 	@Column(name = "county_district")
 	private String countyDistrict;
 
+	@Column(name = "supports_visits")
+	private Boolean supportsVisits;
```
dkayiwa: Creating any new location fails after this upgrade. The changeset makes `supports_visits` NOT NULL, but this field stays null for every caller that never calls `setSupportsVisits`, which is all existing location-creation code (REST clients, admin UI, modules). `Location` has no `@DynamicInsert`, so Hibernate lists the column in the insert and binds an explicit NULL, which bypasses the column default and is rejected by the database. CI cannot see this because context-sensitive tests generate their schema from these annotations via hbm2ddl, which leaves the column nullable there.

I verified both halves on this branch: with the H2 column reshaped to the changeset's `NOT NULL DEFAULT FALSE`, `Context.getLocationService().saveLocation(new Location(...))` fails with `NULL not allowed for column "SUPPORTS_VISITS"`, and initializing the field makes the same save succeed. The ticket's acceptance criteria also ask for exactly this ("required and default to false"), so following the `retired` convention from `BaseOpenmrsMetadata`:

```suggestion
	@Column(name = "supports_visits")
	private Boolean supportsVisits = Boolean.FALSE;
```

One caution: don't copy the `nullable = false` that `retired` also carries, at least not without updating the test datasets. They don't populate this column, so the hbm2ddl test schema then rejects the dbunit loads (I tried it; the standard dataset no longer loads, every location test dies in setup). The initializer alone is the safe fix here.

api/src/main/java/org/openmrs/Location.java
```diff
@@ -857,4 +860,18 @@ public String getAddress15() {
...
+	public Boolean getSupportsVisits() {
+		return supportsVisits;
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public void setSupportsVisits(Boolean supportsVisits) {
+		this.supportsVisits = supportsVisits;
+	}
```
dkayiwa, line 882: There is nothing for these `{@inheritDoc}` tags to inherit, since no supertype of `Location` declares these accessors, so the rendered javadoc comes out empty. Worth swapping for a one-line description; take or leave.

api/src/main/java/org/openmrs/api/VisitService.java
```diff
@@ -425,4 +425,61 @@ public List<Visit> getVisitsByPatient(Patient patient, boolean includeInactive,
...
+	 * <strong>Should</strong> return an existing suitable visit when one is available, otherwise create
+	 * and save a new visit.
+	 *
+	 * @param patient the patient
+	 * @param visitTime the date and time of the visit
+	 * @param location the location where the visit occurred
+	 * @return an existing or newly created {@link Visit}
+	 */
+	@Authorized(PrivilegeConstants.GET_VISITS)
+	Visit ensureVisit(Patient patient, Date visitTime, Location location);
```
dkayiwa, line 443: Could you add `@since 3.0.0` to the javadoc of the new public API (the four methods here, `LocationService#getAllLocationsThatSupportVisits`, the DAO methods, and the new `Location` accessors)? The rest of these interfaces carry `@since`, and module authors rely on it to pick their minimum platform version.

api/src/main/java/org/openmrs/api/impl/VisitServiceImpl.java
```diff
@@ -447,4 +448,181 @@ public List<Class<?>> getRefTypes() {
...
+		if (visit != null) {
+			return visit;
+		}
+
+		VisitType defaultType = getDefaultVisitType(visitList);
+		if (defaultType == null) {
+			throw new APIException("Visit.error.visitType.required");
+		}
+
+		return ensureVisit(patient, visitTime, location, defaultType);
```
dkayiwa: This isn't blocking, but the delegation runs the whole suitable-visit lookup twice: the 4-arg overload immediately re-queries `dao.getSuitableVisits` and re-runs `findSuitableVisit` after this method already did both and found no match. When you rework this block for the default-type problem, it's worth restructuring so the create path reuses the list fetched here.

api/src/test/java/org/openmrs/api/VisitServiceTest.java
```diff
@@ -984,6 +984,80 @@ public void getAllVisitTypes_shouldGetAllVisitTypesBasedOnIncludeRetiredFlag() {
...
+
+		// Verify that the visit is not suitable for an unrelated location.
+		assertFalse(Context.getVisitService().isSuitableVisit(visit, other, new Date()));
+	}
+
+	/**
+	 * @see VisitService#ensureVisit(Patient, Date, Location, VisitType)
+	 */
+	@Test
+	public void ensureVisit_shouldReturnExistingSuitableVisit() {
```
dkayiwa, line 1031: None of the new tests reach the code that writes anything, which is how the broken create path (see the existing comment on `getDefaultVisitType`) went through green. Could you add tests for the writing paths: `ensureVisit` for a patient with no overlapping visit, `ensureActiveVisit` on both branches (it currently has no coverage at all), and the hierarchy fallback where the passed location doesn't support visits but an ancestor does, asserting the new visit lands on the ancestor. One small thing on the tests that do exist: `setSupportsVisits(true)` in the first `isSuitableVisit` test reads as if the flag influenced that method, but `isSuitableVisit` never consults it.

api/src/main/java/org/openmrs/api/impl/VisitServiceImpl.java
```diff
@@ -447,4 +448,181 @@ public List<Class<?>> getRefTypes() {
...
+	 * @return the default visit type
+	 * @throws APIException if no default visit type can be determined
+	 */
+	private VisitType getDefaultVisitType(List<Visit> visitList) {
+		if (visitList.isEmpty()) {
+			throw new APIException("Visit.error.visitType.required");
+		}
+
+		return visitList.getFirst().getVisitType();
+	}
```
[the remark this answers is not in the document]
dkayiwa: There is a second layer here beyond the empty-list throw: the patient's other visits are arguably the wrong source for a default type altogether. TRUNK-5466 asks for these methods to be equivalent to emrapi's AdtService, and emrapi's `buildVisit` never guesses from existing visits; it always uses the configured `emrApiProperties.getAtFacilityVisitType()`. As written, a patient mid-visit at facility A who shows up at an unrelated facility B gets A's visit type copied onto the brand-new visit at B. Since core has no counterpart of that emrapi configuration, I think this needs a deliberate decision rather than a fallback tweak. My preference would be a global property naming the default visit type, which also gives `ensureActiveVisit` something to use, as it takes no type parameter; the alternative is dropping the type-guessing overloads and making callers pass a `VisitType` explicitly.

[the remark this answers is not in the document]
IamMujuziMoses: I’am leaning strongly toward the global property.

Why?

The ticket’s goal is to move functionality from the EMR API module into core. The EMR API already has the concept of a configured default visit type, it just gets it through Metadata Mapping. A global property preserves that idea without introducing a dependency from core to a module whereas 

Removing the overloads is a much bigger API change. Unless there’s a strong desire to redesign the API, it could require updates across modules and potentially affect backward compatibility.

dkayiwa:
@claude review

dkayiwa:
@claude review

Issue #6289, opened 2026-07-10
Events/CDC (TRUNK-6429/6515/6516) sign-off: 13 verified findings across core + artemis/camel/debezium


dkayiwa:
Went through the events/CDC work (core 2.9.x and master, plus artemis/camel/debezium). I confirmed each item below by running it on a 2.9.x install against MySQL, or against the real classes, jars, and bytecode; the file:line references are from the code. Each item notes how I confirmed it.

## Confirmed

1. Outbox events are never delivered when the `@OutboxEventListener` lives in a module, which is the normal case. On a running server I registered a module `@OutboxEventListener`, fired a matching event, and the `outbox_event` row was written but stayed PENDING with the listener never called; `bc2739ac-5bf3-4ecd-8e11-d383d53695bb` was absent from `jobrunr_recurring_jobs`. In the same run `OutboxEventRegistry.hasOutboxListeners()` returned `true` (so the registry does see the module listener), and when I force-scheduled the poller myself the very same row was delivered and marked COMPLETED. So the poller logic is fine; `OutboxTaskSchedulerInitializer` just makes the schedule-or-not decision once at core startup, before modules register listeners, and never revisits it. Same code on master.

2. The outbox poller can't load module-defined event types. Same run: a module-defined `OutboxableEvent` was persisted to the outbox, then the poller's `Class.forName(item.getEventType())` (`OutboxPollingTaskHandler.java:101`; line 107 on master) threw `java.lang.ClassNotFoundException`, the row's `error_count` kept climbing, and the listener was never called. Use `OpenmrsClassLoader`/`Context.loadClass`. (This is what PR #6296 changes.)

3. `CDCEvent` on 2.9.x cannot be serialized by the outbox/broker mapper. I serialized the real `org.openmrs.event.CDCEvent` with the real `JacksonConfig.objectMapper()` and it threw `InvalidDefinitionException: Direct self-reference leading to cycle (... CDCEvent["resolvableType"] -> ResolvableType["componentType"] ...)`, because line 12 imports `org.codehaus.jackson.annotate.JsonIgnore` and FasterXML ignores it. The sibling `SaveServiceEvent` (same `getResolvableType()` getter, FasterXML `@JsonIgnore`) serialized fine in the same test. master already uses the FasterXML import.

[...]

PR #6298, opened 2026-07-13
TRUNK-6516: Fix outbox/broker issues found in events/CDC sign-off

rkorytkowski:
## Summary

Addresses findings 1, 2, 4, 5, and 7 from the events/CDC sign-off issue #6289.

- **Finding 1 — Outbox poller never starts for module listeners:** `OutboxEventRegistry` now owns the scheduler lifecycle. It calls `OutboxTaskSchedulerInitializer.schedule()` after scanning beans at startup, and exposes `registerBean(beanName)` so modules can register their `@OutboxEventListener` beans at runtime and trigger scheduling when the first listener appears. `OutboxTaskSchedulerInitializer` no longer implements `SmartInitializingSingleton` and no longer holds a reference to `OutboxEventRegistry` (no circular coupling).

- **Finding 2 — `ClassNotFoundException` for module-defined event types:** Replaced `Class.forName(item.getEventType())` with `Context.loadClass(item.getEventType())` in `OutboxPollingTaskHandler` so that the `OpenmrsClassLoader` is used and module classes resolve correctly.

- **Finding 4 — `outbox_event.payload` overflows at 64 KB:** Changed `payload` and `completed_listeners` from `TEXT` to `MEDIUMTEXT` in both `liquibase-update-to-latest-2.9.x.xml` and `liquibase-schema-only-2.9.x.xml`. Aligned `OutboxEvent` entity `columnDefinition` to `MEDIUMTEXT` on both fields for consistency.

- **Finding 5 — `BrokerIncomingEvent<?>` dead-letters every message:** `BrokerEventListenerFactory` now throws `IllegalArgumentException` at application startup when a `@BrokerEventListener` uses a wildcard or raw `BrokerIncomingEvent` type, rather than silently wiring a consumer that will fail and dead-letter every message.

- **Finding 7 — Listener exception aborts service call (documented as intentional):** Added Javadoc to `OpenmrsServiceEventAdvice` clarifying that a synchronous `@EventListener` exception propagating out and aborting the service call is by design (the listener acts as a precondition guard), and directing users to `@TransactionalEventListener` or `@OutboxEventListener` for non-blocking behaviour.

## Related

Closes / partially addresses #6289.

## Test plan

- [ ] Verify outbox poller task appears in the scheduler after a module with `@OutboxEventListener` is started at runtime
- [ ] Verify a module-defined `OutboxableEvent` round-trips through the outbox without `ClassNotFoundException`
- [ ] Verify schema creates `payload` and `completed_listeners` as `MEDIUMTEXT`
- [ ] Verify application startup fails fast with a clear error when `BrokerIncomingEvent<?>` is used
- [ ] Verify existing outbox and broker unit/integration tests pass

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
@@ -237,9 +237,8 @@ public class HibernateAdministrationDAO implements AdministrationDAO, Applicatio
 				}
 				Column column = columns.get(0);
 				String columnDefinition = column.getSqlType();
-				if (columnDefinition != null
-				        && (columnDefinition.equalsIgnoreCase("LONGTEXT") || columnDefinition.equalsIgnoreCase("MEDIUMTEXT")
-				                || columnDefinition.equalsIgnoreCase("TEXT") || columnDefinition.equalsIgnoreCase("CLOB"))) {
+				if (columnDefinition != null && (columnDefinition.equalsIgnoreCase("LONGTEXT")
+				        || columnDefinition.equalsIgnoreCase("MEDIUMTEXT"))) {
 					fieldLength = Integer.MAX_VALUE;
 				} else {
 					fieldLength = column.getLength();
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
@@ -237,7 +237,9 @@ public class HibernateAdministrationDAO implements AdministrationDAO, Applicatio
 				}
 				Column column = columns.get(0);
 				String columnDefinition = column.getSqlType();
-				if (columnDefinition != null && columnDefinition.equalsIgnoreCase("LONGTEXT")) {
+				if (columnDefinition != null
+				        && (columnDefinition.equalsIgnoreCase("LONGTEXT") || columnDefinition.equalsIgnoreCase("MEDIUMTEXT")
+				                || columnDefinition.equalsIgnoreCase("TEXT") || columnDefinition.equalsIgnoreCase("CLOB"))) {
 					fieldLength = Integer.MAX_VALUE;
 				} else {
 					fieldLength = column.getLength();
--- a/api/src/main/java/org/openmrs/event/broker/BrokerEventListenerFactory.java
+++ b/api/src/main/java/org/openmrs/event/broker/BrokerEventListenerFactory.java
@@ -56,8 +56,10 @@ public class BrokerEventListenerFactory implements EventListenerFactory, Ordered
 				// If the parameter is wrapped in a generic BrokerIncomingEvent<?>, extract the inner generic type
 				if (payloadType != null && BrokerIncomingEvent.class.isAssignableFrom(payloadType)) {
 					Class<?> genericType = resolvableType.getGeneric(0).resolve();
-					if (genericType == null) {
-						genericType = Object.class;
+					if (genericType == null || genericType == Object.class) {
+						throw new IllegalArgumentException(
+						        "BrokerEventListener on " + method + " must use a concrete type parameter, not <?> or raw. "
+						                + "Use BrokerIncomingEvent<YourDto> instead of BrokerIncomingEvent<?>");
 					}
 
 					payloadType = genericType;
--- a/api/src/main/java/org/openmrs/event/broker/BrokerIncomingEvent.java
+++ b/api/src/main/java/org/openmrs/event/broker/BrokerIncomingEvent.java
@@ -35,8 +35,7 @@ import org.openmrs.serialization.JacksonConfig;
  * {@code InputStream}. The temporary file is automatically scheduled for deletion, when the event
  * is garbage-collected.
  * <p>
- * Values from headers are deserialized to simple values or in case of complex objects from JSON
- * with {@link JacksonConfig#objectMapper()}.
+ * Values from headers are deserialized to simple values.
  *
  * @since 2.9.0
  */
--- a/api/src/main/java/org/openmrs/event/outbox/OutboxEventRegistry.java
+++ b/api/src/main/java/org/openmrs/event/outbox/OutboxEventRegistry.java
@@ -66,6 +66,7 @@ public class OutboxEventRegistry implements SmartInitializingSingleton {
 	@Override
 	public void afterSingletonsInstantiated() {
 		if (!enabled) {
+			schedulerInitializer.deleteScheduledTasks();
 			return;
 		}
 
@@ -77,6 +78,8 @@ public class OutboxEventRegistry implements SmartInitializingSingleton {
 
 		if (hasOutboxListeners()) {
 			schedulerInitializer.schedule();
+		} else {
+			schedulerInitializer.deleteScheduledTasks();
 		}
 	}
 
@@ -102,25 +105,6 @@ public class OutboxEventRegistry implements SmartInitializingSingleton {
 				}
 			}
 		});
-	}
-
-	/**
-	 * Registers any {@link OutboxEventListener} methods found on the given bean and, if this is the
-	 * first listener ever registered, starts the outbox poller. Modules must call this for each of
-	 * their beans when starting up so their listeners are discovered after core startup.
-	 *
-	 * @param beanName the Spring bean name to scan
-	 */
-	public void registerBean(String beanName) {
-		boolean wasEmpty = registry.isEmpty();
-		scanBean(beanName);
-		if (!registry.isEmpty()) {
-			Collections.sort(registry);
-			hasOutboxListenersCache.invalidateAll();
-			if (wasEmpty) {
-				schedulerInitializer.schedule();
-			}
-		}
 	}
 
 	public boolean hasOutboxListeners() {
--- a/api/src/main/java/org/openmrs/event/outbox/OutboxEventRegistry.java
+++ b/api/src/main/java/org/openmrs/event/outbox/OutboxEventRegistry.java
@@ -17,6 +17,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 
+import org.openmrs.event.outbox.tasks.OutboxTaskSchedulerInitializer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.SmartInitializingSingleton;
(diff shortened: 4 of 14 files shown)


api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.9.x.xml
```diff
@@ -57,7 +57,7 @@
 			<column name="event_type" type="VARCHAR(255)">
 				<constraints nullable="false"/>
 			</column>
-			<column name="payload" type="TEXT">
+			<column name="payload" type="MEDIUMTEXT">
```
dkayiwa, line 60: Editing this changeset in place changes its checksum, and Liquibase refuses to start against any database that already ran the original. The 2.9.x branch ships this exact changeset id with TEXT, so every 2.8 install that upgrades through 2.9 to 3.0 (plus any server already tracking this line) fails startup validation once this file differs. The validCheckSum entries scattered through the 2.0.x and 2.2.x update files are scars from past edits like this.

The in-place edit also doesn't reach the databases that most need it: anything initialized from the 2.9.x schema snapshot never re-runs this file (ChangeLogDetective only picks update files for versions greater than the snapshot), so existing installs keep TEXT and the 64 KB rollback from finding 4 stays live for exactly the servers that reported it.

Can you revert this file to TEXT and instead add a new changeset in liquibase-update-to-latest-3.0.x.xml that uses modifyDataType to convert payload and completed_listeners to MEDIUMTEXT? Two details to watch there: MySQL's MODIFY drops NOT NULL, so payload needs addNotNullConstraint (with columnDataType) after the type change, and the 2.9.x branch needs the equivalent changeset if 2.9.0 is meant to ship this fix. The schema-only snapshot change is right as-is for fresh installs.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -237,7 +237,8 @@ public long getMaximumPropertyLength(Class<? extends OpenmrsObject> aClass, Stri
 				}
 				Column column = columns.get(0);
 				String columnDefinition = column.getSqlType();
-				if (columnDefinition != null && columnDefinition.equalsIgnoreCase("LONGTEXT")) {
+				if (columnDefinition != null && (columnDefinition.equalsIgnoreCase("LONGTEXT")
+				        || columnDefinition.equalsIgnoreCase("MEDIUMTEXT"))) {
 					fieldLength = Integer.MAX_VALUE;
 				} else {
```
dkayiwa, line 243: Adding TEXT and CLOB here is what's breaking CI: BaseAttributeTypeValidatorTest.validate_shouldFailValidationIfFieldLengthsAreNotCorrect fails on the Java 25 job because datatype_config and handler_config (TEXT columns, 64 KB in MySQL) now validate as unlimited. Beyond the test, an oversize value that validation waves through comes back later as a MysqlDataTruncation that rolls back the whole business transaction, which is the same failure class finding 4 is fixing for the outbox. Only MEDIUMTEXT needs adding for this PR, and it has a real limit that can be enforced rather than treated as unbounded:

```suggestion
				if (columnDefinition != null && columnDefinition.equalsIgnoreCase("LONGTEXT")) {
					fieldLength = Integer.MAX_VALUE;
				} else if (columnDefinition != null && columnDefinition.equalsIgnoreCase("MEDIUMTEXT")) {
					// MySQL MEDIUMTEXT limit
					fieldLength = 16777215;
```

A companion case in the existing getMaximumPropertyLength tests pinning the 16777215 limit would keep this from regressing.

api/src/main/java/org/openmrs/event/outbox/OutboxEventRegistry.java
```diff
@@ -66,26 +70,57 @@ public void afterSingletonsInstantiated() {
...
-				});
-			}
+			scanBean(beanName);
 		}
 
-		// Sort all listeners once at application startup
 		Collections.sort(registry);
+
+		if (hasOutboxListeners()) {
+			schedulerInitializer.schedule();
```
dkayiwa, line 80: The un-schedule path got lost in the move. The old initializer deleted both recurring tasks when no listeners existed; those are JobRunr jobs persisted in the database, so with the delete gone, a poller that was ever scheduled keeps firing every 15 seconds forever after the last listener module is uninstalled, and likewise after an operator sets outboxevent.enabled=false (the early return above skips this whole block). It's not only wasted polling: rows still PENDING at that point are marked COMPLETED by the task handler without any listener running, so they're silently consumed. Restoring the delete for the no-listeners case and on the disabled early-return keeps the old behavior, and a context refresh or first registration re-schedules when a listener comes back.

api/src/main/java/org/openmrs/event/outbox/OutboxEventRegistry.java
```diff
@@ -66,26 +70,57 @@ public void afterSingletonsInstantiated() {
...
+	}
+
+	/**
+	 * Registers any {@link OutboxEventListener} methods found on the given bean and, if this is the
+	 * first listener ever registered, starts the outbox poller. Modules must call this for each of
+	 * their beans when starting up so their listeners are discovered after core startup.
+	 *
+	 * @param beanName the Spring bean name to scan
+	 */
+	public void registerBean(String beanName) {
```
dkayiwa: Nothing calls this method, and I don't think anything needs to. Starting a module at runtime goes through ModuleUtil.refreshApplicationContext (see ModuleFactory.startModuleInternal), which rebuilds singletons and re-runs afterSingletonsInstantiated, so module listeners are picked up and the poller scheduled without any module-side call; that ordering fix is what actually resolves finding 1. I'd drop registerBean from this PR.

If there's a registration path that skips the refresh and you keep it, it needs work before a module can use it safely: it ignores the enabled flag, so with outboxevent.enabled=false a module calling this still starts the poller and outbox processing behind the operator's back; registry is a plain ArrayList that gets mutated and re-sorted here while the poller iterates it in dispatchOutboxEvent and request threads stream over it in hasOutboxListeners, so a module starting mid-poll can throw ConcurrentModificationException, which the poller then books against the current event's error_count (CopyOnWriteArrayList, as BrokerEventListenerFactory already uses, would cover it); calling it for a bean the startup scan already registered adds duplicate entries; and the runtime-registration scenario from the test plan has no test.

api/src/test/java/org/openmrs/event/broker/BrokerEventListenerTest.java
```diff
@@ -89,7 +87,7 @@ public static class TestListener {
 		private final List<BrokerIncomingEvent> receivedEvents = new ArrayList<>();
 
 		@BrokerEventListener(value = "test-source", broker = "test-broker")
-		public void handleEvent(BrokerIncomingEvent event) {
+		public void handleEvent(BrokerIncomingEvent<String> event) {
 			receivedEvents.add(event);
 		}
```
dkayiwa, line 88: shouldReceiveEventWhenSourceAndBrokerMatch fails on the Java 25 job with this typing: the test event is built without a payload, so BrokerEvent.getResolvableType() falls back to the raw class and Spring never routes it to a listener declared BrokerIncomingEvent<String>. Constructing the events with a String payload fixes it, and it matters for all three tests here, because with no payload the two shouldFilterOut tests now pass vacuously (the event wouldn't be delivered even with matching source and broker, so they no longer exercise the filter).

This also surfaces a behavior question worth a deliberate answer: with concretely typed listeners now mandatory, an incoming message whose payload is null can't match any listener. Is dropping such messages the intended outcome for the broker modules?

api/src/main/java/org/openmrs/event/broker/BrokerEventListenerFactory.java
```diff
@@ -56,8 +56,10 @@ public ApplicationListener<?> createApplicationListener(@NonNull String beanName
 				// If the parameter is wrapped in a generic BrokerIncomingEvent<?>, extract the inner generic type
 				if (payloadType != null && BrokerIncomingEvent.class.isAssignableFrom(payloadType)) {
 					Class<?> genericType = resolvableType.getGeneric(0).resolve();
-					if (genericType == null) {
-						genericType = Object.class;
+					if (genericType == null || genericType == Object.class) {
+						throw new IllegalArgumentException(
```
dkayiwa, line 60: The fail-fast is right for finding 5, but createApplicationListener_shouldNotThrowExceptionIfRawBrokerIncomingEvent still asserts the old lenient contract and is red on the Java 25 job. Inverting it to assertThrows on the raw case (plus one for the <?> wildcard) fixes CI and gives this new rule its only test coverage.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.9.x.xml
rkorytkowski: No need. It wasn't released and it was run probably only by the two of us. Just drop the DB.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
rkorytkowski: Other DBs may have a different limit. I'd rather enforce it by DB.

rkorytkowski:
Addressed all review comments:

- **dkayiwa / Liquibase in-place edit** — No change needed; per @rkorytkowski the changeset hasn't shipped so the in-place edit is safe.
- **dkayiwa / `HibernateAdministrationDAO` MEDIUMTEXT limit** — Kept `Integer.MAX_VALUE` for `MEDIUMTEXT` per @rkorytkowski's preference to let the DB enforce the actual limit rather than hardcoding 16 777 215 in Java. `TEXT` and `CLOB` were never added (they were considered and rejected to avoid breaking `BaseAttributeTypeValidatorTest`).
- **dkayiwa / un-schedule path lost** — Added `deleteScheduledTasks()` to `OutboxTaskSchedulerInitializer` and wired it into `OutboxEventRegistry.afterSingletonsInstantiated()` for both the `outboxevent.enabled=false` early-return path and the no-listeners-found path, restoring the original behaviour of removing the JobRunr jobs when the outbox is not in use.
- **dkayiwa / `registerBean()` never called** — Removed. `ModuleUtil.refreshApplicationContext` already rebuilds singletons and re-runs `afterSingletonsInstantiated()` when a module starts at runtime, so module listeners are picked up without any module-side call.
- **dkayiwa / filter tests pass vacuously** — All three events in `BrokerEventListenerTest` now carry a `String` payload so `getResolvableType()` resolves to `BrokerIncomingEvent<String>` and the source/broker filter logic is actually exercised.
- **dkayiwa / `BrokerEventListenerFactory` test coverage** — The raw-type test was already inverted to `assertThrows`. Added a second test for the `BrokerIncomingEvent<?>` wildcard case, and added a `wildcardListener` method to `TestBean` to drive it.

api/src/main/java/org/openmrs/event/broker/BrokerIncomingEvent.java
```diff
@@ -35,8 +35,7 @@
  * {@code InputStream}. The temporary file is automatically scheduled for deletion, when the event
  * is garbage-collected.
  * <p>
- * Values from headers are deserialized to simple values or in case of complex objects from JSON
- * with {@link JacksonConfig#objectMapper()}.
+ * Values from headers are deserialized to simple values.
```
rkorytkowski, line 38: Handling JSON in header is adding too much complexity... Let's stick with simple values.

PR #6299, opened 2026-07-13
TRUNK-6516: Fix CDCEvent @JsonIgnore import to use FasterXML on 2.9.x

rkorytkowski:
## Summary

- `CDCEvent.getResolvableType()` was annotated with `org.codehaus.jackson.annotate.JsonIgnore` (Codehaus/Jackson 1.x). FasterXML (Jackson 2.x) does not honour that annotation, so the serializer follows the getter and hits a self-referential cycle: `CDCEvent["resolvableType"] -> ResolvableType["componentType"] -> ...`, throwing `InvalidDefinitionException`.
- Fix: replace with `com.fasterxml.jackson.annotation.JsonIgnore` (already correct on master).

This is finding #3 from the events/CDC sign-off issue #6289.

## Test plan

- [x] Serialize a `CDCEvent` instance with `JacksonConfig.objectMapper()` — no `InvalidDefinitionException`
- [x] Existing tests pass

### DIFF
--- a/api/src/test/java/org/openmrs/serialization/JacksonConfigTest.java
+++ b/api/src/test/java/org/openmrs/serialization/JacksonConfigTest.java
@@ -15,10 +15,12 @@ import org.junit.jupiter.api.Test;
 import org.openmrs.Patient;
 import org.openmrs.PatientIdentifier;
 import org.openmrs.PatientIdentifierType;
+import org.openmrs.event.CDCEvent;
 
 import java.util.UUID;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/openmrs/serialization/JacksonConfigTest.java
```diff
@@ -77,4 +79,24 @@ public void objectMapper_shouldSerializeAndDeserializePatientCorrectly() throws
 		assertNotNull(deserializedIdentifier.getPatient());
 		assertEquals(patient.getUuid(), deserializedIdentifier.getPatient().getUuid());
 	}
+
+	@Test
+	public void objectMapper_shouldSerializeCDCEventWithoutSelfReferentialCycle() throws Exception {
```
dkayiwa, line 84: Worth adding this same test on master too. The import there is already correct (it got fixed when the CDC work was ported in #6233), but master's JacksonConfigTest has no CDCEvent case and #6298 doesn't add one, so a future change that breaks CDCEvent serialization again would only surface at runtime when an outbox or broker listener serializes one. A tiny follow-up PR with just this test would close that gap. Doesn't block this backport.

PR #6305, opened 2026-07-13
Recursively clear a module's lib cache when the module changes

dkayiwa:
## What

`ModuleClassLoader.getLibCacheFolderForModule()` is meant to wipe a module's lib cache folder when the module file changed (or on non-optimized startup) so a modified module is fully re-expanded. It used `File.delete()`, which only removes an empty directory and silently returns `false` on a populated one, and the return value was ignored. The cache folder always holds the module jar, `lib/*.jar` and the `.moduleLastModified` marker, so the delete never ran.

The effect: any jar a module dropped between versions stays in the cache and on that module's classpath after an in-place upgrade, leaving stale classes and version or class-identity conflicts that are hard to trace because the omod on disk looks correct. Fresh installs are unaffected (empty cache), and `dispose()` on unload only clears the in-memory map rather than the folder, so the normal upgrade path does not clear it either.

## Fix

Replace both `tmpModuleDir.delete()` calls with a recursive `FileUtils.deleteDirectory()` helper that warns if deletion fails.

## Test

`ModuleClassLoaderTest#getLibCacheFolderForModule_shouldPruneJarsDroppedByAModifiedModule` primes a cache folder with a lib jar and an older `.moduleLastModified` marker, calls `getLibCacheFolderForModule` for a module whose file is newer, and asserts the jar is gone. It fails on the old `File.delete()` (`expected: <false> but was: <true>`) and passes with the recursive delete.

## How this turned up

While reviewing a module that removed a bundled `javax.jms` jar in a new version: after an in-place upgrade the old jar was still on the classpath, so the change looked like it had no effect until the module's lib cache was cleared by hand.

The same code is on the 2.x branches; I can backport once this is agreed.

### DIFF
--- a/api/src/main/java/org/openmrs/module/ModuleClassLoader.java
+++ b/api/src/main/java/org/openmrs/module/ModuleClassLoader.java
@@ -433,7 +433,8 @@ public class ModuleClassLoader extends URLClassLoader {
 	 * ease cleanup when unloading a module while openmrs is running.
 	 * <p>
 	 * The method persists lastModified of a module in .moduleLastModified file. If the value changed,
-	 * the dir will be deleted and re-created.
+	 * the dir will be deleted and re-created. The marker is only refreshed once the previous contents
+	 * were successfully removed, so a failed delete is retried on the next startup.
 	 *
 	 * @param module Module which the cache will be used for
 	 * @return File directory where the files will be placed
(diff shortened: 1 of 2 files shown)


dkayiwa:
@claude review

api/src/main/java/org/openmrs/module/ModuleClassLoader.java
```diff
@@ -488,6 +492,35 @@ public static File getLibCacheFolderForModule(Module module) {
...
+			if (!OpenmrsClassLoader.getLibCacheFolder().getCanonicalFile()
+			        .equals(tmpModuleDir.getCanonicalFile().getParentFile())) {
+				log.warn("Refusing to delete {}; it is not directly inside the lib cache folder", tmpModuleDir);
+				return;
+			}
+			FileUtils.deleteDirectory(tmpModuleDir);
+		} catch (IOException e) {
+			log.warn("Unable to delete the module lib cache folder {}", tmpModuleDir, e);
+		}
+		if (tmpModuleDir.exists()) {
```
dkayiwa, line 535: If the recursive delete only partly succeeds, this warns but getLibCacheFolderForModule still stamps the new lastModified into .moduleLastModified a few lines down, so every later optimized startup sees a matching marker, trusts the cache, and the leftover jar stays on the module classpath even though nothing holds it open anymore. That is the same failure this PR fixes, and after this one warning it goes silent. The realistic trigger is a hot upgrade on Windows: dispose() removes the map entry but never close()s the old classloader, so its jar handles stay locked until the JVM exits, and commons-io deletes what it can before throwing. I would make this method return whether the folder is actually gone (return !tmpModuleDir.exists(), with the guard refusal returning false) and skip the marker write when it is not. The next restart then takes the marker mismatch or missing path, retries the delete with the locks gone, and self-heals without anyone hand-deleting the folder, which is what the warning currently asks the admin to do.

api/src/main/java/org/openmrs/module/ModuleClassLoader.java
```diff
@@ -488,6 +492,35 @@ public static File getLibCacheFolderForModule(Module module) {
...
+	static void deleteLibCacheFolder(File tmpModuleDir) {
+		try {
+			// Guard this recursive delete on a path built from the module id: only ever remove a
+			// direct child of the lib cache root, never something a malformed id resolved to outside it.
+			if (!OpenmrsClassLoader.getLibCacheFolder().getCanonicalFile()
+			        .equals(tmpModuleDir.getCanonicalFile().getParentFile())) {
+				log.warn("Refusing to delete {}; it is not directly inside the lib cache folder", tmpModuleDir);
+				return;
+			}
+			FileUtils.deleteDirectory(tmpModuleDir);
```
dkayiwa: FileUtils.deleteDirectory throws IllegalArgumentException rather than IOException when the target exists but is a regular file (requireDirectoryExists in commons-io 2.22.0), so a stray file named exactly like the module id sitting in the lib cache root would sail past this catch and fail the module's startup, where the old File.delete() quietly removed it and recovered. A tmpModuleDir.isFile() pre-check (or widening the catch) closes it off; take it or leave it given how contrived the trigger is.

PR #6306, opened 2026-07-14
TRUNK-6678: Fix JUnit test assertions and improve disabled test documentation

imandeepduhan:
Updated JUnit test assertions to follow the correct assertEquals(expected, actual) convention and added a descriptive reason to the @Disabled annotation. These changes improve test readability, follow JUnit best practices, and make disabled tests easier to understand.

JIRA TICKET : https://openmrs.atlassian.net/browse/TRUNK-6678

### DIFF
--- a/api/src/test/java/org/openmrs/api/OpenmrsServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/OpenmrsServiceTest.java
@@ -40,7 +40,7 @@ public class OpenmrsServiceTest extends BaseContextSensitiveTest {
 	 * @throws SerializationException
 	 */
 	@Test
-	@Disabled
+	@Disabled("Temporarily disabled because it fails when executed with other tests")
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/openmrs/api/OpenmrsServiceTest.java
```diff
@@ -40,7 +40,7 @@ public class OpenmrsServiceTest extends BaseContextSensitiveTest {
 	 * @throws SerializationException
 	 */
 	@Test
-	@Disabled
+	@Disabled("Temporarily disabled because it fails when executed with other tests")
```
jwnasambu, line 43: Good start @imandeepduhan. The assertion order fix is correct. 
Kindly the @Disabled reason could be more specific  "fails when executed with other tests" is close to what the existing TODO comment already says. If you know which tests cause the conflict or what the error message is including that would help whoever eventually fixes it. Not a blocker though.

PR #6307, opened 2026-07-14
Clean up the module jar copy in ModuleClassLoader

dkayiwa:
## What

Two small cleanups to the module jar copy in `ModuleClassLoader.getModuleUrls()`:

- The copy opened its `FileInputStream`/`FileOutputStream` by hand and closed them in a `finally` that calls `close()` on references still `null` when opening the input stream throws, relying on the surrounding `catch` to swallow the resulting `NullPointerException`. Converted to try-with-resources, which closes both streams (in reverse order, with proper suppressed-exception handling, and only the ones actually opened).
- Dropped the preceding `if (!tmpModuleJar.exists()) createNewFile()` block: the `FileOutputStream` that copies into it creates the file anyway, so the pre-creation is redundant.

Behaviour is unchanged: an `IOException` while copying is still caught and logged. Pure cleanup with no functional change, so no new test; the existing module-loading tests exercise this path.


dkayiwa:
@claude review

PR #6308, opened 2026-07-14
Fail module load when its classpath cannot be built

dkayiwa:
## What

When a module loads, `ModuleClassLoader.getModuleUrls()` copies the module's own jar into the lib cache and expands its `lib` folder. On an `IOException` in either step it only logged a warning, then built the classloader anyway (missing main jar, or an incompletely expanded lib folder). The module was then marked started but had missing classes and failed later with cryptic `ClassNotFound` errors that don't point at the real cause.

This throws a `ModuleException` in both cases instead. `ModuleFactory.startModuleInternal()` already wraps `new ModuleClassLoader(...)` in a try/catch that records the startup error, notifies admins, and rolls the module back via `stopModule`, without aborting the other modules' startup. So a module whose classpath can't be built (unreadable or corrupt omod, disk full) is now reported as failed to start with the real cause, instead of appearing started but broken.

The per-file `MalformedURLException` while adding an individual lib jar to the classpath stays a warning, since it is not a whole-module failure.

## Tests

- `moduleClassLoader_shouldFailToLoadWhenTheModuleJarCannotBeCopied`: points a module at a nonexistent file and asserts construction throws `ModuleException`.
- `moduleClassLoader_shouldFailToLoadWhenTheLibFolderCannotBeExpanded`: points a module at a real file that is not a valid jar (copy succeeds, expand fails) and asserts construction throws `ModuleException`.
- `ModuleFactoryTest` (real module loading) stays green at 9/9, confirming valid modules are unaffected.

## Note

Touches the same jar-copy block as #6307 (the try-with-resources cleanup); whichever merges second needs a trivial rebase.

### DIFF
--- a/api/src/main/java/org/openmrs/module/ModuleClassLoader.java
+++ b/api/src/main/java/org/openmrs/module/ModuleClassLoader.java
@@ -233,7 +233,8 @@ public class ModuleClassLoader extends URLClassLoader {
 			        FileOutputStream out = new FileOutputStream(tmpModuleJar)) {
 				OpenmrsUtil.copyFile(in, out);
 			} catch (IOException io) {
-				log.warn("Unable to copy tmpModuleFile", io);
+				throw new ModuleException(
+				        "Unable to copy the module jar for " + module.getModuleId() + " into the lib cache", io);
 			}
 
 			// add the module jar as a url in the classpath of the classloader
@@ -299,7 +300,7 @@ public class ModuleClassLoader extends URLClassLoader {
 		} catch (MalformedURLException e) {
 			log.warn("Error while adding module 'lib' folder to URL result list");
 		} catch (IOException io) {
-			log.warn("Error while expanding lib folder", io);
+			throw new ModuleException("Unable to expand the lib folder for " + module.getModuleId(), io);
 		}
 
 		// add each xml document to the url list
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/ModuleClassLoader.java
```diff
@@ -244,7 +244,8 @@ private static List<URL> getModuleUrls(final Module module) {
 				out = new FileOutputStream(tmpModuleJar);
 				OpenmrsUtil.copyFile(in, out);
 			} catch (IOException io) {
-				log.warn("Unable to copy tmpModuleFile", io);
+				throw new ModuleException(
+				        "Unable to copy the module jar for " + module.getModuleId() + " into the lib cache", io);
```
ibacher, line 237: I think that logging here is still valuable no? Ditto below...

dkayiwa: Nothing goes unlogged with the throw. The only place that constructs a ModuleClassLoader at runtime is ModuleFactory.startModuleInternal(), and its catch block logs this exception at ERROR with the full stack trace, records it as the module's startup error message, and notifies the super users. Logging a warning here as well would report the same failure twice, once as a warning that looks survivable and then again as the actual startup failure. The lib folder case below surfaces through that same catch block. (getModuleUrls() is also referenced from modulesSetChanged(), but nothing calls that method.)

PR #6230, opened 2026-06-29
TRUNK-6666: Automatic Password Upgrade on Successful Login

solomonfortune:
## Description of what I changed
- Added `spring-security-crypto` dependency for Argon2id password hashing support
- Increased `users.password` column length from 128 to 255 via Liquibase migration to accommodate Argon2id hash output
- Added `isLegacyHash()` to `Security.java` to detect all three legacy hash formats (SHA-512, SHA-1 correct, SHA-1 buggy from ticket #1178)
- Added `Argon2PasswordEncoder` instance to `Security.java` with production-ready default parameters
- Added unit tests for `isLegacyHash()` covering all legacy formats and edge cases

Note: This is a draft PR. Tasks remaining — upgrade hook in `authenticate()`, transaction isolation, and thread safety — will be added in subsequent commits.

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-6666

## Checklist: I completed these to help reviewers :)
- [x] My IDE is configured to follow the code style of this project.
- [x] I have added tests to cover my changes.
- [x] I ran `mvn clean package` right before creating this pull request and added all formatting changes to my commit.
- [x] All new and existing tests passed.
- [ ] My pull request is based on the latest changes of the master branch.

## Performance Impact (2.8.9)

Deployments using `Authorization: Basic` headers on per-request API calls (e.g. webservices.rest `AuthorizationFilter`, fhir2 `AuthenticationFilter`) will experience a one-time increase in authentication latency (~18ms, 16MiB per call) for any account still on a legacy SHA-512/SHA-1 hash, until that account's first successful login upgrades it to Argon2id. After upgrade the cost returns to normal. Integrations making per-request Basic auth calls should be updated to reuse sessions.

PR #6311, opened 2026-07-14
TRUNK-6672: Add Password Hash Algorithm Detection

jwnasambu:

I introduced password hash format detection for Argon2 based passwords allowing the authentication system to identify modern password hashes and route them to the appropriate encoder. This is detection-only — no stored passwords are modified and no password migration occurs.

## Issue I worked on

see https://openmrs.atlassian.net/browse/TRUNK-6672

  No? -> execute above command


jwnasambu:
@claude review

jwnasambu:
@claude review

solomonfortune (commented):
Hey @jwnasambu, this looks like the same salt issue I ran into on #6230 (TRUNK-6666) while building the Argon2id lazy-rehash path. 
Both real callers of hashMatches append the OpenMRS salt before calling it (LoginCredential.checkPassword does pw + getSalt(), and the DB-auth path in HibernateContextDAO does password + saltOnRecord). Argon2's encoded hash already carries its own salt, so Argon2PasswordEncoder.verify() needs the bare password, not password+externalSalt. Passing the concatenated value in will fail verification for every real Argon2 account. 
In #6230 I consolidated this into a single Security.passwordMatches(storedHash, rawPassword, salt) helper that branches on legacy vs. Argon2id and only applies the OpenMRS salt on the legacy branch. Worth noting though is that @ibacher flagged in review that this shouldn't be adding new public API surface, so the exact shape of that helper (and whether it stays public) is still being worked out. Treat it as a reference for the fix pattern rather than a finished API to copy. Happy to sync once it settles if that'd help.

jwnasambu:
@claude review

jwnasambu:
@claude review

jwnasambu:
@dkayiwa, @ibacher, kindly feel free to review my PR at your convenience. Thank you!

solomonfortune (commented):
@jwnasambu 
Thanks for cleaning up the unused SHA_512/SHA_1 placeholders. I think that's resolved. The other three items from the last review still need attention though, most importantly the salt-handling bug in Security.java. hashMatches is still receiving password + salt from both real callers which are LoginCredential.checkPassword and HibernateContextDAO's DB-auth path, but Argon2PasswordEncoder.verify() needs the bare password since Argon2's hash already embeds its own salt. This will fail to authenticate any real Argon2 account. This is the same issue I hit on #6230.
Also, the argon2-jvm LGPL-3.0 license question from the earlier review. This is also not yet addressed and it may be the project's first LGPL runtime dependency. The wipeArray() suggestion on Argon2PasswordEncoder.verify() — that thread shows as "Outdated" but I don't think the fix actually landed, looks like it just got shifted by the merge commit. Worth confirming.

jwnasambu:
@solomonfortune Thank you for the feedback! You're absolutely right about the salt-handling issue—passing password + salt directly to Argon2 will break real logins. Because of this (and the licensing/dependency concerns), I ended up stripping out the verification logic (Argon2PasswordEncoder) in the latest commits to keep this PR strictly limited to the detection utility.

However, as the bot pointed out, this leaves the detector with no production callers. As a way forward, I have kept this PR this PR strictly limited to adding the PasswordHashDetector and the enum, leaving the actual wiring/verification to be integrated alongside your work in #6230. Kindly let me know if the is something I am messing out based on your explanation and I will gladly fix it!"

api/src/main/java/org/openmrs/util/PasswordHashDetector.java
```diff
@@ -0,0 +1,99 @@
...
+	 * @param hashedPassword the stored password hash to check
+	 * @return true if the hash matches the Argon2 format, false otherwise
+	 */
+	public static boolean isArgon2Hash(String hashedPassword) {
+		if (hashedPassword == null || hashedPassword.trim().isEmpty()) {
+			return false;
+		}
+		String s = hashedPassword.trim();
+		// Match standard Argon2 encoded hashes: $argon2id$..., $argon2i$..., $argon2d$...
+		return s.matches("^\\$argon2(?:id|i|d)\\$.*");
```
ibacher: All the regexes are static and should be statically compiled rather than defined inline. I'm not sure we should be building an Argon2Hash detector. There's got to be part of Spring's machinery (which is generating this) we can reuse for this and then fallback to the length-based detection.

jwnasambu:
@dkayiwa, @ibacher, kindly feel free to review my PR at your convenience. Thank you!

ibacher:
Strictly-speaking I don't see a real reason that we need this? The only place we should be loading or caring about the hash at all is in the encoder and that will already recognise these.

PR #6317, opened 2026-07-15
Require privileges for unprotected CohortService methods

dkayiwa:
### Summary
Two methods on `CohortService` were missing the `@Authorized` annotation that guards the rest of the service.

**`purgeCohort(Cohort)`** permanently removes a cohort from the database (not reversible), but unlike every other mutating method on the service it had no `@Authorized` annotation. `voidCohort` (the reversible delete) requires `Delete Cohorts` and `purgeCohortMembership` requires `Edit Cohorts`, yet `purgeCohort` required nothing. A user holding only `Get Patient Cohorts` could therefore permanently delete any cohort through the REST purge path, `DELETE /ws/rest/v1/cohort/{uuid}?purge=true`. A dedicated `Purge Cohorts` privilege already exists in `PrivilegeConstants` but was never applied to this method.

**`getCohorts(String)`** was the only read on the service with no `@Authorized`, reading straight from the DAO while every sibling getter (`getCohort`, `getCohortByName`, `getAllCohorts`, `getCohortByUuid`, and so on) requires `Get Patient Cohorts`. A direct service-layer caller could therefore list cohorts (names, descriptions, member patient ids) with no authorization. The REST search path `GET /ws/rest/v1/cohort?q=` was only covered incidentally, by the resource's required-get privilege being enforced at representation-conversion time, not by the service itself.

### Change
- Annotate `purgeCohort` with `@Authorized({ PrivilegeConstants.PURGE_COHORTS })`.
- Annotate `getCohorts(String)` with `@Authorized({ PrivilegeConstants.GET_PATIENT_COHORTS })`.
- Add regression tests asserting each method throws `APIAuthenticationException` for a caller without the privilege, so dropping either annotation fails the build instead of silently reopening the gap.

### Notes
The REST layer in `openmrs-module-webservices.rest` adds no authorization of its own for the purge path, so the check belongs here at the service layer. Addresses private advisory GHSA-4w59-mg6c-76ch.

### DIFF
--- a/api/src/main/java/org/openmrs/api/CohortService.java
+++ b/api/src/main/java/org/openmrs/api/CohortService.java
@@ -150,6 +150,7 @@ public interface CohortService extends OpenmrsService {
 	 * @return list of cohorts matching the name fragment
 	 * @throws APIException
 	 */
+	@Authorized({ PrivilegeConstants.GET_PATIENT_COHORTS })
 	public List<Cohort> getCohorts(String nameFragment) throws APIException;
 
 	/**
--- a/api/src/main/java/org/openmrs/api/CohortService.java
+++ b/api/src/main/java/org/openmrs/api/CohortService.java
@@ -78,6 +78,7 @@ public interface CohortService extends OpenmrsService {
 	 * @param cohort the Cohort to completely remove from the database
 	 * @throws APIException
 	 */
+	@Authorized({ PrivilegeConstants.PURGE_COHORTS })
 	public Cohort purgeCohort(Cohort cohort) throws APIException;
 
 	/**
--- a/api/src/test/java/org/openmrs/api/CohortServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/CohortServiceTest.java
@@ -123,6 +123,17 @@ public class CohortServiceTest extends BaseContextSensitiveTest {
 	 * @see CohortService#purgeCohort(Cohort)
 	 */
 	@Test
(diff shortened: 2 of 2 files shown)


dkayiwa:
@claude review

api/src/main/java/org/openmrs/api/CohortService.java
```diff
@@ -78,6 +78,7 @@ public interface CohortService extends OpenmrsService {
 	 * @param cohort the Cohort to completely remove from the database
 	 * @throws APIException
 	 */
+	@Authorized({ PrivilegeConstants.PURGE_COHORTS })
```
dkayiwa, line 81: The only purgeCohort test runs as the superuser, so this new authorization gate has no regression guard. If the annotation ever gets dropped in a refactor or a bad merge, nothing will fail: the happy path throws nothing, CI stays green, and the vulnerability quietly comes back. That silent-regression window is why I'd still add a test even though the fix is a one-liner, though I wouldn't hold the merge on it. The test below passes on this branch and fails on master:

```java
/**
 * @see CohortService#purgeCohort(Cohort)
 */
@Test
public void purgeCohort_shouldFailIfUserDoesNotHaveThePurgeCohortsPrivilege() {
	executeDataSet(COHORT_XML);
	Cohort cohort = service.getAllCohorts(true).get(0);
	Context.logout();
	assertThrows(APIAuthenticationException.class, () -> service.purgeCohort(cohort));
}
```

PR #6318, opened 2026-07-15
Require Get Patient Programs privilege for by-uuid program lookups

dkayiwa:
### Summary
`ProgramWorkflowService.getPatientProgramByUuid(String)` and `getPatientStateByUuid(String)` carry no `@Authorized` annotation. OpenMRS enforces service authorization in `AuthorizationAdvice.before()`, which only checks when the method declares at least one privilege; a method with none falls through with no check, not even an authentication requirement. So any authenticated user (and, per the advice logic, effectively any caller reaching the service) can read patient program-enrollment data by uuid.

Every other read on the service already requires `Get Patient Programs`, including `getPatientProgram(Integer)`, `getPatientProgram(...)`, and the other lookups.

### Change
Annotate both by-uuid methods with `@Authorized({ PrivilegeConstants.GET_PATIENT_PROGRAMS })`, matching the existing reads.

### Notes
Addresses private advisory GHSA-gqf9-38mg-wgqg.

### DIFF
--- a/api/src/main/java/org/openmrs/api/ProgramWorkflowService.java
+++ b/api/src/main/java/org/openmrs/api/ProgramWorkflowService.java
@@ -594,6 +594,7 @@ public interface ProgramWorkflowService extends OpenmrsService {
 	@Authorized({ "Get Patient Programs" })
 	public PatientProgramAttribute getPatientProgramAttributeByUuid(String var1);
 
+	@Authorized({ PrivilegeConstants.GET_PATIENT_PROGRAMS })
 	public Map<Object, Object> getPatientProgramAttributeByAttributeName(List<Integer> patients, String attributeName);
 
 	@Transactional(readOnly = true)
--- a/api/src/main/java/org/openmrs/api/ProgramWorkflowService.java
+++ b/api/src/main/java/org/openmrs/api/ProgramWorkflowService.java
@@ -232,6 +232,7 @@ public interface ProgramWorkflowService extends OpenmrsService {
 	 * @param uuid the universally unique identifier
 	 * @return the program which matches the given uuid
 	 */
+	@Authorized({ PrivilegeConstants.GET_PATIENT_PROGRAMS })
 	public PatientState getPatientStateByUuid(String uuid);
 
 	/**
@@ -506,6 +507,7 @@ public interface ProgramWorkflowService extends OpenmrsService {
 	 * @param uuid the universally unique identifier
 	 * @return the patient program which matches the given uuid
 	 */
+	@Authorized({ PrivilegeConstants.GET_PATIENT_PROGRAMS })
 	public PatientProgram getPatientProgramByUuid(String uuid);
 
 	/**
--- a/api/src/test/java/org/openmrs/api/ProgramWorkflowServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/ProgramWorkflowServiceTest.java
@@ -468,6 +468,21 @@ public class ProgramWorkflowServiceTest extends BaseContextSensitiveTest {
 	}
 
 	/**
+	 * @see ProgramWorkflowService#getPatientProgramByUuid(String)
+	 */
+	@Test
+	public void getPatientProgramByUuid_shouldFailIfUserHasNoGetPatientProgramsPrivilege() {
+		// a valid uuid so the missing privilege is the only possible reason for failure
+		String uuid = "2edf272c-bf05-4208-9f93-2fa213ed0415";
+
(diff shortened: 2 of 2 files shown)


api/src/main/java/org/openmrs/api/ProgramWorkflowService.java
```diff
@@ -506,6 +507,7 @@ public void purgeConceptStateConversion(ConceptStateConversion conceptStateConve
 	 * @param uuid the universally unique identifier
 	 * @return the patient program which matches the given uuid
 	 */
+	@Authorized({ PrivilegeConstants.GET_PATIENT_PROGRAMS })
```
dkayiwa, line 510: Worth adding the regression guard that #6317 gave CohortService before this ships: with no test pinning the gate, a future edit that drops one of these annotations reopens GHSA-gqf9-38mg-wgqg with CI green, and nothing throws to make anyone notice. I ran the tests below on this branch and they pass; on the merge-base all three fail, with the logged out calls completing and the two by-uuid ones handing back the patient objects.

```java
@Test
public void getPatientProgramByUuid_shouldFailIfUserDoesNotHaveTheGetPatientProgramsPrivilege() {
	Context.logout();
	APIAuthenticationException exception = assertThrows(APIAuthenticationException.class,
	    () -> Context.getProgramWorkflowService().getPatientProgramByUuid("2edf272c-bf05-4208-9f93-2fa213ed0415"));
	assertThat(exception.getMessage(), containsString(PrivilegeConstants.GET_PATIENT_PROGRAMS));
}

@Test
public void getPatientStateByUuid_shouldFailIfUserDoesNotHaveTheGetPatientProgramsPrivilege() {
	Context.logout();
	APIAuthenticationException exception = assertThrows(APIAuthenticationException.class,
	    () -> Context.getProgramWorkflowService().getPatientStateByUuid("ea89deaa-23cc-4840-92fe-63d199c37e4c"));
	assertThat(exception.getMessage(), containsString(PrivilegeConstants.GET_PATIENT_PROGRAMS));
}

@Test
public void getPatientProgramAttributeByAttributeName_shouldFailIfUserDoesNotHaveTheGetPatientProgramsPrivilege() {
	Context.logout();
	APIAuthenticationException exception = assertThrows(APIAuthenticationException.class,
	    () -> Context.getProgramWorkflowService().getPatientProgramAttributeByAttributeName(
	        Collections.singletonList(2), "some attribute"));
	assertThat(exception.getMessage(), containsString(PrivilegeConstants.GET_PATIENT_PROGRAMS));
}
```

They drop into `ProgramWorkflowServiceTest` next to the existing by-uuid tests and reuse the same dataset uuids. Asserting the message keeps the exact privilege pinned, same reasoning as the #6317 tests.

dkayiwa (commented):
Nothing here blocks merging; the annotations close exactly the gap the advisory describes.

One thing with no diff line to sit on: the metadata reads on this service still carry no `@Authorized` at all. `getProgramByUuid` has none while `getProgram`, `getProgramByName`, `getAllPrograms` and `getPrograms` all require Get Programs; `getConceptStateConversionByUuid` and the workflow-and-trigger overload of `getConceptStateConversion` have none while the Integer variant and `getAllConceptStateConversions` do; `getWorkflow`, `getWorkflowByUuid`, `getState` and `getStateByUuid` have nothing either. With no annotation, `AuthorizationAdvice.before()` skips the check entirely, so all of these stay readable without authentication, through the same mechanism this PR closes for patient data. Program and workflow names are nowhere near patient data in sensitivity, and keeping this PR scoped to the advisory makes sense, but it deserves a follow-up ticket so the rest of the gap does not get lost once the advisory closes.

api/src/main/java/org/openmrs/api/ProgramWorkflowService.java
```diff
@@ -592,6 +594,7 @@ public void purgeConceptStateConversion(ConceptStateConversion conceptStateConve
 	@Authorized({ "Get Patient Programs" })
 	public PatientProgramAttribute getPatientProgramAttributeByUuid(String var1);
 
+	@Authorized({ PrivilegeConstants.GET_PATIENT_PROGRAMS })
```
dkayiwa, line 597: The title and description still cover only the two by-uuid lookups, so this annotation from 821f32ce has no trace in either. The advisory writeup and any backport picks get planned from the description, and a cherry-pick scoped by it would miss this method. Worth updating the description to name `getPatientProgramAttributeByAttributeName` too, and correcting its claim that every other read on the service already requires Get Patient Programs, which this method contradicted until the second commit.

ibacher (approved):

PR #6324, opened 2026-07-15
Require allergy privileges for getAllergies and setAllergies

dkayiwa:
### Summary
`PatientService.getAllergies(Patient)` and `setAllergies(Patient, Allergies)` have no `@Authorized` annotation. Authorization is enforced in `AuthorizationAdvice.before()` only when a method declares at least one privilege; a method with none falls through with no check, not even authentication. So any authenticated caller who can reach these entry points can read another patient's allergy list and persist allergy changes for a patient, even though the dedicated per-item methods are guarded (`getAllergy`/`getAllergyByUuid` require `Get Allergies`; `saveAllergy` requires `Add`/`Edit Allergies`).

### Change
- `getAllergies` → `@Authorized({ GET_ALLERGIES })`
- `setAllergies` → `@Authorized({ ADD_ALLERGIES, EDIT_ALLERGIES })`
- Internal reads that would otherwise transitively demand `Get Allergies` from callers acquire it as a proxy privilege instead: the duplicate-allergen check in `AllergyValidator`, and the allergy reads `mergePatients` makes while reconciling the two patients' allergy lists.
- Regression tests asserting each method now requires its privilege, and that a caller holding only an allergy write privilege can still save a new allergy.
- The merge workflow tests now run `mergePatients` as a non-super user holding only the privileges the merge legitimately requires, with `Get Allergies` and `Get Patient Cohorts` deliberately absent, so the internal-read proxying is guarded by all 27 audit-based merge tests (suggested by @ibacher in review).
- Running the merge as a non-super user surfaced one more leaked internal read: `CohortServiceImpl.notifyPatientVoided` reads cohort memberships through the service proxy, so voiding any patient demanded `Get Patient Cohorts` from the voiding user even though `PatientDataVoidHandler` already proxies `Edit Cohorts` for it (the annotation predates #6317). That read now acquires its own proxy privilege, matching the allergy reads.

### Scope note
The advisory also raises that `setAllergies` does not bind each submitted `Allergy` to the outer `patient` argument (the DAO `persist`s each allergy as-is). That is a separate data-integrity hardening that interacts with the void/recreate reconciliation logic in `setAllergies`, so I'm handling it in a follow-up with its own tests rather than bundling it here. This PR closes the missing-authorization half.

Addresses GHSA-jxvr-3qjq-vfqg.

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -747,8 +747,18 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 		// so moving a duplicate would make the survivor's allergy list fail to load afterwards
 		// TODO: this should be a copy, not a move
 		PatientService patientService = Context.getPatientService();
-		Allergies preferredAllergies = patientService.getAllergies(preferred);
-		for (Allergy allergy : patientService.getAllergies(notPreferred)) {
+		// these are internal reads made on behalf of the merge; mergePatients is authorized with
+		// Edit Patients, so acquire Get Allergies here rather than forcing the merging user to hold it
+		Allergies preferredAllergies;
+		Allergies notPreferredAllergies;
+		try {
+			Context.addProxyPrivilege(PrivilegeConstants.GET_ALLERGIES);
+			preferredAllergies = patientService.getAllergies(preferred);
+			notPreferredAllergies = patientService.getAllergies(notPreferred);
+		} finally {
+			Context.removeProxyPrivilege(PrivilegeConstants.GET_ALLERGIES);
+		}
+		for (Allergy allergy : notPreferredAllergies) {
 			if (preferredAllergies.containsAllergen(allergy)) {
 				log.debug("Skipping allergy {}; preferred patient {} already has the same allergen", allergy.getAllergyId(),
 				    preferred.getPatientId());
--- a/api/src/main/java/org/openmrs/api/PatientService.java
+++ b/api/src/main/java/org/openmrs/api/PatientService.java
@@ -788,6 +788,7 @@ public interface PatientService extends OpenmrsService {
 	 * @param patient the patient
 	 * @return the allergies object
 	 */
+	@Authorized({ PrivilegeConstants.GET_ALLERGIES })
 	Allergies getAllergies(Patient patient);
 
 	/**
@@ -815,6 +816,7 @@ public interface PatientService extends OpenmrsService {
 	 * @param allergies the allergies
 	 * @return the saved allergies
 	 */
+	@Authorized({ PrivilegeConstants.ADD_ALLERGIES, PrivilegeConstants.EDIT_ALLERGIES })
 	Allergies setAllergies(Patient patient, Allergies allergies);
 
 	/**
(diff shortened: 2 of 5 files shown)


api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -747,8 +747,18 @@ private void mergeAllergies(Patient preferred, Patient notPreferred, PersonMerge
...
+		// Edit Patients, so acquire Get Allergies here rather than forcing the merging user to hold it
+		Allergies preferredAllergies;
+		Allergies notPreferredAllergies;
+		try {
+			Context.addProxyPrivilege(PrivilegeConstants.GET_ALLERGIES);
+			preferredAllergies = patientService.getAllergies(preferred);
+			notPreferredAllergies = patientService.getAllergies(notPreferred);
+		} finally {
+			Context.removeProxyPrivilege(PrivilegeConstants.GET_ALLERGIES);
+		}
```
dkayiwa, line 760: This proxy block is the one piece of the change that no test actually exercises. I confirmed it by reverting both wrappings on top of the new annotations: the saveAllergy duplicate-check test then fails with `Privileges required: Get Allergies`, but all four mergePatients allergy tests still pass, because they run as the super-user who already holds Get Allergies. If this block were dropped in a later refactor, CI would stay green while mergePatients started throwing APIAuthenticationException for any user who has Edit Patients but not Get Allergies.

The validator half of the same fix does have a guard test (the butch / Add-Allergies-only case), so this is the asymmetric spot. I would mirror that test here: become a user who holds the merge privileges but not Get Allergies, merge a patient who has an allergy, and assert it succeeds. The wrinkle is that mergePatients pulls in a fairly broad privilege set (relationships, obs, conditions, medication dispenses, patient programs, and so on), so the setup is bulkier than the saveAllergy test. If that is more scaffolding than it is worth, even a short comment here noting that the wrapping is only covered under super-user privileges would flag it for the next reader. Not blocking.

ibacher: It might be useful to ensure that the patient merge workflows are run with a non-super user user (which would patch the missing regression test here).

ibacher (approved):

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
dkayiwa: Done in bf4ef2251. `mergeAndRetrieveAudit` now runs the merge itself as butch (whose role grants no privileges) with proxy grants for only the privileges the merge legitimately requires, and Get Allergies deliberately left out. That closes the gap flagged above: with the proxy block reverted, all 27 audit-based merge workflow tests fail with `Privileges required: Get Allergies`, where previously all of them passed.

Running the merge without super-user privileges immediately paid off beyond allergies: it surfaced that voiding the losing patient demands Get Patient Cohorts from the merging user. `PatientDataVoidHandler` proxies Edit Cohorts around `notifyPatientVoided`, but the `getCohortMemberships` call inside that method goes through the service proxy, and its `@Authorized` annotation predates #6317, so any user voiding a patient without that privilege has been failing silently for a while. I gave that read the same proxy treatment as the allergy reads, and the test's privilege set omits Get Patient Cohorts too, so both internal reads are now guarded the same way (reverting the cohort proxy also fails all 27 tests, with `Privileges required: Get Patient Cohorts`).

The privileges that remain in the set are genuine demands of the workflow, including a few metadata reads (visit attribute types, the overlapping-visits global property, concepts, locations) that validators and change checks make through service proxies while the moved data is saved. Some of those could arguably get the internal-read treatment as well, but they are pre-existing and benign to hold, so the test documents them as-is rather than growing this PR further.

PR #6325, opened 2026-07-15
Check Task type before instantiating scheduler task classes

dkayiwa:
### Summary
`LegacyTask.execute` and `SchedulerFormValidator.validate` load the configured `task_class_name` through `OpenmrsClassLoader` and call `newInstance()` on it **before** checking whether the result implements `Task`:

- `api/.../scheduler/LegacyTask.java` — `taskClass.getDeclaredConstructor().newInstance()` then `if (instance instanceof Task)`
- `api/.../validator/SchedulerFormValidator.java` — `taskClass.newInstance()` then `if (!(o instanceof Task))`

So configuring a scheduler task with any fully-qualified class name on the classpath runs that class's constructor and static initializer as a side effect, even when it isn't a `Task`.

### Change
Check `Task.class.isAssignableFrom(taskClass)` first, and only instantiate classes that implement `Task`. The validator still instantiates a genuine `Task` afterward so the existing instantiation/access error reporting is preserved.

### Severity
Reaching either path requires the `Manage Scheduler` privilege, and such a user can already schedule a real `Task` that runs arbitrary code, so this is defense in depth (no new privilege boundary crossed) rather than a critical bug.

### Tests
- New `validate_shouldRejectANonTaskClassBeforeInstantiatingIt` uses a sentinel non-`Task` class with a constructor side effect and asserts it is rejected without being instantiated.
- The existing `validate_shouldFailValidationIfClassCannotBeInstantiated` now points at an abstract `Task` subtype so the instantiation-failure branch is still exercised after the type check.

Addresses GHSA-6ggw-67pv-j87p.

### DIFF
--- a/api/src/main/java/org/openmrs/validator/SchedulerFormValidator.java
+++ b/api/src/main/java/org/openmrs/validator/SchedulerFormValidator.java
@@ -71,11 +71,12 @@ public class SchedulerFormValidator implements Validator {
 			try {
 				Class<?> taskClass = OpenmrsClassLoader.getInstance().loadClass(taskDefinition.getTaskClass());
 
-				Object o = taskClass.newInstance();
-				if (!(o instanceof Task)) {
+				if (!Task.class.isAssignableFrom(taskClass)) {
 					errors.rejectValue("taskClass", "Scheduler.taskForm.classDoesNotImplementTask",
 					    new Object[] { taskDefinition.getTaskClass(), Task.class.getName() },
 					    "Class does not implement Task interface");
+				} else {
+					taskClass.newInstance();
 				}
 
 			} catch (IllegalAccessException iae) {
--- a/api/src/main/java/org/openmrs/scheduler/LegacyTask.java
+++ b/api/src/main/java/org/openmrs/scheduler/LegacyTask.java
@@ -28,14 +28,12 @@ public class LegacyTask implements TaskHandler<TaskDefinition> {
 		} catch (ClassNotFoundException e) {
 			throw new TaskException("Task class " + taskData.getTaskClass() + " not found", false);
 		}
-		Object instance = taskClass.getDeclaredConstructor().newInstance();
-
-		if (instance instanceof Task) {
-			Task task = (Task) instance;
-			task.initialize(taskData);
-			task.execute();
-		} else {
+		if (!Task.class.isAssignableFrom(taskClass)) {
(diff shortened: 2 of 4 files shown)


api/src/main/java/org/openmrs/validator/SchedulerFormValidator.java
```diff
@@ -71,11 +71,12 @@ public void validate(Object obj, Errors errors) {
...
 				Class<?> taskClass = OpenmrsClassLoader.getInstance().loadClass(taskDefinition.getTaskClass());
 
-				Object o = taskClass.newInstance();
-				if (!(o instanceof Task)) {
+				if (!Task.class.isAssignableFrom(taskClass)) {
 					errors.rejectValue("taskClass", "Scheduler.taskForm.classDoesNotImplementTask",
 					    new Object[] { taskDefinition.getTaskClass(), Task.class.getName() },
 					    "Class does not implement Task interface");
+				} else {
+					taskClass.newInstance();
```
ibacher, line 79: Um... what's the point of calling `newInstance()` here when we do exactly nothing with the instance?

ibacher (commented):
Definitely an improvement, but see my comment below. I don't think we've regarded side-effects of constructors as load-bearing, especially when we assume it's a no-args constructor.

api/src/main/java/org/openmrs/validator/SchedulerFormValidator.java
dkayiwa: Fair, the instance is discarded, and it was in the original too (it only ever used `o instanceof Task`). The one reason it's still here is the throw: if a genuine Task class can't be constructed (abstract, no accessible no-arg constructor, or the constructor itself throws), the catch blocks turn that into the `instantiationException`/`illegalAccessException` field errors, which `validate_shouldFailValidationIfClassCannotBeInstantiated` pins.

I'm not attached to keeping it. If we don't think that save-time check pulls its weight, I'm happy to drop the `newInstance()` and let `LegacyTask` surface instantiation failures at run time, since the type check is the actual fix. Otherwise I'll add a comment so it stops reading as dead code. Your call.

On the broader point, agreed this is defense in depth rather than a class of bug we treat as critical (the advisory is Medium and the PR notes as much). The intent isn't that no-arg constructor side effects are load-bearing, it's that a class which isn't a `Task` shouldn't be constructed here at all, which removes the "construct an arbitrary classpath class" step the report describes.

PR #6327, opened 2026-07-15
Restrict concept reference range criteria SpEL to a safe surface

dkayiwa:
### Summary
Follow-up to CVE-2026-41258 / GHSA-xj4f-8jjg-vx4q. That fix replaced Velocity with SpEL in `ConceptReferenceRangeUtility`, using a `SimpleEvaluationContext` that correctly blocks `T(...)` type references and `new` constructors — so this is **not** RCE. But it binds live patient-graph objects (`$patient`, `$obs`, `$encounter`, …) and, via `DataBindingMethodResolver.forInstanceMethodInvocation()`, permits arbitrary instance-method calls and indexing on them.

A user with `Manage Concepts` can store a criteria like `$patient.uuid.startsWith('a')`, `$patient.uuid[0] == 'a'`, or `$patient.names[0].familyName == 'x'`. Thereafter every clinician with `Get Concepts` who views a matching obs evaluates that expression server-side against **their** patient, and which `referenceRange` is selected leaks one boolean bit. Combined with `startsWith`/`charAt`/`substring`/indexing, that bisects arbitrary string PHI across the whole patient population. I confirmed each of these vectors evaluates on current `master`.

### Fix
Validate the parsed SpEL AST before evaluation (`parseAndValidate` → `validateNode`, run once per unique criteria inside the existing cache):
- reject `ConstructorReference`, `TypeReference`, `BeanReference`, and `Indexer` (indexing, including into strings), and
- reject any `MethodReference` whose name is not in an allow-list.

The allow-list is the `$fn` helper API, a curated set of scalar/value accessors (`getAge`, `getAgeInMonths`, `getGender`, `getAttribute`, `getValueNumeric`, …), and `equals`/`equalsIgnoreCase`. String inspection methods and arbitrary getters are no longer callable, so the oracle collapses to equality comparisons on the values the feature already intends to expose.

### Validation
- All 71 existing `ConceptReferenceRangeUtilityTest` cases still pass, so the documented criteria surface (age, gender, `getAttribute(...).getValue()`, obs values, `$fn.*`, dates) is preserved.
- New `evaluateCriteria_shouldRejectExfiltrationConstructsButAllowValueComparisons` asserts every bisection/exfiltration vector is rejected while the documented comparisons still work.

### For reviewers
The allow-list is derived from the documented/tested criteria surface. Please confirm it covers criteria real deployments have stored — anything relying on other methods will now be rejected (fail-closed). A residual equality oracle on intentionally-exposed fields (gender, attributes) remains by design; if that is also unacceptable, exposing detached value snapshots instead of live entities would be the next step.

Addresses GHSA-cf7m-74mw-c4wv.

### DIFF
--- a/api/src/main/java/org/openmrs/util/ConceptReferenceRangeUtility.java
+++ b/api/src/main/java/org/openmrs/util/ConceptReferenceRangeUtility.java
@@ -43,6 +43,7 @@ import org.springframework.expression.spel.ast.BeanReference;
 import org.springframework.expression.spel.ast.ConstructorReference;
 import org.springframework.expression.spel.ast.Indexer;
 import org.springframework.expression.spel.ast.MethodReference;
+import org.springframework.expression.spel.ast.OperatorMatches;
 import org.springframework.expression.spel.ast.TypeReference;
 import org.springframework.expression.spel.standard.SpelExpression;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
@@ -112,14 +113,15 @@ public class ConceptReferenceRangeUtility {
 	/**
 	 * Rejects expression constructs that let a stored criteria escape the intended "read a few values
 	 * and compare them" surface: constructor calls, type references, bean references, indexing
-	 * (including into strings, e.g. {@code $patient.uuid[0]}) and any method outside
-	 * {@link #ALLOWED_METHODS}.
+	 * (including into strings, e.g. {@code $patient.uuid[0]}), the {@code matches} regex operator (a
+	 * string-inspection primitive equivalent to the blocked String methods, and also a ReDoS vector)
+	 * and any method outside {@link #ALLOWED_METHODS}.
 	 */
 	private static void validateNode(SpelNode node) {
 		if (node instanceof ConstructorReference || node instanceof TypeReference || node instanceof BeanReference
-		        || node instanceof Indexer) {
+		        || node instanceof Indexer || node instanceof OperatorMatches) {
 			throw new IllegalArgumentException(
-			        "Criteria may not use constructors, type references, bean references or indexing");
+			        "Criteria may not use constructors, type references, bean references, indexing or regex matching");
 		}
 		if (node instanceof MethodReference && !ALLOWED_METHODS.contains(((MethodReference) node).getName())) {
 			throw new IllegalArgumentException("Criteria may not call method: " + ((MethodReference) node).getName());
--- a/api/src/main/java/org/openmrs/util/ConceptReferenceRangeUtility.java
+++ b/api/src/main/java/org/openmrs/util/ConceptReferenceRangeUtility.java
@@ -19,6 +19,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.lang3.StringUtils;
@@ -37,6 +38,13 @@ import org.springframework.expression.Expression;
 import org.springframework.expression.ExpressionParser;
 import org.springframework.expression.spel.SpelEvaluationException;
 import org.springframework.expression.spel.SpelMessage;
+import org.springframework.expression.spel.SpelNode;
+import org.springframework.expression.spel.ast.BeanReference;
+import org.springframework.expression.spel.ast.ConstructorReference;
+import org.springframework.expression.spel.ast.Indexer;
+import org.springframework.expression.spel.ast.MethodReference;
+import org.springframework.expression.spel.ast.TypeReference;
+import org.springframework.expression.spel.standard.SpelExpression;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.expression.spel.support.DataBindingMethodResolver;
 import org.springframework.expression.spel.support.DataBindingPropertyAccessor;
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/util/ConceptReferenceRangeUtility.java
```diff
@@ -75,11 +83,52 @@ public class ConceptReferenceRangeUtility {
...
+	}
+
+	/**
+	 * Rejects expression constructs that let a stored criteria escape the intended "read a few values
+	 * and compare them" surface: constructor calls, type references, bean references, indexing
+	 * (including into strings, e.g. {@code $patient.uuid[0]}) and any method outside
+	 * {@link #ALLOWED_METHODS}.
+	 */
+	private static void validateNode(SpelNode node) {
+		if (node instanceof ConstructorReference || node instanceof TypeReference || node instanceof BeanReference
```
ibacher, line 136: So `ConstructorReference`, `TypeReference`, and `BeanReference` should already be excluded by the `SimpleEvaluationContext` itself, which doesn't support these.

ibacher: For the method allow-list, I think a stronger guard here would be to replace the `DataBindingMethodResolver` with a custom version like:

```java
 MethodResolver base = DataBindingMethodResolver.forInstanceMethodInvocation();
 MethodResolver allowListed = (ctx, target, name, argTypes) ->
      ALLOWED_METHODS.contains(name) ? base.resolve(ctx, target, name, argTypes) : null;
```

That leverages SpEL's native mechanism for blocking this. And then we only need to parse the `SpelNode` for `Indexer` instances.

api/src/test/java/org/openmrs/util/ConceptReferenceRangeUtilityTest.java
```diff
@@ -1070,4 +1070,34 @@ private Obs buildObs() {
...
+	public void evaluateCriteria_shouldRejectExfiltrationConstructsButAllowValueComparisons() {
+		Obs obs = buildObs();
+		obs.setPerson(person);
+
+		String[] blocked = { "$patient.uuid.startsWith('a')", "$patient.uuid.charAt(0) == 'a'",
+		        "$patient.uuid.substring(0, 1) == 'a'", "$patient.uuid.length() > 5", "$patient.uuid.compareTo('a') > 0",
+		        "$patient.uuid.getBytes()[0] > 50", "$patient.uuid matches 'a.*'", "$patient.uuid[0] == 'a'",
+		        "$patient.getUuid().startsWith('a')", "$patient.names[0].familyName == 'x'",
+		        "T(java.lang.Runtime).getRuntime()", "new java.net.URL('http://evil')", "$patient.uuid.bytes.length > 5",
+		        "$patient.uuid.Bytes.length > 5", "$patient.uuid.bytes != null", "$patient.uuid.class != null" };
```
dkayiwa, line 1091: The blocked list exercises string inspection called directly on a value, along with indexing and property-syntax getters, but it does not cover the same primitives reached through a selection or projection inner expression, through null-safe navigation, or nested inside a method argument. I confirmed the current code already rejects all of those forms: `$patient.names.?[familyName.startsWith('a')]`, `$patient.names.![familyName.substring(0, 1)]`, `$patient.getGender()?.startsWith('M')`, and a `substring` or constructor placed inside a `$fn.getDaysBetween(...)` argument each throw, because `validateNode` recurses into every child and Spring models those inner expressions and method arguments as children. That recursion is the least obvious part of the guarantee, and a gap there would let an expression slip through silently rather than throw, so nothing in CI would flag the regression. It would be worth adding one selection or projection case and one null-safe case to `blocked` so that behavior is pinned against any future rework of `validateNode`.

api/src/main/java/org/openmrs/util/ConceptReferenceRangeUtility.java
dkayiwa: Good point on the first one. I confirmed the context rejects `T(...)`, `new`, and `@bean` on its own (EL1005E, EL1002E, EL1057E), so those three node checks are really belt and suspenders. I have been inclined to keep them so the rejection happens at parse time with a criteria-specific message, and so the guard does not silently depend on this staying a `SimpleEvaluationContext` (a later swap to a `StandardEvaluationContext` would re-open `T(...)`/`new`), but I am happy to drop them if you would rather lean on the context alone.

On the method allow-list, your resolver does work: I tried it and a non-listed method like `startsWith` fails with EL1004E while a listed one still resolves. The reason the check is in the AST walk is that, since your review, I added two things that are not method calls, so neither a `MethodResolver` nor the context catches them: the `matches` operator (593d5b1) and property-syntax getters like `.bytes`, `.class`, and `.length` (e38d6f4). Even with the allow-list resolver in place I still get `#root matches 'a.*'` returning true and `bytes.length` returning 6, so the walk has to stay for `Indexer`, `OperatorMatches`, and the denied properties regardless. Since it was already there for those, folding the method-name check into the same single pass felt simpler than running a separate resolver alongside it. If you would still prefer the resolver for the method part, it is an easy swap and I will make it.

ibacher: That's reasonable. My thought here is that Spring will already run _a_ MethodResolver, so it's not like skipping that implementation makes this any more efficient and that if we remove the items earlier there's less invalid AST to walk, but it's true that we can't run all these guards without at least some AST walker.

I'm not entirely clear why we want to shutoff the `matches` operator? That actually seems like a reasonable thing to allow if we want to extend this functionality to non-numeric obs.

api/src/test/java/org/openmrs/util/ConceptReferenceRangeUtilityTest.java
ibacher: For the most part, I think this is actually maybe a little over-paranoid to the point where we're breaking useful functionality. There's the marginal fact that this rule could allow an administrator without permission to read a patient record to infer facts about the patient record, but to actually observe the results, they would already need to have the permission to read the resulting obs... and the patient. So it's kind of guarding us against an impossible situation.

ibacher: And, I think we should want to extend this functionality to support categorical and maybe even text data so that we can properly mark Obs with things like "BDL"...

api/src/main/java/org/openmrs/util/ConceptReferenceRangeUtility.java
dkayiwa: Agreed on the walker. On `matches` I need to walk back what I said: I called it a ReDoS vector, but Spring 7.0.8 already caps it. `OperatorMatches` rejects any regex over 1000 characters and has a pattern-access valve that aborts catastrophic backtracking, so the library covers that and my reason does not hold. No objection to allowing it, which also lets criteria match non-numeric obs the way you want. I will fold it into the broader relaxation I describe on the test thread.

api/src/test/java/org/openmrs/util/ConceptReferenceRangeUtilityTest.java
dkayiwa: You are right, and checking the access model I think I was guarding a case core mostly rules out. Obs reads are a single `GET_OBS` privilege with no per-concept granularity, and the criteria also runs at save time through `ObsValidator`, so whoever can see which range applied can generally already read the obs and the patient. The one residual I can construct is narrow: `$patient` is bound from the in-memory person graph and read through property accessors, bypassing service-layer authorization, so a principal with `GET_OBS` but without patient read could infer `$patient.*` demographics. That is an unusual split and not worth breaking legitimate criteria over.

The other thing I confirmed is that the RCE surface never depended on this allow-list. `SimpleEvaluationContext` blocks `new`, `T(...)`, and `@bean`, and the DataBinding accessors filter anything declared on `Object` or `Class` and refuse to resolve against a `Class` target, so `getClass`, `.class`, and the classloader are already unreachable (I checked, they throw on their own). So `validateNode` is really only constraining the exfiltration oracle, and its `class`, constructor, type, and bean entries are redundant with the context.

Given all that, I will relax it so the string and value methods and `matches` are permitted, which gives us the categorical and text criteria you want (equality already covers an exact 'BDL', and this adds prefix and pattern checks on top), and lean on the context plus the DataBinding accessors for RCE, which is where that protection actually lives. That also makes my earlier ask about pinning the selection and projection cases moot, since those become allowed.

Do you want the categorical and text support folded into this PR, or should I keep this one scoped to the relaxation and do the feature separately?

PR #6328, opened 2026-07-15
getMaximumPropertyLength should degrade gracefully instead of throwing

dkayiwa:
### Problem
`HibernateAdministrationDAO.getMaximumPropertyLength(aClass, fieldName)` throws an `APIException` when `metadata.getEntityBinding(aClass.getName())` returns null. That `metadata` is a boot-time Hibernate `Metadata` snapshot captured once in `setApplicationContext`, so if it is ever consulted in an incomplete state the method throws hard.

This method is called by `validate()` on every `OpenmrsObject` save. Two things show the hard throw is an inconsistent outlier:
- `validate()` already guards every use with `if (maxLength >= 0 && ...)`, i.e. a negative return means "length unknown, skip the check".
- The method itself already returns `-1` when it cannot determine a column's length (the `catch` block).

So a transient inability to resolve an entity should behave the same way, `-1`, not throw and fail the save.

### How it surfaced
Intermittent CI failure: `HibernateAdministrationDAOTest.getMaximumPropertyLength_*` errored with `Couldn't find a class in the hibernate configuration named: org.openmrs.event.outbox.OutboxEvent` in some full-suite orderings, while passing in isolation. The throw is the sharp edge; the underlying "the boot Metadata snapshot sometimes can't resolve an annotation-scanned entity" is what varies by test order.

### Change
Return `-1` (unknown length) instead of throwing when the entity binding cannot be resolved, consistent with the existing sentinel and with `validate()`'s `maxLength >= 0` guard. Adds a regression test asserting an unmapped class yields `-1` rather than an exception (it errors on the pre-change code with the same `APIException`, and passes after).

### Scope / what this does and doesn't do
This hardens the production `validate()` path so a transient metadata gap can no longer fail a save, and removes the exception that was erroring the test. It does not, on its own, make assertions that require a specific entity's column lengths deterministic; a test that asserts exact lengths still needs that entity registered. Verified: the class passes (5/5) with the change and spotless is clean; the new test errors on the pre-change code and passes after.

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
@@ -225,10 +225,8 @@ public class HibernateAdministrationDAO implements AdministrationDAO, Applicatio
 	public long getMaximumPropertyLength(Class<? extends OpenmrsObject> aClass, String fieldName) {
 		PersistentClass persistentClass = metadata.getEntityBinding(aClass.getName().split("_")[0]);
 		if (persistentClass == null) {
-			// The boot Metadata is captured once at startup, so it can fail to resolve an entity if it is
-			// ever consulted in an incomplete state (for example after a session factory rebuild). Treat
-			// the length as unknown (-1) instead of throwing: this method is called from validate() on
-			// every save, and validate() already skips the length check when it returns a negative value.
+			// Treat the length as unknown (-1) instead of throwing: this method is called from validate()
+			// on every save, and validate() skips the length check when the value is negative.
 			log.debug("Could not find a hibernate entity binding for {}; treating maximum length as unknown",
 			    aClass.getName());
 			return -1;
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
@@ -36,7 +36,6 @@ import org.hibernate.type.BasicType;
 import org.hibernate.type.Type;
 import org.openmrs.GlobalProperty;
 import org.openmrs.OpenmrsObject;
-import org.openmrs.api.APIException;
 import org.openmrs.api.db.AdministrationDAO;
 import org.openmrs.api.db.DAOException;
 import org.openmrs.util.DatabaseUtil;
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -226,29 +225,34 @@ public List<List<Object>> executeSQL(String sql, boolean selectOnly) throws DAOE
...
-			} catch (Exception e) {
-				log.debug("Could not determine maximum length", e);
-				return -1;
+			// The boot Metadata is captured once at startup, so it can fail to resolve an entity if it is
+			// ever consulted in an incomplete state (for example after a session factory rebuild). Treat
+			// the length as unknown (-1) instead of throwing: this method is called from validate() on
+			// every save, and validate() already skips the length check when it returns a negative value.
+			log.debug("Could not find a hibernate entity binding for {}; treating maximum length as unknown",
+			    aClass.getName());
+			return -1;
```
dkayiwa, line 232: Since this changes what `AdministrationService#getMaximumPropertyLength` observably does for a class with no entity binding (it used to throw `APIException`, now it returns -1), it would be worth updating the `@return` javadoc on the service interface to document the sentinel, something like `@return the max field length of a property, or -1 if it cannot be determined`. Both in-core consumers already key off it (`validate()` checks `maxLength >= 0` and `ValidateUtil.validateFieldLengths` checks `length == -1`), but a module author calling the service has no way to discover it without reading this DAO.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -226,29 +225,34 @@ public List<List<Object>> executeSQL(String sql, boolean selectOnly) throws DAOE
...
-				} else {
-					fieldLength = column.getLength();
-				}
-			} catch (Exception e) {
-				log.debug("Could not determine maximum length", e);
-				return -1;
+			// The boot Metadata is captured once at startup, so it can fail to resolve an entity if it is
+			// ever consulted in an incomplete state (for example after a session factory rebuild). Treat
+			// the length as unknown (-1) instead of throwing: this method is called from validate() on
+			// every save, and validate() already skips the length check when it returns a negative value.
```
ibacher: Isn't this comment somewhat inaccurate? The metadata should be captured at session start time, but that's not the same as "boot" time (e.g., when a module refresh happens. I also don't think it adds much value that can't be gleaned from the log statement.

dkayiwa: You're right that "boot" time is wrong. The snapshot is re-captured whenever this DAO's context is initialized, so a module refresh gives it a fresh one rather than it being a one-time startup capture. I've dropped that framing.

On the value point, I trimmed the comment to the one thing the log line doesn't say, which is why -1 is the right fallback rather than an error: validate() enforces the limit only when the value is non-negative, so returning -1 cleanly skips the check. If you'd still rather have nothing beyond the log statement there, I'm happy to remove it entirely.

PR #6332, opened 2026-07-15
Require Add Patient Identifiers privilege when savePatient adds an identifier

dkayiwa:
### Summary
`PatientService.savePatient` is guarded by `ADD_PATIENTS`/`EDIT_PATIENTS`, and `savePatientIdentifier` separately by `ADD_PATIENT_IDENTIFIERS`/`EDIT_PATIENT_IDENTIFIERS`. But `savePatient` persists the patient's identifiers by Hibernate cascade without applying the identifier privilege, so a user with `Edit Patients` who is deliberately denied `Add Patient Identifiers` can still inject an identifier (e.g. an MRN) through the aggregate save. This breaks the intended separation of duties.

### Change
Before saving, require `ADD_PATIENT_IDENTIFIERS` if the patient carries any not-yet-persisted (transient) identifier. Existing identifiers are not inspected, so editing a patient's demographics still requires only patient privileges — the check keys on `getPatientIdentifierId() == null`, which is exactly a newly added identifier.

### Tests
`savePatient_shouldRequireAddPatientIdentifiersPrivilegeToAddAnIdentifier`: a user with `Edit Patients` (plus the reads validation needs) but not `Add Patient Identifiers` is refused when adding an identifier via `savePatient`. The patient is evicted from the session first so the new identifier stays transient, mirroring a REST-deserialized patient. Full `PatientServiceTest` (182 tests) still passes.

### Scope / follow-up
This closes the add-an-identifier bypass. Modifying or voiding an *existing* identifier through `savePatient` (which would warrant `EDIT`/`DELETE_PATIENT_IDENTIFIERS`) is not covered here: distinguishing a genuine change from `savePatient`'s own internal `preferred`-flag updates needs field-level dirty tracking, which is best done as a separate change. Noting it so it isn't assumed covered.

Addresses GHSA-cw94-x75j-5vwx.

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -168,6 +168,15 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 	 * enforces. Any not-yet-persisted identifier in the graph therefore requires the Add Patient
 	 * Identifiers privilege. Existing identifiers are left untouched so that editing a patient's
 	 * demographics does not require identifier privileges.
+	 * <p>
+	 * This check relies on a newly added identifier still being transient (its id {@code null}) when it
+	 * runs. Requests served through the REST and web layers satisfy that: the session is opened with
+	 * {@code FlushMode.MANUAL} and no read-write transaction wraps the request, so no query flushes the
+	 * pending identifier beforehand, and the identifier-uniqueness validation runs in an advice ordered
+	 * outside this method's transaction. A caller that instead invokes savePatient from within its own
+	 * read-write transaction, on a session-managed patient, could have the new identifier auto-flushed
+	 * and assigned an id before this point, which would skip the check; closing that residual for such
+	 * callers would require enforcing the privilege at the persistence layer rather than here.
 	 */
 	private void requireAppropriatePatientIdentifierPrivilege(Patient patient) {
 		for (PatientIdentifier identifier : patient.getIdentifiers()) {
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -163,20 +163,8 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 	}
 
 	/**
-	 * Adding a patient identifier through the aggregate {@link #savePatient(Patient)} path must not
-	 * bypass the identifier-specific privilege that {@link #savePatientIdentifier(PatientIdentifier)}
-	 * enforces. Any not-yet-persisted identifier in the graph therefore requires the Add Patient
-	 * Identifiers privilege. Existing identifiers are left untouched so that editing a patient's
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -161,6 +162,31 @@ private void requireAppropriatePatientModificationPrivilege(Patient patient) {
...
+	 * pending identifier beforehand, and the identifier-uniqueness validation runs in an advice ordered
+	 * outside this method's transaction. A caller that instead invokes savePatient from within its own
+	 * read-write transaction, on a session-managed patient, could have the new identifier auto-flushed
+	 * and assigned an id before this point, which would skip the check; closing that residual for such
+	 * callers would require enforcing the privilege at the persistence layer rather than here.
+	 */
+	private void requireAppropriatePatientIdentifierPrivilege(Patient patient) {
+		for (PatientIdentifier identifier : patient.getIdentifiers()) {
+			if (identifier.getPatientIdentifierId() == null) {
+				Context.requirePrivilege(PrivilegeConstants.ADD_PATIENT_IDENTIFIERS);
```
dkayiwa, line 172: This fires for any transient identifier, so it also covers a brand new patient's first identifier: a caller with Add Patients but not Add Patient Identifiers can no longer register a patient through savePatient. On master, Add Patients alone was enough, since the cascade persisted the identifier without the identifier privilege, which is the same gap this PR closes on the edit path. Was gating registration intended as well? The summary and the scope note frame the change around the Edit Patients case, so it would help to say explicitly whether new patient registration now requires Add Patient Identifiers, since it changes which roles can register patients and is relevant to any backport.

api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
```diff
@@ -161,6 +162,31 @@ private void requireAppropriatePatientModificationPrivilege(Patient patient) {
...
+	 * <p>
+	 * This check relies on a newly added identifier still being transient (its id {@code null}) when it
+	 * runs. Requests served through the REST and web layers satisfy that: the session is opened with
+	 * {@code FlushMode.MANUAL} and no read-write transaction wraps the request, so no query flushes the
+	 * pending identifier beforehand, and the identifier-uniqueness validation runs in an advice ordered
+	 * outside this method's transaction. A caller that instead invokes savePatient from within its own
+	 * read-write transaction, on a session-managed patient, could have the new identifier auto-flushed
+	 * and assigned an id before this point, which would skip the check; closing that residual for such
+	 * callers would require enforcing the privilege at the persistence layer rather than here.
+	 */
```
ibacher, line 168: I feel like the Javadoc here is an unreadable mess of Claude-over-explanation. All this _really_ needs to say is:

"Ensures that the user has appropriate privileges to add a patient identifier if they try to do so through savePatient()". Pretty much everything else here is way too liable to become out of date and not relevant to understanding _how_ and _when_ to call this method (which is what doc strings are for).

dkayiwa: Done, trimmed it to just that. Thanks Ian.

PR #6152, opened 2026-06-02
TRUNK-6645: Prevent clearPasswords() from wiping wizard credentials mid-flow

solomonfortune:
## Description of what I changed

Added a `passwordsEntered` flag to `InitializationWizardModel` that is set to `true` when the user submits credentials in either the simple or advanced setup path, and reset to `false` inside `clearPasswords()`. 
The `page == null` branch in `doGet()` now only calls `clearPasswords()` when the flag is `false`, i.e. before any credentials have been entered.

Previously, any parameterless GET to `/initialsetup` (caused by a favicon request, browser refresh, Back button, or StartupFilter redirect) would call `clearPasswords()` unconditionally on the shared singleton `wizardModel`, blanking all five password fields even after the user had already entered them. 
Since `SessionModelUtils` deliberately excludes password fields from the session round-trip, the wiped values could not be restored, causing the background install thread to connect with an empty password and fail with: `Access denied for user 'root'@'localhost' (using password: NO)`

Files changed:
- `InitializationWizardModel.java` — added `passwordsEntered` flag
- `InitializationFilter.java` — guard `clearPasswords()`, set flag on credential submission, reset flag inside `clearPasswords()`

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-6645

## Checklist: I completed these to help reviewers :)
- [ ] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.
- [ ] I have **added tests** to cover my changes. (If you refactored existing code that was well tested you do not have to add tests)
- [x] I ran `mvn clean package` right before creating this pull request and added all formatting changes to my commit.
- [ ] All new and existing **tests passed**.
- [x] My pull request is **based on the latest changes** of the master branch.

PR #6333, opened 2026-07-15
Keep the setup wizard closed when a configured database is unreachable

dkayiwa:
### Summary
`DatabaseDetective.isDatabaseEmpty(Properties)` treated any exception while connecting as "the database is empty" (`catch (Exception e) { return true; }`). Both callers, `Listener` at startup and `InitializationFilter`, use that to decide whether to show the unauthenticated setup wizard. So on an already-configured deployment, a database-connectivity failure at startup (a DB outage or restart) made the instance report itself empty and **reopen the unauthenticated setup wizard**, which renders the stored DB credentials and lets an unauthenticated user reconfigure the instance and persist their own settings back to `openmrs-runtime.properties`.

### Change
On a connection failure, only treat the instance as empty when it is genuinely unconfigured (no `connection.url`). A configured instance that cannot be reached is reported as not empty, so startup fails with a connection error instead of reopening setup. Unconfigured/fresh installs (null or empty properties) still return empty, so the normal first-run wizard is unaffected. Also removed the `return true` inside the `finally` block's `close()` handler, which could otherwise override a non-empty result.

### Tests
`DatabaseDetectiveTest.shouldNotReportConfiguredButUnreachableDatabaseAsEmpty`: a configured instance whose database is unreachable is not reported as empty. The existing null-properties and empty-properties cases (still empty) continue to pass.

Addresses GHSA-jqhq-p753-2hwj. This is the reopen root cause; it is distinct from #6152, which fixes credential wiping mid-flow (TRUNK-6645).

### DIFF
--- a/web/src/main/java/org/openmrs/web/filter/initialization/DatabaseDetective.java
+++ b/web/src/main/java/org/openmrs/web/filter/initialization/DatabaseDetective.java
@@ -73,7 +73,7 @@ public class DatabaseDetective {
 			// already-configured deployment during a database outage, so only a genuinely
 			// unconfigured instance (no connection URL) counts as empty.
 			String url = props.getProperty(CONNECTION_URL);
-			return url == null || url.trim().isEmpty();
+			return url == null || url.isBlank();
 		} finally {
 			if (connection != null) {
 				try {
--- a/web/src/main/java/org/openmrs/web/filter/initialization/DatabaseDetective.java
+++ b/web/src/main/java/org/openmrs/web/filter/initialization/DatabaseDetective.java
@@ -68,16 +68,19 @@ public class DatabaseDetective {
 			}
 			return true;
 		} catch (Exception e) {
-			// consider the database to be empty
-			return true;
+			// A configured instance (connection URL set) that we cannot reach is unreachable, not
+			// empty. Reporting it empty here would reopen the unauthenticated setup wizard on an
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/openmrs/web/filter/initialization/DatabaseDetective.java
```diff
@@ -68,16 +68,19 @@ public boolean isDatabaseEmpty(Properties props) {
...
 			return true;
 		} catch (Exception e) {
-			// consider the database to be empty
-			return true;
+			// A configured instance (connection URL set) that we cannot reach is unreachable, not
+			// empty. Reporting it empty here would reopen the unauthenticated setup wizard on an
+			// already-configured deployment during a database outage, so only a genuinely
+			// unconfigured instance (no connection URL) counts as empty.
+			String url = props.getProperty(CONNECTION_URL);
+			return url == null || url.trim().isEmpty();
```
ibacher: `isBlank()` exists and is better than `.trim().isEmpty()` for this purpose since it doesn't need to allocate a new string.

ibacher (approved):

web/src/main/java/org/openmrs/web/filter/initialization/DatabaseDetective.java
dkayiwa: Good point! Done in c9bbc6b05.

PR #6335, opened 2026-07-15
Bind caller-supplied names as query parameters instead of concatenating

dkayiwa:
### Summary
An audit for SQL string concatenation turned up two native/JDBC queries that concatenated a caller-supplied string directly into the SQL text instead of binding it. Every other value in these queries is already bound or type-safe, so these were the two places a caller-supplied string reached the SQL as literal text.

### Change
- `HibernateProgramWorkflowDAO.getPatientProgramAttributeByAttributeName`: bind `attributeName` as a named parameter (`:attributeName`) so it is always handled as a literal attribute-type name. The patient-id list stays inlined since those elements are `Integer`s.
- `FilterUtil.getGlobalPropertyValue(String)`: replace the concatenated `global_property` lookup with a `PreparedStatement` that binds the property name. This runs before the Spring context is up (setup/upgrade wizard) via a raw JDBC connection, so it does not go through the usual service layer. Not exploitable today since the only caller passes a hardcoded constant, but the signature accepts an arbitrary string.

There is no behaviour change for legitimate attribute or property names.

### Tests
- `ProgramWorkflowServiceTest#getPatientProgramAttributeByAttributeName_shouldMatchTheAttributeNameLiterally`: stores a `ProgramId` attribute on an existing enrollment and checks that a real name resolves the stored value, while a name containing quote characters is compared literally and resolves no attribute type.
- `FilterUtilTest#getGlobalPropertyValue_shouldMatchThePropertyNameLiterally`: a real property name resolves its stored value, while a name carrying SQL control characters is matched literally and returns null. The payload was chosen so it resolved a single row (returning a value) on the old concatenated query, so the test fails on the pre-change code and passes on the fix.

`ProgramWorkflowServiceTest` and `FilterUtilTest` pass and spotless is clean.

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateProgramWorkflowDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateProgramWorkflowDAO.java
@@ -475,12 +475,12 @@ public class HibernateProgramWorkflowDAO implements ProgramWorkflowDAO {
 		        + " from patient p " + " join patient_program pp on p.patient_id = pp.patient_id and p.patient_id in ("
 		        + commaSeperatedPatientIds + ")"
 		        + " join patient_program_attribute ppa on pp.patient_program_id = ppa.patient_program_id and ppa.voided=0"
-		        + " join program_attribute_type ppt on ppa.attribute_type_id = ppt.program_attribute_type_id and ppt.name ='"
-		        + attributeName + "' "
+		        + " join program_attribute_type ppt on ppa.attribute_type_id = ppt.program_attribute_type_id and ppt.name = :attributeName "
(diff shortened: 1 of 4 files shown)


dkayiwa:
@claude review

ibacher (approved):

ibacher:
Have we run an audit for any other similar string concatentations? I recall seeing a few.

dkayiwa:
Yeah, I went through api and web looking for concatenated strings going into createNativeQuery/createSQLQuery/createQuery. This DAO method was the only place where a caller-supplied string actually ended up in the SQL as a literal.

The other concatenation spots are fine: some inline integers only, HibernatePatientDAO.getDuplicatePatientsByAttributes whitelists the column names against the entity fields, a few use hardcoded table/column names, and the rest either concatenate parameter names or just toggle clauses on booleans while keeping the values bound.

The one exception was FilterUtil.getGlobalPropertyValue(String), which built raw JDBC SQL by concatenating the property name. It wasn't exploitable since the only caller passes a hardcoded constant, but the signature takes any string and it runs before the app context is up, so I've folded a fix into this PR that binds it via a PreparedStatement, plus a regression test.

PR #6341, opened 2026-07-16
TRUNK-6688: Fix circular initialization error in logging setup

ibacher:

This fixes an issue where, during initialization of the `ServiceContext` if a logger is configured, it would attempt to read a global property (for the log settings) which, if the `ServiceContext` was not fully configured before this happened, leads to a runtime NPE. This only triggers on Java 25, but it's a latent issue on earlier JVM versions.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6688

  No? -> execute above command

### DIFF
--- a/api/src/test/java/org/openmrs/logging/OpenmrsConfigurationFactoryTest.java
+++ b/api/src/test/java/org/openmrs/logging/OpenmrsConfigurationFactoryTest.java
@@ -310,7 +310,7 @@ class OpenmrsConfigurationFactoryTest {
 	 * Log4j2 configuration is triggered from within {@code ServiceContext}'s static initializer, so a
 	 * session appears open but the {@code ServiceContext} singleton has not been created yet. In that
 	 * state {@code doOpenmrsCustomisations} must not try to read global properties, which triggers a
-	 * message logged using the {@code ServiceContext}'s logger.'
+	 * message logged using the {@code ServiceContext}'s logger.
 	 */
 	@Test
 	void doOpenmrsCustomisations_shouldNotForceServiceContextInitializationWhenNotYetAvailable() {
--- a/api/src/test/java/org/openmrs/logging/OpenmrsConfigurationFactoryTest.java
+++ b/api/src/test/java/org/openmrs/logging/OpenmrsConfigurationFactoryTest.java
@@ -31,6 +31,7 @@ import org.mockito.InOrder;
 import org.mockito.MockedStatic;
 import org.mockito.Mockito;
 import org.openmrs.api.context.Context;
+import org.openmrs.api.context.ServiceContext;
 import org.openmrs.util.ConfigUtil;
 import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.OpenmrsUtil;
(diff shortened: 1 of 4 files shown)


api/src/test/java/org/openmrs/logging/OpenmrsConfigurationFactoryTest.java
```diff
@@ -304,6 +305,65 @@ void doOpenmrsCustomisations_shouldSkipGlobalPropertyWhenSessionClosed() {
 		}
 	}
 
+	/**
+	 * Regression test for TRUNK-6688. Reproduces the conditions of the re-entrant initialization path:
+	 * Log4j2 configuration is triggered from within {@code ServiceContext}'s static initializer, so a
+	 * session appears open but the {@code ServiceContext} singleton has not been created yet. In that
+	 * state {@code doOpenmrsCustomisations} must not try to read global properties, which triggers a
+	 * message logged using the {@code ServiceContext}'s logger.
```
dkayiwa, line 313: There's a stray quote at the end of this sentence, and the "which triggers a message" clause reads inside out to me: the point is that the GP read forces `ServiceContext` creation, whose constructor then logs through the still-unassigned static `log` (the NPE from the ticket). Maybe:

```suggestion
	 * state {@code doOpenmrsCustomisations} must not try to read global properties, since that would force
	 * {@code ServiceContext} creation, whose constructor logs through the still-unassigned static logger.
```

Take or leave the rewording, the stray quote is the main thing.

api/src/test/java/org/openmrs/logging/OpenmrsConfigurationFactoryTest.java
```diff
@@ -304,6 +305,65 @@ void doOpenmrsCustomisations_shouldSkipGlobalPropertyWhenSessionClosed() {
 		}
 	}
 
+	/**
+	 * Regression test for TRUNK-6688. Reproduces the conditions of the re-entrant initialization path:
```
dkayiwa, line 309: Was it a deliberate call to cover this with mocked statics rather than the real re-entrant path? The ticket's acceptance criteria ask for a test that triggers Log4j2 config from inside `ServiceContext.<clinit>` and asserts it initializes cleanly. This test pins the guard well (I checked it fails if the `isInstantiated()` check is reverted), but the clinit re-entrancy itself stays unexercised, so a future cycle through a different route in the config path would slip past it. If a forked-JVM repro was considered and dropped as not worth the plumbing, works for me; asking mainly so the tradeoff is on the record.

api/src/test/java/org/openmrs/logging/OpenmrsConfigurationFactoryTest.java
ibacher: Yeah, caught by Copilot above and already fixed 😁

api/src/test/java/org/openmrs/logging/OpenmrsConfigurationFactoryTest.java
ibacher: Yeah, considered and dropped. The options here are really heavy-weight, i.e., forking the JVM just to run one test or creating a ClassLoader headache, so I opted for the a test that will at least flag things if the code changes too much.

PR #6301, opened 2026-07-13
Trunk-6663: Implement argon 2 id password encoder with configurable work factors

EDSONZ-WASSWA:
## Summary

This PR implements **Argon2id as the default password encoder** for OpenMRS while preserving backward compatibility with existing password hashes.

Argon2 work factors (memory, parallelism, iterations, hash length, and salt length) are configurable through Global Properties with documented defaults. The encoder automatically refreshes when configuration changes, allowing updated settings to take effect without requiring a server restart.

The implementation validates all configurable work factor values, ensuring they are positive and falling back to safe defaults with warning logs when invalid values are configured. Existing SHA-512 and SHA-1 password hashes continue to verify correctly, while newly encoded passwords use Argon2id.

To preserve existing deterministic hashing behaviour outside password storage, a dedicated `Security.encodeStringSHA512()` helper was introduced. Existing consumers such as activation keys and secret-answer hashing continue using SHA-512, preventing regressions in features that rely on deterministic hashes while allowing password hashing to transition to Argon2id.

The configuration-loading logic was also updated to safely fall back to the default Argon2 parameters when the application service layer is not yet available (for example during startup or database upgrades), while automatically using Global Property values once the service layer has been initialized.

## Changes

- Added the required `spring-security-crypto` and `argon2-jvm` dependencies.
- Implemented Argon2id password encoding in `Security.encodeString()`.
- Added configurable Argon2 Global Properties and their Javadocs.
- Updated `Security.hashMatches()` to verify Argon2 hashes while maintaining compatibility with legacy hashes.
- Introduced `Security.encodeStringSHA512()` for deterministic hashing use cases.
- Added validation for configurable Argon2 work factors, with safe fallback to documented defaults.
- Added automatic encoder refresh when Argon2 configuration changes.
- Updated deterministic hashing consumers (activation keys and secret-answer hashing) to continue using SHA-512.
- Made configuration loading resilient during application startup and database upgrade scenarios.
- Added unit tests covering Argon2 encoding, configuration updates, and invalid configuration values.

## Backward Compatibility

- Existing SHA-512 and SHA-1 password hashes continue to verify correctly through the existing password verification mechanisms.
- New password hashes are encoded using Argon2id.
- Existing deterministic hashing consumers (such as activation keys and secret answers) continue using SHA-512 to preserve their current behaviour.
- Invalid or unavailable Argon2 configuration safely falls back to documented defaults.
## Issue I worked on

[...]

PR #6219, opened 2026-06-22
TRUNK-6662: Get unchanged password hashing working using Spring's interfaces

jwnasambu:
## Description of what I changed

I integrated Spring Security's `PasswordEncoder` interfaces to manage password hashing within the authentication flow and ensure that unchanged password validation and hashing mechanisms remain functional using Spring's native interfaces.

## Issue I worked on

https://openmrs.atlassian.net/browse/TRUNK-6662

## Checklist: I completed these to help reviewers :)

- [ ] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

  No? Unsure? -> [configure your IDE](https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE), format the code and add the changes with `git add . && git commit --amend`

- [ ] I have **added tests** to cover my changes. (If you refactored
  existing code that was well tested you do not have to add tests)

  No? -> write tests and add them to this commit `git add . && git commit --amend`

- [ ] I ran `mvn clean package` right before creating this pull request and
  added all formatting changes to my commit.

  No? -> execute above command

- [ ] All new and existing **tests passed**.

  No? -> figure out why and add the fix to your commit. It is your responsibility to make sure your code works.

- [ ] My pull request is **based on the latest changes** of the master branch.

  No? Unsure? -> execute command `git pull --rebase upstream master`

PR #6345, opened 2026-07-17
TRUNK-6693: Upgrade secrets to modern password encoding

jwnasambu:

I extended the new Argon2id password encoding mechanism to support secret answers, which are human-controlled, low-entropy secrets that benefit from the same secure hashing approach as passwords. Activation keys remain on SHA-512 as they are short-lived and randomly generated tokens requiring deterministic lookup.

## Issue I worked on

https://openmrs.atlassian.net/browse/TRUNK-6693

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
@@ -152,7 +152,7 @@ public class HibernateUserDAO implements UserDAO {
 	 */
 	@Override
 	public LoginCredential getLoginCredentialByActivationKey(String activationKey) {
-		String key = Security.encodeString(activationKey);
+		String key = Security.encodeStringSHA512(activationKey);
 		Session session = sessionFactory.getCurrentSession();
 		CriteriaBuilder cb = session.getCriteriaBuilder();
 		CriteriaQuery<LoginCredential> cq = cb.createQuery(LoginCredential.class);
@@ -477,15 +477,26 @@ public class HibernateUserDAO implements UserDAO {
 	 */
 	@Override
 	public boolean isSecretAnswer(User u, String answer) throws DAOException {
-		
+
 		if (StringUtils.isEmpty(answer)) {
 			return false;
 		}
-		
+
 		LoginCredential credentials = getLoginCredential(u);
 		String answerOnRecord = credentials.getSecretAnswer();
-		String hashedAnswer = Security.encodeString(answer.toLowerCase() + credentials.getSalt());
-		return (hashedAnswer.equals(answerOnRecord));
+		if (StringUtils.isEmpty(answerOnRecord)) {
+			return false;
+		}
+
+		boolean matches = Security.hashMatches(answerOnRecord, answer.toLowerCase() + credentials.getSalt());
+
+		if (matches && !answerOnRecord.startsWith("$argon2id$")) {
+			String rehashedAnswer = Security.encodeString(answer.toLowerCase() + credentials.getSalt());
+			credentials.setSecretAnswer(rehashedAnswer);
+			updateLoginCredential(credentials);
+		}
+
+		return matches;
 	}
 	
 	/**
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
@@ -103,7 +103,7 @@ public class HibernateUserDAO implements UserDAO {
 			
 			//update the new user with the password
 			String salt = Security.getRandomToken();
-			String hashedPassword = Security.encodeString(password + salt);
+			String hashedPassword = Security.encodeCredential(password + salt);
 			
 			updateUserPassword(hashedPassword, salt, Context.getAuthenticatedUser().getUserId(), new Date(), user
 			        .getUserId());
@@ -152,7 +152,7 @@ public class HibernateUserDAO implements UserDAO {
 	 */
 	@Override
 	public LoginCredential getLoginCredentialByActivationKey(String activationKey) {
-		String key = Security.encodeStringSHA512(activationKey);
+		String key = Security.encodeString(activationKey);
 		Session session = sessionFactory.getCurrentSession();
 		CriteriaBuilder cb = session.getCriteriaBuilder();
 		CriteriaQuery<LoginCredential> cq = cb.createQuery(LoginCredential.class);
@@ -350,7 +350,7 @@ public class HibernateUserDAO implements UserDAO {
 		if (StringUtils.isBlank(salt)) {
 			salt = Security.getRandomToken();
 		}
-		String newHashedPassword = Security.encodeString(pw + salt);
+		String newHashedPassword = Security.encodeCredential(pw + salt);
 		
 		updateUserPassword(newHashedPassword, salt, authUser.getUserId(), new Date(), u.getUserId());
 	}
@@ -434,7 +434,7 @@ public class HibernateUserDAO implements UserDAO {
 		
 		// update the user with the new password
 		String salt = credentials.getSalt();
-		String newHashedPassword = Security.encodeString(newPassword + salt);
+		String newHashedPassword = Security.encodeCredential(newPassword + salt);
 		updateUserPassword(newHashedPassword, salt, u.getUserId(), new Date(), u.getUserId());
 	}
 	
@@ -464,7 +464,7 @@ public class HibernateUserDAO implements UserDAO {
 		
 		LoginCredential credentials = getLoginCredential(u);
 		credentials.setSecretQuestion(question);
-		String hashedAnswer = Security.encodeString(answer.toLowerCase() + credentials.getSalt());
+		String hashedAnswer = Security.encodeCredential(answer.toLowerCase() + credentials.getSalt());
 		credentials.setSecretAnswer(hashedAnswer);
 		credentials.setDateChanged(new Date());
 		credentials.setChangedBy(u);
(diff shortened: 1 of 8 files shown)


api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -434,6 +434,31 @@ public static final Collection<String> AUTO_ROLES() {
...
+
+	/**
+	 * Global property name for Argon2 hash length in bytes
+	 */
+	public static final String GP_ARGON2_HASH_LENGTH = "security.argon2.hashLength";
+
+	/**
+	 * Global property name for Argon2 salt length in bytes
+	 */
+	public static final String GP_ARGON2_SALT_LENGTH = "security.argon2.saltLength";
```
ibacher: We shouldn't make these global properties, I think. Rather, we should prefer to get these from the runtime properties (which is where we traditionally store the more security-sensitive stuff).

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -737,7 +737,7 @@ public User getUserByActivationKey(String activationKey) {
 	public User setUserActivationKey(User user) throws MessageException {
 		String token = RandomStringUtils.randomAlphanumeric(20);
 		long time = System.currentTimeMillis() + getValidTime();
-		String hashedKey = Security.encodeString(token);
+		String hashedKey = Security.encodeStringSHA512(token);
```
ibacher: ???

api/src/main/java/org/openmrs/util/databasechange/EncryptSecretAnswersChangeSet.java
```diff
@@ -51,7 +51,7 @@ public void execute(Database database) throws CustomChangeException {
 			while (rs.next()) {
 				String answer = rs.getString("secret_answer");
 				String salt = rs.getString("salt");
-				String encryptedAnswer = Security.encodeString(answer.toLowerCase() + salt);
+				String encryptedAnswer = Security.encodeStringSHA512(answer.toLowerCase() + salt);
```
ibacher: Also this?

jwnasambu:
@claude review

jwnasambu:
@claude review

jwnasambu:
@claude review

jwnasambu:
@ibacher and @dkayiwa Kindly feel free to review this PR at you convenient time please!

api/src/main/java/org/openmrs/util/Argon2PasswordEncoder.java
```diff
@@ -0,0 +1,458 @@
...
+	boolean matches(String rawPassword, String encodedHash) {
+		String[] parts = parseHash(encodedHash);
+		if (parts.length < 2) {
+			return false;
+		}
+		byte[] salt = Base64.getUrlDecoder().decode(parts[0]);
+		byte[] expectedHash = Base64.getUrlDecoder().decode(parts[1]);
+		byte[] actualHash = argon2id(rawPassword.getBytes(StandardCharsets.UTF_8), salt, expectedHash.length);
+		return constantTimeEquals(expectedHash, actualHash);
+	}
```
dkayiwa, line 63: `matches()` recomputes the hash with whatever memory, iterations and parallelism this encoder instance is currently configured with, and `parseHash()` only ever reads the salt and hash back out of the stored string, so the `m=...,t=...,p=...` section each hash carries is written but never read. The runtime properties this PR adds are exactly the knobs an admin is invited to tune, and the encoder cache in `Security` is even built to pick changes up without a restart. The first time someone tunes them, every credential stored under the old settings stops verifying: `hashMatches` sees the `$argon2id$` prefix, recomputes under the new settings, gets a different result, and returns false with no error logged anywhere. For passwords there is no legacy fallback once the stored value starts with `$argon2id$`, so that locks out every user who set or changed a password or secret answer under the old settings, until someone guesses to revert the config. I verified this on the branch: a hash encoded at t=3 fails to verify after bumping iterations to 4, a memory change does the same, and a stored string whose parameter section I tampered to `m=999999,t=42,p=7` still verifies under the original config.

Verification has to use the m/t/p parsed from the stored hash (that is what the PHC format embeds them for), with the configured values applying only to new encodes. If this class gets replaced by the library encoder from #6301 per the other thread, that covers this too, since argon2-jvm verifies against the parameters embedded in the hash. And if the tunables survive in some form, note that `users.password` is varchar(128) while the PHC string outgrows it once `security.argon2.hashLength` goes past 55 bytes, so password writes start failing at larger configured hash lengths.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -484,8 +484,19 @@ public boolean isSecretAnswer(User u, String answer) throws DAOException {
...
+			return false;
+		}
+
+		boolean matches = Security.hashMatches(answerOnRecord, answer.toLowerCase() + credentials.getSalt());
+
+		if (matches && !answerOnRecord.startsWith("$argon2id$")) {
+			String rehashedAnswer = Security.encodeCredential(answer.toLowerCase() + credentials.getSalt());
+			credentials.setSecretAnswer(rehashedAnswer);
+			updateLoginCredential(credentials);
+		}
```
dkayiwa, line 497: This write is silently dropped on the public API path. `UserServiceImpl.isSecretAnswer` is `@Transactional(readOnly = true)`, so when the check comes in through the service proxy, Hibernate runs with manual flush and a read-only session, and the `session.update()` inside `updateLoginCredential()` is discarded at commit with no error. I reproduced it on the branch: seed a legacy SHA-512 answer, call `Context.getUserService().isSecretAnswer(...)` (it returns true), re-read the credential in a fresh transaction, and it is still the SHA-512 hash. The upgrade only persists when the check happens to run inside an outer read-write call, like `changePasswordUsingSecretAnswer` invoking `isSecretAnswer` on itself past the proxy. The new UserDAOTest upgrade tests don't catch this because they call the DAO inside the test's own read-write transaction. So as merged, the transparent-upgrade criterion on TRUNK-6693 mostly just doesn't happen in production, and nothing surfaces it.

The smallest fix is to drop `readOnly = true` from `UserServiceImpl.isSecretAnswer`, which now does write, and to add a service-level test asserting the upgraded hash actually gets persisted. Keeping `isSecretAnswer` read-only and doing the upgrade only in the read-write credential flows would also work, but I would go with the first so the upgrade happens at the point of verification, where the plaintext answer is in hand.

dkayiwa (commented):
Stepping back from the diff for a moment: I think this is re-implementing infrastructure that other open PRs in the TRUNK-6661 epic already own, rather than building on them. TRUNK-6693's own acceptance criteria say to reuse the existing Argon2id credential encoder, to introduce no new hashing implementation for secret answers, and to leave `Security.encodeString()` alone. The encoder it refers to is the library-backed one #6301 (TRUNK-6663) adds via spring-security-crypto and argon2-jvm, with #6219 (TRUNK-6662) putting the PasswordEncoder seam in place first, and #6230 (TRUNK-6666) owning the lazy rehash on login. Since none of those have merged yet, this PR ships its own from-scratch encoder, its own work-factor properties, its own hashMatches wiring, and it also moves the password write paths (saveUser and both changePassword overloads) to Argon2, which is TRUNK-6663 territory. Whichever of these PRs merges first, the others now conflict.

My suggestion is to hold this one until #6219 and #6301 are in, then cut it down to the secret-answer layer (changeQuestionAnswer, isSecretAnswer, the lazy upgrade and its tests) on top of the merged encoder. That would also resolve the open thread on the hand-rolled Argon2id implementation, since that class would go away entirely.

api/src/test/java/org/openmrs/api/db/UserDAOTest.java
```diff
@@ -203,4 +200,63 @@ public void isSecretAnswer_shouldReturnFalseWhenTheAnswerDoesNotMatch() {
...
+		dao.updateLoginCredential(lc);
+		
+		assertTrue(dao.isSecretAnswer(userJoe, SECRET_ANSWER), "legacy SHA-1 hash should still verify");
+		
+		lc = dao.getLoginCredential(userJoe);
+		assertTrue(lc.getSecretAnswer().startsWith("$argon2id$"), "legacy SHA-1 hash should be upgraded to Argon2id after verification");
+	}
+	
+	@Test
+	public void setUserActivationKey_shouldStillWorkAfterSecretAnswerMigration() {
```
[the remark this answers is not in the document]
dkayiwa: The reworked test still can't fail: it hashes the token itself with `Security.encodeString`, sets the result on `lc`, and then asserts on `lc.getActivationKey()`, the same value on the same object it just set, so it passes no matter what the production code does with activation keys. To actually pin the SHA-512 contract it needs to drive the production path, along the lines of the existing `UserServiceTest.setUserActivationKey_shouldCreateUserActivationKey` (call `userService.setUserActivationKey(user)`, tolerate the MessageException from the missing mail config, then assert on the shape of the activation key re-read from the DAO).

jwnasambu:
@dkayiwa Thanks for the review. I will put on hold on this one until https://github.com/openmrs/openmrs-core/pull/6219 and https://github.com/openmrs/openmrs-core/pull/6301 are in

ibacher (changes requested):
So this implements it's own Argon2PasswordEncoder and that is an absolute non-starter. OpenMRS should not own the Argon2 implementation. Instead, we should be getting an existing implementation ideally from Spring or where Spring will not work from another well-known open source library like Password4J or BouncyCastle. Introducing our own implementation is just too much risk in a critical path. Half the reason we're in this mess that we're trying to clean-up here is because we've gone too far into implementing our own security solutions.

PR #6348, opened 2026-07-17
TRUNK-6694 & TRUNK-6695: Avoid synchronization in service fetching

ibacher:

This implements two changes:

1. `ServiceContext#getService()` was synchronized which forced threads to access it synchronously. The problem is, that getting service instances is something we will wind up doing in most request processes, often multiple times and the only reason we needed the synchronization guard was to check if we were refreshing context (while the context is being refreshed, thread legitimately need to wait until all services are available). Consequently, this winds up being one of the hottest paths in the code base. With this PR, we switch from synchronization to making the context refresh indicator `volatile` and only synchronize if we are refreshing the context. This makes getting a registered service faster in the normal case.
2. The map of services was not correctly synchronized. It is now switched for a `ConcurrentHashMap`. Since this map is read-mostly, this shouldn't drastically affect things.

Since this is a performance consideration, here's the result of a benchmark on my 16-core M3:

| Threads | OLD (`synchronized` + `HashMap`) | NEW (`volatile` fast path + `ConcurrentHashMap`) | Speedup |
|--------:|--------------------------------:|-------------------------------------------------:|--------:|
| 1  | 240 M ops/s | 271 M ops/s  | 1.1× |
| 2  | 100 M ops/s | 518 M ops/s  | 5.2× |
| 4  | 19 M ops/s  | 1.01 B ops/s | 54×  |
| 8  | 23 M ops/s  | 2.02 B ops/s | 87×  |
| 16 | 24 M ops/s  | 1.62 B ops/s | 67×  |
| 32 | 24 M ops/s  | 2.95 B ops/s | 123× |

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6694 & https://issues.openmrs.org/browse/TRUNK-6695

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/ServiceContext.java
+++ b/api/src/main/java/org/openmrs/api/context/ServiceContext.java
@@ -17,6 +17,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.aopalliance.aop.Advice;
 import org.openmrs.api.APIException;
@@ -82,7 +83,7 @@ public class ServiceContext implements ApplicationContextAware {
 
 	private ApplicationContext applicationContext;
 
-	private static boolean refreshingContext = false;
+	private static volatile boolean refreshingContext = false;
 
 	private static final Object refreshingContextLock = new Object();
 
@@ -93,7 +94,7 @@ public class ServiceContext implements ApplicationContextAware {
 	private boolean useSystemClassLoader = false;
 
 	// Cached service objects
-	Map<Class, Object> services = new HashMap<>();
+	Map<Class, Object> services = new ConcurrentHashMap<>();
 
 	// Advisors added to services by this service
 	Map<Class, Set<Advisor>> addedAdvisors = new HashMap<>();
@@ -680,20 +681,23 @@ public class ServiceContext implements ApplicationContextAware {
 			log.trace("Getting service: " + cls);
 		}
 
-		// if the context is refreshing, wait until it is
-		// done -- otherwise a null service might be returned
-		synchronized (refreshingContextLock) {
-			try {
-				while (refreshingContext) {
-					log.debug("Waiting to get service: {} while the context is being refreshed", cls);
+		// Only synchronize if we're refreshing the context; during context refreshes, we need
+		// to wait for services to be available. Otherwise we can take the fast-path.
+		if (refreshingContext) {
+			synchronized (refreshingContextLock) {
+				try {
+					while (refreshingContext) {
+						log.debug("Waiting to get service: {} while the context is being refreshed", cls);
 
-					refreshingContextLock.wait();
+						refreshingContextLock.wait();
 
-					log.debug("Finished waiting to get service {} while the context was being refreshed", cls);
+						log.debug("Finished waiting to get service {} while the context was being refreshed", cls);
+					}
+
+				} catch (InterruptedException e) {
+					log.warn("Refresh lock was interrupted", e);
+					Thread.currentThread().interrupt();
 				}
-
-			} catch (InterruptedException e) {
-				log.warn("Refresh lock was interrupted", e);
 			}
 		}
 
--- a/api/src/main/java/org/openmrs/api/context/ServiceContext.java
+++ b/api/src/main/java/org/openmrs/api/context/ServiceContext.java
@@ -696,7 +696,6 @@ public class ServiceContext implements ApplicationContextAware {
 
 				} catch (InterruptedException e) {
 					log.warn("Refresh lock was interrupted", e);
-					Thread.currentThread().interrupt();
 				}
 			}
 		}
@@ -869,9 +868,7 @@ public class ServiceContext implements ApplicationContextAware {
 	 *         doneRefreshingContext()
 	 */
 	public boolean isRefreshingContext() {
-		synchronized (refreshingContextLock) {
-			return refreshingContext;
-		}
+		return refreshingContext;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


ibacher:
@claude review

api/src/main/java/org/openmrs/api/context/ServiceContext.java
```diff
@@ -82,7 +83,7 @@ public class ServiceContext implements ApplicationContextAware {
 
 	private ApplicationContext applicationContext;
 
-	private static boolean refreshingContext = false;
+	private static volatile boolean refreshingContext = false;
```
dkayiwa, line 86: With this field volatile now, the synchronized block in `isRefreshingContext()` is a lock around a single volatile read, which buys nothing. A bare `return refreshingContext;` there would finish the de-synchronization this PR is doing. Take it or leave it.

api/src/main/java/org/openmrs/api/context/ServiceContext.java
```diff
@@ -680,20 +681,23 @@ public <T> T getService(Class<? extends T> cls) {
...
-					log.debug("Finished waiting to get service {} while the context was being refreshed", cls);
-				}
+						log.debug("Finished waiting to get service {} while the context was being refreshed", cls);
+					}
 
-			} catch (InterruptedException e) {
-				log.warn("Refresh lock was interrupted", e);
+				} catch (InterruptedException e) {
+					log.warn("Refresh lock was interrupted", e);
+					Thread.currentThread().interrupt();
```
dkayiwa: This catch is the only new code no test reaches. Running just this test class, jacoco shows 697-699 as the sole missed lines in the new `getService()` body, which is what the codecov patch warning points at. It can be covered deterministically: start a refresh, spawn a lookup thread, spin until it reaches `Thread.State.WAITING`, interrupt it, then assert the lookup completed without `doneRefreshingContext()` ever running, returned the service, and left the thread's interrupt flag set. That would also pin down a subtle semantic this path preserves: an interrupted waiter gives up waiting and reads the registry mid-refresh, which is worth documenting in a test rather than leaving to archaeology. Since this PR already improves interrupt handling here, the same catch in `runOpenmrsServiceOnStartup()` further down still swallows the flag and could take the same one-liner. Not blocking.

api/src/test/java/org/openmrs/api/context/ServiceContextTest.java
```diff
@@ -102,4 +107,82 @@ public void getModuleOpenmrsServices_shouldRaiseApiExceptionWithNullClassInstanc
...
+		assertTrue(lookupComplete.await(5, TimeUnit.SECONDS));
+		assertNull(thrown.get());
+		assertNotNull(retrievedService.get());
+	}
+
+	/**
+	 * @see ServiceContext#setService(Class, Object)
+	 */
+	@Test
+	public void setService_shouldMakeRegistrationsVisibleToOtherThreads() throws Exception {
```
dkayiwa, line 172: I ran this exact test file against master without the ServiceContext changes, and this test passes there too. `setService()` completes before `reader.start()`, and `Thread.start()` guarantees the started thread sees everything written before it (JLS 17.4.5), so the assertion holds however badly `services` is synchronized. As written it can't catch the unsafe-publication problem the `ConcurrentHashMap` switch fixes. To give it teeth, invert the order: start the reader first, have it poll `getService()` in a loop (catching `ServiceNotFoundException`) against a deadline, then register the service from the test thread once the reader is running. No unit test can prove visibility deterministically, but that way the read at least races the write instead of being ordered behind it. That is what I'd do here; the lighter alternative is renaming this to the setService/getService round-trip it does verify. Either way, not blocking.

api/src/test/java/org/openmrs/api/context/ServiceContextTest.java
```diff
@@ -102,4 +107,86 @@ public void getModuleOpenmrsServices_shouldRaiseApiExceptionWithNullClassInstanc
...
+
+		try {
+			// wait for the lookup thread to reach getService, then give it time to block
+			assertTrue(lookupStarted.await(5, TimeUnit.SECONDS));
+			Thread.sleep(500);
+
+			// the lookup must not have returned while the refresh is in progress
+			assertNull(retrievedService.get());
+			assertTrue(lookup.isAlive());
+		} finally {
```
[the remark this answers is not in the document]
dkayiwa: One corner of this survives the CountDownLatch rewrite: `assertFalse(lookupComplete.await(500, TimeUnit.MILLISECONDS))` still assumes the lookup thread gets scheduled and reaches the wait within 500 ms. On a stalled runner where the thread starts later than that, a non-blocking `getService()` would still pass, since the latch only counts down after the window has already been checked. Spinning until the thread reports `Thread.State.WAITING` (and failing fast if the latch drops first) proves it parked in the wait loop, is immune to scheduling delays, and typically finishes in milliseconds instead of a flat 500 ms per run. Optional; the current form only misses in that pathological case.

api/src/main/java/org/openmrs/api/context/ServiceContext.java
ibacher: Tbh, I'm more inclined to just rollback the one newly added line here. It's tangential to the goal of this PR and the code path isn't meaningfully different or untested.

api/src/test/java/org/openmrs/api/context/ServiceContextTest.java
ibacher: Yeah, this is a reasonable point, though, tbf, this wasn't exactly a test of resource contention, but it's easy to make it do so.

api/src/main/java/org/openmrs/api/context/ServiceContext.java
ibacher: Fixed

PR #6350, opened 2026-07-18
Fix operator precedence bug in getRefByUuid() exception message

Kapil-kundu:
The **getRefByUuid()** method contains an operator precedence issue in the **APIException** message construction:

`throw new APIException("Unsupported type for getRefByUuid: " + type != null ? type.getName() : "null");`

Due to Java operator precedence, the ternary operator is evaluated incorrectly, causing the exception message to be wrong and potentially leading to a **NullPointerException** when **type** is **null**.

**Proposed Fix**
Wrap the ternary expression in parentheses:

`throw new APIException(
    "Unsupported type for getRefByUuid: "
        + (type != null ? type.getName() : "null")
);`

**Impact**

- Generates the correct error message.
- Prevents incorrect evaluation of the ternary expression.
- Avoids a possible NullPointerException when type is null.

JIra Ticker : https://openmrs.atlassian.net/browse/TRUNK-6714


Kapil-kundu:
Some files added mistakely

PR #6354, opened 2026-07-20
TRUNK-6706: Disable hibernate.generate_statistics by default

Janani-Balasooriya:

Changed the default value of `hibernate.generate_statistics` in `api/src/main/resources/hibernate.default.properties` from `true` to `false`. Statistics generation maintains counters on every query, entity, collection, and cache operation, and the only consumer in core is a shutdown-time debug log dump, so this default was adding overhead for data that's almost never read.

No code changes were needed to preserve explicit overrides:
`HibernateContextDAO.mergeDefaultRuntimeProperties()` already only applies a default when the key isn't already present in the user's runtime properties, so anyone setting `hibernate.generate_statistics=true` themselves will have that value preserved.

Added two unit tests to `ContextDAOTest.java`:
- `should_defaultGenerateStatisticsToFalse()`
- `should_notOverrideExplicitGenerateStatisticsSetting()`

Note: I noticed three tests in `ContextTest.java` check cache eviction using Hibernate's statistics counter, which no longer updates by default after this change. Flagging in case a follow-up is wanted, happy to help fix if so.

I noticed this ticket is labeled `backport`. Should this fix also be applied to older maintained branches, or is that handled separately by the team?

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6706

  No? -> execute above command

### DIFF
--- a/api/src/main/resources/hibernate.default.properties
+++ b/api/src/main/resources/hibernate.default.properties
@@ -26,7 +26,7 @@ hibernate.c3p0.acquire_increment=1
 hibernate.c3p0.checkoutTimeout=30000
 
 # Hibernate debugging options
-hibernate.generate_statistics=true
+hibernate.generate_statistics=false
 hibernate.cache.use_structured_entries=false
 
 # Hibernate cache
--- a/api/src/test/java/org/openmrs/api/context/ContextTest.java
+++ b/api/src/test/java/org/openmrs/api/context/ContextTest.java
@@ -297,20 +297,27 @@ public class ContextTest extends BaseContextSensitiveTest {
 	public void evictEntity_shouldClearTheEntityFromCaches() {
 		TestTransaction.end();
 
-		// Load the person name so that it is stored in the cache
-		PersonName name = Context.getPersonService().getPersonName(PERSON_NAME_ID_2);
+		// Statistics are disabled by default; enable them so hit counts below are meaningful
+		boolean statisticsEnabled = sf.getStatistics().isStatisticsEnabled();
+		sf.getStatistics().setStatisticsEnabled(true);
+		try {
+			// Load the person name so that it is stored in the cache
+			PersonName name = Context.getPersonService().getPersonName(PERSON_NAME_ID_2);
 
-		// Clear session so that the first-level cache is empty
-		Context.flushSession();
-		Context.clearSession();
+			// Clear session so that the first-level cache is empty
+			Context.flushSession();
+			Context.clearSession();
 
-		// evictEntity
-		Context.evictEntity(name);
+			// evictEntity
+			Context.evictEntity(name);
 
-		// Assert that the entity name has been removed from cache
-		long hitCount = sf.getStatistics().getSecondLevelCacheHitCount();
-		Context.getPersonService().getPersonName(PERSON_NAME_ID_2);
-		assertThat(sf.getStatistics().getSecondLevelCacheHitCount(), is(hitCount));
+			// Assert that the entity name has been removed from cache
+			long hitCount = sf.getStatistics().getSecondLevelCacheHitCount();
+			Context.getPersonService().getPersonName(PERSON_NAME_ID_2);
+			assertThat(sf.getStatistics().getSecondLevelCacheHitCount(), is(hitCount));
+		} finally {
+			sf.getStatistics().setStatisticsEnabled(statisticsEnabled);
+		}
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


dkayiwa:
@claude review

api/src/main/resources/hibernate.default.properties
```diff
@@ -26,7 +26,7 @@ hibernate.c3p0.acquire_increment=1
 hibernate.c3p0.checkoutTimeout=30000
 
 # Hibernate debugging options
-hibernate.generate_statistics=true
+hibernate.generate_statistics=false
```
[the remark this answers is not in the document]
Janani-Balasooriya: Fixed in the latest commit: the three cache tests (evictEntity_shouldClearTheEntityFromCaches,
evictAllEntities_shouldClearAllEntityFromCaches, clearEntireCache_shouldClearEntireCache) now enable
sf.getStatistics().setStatisticsEnabled(true) at the start of their own test body and restore the prior state in a finally block, so they keep exercising real getSecondLevelCacheHitCount() values independent of the
new global default.

Ran ContextTest and the full mvn clean install, everything passes except one pre-existing failure in isSessionOpen_shouldReturnTrueIfSessionIsClosed that I confirmed is unrelated to this PR, it reproduces identically on unmodified upstream/master with none of my changes present.

jwnasambu:
@claude review

jwnasambu (approved):

jwnasambu:
@dkayiwa, @ibacher kindly help review this PR please! hahaha though I approved it I don't have a clear picture of the expectations now.

api/src/test/java/org/openmrs/api/context/ContextTest.java
```diff
@@ -297,20 +297,27 @@ public void becomeUser_shouldChangeLocaleWhenBecomeAnotherUser() {
 	public void evictEntity_shouldClearTheEntityFromCaches() {
 		TestTransaction.end();
 
-		// Load the person name so that it is stored in the cache
-		PersonName name = Context.getPersonService().getPersonName(PERSON_NAME_ID_2);
+		// Statistics are disabled by default; enable them so hit counts below are meaningful
```
ibacher, line 300: This would be better done by moving these tests into a `@Nested` inner class with a `@BeforeEach` and `@AfterEach` that do the setup and tear-down.

api/src/main/resources/hibernate.default.properties
dkayiwa, line 29: This needs handling before merge, and it is not visible from this branch. Since the 21 July merge base, master has picked up five test classes that read Hibernate's statistics counters, none of which touched statistics when you branched. I applied only this one line to current master (3310bc1d9) and ran them.

Six tests fail outright. `ObsServiceTest` routes its paging assertions through a helper that first asserts `statistics.isStatisticsEnabled()`, on the grounds that "a test that silently measures nothing is worse than no test", so five of them fail immediately. `PatientIdentifierBatchLoadTest.getPatients_shouldLoadIdentifiersInBoundedNumberOfQueries` asserts a collection fetch count of 1 and gets 0.

Four more keep passing while measuring nothing, which is the worse half. `HibernateLocationDAOTest` bounds `getEntityLoadCount()` at 10, and `HibernatePatientDAOTest` and `HibernateConceptDAOTest` compare `getPrepareStatementCount()` for a small page against a large one. I dropped a temporary `assertTrue(count > 0)` next to each of those and all four failed, so the counters really do sit at 0 and the bounds hold trivially. The N+1 guards from TRUNK-6681, TRUNK-6729 and TRUNK-6734 would report green while checking nothing, so the batch loading they protect could regress unnoticed.

Rather than repeat the try/finally in five more classes, I would enable statistics once for context-sensitive tests. `getRuntimeProperties()` in `BaseContextSensitiveNonTransactionalTest` already injects test-only Hibernate settings (dialect, url, hbm2ddl_auto), and runtime properties beat `hibernate.default.properties` in `HibernateSessionFactoryBean`, so a single `runtimeProperties.setProperty("hibernate.generate_statistics", "true")` there keeps the shipped default false while leaving every counting test working, including ones added later. I tried exactly that on top of the flipped default and the six failures and the four probes all go green. That route also removes the need for the `ContextTest` wrapping altogether, since the suite gets its statistics back the way master has them today, so it is worth settling before restructuring those tests the way @ibacher suggests.

Either way this wants a rebase on master first, since none of it reproduces on the current branch.

dkayiwa (commented):
On your backport question: yes, the `backport` label means this should reach the maintained release branch too, but raise that as a separate PR once this one has merged rather than now.

api/src/main/resources/hibernate.default.properties
ibacher: That's a good point about the other tests that have landed. I resisted suggesting overriding the runtime property by _default_ only because while that fixes things for _tests_, it could also leave us writing code that passes in test but fails in production if, for whatever reason, a production code pathway started analyzing Hibernate statistics. The issue being that our context-sensitive defaults affect not only this repo, but the entire ecosystem.

Maybe we create a new sub-hierarchy of classes with this enabled to make it an explicit opt-in rather than a default? E.g., `BaseContextSensitiveHibernateStatisticsTest` and maybe `BaseModuleContextSensitiveHibernateStatisticsTest`.

ibacher (approved):

PR #6355, opened 2026-07-20
TRUNK-6664: Implement deterministic bootstrap password mechanism

praisekavuma4:
- Added PBKDF2-based deterministic password generation
- Added bootstrap password methods to UserService
- Added unit tests (15/15 passing)

Summary
This PR implements a deterministic bootstrap password mechanism for OpenMRS, allowing administrators to generate secure, temporary passwords for user provisioning and controlled resets. The password is derived from the user's UUID and a system-wide salt using PBKDF2, ensuring the same user always gets the same password without storing it in the database.

What This PR Adds
1. Core PBKDF2 Logic (Security.java)
generateDeterministicHash(String input, int iterations) – PBKDF2-based derivation
getBootstrapSystemSalt() – Retrieves system salt from global properties
getBootstrapIterations() – Retrieves iteration count from global properties
generateBootstrapPassword(User user) – Generates deterministic password from UUID + salt
validateBootstrapPassword(User user, String password) – Validates a password against the user's bootstrap password
forcePasswordChange(User user) – Sets the forcePassword user property to force password change on next login
isBootstrapPasswordExpired(User user) – Checks if a bootstrap password has expired

2. Service Layer Integration (UserService.java / UserServiceImpl.java)
Added 5 new methods to the UserService interface:
generateBootstrapPassword(User user)
validateBootstrapPassword(User user, String password)
forcePasswordChange(User user)
isBootstrapPassword(User user, String password)
isBootstrapPasswordExpired(User user)

3. Global Properties (OpenmrsConstants.java)
security.bootstrap.systemSalt – System salt for password derivation (must be kept secret)
security.bootstrap.iterations – PBKDF2 iteration count (default: 600,000)

4. Unit Tests (BootstrapPasswordTest.java)
15 unit tests covering:
Deterministic hash generation
Bootstrap password generation and validation
Error handling (null user, missing UUID, missing salt)
Expiry detection via forcePassword property
All tests passing: Tests run: 15, Failures: 0, Errors: 0

### DIFF
--- a/api/src/main/java/org/openmrs/api/UserService.java
+++ b/api/src/main/java/org/openmrs/api/UserService.java
@@ -643,17 +643,6 @@ public interface UserService extends OpenmrsService {
 	void forcePasswordChange(User user);
 	
 	/**
-	 * Checks if a user is currently using their bootstrap password.
-	 * 
-	 * @param user the user
-	 * @param password the password to check
-	 * @return true if the password is the user's bootstrap password
-	 * @since 2.8.8
-	 */
-	@Authorized( { PrivilegeConstants.GET_USERS })
-	boolean isBootstrapPassword(User user, String password);
-	
-	/**
 	 * Checks if a user's bootstrap password has expired.
 	 * 
 	 * Bootstrap passwords are considered expired when the user has been
--- a/api/src/main/java/org/openmrs/api/UserService.java
+++ b/api/src/main/java/org/openmrs/api/UserService.java
@@ -605,4 +605,64 @@ public interface UserService extends OpenmrsService {
 	 */
 	@Authorized
 	String getLastLoginTime(User user);
+
+		/**
+	 * Generates a deterministic bootstrap password for a user.
+	 * 
+	 * The password is derived from the user's UUID and a system-wide salt using PBKDF2.
+	 * It is not stored in the database and can be regenerated at any time.
+	 * 
+	 * @param user the user for whom to generate the bootstrap password
+	 * @return the generated bootstrap password
+	 * @throws APIException if the user is null, has no UUID, or system salt is not configured
+	 * @since 2.8.8
+	 */
+	@Authorized( { PrivilegeConstants.GET_USERS })
+	String generateBootstrapPassword(User user);
+	
+	/**
+	 * Validates a password against a user's bootstrap password.
+	 * 
+	 * @param user the user
+	 * @param password the password to validate
+	 * @return true if the password matches the user's bootstrap password
+	 * @since 2.8.8
+	 */
+	@Authorized( { PrivilegeConstants.GET_USERS })
+	boolean validateBootstrapPassword(User user, String password);
+	
+	/**
+	 * Forces a user to change their password on next login.
+	 * 
+	 * This is typically called after a user successfully logs in with a bootstrap password.
+	 * 
+	 * @param user the user whose password change should be forced
+	 * @since 2.8.8
+	 */
+	@Authorized( { PrivilegeConstants.EDIT_USER_PASSWORDS })
+	void forcePasswordChange(User user);
+	
+	/**
+	 * Checks if a user is currently using their bootstrap password.
+	 * 
+	 * @param user the user
+	 * @param password the password to check
+	 * @return true if the password is the user's bootstrap password
+	 * @since 2.8.8
+	 */
+	@Authorized( { PrivilegeConstants.GET_USERS })
+	boolean isBootstrapPassword(User user, String password);
+	
+	/**
+	 * Checks if a user's bootstrap password has expired.
+	 * 
+	 * Bootstrap passwords are considered expired when the user has been
+	 * successfully authenticated with them and was forced to change their password.
+	 * 
+	 * @param user the user
+	 * @return true if the bootstrap password is expired
+	 * @since 2.8.8
+	 */
+	@Authorized( { PrivilegeConstants.GET_USERS })
+	boolean isBootstrapPasswordExpired(User user);
 }
--- a/api/src/main/java/org/openmrs/api/UserService.java
+++ b/api/src/main/java/org/openmrs/api/UserService.java
@@ -607,4 +607,64 @@ public interface UserService extends OpenmrsService {
 	 */
 	@Authorized
 	String getLastLoginTime(User user);
+
+		/**
+	 * Generates a deterministic bootstrap password for a user.
+	 * 
+	 * The password is derived from the user's UUID and a system-wide salt using PBKDF2.
+	 * It is not stored in the database and can be regenerated at any time.
+	 * 
+	 * @param user the user for whom to generate the bootstrap password
+	 * @return the generated bootstrap password
+	 * @throws APIException if the user is null, has no UUID, or system salt is not configured
+	 * @since 2.8.8
+	 */
+	@Authorized( { PrivilegeConstants.GET_USERS })
+	String generateBootstrapPassword(User user);
+	
+	/**
+	 * Validates a password against a user's bootstrap password.
+	 * 
+	 * @param user the user
+	 * @param password the password to validate
+	 * @return true if the password matches the user's bootstrap password
+	 * @since 2.8.8
+	 */
+	@Authorized( { PrivilegeConstants.GET_USERS })
+	boolean validateBootstrapPassword(User user, String password);
+	
+	/**
+	 * Forces a user to change their password on next login.
+	 * 
+	 * This is typically called after a user successfully logs in with a bootstrap password.
+	 * 
+	 * @param user the user whose password change should be forced
+	 * @since 2.8.8
+	 */
+	@Authorized( { PrivilegeConstants.EDIT_USER_PASSWORDS })
+	void forcePasswordChange(User user);
+	
+	/**
+	 * Checks if a user is currently using their bootstrap password.
+	 * 
+	 * @param user the user
+	 * @param password the password to check
+	 * @return true if the password is the user's bootstrap password
+	 * @since 2.8.8
+	 */
+	@Authorized( { PrivilegeConstants.GET_USERS })
+	boolean isBootstrapPassword(User user, String password);
+	
+	/**
+	 * Checks if a user's bootstrap password has expired.
+	 * 
+	 * Bootstrap passwords are considered expired when the user has been
+	 * successfully authenticated with them and was forced to change their password.
+	 * 
+	 * @param user the user
+	 * @return true if the bootstrap password is expired
+	 * @since 2.8.8
+	 */
+	@Authorized( { PrivilegeConstants.GET_USERS })
+	boolean isBootstrapPasswordExpired(User user);
 }
--- a/api/src/main/java/org/openmrs/api/UserService.java
+++ b/api/src/main/java/org/openmrs/api/UserService.java
@@ -645,17 +645,6 @@ public interface UserService extends OpenmrsService {
 	void forcePasswordChange(User user);
 	
 	/**
-	 * Checks if a user is currently using their bootstrap password.
-	 * 
-	 * @param user the user
-	 * @param password the password to check
-	 * @return true if the password is the user's bootstrap password
-	 * @since 2.8.8
-	 */
-	@Authorized( { PrivilegeConstants.GET_USERS })
-	boolean isBootstrapPassword(User user, String password);
-	
-	/**
 	 * Checks if a user's bootstrap password has expired.
 	 * 
 	 * Bootstrap passwords are considered expired when the user has been
--- a/api/src/main/java/org/openmrs/api/UserService.java
+++ b/api/src/main/java/org/openmrs/api/UserService.java
@@ -617,7 +617,7 @@ public interface UserService extends OpenmrsService {
 	 * @param user the user for whom to generate the bootstrap password
 	 * @return the generated bootstrap password
 	 * @throws APIException if the user is null, has no UUID, or pepper is not configured
-	 * @since 2.8.9
+	 * @since 2.8.10
 	 */
 	@Authorized({ PrivilegeConstants.EDIT_USER_PASSWORDS })
 	String generateBootstrapPassword(User user);
@@ -627,8 +627,8 @@ public interface UserService extends OpenmrsService {
 	 *
 	 * @param user the user
 	 * @param password the password to validate
-	 * @return true if the password matches the user's bootstrap password
-	 * @since 2.8.9
+	 * @return true if the password matches the user's bootstrap password and the bootstrap password has not expired
+	 * @since 2.8.10
 	 */
 	@Authorized({ PrivilegeConstants.GET_USERS })
 	@Logging(ignoredArgumentIndexes = { 1 })
@@ -638,9 +638,11 @@ public interface UserService extends OpenmrsService {
 	 * Forces a user to change their password on next login.
 	 *
 	 * This is typically called after a user successfully logs in with a bootstrap password.
+	 * It also marks any issued bootstrap password as expired, so the consumed
+	 * bootstrap credential cannot be used again.
 	 *
 	 * @param user the user whose password change should be forced
-	 * @since 2.8.9
+	 * @since 2.8.10
 	 */
 	@Authorized({ PrivilegeConstants.EDIT_USER_PASSWORDS })
 	void forcePasswordChange(User user);
@@ -650,10 +652,13 @@ public interface UserService extends OpenmrsService {
 	 *
 	 * Bootstrap passwords are considered expired when the user has been
 	 * successfully authenticated with them and was forced to change their password.
+	 * Expiry is tracked in a dedicated user property independent of the general
+	 * force-password flag and only ever moves one way, so a consumed bootstrap
+	 * password stays expired.
 	 *
 	 * @param user the user
 	 * @return true if the bootstrap password is expired
-	 * @since 2.8.9
+	 * @since 2.8.10
 	 */
 	@Authorized({ PrivilegeConstants.GET_USERS })
 	boolean isBootstrapPasswordExpired(User user);
--- a/api/src/main/java/org/openmrs/api/UserService.java
+++ b/api/src/main/java/org/openmrs/api/UserService.java
@@ -608,30 +608,31 @@ public interface UserService extends OpenmrsService {
 	@Authorized
 	String getLastLoginTime(User user);
 
-		/**
+	/**
 	 * Generates a deterministic bootstrap password for a user.
 	 * 
-	 * The password is derived from the user's UUID and a system-wide salt using PBKDF2.
+	 * The password is derived from the user's UUID and a system-wide pepper using PBKDF2.
 	 * It is not stored in the database and can be regenerated at any time.
 	 * 
 	 * @param user the user for whom to generate the bootstrap password
 	 * @return the generated bootstrap password
-	 * @throws APIException if the user is null, has no UUID, or system salt is not configured
-	 * @since 2.8.8
+	 * @throws APIException if the user is null, has no UUID, or pepper is not configured
+	 * @since 2.8.9
 	 */
-	@Authorized( { PrivilegeConstants.GET_USERS })
+	@Authorized( { PrivilegeConstants.EDIT_USER_PASSWORDS })   
 	String generateBootstrapPassword(User user);
 	
 	/**
-	 * Validates a password against a user's bootstrap password.
-	 * 
-	 * @param user the user
-	 * @param password the password to validate
-	 * @return true if the password matches the user's bootstrap password
-	 * @since 2.8.8
-	 */
-	@Authorized( { PrivilegeConstants.GET_USERS })
-	boolean validateBootstrapPassword(User user, String password);
+     * Validates a password against a user's bootstrap password.
+     * 
+     * @param user the user
+     * @param password the password to validate
+     * @return true if the password matches the user's bootstrap password
+     * @since 2.8.9
+     */
+    @Authorized( { PrivilegeConstants.GET_USERS })
+    @Logging(ignoredArgumentIndexes = { 1 })
+    boolean validateBootstrapPassword(User user, String password);
 	
 	/**
 	 * Forces a user to change their password on next login.
@@ -639,7 +640,7 @@ public interface UserService extends OpenmrsService {
 	 * This is typically called after a user successfully logs in with a bootstrap password.
 	 * 
 	 * @param user the user whose password change should be forced
-	 * @since 2.8.8
+	 * @since 2.8.9
 	 */
 	@Authorized( { PrivilegeConstants.EDIT_USER_PASSWORDS })
 	void forcePasswordChange(User user);
@@ -652,7 +653,7 @@ public interface UserService extends OpenmrsService {
 	 * 
 	 * @param user the user
 	 * @return true if the bootstrap password is expired
-	 * @since 2.8.8
+	 * @since 2.8.9
 	 */
 	@Authorized( { PrivilegeConstants.GET_USERS })
 	boolean isBootstrapPasswordExpired(User user);
--- a/api/src/main/java/org/openmrs/api/UserService.java
+++ b/api/src/main/java/org/openmrs/api/UserService.java
@@ -610,51 +610,51 @@ public interface UserService extends OpenmrsService {
 
 	/**
 	 * Generates a deterministic bootstrap password for a user.
-	 * 
+	 *
 	 * The password is derived from the user's UUID and a system-wide pepper using PBKDF2.
 	 * It is not stored in the database and can be regenerated at any time.
-	 * 
+	 *
 	 * @param user the user for whom to generate the bootstrap password
 	 * @return the generated bootstrap password
 	 * @throws APIException if the user is null, has no UUID, or pepper is not configured
 	 * @since 2.8.9
 	 */
-	@Authorized( { PrivilegeConstants.EDIT_USER_PASSWORDS })   
+	@Authorized({ PrivilegeConstants.EDIT_USER_PASSWORDS })
 	String generateBootstrapPassword(User user);
-	
+
 	/**
-     * Validates a password against a user's bootstrap password.
-     * 
-     * @param user the user
-     * @param password the password to validate
-     * @return true if the password matches the user's bootstrap password
-     * @since 2.8.9
-     */
-    @Authorized( { PrivilegeConstants.GET_USERS })
-    @Logging(ignoredArgumentIndexes = { 1 })
-    boolean validateBootstrapPassword(User user, String password);
-	
+	 * Validates a password against a user's bootstrap password.
+	 *
+	 * @param user the user
+	 * @param password the password to validate
+	 * @return true if the password matches the user's bootstrap password
+	 * @since 2.8.9
+	 */
+	@Authorized({ PrivilegeConstants.GET_USERS })
+	@Logging(ignoredArgumentIndexes = { 1 })
+	boolean validateBootstrapPassword(User user, String password);
+
 	/**
 	 * Forces a user to change their password on next login.
-	 * 
+	 *
 	 * This is typically called after a user successfully logs in with a bootstrap password.
-	 * 
+	 *
 	 * @param user the user whose password change should be forced
 	 * @since 2.8.9
 	 */
-	@Authorized( { PrivilegeConstants.EDIT_USER_PASSWORDS })
+	@Authorized({ PrivilegeConstants.EDIT_USER_PASSWORDS })
 	void forcePasswordChange(User user);
-	
+
 	/**
 	 * Checks if a user's bootstrap password has expired.
-	 * 
+	 *
 	 * Bootstrap passwords are considered expired when the user has been
 	 * successfully authenticated with them and was forced to change their password.
-	 * 
+	 *
 	 * @param user the user
 	 * @return true if the bootstrap password is expired
 	 * @since 2.8.9
 	 */
-	@Authorized( { PrivilegeConstants.GET_USERS })
+	@Authorized({ PrivilegeConstants.GET_USERS })
 	boolean isBootstrapPasswordExpired(User user);
 }
--- a/api/src/main/java/org/openmrs/api/UserService.java
+++ b/api/src/main/java/org/openmrs/api/UserService.java
@@ -606,30 +606,31 @@ public interface UserService extends OpenmrsService {
 	@Authorized
 	String getLastLoginTime(User user);
 
-		/**
+	/**
 	 * Generates a deterministic bootstrap password for a user.
 	 * 
-	 * The password is derived from the user's UUID and a system-wide salt using PBKDF2.
+	 * The password is derived from the user's UUID and a system-wide pepper using PBKDF2.
 	 * It is not stored in the database and can be regenerated at any time.
 	 * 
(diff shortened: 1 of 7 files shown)


praisekavuma4:
@ibacher , @dkayiwa , @jwnasambu  kindly check on this PR.

jwnasambu:
@claude review

praisekavuma4:
"Addressed feedback: renamed 'system salt' → 'pepper' and moved to runtime properties. All 15 tests passing. Ready for review."

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -822,4 +822,48 @@ public void changePasswordUsingActivationKey(String activationKey, String newPas
 	public String getLastLoginTime(User user) {
 		return dao.getLastLoginTime(user);
 	}
+
+	/**
+	 * @see org.openmrs.api.UserService#generateBootstrapPassword(User)
+	 */
+	@Override
+	public String generateBootstrapPassword(User user) {
```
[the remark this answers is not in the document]
praisekavuma4: Wiring the bootstrap password check into the authentication flow is planned as a follow-up ticket. This PR focuses on the backend service layer. We'll coordinate with the auth framework team and see how to  integrate it.

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -69,6 +79,216 @@ public static boolean hashMatches(String hashedPassword, String passwordToHash)
...
+        if (!StringUtils.hasText(uuid)) {
+            throw new APIException("bootstrap.user.uuid.missing", (Object[]) null);
+        }
+        
+        String salt = getBootstrapSystemSalt();
+        if (!StringUtils.hasText(salt)) {
+            throw new APIException("bootstrap.system.salt.missing", (Object[]) null);
+        }
+        
+        String input = uuid + salt;
```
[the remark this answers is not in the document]
praisekavuma4: good

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -69,6 +79,216 @@
...
+        int actualIterations = iterations > 0 ? iterations : PBKDF2_DEFAULT_ITERATIONS;
+        
+        try {
+            SecretKeyFactory factory = SecretKeyFactory.getInstance(PBKDF2_ALGORITHM);
+            KeySpec spec = new PBEKeySpec(
+				input.toCharArray(),
+				PBKDF2_SALT,   // Non-empty fixed salt
+				actualIterations,
+				PBKDF2_KEY_LENGTH
+			);
```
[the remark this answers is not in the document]
praisekavuma4: intentionally a fixed salt was used because bootstrap passwords must be deterministic (same input → same output). The system pepper, stored in runtime.properties, provides the unpredictability per installation. This is a deliberate design trade-off.

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -69,6 +79,216 @@ public static boolean hashMatches(String hashedPassword, String passwordToHash)
...
+     * @since 2.8.8
+     */
+    public static boolean validateBootstrapPassword(org.openmrs.User user, String password) {
+        if (user == null || password == null) {
+            return false;
+        }
+        
+        try {
+            String expected = generateBootstrapPassword(user);
+            return expected.equals(password);
```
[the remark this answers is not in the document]
praisekavuma4: fixed, validateBootstrapPassword now returns false if isBootstrapPasswordExpired(user) is true. thank you

solomonfortune (commented):
@praisekavuma4 could you please change your PR title to "TRUNK-6664" rather than "TRUNK-6666" which belongs to me

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -646,6 +646,19 @@ public static final Collection<String> AUTO_ROLES() {
 	 */
 	public static final String GP_PASSWORD_RESET_URL = "security.passwordResetUrl";
 
+	/**
+	 * Global property that stores the system salt used for deterministic bootstrap password generation.
+	 * This salt is used to derive bootstrap passwords from user UUIDs and must be kept secret.
+	 */
+	public static final String GP_BOOTSTRAP_SYSTEM_SALT = "security.bootstrap.systemSalt";
```
jwnasambu: Thanks @praisekavuma4 for this progress. Kindly remove this line of code. OpenmrsConstants.java defines GP_BOOTSTRAP_SYSTEM_SALT on this line and registers it as a core global property on line 751 but nothing in the codebase actually reads this property. The real secret is pepper comes from the runtime property openmrs.bootstrap.pepper (Security.java:145-148). This means if an admin sets security.bootstrap.systemSalt via the UI, they'll believe they've configured the bootstrap secret, but it has no effect.

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -69,6 +79,206 @@ public static boolean hashMatches(String hashedPassword, String passwordToHash)
...
+	 * Checks if a password matches a user's bootstrap password.
+	 * 
+	 * @param user the user
+	 * @param password the password to check
+	 * @return true if the password is the user's bootstrap password
+	 * @since 2.8.8
+	 */
+	public static boolean isBootstrapPassword(org.openmrs.User user, String password) {
+		return validateBootstrapPassword(user, password);
+	}
```
jwnasambu: Kindly `isBootstrapPassword` is an exact duplicate of validateBootstrapPassword. The UserService interface exposes both as separate public API methods (UserService.java:632 and UserService.java:654) and UserServiceImpl delegates both to the same static methods. This doubles the public API surface for no reason. Pick one name and remove the other.

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
praisekavuma4: Thank you @jwnasambu  here, i had made the changes but dint save while i was pushing the code.

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -69,6 +79,206 @@ public static boolean hashMatches(String hashedPassword, String passwordToHash)
...
+		}
+		
+		// If the user has already been forced to change, the bootstrap password is no longer valid
+		if (isBootstrapPasswordExpired(user)) {
+			return false;
+		}
+		
+		try {
+			String expected = generateBootstrapPassword(user);
+			return expected.equals(password);
```
jwnasambu: Kindly` expected.equals(password)` uses a regular string comparison, which can leak timing information. While the risk is lower here because the bootstrap password is regenerated rather than looked up from a database, consider using MessageDigest.isEqual(byte[], byte[]) or another constant-time comparison utility for consistency and defense in depth.

api/src/main/java/org/openmrs/util/Security.java
praisekavuma4: fixed well. thank you

api/src/main/java/org/openmrs/util/Security.java
praisekavuma4: @jwnasambu thank you.  i am sure it has now been solved

jwnasambu (commented):
@praisekavuma4 SonarCloud flagged a couple of issues on Security.java that need addressing before this can be merged:

Line 108–109 – Password hashing salt is unpredictable-flagged. Could you use a cryptographically secure random salt generator (e.g. SecureRandom in Java) instead of the current approach?
Line 32 – There's a duplicated import that should be removed.

Details/links from SonarCloud:

[Salt issue](https://sonarcloud.io/project/issues?id=openmrs_openmrs-core&issues=AZ-AOqQF-TchFAS-RJ1G&open=AZ-AOqQF-TchFAS-RJ1G&pullRequest=6355)
[Duplicate import](https://sonarcloud.io/project/issues?id=openmrs_openmrs-core&issues=AZ-FNdpp1YFO5dkuhfuN&open=AZ-FNdpp1YFO5dkuhfuN&pullRequest=6355)

api/src/test/java/org/openmrs/util/BootstrapPasswordTest.java
```diff
@@ -0,0 +1,136 @@
...
+    public void testIsBootstrapPasswordExpired_shouldReturnFalseWhenUserDoesNotHaveChangePasswordFlag() {
+        user.setUserProperty(OpenmrsConstants.USER_PROPERTY_CHANGE_PASSWORD, "false");
+        assertFalse(Security.isBootstrapPasswordExpired(user));
+    }
+    
+    @Test
+    public void testIsBootstrapPasswordExpired_shouldReturnFalseForNullUser() {
+        assertFalse(Security.isBootstrapPasswordExpired(null));
+    }
+}
```
jwnasambu: Nit: Could you please leave one blank line at the very end of the file, after the last line of code?

solomonfortune (commented):
Thanks for the work here — the crypto/service layer is solid, but comparing against the TRUNK-6664 acceptance criteria, a few things from the ticket aren't in this PR yet:
- Bootstrap password isn't actually used during initial user provisioning
- No admin action/endpoint to trigger a controlled reset
- No forced password change on first login (this is separate from the earlier login-wiring question — it's explicitly required by the ticket)
- No CLI tool for resetting a password this way
- Salt isn't auto-generated on install
Also, on the salt/pepper change: renaming `systemSalt` → `pepper` and moving it to runtime.properties didn't actually fix the SonarCloud finding — the salt passed into `PBEKeySpec` is still the hardcoded `PBKDF2_SALT` constant, unchanged. The pepper is just part of the input string now, not the salt itself. That check is still failing on the latest commit.
One more small thing: `GP_BOOTSTRAP_SYSTEM_SALT` is still defined and registered as a global property in `OpenmrsConstants.java`, but the code doesn't use it anymore now that the pepper comes from runtime.properties. Worth removing so there isn't a second, unused "secret" sitting in the admin UI.

I came up with this review yesterday.... Hope it does not conflict or make any repetitions in other reviews

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -41,6 +46,12 @@ public class Security {
 	
 	private static final Random RANDOM = new SecureRandom();
 
+	private static final String PBKDF2_ALGORITHM = "PBKDF2WithHmacSHA256";
+    private static final int PBKDF2_DEFAULT_ITERATIONS = 600000; 
+    private static final int PBKDF2_KEY_LENGTH = 256;
+    private static final int BOOTSTRAP_MAX_PASSWORD_LENGTH = 20;
+	private static final byte[] PBKDF2_SALT = "OpenMRS-Bootstrap-PBKDF2".getBytes(StandardCharsets.UTF_8);
```
jwnasambu: The pepper (read from runtime.properties) is concatenated into the PBKDF2 input string, but the salt passed to PBEKeySpec is still the hardcoded constant above. This is what SonarCloud is flagging. The comment on the generateDeterministicHash call says "Non-empty fixed salt" but PBKDF2 salt is supposed to be unpredictable per-password (or at minimum per-installation). The pepper provides per-installation variation, but a leaked pepper plus any UUID is a full account compromise and there's no way to rotate the pepper without invalidating every bootstrap password simultaneously. Please acknowledge this trust model explicitly or use a per-user salt (the UUID could serve as the PBKDF2 salt, since the pepper already provides the secret). You'd need to rename the field and update the generateDeterministicHash signature, but it's a cleaner design.

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -29,6 +29,11 @@
 import org.slf4j.LoggerFactory;
 import org.springframework.util.StringUtils;
 
+import java.security.MessageDigest;
```
jwnasambu: Kindly remove this import since its already imported at line 14.

praisekavuma4:
> Thanks for the work here — the crypto/service layer is solid, but comparing against the [TRUNK-6664](https://issues.openmrs.org/browse/TRUNK-6664) acceptance criteria, a few things from the ticket aren't in this PR yet:
> 
> * Bootstrap password isn't actually used during initial user provisioning
> * No admin action/endpoint to trigger a controlled reset
> * No forced password change on first login (this is separate from the earlier login-wiring question — it's explicitly required by the ticket)
> * No CLI tool for resetting a password this way
> * Salt isn't auto-generated on install
>   Also, on the salt/pepper change: renaming `systemSalt` → `pepper` and moving it to runtime.properties didn't actually fix the SonarCloud finding — the salt passed into `PBEKeySpec` is still the hardcoded `PBKDF2_SALT` constant, unchanged. The pepper is just part of the input string now, not the salt itself. That check is still failing on the latest commit.
>   One more small thing: `GP_BOOTSTRAP_SYSTEM_SALT` is still defined and registered as a global property in `OpenmrsConstants.java`, but the code doesn't use it anymore now that the pepper comes from runtime.properties. Worth removing so there isn't a second, unused "secret" sitting in the admin UI.
> 
> I came up with this review yesterday.... Hope it does not conflict or make any repetitions in other reviews

These are valid points @solomonfortune ,  the items you mentioned are planned as follow up tickets.

[TRUNK-6664]: https://openmrs.atlassian.net/browse/TRUNK-6664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

api/src/main/java/org/openmrs/util/Security.java
praisekavuma4: @jwnasambu  i have been able to fix these areas

api/src/test/java/org/openmrs/util/BootstrapPasswordTest.java
```diff
@@ -0,0 +1,138 @@
+package org.openmrs.util;
```
dkayiwa, line 1: This file has no MPL header, and that is what is turning CI red: `license-maven-plugin:check` fails the `openmrs-api` module on it, so `openmrs-web`, `openmrs-webapp` and everything after it are skipped and the build never gets as far as running tests. I reproduced it locally on f87d9d4b2:

```
[WARNING] Missing header in: .../api/src/test/java/org/openmrs/util/BootstrapPasswordTest.java
[INFO] openmrs-api ........................................ FAILURE
[ERROR] Failed to execute goal com.mycila:license-maven-plugin:3.0:check
```

and the same check passes clean on 2.8.x at 50f481786, so this file is the whole of it. `mvn license:format -pl api` will add the header, or paste it in here:

```suggestion
/**
 * This Source Code Form is subject to the terms of the Mozilla Public License,
 * v. 2.0. If a copy of the MPL was not distributed with this file, You can
 * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
 * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
 *
 * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
 * graphic logo is a trademark of OpenMRS Inc.
 */
package org.openmrs.util;
```

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -822,4 +822,40 @@ public void changePasswordUsingActivationKey(String activationKey, String newPas
...
+	/**
+     * @see org.openmrs.api.UserService#forcePasswordChange(User)
+     */
+    @Override
+    @Authorized( { PrivilegeConstants.EDIT_USER_PASSWORDS })
+    public void forcePasswordChange(User user) {
+        if (user != null) {
+            Security.forcePasswordChange(user); // Sets the property
+            dao.saveUser(user, null);
+        }
```
dkayiwa, line 851: Going straight to the DAO here skips the guard that stops a caller handing out privileges they do not hold. `UserService.saveUser` requires `Edit Users` and then runs `checkPrivileges(user)` and `checkSuperUserPrivilege(role)`; `dao.saveUser` runs neither, but it still persists the whole detached `User` graph, roles included. So if this merges as is, anyone holding only `Edit User Passwords` can promote any account to System Developer: read the user, `addRole(superuser)`, call `forcePasswordChange(user)`.

I ran that on f87d9d4b2 as `butch` (a Provider) with just `Edit User Passwords` proxied in, targeting `bruno`:

```
PROBE caller=butch hasEDIT_USERS=false hasASSIGN_SYS_DEV=false
PROBE forcePasswordChange RETURNED NORMALLY
PROBE saveUser THREW APIAuthenticationException: Privileges required: Edit Users
PROBE afterRoles=[Provider, System Developer] afterSuperuser=true
```

Same object, same session: `saveUser` refuses the write and `forcePasswordChange` performs it. Nothing throws and nothing is logged, so there is no trace of it afterwards either.

Routing the save back through the service fixes it:

```suggestion
        if (user != null) {
            Security.forcePasswordChange(user);
            Context.getUserService().saveUser(user);
        }
```

`setUserProperty(User, String, String)` is the other option, since it already does both halves of this (sets the property, saves through `saveUser`, checks authorization itself). Either way the effective requirement becomes `Edit Users`, which is a deliberate decision to make, but it should not end up below what `saveUser` demands for the same write.

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -69,6 +78,211 @@ public static boolean hashMatches(String hashedPassword, String passwordToHash)
...
+     * 
+     * Reads from global property "security.bootstrap.iterations".
+     * Falls back to default (600,000) if not configured or invalid.
+     * 
+     * @return the number of iterations (always > 0)
+     * @since 2.8.8
+     */
+    public static int getBootstrapIterations() {
+        try {
+            String value = Context.getAdministrationService()
```
dkayiwa, line 206: Reading the iteration count from a global property makes the derivation depend on who is asking. `AdministrationService.getGlobalProperty` is `@Authorized(GET_GLOBAL_PROPERTIES)`, and the `catch (Exception e)` below swallows the resulting `APIAuthenticationException` at debug level and falls back to 600000. So the same user gets a different bootstrap password depending on the caller's privileges. On f87d9d4b2 with the property set to 1000, generating for user 1:

```
PROBE as-superuser iterations=1000
PROBE as-superuser password=EXPGruhFx39irq451Aon
PROBE as-limited-user has GET_GLOBAL_PROPERTIES=false
PROBE as-limited-user iterations=600000
PROBE as-limited-user password=XcR3GfGgDibZBk3TDjCP
PROBE validate(adminPassword) as limited user=false
```

Merged as is, an admin hands out a bootstrap password and whatever checks it rejects it, or the reverse, with nothing above debug level to explain why. `Get Users` is all it takes to reach `generateBootstrapPassword`; `Get Global Properties` is a separate privilege that plenty of callers will not hold. The same root cause bites even where every caller is equally privileged, because the property is editable from the admin UI: changing it silently invalidates every bootstrap password already issued.

The iteration count needs to be a fixed input to the derivation rather than a mutable, privilege-gated one. Reading it from runtime.properties alongside the pepper keeps it configurable per installation without either problem, and it lets the blanket `catch (Exception e)` go. Whichever way you take it, this method has no test coverage at all right now, so it would be good to pin the chosen behaviour down with one.

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -605,4 +605,53 @@ List<User> getUsers(String name, List<Role> roles, boolean includeRetired, Integ
...
+	 * 
+	 * The password is derived from the user's UUID and a system-wide salt using PBKDF2.
+	 * It is not stored in the database and can be regenerated at any time.
+	 * 
+	 * @param user the user for whom to generate the bootstrap password
+	 * @return the generated bootstrap password
+	 * @throws APIException if the user is null, has no UUID, or system salt is not configured
+	 * @since 2.8.8
+	 */
+	@Authorized( { PrivilegeConstants.GET_USERS })
```
dkayiwa, line 620: `Get Users` is a read privilege, and this method returns a working credential derived from the system pepper, so anyone who can list users can compute a bootstrap password for every account including admin. This annotation is the real gate rather than documentation: `AuthorizationAdvice` resolves `@Authorized` off the interface method, which I confirmed in the same test setup by watching a caller without `Edit Users` get `APIAuthenticationException: Privileges required: Edit Users` out of `saveUser`. So `Get Users` is exactly and only what this method will ask for.

Nothing in core authenticates with these yet, so the door is not open today. But a bootstrap password is a pure function of UUID plus pepper and never changes, so anything harvested now starts working the moment the login wiring you mentioned lands, and the only way to revoke it is rotating the pepper for everybody. The privilege on a released interface is also the contract downstream modules will code against, so it is much cheaper to get right before 2.8.9 ships than to tighten afterwards.

`Edit User Passwords` is what `changePassword` requires for a strictly weaker capability, and what `forcePasswordChange` already requires two methods below:

```suggestion
	@Authorized( { PrivilegeConstants.EDIT_USER_PASSWORDS })
```

`isBootstrapPasswordExpired` is a genuine read and can stay where it is.

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -605,4 +605,53 @@ List<User> getUsers(String name, List<Role> roles, boolean includeRetired, Integ
...
+	/**
+	 * Validates a password against a user's bootstrap password.
+	 * 
+	 * @param user the user
+	 * @param password the password to validate
+	 * @return true if the password matches the user's bootstrap password
+	 * @since 2.8.8
+	 */
+	@Authorized( { PrivilegeConstants.GET_USERS })
+	boolean validateBootstrapPassword(User user, String password);
```
dkayiwa, line 632: The password argument reaches the logs here. `LoggingAdvice` prints every non-ignored argument at TRACE on the `org.openmrs.api` logger, which is why the other password-taking declarations in this interface carry `@Logging(ignoredArgumentIndexes = ...)`. With TRACE enabled I get both of these:

```
In method UserService.validateBootstrapPassword. Arguments: User=bruno, String=2i7c5Q5N6CKM26VUKW1d,
In method UserService.changePassword. Arguments: User=bruno, String=<Arg value ignored>, String=<Arg value ignored>,
```

The annotation has to go on this interface declaration rather than on the impl, since `LoggingAdvice` resolves it from the interface method:

```suggestion
	@Logging(ignoredArgumentIndexes = { 1 })
	boolean validateBootstrapPassword(User user, String password);
```

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -69,6 +78,211 @@ public static boolean hashMatches(String hashedPassword, String passwordToHash)
...
+		}
+		
+		String uuid = user.getUuid();
+		if (!StringUtils.hasText(uuid)) {
+			throw new APIException("bootstrap.user.uuid.missing", (Object[]) null);
+		}
+		
+		String pepper = getBootstrapPepper();
+		if (!StringUtils.hasText(pepper)) {
+			throw new APIException("bootstrap.pepper.missing", (Object[]) null);
```
dkayiwa, line 111: None of the five new keys exist in `api/src/main/resources/messages.properties`: `bootstrap.user.null`, `bootstrap.user.uuid.missing`, `bootstrap.pepper.missing`, `bootstrap.input.null`, `bootstrap.hash.generation.failed`. The message source is built with `setUseCodeAsDefaultMessage(true)` (`OpenmrsApplicationContextConfig.java:104`), so rather than throwing, it hands back the key itself: an admin who has not set the pepper sees the literal text `bootstrap.pepper.missing`, which does not tell them the property is `openmrs.bootstrap.pepper` in runtime.properties. Adding the five entries sorts it out.

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -69,6 +78,211 @@ public static boolean hashMatches(String hashedPassword, String passwordToHash)
...
+     * 
+     * This method is used for bootstrap password generation, where deterministic output is required
+     * for a given user (same user → same password). The UUID serves as a per-user salt, while the
+     * pepper (system secret) is concatenated into the input string. This ensures that even if the
+     * pepper is compromised, each user's password remains unique due to the per-user salt.
+     * 
+     * Unlike {@link #encodeString(String)}, which uses SHA-512, this method uses PBKDF2 with a high
+     * iteration count, making it suitable for password derivation.
+     * 
+     * @param input      the input string to derive from (typically user UUID + system pepper)
```
dkayiwa, line 92: These three `@param` tags describe the three-arg `generateDeterministicHash`, not `generateBootstrapPassword`, which takes a single `User`. The whole comment block looks like it was copied down from that method, so the javadoc published for this one documents parameters it does not have and says nothing about the one it does.

Two smaller things in the same block. `@since` should be 2.8.9, not 2.8.8, throughout the new code here and in `UserService`: 2.8.8 was tagged on 2026-07-09 and this branch is on 2.8.9-SNAPSHOT. And the indentation through the new methods mixes spaces with the tabs the rest of the file uses; checkstyle only sits in `pluginManagement` in the parent pom, so nothing is going to tidy that up on its own.

api/src/main/java/org/openmrs/util/Security.java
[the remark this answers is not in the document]
dkayiwa: `forcePassword` will not hold that line, though. It is the general-purpose "must change password at next login" flag, and every upstream flow clears it once the change succeeds. legacyui's `ChangePasswordFormController` does `new UserProperties(currentUser.getUserProperties()).setSupposedToChangePassword(false)` right after `changePassword`, and `false` there means *remove the property* (`UserProperties.java:43-49`); authenticationui's `ChangePasswordPageController` sets it to `"false"`. Its own test pins that behaviour down ("set the user property forcePassword to false after successful password change"). So the sequence is: bootstrap login, `forcePasswordChange` sets the flag, the user changes their password, the flag is cleared, `isBootstrapPasswordExpired` goes back to false, and the bootstrap password validates again. Permanently, since it is still a pure function of UUID plus pepper.

It reads wrong in the other direction too. adminui's `AccountPageController` defaults `forceChangePassword` to true for every new account (lines 241-244 and 325-326), legacyui's user form exposes it as a checkbox on every save, and authenticationui's admin-initiated reset ships that checkbox pre-checked. Plenty of users therefore carry `forcePassword=true` having never had a bootstrap password issued, and this method reports their nonexistent bootstrap password as expired.

Expiry needs state of its own that only moves one way. A distinct user property recording that the bootstrap password was consumed would do it, or a per-user generation counter folded into the derivation so that consuming one changes the value that gets derived next time. Worth a test on whichever you pick, since nothing currently asserts that `validateBootstrapPassword` returns false for an expired user.

praisekavuma4:
@dkayiwa i have made the necessary fixes, kindly verify if there is no other.

dkayiwa:
@claude review

praisekavuma4:
@dkayiwa and @ibacher kindly verify these fixes

jwnasambu:
Hey @dkayiwa @ibacher, could you kindly approve or re-run the 'Build with Maven' workflow when you get a chance?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -826,4 +826,41 @@ public void changePasswordUsingActivationKey(String activationKey, String newPas
...
+	 * @see org.openmrs.api.UserService#forcePasswordChange(User)
+	 */
+	@Override
+	@Authorized({ PrivilegeConstants.EDIT_USER_PASSWORDS })
+	public void forcePasswordChange(User user) {
+		if (user != null) {
+			Security.forcePasswordChange(user);
+			Context.getUserService().saveUser(user);
+		}
+	}
```
dkayiwa, line 854: The privilege this declares is lower than the one it actually needs. `saveUser` requires `Edit Users`, so a caller holding exactly the `Edit User Passwords` that the annotation and the javadoc ask for gets an exception out of the middle of the call instead of the operation they were told they could perform.

Same caller, same session, same target on 6069109e2, authenticated as `test_user` with only `Edit User Passwords` proxied in:

```
PROBE caller=test_user EDIT_USER_PASSWORDS=true EDIT_USERS=false
PROBE changePassword(target, "Bootstrap123")  -> RETURNED NORMALLY
PROBE forcePasswordChange(target)             -> APIAuthenticationException: Privileges required: Edit Users
```

So changing somebody's password outright goes through, and merely flagging it for change does not, which is the wrong way round. `Security.forcePasswordChange(user)` has also already run by the time it throws, so the caller is handed back a `User` carrying `forcePassword=true` and `bootstrapPasswordExpired=true` while the database has neither.

I would keep `Edit User Passwords` as the gate, since that is what `changePassword(User, String)` asks for the stronger capability, and proxy the inner save the way `setUserProperty` in this same class already does:

```suggestion
	public void forcePasswordChange(User user) {
		if (user != null) {
			Security.forcePasswordChange(user);
			try {
				Context.addProxyPrivilege(PrivilegeConstants.EDIT_USERS);
				Context.getUserService().saveUser(user);
			}
			finally {
				Context.removeProxyPrivilege(PrivilegeConstants.EDIT_USERS);
			}
		}
	}
```

I applied exactly that and re-ran the probe, and both calls then return normally. Declaring `Edit Users` on the `@Authorized` is the other way out of it, but that makes forcing a password change cost more privilege than performing one.

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -109,6 +118,212 @@ public static boolean hashMatches(String hashedPassword, String passwordToHash)
...
+				input.toCharArray(),
+				salt.getBytes(StandardCharsets.UTF_8),
+				actualIterations,
+				PBKDF2_KEY_LENGTH
+			);
+			byte[] derived = factory.generateSecret(spec).getEncoded();
+
+			String base64 = Base64.getEncoder().withoutPadding().encodeToString(derived);
+			String cleaned = base64.replaceAll("[0OIl]", "");
+			return cleaned.substring(0, Math.min(cleaned.length(), BOOTSTRAP_MAX_PASSWORD_LENGTH));
```
dkayiwa: `0`, `O`, `I` and `l` are stripped here so the result can be read off a screen and typed back in, but `Base64.getEncoder()` also emits `+` and `/`, and those stay. Over 5000 generations, 2460 of them (49.2%) contain one, e.g. `E+xxj+UCbx8x/bsRxQMW`.

Digits are the other half of the same question. 196 of those 5000 (3.92%) are rejected outright by `OpenmrsUtil.validatePassword`, which is what `UserServiceImpl.updatePassword` runs before storing any password: `security.passwordRequiresDigit` is seeded `true` in core data, and a 20 character sample of this alphabet has no digit about one time in twenty five.

Neither of those bites today, since nothing issues these yet, and that is the argument for settling it now rather than later. The alphabet is an input to the derivation, so once installations start handing bootstrap passwords out, changing it invalidates every one still outstanding. For the readable-characters half I would just widen the filter to `[0OIl+/]`; I ran 200000 derivations that way and the shortest post-filter string was 29 characters, so the 20 character truncation stays safe. The digit half only matters if a bootstrap password is ever meant to reach `changePassword`, so it is worth deciding which of the two you want while the derivation is still free to change.

jwnasambu:
@dkayiwa and @ibacher kindly help review this PR at your convenient time please!

jwnasambu:
@dkayiwa and @ibacher kindly help review this PR at your convenient time please!

jwnasambu:
@claude review

jwnasambu:
@dkayiwa and @ibacher kindly help review this PR at your convenient time please!

PR #6363, opened 2026-07-21
TRUNK-6711: Batch-load search result pages

ibacher:

When using the search index to retrieve items, we previously hydrated the results by loading the corresponding rows 1-by-1 from the database, creating a bottle-neck. This reworks these queries to bulk-load items from the database with some additional handling for nulls.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6711

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
@@ -799,7 +799,7 @@ public class HibernatePatientDAO implements PatientDAO {
 			    b.filter(f.match().field("patient.isPatient").matching(true));
 			    b.filter(f.match().field("voided").matching(false));
 		    }).toPredicate(), "patient.personId",
-		        (List<PatientIdentifier> pIs) -> multiLoadPatients(pIs, pI -> pI.getPatient().getPatientId())),
+		        (List<PatientIdentifier> pIs) -> multiLoadPatients(pIs, pI -> pI.getPatient().getPersonId())),
 		    tmpStart, tmpLength);
 	}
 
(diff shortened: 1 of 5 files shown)


ibacher:
@claude review

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -821,16 +828,22 @@ public List<Patient> findPatients(String query, boolean includeVoided, Integer s
...
-		                    f -> personQuery.getPatientAttributeQuery(f, query, includeVoided), "person.personId",
-		                    pA -> getPatient(pA.getPerson().getId())))),
-		    start, length);
+		patients = SearchQueryUnique
+		        .search(searchSessionFactory,
+		            SearchQueryUnique
+		                    .newQuery(PatientIdentifier.class,
+		                        f -> newPatientIdentifierSearchPredicate(f, query, includeVoided, false), "patient.personId",
+		                        (List<PatientIdentifier> pIs) -> multiLoadPatients(pIs,
+		                            pI -> pI.getPatient().getPatientId()))
```
dkayiwa: I profiled a 25-hit page on H2 with Hibernate statistics (prepared-statement counts around each DAO call), this branch against master. The name and attribute paths get the full win: `findPatients` on a name went from 116 statements to 20, and `getPeople` now hands back initialized persons in a bounded number of selects instead of lazy proxies that callers then initialize one at a time.

This identifier sub-query stays flat though (44 to 45; the concept searches similarly 11 to 12): `PatientIdentifier.patient` and `ConceptName.concept` are default-eager `@ManyToOne`s, so hit loading already fetches each patient one by one while hydrating the `PatientIdentifier` hits, before the mapper runs, and the `findMultiple` here mostly re-selects rows already in the session. Nothing regresses, so I don't think this needs to hold the PR, but identifier lookup is arguably the hottest search TRUNK-6711 is about, so it seems worth a follow-up to let the batch loader actually kick in here, either an entity graph in the hit-loading options (probably the safer first step) or making those associations lazy.

If `patient` does go lazy at some point, this extractor will need `pI.getPatient().getPersonId()` instead: `getPatientId()` is a plain property (`Patient` is a joined subclass of `Person`, whose id property is `personId`), so on an uninitialized proxy it triggers the very per-hit load this batching is meant to avoid, while the id getter is proxy-safe. Same value either way.

ibacher: FWIW, the main target here is search by name, which is, I think, the most used search in the OMRS ecosystem and which this substantially improves. I see what you're saying about laziness (and, indeed, as part of this series, I opened [TRUNK-6680](https://openmrs.atlassian.net/browse/TRUNK-6680) to restore some laziness to the models, albeit in places we accidentally dropped it).

I don't think there's a strong reason to make `PatientIdentifier.patient` lazy, at least as a default. It's a net gain _here_, but it's not clearly one in other places, i.e., existing code may depend on `pi.getPatient()` returning a hydrated object without running a new query, since that's been the behaviour forever.

What might be possible to make improvements for these other searches is something like [TRUNK-6729](https://openmrs.atlassian.net/browse/TRUNK-6729), which suggests two avenues to explore for ensuring that we're lazy-loading `PatientIdentifier.patient` and `ConceptName.concept` in the context of these bulk-search queries only.

PR #6364, opened 2026-07-21
mergePatients does not reassign Conditions, Allergies, or MedicationDispenses to the surviving patient

ibacher:

This is a backport to 2.8.x of #6198, fixing #6194. Including this backport for discussion since it's a clear correctness issue.

## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -27,8 +27,10 @@ import org.openmrs.Allergies;
 import org.openmrs.Allergy;
 import org.openmrs.BaseOpenmrsMetadata;
 import org.openmrs.Concept;
+import org.openmrs.Condition;
 import org.openmrs.Encounter;
 import org.openmrs.Location;
+import org.openmrs.MedicationDispense;
 import org.openmrs.Obs;
 import org.openmrs.Order;
 import org.openmrs.Patient;
@@ -44,9 +46,11 @@ import org.openmrs.User;
 import org.openmrs.Visit;
 import org.openmrs.api.APIException;
 import org.openmrs.api.BlankIdentifierException;
+import org.openmrs.api.ConditionService;
 import org.openmrs.api.DuplicateIdentifierException;
 import org.openmrs.api.EncounterService;
 import org.openmrs.api.InsufficientIdentifiersException;
+import org.openmrs.api.MedicationDispenseService;
 import org.openmrs.api.MissingRequiredIdentifierException;
 import org.openmrs.api.ObsService;
 import org.openmrs.api.PatientIdentifierException;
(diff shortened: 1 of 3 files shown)


ibacher:
@dkayiwa Two questions:

1. Do you think backporting to 2.8 is a good idea? I'm inclined to do so because I know DRC is heavily reliant on the patient-merge functionality, but this is also a behavioural change that I'd usually be suspicious of.
2. I see we have a GitHub issue. Do we have a corresponding TRUNK ticket for records sake?

dkayiwa (commented):
On the two questions:

1. I would take the backport, since what it fixes is silent clinical data loss on merge, but with eyes open about one behavioural edge. The three helpers are identical to what 2.9.x has today (including the Get Allergies proxy block that the #6324 backport made necessary), so this is a faithful port. The catch is the privilege envelope of mergePatients on a patch line. I tried the merge as a non-superuser holding the privilege set that is sufficient on current 2.8.x (our MERGE_PRIVILEGES test list minus the five this change starts using): on the 2.8.x head the merge passes, while on this branch every merge fails up front with "Privileges required: Get Conditions", even for patients that have no conditions at all, because the condition and dispense lookups run unconditionally. With actual data on the losing patient, Edit Conditions, Edit Allergies (or Add Allergies) and Edit Medication Dispense are demanded as well. The failure is at least loud and clean: mergePatients is a single transaction, so it rolls back rather than half-merging. But any non-superuser merge role that predates these privileges (the Medication Dispense pair only exists since 2.6) will fail every merge after upgrading to 2.8.9 until the role is edited. So my answer is yes, backport, but with release notes that spell out what a merge-capable role now needs: Get Conditions, Edit Conditions, Edit Allergies, Get Medication Dispense, Edit Medication Dispense. That matches the position we took in the #6324 merge test, that these are privileges a merging user should legitimately hold. The alternative is to proxy the reads and writes in these three helpers the way the allergy reads are (the two Copilot comments point at the read half of that); if we prefer that for the patch line, I would first make the same change on 2.9.x and master and keep this branch identical rather than diverge here.

2. I did not find a TRUNK ticket for this one. The closest is TRUNK-6641, which is the copy-instead-of-move design question these helpers reference in their TODOs. I can create one for the record.

One small thing: the 2.9.x commit also added the new "Should" lines to the mergePatients javadoc in PatientService; carrying those over keeps the interface doc and future branch diffs aligned.

ibacher:
In the future, can you break up the text a bit more. The prose in that format is borderline unreadable.

> The catch is the privilege envelope of mergePatients on a patch line.

I think this is mischaracterizing things greatly. This is (apart from the should comments which are of questionable value) a straight-forward port of the code. The problem this is highlighting _also_ exists in Core 3.0.0 and Core 2.9.0. If the solution is to add new proxy privileges, then we should do that there and backport it here.

> I did not find a TRUNK ticket for this one

I was asking if you _created_ one for this.

PR #6365, opened 2026-07-21
TRUNK-6709: Remove internal copy amplification in Concept name resolution

Twiineenock:

Rework the internal call graph in `Concept.java` so name resolution iterates the underlying `names` collection directly with voided/locale predicates instead of building intermediate copies.

**What was added:**
- Four private stream helpers (`nonVoidedNames()`, `nonVoidedNamesIn(Locale)`, `partiallyCompatibleNamesFor(Locale)`, `hasNonVoidedNames()`) that iterate the raw `names` field without allocating collections.

**What was changed:**
- 15 internal methods (`getName()`, `getPreferredName()`, `getFullySpecifiedName()`, `getNameInLocale()`, `getShortNameInLocale()`, `getShortestName()`, `findNameTaggedWith()`, `hasName()`, `isNamed()`, `setPreferredName()`, `setFullySpecifiedName()`, `setShortName()`, `getName(Locale, boolean)`, `getName(Locale, type, tag)`, `getCompatibleNames()`) now use the private stream helpers instead of the public copy-producing accessors.

**What was removed:**
- `getPartiallyCompatibleNames(Locale)` — dead code after the refactoring.

**What was NOT changed:**
- All public collection-returning accessors (`getNames()`, `getNames(Locale)`, `getSynonyms()`, `getSynonyms(Locale)`, `getShortNames()`, `getIndexTerms()`) still return defensive copies.
- All resolution behavior, including locale fallback (country/variant stripping) and preferred/fully-specified/synonym ordering.

**Result:** A single `getName()` call that previously created ~75 throwaway HashSets across 3 allowed locales now creates zero intermediate collections. Net change: 139 insertions, 144 deletions (-5 lines).

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6709

- [ ] I have **added tests** to cover my changes. (If you refactored existing code that was well tested you do not have to add tests)

### DIFF
--- a/api/src/main/java/org/openmrs/Concept.java
+++ b/api/src/main/java/org/openmrs/Concept.java
@@ -16,10 +16,12 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


Twiineenock:
CC @ibacher , @dkayiwa

dkayiwa:
@claude review

jwnasambu (commented):
Thanks @Twiineenock for the fix. This LGTM. 

Nit: SonarCloud flags ObjectUtils.nullSafeEquals(n.getLocalePreferred(), true) as unnecessary boxing -- getLocalePreferred() returns a Boolean wrapper, and comparing it against the primitive true triggers the R550 warning. Since LocalePreferred is a nullable Boolean field mapped by Hibernate, the safest fix is to use Boolean.TRUE.equals(n.getLocalePreferred()) which is null-safe and avoids the boxed comparison. This applies to the two occurrences in getPreferredName and the one in getName(Locale, ConceptNameType, ConceptNameTag). Though not a blocker but I belief it will be good to fix it.

ibacher (approved):
Thanks @Twiineenock! This was a real improvement! I discovered when reviewing this that we can squeeze slightly more speed out by using for loops and some duplication of checks, simply because there's no need to allocate any objects to do so, so I've somewhat modified the final version of this.

PR #6371, opened 2026-07-23
TRUNK-6714 ->  Fix operator precedence bug in getRefByUuid() exception message

Kapil-kundu:
The **getRefByUuid()** method contains an operator precedence issue in the **APIException** message construction:

`throw new APIException("Unsupported type for getRefByUuid: " + type != null ? type.getName() : "null");`

Due to Java operator precedence, the ternary operator is evaluated incorrectly, causing the exception message to be wrong and potentially leading to a **NullPointerException** when **type** is **null**.

**Proposed Fix**
Wrap the ternary expression in parentheses:

`throw new APIException(
    "Unsupported type for getRefByUuid: "
        + (type != null ? type.getName() : "null")
);`

**Impact**

- Generates the correct error message.
- Prevents incorrect evaluation of the ternary expression.
- Avoids a possible NullPointerException when type is null.

**Tests**
-All test cases passed

JIra Ticker : https://openmrs.atlassian.net/browse/TRUNK-6714

### DIFF
--- a/api/src/test/java/org/openmrs/api/impl/ObsServiceImplTest.java
+++ b/api/src/test/java/org/openmrs/api/impl/ObsServiceImplTest.java
@@ -1,0 +1,33 @@
+
+import org.junit.jupiter.api.Test;
+import org.openmrs.api.APIException;
+import org.openmrs.test.jupiter.BaseContextSensitiveTest;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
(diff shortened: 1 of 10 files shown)


Kapil-kundu:
@jwnasambu review required

jwnasambu:
@claude review

api/src/test/java/org/openmrs/api/impl/ObsServiceImplTest.java
```diff
@@ -0,0 +1,33 @@
+
```
jwnasambu, line 2: @Kapil-kundu Thanks for the progress on this task. Kindly add the missing package` org.openmrs.api.impl;` declaration on line 1. java requires a package declaration before import statements. Without it the compiler treats the class as being in the default package, causing the build to fail. All other test files in this PR (OrderServiceImplTest.java, PersonServiceImplTest.java, etc.) correctly include package org.openmrs.api.impl; — this one was overlooked.

Kapil-kundu:
@ibacher Can you please further review this PR

PR #6372, opened 2026-07-23
TRUNK-6712: Cache flattened role privileges for privilege and superuser checks

Rocker810:
<!--- 'TRUNK-6712: Cache flattened role privileges for privilege and superuser checks' -->
  Cache each role's fully-flattened privilege set (keyed by role name) so that                                                                                                       
  `UserContext.hasPrivilege` performs a cheap lookup instead of rebuilding the                                                                                                       
  user's entire expanded role tree and string-scanning privileges on every call.                                                                                                     
                                                                                                                                                                                     
  - Register a `rolePrivileges` cache in `cache-api.yaml`, reusing the shared                                                                                                        
    `entity` template so it inherits TTL expiry and clustered invalidation.                                                                                                          
  - Add an immutable, `Serializable` `RolePrivileges` value object (lowercased                                                                                                       
    privilege set + `grantsSuperuser` flag).                                                                                                                                         
  - Add a `RolePrivilegeCache` `@Component` that flattens a role's inheritance                                                                                                       
    chain once on a cache miss (with a cycle guard) and reads from the cache                                                                                                         
    thereafter. It is a plain component, so the read path does not re-enter the                                                                                                      
    `@Authorized` service proxy.                                                                                                                                                     
  - Rewrite `UserContext.hasPrivilege` to check each of the user's direct roles'                                                                                                     
    cached, fully-flattened entries. Behaviour is preserved: case-insensitive                                                                                                        
    matching, inherited roles, inherited superuser status, proxy privileges, and                                                                                                     
    the anonymous/authenticated roles.                                                                                                                                               
  - Evict the cache via `@CacheEvict(allEntries = true)` on `saveRole`,                                                                                                              
    `purgeRole`, `savePrivilege`, and `purgePrivilege`.                                                                                                                              
  - Add unit tests (population, cache hit, inherited superuser, inheritance                                                                                                          
    cycles) and a context-sensitive test proving eviction fires through the                                                                                                          
    service.  

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-6712

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/UserContext.java
+++ b/api/src/main/java/org/openmrs/api/context/UserContext.java
@@ -394,53 +394,53 @@ public class UserContext implements Serializable {
 	 * @return true if authenticated user has given privilege
 	 */
 	public boolean hasPrivilege(String privilege) {
-		log.debug("Checking '{}' against proxies: {}", privilege, proxies);
-		// check proxied privileges
+		boolean authorized = isAuthorized(privilege);
+		notifyPrivilegeListeners(getAuthenticatedUser(), privilege, authorized);
+		return authorized;
+	}
+
+	/**
+	 * Resolves whether the current user has the given privilege, checking, in order: proxy privileges,
+	 * the authenticated user's own roles, the authenticated role, and the anonymous role.
+	 */
+	private boolean isAuthorized(String privilege) {
 		for (String s : new ArrayList<>(proxies)) {
 			if (s.equals(privilege)) {
-				notifyPrivilegeListeners(getAuthenticatedUser(), privilege, true);
 				return true;
 			}
 		}
 
-		// grab the cache component (a plain component lookup, NOT a service proxy)
+		// a plain component lookup, so the read path does not re-enter the service proxy
 		RolePrivilegeCache cache = Context.getRegisteredComponent("rolePrivilegeCache", RolePrivilegeCache.class);
 		String normalizedPrivilege = (privilege == null) ? null : privilege.toLowerCase(Locale.ENGLISH);
 
-		// if a user has logged in, check their own roles plus the authenticated role
 		if (isAuthenticated()) {
-			User authenticatedUser = getAuthenticatedUser();
-
-			// keep the old rule: an empty privilege string always passes for logged-in users
+			// an empty privilege always passes for logged-in users
 			if (StringUtils.isEmpty(privilege)) {
-				notifyPrivilegeListeners(authenticatedUser, privilege, true);
 				return true;
 			}
-
-			if (authenticatedUser.getRoles() != null) {
-				// the user's DIRECT roles; each role's cached entry already contains its inherited closure
-				for (Role role : authenticatedUser.getRoles()) {
-					if (grants(cache, role, normalizedPrivilege)) {
-						notifyPrivilegeListeners(authenticatedUser, privilege, true);
-						return true;
-					}
-				}
-			}
-
-			if (grants(cache, getAuthenticatedRole(), normalizedPrivilege)) {
-				notifyPrivilegeListeners(authenticatedUser, privilege, true);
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -392,35 +394,65 @@ public Set<Role> getAllRoles(User user) throws Exception {
...
+			if (userRolesGrant(cache, normalizedPrivilege) || grants(cache, getAuthenticatedRole(), normalizedPrivilege)) {
+				return true;
+			}
 		}
 
-		if (getAnonymousRole().hasPrivilege(privilege)) {
-			notifyPrivilegeListeners(getAuthenticatedUser(), privilege, true);
-			return true;
-		}
+		return grants(cache, getAnonymousRole(), normalizedPrivilege);
```
dkayiwa, line 427: Before this change the authenticated-role check here (line 422) and the anonymous-role check on this line went through `Role.hasPrivilege`, which by its own contract only looks at that role's direct privileges — its Javadoc says it "does not recurse through the inherited roles." `grants(...)` consults the flattened cache, whose entry for a role is its full inherited closure. So if the Authenticated or Anonymous role is ever configured to inherit another role, this now grants that parent role's privileges where the old code did not, and for the Anonymous branch that means unauthenticated callers pick them up. The user's own roles aren't affected (that path already went through `User.hasPrivilege`/`getAllRoles`, which does include inheritance), so the widening is specific to these two role checks.

The PR description says behaviour is preserved, so is this widening intended? It reads more like a reasonable fix than a regression (arguably inherited privileges *should* count), but it is a behaviour change on the authorization path and worth calling out explicitly. In core's default config neither role inherits anything, so the real-world trigger is low.

Rocker810: this is a real behavior change. The user's-own-roles path already recursed via `getAllRoles()`, but the Authenticated and Anonymous checks went through                  
  `Role.hasPrivilege`, which by its `Javadoc` doesn't recurse inherited roles. The new `grants(...)` path uses the flattened cache (full inherited closure), so if either role ever has a parent, we now grant its privileges — including to unauthenticated callers on the Anonymous branch. Low real-world trigger (neither has a parent in default config), but it contradicts the "behavior preserved" claim in the PR description.                                                                                                                                                                                                                                                                                                       
  How do you want me to resolve it?                                                                                                                                                     
  Keep those two checks non-recursive → pure perf change, parity preserved or Keep the widening as an intentional consistency fix → update the PR description to call it out.

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -392,35 +394,65 @@ public Set<Role> getAllRoles(User user) throws Exception {
...
-		// if a user has logged in, check their privileges
-		if (isAuthenticated()
-		        && (getAuthenticatedUser().hasPrivilege(privilege) || getAuthenticatedRole().hasPrivilege(privilege))) {
-
-			// check user's privileges
-			notifyPrivilegeListeners(getAuthenticatedUser(), privilege, true);
-			return true;
+		// a plain component lookup, so the read path does not re-enter the service proxy
+		RolePrivilegeCache cache = Context.getRegisteredComponent("rolePrivilegeCache", RolePrivilegeCache.class);
+		String normalizedPrivilege = (privilege == null) ? null : privilege.toLowerCase(Locale.ENGLISH);
```
ibacher, line 415: I don't think we should be normalizing privileges. OpenMRS privileges are meant to be case-sensitive.

ibacher:
@Rocker810 Thanks for this! I realized when reviewing this that there are a bunch of subtle issues I didn't fully lay out in the ticket, having to do with accidentally having two privilege resolution mechanisms that disagree slightly and the need to ensure that all privileges are resolved against the DB regardless of the users privilege level. I'm going to take over this ticket, but thank you for the contribution. My version will be based on this.

PR #6375, opened 2026-07-23
TRUNK-6719: Fix reference-equality String comparison in NameSupport u…

Mahesh5f4:

- Replaced the reference-based `String` comparison in `NameSupport.updateLayoutTemplates()` with a null-safe value comparison using `Objects.equals()`.
- Added a unit test to verify that an existing template is replaced when a new template with the same code name (equal value but different `String` reference) is added.

## Issue I worked on

TRUNK-6719

https://issues.openmrs.org/browse/TRUNK-6719


jwnasambu (commented):
So sorry @Mahesh5f4, kindly fix the merge conflicts on this PR

PR #6379, opened 2026-07-23
TRUNK-6680: Restore lazy fetch semantics after JPA migration

Mahesh5f4:

- Restored lazy fetching for associations that unintentionally became eager during the hbm.xml to JPA annotation migration.
- Added regression tests to verify that key associations remain lazily initialized until accessed.

## Issue I worked on

TRUNK-6680

https://issues.openmrs.org/browse/TRUNK-6680

### DIFF
--- a/api/src/main/java/org/openmrs/DrugIngredient.java
+++ b/api/src/main/java/org/openmrs/DrugIngredient.java
@@ -13,6 +13,7 @@ import java.io.Serializable;
 
 import jakarta.persistence.Column;
 import jakarta.persistence.Entity;
+import jakarta.persistence.FetchType;
 import jakarta.persistence.Id;
 import jakarta.persistence.JoinColumn;
 import jakarta.persistence.ManyToOne;
@@ -31,12 +32,12 @@ public class DrugIngredient extends BaseOpenmrsObject implements Serializable, O
 	public static final long serialVersionUID = 94023L;
 
 	// Fields
-	@ManyToOne
+	@ManyToOne(fetch = FetchType.LAZY)
 	@JoinColumn(name = "drug_id", updatable = false, insertable = false)
 	@Id
 	private Drug drug;
 
-	@ManyToOne
+	@ManyToOne(fetch = FetchType.LAZY)
 	@JoinColumn(name = "ingredient_id", updatable = false, insertable = false)
 	@Id
 	private Concept ingredient;
@@ -45,7 +46,7 @@ public class DrugIngredient extends BaseOpenmrsObject implements Serializable, O
 	private Double strength;
 
 	@JoinColumn(name = "units")
-	@ManyToOne
+	@ManyToOne(fetch = FetchType.LAZY)
 	private Concept units;
 
 	// Constructors
--- a/api/src/main/java/org/openmrs/DrugIngredient.java
+++ b/api/src/main/java/org/openmrs/DrugIngredient.java
@@ -32,12 +32,12 @@ public class DrugIngredient extends BaseOpenmrsObject implements Serializable, O
 	public static final long serialVersionUID = 94023L;
 
 	// Fields
-	@ManyToOne(fetch = FetchType.LAZY)
+	@ManyToOne
 	@JoinColumn(name = "drug_id", updatable = false, insertable = false)
 	@Id
 	private Drug drug;
 
-	@ManyToOne(fetch = FetchType.LAZY)
+	@ManyToOne
 	@JoinColumn(name = "ingredient_id", updatable = false, insertable = false)
 	@Id
 	private Concept ingredient;
--- a/api/src/main/java/org/openmrs/Relationship.java
+++ b/api/src/main/java/org/openmrs/Relationship.java
@@ -13,6 +13,7 @@ import java.util.Date;
 
 import jakarta.persistence.Column;
 import jakarta.persistence.Entity;
+import jakarta.persistence.FetchType;
 import jakarta.persistence.GeneratedValue;
 import jakarta.persistence.GenerationType;
 import jakarta.persistence.Id;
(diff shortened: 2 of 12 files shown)


api/src/main/java/org/openmrs/Relationship.java
```diff
@@ -38,15 +39,15 @@ public class Relationship extends BaseChangeableOpenmrsData {
 	@Column(name = "relationship_id")
 	private Integer relationshipId;
 
-	@ManyToOne(optional = false)
+	@ManyToOne(optional = false, fetch = FetchType.LAZY)
```
dkayiwa, line 42: With the proxy in place, `getPersonA()`/`getPersonB()` stop reporting the concrete subtype. On master, `getRelationship(1).getPersonB()` comes back as `org.openmrs.Patient` and `instanceof Patient` is true; on this branch it is `org.openmrs.Person$HibernateProxy` and `instanceof Patient` is false. It flips back to true whenever that person already happens to be in the session (`getRelationshipsByPerson(person)` does exactly that), so a caller sees it intermittently rather than consistently. Same shape for the `Concept` targets in this PR (`ConceptAnswer.concept`/`answerConcept`, `DrugIngredient.ingredient`, `Location.type`) hiding `ConceptNumeric` and `ConceptComplex`.

I don't think it blocks. It is what the 2.6.x hbm mappings did, core already deproxies where it matters (`PatientServiceImpl.getPatient`, `Obs.getValueCoded`), and webservices.rest resolves a proxy class up to its superclass resource. But nothing in core's test suite would catch a module that depends on the concrete type. Is this worth a line on TRUNK-6680 or in the 3.0 release notes, so whoever runs into it has something to find?

api/src/main/java/org/openmrs/DrugIngredient.java
```diff
@@ -31,12 +32,12 @@ public class DrugIngredient extends BaseOpenmrsObject implements Serializable, O
...
+	@ManyToOne(fetch = FetchType.LAZY)
 	@JoinColumn(name = "drug_id", updatable = false, insertable = false)
 	@Id
 	private Drug drug;
 
-	@ManyToOne
+	@ManyToOne(fetch = FetchType.LAZY)
 	@JoinColumn(name = "ingredient_id", updatable = false, insertable = false)
 	@Id
 	private Concept ingredient;
```
dkayiwa: TRUNK-6680 lists only `DrugIngredient.units` for this class, and these two are the composite identifier rather than ordinary associations. In the 2.6.x mapping they were `<key-many-to-one>` inside `<composite-id>`, not `<many-to-one>`, so "match the hbm semantics" doesn't obviously reach them.

Hibernate 7 does honour the hint here: a `from DrugIngredient` query now returns `Drug$HibernateProxy` and `Concept$HibernateProxy` for both, where master returns the real entities. I saved a drug with a new ingredient on this branch and it still round-trips, so nothing looks broken.

Was widening it to the identifier deliberate? If it wasn't, I would keep this class to `units` and leave the two id properties as they were, since an entity's id is a surprising place to pick up new fetch semantics as a side effect.

api/src/test/java/org/openmrs/api/LocationServiceTest.java
```diff
@@ -1446,4 +1446,20 @@ public void getLocations_shouldApplyPaginationAfterOtherFilters() {
...
+		child.setName("Child");
+		Location parent = ls.getLocation(1);
+		child.setParentLocation(parent);
+		ls.saveLocation(child);
+
+		Context.flushSession();
+		Context.clearSession();
+
+		Location fetchedChild = ls.getLocation(child.getId());
+		org.junit.jupiter.api.Assertions.assertFalse(org.hibernate.Hibernate.isInitialized(fetchedChild.getParentLocation()),
```
dkayiwa: `assertFalse` is already statically imported at the top of this file, and every other assertion here uses the short form, so the fully qualified call stands out.

```suggestion
		assertFalse(org.hibernate.Hibernate.isInitialized(fetchedChild.getParentLocation()),
```

Adding `import org.hibernate.Hibernate;` and dropping that package prefix too would tidy the rest of it, and the same applies to the new test in `ConceptTest`. Trivial either way.

api/src/main/java/org/openmrs/Relationship.java
Mahesh5f4: Thanks for the explanation. My goal was to restore the previous hbm.xml lazy-loading semantics, so I understand that this proxy behavior is consistent with the historical mapping. I agree that adding a note to the issue or the 3.0 release notes would be helpful for anyone who runs into this behavior.

api/src/main/java/org/openmrs/DrugIngredient.java
Mahesh5f4: Thanks for the clarification. This wasn't deliberate. My intention was to restore the lazy fetch semantics only for the associations covered by TRUNK-6680. I've updated the implementation by removing fetch = FetchType.LAZY from the composite identifier properties (drug and ingredient) and kept the change limited to units.

api/src/test/java/org/openmrs/api/LocationServiceTest.java
Mahesh5f4: Thanks! I've updated both LocationServiceTest and ConceptTest to use the existing static assertFalse import and imported org.hibernate.Hibernate, removing the fully qualified references for consistency with the existing code style.

jwnasambu:
@claude review

jwnasambu (approved):

api/src/test/java/org/openmrs/ConceptTest.java
```diff
@@ -1264,4 +1265,12 @@ public void addSetMember_shouldAppendConceptToExistingConceptSetHavingRetiredCon
 		assertThat(concept.getSetMembers(), hasItem(setMember3));
 		assertThat(concept.getSetMembers().size(), is(3));
 	}
+
+	@Test
+	public void shouldLoadConceptAnswersLazily() throws Exception {
```
ibacher: nit, but important: please match the test naming convention in this file, i.e., `<methodName>_should...`

ibacher (changes requested):

ibacher:
@Mahesh5f4 In future PRs, please leave the PR template in place and fill it in. The template is there to help both you ensure you've followed our conventions and to help reviewers understand the state of the PR.

api/src/main/java/org/openmrs/Relationship.java
ibacher: As said in the ticket I was hoping to backport this to 2.8, but maybe that's a bad idea... The main idea is that since we more or less mistakenly introduced the non-laziness in 2.7, that there's unlikely to be anything that depends on this in a load-bearing way, but maybe we save this for 2.9?

api/src/test/java/org/openmrs/api/LocationServiceTest.java
```diff
@@ -1446,4 +1447,19 @@ public void getLocations_shouldApplyPaginationAfterOtherFilters() {
 		assertEquals(filtered.get(1).getLocationId(), paged.get(0).getLocationId());
 	}
 
+	@Test
+	public void shouldLoadLocationParentLazily() throws Exception {
```
ibacher:
```suggestion
	public void getLocation_shouldLoadLocationParentLazily() throws Exception {
```

ibacher (approved):

PR #6384, opened 2026-07-24
TRUNK-6730 : Add Application Event Publishers for Module action events

sudhanshu-raj:
This is publishing the module action events  load, stop, start or unload of modules, and for auditing the module events in the audit log web module and  related to this [PR](https://github.com/openmrs/openmrs-module-auditlogweb/pull/57). 

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6730

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/context/ServiceContext.java
+++ b/api/src/main/java/org/openmrs/api/context/ServiceContext.java
@@ -46,7 +46,6 @@ import org.openmrs.hl7.HL7Service;
 import org.openmrs.logic.LogicService;
 import org.openmrs.messagesource.MessageSourceService;
 import org.openmrs.messagesource.impl.DefaultMessageSourceServiceImpl;
-import org.openmrs.module.ModuleFactory;
 import org.openmrs.notification.AlertService;
 import org.openmrs.notification.MessageService;
 import org.openmrs.scheduler.SchedulerService;
--- a/api/src/main/java/org/openmrs/api/context/ServiceContext.java
+++ b/api/src/main/java/org/openmrs/api/context/ServiceContext.java
@@ -933,7 +933,6 @@ public class ServiceContext implements ApplicationContextAware {
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) {
 		this.applicationContext = applicationContext;
-		ModuleFactory.setApplicationEventPublisher(applicationContext);
 	}
 	
 	public ApplicationContext getApplicationContext() {
--- a/api/src/main/java/org/openmrs/api/context/ServiceContext.java
+++ b/api/src/main/java/org/openmrs/api/context/ServiceContext.java
@@ -46,6 +46,7 @@ import org.openmrs.hl7.HL7Service;
 import org.openmrs.logic.LogicService;
 import org.openmrs.messagesource.MessageSourceService;
 import org.openmrs.messagesource.impl.DefaultMessageSourceServiceImpl;
+import org.openmrs.module.ModuleFactory;
 import org.openmrs.notification.AlertService;
 import org.openmrs.notification.MessageService;
 import org.openmrs.scheduler.SchedulerService;
@@ -932,6 +933,7 @@ public class ServiceContext implements ApplicationContextAware {
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) {
 		this.applicationContext = applicationContext;
+		ModuleFactory.setApplicationEventPublisher(applicationContext);
 	}
 	
 	public ApplicationContext getApplicationContext() {
--- a/api/src/main/java/org/openmrs/module/AbstractModuleEvent.java
+++ b/api/src/main/java/org/openmrs/module/AbstractModuleEvent.java
@@ -33,8 +33,8 @@ public abstract class AbstractModuleEvent extends ApplicationEvent {
 	
 	private final String failureReason;
 	
-	protected AbstractModuleEvent(Object source, String moduleId, String moduleName, String moduleVersion, boolean isSuccess,
-	    String failureReason) {
+	protected AbstractModuleEvent(Object source, String moduleId, String moduleName, String moduleVersion,
+	    boolean isSuccess, String failureReason) {
 		super(source);
 		this.moduleId = moduleId;
 		this.moduleName = moduleName;
--- a/api/src/main/java/org/openmrs/module/AbstractModuleEvent.java
+++ b/api/src/main/java/org/openmrs/module/AbstractModuleEvent.java
@@ -33,7 +33,7 @@ public abstract class AbstractModuleEvent extends ApplicationEvent {
 	
 	private final String failureReason;
 	
-	public AbstractModuleEvent(Object source, String moduleId, String moduleName, String moduleVersion, boolean isSuccess,
+	protected AbstractModuleEvent(Object source, String moduleId, String moduleName, String moduleVersion, boolean isSuccess,
 	    String failureReason) {
 		super(source);
 		this.moduleId = moduleId;
--- a/api/src/main/java/org/openmrs/module/AbstractModuleEvent.java
+++ b/api/src/main/java/org/openmrs/module/AbstractModuleEvent.java
@@ -12,14 +12,16 @@ package org.openmrs.module;
 import org.springframework.context.ApplicationEvent;
 
 /**
- * A POJO representing a module action event which
- * carries data for module-related events so they can be published across the app.
+ * Base {@link ApplicationEvent} for module lifecycle actions executed by {@link ModuleFactory}.
+ * Subclasses represent concrete lifecycle actions such as load, start, stop, and unload.
+ * <p>
+ * The payload includes the event payload and the final action outcome.
+ * It will use by listener to get the module event metadata. 
+ * </p>
  * 
  * @since 2.7.10
  */
-public class ModuleActionEvent extends ApplicationEvent {
-	
-	private final ModuleEventType eventType;
+public abstract class AbstractModuleEvent extends ApplicationEvent {
 	
 	private final String moduleId;
 	
@@ -31,19 +33,14 @@ public class ModuleActionEvent extends ApplicationEvent {
 	
 	private final String failureReason;
 	
-	public ModuleActionEvent(Object source, ModuleEventType eventType, String moduleId, String moduleName,
-	    String moduleVersion, boolean isSuccess, String failureReason) {
+	public AbstractModuleEvent(Object source, String moduleId, String moduleName, String moduleVersion, boolean isSuccess,
+	    String failureReason) {
 		super(source);
-		this.eventType = eventType;
 		this.moduleId = moduleId;
 		this.moduleName = moduleName;
 		this.moduleVersion = moduleVersion;
 		this.isSuccess = isSuccess;
 		this.failureReason = failureReason;
-	}
-	
-	public ModuleEventType getEventType() {
-		return eventType;
 	}
 	
 	public String getModuleId() {
@@ -65,5 +62,4 @@ public class ModuleActionEvent extends ApplicationEvent {
 	public String getFailureReason() {
 		return failureReason;
 	}
-	
 }
--- a/api/src/main/java/org/openmrs/module/ModuleActionEvent.java
+++ b/api/src/main/java/org/openmrs/module/ModuleActionEvent.java
@@ -11,6 +11,12 @@ package org.openmrs.module;
 
 import org.springframework.context.ApplicationEvent;
 
+/**
+ * A POJO representing a module action event which
+ * carries data for module-related events so they can be published across the app.
+ * 
+ * @since 2.7.10
+ */
 public class ModuleActionEvent extends ApplicationEvent {
 	
 	private ModuleEventType eventType;
@@ -38,13 +44,4 @@ public class ModuleActionEvent extends ApplicationEvent {
 		return isSuccess;
 	}
 	
-}
-
-enum ModuleEventType {
-
-	MODULE_START,
-	MODULE_STOP,
-	MODULE_LOAD,
-	MODULE_UNLOAD;
-
 }
--- a/api/src/main/java/org/openmrs/module/ModuleActionEvent.java
+++ b/api/src/main/java/org/openmrs/module/ModuleActionEvent.java
@@ -21,14 +21,21 @@ public class ModuleActionEvent extends ApplicationEvent {
 	
 	private ModuleEventType eventType;
 	
+	private String moduleId;
+	
 	private String moduleName;
+	
+	private String moduleVersion;
 	
 	private boolean isSuccess;
 	
-	public ModuleActionEvent(Object source, ModuleEventType eventType, String moduleName, boolean isSuccess) {
+	public ModuleActionEvent(Object source, ModuleEventType eventType, String moduleId, String moduleName,
+	    String moduleVersion, boolean isSuccess) {
 		super(source);
 		this.eventType = eventType;
+		this.moduleId = moduleId;
 		this.moduleName = moduleName;
+		this.moduleVersion = moduleVersion;
 		this.isSuccess = isSuccess;
 	}
 	
@@ -36,8 +43,16 @@ public class ModuleActionEvent extends ApplicationEvent {
 		return eventType;
 	}
 	
+	public String getModuleId() {
+		return moduleId;
+	}
+	
 	public String getModuleName() {
 		return moduleName;
+	}
+	
+	public String getModuleVersion() {
+		return moduleVersion;
 	}
 	
 	public boolean isSuccess() {
--- a/api/src/main/java/org/openmrs/module/ModuleActionEvent.java
+++ b/api/src/main/java/org/openmrs/module/ModuleActionEvent.java
@@ -19,15 +19,15 @@ import org.springframework.context.ApplicationEvent;
  */
 public class ModuleActionEvent extends ApplicationEvent {
 	
-	private ModuleEventType eventType;
+	private final ModuleEventType eventType;
 	
-	private String moduleId;
+	private final String moduleId;
 	
-	private String moduleName;
+	private final String moduleName;
 	
-	private String moduleVersion;
+	private final String moduleVersion;
 	
-	private boolean isSuccess;
+	private final boolean isSuccess;
 	
 	public ModuleActionEvent(Object source, ModuleEventType eventType, String moduleId, String moduleName,
 	    String moduleVersion, boolean isSuccess) {
@@ -39,7 +39,7 @@ public class ModuleActionEvent extends ApplicationEvent {
 		this.isSuccess = isSuccess;
 	}
 	
-	public ModuleEventType getActionType() {
+	public ModuleEventType getEventType() {
 		return eventType;
 	}
 	
--- a/api/src/main/java/org/openmrs/module/ModuleActionEvent.java
+++ b/api/src/main/java/org/openmrs/module/ModuleActionEvent.java
@@ -29,14 +29,17 @@ public class ModuleActionEvent extends ApplicationEvent {
 	
 	private final boolean isSuccess;
 	
+	private final String failureReason;
+	
 	public ModuleActionEvent(Object source, ModuleEventType eventType, String moduleId, String moduleName,
-	    String moduleVersion, boolean isSuccess) {
+	    String moduleVersion, boolean isSuccess, String failureReason) {
 		super(source);
 		this.eventType = eventType;
 		this.moduleId = moduleId;
 		this.moduleName = moduleName;
 		this.moduleVersion = moduleVersion;
 		this.isSuccess = isSuccess;
+		this.failureReason = failureReason;
 	}
 	
 	public ModuleEventType getEventType() {
@@ -57,6 +60,10 @@ public class ModuleActionEvent extends ApplicationEvent {
 	
 	public boolean isSuccess() {
 		return isSuccess;
+	}
+	
+	public String getFailureReason() {
+		return failureReason;
 	}
 	
 }
--- a/api/src/main/java/org/openmrs/module/ModuleActionEvent.java
+++ b/api/src/main/java/org/openmrs/module/ModuleActionEvent.java
@@ -1,0 +1,50 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module;
+
+import org.springframework.context.ApplicationEvent;
+
+public class ModuleActionEvent extends ApplicationEvent {
+	
+	private ModuleEventType eventType;
+	
+	private String moduleName;
+	
+	private boolean isSuccess;
+	
+	public ModuleActionEvent(Object source, ModuleEventType eventType, String moduleName, boolean isSuccess) {
+		super(source);
+		this.eventType = eventType;
+		this.moduleName = moduleName;
+		this.isSuccess = isSuccess;
+	}
+	
+	public ModuleEventType getActionType() {
+		return eventType;
+	}
+	
+	public String getModuleName() {
+		return moduleName;
+	}
+	
+	public boolean isSuccess() {
+		return isSuccess;
+	}
+	
+}
+
+enum ModuleEventType {
+
+	MODULE_START,
+	MODULE_STOP,
+	MODULE_LOAD,
+	MODULE_UNLOAD;
+
+}
--- a/api/src/main/java/org/openmrs/module/ModuleFactory.java
+++ b/api/src/main/java/org/openmrs/module/ModuleFactory.java
@@ -116,7 +116,6 @@ public class ModuleFactory {
 		Module module = null;
 		boolean isModuleLoaded = true;
 		String failureReason = null;
-		String fileName;
 
 		try {
 			module = new ModuleFileParser(Context.getMessageSourceService()).parse(moduleFile);
@@ -129,25 +128,12 @@ public class ModuleFactory {
 			failureReason = ex.getMessage();
 			throw ex;
 		} finally {
-			Module moduleToPublish = module;
-			if (module == null) {
-				fileName = ModuleUtil.getModuleNameAndVersionFromFileName(moduleFile.getName());
-				isModuleLoaded = false;
-				String name = null;
-				String version = null;
-
-				if (fileName != null) {
-					String[] parts = fileName.split(":");
-
-					name = parts[0];
-					version = (parts.length > 1) ? parts[1] : null;
-				}
-				moduleToPublish = new Module(name, null, null, null, null, version, null);
-			}
-			publishModuleEvents(new ModuleLoadEvent(ModuleFactory.class, moduleToPublish.getModuleId(), moduleToPublish.getName(),
-			    moduleToPublish.getVersion(), isModuleLoaded, failureReason));
+			publishModuleEvents(new ModuleLoadEvent(ModuleFactory.class,
+				module != null ? module.getModuleId() : null, 
+				module !=null ? module.getName() : null,
+				module !=null ? module.getVersion() : null,
+				moduleFile.getName(), isModuleLoaded, failureReason));
 		}
-		
 		return module;
 	}
 	
@@ -155,14 +141,20 @@ public class ModuleFactory {
 	 * Add a module to the list of openmrs modules
 	 *
 	 * @param module
-	 * @param replaceIfExists unload a module that has the same moduleId if one is loaded already
-	 *            <strong>Should</strong> load module if it is currently not loaded
-	 *            <strong>Should</strong> not load module if already loaded <strong>Should</strong>
-	 *            always load module if replacement is wanted <strong>Should</strong> not load an older
-	 *            version of the same module <strong>Should</strong> load a newer version of the same
-	 *            module
-	 * @return module the module that was loaded or if the module exists already with the same version,
-	 *         the old module
+	 * @param replaceIfExists unload a module that has the same moduleId if one is
+	 *                        loaded already
+	 *                        <strong>Should</strong> load module if it is currently
+	 *                        not loaded
+	 *                        <strong>Should</strong> not load module if already
+	 *                        loaded <strong>Should</strong>
+	 *                        always load module if replacement is wanted
+	 *                        <strong>Should</strong> not load an older
+	 *                        version of the same module <strong>Should</strong>
+	 *                        load a newer version of the same
+	 *                        module
+	 * @return module the module that was loaded
+	 * @throws ModuleException if the module exists already with the same or newer
+	 *                         version
 	 */
 	public static Module loadModule(Module module, Boolean replaceIfExists) throws ModuleException {
 		
@@ -183,7 +175,7 @@ public class ModuleFactory {
 					throw new ModuleException("A module with the same id and version already exists", module.getModuleId());
 				}
 			} else {
-				throw new ModuleException("There exists a same module with latest version than this module version");
+				throw new ModuleException("A newer version of this module is already loaded");
 			}
 		}
 		
@@ -719,7 +711,7 @@ public class ModuleFactory {
 				
 				// Sort this module's extensions, and merge them into the full extensions map
 				Comparator<Extension> sortOrder = (e1, e2) -> Integer.valueOf(e1.getOrder()).compareTo(e2.getOrder());
-				for (Entry<String, List<Extension>> moduleExtensionEntry : moduleExtensionMap.entrySet()) {
+				for (Map.Entry<String, List<Extension>> moduleExtensionEntry : moduleExtensionMap.entrySet()) {
 					// Sort this module's extensions for current extension point
 					List<Extension> sortedModuleExtensions = moduleExtensionEntry.getValue();
 					sortedModuleExtensions.sort(sortOrder);
@@ -746,7 +738,7 @@ public class ModuleFactory {
 					// be nobody because this is being run at startup)
 					Context.addProxyPrivilege("");
 					
-					for (Entry<String, String> entry : diffs.entrySet()) {
+					for (Map.Entry<String, String> entry : diffs.entrySet()) {
 						String version = entry.getKey();
 						String sql = entry.getValue();
 						if (StringUtils.hasText(sql)) {
@@ -1066,8 +1058,19 @@ public class ModuleFactory {
 	}
 
 	/**
-	 * It's a wrapper over {@link ModuleFactory#doStopModule}, 
-	 * which needed to separate the event publish part from the actual stop module logic.
+	 * Runs through the advice and extension points and removes from api.<br>
+	 * <code>skipOverStartedProperty</code> should only be true when openmrs is stopping modules because
+	 * it is shutting down. When normally stopping a module, use {@link #stopModule(Module)} (or leave
+	 * value as false). This property controls whether the globalproperty is set for startup/shutdown.
+	 * <br>
+	 * Also calls module's {@link ModuleActivator#stopped()}
+	 *
+	 * @param mod module to stop
+	 * @param skipOverStartedProperty true if we don't want to set &lt;moduleid&gt;.started to false
+	 * @param isFailedStartup true if this is being called as a cleanup because of a failed module
+	 *            startup
+	 * @return list of dependent modules that were stopped because this module was stopped. This will
+	 *         never be null.
 	 */
 	public static List<Module> stopModule(Module mod, boolean skipOverStartedProperty, boolean isFailedStartup)
 		throws ModuleMustStartException {
@@ -1088,27 +1091,14 @@ public class ModuleFactory {
 			throw ex;
 		}
 		finally {
-			publishModuleEvents(new ModuleStopEvent(ModuleFactory.class, mod.getModuleId(), mod.getName(), mod.getVersion(), isStoppedSuccess && !isModuleStarted(mod), failureReason));
+			publishModuleEvents(new ModuleStopEvent(ModuleFactory.class, mod.getModuleId(), mod.getName(), mod.getVersion(),
+							isStoppedSuccess && !isModuleStarted(mod), failureReason));
 		}
 		return dependentModulesStopped;
 	}
-
-
-	/**
-	 * Runs through the advice and extension points and removes from api.<br>
-	 * <code>skipOverStartedProperty</code> should only be true when openmrs is stopping modules because
-	 * it is shutting down. When normally stopping a module, use {@link #stopModule(Module)} (or leave
-	 * value as false). This property controls whether the globalproperty is set for startup/shutdown.
-	 * <br>
-	 * Also calls module's {@link ModuleActivator#stopped()}
-	 *
-	 * @param mod module to stop
-	 * @param skipOverStartedProperty true if we don't want to set &lt;moduleid&gt;.started to false
-	 * @param isFailedStartup true if this is being called as a cleanup because of a failed module
-	 *            startup
-	 * @return list of dependent modules that were stopped because this module was stopped. This will
-	 *         never be null.
-	 */
+	
+	
+	// Performs the actual work of stopping a module, separated from event publishing
 	private static List<Module> doStopModule(Module mod, boolean skipOverStartedProperty, boolean isFailedStartup)
 		throws ModuleMustStartException {
 		
@@ -1286,6 +1276,9 @@ public class ModuleFactory {
 	 * @param mod module to unload
 	 */
 	public static void unloadModule(Module mod) {
+		if(mod == null) {
+			return;
+		}
 		boolean isEventSuccess = true;
 		String failureReason = null;
 		try {
@@ -1297,17 +1290,15 @@ public class ModuleFactory {
 			// remove from list of loaded modules
 			getLoadedModules().remove(mod);
 
-			if (mod != null) {
-				// remove the file from the module repository
-				File file = mod.getFile();
-				
-				boolean deleted = file.delete();
-				if (!deleted) {
-					file.deleteOnExit();
-					log.warn("Could not delete " + file.getAbsolutePath());
-				}
+			// remove the file from the module repository
+			File file = mod.getFile();
+
+			boolean deleted = file.delete();
+			if (!deleted) {
+				file.deleteOnExit();
+				log.warn("Could not delete " + file.getAbsolutePath());
 			}
-			
+
 		}
 		catch(Exception ex) {
 			isEventSuccess = false;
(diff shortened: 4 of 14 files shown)


sudhanshu-raj:
Hi @dkayiwa @wikumChamith @ibacher @ManojLL ,  please review the changes.

api/src/main/java/org/openmrs/module/ModuleActionEvent.java
```diff
@@ -0,0 +1,50 @@
...
+		return moduleName;
+	}
+	
+	public boolean isSuccess() {
+		return isSuccess;
+	}
+	
+}
+
+enum ModuleEventType {
```
dkayiwa: `ModuleEventType` is package private, so nothing outside `org.openmrs.module` can name it. If this merges as is, the discriminator on the event is unusable by the very modules the event exists for. A consumer that writes the obvious thing:

```java
ModuleEventType type = event.getActionType();
```

fails to compile with `ModuleEventType is not public in org.openmrs.module; cannot be accessed from outside package` (I compiled this class on its own and then a consumer in another package to check). The auditlogweb PR this pairs with is already living with the consequence: it reaches the value as `String.valueOf(moduleActionEvent.getActionType())` and keeps its own copy of the enum in `auditlogweb/api/utils/ModuleEventType`, complete with a `fromName(String)` parser and an `UNKNOWN` bucket, to get a typed value back. Every future consumer will have to do the same, and they will all drift apart. A public getter whose return type is inaccessible is also just a broken signature.

Could you make the enum public? That moves it into its own `ModuleEventType.java` anyway, since javac will not take a public top level type in a file named after something else. While you are in there, `ModuleActionEvent` and the enum are new public API with no javadoc at all, and this branch is 2.7.10-SNAPSHOT, so they want a short class comment and `@since 2.7.10`. The `@since 2.7` on the private `publishModuleEvents` helper can go, since `@since` on a private method does not tell anyone anything.

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -135,29 +143,36 @@ public static Module loadModule(File moduleFile, Boolean replaceIfExists) throws
...
+					isModuleLoaded = false;
+					// if the older (already loaded) module is newer, keep that original one that was loaded. return that one.
+					return oldModule;
 				}
-			} else {
-				// if the older (already loaded) module is newer, keep that original one that was loaded. return that one.
-				return oldModule;
 			}
+
+			getLoadedModulesMap().put(module.getModuleId(), module);
```
dkayiwa: `isModuleLoaded` starts out `true` and only becomes `false` on the two paths that bail out deliberately, so anything else thrown inside this `try` leaves it `true` and the `finally` publishes a *successful* load. If this merges as is, an audit log fed by these events records MODULE_LOAD / success for a load that threw and loaded nothing, and because nothing blows up there is no way to notice from the logs.

I reproduced it on your branch. Load a module, mark it mandatory, then load the same version again with `replaceIfExists` true. `unloadModule(oldModule)` throws `MandatoryModuleException`, the old module stays in the loaded map, the new one never gets put there, `loadModule` propagates the exception to its caller, and the listener still receives:

```
Test1 Module:MODULE_STOP:false, Test1 Module:MODULE_UNLOAD:false, Test1 Module:MODULE_LOAD:true
```

Flipping the default so the flag is only raised once the module is actually in the map fixes it, and the two explicit `false` assignments fall out:

```suggestion
		boolean isModuleLoaded = false;
		
		try {
			Module oldModule = getLoadedModulesMap().get(module.getModuleId());
			if (oldModule != null) {
				int versionComparison = ModuleUtil.compareVersion(oldModule.getVersion(), module.getVersion());
				if (versionComparison < 0) {
					// if oldModule version is lower, unload it and use the new
					unloadModule(oldModule);
				} else if (versionComparison == 0) {
					if (replaceIfExists) {
						// if the versions are the same and we're told to replaceIfExists, use the new
						unloadModule(oldModule);
					} else {
						// if the versions are equal and we're not told to replaceIfExists, jump out of here in a bad way
						throw new ModuleException("A module with the same id and version already exists", module.getModuleId());
					}
				} else {
					// if the older (already loaded) module is newer, keep that original one that was loaded. return that one.
					return oldModule;
				}
			}

			getLoadedModulesMap().put(module.getModuleId(), module);
			isModuleLoaded = true;
```

Worth a test as well, since none of the four new load tests exercise a throw coming out of `unloadModule`.

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -1034,156 +1063,156 @@ public static void stopModule(Module mod, boolean isShuttingDown) {
...
-				// if extends BaseModuleActivator
-				if (mod.getModuleActivator() != null) {
-					mod.getModuleActivator().willStop();
+		boolean isStoppedSuccess = true;
+		try {
+			if (mod != null) {
+
+				if (!ModuleFactory.isModuleStarted(mod)) {
+					return dependentModulesStopped;
 				}
```
dkayiwa: The early return for a module that is not running sits inside the `try`, so the `finally` still fires, and at that point `isStoppedSuccess && !isModuleStarted(mod)` evaluates to `true`. If this merges as is, the audit log gets a successful MODULE_STOP entry for a module that was never running.

The path that makes this matter is not somebody stopping a module twice, it is a failed start. When `startModuleInternal` throws before it reaches `getStartedModulesMap().put(...)` (a `require_version` mismatch against the running core is the everyday case) its catch block calls `stopModule(module, false, true)` to undo the startup, that call returns right here, and the listener sees:

```
Test1 Module:MODULE_STOP:true, Test1 Module:MODULE_START:false
```

I ran both that case and a plain `stopModule` on a loaded but never started module, and each produced a successful stop event. Someone auditing a server later cannot tell those apart from a real stop.

Hoisting the guard above the `try` fixes it:

```suggestion
		List<Module> dependentModulesStopped = new ArrayList<>();
		if (mod == null || !ModuleFactory.isModuleStarted(mod)) {
			return dependentModulesStopped;
		}
		
		boolean isStoppedSuccess = true;
		try {
			if (mod != null) {
```

After that the `if (mod != null)` wrapper is redundant, and so is the `mod != null` check in the `finally`, so both can go and the body loses a level of indentation.

One thing you might do while you are in here: wrapping the existing body in place is what forced the ~150 lines of re-indentation that make up most of this diff, and it rewrites the blame for the method. Pulling the old body into a private `doStopModule(...)` and leaving `stopModule` as the guard plus a try/finally around that call would keep the actual change visible, and it should also clear the five `java:S1141` "extract this nested try block" findings Sonar raised on this file.

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -1216,25 +1245,31 @@ private static ModuleClassLoader removeClassLoader(Module mod) {
...
+
+			if (mod != null) {
+				// remove the file from the module repository
+				File file = mod.getFile();
+				boolean deleted = file.delete();
+				if (!deleted) {
+					file.deleteOnExit();
+					isEventSuccess = false;
+					log.warn("Could not delete " + file.getAbsolutePath());
+				}
```
dkayiwa, line 1319: Not being able to delete the .omod file does not mean the unload failed. By the time we reach this branch the module has been stopped, removed from the loaded modules map and its class loader disposed, and `deleteOnExit()` has just been registered so the leftover file goes away at JVM exit. The surrounding code has always treated a failed delete as expected and recoverable, which is exactly why that fallback is there.

If this merges as is, a completed unload is recorded as a failed one. I forced `file.delete()` to return false on your branch and got:

```
Test1 Module:MODULE_STOP:true, Test1 Module:MODULE_UNLOAD:false
```

with the module genuinely gone from `getLoadedModules()`. Whoever reads that audit entry later concludes a module failed to unload when it did unload, and the only real symptom, a stale file in the module repository, is not what the entry says.

I would drop the flag here and let the existing warn carry the undeleted file:

```suggestion
				if (!deleted) {
					file.deleteOnExit();
					log.warn("Could not delete " + file.getAbsolutePath());
				}
```

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -62,8 +63,15 @@
  */
 public class ModuleFactory {
 	
+	private static Logger logger = LoggerFactory.getLogger(ModuleFactory.class);
```
dkayiwa: This `logger` is never read. The class declares `log` ten lines further down and all 51 logging calls in the file, including the new `publishModuleEvents`, go through that one. It is also the only non final logger field here. Worth deleting.

api/src/main/java/org/openmrs/module/ModuleActionEvent.java
```diff
@@ -0,0 +1,50 @@
...
+
+public class ModuleActionEvent extends ApplicationEvent {
+	
+	private ModuleEventType eventType;
+	
+	private String moduleName;
+	
+	private boolean isSuccess;
+	
+	public ModuleActionEvent(Object source, ModuleEventType eventType, String moduleName, boolean isSuccess) {
```
dkayiwa: Should this carry the module id alongside the display name? `Module.getName()` is the `<name>` from config.xml, a human label that can change between versions, whereas `moduleId` is the stable identifier the rest of the module API keys on (`getModuleById`, the `<moduleId>.started` global property, `ModuleConstants.CORE_MODULES`). An audit row that only says "Test1 Module" is awkward to join back to anything, and for load and start events the module version would be worth having too.

Since this is brand new API, adding `moduleId` (and `version`) to the constructor now is a lot cheaper than after 2.7.10 ships with the current shape.

api/src/main/java/org/openmrs/api/context/ServiceContext.java
```diff
@@ -932,6 +933,7 @@ private <T> Map<String, T> getRegisteredComponents(ApplicationContext context, C
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) {
 		this.applicationContext = applicationContext;
+		ModuleFactory.setApplicationEventPublisher(applicationContext);
```
dkayiwa: `ServiceContext` already holds this context and hands it out through `getApplicationContext()`, so `ModuleFactory` can read it when it publishes rather than keeping a second copy in a static field:

```java
ApplicationEventPublisher publisher = ServiceContext.getInstance().getApplicationContext();
if (publisher != null) {
	publisher.publishEvent(new ModuleActionEvent(ModuleFactory.class, eventType, moduleName, isSuccess));
}
```

That takes out this line and the whole change to this file, drops the static field, and keeps `setApplicationEventPublisher` off `ModuleFactory`'s public surface. As written, any caller can hand `ModuleFactory` a different publisher or a null one and quietly switch module auditing off, which is an odd thing to expose on the class this feature is meant to make auditable. It also avoids `ServiceContext` picking up its first dependency on `org.openmrs.module`, which is the awkward direction given `ModuleFactory` already depends on `org.openmrs.api.context`.

I tried it on your branch: with this file reverted and `publishModuleEvents` reading the publisher off `ServiceContext`, all 19 `ModuleFactoryTest` tests pass, so the timing works out the same. Both routes are fed by this same callback. Not a blocker, but I think it is the better shape.

web/src/main/java/org/openmrs/web/filter/initialization/DatabaseDetective.java
```diff
@@ -42,7 +42,8 @@ public boolean isDatabaseEmpty(Properties props) {
 		Connection connection = null;
 		
 		try {
-			DatabaseUtil.loadDatabaseDriver(props.getProperty(CONNECTION_URL), props.getProperty(CONNECTION_DRIVER_CLASS, null));
+			DatabaseUtil.loadDatabaseDriver(props.getProperty(CONNECTION_URL), props.getProperty(CONNECTION_DRIVER_CLASS,
+			    null));
```
dkayiwa: This file has nothing to do with TRUNK-6730. The two changes in it are this line wrap and a tab added to an otherwise blank line near the top, which look like an IDE formatter pass on a file that happened to be open. Reverting it keeps the PR inside the api module.

api/src/main/java/org/openmrs/module/ModuleActionEvent.java
sudhanshu-raj: agree, fixed it

api/src/main/java/org/openmrs/module/ModuleFactory.java
sudhanshu-raj: flipped the success var value and added test cases too

api/src/main/java/org/openmrs/module/ModuleFactory.java
sudhanshu-raj: It's good edge case,  fixed and added test case too

api/src/main/java/org/openmrs/module/ModuleFactory.java
sudhanshu-raj: I guessed it same, anyway fixed

api/src/main/java/org/openmrs/module/ModuleFactory.java
sudhanshu-raj: My bad, didn't see !

api/src/main/java/org/openmrs/api/context/ServiceContext.java
sudhanshu-raj: Ahh new learning for me, thnx

api/src/main/java/org/openmrs/module/ModuleActionEvent.java
sudhanshu-raj: True the module name itself can't tell more  about itself, so version is needed and module id also can be useful for the easy API queries, added both.

web/src/main/java/org/openmrs/web/filter/initialization/DatabaseDetective.java
sudhanshu-raj: fixed

api/src/main/java/org/openmrs/api/context/ServiceContext.java
```diff
@@ -46,6 +46,7 @@
 import org.openmrs.logic.LogicService;
 import org.openmrs.messagesource.MessageSourceService;
 import org.openmrs.messagesource.impl.DefaultMessageSourceServiceImpl;
+import org.openmrs.module.ModuleFactory;
```
wikumChamith: Do we need this import?

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -1016,7 +1038,31 @@ public static void stopModule(Module mod) {
...
+		List<Module> dependentModulesStopped = new ArrayList<>();
+
+		if (mod == null || !ModuleFactory.isModuleStarted(mod)) {
+			return dependentModulesStopped;
+		}
+		
+		boolean isStoppedSuccess = true;
+		try {
+			dependentModulesStopped = doStopModule(mod, skipOverStartedProperty, isFailedStartup);
+		} catch(ModuleMustStartException ex) {
```
wikumChamith: What will happen if we get an another type of exception here?

api/src/main/java/org/openmrs/api/context/ServiceContext.java
sudhanshu-raj: Removed

api/src/main/java/org/openmrs/module/ModuleFactory.java
sudhanshu-raj: Pivoted to more general exception now, I thought  this module exception could be only reason for failure event

wikumChamith (commented):
@sudhanshu-raj this PR contains a lot of formatting changes, and that's making it harder to review. Can we revert those changes?

sudhanshu-raj:
> @sudhanshu-raj this PR contains a lot of formatting changes, and that's making it harder to review. Can we revert those changes?

Not sure why getting such view here,  I tried now some changes but seems not that helpful,  any suggestions ? Though if view from the IntelliJ compare with branch feature its not that cluttered and it's get very clear the new changes.

api/src/main/java/org/openmrs/module/ModuleActionEvent.java
```diff
@@ -0,0 +1,62 @@
...
+	    String moduleVersion, boolean isSuccess) {
+		super(source);
+		this.eventType = eventType;
+		this.moduleId = moduleId;
+		this.moduleName = moduleName;
+		this.moduleVersion = moduleVersion;
+		this.isSuccess = isSuccess;
+	}
+	
+	public ModuleEventType getActionType() {
```
wikumChamith: shouldn't this be `getEventType()`?

api/src/main/java/org/openmrs/module/ModuleActionEvent.java
```diff
@@ -0,0 +1,62 @@
...
+	
+	private ModuleEventType eventType;
+	
+	private String moduleId;
+	
+	private String moduleName;
+	
+	private String moduleVersion;
+	
+	private boolean isSuccess;
```
wikumChamith: I think we can make these final.

api/src/main/java/org/openmrs/module/ModuleFactory.java
wikumChamith: What's gonna happen if this parser fails due to a corruption? Don't we need to log that?

api/src/main/java/org/openmrs/module/ModuleActionEvent.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.springframework.context.ApplicationEvent;
+
+/**
+ * A POJO representing a module action event which
+ * carries data for module-related events so they can be published across the app.
+ * 
+ * @since 2.7.10
+ */
+public class ModuleActionEvent extends ApplicationEvent {
```
wikumChamith, line 20: What if we add a `failureReason` field? That way audit log can say why something failed.

api/src/main/java/org/openmrs/module/ModuleActionEvent.java
sudhanshu-raj: Done!

api/src/main/java/org/openmrs/module/ModuleActionEvent.java
sudhanshu-raj: Done!

api/src/main/java/org/openmrs/module/ModuleFactory.java
sudhanshu-raj: I added the tracker for this so when ever the module file parsing will fail, we will mark with fail event and will extract the module name and version(whatever is possible) from the module filename.

api/src/main/java/org/openmrs/module/ModuleActionEvent.java
sudhanshu-raj: Added

sudhanshu-raj: Added

wikumChamith:
@sudhanshu-raj what do you think about the two issues SonarCloud is pointing at?

sudhanshu-raj:
> @sudhanshu-raj what do you think about the two issues SonarCloud is pointing at?

One is for the cognitive complexity for which it saying to re-factor the whole method to make easy to understand but this is not new method I just added new name to it  and another issue is to simplify the regular expression for the module file name pattern due to lazy quantifiers which says it can take increase runtime incase we are matching this regex against the large string but in this case string is just the module name which should not go in that way and thing is this lazy quantifier (*?) is needed to extract the proper module name and version format.

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -110,10 +112,44 @@ public static Module loadModule(File moduleFile) throws ModuleException {
...
+					   if (versionComparison > 0) {
+						   isModuleLoaded = false;
+						   failureReason = "There exists a same module with latest version than this module version";
+					   }
+				   }
+			   } 
+			   if(module == null) {
+				   fileName = ModuleUtil.getModuleNameAndVersionFromFileName(moduleFile.getName());
+				   isModuleLoaded = false;
+				   String[] parts = fileName.split(":");
```
wikumChamith: `getModuleNameAndVersionFromFileName` returns null for an empty filename (e.g. `new File("/")`), so this split can NPE inside the `finally`, replacing the real `ModuleException` and publishing no event at all. Please null-guard the result and wrap this whole block in a try/catch so event metadata can never outrank the actual failure.

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -110,10 +112,44 @@ public static Module loadModule(File moduleFile) throws ModuleException {
...
+
+				   String name = parts[0];
+				   String version = (parts.length > 1) ? parts[1] : null;
+				   module = new Module(name, null, null, null, null, version, null);
+			   }
+			   publishModuleEvents(ModuleEventType.MODULE_LOAD, module, isModuleLoaded, failureReason);
```
wikumChamith: The activator is assigned to `module` and returned to the caller, breaking the documented "returns null on error" contract. `loadModules()` would hold onto it and fail much later somewhere confusing. Can we keep the synthetic module local to the event publish and keep the return value honest?

api/src/main/java/org/openmrs/module/ModuleFactory.java
sudhanshu-raj: Fixed !

api/src/main/java/org/openmrs/module/ModuleFactory.java
sudhanshu-raj: Fixed, not tweaking the original return module .

sudhanshu-raj:
Hey @wikumChamith @ManojLL @dkayiwa,  please let me know if any further things I need to modify.

api/src/main/java/org/openmrs/module/AbstractModuleEvent.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+package org.openmrs.module;
+
+import org.springframework.context.ApplicationEvent;
+
+/**
+ * A POJO representing a module action event which
+ * carries data for module-related events so they can be published across the app.
+ * 
+ * @since 2.7.10
```
ibacher, line 22: Why are we targeting 2.7 for this?

api/src/main/java/org/openmrs/module/ModuleActionEvent.java
```diff
@@ -0,0 +1,69 @@
...
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module;
+
+import org.springframework.context.ApplicationEvent;
+
+/**
+ * A POJO representing a module action event which
+ * carries data for module-related events so they can be published across the app.
```
ibacher:
```suggestion
 * A {@code ModuleActionEvent} is...
```

ibacher: When does this fire? What does it do? What is it used for?

api/src/main/java/org/openmrs/module/ModuleActionEvent.java
```diff
@@ -0,0 +1,69 @@
...
+ * 
+ * @since 2.7.10
+ */
+public class ModuleActionEvent extends ApplicationEvent {
+	
+	private final ModuleEventType eventType;
```
ibacher: In Spring's `ApplicationEvent` instead of communicating `eventType` as a property, these are usually communicated as a subclass. For example, Spring has an `ApplicationContextEvent` abstract class with concrete sub-types `ContextClosedEvent`, `ContextRefreshedEvent`, etc. This allows `EventListener`s to only need to process the events that they are interested in, so it's an efficient form of filtering. It would seem better to use that than to assign an enum like this.

api/src/main/java/org/openmrs/module/AbstractModuleEvent.java
```diff
@@ -0,0 +1,69 @@
...
+public class ModuleActionEvent extends ApplicationEvent {
+	
+	private final ModuleEventType eventType;
+	
+	private final String moduleId;
+	
+	private final String moduleName;
+	
+	private final String moduleVersion;
+
```
ibacher, line 31: I think we only really need to communicate the `moduleId`. The name and version can be grabbed using the `ModuleFactory` itself.

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -110,10 +112,50 @@ public static Module loadModule(File moduleFile) throws ModuleException {
 	 * @return Module
 	 */
 	public static Module loadModule(File moduleFile, Boolean replaceIfExists) throws ModuleException {
-		Module module = new ModuleFileParser(Context.getMessageSourceService()).parse(moduleFile);
-		
-		if (module != null) {
-			loadModule(module, replaceIfExists);
+
+		Module module = null;
+		Module laodedModule = null;
```
ibacher:
```suggestion
		Module loadedModule = null;
```

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -110,10 +112,50 @@ public static Module loadModule(File moduleFile) throws ModuleException {
...
+
+			if (module != null) {
+				laodedModule = loadModule(module, replaceIfExists);
+			}
+		} catch (Exception ex) {
+			isModuleLoaded = false;
+			failureReason = ex.getMessage();
+			throw ex;
+		} finally {
+			   if (laodedModule != null) {
```
ibacher: Instead of all this detective work, the smarter thing to do would be to have `loadModule()` itself throw an exception.

This _definitely_ doesn't belong in a `finally` block. It's nothing to do with clean-up.

ibacher: This same set of comments applies broadly to this whole file.

api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -61,6 +61,8 @@ private ModuleUtil() {
 	}
 	
 	private static final Logger log = LoggerFactory.getLogger(ModuleUtil.class);
+
+	private static final Pattern MODULE_FILNAME_PATTERN = Pattern.compile("^(.+?)(?:-(\\d[\\w.-]*?))?(?:\\.([a-zA-Z0-9]+))?$");
```
ibacher:
```suggestion
	private static final Pattern MODULE_FILENAME_PATTERN = Pattern.compile("^(.+?)(?:-(\\d[\\w.-]*?))?(?:\\.([a-zA-Z0-9]+))?$");
```

ibacher: This seems way more complicated than what it needs to be.

api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -1250,4 +1252,31 @@ public static File getDevelopmentDirectory(String moduleId) {
...
 		return null;
 	}
+
+	/**
+	 * Gets the module name and version if possible from the filename 
+	 * 
+	 * @param fileName gets the file name from the uploaded module file
+	 * @return the module name and version in format like event:1.0.0 or just the original filename
+	 */
+	public static String getModuleNameAndVersionFromFileName(String fileName) {
```
ibacher: Why would we get this from the filename? We should be using the module descriptor for this, not the file name.

ibacher (changes requested):
Thanks for the work on this, but I feel like there's a lot of clean-up that needs to be done before this is mergeable.

api/src/main/java/org/openmrs/module/AbstractModuleEvent.java
sudhanshu-raj: Isn't because it born on the 2.7 ? And then later we will carry this in latest versions.

api/src/main/java/org/openmrs/module/ModuleUtil.java
sudhanshu-raj: This is for the fallback case for what if the we get exception early even before getting the module object, which currently happening due to the wrong module extension so in that case I'm trying to get best possible metadata like name and version which I think I can get only from the original filename and  the regex pattern for this match is seems complex because I am trying to get the metadata from whether it's a image or wrong extension module file.  Though we can completely skip this part and put the things null in that case.

api/src/main/java/org/openmrs/module/ModuleFactory.java
sudhanshu-raj: I removed the manually checking of the version instead throwing the exception for the loadModule.

api/src/main/java/org/openmrs/module/ModuleActionEvent.java
sudhanshu-raj: Done, removed the enum and made the event type bases on the sub-types

api/src/main/java/org/openmrs/module/AbstractModuleEvent.java
sudhanshu-raj: yeah we can put the `moduleId ` only but what if for the case during loading module,  module parsing throws the exception and it's invalid module due to may be wrong extension. In that way we can't fetch the module metadata, but here I'm trying to get the metadata from the filename so that we can also know what exactly user trying to upload . Again this can be optional and can replace with the null values.

api/src/main/java/org/openmrs/module/AbstractModuleEvent.java
sudhanshu-raj: And another thing I want to add, in the audit log web we are using core till 2.7, so that's probably we should use this.

api/src/main/java/org/openmrs/module/ModuleUtil.java
sudhanshu-raj: Here I'm trying to get a group of three elements possible from whatever filename we will get to a best possible filename we can use to audit, please check those test cases for it for the possible cases. And I would love  your more detailed feedback on this, of why ?

api/src/main/java/org/openmrs/module/AbstractModuleEvent.java
ibacher: In that case, this needs to be something like `@since 2.7.10, 2.8.10` (just a coincidence they're on the same point release).

ibacher: My main concern, though, is that we have a reworked event system in 2.9.x, and I think this should harmonize with that.

sudhanshu-raj: yeah after viewing those changes on 2.9.x, it seems those new changes are  perfect for publishing the events for service methods and since those methods are wrapped on the transactions too so the outbox pattern also gonna useful for that . And it explicitly intercepting the service methods like save , void etc and hibernate actions and then publishing the events but in this  case `ModuleFactory`  is not comes in that case, it can't work like that so as far I see this current changes not gonna effect or harmonize  any of existing things on 2.9.x

sudhanshu-raj:
Hi @ibacher , thoughts ?

sudhanshu-raj:
Hey @ibacher @wikumChamith @dkayiwa , still waiting for further feedbacks if any, as this one is blocker for the module auditing  PR in the audit module.

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -151,8 +183,7 @@ public static Module loadModule(Module module, Boolean replaceIfExists) throws M
 					throw new ModuleException("A module with the same id and version already exists", module.getModuleId());
 				}
 			} else {
-				// if the older (already loaded) module is newer, keep that original one that was loaded. return that one.
-				return oldModule;
+				throw new ModuleException("There exists a same module with latest version than this module version");
```
dkayiwa: The javadoc on this method still promises the old contract for exactly this branch: `@return module the module that was loaded or if the module exists already with the same version, the old module`. Nothing comes back any more, so that clause wants replacing with a `@throws ModuleException` line saying a newer version is already loaded, otherwise the next person reads the doc and writes a caller that never handles the exception.

The message itself is not an internal string. legacyui catches `ModuleException` around its `ModuleFactory.loadModule(moduleFile)` call and puts `me.getMessage()` straight on the admin's screen as the upload error, and your own `loadModule_shouldPublishFailedLoadModuleEventIfOldModuleExistsOfNewerVersion` asserts it lands on the event as `failureReason`, so it ends up in the audit rows too. As written it is not grammatical:

```suggestion
				throw new ModuleException("A newer version of this module is already loaded");
```

That test hardcodes the old wording in its `failureReason` local, so it needs the same edit.

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -671,7 +719,7 @@ public static Module startModuleInternal(Module module, boolean isOpenmrsStartup
 				
 				// Sort this module's extensions, and merge them into the full extensions map
 				Comparator<Extension> sortOrder = (e1, e2) -> Integer.valueOf(e1.getOrder()).compareTo(e2.getOrder());
-				for (Map.Entry<String, List<Extension>> moduleExtensionEntry : moduleExtensionMap.entrySet()) {
+				for (Entry<String, List<Extension>> moduleExtensionEntry : moduleExtensionMap.entrySet()) {
```
dkayiwa: This one and three others in the file have nothing to do with TRUNK-6730 and look like an IDE "simplify qualified name" pass: line 749 makes the same `Map.Entry` to `Entry` change, and lines 1348 and 1376 drop the `Extension.` prefix from `MEDIA_TYPE`. Both `java.util.Map.Entry` and `org.openmrs.module.Extension.MEDIA_TYPE` are already imported on 2.7.x, so all four are no-ops.

Being harmless is the reason to take them out rather than a reason to keep them: a reviewer still has to work that out one hunk at a time. This is also the concrete part of what @wikumChamith was pointing at when he said the formatting changes made the PR hard to read, so it may answer your question there. The rest of the churn in this file is the `doStopModule` extraction, which is real work and reads fine.

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -1016,7 +1064,36 @@ public static void stopModule(Module mod) {
 	public static void stopModule(Module mod, boolean isShuttingDown) {
 		stopModule(mod, isShuttingDown, false);
 	}
-	
+
+	/**
+	 * It's a wrapper over {@link ModuleFactory#doStopModule}, 
+	 * which needed to separate the event publish part from the actual stop module logic.
+	 */
```
dkayiwa, line 1074: The three argument `stopModule` is the one legacyui and webservices.rest both call, and its documented contract has just moved out of reach: the `@param` descriptions, the warning that `skipOverStartedProperty` should only be true while OpenMRS is shutting down, and `@return list of dependent modules that were stopped ... will never be null` now sit on the private `doStopModule`. The two overloads above still carry `@see #stopModule(Module, boolean, boolean)`, so following that link lands the reader on a note about an internal wrapper, and `{@link ModuleFactory#doStopModule}` points at a private member that will not resolve in generated javadoc.

Could you leave the original block here and give `doStopModule` the one liner instead? The public contract is the half worth keeping visible.

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -1216,26 +1286,43 @@ private static ModuleClassLoader removeClassLoader(Module mod) {
...
+		}
+		finally {
+			if (mod == null) {
+				publishModuleEvents(new ModuleUnloadEvent(ModuleFactory.class, null, null, null,  isEventSuccess, failureReason));
+			} 
+			else {
+				publishModuleEvents(new ModuleUnloadEvent(ModuleFactory.class, mod.getModuleId(), mod.getName(), mod.getVersion(),  isEventSuccess, failureReason));
+
+			}
+		}
```
dkayiwa, line 1311: `unloadModule(null)` reaches this branch with `isEventSuccess` still true, so a listener receives `ModuleUnloadEvent[id=null, name=null, version=null, success=true]`. I ran it on your branch to check that is really what comes out. An audit row saying a module was unloaded successfully, carrying no id, no name and no version, is worse than no row at all: whoever reads the log later cannot tell what it refers to, and it looks the same as a real unload whose metadata went missing.

`stopModule` makes the opposite call for the same situation a couple of hundred lines up, returning early and publishing nothing, and I think that is the right one here too. An `if (mod == null) { return; }` at the top of the method keeps today's behaviour exactly (the whole body is already a no-op for null) and makes the `if (mod != null)` guard further down redundant, which collapses this finally to a single publish:

```suggestion
		finally {
			publishModuleEvents(new ModuleUnloadEvent(ModuleFactory.class, mod.getModuleId(), mod.getName(),
			    mod.getVersion(), isEventSuccess, failureReason));
		}
```

api/src/main/java/org/openmrs/module/ModuleFactory.java
dkayiwa: The split is guarded now, but the other half of this, wrapping the block so event metadata cannot outrank the actual failure, is still open, and `moduleFile` itself is the case that gets you: it is dereferenced unguarded in the `finally`.

On 2.7.x, `ModuleFactory.loadModule(null, true)` throws `ModuleException: Module file cannot be null` from `ModuleFileParser.validateFileIsNotNull`. On this branch the same call throws `java.lang.NullPointerException: Cannot invoke "java.io.File.getName()" because "moduleFile" is null`, and the ModuleException is discarded by the finally before it can propagate. I ran both. The substituted exception is a worse outcome than the message change suggests, because callers that catch `ModuleException` specifically, which is how legacyui's module upload handler is written, would not catch it at all.

api/src/main/java/org/openmrs/module/ModuleUtil.java
dkayiwa: The trouble with taking the name and version from the filename is not that it is a bad guess, it is that the guessed values land in fields that mean something else everywhere else. On a successful load the event carries `moduleName` = "Test1 Module", the `<name>` from config.xml. On a parse failure it carries `moduleName` = "webservices.rest", a filename stem, and `moduleId` is null because a filename cannot give you one. Both shapes are in the new tests. Nothing on the event says which kind of row you are looking at, so a consumer grouping or joining by module name gets config.xml names and file stems mixed together with no way to separate them.

The round trip is lossy too, since `getModuleNameAndVersionFromFileName` glues the two values with ":" and `loadModule` splits them back apart. I ran `weird:name-1.0.omod` through it and got name "weird", version "name".

What I would do instead is carry the file name on the event as its own field, say `getModuleFileName()`, and leave id, name and version null when the descriptor never parsed. A consumer can then say "failed to load foo-1.0.omod" and be exactly right, the three module fields keep a single meaning, and both the regex and the new public method on `ModuleUtil` disappear, which also settles the separate point about the pattern being more complicated than it needs to be.

sudhanshu-raj: yeah this clear way, added the direct filename for the loadModule event only.

sudhanshu-raj:
Hey @dkayiwa,  just a quick reminder on the PR review. Thanks!

sudhanshu-raj:
Hey @dkayiwa , just a quick reminder for review, thanks

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -110,27 +112,49 @@ public static Module loadModule(File moduleFile) throws ModuleException {
...
+	 *                        <strong>Should</strong> not load module if already
+	 *                        loaded <strong>Should</strong>
+	 *                        always load module if replacement is wanted
+	 *                        <strong>Should</strong> not load an older
+	 *                        version of the same module <strong>Should</strong>
+	 *                        load a newer version of the same
+	 *                        module
+	 * @return module the module that was loaded
+	 * @throws ModuleException if the module exists already with the same or newer
+	 *                         version
```
dkayiwa, line 157: Only the `@return` and `@throws` lines here are ticket work. The `@param replaceIfExists` text above them is the same wording reflowed to a narrower column with a 24 space continuation indent, which turns a two line edit into thirteen. Lines 145 to 154 are now the only javadoc continuations in the file at that indent; everything else sits at 12 or less, so this one block reads differently from the rest of `ModuleFactory`. It came in with 651d974d, after the earlier formatting had already been cleaned out of the diff, and it is the concrete version of what @wikumChamith was asking about in August. Could you put the original wrapping back?

The new `@throws` wording is a little off as well. With `replaceIfExists` true and the same version the old module is unloaded and nothing is thrown, so the same-version case only throws when the flag is false.

```suggestion
	 * @param replaceIfExists unload a module that has the same moduleId if one is loaded already
	 *            <strong>Should</strong> load module if it is currently not loaded
	 *            <strong>Should</strong> not load module if already loaded <strong>Should</strong>
	 *            always load module if replacement is wanted <strong>Should</strong> not load an older
	 *            version of the same module <strong>Should</strong> load a newer version of the same
	 *            module
	 * @return module the module that was loaded
	 * @throws ModuleException if a module with the same id is already loaded, at a newer version or
	 *             at the same version with replaceIfExists false
```

api/src/main/java/org/openmrs/module/ModuleFactory.java
```diff
@@ -1642,4 +1715,26 @@ public static List<String> getDependencies(String moduleId) {
...
+			}
+			ApplicationEventPublisher applicationEventPublisher = ServiceContext.getInstance().getApplicationContext();
+			if (applicationEventPublisher != null) {
+				applicationEventPublisher.publishEvent(moduleEvent);
+			}
+		} 
+		catch (Exception e) {
+			log.warn("Unable to publish module event", e);
+		}
+	}
```
dkayiwa, line 1738: This catch is what stops a badly behaved listener from breaking module management, and nothing in the suite exercises it. I measured how much it is carrying: with a second `@EventListener` added to `ModuleFactoryTest` that throws `IllegalStateException`, `ModuleFactory.loadModule(file)` still returns the module; with the try/catch taken out, the same call fails with `IllegalStateException: listener blew up` coming out of `loadModule`'s finally via `SimpleApplicationEventMulticaster.doInvokeListener`, so Spring is not isolating listeners for us here.

Worth a test that registers a throwing listener alongside `testingModuleEventListener` and asserts the load still succeeds. Without one, whoever tidies this method up later can drop the catch with the whole suite still green, and the first person to notice is an admin whose module uploads start failing because the audit module happens to be installed.

api/src/main/java/org/openmrs/module/ModuleFactory.java
dkayiwa: The block got rewritten in 651d974d and `moduleFile` is still dereferenced unguarded, so the finally can still throw over the real exception. This one is reachable from the admin UI. In legacyui's `ModuleListController.onSubmit`, `File moduleFile = null` is only assigned inside the download branch or inside `if (multipartModuleFile != null && !multipartModuleFile.isEmpty())`, and `ModuleFactory.loadModule(moduleFile)` runs unconditionally after both. Neither upload form in `moduleList.jsp` marks the file input required and there is no javascript on either form, so an admin who clicks Upload without choosing a file arrives at that call with null.

I ran legacyui's exact shape, `loadModule(moduleFile)` with `moduleFile` null inside a `try` / `catch (ModuleException)`:

```
2.7.x:     caught ModuleException, "Module file cannot be null"  (what the admin sees on the page)
this head: escaped the handler, java.lang.NullPointerException: Cannot invoke
           "java.io.File.getName()" because "moduleFile" is null
```

So on 2.7.10 that click turns into a 500 page instead of the inline error, and nothing goes red because no test calls `loadModule` with a null file.

One ternary settles it, and it is the last place where argument evaluation for a publish sits outside `publishModuleEvents`' own try/catch:

```java
				moduleFile != null ? moduleFile.getName() : null, isModuleLoaded, failureReason));
```

While you are on those lines, `module !=null` on the two above is missing a space.

PR #6385, opened 2026-07-24
TRUNK-6509: Do not re-copy and re-expand unchanged module jars on every restart

dkayiwa:
## Summary

TRUNK-6509 ("Do not expand jars every restart") stopped deleting the whole lib cache and skipped Liquibase when the core version is unchanged, but `ModuleClassLoader.getUrls()` still **re-copied every module jar into the lib cache and re-expanded its `/lib` folder on every startup**, regardless of whether the module had changed. So the expansion the ticket is named after was never actually skipped.

On a module-heavy distribution that is several seconds of redundant disk I/O on every restart. 2.8.x does the same work, but 2.9.x was supposed to avoid it, so this quietly erodes the startup gains 2.9.x was expected to have over 2.8.x.

## Change

`getUrls()` now reads the module's `.moduleLastModified` marker (via a small `isModuleCacheUpToDate` helper) **before** `getLibCacheFolderForModule()` rewrites it, and skips both the jar copy and the `/lib` expansion when the cached expansion is already current.

Correctness is preserved: `getLibCacheFolderForModule()` still deletes the module's cache folder when the module actually changed (TRUNK-6675), so work is skipped only for genuinely unchanged modules. Changed modules are re-copied and re-expanded exactly as before, so the "module not updated when omod changes" fix is unaffected. When `optimized.startup` is `false`, nothing is skipped.

## Measured impact

Reference-application-3-backend (core 2.9, ~40 modules), same war built two ways (differing only by this change), mounted into the same image, steady-state restarts, Tomcat's own "Server startup in [N] ms":

| | restart 1 | restart 2 | mean |
|---|---|---|---|
| before | 22996 ms | 23086 ms | ~23.0 s |
| after | 20888 ms | 21054 ms | ~21.0 s |

~2.0 s (~9%) faster, entirely within the module-starting phase. Bare `openmrs-core` (no modules) is unaffected, as expected.

## Testing

- New `ModuleClassLoaderTest.getUrls_shouldNotReExpandLibForUnchangedModuleButShouldForChangedModule` asserts that an unchanged module is **not** re-expanded on a simulated restart while a changed module **is**.
- Full `ModuleClassLoaderTest` passes (42/42), including the existing TRUNK-6675 cases.
- Verified end-to-end on the O3 backend image: boots healthy, all modules load, no new errors and no `NoClassDefFound`/`ClassNotFound` compared to the unfixed war.

https://claude.ai/code/session_01RamM5GdnGvTDRs7asPhNtV

### DIFF
--- a/api/src/main/java/org/openmrs/module/ModuleClassLoader.java
+++ b/api/src/main/java/org/openmrs/module/ModuleClassLoader.java
@@ -231,41 +231,41 @@ public class ModuleClassLoader extends URLClassLoader {
 			log.error("Failed to add development folder to the classpath", ex);
 		}
 		
+		// Whether the module's previously expanded cache is still up to date. This is read before
+		// getLibCacheFolderForModule() below because that method rewrites the .moduleLastModified
+		// marker it relies on. When the cache is up to date we skip re-copying the module jar and
+		// re-expanding its /lib folder, which otherwise happens on every restart (TRUNK-6509).
+		// getLibCacheFolderForModule() still deletes the folder when the module actually changed
+		// (TRUNK-6675), so work is only skipped for genuinely unchanged modules.
+		boolean cacheUpToDate = isModuleCacheUpToDate(module);
 		File tmpModuleDir = getLibCacheFolderForModule(module);
-		
+
 		//add module jar to classpath only if we are not in dev mode
 		if (devDir == null) {
 			File tmpModuleJar = new File(tmpModuleDir, module.getModuleId() + ".jar");
 			
-			if (!tmpModuleJar.exists()) {
+			// copy the module jar into that temporary folder, unless an up-to-date copy is already there
+			if (!(cacheUpToDate && tmpModuleJar.exists())) {
+				FileInputStream in = null;
+				FileOutputStream out = null;
 				try {
-					tmpModuleJar.createNewFile();
+					in = new FileInputStream(module.getFile());
+					out = new FileOutputStream(tmpModuleJar);
+					OpenmrsUtil.copyFile(in, out);
 				}
 				catch (IOException io) {
-					log.warn("Unable to create tmpModuleFile", io);
+					log.warn("Unable to copy tmpModuleFile", io);
 				}
-			}
-			
-			// copy the module jar into that temporary folder
-			FileInputStream in = null;
-			FileOutputStream out = null;
-			try {
-				in = new FileInputStream(module.getFile());
-				out = new FileOutputStream(tmpModuleJar);
-				OpenmrsUtil.copyFile(in, out);
-			}
-			catch (IOException io) {
-				log.warn("Unable to copy tmpModuleFile", io);
-			}
-			finally {
-				try {
-					in.close();
+				finally {
+					try {
+						in.close();
+					}
+					catch (Exception e) { /* pass */}
+					try {
+						out.close();
+					}
+					catch (Exception e) { /* pass */}
 				}
-				catch (Exception e) { /* pass */}
-				try {
-					out.close();
-				}
-				catch (Exception e) { /* pass */}
 			}
 			
 			// add the module jar as a url in the classpath of the classloader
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/ModuleClassLoader.java
```diff
@@ -281,11 +281,14 @@ private static List<URL> getUrls(final Module module) {
...
-			
-			ModuleUtil.expandJar(module.getFile(), tmpModuleDir, "lib", true);
 			File libdir = new File(tmpModuleDir, "lib");
-			
+
+			// expand the module's /lib folder, unless it was already expanded for this version of the module
+			if (!(cacheUpToDate && libdir.exists())) {
+				log.debug("Expanding /lib folder in module");
+				ModuleUtil.expandJar(module.getFile(), tmpModuleDir, "lib", true);
+			}
```
dkayiwa, line 290: Before this PR the jar copy and this `/lib` expansion ran on every startup, so a cache left half-written by an earlier crash healed itself on the next restart. Now both are skipped whenever `cacheUpToDate && ...exists()`, and the `.moduleLastModified` marker they trust is written by `getLibCacheFolderForModule` (line 503) *before* the jar copy at line 248 and before this expansion runs. So if the process dies, or the disk fills, after the marker is written but partway through copying the jar or extracting a multi-jar lib folder, the next restart sees the marker match and the file/dir exist, skips the redo, and keeps the truncated jar or partial lib on the classpath. Since the module file hasn't changed it never self-heals; clearing it needs a version bump or `optimized.startup=false`, and the symptom is a silent ClassNotFound/NoClassDefFound at class load rather than an error at expansion time.

The happy path is correct and I haven't reproduced this, so treat it as a question about how defensive to be: is that window worth closing by writing the completion marker only after the copy and expand succeed (or via a separate `.complete` sentinel), so the marker means "fully populated" rather than "seen"? Not blocking.

api/src/main/java/org/openmrs/module/ModuleClassLoader.java
```diff
@@ -517,6 +520,39 @@ private static boolean deleteLibCacheDir(File dir, String moduleId) {
...
+
+		File tmpModuleDir = new File(OpenmrsClassLoader.getLibCacheFolder(), module.getModuleId());
+		File moduleLastModifiedFile = new File(tmpModuleDir, ".moduleLastModified");
+		if (!moduleLastModifiedFile.exists()) {
+			return false;
+		}
+
+		try {
+			String savedLastModified = FileUtils.readFileToString(moduleLastModifiedFile, Charset.defaultCharset());
+			return Long.valueOf(savedLastModified).equals(module.getFile().lastModified());
```
dkayiwa, line 549: These lines (the cache-folder path, the `.moduleLastModified` filename, and the `Long.valueOf(saved).equals(file.lastModified())` comparison) are duplicated from `getLibCacheFolderForModule` (lines 473, 476, 484). The two have to agree exactly or the optimization silently misbehaves: diverge one way and it never skips, the other way and it skips a module that actually changed. Nothing ties them together today, so a future edit to the marker scheme in one place won't flag the other. Worth pulling the marker-file resolution and the saved-vs-current comparison into a small shared helper that both call, so they can't drift apart. Non-blocking.

api/src/main/java/org/openmrs/module/ModuleClassLoader.java
```diff
@@ -231,41 +231,41 @@ private static List<URL> getUrls(final Module module) {
...
-					out.close();
+				finally {
+					try {
+						in.close();
+					}
+					catch (Exception e) { /* pass */}
+					try {
+						out.close();
+					}
+					catch (Exception e) { /* pass */}
```
ibacher, line 267: Nit, but while we're fixing this:

```suggestion
					try {
						in.close();
					}
					catch (Exception e) { /* pass */}
					finally {
						try {
							out.close();
						}
						catch (Exception e) { /* pass */}
					}
```

api/src/main/java/org/openmrs/module/ModuleClassLoader.java
```diff
@@ -231,41 +231,41 @@ private static List<URL> getUrls(final Module module) {
...
 		File tmpModuleDir = getLibCacheFolderForModule(module);
-		
+
 		//add module jar to classpath only if we are not in dev mode
 		if (devDir == null) {
 			File tmpModuleJar = new File(tmpModuleDir, module.getModuleId() + ".jar");
 			
-			if (!tmpModuleJar.exists()) {
+			// copy the module jar into that temporary folder, unless an up-to-date copy is already there
+			if (!(cacheUpToDate && tmpModuleJar.exists())) {
```
ibacher, line 248: So one danger here: `getLibCacheFolderForModule()` updates the `.moduleLastModified`, which with this new change means that if the expansion process fails, the module will consistently be a broken state, because a subsequent run will see the code as up-to-date. Implementing this code change seems to require us to only update `.moduleLastModified` _after_ it has successfully expanded; that way the module can't end up in a partially-expanded-but-corrupted state.

api/src/main/java/org/openmrs/module/ModuleClassLoader.java
```diff
@@ -231,41 +231,41 @@ private static List<URL> getUrls(final Module module) {
...
+				FileInputStream in = null;
+				FileOutputStream out = null;
```
ibacher, line 250: We should just adopt the try-with-resources version of this from `master`. Will make it easier to forward-port this too.

PR #6390, opened 2026-07-25
Fix duplicate savePatient and PersonMergeLog in mergePatients

### LINKED ISSUE #6336: mergePatients runs the merge tail twice (duplicate savePatient and PersonMergeLog)
### Summary
`PatientServiceImpl.mergePatients(Patient, Patient)` runs the final merge tail twice. `mergeAddresses(...)` — despite its name — executes the whole tail itself: it voids the non-preferred patient and person, calls `changeUserAssociations`, saves the preferred patient, and writes a `PersonMergeLog`. `mergePatients` then calls `mergeAddresses(...)` and repeats that same sequence inline.

### Effect
Every patient merge:
- saves the survivor twice (`savePatient(preferred)`), and
- writes **two** `PersonMergeLog` rows for a single merge (plus redundant `voidPatient` / `voidPerson` / `changeUserAssociations`).

### Where (origin/master, `api/.../api/impl/PatientServiceImpl.java`)
- `mergePatients` calls `mergeAddresses` at ~L584, then `savePatient(preferred)` at ~L603 and `savePersonMergeLog(...)` at ~L610.
- `mergeAddresses` itself does `savePatient(preferred)` at ~L971 and `savePersonMergeLog(...)` at ~L978.

Looks like a refactor left the tail in both `mergeAddresses` and its caller. The fix is presumably to keep the tail in one place (remove the duplicate). Merge is sensitive, so worth confirming there is no ordering reason for the earlier save before removing it, and adding a test asserting exactly one `PersonMergeLog` per merge.

### Related
Surfaced while reviewing #6332 (require Add Patient Identifiers when `savePatient` adds a transient identifier). In a manual-flush (web) context that PR's transient-identifier check would fire at the *first* save — the one inside `mergeAddresses` — so the merge-side privilege handling should be finalized on top of a single-save structure once this is resolved. No security exploit; filing as a correctness bug.

shubhada-desh07:
Fixes #6336

## Problem
`mergeAddresses` was doing more than just merging addresses. Besides the address merge, it was also voiding the non-preferred patient and person, calling `changeUserAssociations`, saving the preferred patient, and creating a `PersonMergeLog`.

The problem is that `mergePatients` already performs that same sequence immediately after calling `mergeAddresses`. As a result, the merge completion logic was executed twice for every patient merge, even though the merge itself only happened once.

In practice this meant every merge:
- saved the preferred patient twice,
- created two `PersonMergeLog` entries for a single merge,
- and made redundant `voidPatient`, `voidPerson`, and `changeUserAssociations` calls.

## Fix
Removed the duplicate merge completion logic from `mergeAddresses` so it now only handles address merging.

The completion logic remains in `mergePatients`, making it the single place responsible for finalizing the merge. The merge flow and ordering remain unchanged; only the duplicated work was removed.

While working in this area, I also noticed that `mergeDateOfDeath` copied `deathDate` but not `deathdateEstimated`. Since `mergeDateOfBirth` already transfers the equivalent estimated flag for birth dates, this looked like an accidental omission. I included that fix in the same PR because it was directly adjacent to the code being modified.

## Tests
- Added `mergePatients_shouldCallSaveVoidPatientAndVoidPersonExactlyOnceAndCreateLog`, which verifies that a merge creates exactly one `PersonMergeLog` and that the non-preferred patient and person are correctly voided.
- Added `mergePatients_shouldTransferDeathdateEstimatedToPreferred` to verify that `deathdateEstimated` is transferred during the merge.

Also ran the full `PatientServiceTest` suite after merging in the latest `master`; all 186 tests pass.


jwnasambu:
@claude review

jwnasambu (commented):
Please resolve the merge conflicts on this PR before we begin the code review.

PR #6391, opened 2026-07-25
TRUNK-6733: Use java-uuid-generator for entity UUIDs instead of UUID.randomUUID()

sudeshkar:

`BaseOpenmrsObject` generates its uuid in a field initialiser:

```java
private String uuid = UUID.randomUUID().toString();
```

Field initialisers run on every construction, which includes every Hibernate hydration of every
entity, where the generated value is then immediately overwritten by the uuid column read from the
database. `UUID.randomUUID()` draws from a single, JVM wide `SecureRandom` whose `engineNextBytes`
is synchronized, so every entity instantiation contends on one JVM wide monitor and the cost scales
with the number of rows a request loads.

A cryptographically strong source is not required here: uuids identify objects rather than protect
them. Access to a record is granted by privilege checks rather than by a uuid being unguessable,
and security sensitive tokens are generated separately (user activation keys use
`RandomStringUtils.randomAlphanumeric(20)`).

### What this does

All uuid generation in `api/src/main` now goes through a single helper, `org.openmrs.util.UuidUtil`,
which hands java-uuid-generator a `Random` backed by `ThreadLocalRandom`. Every thread therefore
draws from its own stream and none of them queue up behind a shared generator.

Routing everything through one helper also means the scheme can be changed in one place later.

### Three things worth flagging for reviewers

**`Generators.randomBasedGenerator()` on its own does not fix this.** With no argument, JUG falls
back to a shared `SecureRandom` singleton of its own, so the queue simply moves from one JVM wide
monitor to another. Measured at 32 threads it is 499.9 ns/uuid against a 539.8 ns/uuid baseline,
i.e. effectively no improvement. The random number generator that is passed in is what matters.

**The generator cannot be a captured `ThreadLocalRandom.current()`.** `Generators.randomBasedGenerator(Random)`
stores whatever it is handed in a field, and every thread then draws through that one reference,
whereas `current()` is documented as "Methods of this object should be called only by the current
thread, not by other threads." `UuidUtil.ThreadLocalRandomSource` is therefore a small `Random` that
calls `current()` on each draw. The JDK does the same thing internally:
`ThreadLocalRandom.ThreadLocalRandomProxy` is a private `Random` subclass whose `nextInt`,
`nextLong` and `nextDouble` delegate to `ThreadLocalRandom.current()`.

**Version 7 was measured and not used.** `TimeBasedEpochGenerator.construct()` holds a
`synchronized` block on a shared entropy buffer for every call, which reintroduces exactly the
shared monitor contention this ticket is about (111.9 ns/uuid at 32 threads). Version 4 also keeps
uuids from revealing when a record was created, which matters when they appear in URLs. `UuidUtil`
documents the single place to change should v7 be wanted later.

### Compatibility

Generated values are unchanged in form: canonical, 36 character, version 4 uuid strings that fit
the existing 38 character `uuid` column. Existing rows and their uuids are unaffected and there are
no liquibase changes in this PR, so no migration is required.

### Results

Measured over 96 million uuids generated on 32 threads, with a JFR recording using
`settings=profile` (10 ms monitor threshold, the same floor the ticket quotes):

| | before | after |
| --- | --- | --- |
| `jdk.JavaMonitorEnter` events | 49, all on `sun.security.provider.HashDrbg` | 0 |
| wall clock | 56.7 s | 0.6 s |
| uuid generation on the entity path | 539.8 ns/uuid | 5.9 ns/uuid |

A complete `new Location()` now costs 21.9 ns at 32 threads, less than 5% of what the uuid alone
used to cost.

### What the speed-up rests on, and how it is pinned

Supplying a generator only helps if java-uuid-generator actually draws every uuid from it, without
taking its monitor. Both hold in `RandomBasedGenerator` as of 5.2.0 — `generate()` contains no
`synchronized` block, and since `_secureRandom` is `(rnd instanceof SecureRandom)`, the supplied
generator takes the `_random.nextLong()` path rather than the `nextBytes(byte[])` one.

Neither property was pinned by anything, though, and both belong to a third-party class. A future
bump that reintroduced the lock, or changed the `_secureRandom` heuristic, would quietly serialise
generation on one monitor again with no test failing — the win would be gone and nothing would say
so. Two guards now cover it:

- `randomBasedGenerator_shouldDrawEveryUuidFromTheSuppliedGenerator` passes a counting `Random` and
  asserts generation made exactly two `nextLong()` calls and zero `nextBytes()` calls.
- `randomBasedGenerator_shouldNotSynchroniseOnTheSuppliedGenerator` holds the monitor of the
  supplied generator on one thread and asserts a second thread can still generate.

The second guard was checked against a deliberately synchronizing generator to confirm generation
blocks under it, so it discriminates rather than passing vacuously.

### Tests

`UuidUtilTest` covers canonical format and that values fit the uuid column, version 4 and IETF
variant, uniqueness across 200,000 generated uuids, uniqueness across 8 threads generating 25,000
each, and the two guards above. `BaseOpenmrsObjectTest` gains two cases covering `equals`/`hashCode`
for newly constructed objects and for entities rehydrated from the database.

`java-uuid-generator` 5.2.0 is Apache 2.0 licensed and its only compile dependency is the slf4j-api
version already in use. It has been added to the BOM and to `NOTICE.md`.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6733

Full `api` module suite on the rebased branch: 5217 tests, 0 failures, 0 errors, 46 skipped.

### DIFF
--- a/api/src/main/java/org/openmrs/util/UuidUtil.java
+++ b/api/src/main/java/org/openmrs/util/UuidUtil.java
@@ -1,0 +1,145 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.util;
+
+import java.security.SecureRandom;
+import java.util.Random;
+import java.util.SplittableRandom;
+import java.util.UUID;
+
+import com.fasterxml.uuid.Generators;
+import com.fasterxml.uuid.NoArgGenerator;
+
+/**
+ * Generates the uuids that identify OpenMRS objects.
+ * <p>
+ * All uuid generation should go through this class rather than calling {@link UUID#randomUUID()},
+ * which draws from a single, JVM-wide {@link SecureRandom} whose {@code nextBytes} is synchronized.
+ * Since {@link org.openmrs.BaseOpenmrsObject} generates a uuid in a field initializer, that one
+ * monitor is entered for every object Hibernate hydrates - even though the generated value is then
+ * immediately overwritten by the uuid read from the database - which makes it a significant point
+ * of contention on read paths that load many rows.
+ * <p>
+ * A cryptographically strong source is not needed here: uuids identify objects, they are not
+ * secrets. Access to a record is granted by privilege checks rather than by a uuid being
+ * unguessable, and security-sensitive tokens (such as user activation keys) are generated
+ * separately. This class therefore hands java-uuid-generator a thread-confined random number
+ * generator: each thread draws from its own {@link SplittableRandom} stream, so threads never queue
+ * up behind a shared generator. {@link SecureRandom} is read exactly once, to seed the root stream
+ * that all of the others are split from, which is what keeps uuids unique across threads, restarts
+ * and servers.
+ * <p>
+ * Generated values are canonical, 36 character, version 4 (random) uuid strings, exactly as before,
+ * so they still fit the 38 character uuid columns and existing uuids are unaffected. Should
+ * time-ordered (version 7) uuids ever be wanted for their better insert locality on the uuid
+ * indexes, the generator below is the only place that needs to change.
+ *
+ * @since 3.0.0
+ */
+public class UuidUtil {
+
+	/**
+	 * Seeded once per JVM from a cryptographically strong source and thereafter only ever split. The
+	 * cost of the strong source does not matter for a single draw, while the seed quality it provides
+	 * does: it is what stops the streams of different JVMs and servers from overlapping.
+	 * <p>
+	 * Reading {@link SecureRandom} exactly once and then letting it go is the whole point here rather
+	 * than an oversight, so the "save and re-use this Random" warning is suppressed.
+	 */
+	@SuppressWarnings("java:S2119")
+	private static final SplittableRandom ROOT = new SplittableRandom(new SecureRandom().nextLong());
+
+	/**
+	 * The random number stream a thread draws its uuids from.
+	 */
+	private static final ThreadLocal<SplittableRandom> STREAMS = ThreadLocal.withInitial(UuidUtil::newStream);
+
+	/**
+	 * java-uuid-generator's version 4 generator, drawing from the thread-confined streams above.
+	 * <p>
+	 * Two properties of {@code RandomBasedGenerator} are relied on here, and both are pinned by
+	 * {@code UuidUtilTest}, because supplying a generator buys nothing if either of them changes: it
+	 * does not synchronize on the generator it is handed, and it draws through
+	 * {@link Random#nextLong()} rather than {@link Random#nextBytes(byte[])} for anything that is not a
+	 * {@link SecureRandom}. Both hold as of 5.2.0.
+	 */
+	private static final NoArgGenerator UUID_GENERATOR = Generators.randomBasedGenerator(new ThreadConfinedRandom());
+
+	private UuidUtil() {
+	}
+
+	/**
+	 * Hands out the stream for a thread that is generating its first uuid. Splitting gives the stream
+	 * its own increment as well as its own starting point, so that streams stay independent of one
+	 * another rather than being offsets into a single shared sequence.
+	 * <p>
+	 * The root has to be guarded because {@link SplittableRandom#split()} advances the instance it is
+	 * called on, but the monitor is entered once per thread rather than once per uuid, and it is held
+	 * for nothing more than a few arithmetic operations.
+	 *
+	 * @return the calling thread's own random number stream
+	 */
+	private static SplittableRandom newStream() {
+		synchronized (ROOT) {
+			return ROOT.split();
+		}
+	}
+
+	/**
+	 * Generates a new uuid.
+	 *
+	 * @return a newly generated version 4 uuid
+	 * @since 3.0.0
+	 */
+	public static UUID newUuid() {
+		return UUID_GENERATOR.generate();
+	}
+
+	/**
+	 * Generates a new uuid in the canonical 36 character string form used by the uuid columns, for
+	 * example {@code 6e9a0b1c-6d3a-4f0e-8a1b-2c3d4e5f6a7b}.
+	 *
+	 * @return a newly generated version 4 uuid as a string
+	 * @since 3.0.0
+	 */
+	public static String newUuidString() {
+		return newUuid().toString();
+	}
+
+	/**
+	 * A {@link Random} that serves every draw from the calling thread's own stream. It holds no state
+	 * of its own, so handing the single generator instance to every thread is safe and, unlike the
+	 * {@link SecureRandom} java-uuid-generator would use by default, involves no locking.
+	 */
+	private static final class ThreadConfinedRandom extends Random {
+
+		private static final long serialVersionUID = 1L;
+
+		@Override
+		public long nextLong() {
+			return STREAMS.get().nextLong();
+		}
+
+		@Override
+		public int nextInt() {
+			return STREAMS.get().nextInt();
+		}
+
+		@Override
+		public void nextBytes(byte[] bytes) {
+			STREAMS.get().nextBytes(bytes);
+		}
+
+		@Override
+		protected int next(int bits) {
+			return STREAMS.get().nextInt() >>> (Integer.SIZE - bits);
+		}
+	}
+}
--- a/api/src/main/java/org/openmrs/util/UuidUtil.java
+++ b/api/src/main/java/org/openmrs/util/UuidUtil.java
@@ -9,10 +9,9 @@
  */
 package org.openmrs.util;
 
-import java.security.SecureRandom;
 import java.util.Random;
-import java.util.SplittableRandom;
 import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
 
 import com.fasterxml.uuid.Generators;
 import com.fasterxml.uuid.NoArgGenerator;
@@ -20,74 +19,24 @@ import com.fasterxml.uuid.NoArgGenerator;
 /**
  * Generates the uuids that identify OpenMRS objects.
  * <p>
- * It is preferable to use this class to generate random UUIDs, as the JDK's default implementation can be slow under a high volume of requests.
- * <p>
- * A cryptographically strong source is not needed here: uuids identify objects, they are not
- * secrets. Access to a record is granted by privilege checks rather than by a uuid being
- * unguessable, and security-sensitive tokens (such as user activation keys) are generated
- * separately. This class therefore hands java-uuid-generator a thread-confined random number
- * generator: each thread draws from its own {@link SplittableRandom} stream, so threads never queue
- * up behind a shared generator. {@link SecureRandom} is read exactly once, to seed the root stream
- * that all of the others are split from, which is what keeps uuids unique across threads, restarts
- * and servers.
- * <p>
- * Generated values are canonical, 36 character, version 4 (random) uuid strings, exactly as before,
- * so they still fit the 38 character uuid columns and existing uuids are unaffected. Should
- * time-ordered (version 7) uuids ever be wanted for their better insert locality on the uuid
- * indexes, the generator below is the only place that needs to change.
+ * It is preferable to use this class to generate random UUIDs, as the JDK's default implementation
+ * can be slow under a high volume of requests. The uuids it generates are not suitable for
+ * security-sensitive purposes.
  *
  * @since 3.0.0, 2.9.0
  */
 public class UuidUtil {
 
-	/**
-	 * Seeded once per JVM from a cryptographically strong source and thereafter only ever split. The
-	 * cost of the strong source does not matter for a single draw, while the seed quality it provides
-	 * does: it is what stops the streams of different JVMs and servers from overlapping.
-	 * <p>
-	 * Reading {@link SecureRandom} exactly once and then letting it go is the whole point here rather
-	 * than an oversight, so the "save and re-use this Random" warning is suppressed.
-	 */
-	@SuppressWarnings("java:S2119")
-	private static final SplittableRandom ROOT = new SplittableRandom(new SecureRandom().nextLong());
-
-	/**
-	 * The random number stream a thread draws its uuids from.
-	 */
-	private static final ThreadLocal<SplittableRandom> STREAMS = ThreadLocal.withInitial(UuidUtil::newStream);
-
-	/**
-	 * java-uuid-generator's version 4 generator, drawing from the thread-confined streams above.
-	 * <p>
-	 * Two properties of {@code RandomBasedGenerator} are relied on here, and both are pinned by
-	 * {@code UuidUtilTest}, because supplying a generator buys nothing if either of them changes: it
-	 * does not synchronize on the generator it is handed, and it draws through
-	 * {@link Random#nextLong()} rather than {@link Random#nextBytes(byte[])} for anything that is not a
-	 * {@link SecureRandom}. Both hold as of 5.2.0.
-	 */
-	private static final NoArgGenerator UUID_GENERATOR = Generators.randomBasedGenerator(new ThreadConfinedRandom());
+	private static final NoArgGenerator UUID_GENERATOR = Generators.randomBasedGenerator(new ThreadLocalRandomSource());
 
 	private UuidUtil() {
-	}
-
-	/**
-	 * Used to create a new per-thread random generator. Needs to be
-	 * synchronized because {@code SplittableRandom} itself is not
-	 * thread-safe.
-	 *
-	 * @return the calling thread's own random number stream
-	 */
-	private static SplittableRandom newStream() {
-		synchronized (ROOT) {
-			return ROOT.split();
-		}
 	}
 
 	/**
 	 * Generates a new uuid.
 	 *
 	 * @return a newly generated version 4 uuid
-	 * @since 3.0.0
+	 * @since 3.0.0, 2.9.0
 	 */
 	public static UUID newUuid() {
 		return UUID_GENERATOR.generate();
@@ -98,39 +47,40 @@ public class UuidUtil {
 	 * example {@code 6e9a0b1c-6d3a-4f0e-8a1b-2c3d4e5f6a7b}.
 	 *
 	 * @return a newly generated version 4 uuid as a string
-	 * @since 3.0.0
+	 * @since 3.0.0, 2.9.0
 	 */
 	public static String newUuidString() {
 		return newUuid().toString();
 	}
 
 	/**
-	 * A {@link Random} that serves every draw from the calling thread's own stream. It holds no state
-	 * of its own, so handing the single generator instance to every thread is safe and, unlike the
-	 * {@link SecureRandom} java-uuid-generator would use by default, involves no locking.
+	 * Presents {@link ThreadLocalRandom} as the {@link Random} that java-uuid-generator takes. A single
+	 * generator is shared by every thread, so each draw goes through
+	 * {@link ThreadLocalRandom#current()} rather than a stored instance, which is only valid on the
+	 * thread that obtained it.
 	 */
-	private static final class ThreadConfinedRandom extends Random {
+	private static final class ThreadLocalRandomSource extends Random {
 
 		private static final long serialVersionUID = 1L;
 
 		@Override
 		public long nextLong() {
-			return STREAMS.get().nextLong();
+			return ThreadLocalRandom.current().nextLong();
 		}
 
 		@Override
 		public int nextInt() {
-			return STREAMS.get().nextInt();
+			return ThreadLocalRandom.current().nextInt();
 		}
 
 		@Override
 		public void nextBytes(byte[] bytes) {
-			STREAMS.get().nextBytes(bytes);
+			ThreadLocalRandom.current().nextBytes(bytes);
 		}
 
 		@Override
 		protected int next(int bits) {
-			return STREAMS.get().nextInt() >>> (Integer.SIZE - bits);
+			return ThreadLocalRandom.current().nextInt() >>> (Integer.SIZE - bits);
 		}
 	}
 }
--- a/api/src/main/java/org/openmrs/util/UuidUtil.java
+++ b/api/src/main/java/org/openmrs/util/UuidUtil.java
@@ -36,7 +36,7 @@ import com.fasterxml.uuid.NoArgGenerator;
  * time-ordered (version 7) uuids ever be wanted for their better insert locality on the uuid
  * indexes, the generator below is the only place that needs to change.
  *
- * @since 3.0.0
+ * @since 3.0.0, 2.9.0
  */
 public class UuidUtil {
 
--- a/api/src/main/java/org/openmrs/util/UuidUtil.java
+++ b/api/src/main/java/org/openmrs/util/UuidUtil.java
@@ -71,13 +71,9 @@ public class UuidUtil {
 	}
 
 	/**
-	 * Hands out the stream for a thread that is generating its first uuid. Splitting gives the stream
-	 * its own increment as well as its own starting point, so that streams stay independent of one
-	 * another rather than being offsets into a single shared sequence.
-	 * <p>
-	 * The root has to be guarded because {@link SplittableRandom#split()} advances the instance it is
-	 * called on, but the monitor is entered once per thread rather than once per uuid, and it is held
-	 * for nothing more than a few arithmetic operations.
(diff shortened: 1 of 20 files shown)


jwnasambu:
@claude review

api/src/main/java/org/openmrs/util/UuidUtil.java
```diff
@@ -0,0 +1,136 @@
...
+	 */
+	@SuppressWarnings("java:S2119")
+	private static final SplittableRandom ROOT = new SplittableRandom(new SecureRandom().nextLong());
+
+	/**
+	 * The random number stream a thread draws its uuids from.
+	 */
+	private static final ThreadLocal<SplittableRandom> STREAMS = ThreadLocal.withInitial(UuidUtil::newStream);
+
+	private static final NoArgGenerator UUID_GENERATOR = Generators.randomBasedGenerator(new ThreadConfinedRandom());
```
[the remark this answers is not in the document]
sudeshkar: Good question to raise rather than assume, so I went and checked the 5.2.0 sources jar directly. **The synchronization is not there in 5.2.0** — `RandomBasedGenerator.generate()` is:

```java
@Override
public UUID generate()
{
    // 14-Oct-2010, tatu: Surprisingly, variant for reading byte array is
    //   tad faster for SecureRandom... so let's use that then

    long r1, r2;

    if (_secureRandom) {
        final byte[] buffer = new byte[16];
        _random.nextBytes(buffer);
        r1 = _toLong(buffer, 0);
        r2 = _toLong(buffer, 1);
    } else {
        r1 = _random.nextLong();
        r2 = _random.nextLong();
    }
    return UUIDUtil.constructUUID(UUIDType.RANDOM_BASED, r1, r2);
}
```

No `synchronized` block, and grepping the whole library confirms `RandomBasedGenerator` contains none at all — the `synchronized` uses live in `NameBasedGenerator` (on `_digester`), `TimeBasedEpochGenerator` (on `_lastEntropy`, which is the v7 contention I flagged in the description), `UUIDTimer`, `LazyRandom` and `EthernetAddress`. So `ThreadConfinedRandom`'s per-thread dispatch via `STREAMS.get()` is reached concurrently, and the measured numbers stand.

Worth noting the second half of that snippet too: `_secureRandom` is set in the constructor as `(rnd instanceof SecureRandom)`. `ThreadConfinedRandom` extends `java.util.Random`, so it is `false`, and generation takes the `_random.nextLong()` path — which is exactly the method `ThreadConfinedRandom` overrides for the hot path.

**That said, your underlying point was right even though the specific mechanism was not, so I have fixed it.** The entire performance case rested on two properties of a third-party class that nothing in the build was pinning. If a future JUG bump reintroduced the lock, or changed the `_secureRandom` heuristic so generation went through `nextBytes`, the win would silently evaporate and no test would fail. That is a bad thing to leave resting on a library implementation detail.

Two guards added to `UuidUtilTest`:

- `randomBasedGenerator_shouldDrawEveryUuidFromTheSuppliedGenerator` — passes a counting `Random` and asserts generation made exactly two `nextLong()` calls and zero `nextBytes()` calls, pinning both that our generator is the source and that the `nextLong` path is taken.
- `randomBasedGenerator_shouldNotSynchroniseOnTheSuppliedGenerator` — holds the monitor of the supplied generator on one thread and asserts a second thread can still generate. This is a direct test of your hypothesis: if `generate()` ever wraps in `synchronized (_random)`, it fails.

I checked that second guard is not vacuous rather than trusting a green tick — ran the same structure against a deliberately synchronizing generator and confirmed generation blocks, so the assertion genuinely discriminates.

The assumption and the JUG version it was verified against are now recorded on the `UUID_GENERATOR` field, so whoever bumps the dependency knows what to re-check.

api/src/main/java/org/openmrs/util/UuidUtil.java
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Generates the uuids that identify OpenMRS objects.
+ * <p>
+ * All uuid generation should go through this class rather than calling {@link UUID#randomUUID()},
+ * which draws from a single, JVM-wide {@link SecureRandom} whose {@code nextBytes} is synchronized.
+ * Since {@link org.openmrs.BaseOpenmrsObject} generates a uuid in a field initializer, that one
+ * monitor is entered for every object Hibernate hydrates - even though the generated value is then
+ * immediately overwritten by the uuid read from the database - which makes it a significant point
+ * of contention on read paths that load many rows.
```
ibacher: Point here: not all UUID generation is random; in some cases, we use other UUID types for various reasons. It is true that this should be preferred for _random_ UUID generation, which is the most common.

```suggestion
 * It is preferable to use this class to generate random UUIDs, as the JDK's default implementation can be slow under a high volume of requests.
```

api/src/main/java/org/openmrs/util/UuidUtil.java
```diff
@@ -0,0 +1,145 @@
...
+ * <p>
+ * A cryptographically strong source is not needed here: uuids identify objects, they are not
+ * secrets. Access to a record is granted by privilege checks rather than by a uuid being
+ * unguessable, and security-sensitive tokens (such as user activation keys) are generated
+ * separately. This class therefore hands java-uuid-generator a thread-confined random number
+ * generator: each thread draws from its own {@link SplittableRandom} stream, so threads never queue
+ * up behind a shared generator. {@link SecureRandom} is read exactly once, to seed the root stream
+ * that all of the others are split from, which is what keeps uuids unique across threads, restarts
+ * and servers.
```
ibacher: This should either be removed or reduced to a comment noting that the default generated UUIDs should not be used for security-sensitive purposes.

api/src/main/java/org/openmrs/util/UuidUtil.java
```diff
@@ -0,0 +1,145 @@
...
+ * <p>
+ * Generated values are canonical, 36 character, version 4 (random) uuid strings, exactly as before,
+ * so they still fit the 38 character uuid columns and existing uuids are unaffected. Should
+ * time-ordered (version 7) uuids ever be wanted for their better insert locality on the uuid
+ * indexes, the generator below is the only place that needs to change.
```
ibacher: Remove this.

api/src/main/java/org/openmrs/util/UuidUtil.java
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * @since 3.0.0
```
ibacher:
```suggestion
 * @since 3.0.0, 2.9.0
```

api/src/main/java/org/openmrs/util/UuidUtil.java
```diff
@@ -0,0 +1,145 @@
...
+public class UuidUtil {
+
+	/**
+	 * Seeded once per JVM from a cryptographically strong source and thereafter only ever split. The
+	 * cost of the strong source does not matter for a single draw, while the seed quality it provides
+	 * does: it is what stops the streams of different JVMs and servers from overlapping.
+	 * <p>
+	 * Reading {@link SecureRandom} exactly once and then letting it go is the whole point here rather
+	 * than an oversight, so the "save and re-use this Random" warning is suppressed.
+	 */
```
ibacher: I don't think this comment adds much value

api/src/main/java/org/openmrs/util/UuidUtil.java
```diff
@@ -0,0 +1,145 @@
...
+	@SuppressWarnings("java:S2119")
+	private static final SplittableRandom ROOT = new SplittableRandom(new SecureRandom().nextLong());
+
+	/**
+	 * The random number stream a thread draws its uuids from.
+	 */
```
ibacher: Ditto

api/src/main/java/org/openmrs/util/UuidUtil.java
```diff
@@ -0,0 +1,145 @@
...
+
+	/**
+	 * java-uuid-generator's version 4 generator, drawing from the thread-confined streams above.
+	 * <p>
+	 * Two properties of {@code RandomBasedGenerator} are relied on here, and both are pinned by
+	 * {@code UuidUtilTest}, because supplying a generator buys nothing if either of them changes: it
+	 * does not synchronize on the generator it is handed, and it draws through
+	 * {@link Random#nextLong()} rather than {@link Random#nextBytes(byte[])} for anything that is not a
+	 * {@link SecureRandom}. Both hold as of 5.2.0.
+	 */
```
ibacher: Ditto

api/src/main/java/org/openmrs/util/UuidUtil.java
```diff
@@ -0,0 +1,145 @@
...
+	 * its own increment as well as its own starting point, so that streams stay independent of one
+	 * another rather than being offsets into a single shared sequence.
+	 * <p>
+	 * The root has to be guarded because {@link SplittableRandom#split()} advances the instance it is
+	 * called on, but the monitor is entered once per thread rather than once per uuid, and it is held
+	 * for nothing more than a few arithmetic operations.
+	 *
+	 * @return the calling thread's own random number stream
+	 */
+	private static SplittableRandom newStream() {
```
ibacher: It seems like it would be simpler to just mark this method itself as synchronized

api/src/main/java/org/openmrs/util/UuidUtil.java
```diff
@@ -0,0 +1,145 @@
...
+	/**
+	 * Hands out the stream for a thread that is generating its first uuid. Splitting gives the stream
+	 * its own increment as well as its own starting point, so that streams stay independent of one
+	 * another rather than being offsets into a single shared sequence.
+	 * <p>
+	 * The root has to be guarded because {@link SplittableRandom#split()} advances the instance it is
+	 * called on, but the monitor is entered once per thread rather than once per uuid, and it is held
+	 * for nothing more than a few arithmetic operations.
+	 *
+	 * @return the calling thread's own random number stream
```
ibacher:
```suggestion
	/**
	 * Used to create a new per-thread random generator. Needs to be
	 * synchronized because {@code SplittableRandom} itself is not
	 * thread-safe.
	 *
	 * @return the calling thread's own random number stream
```

api/src/main/java/org/openmrs/util/UuidUtil.java
ibacher, line 1: There's a lot of ceremony in this class to create thread-local random generators and it's not clear to me that this has any advantages over using `ThreadLocalRandom` that is already built into the JVM and seems like it would do the same job?

spotbugs-exclude.xml
```diff
@@ -70,4 +70,20 @@
...
 	</Match>
+
+	<!--
+		Uuids identify objects, they do not protect them: access to a record is granted by
+		privilege checks rather than by a uuid being unguessable. UuidUtil therefore generates
+		them from a per thread random number generator on purpose, rather than from the single
+		shared SecureRandom that every caller of UUID.randomUUID() has to queue up behind.
+		SecureRandom is deliberately read only once, to seed the root stream, which is what
+		DMI_RANDOM_USED_ONLY_ONCE reports.
+	-->
```
ibacher: This comment could also be substantially shortened.

ibacher (changes requested):
Thanks. I think the largely looks ok. 1 major comment (please use `ThreadLocalRandom` unless there's a reason we can't) and several minor comments to clean-up the over-commenting in this PR.

api/src/main/java/org/openmrs/util/UuidUtil.java
sudeshkar: Switched to `ThreadLocalRandom` - you're right, it does the same job. `ROOT`, `STREAMS`, `newStream()` and the one-off `SecureRandom` seeding are all gone, and the class is down from 145 lines to 86.

One detail worth recording, since it is why a small `Random` subclass is still here rather than nothing at all: the generator cannot be a captured `ThreadLocalRandom.current()`. `Generators.randomBasedGenerator(Random)` stores whatever it is handed in a field, and every thread then draws through that one reference, but `current()` is documented as "Methods of this object should be called only by the current thread, not by other threads." So `ThreadLocalRandomSource` is just a `Random` that calls `current()` on each draw. The JDK does exactly this internally - `ThreadLocalRandom.ThreadLocalRandomProxy` is a private `Random` subclass whose `nextInt`/`nextLong`/`nextDouble` delegate to `ThreadLocalRandom.current()`.

(On JDK 21 a stored reference does happen to work, because `nextSeed()` folds in `t.threadId()` and so still separates threads whose seed was never initialised, but that is an implementation detail rather than something to depend on.)

The speed-up is unaffected by the change.

api/src/main/java/org/openmrs/util/UuidUtil.java
sudeshkar: Applied, just re-wrapped to the 100 character comment width the formatter uses so spotless leaves it alone.

api/src/main/java/org/openmrs/util/UuidUtil.java
sudeshkar: Reduced to one sentence on the class javadoc: "The uuids it generates are not suitable for security-sensitive purposes."

api/src/main/java/org/openmrs/util/UuidUtil.java
sudeshkar: Removed.

api/src/main/java/org/openmrs/util/UuidUtil.java
sudeshkar: Applied, and on `newUuid()` and `newUuidString()` as well.

api/src/main/java/org/openmrs/util/UuidUtil.java
sudeshkar: Gone - the field itself no longer exists now that draws come from `ThreadLocalRandom`.

api/src/main/java/org/openmrs/util/UuidUtil.java
sudeshkar: Also gone with the switch to `ThreadLocalRandom`.

api/src/main/java/org/openmrs/util/UuidUtil.java
sudeshkar: Removed - the field is a single uncommented line now. The two assumptions it described are still covered by `UuidUtilTest`, which seemed the better place for them than a comment restating what the tests assert.

api/src/main/java/org/openmrs/util/UuidUtil.java
sudeshkar: The method is gone entirely - `ThreadLocalRandom` removes the need to hand out per-thread streams at all, so there is nothing left to guard.

api/src/main/java/org/openmrs/util/UuidUtil.java
sudeshkar: Took this one, then the method went away with the rest of the `SplittableRandom` machinery, so the javadoc went with it.

spotbugs-exclude.xml
sudeshkar: Down to a single line. Also dropped the `DMI_RANDOM_USED_ONLY_ONCE` entry, which no longer applies now that `SecureRandom` is not read at all.

api/src/test/java/org/openmrs/BaseOpenmrsObjectTest.java
```diff
@@ -418,6 +422,53 @@ public void setId(Integer id) {
...
+		Location location = session.get(Location.class, 1);
+		String uuid = location.getUuid();
+
+		//drop the first level cache so that the row is hydrated into a second instance
+		session.clear();
+		Location reloaded = session.get(Location.class, 1);
+
+		assertNotSame(location, reloaded);
+		assertEquals(uuid, reloaded.getUuid());
+		assertTrue(location.equals(reloaded));
```
ibacher:
```suggestion
		assertEquals(location, reloaded);
```

api/src/test/java/org/openmrs/BaseOpenmrsObjectTest.java
```diff
@@ -418,6 +422,53 @@ public void setId(Integer id) {
...
+		BaseOpenmrsObject o = new BaseOpenmrsObjectMock();
+		BaseOpenmrsObject other = new BaseOpenmrsObjectMock();
+
+		assertEquals(4, UUID.fromString(o.getUuid()).version());
+		assertTrue(o.getUuid().length() <= 38, "the uuid has to fit the uuid column");
+
+		//two objects that were just constructed are distinct, and stay distinct as far as
+		//equals and hashCode are concerned
+		assertNotEquals(o.getUuid(), other.getUuid());
+		assertFalse(o.equals(other));
```
ibacher:
```suggestion
		assertNotEquals(o, other);
```

api/src/test/java/org/openmrs/BaseOpenmrsObjectTest.java
```diff
@@ -418,6 +422,53 @@ public void setId(Integer id) {
...
+
+		assertNotSame(location, reloaded);
+		assertEquals(uuid, reloaded.getUuid());
+		assertTrue(location.equals(reloaded));
+		assertEquals(location.hashCode(), reloaded.hashCode());
+
+		//a newly constructed object of the same type does not collide with the hydrated one
+		Location constructed = new Location();
+		assertNotEquals(uuid, constructed.getUuid());
+		assertFalse(location.equals(constructed));
```
ibacher:
```suggestion
		assertNotEquals(location, constructed);
```

api/src/test/java/org/openmrs/util/UuidUtilTest.java
```diff
@@ -0,0 +1,225 @@
...
+		assertNotEquals(UuidUtil.newUuid(), UuidUtil.newUuid());
+	}
+
+	/**
+	 * Supplying a generator buys nothing unless every uuid is actually drawn from it, so this pins that
+	 * {@code RandomBasedGenerator} draws through {@link Random#nextLong()} rather than from a source of
+	 * its own.
+	 */
+	@Test
+	public void randomBasedGenerator_shouldDrawEveryUuidFromTheSuppliedGenerator() {
```
ibacher: This seems like a silly test for us to have. We're testing the contract of a library, which is not really something that's saving us from anything sensible.

sudeshkar: Removed. I also dropped the neighbouring `randomBasedGenerator_shouldNotSynchroniseOnTheSuppliedGenerator` test for the same reason — both were pinning java-uuid-generator's behaviour rather than ours. What's left in `UuidUtilTest` is uniqueness (single- and multi-threaded), format and version validity.

sudeshkar:
All review comments are addressed and every thread is resolved. The branch is also rebased onto current master (it had drifted 9 commits behind; `bom/pom.xml` was the only overlap and it rebased cleanly).

`UuidUtilTest` and `BaseOpenmrsObjectTest` pass locally on the rebased branch (27 tests).

@ibacher when you get a chance, this needs a maintainer to approve the workflow runs — Build with Maven, CodeQL and OWASP Dependency Check are all sitting at `action_required`, so only SonarCloud has run on the recent commits (it passed). Worth noting the `dependencycheck_test` failure on the earlier commit was `mysql-connector-j-9.7.0` CVEs, which is pre-existing on master and unrelated to this change. A re-review to clear the requested-changes status would also be appreciated.

api/src/test/java/org/openmrs/util/UuidUtilTest.java
```diff
@@ -0,0 +1,137 @@
...
+		}
+	}
+
+	/**
+	 * @see UuidUtil#newUuidString()
+	 */
+	@Test
+	public void newUuidString_shouldNotRepeatItselfOverALargeNumberOfUuids() {
+		int count = 200000;
+		Set<String> uuids = new HashSet<>(count * 2);
```
ibacher, line 83: nit: there's no reason to apply a scaling factor here; there can only ever be `count` elements in this list.

api/src/test/java/org/openmrs/util/UuidUtilTest.java
```diff
@@ -0,0 +1,137 @@
...
+	 * values have to fit into.
+	 */
+	private static final int UUID_COLUMN_LENGTH = 38;
+
+	/**
+	 * @see UuidUtil#newUuidString()
+	 */
+	@Test
+	public void newUuidString_shouldReturnACanonicalUuidThatFitsTheUuidColumn() {
+		for (int i = 0; i < 1000; i++) {
```
ibacher, line 53: I don't know that these tests are functionally useful run 1000 times (this and below). Either the code is deterministic enough that a single check would catch the error or the window of 1000 is far to small to reasonably fail.

api/src/test/java/org/openmrs/util/UuidUtilTest.java
```diff
@@ -0,0 +1,137 @@
...
+			String uuid = UuidUtil.newUuidString();
+
+			assertTrue(CANONICAL_VERSION_4_UUID.matcher(uuid).matches(), uuid + " is not a canonical uuid");
```
ibacher, line 56: The test states it's for a "Canonical UUID", but here we match against a type 4 UUID. That seems inconsistent.

api/src/test/java/org/openmrs/util/UuidUtilTest.java
```diff
@@ -0,0 +1,137 @@
...
+			assertEquals(36, uuid.length());
+			assertTrue(uuid.length() <= UUID_COLUMN_LENGTH);
```
ibacher, line 58: This assertion is always redundant because 36 <= UUID_COLUMN_LENGTH and we already verified uuid.length() == 36.

api/src/test/java/org/openmrs/util/UuidUtilTest.java
```diff
@@ -0,0 +1,137 @@
...
+			// the string form has to survive a round trip through java.util.UUID unchanged
+			assertEquals(uuid, UUID.fromString(uuid).toString());
```
ibacher, line 60: Strictly, the only assertion we need here is:

```
// UUID.fromString() throws if this string is not a valid UUID
UUID.fromString(uuid);
```

Every other check here is strictly redundant (and round-tripping through `java.util.UUID` is outside of the scope of this test.

api/src/test/java/org/openmrs/BaseOpenmrsObjectTest.java
```diff
@@ -418,6 +422,53 @@ public void setId(Integer id) {
...
 
+	/**
+	 * @see BaseOpenmrsObject#getUuid()
+	 */
+	@Test
+	public void getUuid_shouldGiveEveryNewlyConstructedObjectItsOwnValidUuid() {
+		BaseOpenmrsObject o = new BaseOpenmrsObjectMock();
+		BaseOpenmrsObject other = new BaseOpenmrsObjectMock();
+
+		assertEquals(4, UUID.fromString(o.getUuid()).version());
```
ibacher, line 433: This test enforces a rule that we neither have enforced before nor that is strictly relevant to these changes. Once again, it is simply sufficient to show `UUID.fromString()` does not throw to show this UUID is valie.

api/src/test/java/org/openmrs/BaseOpenmrsObjectTest.java
```diff
@@ -418,6 +422,53 @@ public void setId(Integer id) {
...
+		assertTrue(o.getUuid().length() <= 38, "the uuid has to fit the uuid column");
+
+		//two objects that were just constructed are distinct, and stay distinct as far as
+		//equals and hashCode are concerned
+		assertNotEquals(o.getUuid(), other.getUuid());
+		assertNotEquals(o, other);
+		assertEquals(o.getUuid().hashCode(), o.hashCode());
```
ibacher, line 440: Both of these assertions are not germane to what this tests claims to be about. While they pass, these are different properties (of the `equals()` and `hashCode()` implementation details) that a test about `getUuid()` should not make any assertions about.

api/src/test/java/org/openmrs/BaseOpenmrsObjectTest.java
```diff
@@ -418,6 +422,53 @@ public void setId(Integer id) {
...
+	}
+
+	/**
+	 * Hibernate overwrites the uuid generated in the field initialiser with the one read from the
+	 * database, so hydrated objects have to keep comparing equal by their stored uuid.
+	 *
+	 * @see BaseOpenmrsObject#equals(Object)
+	 */
+	@Test
+	public void equals_shouldUseTheUuidReadFromTheDatabaseForHydratedObjects() {
```
ibacher, line 450: This test is an assertion about internals not germane to the code in this PR. Please remove it.

api/src/test/java/org/openmrs/BaseOpenmrsObjectTest.java
```diff
@@ -418,6 +422,53 @@ public void setId(Integer id) {
...
+		String uuid = location.getUuid();
+
+		//drop the first level cache so that the row is hydrated into a second instance
+		session.clear();
+		Location reloaded = session.get(Location.class, 1);
+
+		assertNotSame(location, reloaded);
+		assertEquals(uuid, reloaded.getUuid());
+		assertEquals(location, reloaded);
+		assertEquals(location.hashCode(), reloaded.hashCode());
```
ibacher, line 464: For what its worth, these three properties being true is, once again, three separate concerns that no single unit test should be enforcing. All this does it make it harder to refactor code with no gains in correctness.

ibacher (approved):

PR #6392, opened 2026-07-26
TRUNK-6727: Exclude descendant locations when the ancestor itself is retired

sudeshkar:

`HibernateLocationDAO.getLocations(LocationSearchCriteria)` supports a "descendant of" filter. When `includeRetired` is `false`, retired locations are meant to be pruned from the hierarchy together with their subtrees, but the retired state of the ancestor passed as the search root was never considered: the recursive CTE was seeded with the *children* of the ancestor, so a retired ancestor still returned all of its descendants.

The in-memory implementation that preceded the TRUNK-6681 performance rewrite excluded them, because it walked parent pointers upward and checked the retired flag of every node before matching the ancestor. The DB-side rewrite dropped that behaviour, and because this edge case was untested the divergence went unnoticed.

This seeds the recursive CTE with the ancestor row itself, so the existing `AND retired = false` filter prunes its whole subtree when the ancestor is retired, and drops the ancestor from the returned ids to keep the existing root-excluded semantics:

```sql
WITH RECURSIVE descendants (location_id) AS (
    SELECT location_id FROM location WHERE location_id = :locationId AND retired = false
    UNION ALL
    SELECT l.location_id FROM location l
    INNER JOIN descendants d ON l.parent_location = d.location_id AND retired = false
) SELECT location_id FROM descendants WHERE location_id <> :locationId
```

No extra query is issued, and the `includeRetired = true` path is unchanged: the filter disappears from both branches of the CTE and the ancestor is still excluded from the result.

| `descendantOfLocation` | `includeRetired` | before | after |
| --- | --- | --- | --- |
| retired | `false` | all non-retired descendants | empty list |
| retired | `true` | all descendants | all descendants (unchanged) |
| not retired | either | unchanged | unchanged |

### The CTE also needed a query space

Splitting the test during review (see below) surfaced a second problem, without which the fix above does not reliably hold.

`getDescendantIds` runs through `session.createNativeQuery(...)`, and a native query does not auto-flush on its own. `retireLocation` leaves a pending `UPDATE` in the session — the `INSERT`s are already out, since `location_id` is `IDENTITY`-generated — so the CTE read the ancestor as `retired = false` and walked its subtree anyway. The criteria query at the end of `getLocations` *does* flush, which is why the original single test passed: its `includeRetired = true` half flushed the session before the `includeRetired = false` half ran. On its own, the `includeRetired = false` case failed.

Declaring the location table as a query space makes `FlushMode.AUTO` apply to the native query as well:

```java
// a native query does not auto-flush on its own, so the location table is declared as a query
// space to make pending changes (a location retired earlier in the same transaction) visible
return session.createNativeQuery(cteSql, Integer.class).addSynchronizedEntityClass(Location.class)
        .setParameter("locationId", criteria.getDescendantOfLocation().getLocationId()).list();
```

Any caller that retires a location and then queries its descendants within the same transaction was reading stale data before this.

### Tests

- `getLocations_withDescendantOf_shouldReturnDescendantsOfARetiredAncestorWhenIncludeRetiredIsTrue` — both descendants are returned, and the retired ancestor is explicitly asserted *not* to be among them, pinning the `WHERE location_id <> :locationId` root exclusion.

The second test fails without the DAO change, on either half of it.

### Documentation

### Backports

2.9.x #6393 and 2.8.x #6394. Both branches still resolve descendants in memory (TRUNK-6681 has not landed there yet) and therefore already behave correctly, so those backports carry the regression tests and the javadoc only, to pin the behaviour for when the recursive-CTE lookup gets backported. The test code is identical across all three branches.

## Issue I worked on

I ran the full `api` module suite (`./mvnw -pl api -am test`) rather than `./mvnw clean package`, since the change is confined to that module: 5210 tests, 0 failures, 0 errors, 46 skipped.

### DIFF
--- a/api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
+++ b/api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
@@ -48,7 +48,8 @@ public class LocationSearchCriteria {
 
 	/**
 	 * @param descendantOfLocation the ancestor location; only its descendants are returned (root
-	 *            excluded)
+	 *            excluded). When retired locations are excluded, a retired location prunes its whole
+	 *            subtree, so a retired ancestor contributes no descendants at all.
 	 */
 	public void setDescendantOfLocation(Location descendantOfLocation) {
 		this.descendantOfLocation = descendantOfLocation;
@@ -97,7 +98,9 @@ public class LocationSearchCriteria {
 	}
 
 	/**
-	 * @return the ancestor location; only descendants of that location are returned (root excluded)
+	 * @return the ancestor location; only descendants of that location are returned (root excluded).
+	 *         When retired locations are excluded, a retired location prunes its whole subtree, so a
+	 *         retired ancestor contributes no descendants at all.
 	 */
 	public Location getDescendantOfLocation() {
 		return descendantOfLocation;
--- a/api/src/test/java/org/openmrs/api/LocationServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/LocationServiceTest.java
@@ -1272,6 +1272,105 @@ public class LocationServiceTest extends BaseContextSensitiveTest {
 	 * @see LocationService#getLocations(LocationSearchCriteria)
 	 */
 	@Test
+	public void getLocations_withDescendantOf_shouldReturnDescendantsOfARetiredAncestorWhenIncludeRetiredIsTrue() {
+		LocationService ls = Context.getLocationService();
+		List<Location> tree = saveTreeUnderRetiredAncestor(ls);
+		Location retiredRoot = tree.get(0);
+		Location child = tree.get(1);
+		Location grandchild = tree.get(2);
+
+		LocationSearchCriteria criteria = new LocationSearchCriteria();
+		criteria.setDescendantOfLocation(retiredRoot);
+		criteria.setIncludeRetired(true);
+		List<Location> result = ls.getLocations(criteria);
+
+		assertEquals(2, result.size());
+		assertTrue(result.contains(child));
+		assertTrue(result.contains(grandchild));
+		// the ancestor is the search root and is never part of its own descendants
+		assertFalse(result.contains(retiredRoot));
+		// the convenience method builds the same criteria, so it has to agree
+		assertEquals(result, ls.getDescendantLocations(retiredRoot, true));
+	}
+
+	/**
+	 * @see LocationService#getLocations(LocationSearchCriteria)
+	 */
+	@Test
+	public void getLocations_withDescendantOf_shouldExcludeAllDescendantsWhenAncestorIsRetiredAndIncludeRetiredIsFalse() {
+		LocationService ls = Context.getLocationService();
+		List<Location> tree = saveTreeUnderRetiredAncestor(ls);
+		Location retiredRoot = tree.get(0);
+		Location child = tree.get(1);
+		Location grandchild = tree.get(2);
+
+		// the descendants are not retired themselves, but their retired ancestor prunes the whole subtree
+		assertFalse(child.getRetired());
+		assertFalse(grandchild.getRetired());
+
+		LocationSearchCriteria criteria = new LocationSearchCriteria();
+		criteria.setDescendantOfLocation(retiredRoot);
+		criteria.setIncludeRetired(false);
+
+		assertTrue(ls.getLocations(criteria).isEmpty());
+		// the convenience method builds the same criteria, so it has to agree
+		assertTrue(ls.getDescendantLocations(retiredRoot, false).isEmpty());
+	}
+
+	/**
+	 * The deprecated {@link Location#getDescendantLocations(boolean)} names
+	 * {@link LocationService#getDescendantLocations(Location, boolean)} as its replacement, but it
+	 * starts from the child collection and so never examines the ancestor's own retired flag. The two
+	 * therefore disagree for a retired ancestor. That divergence is documented rather than changed -
+	 * changing it would silently alter results for everyone still on the deprecated method - so it is
+	 * pinned here instead.
+	 *
+	 * @see Location#getDescendantLocations(boolean)
+	 */
+	@Test
+	public void getDescendantLocations_shouldDisagreeWithTheDeprecatedEntityMethodForARetiredAncestor() {
+		LocationService ls = Context.getLocationService();
+		Location retiredRoot = saveTreeUnderRetiredAncestor(ls).get(0);
+
+		// the entity method walks the child collection, which is not loaded on the just-saved root
+		Context.flushSession();
+		Context.clearSession();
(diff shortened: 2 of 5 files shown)


api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
```diff
@@ -48,7 +48,8 @@ public enum TagMatchMode {
 
 	/**
 	 * @param descendantOfLocation the ancestor location; only its descendants are returned (root
-	 *            excluded)
+	 *            excluded). If the ancestor itself is retired and retired locations are excluded, none
+	 *            of its descendants are returned.
 	 */
```
jwnasambu, line 53: Thanks for the progress, @sudeshkar. This looks good. Though not a blocker, kindly update the getter javadoc to match the setter:

```

 /**
 * @return the ancestor location; only descendants of that location are returned (root
 *         excluded). If the ancestor itself is retired and retired locations are excluded, none
 *         of its descendants are returned.
 */
```
The getter and setter should document the same contract. Callers reading the getter's return documentation won't learn about the retired-ancestor pruning behavior, which could lead to incorrect assumptions about results.

api/src/test/java/org/openmrs/api/LocationServiceTest.java
```diff
@@ -1268,6 +1268,48 @@ public void getLocations_withDescendantOf_shouldExcludeRetiredSubtreesWhenInclud
...
+		ls.retireLocation(retiredRoot, "test");
+
+		LocationSearchCriteria withRetiredCriteria = new LocationSearchCriteria();
+		withRetiredCriteria.setDescendantOfLocation(retiredRoot);
+		withRetiredCriteria.setIncludeRetired(true);
+		List<Location> withRetired = ls.getLocations(withRetiredCriteria);
+		assertEquals(2, withRetired.size());
+		assertTrue(withRetired.contains(child));
+		assertTrue(withRetired.contains(grandchild));
+
```
jwnasambu:
```suggestion
assertFalse(withRetired.contains(retiredRoot));
```
The test verifies descendants are returned and checks the count is 2, but never explicitly asserts the retired ancestor is excluded. The CTE's WHERE location_id <> :locationId filter is the mechanism that excludes the root — this assertion pins that contract so a future refactor removing the filter would fail the test.

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
sudeshkar: Done, thanks — the getter now carries the same sentence as the setter:

```java
/**
 * @return the ancestor location; only descendants of that location are returned (root excluded). If
 *         the ancestor itself is retired and retired locations are excluded, none of its
 *         descendants are returned.
 */
```

I took the point further than the getter: `LocationService#getDescendantLocations(Location, boolean)` now documents it too, since callers of that convenience method get the criteria built for them and never read `LocationSearchCriteria` at all. Same sentence is now on all three branches.

api/src/test/java/org/openmrs/api/LocationServiceTest.java
sudeshkar: Added — good catch, the root-exclusion was indeed only implied by the count. It is now asserted explicitly:

```java
assertEquals(2, result.size());
assertTrue(result.contains(child));
assertTrue(result.contains(grandchild));
// the ancestor is the search root and is never part of its own descendants
assertFalse(result.contains(retiredRoot));
```

I also split the test in two, following the feedback on #6394 that one name was covering both `includeRetired` values. The two halves now share a `saveTreeUnderRetiredAncestor` helper.

**That split turned out to matter more than cosmetics — it exposed a real bug in the fix**, so please re-read the DAO change.

Running the `includeRetired = false` case on its own, with no query before it, it failed: the subtree came back instead of being pruned. The reason is that `getDescendantIds` uses `createNativeQuery`, which does not auto-flush by itself. `retireLocation` leaves a pending UPDATE in the session (the INSERTs are already out because `location_id` is IDENTITY-generated), so the CTE read `retired = false` for the ancestor and happily walked its subtree. In the old combined test the `includeRetired = true` half ran first, and the criteria query at the end of `getLocations` flushed the session — so the second half only passed by accident of ordering.

Fixed by declaring the query space, which is what makes `FlushMode.AUTO` apply to the native query:

```java
// a native query does not auto-flush on its own, so the location table is declared as a query
// space to make pending changes (a location retired earlier in the same transaction) visible
return session.createNativeQuery(cteSql, Integer.class).addSynchronizedEntityClass(Location.class)
        .setParameter("locationId", criteria.getDescendantOfLocation().getLocationId()).list();
```

Without it, any caller that retires a location and then queries its descendants in the same transaction reads stale data — so the contract we are documenting would not actually have held. Both tests are green now.

sudeshkar:
Follow-up from the review on #6393, applied here too so the three branches stay identical:

- The contract sentence is widened on the `LocationSearchCriteria` setter and getter and on `LocationService#getDescendantLocations`. The retired-ancestor case is not a rule of its own — with retired locations excluded, *any* retired location prunes its whole subtree, and a retired ancestor contributing nothing is that rule applied to the root. The old wording stated only the special case, which left a reader free to assume they otherwise get every non-retired descendant. The `@deprecated` caveat on `Location#getDescendantLocations(boolean)` still speaks about the ancestor specifically, because there it really is the only divergence — that method does skip retired children and their subtrees, it just never looks at the root.

- Both new tests now also assert through `LocationService#getDescendantLocations`. `git grep` turns up no reference to that method anywhere under `api/src/test` on any of the three branches, so the replacement the `@deprecated` note points migrating callers at had no test at all, and its delegation to the criteria path was unpinned. It sets only `descendantOfLocation` and `includeRetired` and calls `getLocations`, so comparing the lists directly pins ordering as well as membership.

`LocationServiceTest` green on all three.

sudeshkar:
Follow-up from the review on #6393, applied here too so the three branches stay identical:

- The `@deprecated` note on `Location#getDescendantLocations(boolean)` said the method "returns the non-retired ones", which overclaims. Reading the loop, it skips a retired child and does not recurse into it, so a non-retired descendant under a retired one is dropped as well — the ancestor is the *only* place the two methods disagree. It now says it "never examines the ancestor's own retired flag", which is both exact and exhaustive.

- New test `getDescendantLocations_shouldDisagreeWithTheDeprecatedEntityMethodForARetiredAncestor` pins that claim: the entity method returns the subtree of a retired ancestor (2), the service method returns nothing (0), same input. Nothing anywhere under `api/src/test` called the deprecated method before, which is notable given it is the divergence the javadoc now describes, and given that the absence of tests on it was the reason for documenting the divergence rather than changing it.

It reloads the root before calling the entity method, because that method walks the child collection, which is not loaded on the just-saved root. Populating it in the helper instead would make `retireLocation` cascade onto the children and break the `assertFalse(child.getRetired())` guard in the test above.

`LocationServiceTest` green: 86 on master, 84 on each backport.

PR #6393, opened 2026-07-26
TRUNK-6727: Exclude descendant locations when the ancestor itself is retired (2.9.x backport)

sudeshkar:

2.9.x backport of #6392.

On master, `HibernateLocationDAO` resolves the "descendant of" filter with a recursive CTE (TRUNK-6681) that was seeded with the *children* of the ancestor, so the retired flag of the ancestor itself was never considered and its descendants were still returned with `includeRetired = false`.

**This branch is not affected in production code.** It still resolves descendants in memory: `isDescendantOf` walks parent pointers upward and returns `false` as soon as it meets a retired node — the ancestor included, since the retired check happens before the ancestor comparison — so a retired ancestor already prunes its whole subtree here.

What this backports is therefore the regression tests and the documented contract, so the behaviour is pinned before the recursive-CTE lookup from TRUNK-6681 is backported to this branch.

The master fix also needed `addSynchronizedEntityClass(Location.class)` on the native CTE query, because a native query does not auto-flush and a pending `retireLocation` update was not visible to it. That does not apply here — the in-memory walk reads the session's own entities — so no production code changes on this branch.

### Tests

- `getLocations_withDescendantOf_shouldReturnDescendantsOfARetiredAncestorWhenIncludeRetiredIsTrue` — both descendants are returned, and the retired ancestor is explicitly asserted *not* to be among them.

Both pass against the current implementation, and the second fails against the master implementation before #6392. The test code is identical across all three branches.

### Documentation

## Issue I worked on

- [x] My pull request is **based on the latest changes** of the 2.9.x branch.

I ran the full `LocationServiceTest` class on this branch (83 tests, all green) rather than the whole build; this change adds tests and javadoc and touches no production code, so CI covers the rest.

### DIFF
--- a/api/src/main/java/org/openmrs/Location.java
+++ b/api/src/main/java/org/openmrs/Location.java
@@ -464,8 +464,8 @@ public class Location extends BaseCustomizableMetadata<LocationAttribute> implem
 	 * @since 1.10
 	 * @deprecated since 2.8.7 in favor of {@link LocationService#getDescendantLocations(Location, boolean)},
 	 *             which differs when the ancestor itself is retired: with {@code includeRetired} false
-	 *             it returns none of the descendants of a retired ancestor, whereas this method returns
-	 *             the non-retired ones.
+	 *             it returns none of the descendants of a retired ancestor, whereas this method never
+	 *             examines the ancestor's own retired flag.
 	 */
 	@Deprecated
 	public Set<Location> getDescendantLocations(boolean includeRetired) {
--- a/api/src/main/java/org/openmrs/Location.java
+++ b/api/src/main/java/org/openmrs/Location.java
@@ -462,7 +462,10 @@ public class Location extends BaseCustomizableMetadata<LocationAttribute> implem
 	 * @param includeRetired specifies whether or not to include voided childLocations
 	 * @return Returns a Set&lt;Location&gt; of the descendant location.
 	 * @since 1.10
-	 * @deprecated since 2.8.7 in favor of {@link LocationService#getDescendantLocations(Location, boolean)}
+	 * @deprecated since 2.8.7 in favor of {@link LocationService#getDescendantLocations(Location, boolean)},
+	 *             which differs when the ancestor itself is retired: with {@code includeRetired} false
+	 *             it returns none of the descendants of a retired ancestor, whereas this method returns
+	 *             the non-retired ones.
 	 */
 	@Deprecated
 	public Set<Location> getDescendantLocations(boolean includeRetired) {
--- a/api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
+++ b/api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
@@ -43,7 +43,9 @@ public class LocationSearchCriteria {
 	private boolean includeRetired = false;
 
 	/**
-	 * @param descendantOfLocation the ancestor location; only its descendants are returned (root excluded)
+	 * @param descendantOfLocation the ancestor location; only its descendants are returned (root excluded).
+	 *            When retired locations are excluded, a retired location prunes its whole subtree, so a
+	 *            retired ancestor contributes no descendants at all.
 	 */
 	public void setDescendantOfLocation(Location descendantOfLocation) {
 		this.descendantOfLocation = descendantOfLocation;
@@ -78,7 +80,9 @@ public class LocationSearchCriteria {
 	}
 
 	/**
-	 * @return the ancestor location; only descendants of that location are returned (root excluded)
+	 * @return the ancestor location; only descendants of that location are returned (root excluded).
+	 *         When retired locations are excluded, a retired location prunes its whole subtree, so a
+	 *         retired ancestor contributes no descendants at all.
 	 */
 	public Location getDescendantOfLocation() {
 		return descendantOfLocation;
--- a/api/src/test/java/org/openmrs/api/LocationServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/LocationServiceTest.java
@@ -1314,6 +1314,7 @@ public class LocationServiceTest extends BaseContextSensitiveTest {
 		criteria.setIncludeRetired(false);
 
 		assertTrue(ls.getLocations(criteria).isEmpty());
+		assertTrue(ls.getDescendantLocations(retiredRoot, false).isEmpty());
 		// the convenience method builds the same criteria, so it has to agree
 		assertTrue(ls.getDescendantLocations(retiredRoot, false).isEmpty());
 
--- a/api/src/test/java/org/openmrs/api/LocationServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/LocationServiceTest.java
@@ -1273,6 +1273,79 @@ public class LocationServiceTest extends BaseContextSensitiveTest {
 	 * @see LocationService#getLocations(LocationSearchCriteria)
 	 */
 	@Test
+	public void getLocations_withDescendantOf_shouldReturnDescendantsOfARetiredAncestorWhenIncludeRetiredIsTrue() {
+		LocationService ls = Context.getLocationService();
+		List<Location> tree = saveTreeUnderRetiredAncestor(ls);
+		Location retiredRoot = tree.get(0);
+		Location child = tree.get(1);
+		Location grandchild = tree.get(2);
+
+		LocationSearchCriteria criteria = new LocationSearchCriteria();
+		criteria.setDescendantOfLocation(retiredRoot);
+		criteria.setIncludeRetired(true);
+		List<Location> result = ls.getLocations(criteria);
+
+		assertEquals(2, result.size());
+		assertTrue(result.contains(child));
+		assertTrue(result.contains(grandchild));
+		// the ancestor is the search root and is never part of its own descendants
+		assertFalse(result.contains(retiredRoot));
+		// the convenience method builds the same criteria, so it has to agree
+		assertEquals(result, ls.getDescendantLocations(retiredRoot, true));
+	}
+
+	/**
+	 * @see LocationService#getLocations(LocationSearchCriteria)
+	 */
+	@Test
+	public void getLocations_withDescendantOf_shouldExcludeAllDescendantsWhenAncestorIsRetiredAndIncludeRetiredIsFalse() {
+		LocationService ls = Context.getLocationService();
+		List<Location> tree = saveTreeUnderRetiredAncestor(ls);
+		Location retiredRoot = tree.get(0);
+		Location child = tree.get(1);
+		Location grandchild = tree.get(2);
+
+		// the descendants are not retired themselves, but their retired ancestor prunes the whole subtree
+		assertFalse(child.getRetired());
+		assertFalse(grandchild.getRetired());
+
+		LocationSearchCriteria criteria = new LocationSearchCriteria();
+		criteria.setDescendantOfLocation(retiredRoot);
+		criteria.setIncludeRetired(false);
+
+		assertTrue(ls.getLocations(criteria).isEmpty());
+		// the convenience method builds the same criteria, so it has to agree
+		assertTrue(ls.getDescendantLocations(retiredRoot, false).isEmpty());
+	}
+
+	/**
+	 * Builds retiredRoot (retired) → child (active) → grandchild (active) and returns the three
+	 * locations in that order.
+	 */
+	private List<Location> saveTreeUnderRetiredAncestor(LocationService ls) {
+		Location retiredRoot = new Location();
+		retiredRoot.setName("Retired Ancestor");
+		ls.saveLocation(retiredRoot);
+
+		Location child = new Location();
+		child.setName("Child of Retired Ancestor");
+		child.setParentLocation(retiredRoot);
+		ls.saveLocation(child);
+
+		Location grandchild = new Location();
+		grandchild.setName("Grandchild of Retired Ancestor");
+		grandchild.setParentLocation(child);
+		ls.saveLocation(grandchild);
+
+		ls.retireLocation(retiredRoot, "test");
+
+		return Arrays.asList(retiredRoot, child, grandchild);
+	}
+
+	/**
+	 * @see LocationService#getLocations(LocationSearchCriteria)
+	 */
+	@Test
 	public void getLocations_withDescendantOf_shouldReturnEmptyListForLeafLocation() {
 		LocationService ls = Context.getLocationService();
 
--- a/api/src/test/java/org/openmrs/api/LocationServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/LocationServiceTest.java
@@ -1314,15 +1314,34 @@ public class LocationServiceTest extends BaseContextSensitiveTest {
 		criteria.setIncludeRetired(false);
 
 		assertTrue(ls.getLocations(criteria).isEmpty());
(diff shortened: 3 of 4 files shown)


jwnasambu:
@claude review

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
```diff
@@ -43,7 +43,9 @@ public enum TagMatchMode {
 	private boolean includeRetired = false;
 
 	/**
-	 * @param descendantOfLocation the ancestor location; only its descendants are returned (root excluded)
+	 * @param descendantOfLocation the ancestor location; only its descendants are returned (root excluded).
+	 *            If the ancestor itself is retired and retired locations are excluded, none of its
+	 *            descendants are returned.
```
dkayiwa: Worth repeating this sentence on `LocationService#getDescendantLocations(Location, boolean)` as well. Callers of that method get the criteria built for them by the service and never read `LocationSearchCriteria`, and its own javadoc still promises "its children, grandchildren, etc." with nothing about a retired ancestor returning none of them.

The reason I'd bother: the deprecated `Location#getDescendantLocations(boolean)` names that service method as its replacement, and the two disagree precisely on the case you're pinning. I built your `retiredRoot (retired)` -> `child (active)` -> `grandchild (active)` tree on this branch, reloaded the root so its child collection was populated, and called both with `includeRetired = false`. The deprecated entity method returned 2 (`[Grandchild, Child]`), the service method returned 0. So someone migrating off the deprecated method exactly the way its javadoc tells them to silently loses the active descendants.

None of that is introduced here and it doesn't need to hold the PR up. But since the documented contract is the deliverable, I'd put the sentence on the service method too, plus a short "differs when the ancestor is retired" caveat on the `@deprecated` line. Same shape on #6392 and #6394, so the three branches don't drift.

sudeshkar: Thanks for actually building the tree and running both methods — that is a much more convincing argument than the javadoc mismatch alone, and you are right that someone migrating off the deprecated method the way its javadoc instructs would silently lose the active descendants.

Done on all three branches (#6392, #6393, #6394), identical wording, so they do not drift.

`LocationService#getDescendantLocations(Location, boolean)`:

```java
 * @param location the location whose descendants should be returned. If this location itself is
 *            retired and {@code includeRetired} is false, none of its descendants are returned.
```

`Location#getDescendantLocations(boolean)`:

```java
 * @deprecated since 2.8.7 in favor of {@link LocationService#getDescendantLocations(Location, boolean)},
 *             which differs when the ancestor itself is retired: with {@code includeRetired} false
 *             it returns none of the descendants of a retired ancestor, whereas this method returns
 *             the non-retired ones.
```

I stated the divergence in the direction a migrating caller will hit it, rather than as a generic "differs", so the `@deprecated` line itself says what they will lose.

Agreed it is pre-existing and not introduced here, so I have documented it rather than changed either behaviour — reconciling them is a behaviour change for the deprecated method and belongs in its own ticket. Happy to raise one if you think it is worth doing before the method is removed.

dkayiwa: Worth widening this clause rather than leaving it on the ancestor alone. With retired locations excluded, *any* retired location prunes its whole subtree: `isDescendantOf` walks up from each candidate and bails on the first retired node it meets, so an active grandchild under a retired middle child is dropped too. `getLocations_withDescendantOf_shouldExcludeRetiredSubtreesWhenIncludeRetiredIsFalse`, a few lines above your new tests, already pins that (root -> childA active -> childB retired -> grandchild active returns just childA). Master's CTE keeps the same rule, since `retiredFilter` sits on the recursive step as well, so retired rows never enter `descendants` and their children have nothing to join to.

As written the sentence states only the special case, and a reader can fairly take the ancestor to be the one node whose retired flag prunes, and conclude they otherwise get every non-retired descendant. One clause covers both:

```suggestion
	 * @param descendantOfLocation the ancestor location; only its descendants are returned (root excluded).
	 *            When retired locations are excluded, a retired location prunes its whole subtree, so a
	 *            retired ancestor contributes no descendants at all.
```

The getter below and the `@param location` on `LocationService#getDescendantLocations` carry the same sentence and want the same edit, on all three branches. Nothing here blocks the backport.

api/src/test/java/org/openmrs/api/LocationServiceTest.java
```diff
@@ -1269,6 +1269,75 @@ public void getLocations_withDescendantOf_shouldExcludeRetiredSubtreesWhenInclud
...
+
+		// the descendants are not retired themselves, but their retired ancestor prunes the whole subtree
+		assertFalse(child.getRetired());
+		assertFalse(grandchild.getRetired());
+
+		LocationSearchCriteria criteria = new LocationSearchCriteria();
+		criteria.setDescendantOfLocation(retiredRoot);
+		criteria.setIncludeRetired(false);
+
+		assertTrue(ls.getLocations(criteria).isEmpty());
```
dkayiwa, line 1316: Now that the contract sentence sits on `LocationService#getDescendantLocations(Location, boolean)`, it would be worth asserting it through that method too. Nothing does today: grepping every `src/test` tree turns up zero references to `getDescendantLocations`, here or on master or 2.8.x, so the method the `@deprecated` note points migrating callers at has no test at all. The criteria-level tests cover the filtering; what is unpinned is the delegation, which is exactly what a caller reading that new `@param` sentence is trusting.

The fixture is already built here, so it costs a line:

```suggestion
		assertTrue(ls.getLocations(criteria).isEmpty());
		assertTrue(ls.getDescendantLocations(retiredRoot, false).isEmpty());
```

and `assertEquals(result, ls.getDescendantLocations(retiredRoot, true));` on the `includeRetired = true` test above. I applied both locally and ran the pair, green. I would not hold the backport for this.

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
dkayiwa: No, I would not raise that ticket. Making the deprecated method prune on a retired ancestor would silently change results for everyone who has not migrated yet, which is the same trap pointed in the other direction, and there is no test anywhere on `Location#getDescendantLocations(boolean)` to catch what such a change breaks (zero hits in every `src/test` tree, on all three branches). An untested behaviour change to an API that is on its way out costs more than the divergence does. Documenting it is the resolution: it is stated at the point where someone decides to migrate, and it disappears when the method does.

sudeshkar: You are right, and the way you put it is better than what I had: the ancestor is not a special rule, it is the special case of the general one. I checked both implementations before taking the wording, since the sentence is only worth widening if it is actually true on both branches.

On this branch `isDescendantOf` bails on the first retired node walking up, so an active grandchild under a retired middle child is dropped. On master, `retiredFilter` is concatenated onto the recursive step as well as the seed, so retired rows never enter `descendants` and their children have nothing left to join to. Same rule, and `getLocations_withDescendantOf_shouldExcludeRetiredSubtreesWhenIncludeRetiredIsFalse` already pins it, as you say.

Taken as suggested, on the setter, the getter and the `@param location` on `LocationService#getDescendantLocations`, all three branches:

```java
 * @param descendantOfLocation the ancestor location; only its descendants are returned (root excluded).
 *            When retired locations are excluded, a retired location prunes its whole subtree, so a
 *            retired ancestor contributes no descendants at all.
```

One thing I deliberately left alone: the `@deprecated` caveat on `Location#getDescendantLocations(boolean)` still speaks about the ancestor specifically, because there the ancestor really is the only point of divergence. That method does prune retired subtrees — it skips a retired child and does not recurse into it — so it agrees with the general rule everywhere except the root, which it never examines because it starts from `getChildLocations()`. Widening that sentence too would have made it say something false.

api/src/test/java/org/openmrs/api/LocationServiceTest.java
sudeshkar: Applied, both of them, on all three branches. I confirmed the gap you describe before adding them — `git grep getDescendantLocations` across `api/src/test` returns nothing on master, 2.9.x or 2.8.x, so the method the `@deprecated` note sends people to genuinely had no test at all. Documenting a contract on a method while leaving its delegation unpinned was the weaker half of what I did earlier, so this is a fair catch.

```java
assertEquals(result, ls.getDescendantLocations(retiredRoot, true));
...
assertTrue(ls.getDescendantLocations(retiredRoot, false).isEmpty());
```

I checked the `assertEquals` is sound rather than just green: `LocationServiceImpl.getDescendantLocations` sets exactly `descendantOfLocation` and `includeRetired` and calls `getLocations(criteria)`, so it is the same query and the same `order by name`, and comparing the lists directly pins ordering too rather than just membership.

Green on all three: 85 tests on master, 83 on each backport.

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
sudeshkar: Agreed, and the argument is stronger than the one I offered. I was thinking of the divergence as the defect and the ticket as the fix, but silently changing what the deprecated method returns would hit exactly the people who have not migrated — the ones least likely to be watching for it — and with zero tests on `Location#getDescendantLocations(boolean)` there is nothing to tell us what such a change breaks. That is a worse trade than the divergence.

No ticket then. Documenting it at the point of migration, where it disappears along with the method, is the resolution.

api/src/main/java/org/openmrs/Location.java
```diff
@@ -462,7 +462,10 @@ public Set<Location> getChildLocations(boolean includeRetired) {
 	 * @param includeRetired specifies whether or not to include voided childLocations
 	 * @return Returns a Set&lt;Location&gt; of the descendant location.
 	 * @since 1.10
-	 * @deprecated since 2.8.7 in favor of {@link LocationService#getDescendantLocations(Location, boolean)}
+	 * @deprecated since 2.8.7 in favor of {@link LocationService#getDescendantLocations(Location, boolean)},
+	 *             which differs when the ancestor itself is retired: with {@code includeRetired} false
+	 *             it returns none of the descendants of a retired ancestor, whereas this method returns
+	 *             the non-retired ones.
```
dkayiwa: `whereas this method returns the non-retired ones` claims a bit more than the method does. It skips a retired child and does not recurse into it, so a non-retired descendant sitting under a retired one is dropped as well. I built `retiredRoot (retired)` → `childA (active)` → `childB (retired)` → `grandchild (active)` on this branch and called it with `includeRetired = false`: it gave me `[childA]`, not `[childA, grandchild]`.

Your note in the thread above already says the method prunes retired subtrees, so it is really the wording that has not caught up. Naming what it skips rather than what it returns keeps it exact, and that is also the part a migrating caller acts on:

```suggestion
	 *             it returns none of the descendants of a retired ancestor, whereas this method never
	 *             examines the ancestor's own retired flag.
```

I applied that and ran the class, still green. Nothing here holds the backport up, but the same edit is wanted on #6392 and #6394 so the three do not drift.

api/src/test/java/org/openmrs/api/LocationServiceTest.java
```diff
@@ -1269,6 +1269,79 @@ public void getLocations_withDescendantOf_shouldExcludeRetiredSubtreesWhenInclud
...
+		// the convenience method builds the same criteria, so it has to agree
+		assertTrue(ls.getDescendantLocations(retiredRoot, false).isEmpty());
```
dkayiwa, line 1318: The `@deprecated` note now states what `Location#getDescendantLocations(boolean)` returns for a retired ancestor, and nothing in any `src/test` tree calls that method. Same gap you pointed at last round for the service method, and it is also the reason you gave for not reconciling the two: with no test on it, nothing tells us what a change would break. The fixture here is already built, so it costs a few lines.

The reload is not decorative. The entity method walks `getChildLocations()`, which on the just-saved root is still null, so the call NPEs without it. Populating that collection in the helper instead would make `retireLocation` cascade onto the children and break the `assertFalse(child.getRetired())` guard above.

```suggestion
		assertTrue(ls.getDescendantLocations(retiredRoot, false).isEmpty());

		// the deprecated entity method never looks at the ancestor's own retired flag, so it still
		// returns the subtree; reload because it walks the child collection, which is unloaded here
		Context.flushSession();
		Context.clearSession();
		assertEquals(2, ls.getLocation(retiredRoot.getLocationId()).getDescendantLocations(false).size());
```

83 green with it. Reasonable to skip, though: the CTE backport this PR is guarding against cannot touch the entity method, so unlike the criteria contract there is no incoming change here for a test to catch.

api/src/main/java/org/openmrs/Location.java
sudeshkar: You are right, and I checked the method body rather than take it on trust — the loop skips a retired child and does not recurse into it, so `[childA]` is what comes back, exactly as you measured. "returns the non-retired ones" claimed the method ignores retired state entirely, which it does not. My own reply in the thread above already said it prunes retired subtrees, so the sentence was contradicting what I had worked out one comment earlier. Taken as suggested:

```java
 *             it returns none of the descendants of a retired ancestor, whereas this method never
 *             examines the ancestor's own retired flag.
```

Naming what it skips is the better framing for the further reason that it is the whole of the divergence — everywhere else the two agree, so "never examines the ancestor's own retired flag" is not just exact, it is exhaustive.

Applied on all three branches.

api/src/test/java/org/openmrs/api/LocationServiceTest.java
sudeshkar: Added, and I did not skip it — your own argument for leaving the two methods alone was that no test tells us what changing the deprecated one would break, and the `@deprecated` note now makes a factual claim about what it returns. Leaving that claim unpinned is the same gap, one level down.

Two adjustments to how it landed, both worth flagging since you suggested it inline.

**The suggestion was applied here through the web UI and duplicated an assertion.** It was written against the revision before the convenience-method line went in, so committing it produced:

```java
assertTrue(ls.getLocations(criteria).isEmpty());
assertTrue(ls.getDescendantLocations(retiredRoot, false).isEmpty());
// the convenience method builds the same criteria, so it has to agree
assertTrue(ls.getDescendantLocations(retiredRoot, false).isEmpty());
```

Same call twice, with the comment describing it stranded between the two. Removed.

**The assertions are now their own test** rather than the tail of `getLocations_withDescendantOf_shouldExcludeAllDescendantsWhenAncestorIsRetiredAndIncludeRetiredIsFalse`. The disagreement between the two methods is a separate claim from what `getLocations` returns, and a test asserting the entity method under that name is the naming problem already raised on this class a round ago. It is now `getDescendantLocations_shouldDisagreeWithTheDeprecatedEntityMethodForARetiredAncestor`, asserting both halves together so the divergence is pinned in one place: 2 from the entity method, 0 from the service method, same input.

Your point about the reload is exactly right and is kept, with the reason in a comment — the child collection is unloaded on the just-saved root, and populating it in the helper instead would make `retireLocation` cascade and break the `assertFalse(child.getRetired())` guard.

I built this on top of the accepted commit rather than force-pushing over it, so that commit and its co-authorship stay in the history. 84 green here, 86 on master, 84 on 2.8.x, and the test is byte-identical on all three.

ibacher:
I think we should move discussion here to #6392. We can backport once the shape of the initial fix is agreed.

PR #6394, opened 2026-07-26
TRUNK-6727: Exclude descendant locations when the ancestor itself is retired (2.8.x backport)

sudeshkar:

2.8.x backport of #6392.

On master, `HibernateLocationDAO` resolves the "descendant of" filter with a recursive CTE (TRUNK-6681) that was seeded with the *children* of the ancestor, so the retired flag of the ancestor itself was never considered and its descendants were still returned with `includeRetired = false`.

**This branch is not affected in production code.** It still resolves descendants in memory: `isDescendantOf` walks parent pointers upward and returns `false` as soon as it meets a retired node — the ancestor included, since the retired check happens before the ancestor comparison — so a retired ancestor already prunes its whole subtree here.

What this backports is therefore the regression tests and the documented contract, so the behaviour is pinned before the recursive-CTE lookup from TRUNK-6681 is backported to this branch.

The master fix also needed `addSynchronizedEntityClass(Location.class)` on the native CTE query, because a native query does not auto-flush and a pending `retireLocation` update was not visible to it. That does not apply here — the in-memory walk reads the session's own entities — so no production code changes on this branch.

### Tests

- `getLocations_withDescendantOf_shouldReturnDescendantsOfARetiredAncestorWhenIncludeRetiredIsTrue` — both descendants are returned, and the retired ancestor is explicitly asserted *not* to be among them.

Both pass against the current implementation, and the second fails against the master implementation before #6392. The test code is identical across all three branches.

### Documentation

## Issue I worked on

I ran the full `LocationServiceTest` class on this branch (83 tests, all green) rather than the whole build; this change adds tests and javadoc and touches no production code, so CI covers the rest.

### DIFF
--- a/api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
+++ b/api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
@@ -43,7 +43,9 @@ public class LocationSearchCriteria {
 	private boolean includeRetired = false;
 
 	/**
-	 * @param descendantOfLocation the ancestor location; only its descendants are returned (root excluded)
+	 * @param descendantOfLocation the ancestor location; only its descendants are returned (root excluded).
+	 *            When retired locations are excluded, a retired location prunes its whole subtree, so a
+	 *            retired ancestor contributes no descendants at all.
 	 */
 	public void setDescendantOfLocation(Location descendantOfLocation) {
 		this.descendantOfLocation = descendantOfLocation;
@@ -78,7 +80,9 @@ public class LocationSearchCriteria {
 	}
 
 	/**
-	 * @return the ancestor location; only descendants of that location are returned (root excluded)
+	 * @return the ancestor location; only descendants of that location are returned (root excluded).
+	 *         When retired locations are excluded, a retired location prunes its whole subtree, so a
+	 *         retired ancestor contributes no descendants at all.
 	 */
 	public Location getDescendantOfLocation() {
 		return descendantOfLocation;
--- a/api/src/test/java/org/openmrs/api/LocationServiceTest.java
+++ b/api/src/test/java/org/openmrs/api/LocationServiceTest.java
@@ -1273,6 +1273,105 @@ public class LocationServiceTest extends BaseContextSensitiveTest {
 	 * @see LocationService#getLocations(LocationSearchCriteria)
 	 */
 	@Test
+	public void getLocations_withDescendantOf_shouldReturnDescendantsOfARetiredAncestorWhenIncludeRetiredIsTrue() {
+		LocationService ls = Context.getLocationService();
+		List<Location> tree = saveTreeUnderRetiredAncestor(ls);
+		Location retiredRoot = tree.get(0);
+		Location child = tree.get(1);
+		Location grandchild = tree.get(2);
+
+		LocationSearchCriteria criteria = new LocationSearchCriteria();
+		criteria.setDescendantOfLocation(retiredRoot);
+		criteria.setIncludeRetired(true);
+		List<Location> result = ls.getLocations(criteria);
+
+		assertEquals(2, result.size());
+		assertTrue(result.contains(child));
+		assertTrue(result.contains(grandchild));
+		// the ancestor is the search root and is never part of its own descendants
+		assertFalse(result.contains(retiredRoot));
+		// the convenience method builds the same criteria, so it has to agree
+		assertEquals(result, ls.getDescendantLocations(retiredRoot, true));
+	}
+
+	/**
+	 * @see LocationService#getLocations(LocationSearchCriteria)
+	 */
+	@Test
+	public void getLocations_withDescendantOf_shouldExcludeAllDescendantsWhenAncestorIsRetiredAndIncludeRetiredIsFalse() {
+		LocationService ls = Context.getLocationService();
+		List<Location> tree = saveTreeUnderRetiredAncestor(ls);
+		Location retiredRoot = tree.get(0);
+		Location child = tree.get(1);
+		Location grandchild = tree.get(2);
+
+		// the descendants are not retired themselves, but their retired ancestor prunes the whole subtree
+		assertFalse(child.getRetired());
+		assertFalse(grandchild.getRetired());
+
+		LocationSearchCriteria criteria = new LocationSearchCriteria();
+		criteria.setDescendantOfLocation(retiredRoot);
+		criteria.setIncludeRetired(false);
+
+		assertTrue(ls.getLocations(criteria).isEmpty());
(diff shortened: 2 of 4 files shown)


api/src/test/java/org/openmrs/api/LocationServiceTest.java
```diff
@@ -1269,6 +1269,48 @@ public void getLocations_withDescendantOf_shouldExcludeRetiredSubtreesWhenInclud
 		assertFalse(withoutRetired.contains(grandchild));
 	}
 
+	/**
+	 * @see LocationService#getLocations(LocationSearchCriteria)
+	 */
+	@Test
+	public void getLocations_withDescendantOf_shouldExcludeAllDescendantsWhenAncestorIsRetired() {
```
jwnasambu: 1. Thanks @sudeshkar for this fix.  Kindly do you fixing this please! `getLocations_withDescendantOf_shouldExcludeAllDescendantsWhenAncestorIsRetired` only describes the includeRetired=false half, but the test also asserts includeRetired=true behavior at lines 1296-1302. The name should reflect both cases `getLocations_withDescendantOf_shouldFilterDescendantsByRetiredAncestorAndIncludeRetiredFlag` or the test should be split into two.

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
jwnasambu, line 86: The setter on line 46 was updated with the retired-ancestor caveat, but the getter on this line still reads:
`@return the ancestor location; only descendants of that location are returned (root excluded)` 
Kindly change to:
```suggestion
/**
 * @return the ancestor location; only descendants of that location are returned (root excluded).
 *         If the ancestor itself is retired and retired locations are excluded, none of its
 *         descendants are returned.
 */
```

jwnasambu:
Kindly also update convenience method Javadoc on this file `LocationService.java:383-393`
Because `getDescendantLocations(Location, boolean) delegates` to the same getLocations criteria logic, but its Javadoc (@param includeRetired whether or not to include retired locations) doesn't mention that a retired ancestor prunes its entire subtree when includeRetired=false. Callers using this convenience method won't discover the interaction without reading the source.
Change the @param location line to:
 * @param location the location whose descendants should be returned.
 *            If this location itself is retired and {@code includeRetired} is false,
 *            no descendants are returned.

jwnasambu:
@claude review

api/src/test/java/org/openmrs/api/LocationServiceTest.java
sudeshkar: Agreed, and I went with your second option — split into two, since one name honestly covering both `includeRetired` values was always going to be a mouthful:

- `getLocations_withDescendantOf_shouldReturnDescendantsOfARetiredAncestorWhenIncludeRetiredIsTrue`
- `getLocations_withDescendantOf_shouldExcludeAllDescendantsWhenAncestorIsRetiredAndIncludeRetiredIsFalse`

They share a `saveTreeUnderRetiredAncestor(ls)` helper that builds `retiredRoot (retired) → child (active) → grandchild (active)` and returns the three locations, so no setup is duplicated. This also lines the names up with the existing pair just above them, which is already split by `includeRetired` value.

Worth flagging that the split was not just cosmetic: on master (#6392) it caught a real bug. Once the `includeRetired = false` case ran on its own, with no query before it, the subtree came back instead of being pruned — `createNativeQuery` does not auto-flush, so the pending `retireLocation` UPDATE had not reached the DB and the CTE still read the ancestor as active. The old combined test only passed because its `includeRetired = true` half flushed the session first. Fixed there with `addSynchronizedEntityClass(Location.class)`.

This branch is unaffected by that bug — descendants are still resolved in memory here — so the backport stays test-and-javadoc only.

api/src/main/java/org/openmrs/parameter/LocationSearchCriteria.java
sudeshkar: Applied, thanks — the getter now matches the setter on all three branches:

```java
/**
 * @return the ancestor location; only descendants of that location are returned (root excluded).
 *         If the ancestor itself is retired and retired locations are excluded, none of its
 *         descendants are returned.
 */
```

Your follow-up on `LocationService.java:383-393` is done as well, with the `@param location` wording you suggested — that was the same point dkayiwa raised on #6393, so it is on all three branches too, along with a note on the deprecated `Location#getDescendantLocations(boolean)` about where it and the service method disagree.

sudeshkar:
Follow-up from the review on #6393, applied here too so the three branches stay identical:

- The contract sentence is widened on the `LocationSearchCriteria` setter and getter and on `LocationService#getDescendantLocations`. The retired-ancestor case is not a rule of its own — with retired locations excluded, *any* retired location prunes its whole subtree, and a retired ancestor contributing nothing is that rule applied to the root. The old wording stated only the special case, which left a reader free to assume they otherwise get every non-retired descendant. The `@deprecated` caveat on `Location#getDescendantLocations(boolean)` still speaks about the ancestor specifically, because there it really is the only divergence — that method does skip retired children and their subtrees, it just never looks at the root.

- Both new tests now also assert through `LocationService#getDescendantLocations`. `git grep` turns up no reference to that method anywhere under `api/src/test` on any of the three branches, so the replacement the `@deprecated` note points migrating callers at had no test at all, and its delegation to the criteria path was unpinned. It sets only `descendantOfLocation` and `includeRetired` and calls `getLocations`, so comparing the lists directly pins ordering as well as membership.

`LocationServiceTest` green on all three.

sudeshkar:
Follow-up from the review on #6393, applied here too so the three branches stay identical:

- The `@deprecated` note on `Location#getDescendantLocations(boolean)` said the method "returns the non-retired ones", which overclaims. Reading the loop, it skips a retired child and does not recurse into it, so a non-retired descendant under a retired one is dropped as well — the ancestor is the *only* place the two methods disagree. It now says it "never examines the ancestor's own retired flag", which is both exact and exhaustive.

- New test `getDescendantLocations_shouldDisagreeWithTheDeprecatedEntityMethodForARetiredAncestor` pins that claim: the entity method returns the subtree of a retired ancestor (2), the service method returns nothing (0), same input. Nothing anywhere under `api/src/test` called the deprecated method before, which is notable given it is the divergence the javadoc now describes, and given that the absence of tests on it was the reason for documenting the divergence rather than changing it.

It reloads the root before calling the entity method, because that method walks the child collection, which is not loaded on the just-saved root. Populating it in the helper instead would make `retireLocation` cascade onto the children and break the `assertFalse(child.getRetired())` guard in the test above.

`LocationServiceTest` green: 86 on master, 84 on each backport.

ibacher:
I think we should move discussion here to #6392. We can backport once the shape of the initial fix is agreed.

PR #6399, opened 2026-07-28
Make the openmrs-api surefire crash on 2.8.x explain itself

dkayiwa:

Since 24 July the `openmrs-api` surefire fork has been aborting with `SIGABRT` (`Process Exit Code: 134`, `Aborted (core dumped)`) partway through the suite on 2.8.x, failing nearly every pull request against the branch (#6394, #6262, #6230, #6219 and the last two pushes to 2.8.x itself). Contributors have started pushing empty "re-run CI" commits to get around it.

Every test that ran passed, so the console log had nothing to point at:

```
[INFO] Running org.openmrs.liquibase.ChangeLogDetectiveTest
Aborted (core dumped)
[WARNING] Tests run: 1765, Failures: 0, Errors: 0, Skipped: 23
[WARNING] Corrupted channel by directly writing to native stream in forked JVM 1. See ... the dump file ....dumpstream
[ERROR] The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
```

That "corrupted channel" warning is the JVM writing its own explanation for the abort to stdout. Surefire diverts it into a `.dumpstream` file instead of the console, and the runner is torn down before anyone can read it, so the one piece of evidence that identifies the cause was being discarded on every failure.

This PR keeps that evidence: on a failed build it prints the surefire dump files and any `hs_err_pid*.log` into the job log, uploads them with `target/surefire-reports` as a per-job artifact, and reports free disk and memory. It also sets `fail-fast: false`, because the first crashing job was cancelling the other four, so a red build could not show whether a crash was specific to one Java version.

Nothing here changes what gets built or tested; it is all gated on `if: failure()`.

## What the first run found

It worked, and the crash is no longer a mystery. **These are two different JVM bugs, not a test failure.**

`build (ubuntu-latest, 11)`, Temurin 11.0.31+11:

```
#  Internal Error (moduleEntry.cpp:263), pid=2394, tid=2395
#  guarantee(java_lang_Module::is_instance(module)) failed: The unnamed module for ClassLoader
#  jdk.internal.reflect.DelegatingClassLoader, is null or not an instance of java.lang.Module.
#  The class loader has not been initialized correctly.
```

with the crash reached through reflection accessor generation from Mockito's inline mock maker:

```
V  [libjvm.so+0xc1828a]  ModuleEntry::create_unnamed_module(ClassLoaderData*)+0xca
V  [libjvm.so+0xf139db]  Unsafe_DefineClass0+0x37b
J  jdk.internal.reflect.MethodAccessorGenerator.generate(...)
J  jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(...)
J  org.mockito.internal.util.reflection.ReflectionMemberAccessor$$Lambda$1524.newInstance()
j  org.mockito.internal.creation.bytebuddy.InlineDelegateByteBuddyMockMaker.lambda$newInstance$4(...)
```

`build (ubuntu-latest, 8)`, Temurin 8.0.492-b09 — a C2 crash, on the compiler thread:

```
#  SIGSEGV (0xb) at pc=0x00007f07f4e0468d, pid=2414
#  Problematic frame:
#  V  [libjvm.so+0x40468d]  ciMemberName::get_vmtarget() const+0xad
Current thread: JavaThread "C2 CompilerThread0" daemon [_thread_in_vm]
```

And because the matrix no longer cancels itself, we can now see the shape of it:

| job | result |
| --- | --- |
| ubuntu-latest, Java 8 | **failure** — SIGSEGV in C2 |
| ubuntu-latest, Java 11 | **failure** — `moduleEntry.cpp` guarantee |
| ubuntu-latest, Java 17 | success |
| ubuntu-latest, Java 21 | success |
| windows-latest, Java 8 | success |

Two things worth recording, because both are tempting wrong turns:

- **It is not a JDK upgrade.** The last green run on 23 July used the *same* builds, Temurin 11.0.31-11 and 8.0.492-9. Nor is it the runner image bump: that run had its Java 8 job on the newer `ubuntu24/20260720.247` image and passed. The crash is intermittent, which is also why the failure rate looked like it started on a particular day.
- **It does not reproduce off the runners.** The full api suite passes locally on Java 11 with identical fork arguments (`-Xmx1g`, same JaCoCo agent), 5116 tests, and open files, threads and mapped regions stay flat, so it is not resource accumulation.

The leading lead for a fix is that the two failing jobs are exactly the ones on the old test toolchain. `mockitoVersion` is 3.12.4 with `byteBuddyVersion` 1.12.18 for Java below 17, against Mockito 5.6.0 and Byte Buddy 1.17.6 for 17 and up, and `api` forces `mock-maker-inline`, whose class generation is where the Java 11 crash lands. That correlation is not the whole story, since `windows-latest, 8` uses the old toolchain and passes, so I would rather test a toolchain bump against this now-reliable reproduction than assume it.

## Issue I worked on

No Jira issue yet; happy to file one if we would rather track it there.

- [x] I have **added tests** to cover my changes. CI-only change, so verified by extracting the step's script straight out of the workflow and running it under GitHub's exact shell (`bash --noprofile --norc -e -o pipefail`) against the real `hs_err` and `.dumpstream` files this PR captured, with a deliberately unreadable directory in the tree. It exits 0, still surfaces `find`'s own errors, and prints the Mockito frame that identifies the crash in 487 lines with no memory-map bloat; the first version of the script exits 1 and prints no fatal error log at all.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -18,6 +18,9 @@ permissions: read-all
 jobs:
   build:
     strategy:
+      # one crashing job used to cancel the rest, so a failure told us nothing about whether
+      # it was specific to a single Java version or hit the whole matrix
+      fail-fast: false
       matrix:
         platform:
           - ubuntu-latest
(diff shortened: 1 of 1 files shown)


jwnasambu:
@claude review

PR #6400, opened 2026-07-28
TRUNK-6734: Batch-fetch Patient.identifiers so patient search does not issue one query per result

jwnasambu:

I added batch-size="1000" to Patient.identifiers in Patient.hbm.xml, matching the existing configuration for Person.names, Person.addresses, and Person.attributes. This resolves the N+1 identifier query issue during patient search (previously causing 50 additional SELECT queries for a 50-patient page) without changing query results, ordering, or preferred-identifier logic. I also added PatientIdentifierBatchLoadTest with 7 tests, along with a supporting test dataset, to verify query counts, identifier-count edge cases, and voided/preferred-identifier behavior.

The second-level cache configuration originally included in the acceptance criteria was dropped from this PR (#6400) because second-level cache is not consistently enabled across all deployments. The batch-size change alone addresses the N+1 query issue.

## Issue I worked on

https://openmrs.atlassian.net/browse/TRUNK-6734

  No? -> execute above command

### DIFF
--- a/api/src/main/resources/org/openmrs/api/db/hibernate/Patient.hbm.xml
+++ b/api/src/main/resources/org/openmrs/api/db/hibernate/Patient.hbm.xml
@@ -50,7 +50,8 @@
 		<!-- bi-directional one-to-many association to PatientIdentifier -->
 		<!-- list is unsorted due so that its put into a LinkedHashSet which uses .hashcode() instead of .compareTo() -->
 		<set name="identifiers" lazy="true" cascade="all-delete-orphan"
-			table="patient_identifier" inverse="true" sort="natural">
+			table="patient_identifier" inverse="true" sort="natural" batch-size="1000">
+			<cache usage="read-write"/>
 			<key not-null="true" column="patient_id" />
 			<one-to-many class="PatientIdentifier" />
 		</set>
--- a/api/src/main/resources/org/openmrs/api/db/hibernate/Patient.hbm.xml
+++ b/api/src/main/resources/org/openmrs/api/db/hibernate/Patient.hbm.xml
@@ -51,7 +51,6 @@
 		<!-- list is unsorted due so that its put into a LinkedHashSet which uses .hashcode() instead of .compareTo() -->
 		<set name="identifiers" lazy="true" cascade="all-delete-orphan"
 			table="patient_identifier" inverse="true" sort="natural" batch-size="1000">
-			<cache usage="read-write"/>
 			<key not-null="true" column="patient_id" />
 			<one-to-many class="PatientIdentifier" />
 		</set>
--- a/api/src/test/java/org/openmrs/api/PatientIdentifierBatchLoadTest.java
+++ b/api/src/test/java/org/openmrs/api/PatientIdentifierBatchLoadTest.java
@@ -57,7 +57,7 @@ public class PatientIdentifierBatchLoadTest extends BaseContextSensitiveTest {
 		Statistics stats = sessionFactory.getStatistics();
 		stats.clear();
 
-		// Search for "Batch" patients (504, 505, 506, 507) - 4 patients
+		// Search for "Batch" patients (7004, 7005, 7006, 7007) - 4 patients
 		List<Patient> patients = patientService.getPatients("Batch", null, null, false);
 		assertEquals(4, patients.size(), "Should find all 4 Batch patients");
 
@@ -76,8 +76,8 @@ public class PatientIdentifierBatchLoadTest extends BaseContextSensitiveTest {
 		executeDataSet(IDENTIFIERS_DATASET);
 		updateSearchIndex();
 
-		Patient patient = patientService.getPatient(501);
-		assertNotNull(patient, "Patient 501 should exist");
+		Patient patient = patientService.getPatient(7001);
+		assertNotNull(patient, "Patient 7001 should exist");
 		assertEquals("NoIdent", patient.getGivenName());
 		assertNotNull(patient.getIdentifiers(), "Identifiers collection should not be null");
 		assertTrue(patient.getIdentifiers().isEmpty(), "Patient with no identifiers should have empty set");
@@ -91,8 +91,8 @@ public class PatientIdentifierBatchLoadTest extends BaseContextSensitiveTest {
 		executeDataSet(IDENTIFIERS_DATASET);
 		updateSearchIndex();
 
-		Patient patient = patientService.getPatient(502);
-		assertNotNull(patient, "Patient 502 should exist");
+		Patient patient = patientService.getPatient(7002);
+		assertNotNull(patient, "Patient 7002 should exist");
 		assertEquals("Single", patient.getGivenName());
 		assertEquals(1, patient.getIdentifiers().size(), "Patient should have exactly 1 identifier");
 
@@ -111,8 +111,8 @@ public class PatientIdentifierBatchLoadTest extends BaseContextSensitiveTest {
 		executeDataSet(IDENTIFIERS_DATASET);
 		updateSearchIndex();
 
-		Patient patient = patientService.getPatient(503);
-		assertNotNull(patient, "Patient 503 should exist");
+		Patient patient = patientService.getPatient(7003);
+		assertNotNull(patient, "Patient 7003 should exist");
 		assertEquals("Multi", patient.getGivenName());
 
 		// getIdentifiers() returns all (3 total: 2 non-voided + 1 voided)
@@ -144,28 +144,24 @@ public class PatientIdentifierBatchLoadTest extends BaseContextSensitiveTest {
 	}
 
 	/**
-	 * Verifies that identifier ordering follows natural ordering (PatientIdentifier's compareTo).
-	 * Uses getActiveIdentifiers() to get only non-voided identifiers in natural order.
-	 * Natural order: voided (false first), then preferred (true first), then dateCreated, then type, then identifier.
+	 * Verifies that the mapped collection comes back in PatientIdentifier's natural order:
+	 * non-voided first, then preferred ascending (Boolean.compareTo puts false before true),
+	 * then dateCreated, then type, then identifier.
 	 */
 	@Test
 	public void getPatient_shouldReturnIdentifiersInNaturalOrder() {
 		executeDataSet(IDENTIFIERS_DATASET);
 		updateSearchIndex();
 
-		Patient patient = patientService.getPatient(503);
+		Patient patient = patientService.getPatient(7003);
 		assertNotNull(patient);
 
-		// getActiveIdentifiers() returns non-voided identifiers in natural order
-		PatientIdentifier[] identifiers = patient.getActiveIdentifiers().toArray(new PatientIdentifier[0]);
-		assertEquals(2, identifiers.length);
+		PatientIdentifier[] identifiers = patient.getIdentifiers().toArray(new PatientIdentifier[0]);
+		assertEquals(3, identifiers.length);
 
-		// Natural order: preferred=true comes before preferred=false
-		// ID-503-B (preferred=true) comes before ID-503-A (preferred=false)
-		assertEquals("ID-503-B", identifiers[0].getIdentifier(),
-		    "First identifier should be ID-503-B (preferred=true comes first)");
-		assertEquals("ID-503-A", identifiers[1].getIdentifier(),
-		    "Second identifier should be ID-503-A (preferred=false comes second)");
+		assertEquals("ID-503-A", identifiers[0].getIdentifier());
+		assertEquals("ID-503-B", identifiers[1].getIdentifier());
+		assertEquals("ID-503-C-VOIDED", identifiers[2].getIdentifier());
 	}
 
 	/**
@@ -177,7 +173,7 @@ public class PatientIdentifierBatchLoadTest extends BaseContextSensitiveTest {
 		updateSearchIndex();
 
 		// Patient with multiple identifiers, preferred is ID-503-B
-		Patient patient = patientService.getPatient(503);
+		Patient patient = patientService.getPatient(7003);
 		assertNotNull(patient);
 		PatientIdentifier preferred = patient.getPatientIdentifier();
 		assertNotNull(preferred, "Preferred identifier should not be null");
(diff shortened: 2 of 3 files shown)


jwnasambu:
@claude review

jwnasambu:

api/src/main/resources/org/openmrs/api/db/hibernate/Patient.hbm.xml
```diff
@@ -50,7 +50,8 @@
 		<!-- bi-directional one-to-many association to PatientIdentifier -->
 		<!-- list is unsorted due so that its put into a LinkedHashSet which uses .hashcode() instead of .compareTo() -->
 		<set name="identifiers" lazy="true" cascade="all-delete-orphan"
-			table="patient_identifier" inverse="true" sort="natural">
+			table="patient_identifier" inverse="true" sort="natural" batch-size="1000">
+			<cache usage="read-write"/>
```
dkayiwa: This line needs to come out before merge. With it in place, adding an identifier through `PatientService.savePatientIdentifier()` and then reading `patient.getIdentifiers()` in a later session gives back the old list: the identifier is in the database but invisible to the API, and nothing throws. I ran the same probe three ways, against 2.8.x, against 2.8.x with only `batch-size` added, and against this branch. The first two go from 1 identifier to 2 after the save. This branch stays at 1 while `patient_identifier` holds 2 rows for that patient.

The collection is `inverse="true"` and `savePatientIdentifier` just does a `saveOrUpdate` on the child without touching `patient.getIdentifiers()`, so Hibernate queues no collection action and never invalidates the `org.openmrs.Patient.identifiers` region. The description is right that `Person.names`, `addresses` and `attributes` are mapped this way, but they have the same hole today (I reproduced it on 2.8.x through `savePersonName`), so I read that as a reason not to add a fourth cached collection rather than as a precedent. Fixing Person is its own ticket.

The cache also cancels out the batching this PR is for. Once the region is populated the collection is served from it, so `batch-size` never applies, and since `PatientIdentifier` has no second-level cache of its own (`PersonName`, `PersonAddress` and `PersonAttribute` are all `@Cache(READ_WRITE)`, this one isn't) every element then gets fetched by id. On 4 patients holding 7 identifiers I measured 7 single-row `select ... from patient_identifier where patient_identifier_id=?` and 12 JDBC statements on this branch, against 1 batched select and 6 statements with `batch-size` alone.

`batch-size` on its own gets the whole win the ticket is after, one batched select cold and warm, with none of the above:

```suggestion
			table="patient_identifier" inverse="true" sort="natural" batch-size="1000">
```

api/src/test/java/org/openmrs/api/PatientIdentifierBatchLoadTest.java
```diff
@@ -0,0 +1,233 @@
...
+		for (String query : queries) {
+			if (query.contains("patient_identifier")) {
+				identifierQueries += stats.getQueryStatistics(query).getExecutionCount();
+			}
+		}
+
+		// With batch-size="1000", all 4 patients' identifiers should load in 1 query
+		// (or 0 if cached). Without the fix, this would be 4 queries.
+		assertTrue(identifierQueries <= 1,
+		    "Expected at most 1 batched identifier query for 4 patients, but got " + identifierQueries);
```
dkayiwa: This assertion cannot fail, so nothing here guards the mapping change and that needs fixing before merge. `Statistics.getQueries()` holds the HQL and native query strings that went through the query API, not the SQL Hibernate emits when it initialises a lazy collection, so nothing in it ever mentions `patient_identifier` on this path. I printed the array on both branches for this exact `getPatients("Batch", ...)` call and got three entries, two `PersonName` HQL and one `GlobalProperty` HQL, with `identifierQueries` coming out 0 either way. On top of that `identifiers` is `lazy="true"` and nothing in this method touches `getIdentifiers()`, so no identifier loading happens inside the measured window at all.

I ran all seven tests in this class against 2.8.x without the `Patient.hbm.xml` change and they all pass, so if someone drops `batch-size="1000"` again nothing here goes red.

The collection statistics for the role do see it. Hydrating the page and asserting on the fetch count gives 4 on 2.8.x and 1 with the change, which I confirmed by running this replacement on both branches:

```suggestion
		// hydrate the identifiers for the whole page
		patients.forEach(p -> p.getIdentifiers().size());

		assertEquals(1, stats.getCollectionStatistics("org.openmrs.Patient.identifiers").getFetchCount(),
		    "a page of patients should load its identifiers in one batched select");
```

api/src/test/java/org/openmrs/api/PatientIdentifierBatchLoadTest.java
```diff
@@ -0,0 +1,233 @@
...
+
+		stats.setStatisticsEnabled(false);
```
dkayiwa: Worth dropping this line. It turns statistics off on the shared `SessionFactory` for the rest of the JVM run, and surefire runs the whole api module in one fork with one Spring context, so every test class scheduled after this one sees them off. `ContextTest` compares `getSecondLevelCacheHitCount()` before and after `evictEntity`, `evictAllEntities` and `clearEntireCache`; with statistics off that counter never moves and all three assertions hold no matter what the cache does. I checked both halves locally: a probe class running after this one reports `isStatisticsEnabled()` as false, and `ContextTest` still passes 26 of 26 in that order, so nothing goes red. `hibernate.generate_statistics=true` is already the default, so neither this call nor the `setStatisticsEnabled(true)` above is needed.

api/src/test/resources/org/openmrs/api/include/PatientServiceTest-identifiers.xml
```diff
@@ -0,0 +1,53 @@
...
+<dataset>
+  <!-- Patient with zero identifiers -->
+  <person person_id="501" gender="M" dead="false" creator="1" date_created="2026-01-01 00:00:00.0" voided="false" uuid="a0000000-0000-0000-0000-000000000501"/>
+  <person_name person_name_id="501" preferred="true" person_id="501" given_name="NoIdent" middle_name=" " family_name="Patient" creator="1" date_created="2026-01-01 00:00:00.0" voided="false" uuid="b0000000-0000-0000-0000-000000000501"/>
+  <patient patient_id="501" creator="1" date_created="2026-01-01 00:00:00.0" voided="false"/>
+
+  <!-- Patient with one identifier -->
+  <person person_id="502" gender="F" dead="false" creator="1" date_created="2026-01-01 00:00:00.0" voided="false" uuid="a0000000-0000-0000-0000-000000000502"/>
+  <person_name person_name_id="502" preferred="true" person_id="502" given_name="Single" middle_name=" " family_name="Ident" creator="1" date_created="2026-01-01 00:00:00.0" voided="false" uuid="b0000000-0000-0000-0000-000000000502"/>
+  <patient patient_id="502" creator="1" date_created="2026-01-01 00:00:00.0" voided="false"/>
```
dkayiwa: These ids need renumbering. Persons 501 and 502 already exist in `standardTestDataset.xml`, and `executeDataSet` runs `DatabaseOperation.REFRESH`, which updates a row whose primary key is already present instead of inserting a new one. So this block does not add two people, it rewrites two.

I printed both persons either side of loading this dataset:

```
BEFORE 501 gender=F uuid=df8ae447-6745-45be-b859-403241d9913c names=[Bruno Otterbourg]
AFTER  501 gender=M uuid=a0000000-0000-0000-0000-000000000501 names=[NoIdent Patient, Bruno Otterbourg]
BEFORE 502 gender=M uuid=341b4e41-790c-484f-b6ed-71dc8da222de names=[Hippocrates of Cos]
AFTER  502 gender=F uuid=a0000000-0000-0000-0000-000000000502 names=[Single Ident, Hippocrates of Cos]
```

Those two persons back the users `bruno` and `butch`, and the `<patient>` rows here turn both of them into patients, so after the load `getUser(501).getPersonName()` reads "NoIdent Patient". Nothing goes red today. `@Rollback` undoes all of it, and I ran this class ahead of PatientServiceTest, PersonServiceTest and UserServiceTest in one JVM with nothing failing across the 457. The cost lands later: `getPatient_shouldHandlePatientWithZeroIdentifiers` starts failing the day someone gives person 501 an identifier or a preferred name in the standard dataset, and the failure points at this file rather than at the change that caused it.

Only 501 and 502 actually collide, 503 through 507 are free, but I would move the whole block so the numbering stays uniform. 7001 through 7008 is unused everywhere under `api/src/test/resources`, so that plus the matching ids in `PatientIdentifierBatchLoadTest` covers it.

api/src/test/java/org/openmrs/api/PatientIdentifierBatchLoadTest.java
```diff
@@ -0,0 +1,222 @@
...
+		PatientIdentifier[] identifiers = patient.getActiveIdentifiers().toArray(new PatientIdentifier[0]);
+		assertEquals(2, identifiers.length);
+
+		// Natural order: preferred=true comes before preferred=false
+		// ID-503-B (preferred=true) comes before ID-503-A (preferred=false)
+		assertEquals("ID-503-B", identifiers[0].getIdentifier(),
+		    "First identifier should be ID-503-B (preferred=true comes first)");
+		assertEquals("ID-503-A", identifiers[1].getIdentifier(),
+		    "Second identifier should be ID-503-A (preferred=false comes second)");
+	}
```
dkayiwa, line 165: Natural order puts `preferred=false` first, not `preferred=true`, so line 149, the comments on 163 and 164, and both assertion messages have the rule backwards. `PatientIdentifier.DefaultComparator` compares `pi1.getPreferred().compareTo(pi2.getPreferred())`, and `Boolean.FALSE.compareTo(Boolean.TRUE)` is -1. Printing the mapped collection for patient 503 gives:

```
ID-503-A[voided=false,pref=false]  ID-503-B[voided=false,pref=true]  ID-503-C-VOIDED[voided=true,pref=false]
```

The assertions pass anyway because they read `getActiveIdentifiers()`, which builds its own list preferred-first in Java (`Patient.java:346-360`) and throws away whatever order the collection arrived in. So in spite of the name, this test never looks at the collection the mapping controls.

Reading `getIdentifiers()` instead puts the assertions on that collection and gets the documented rule the right way round. I ran the replacement and it passes on the branch. Being straight about its limits, it will not catch `sort="natural"` going missing, since these three rows come back in the same order either way, but it does at least describe what the mapping produces:

```suggestion
	/**
	 * Verifies that the mapped collection comes back in PatientIdentifier's natural order:
	 * non-voided first, then preferred ascending (Boolean.compareTo puts false before true),
	 * then dateCreated, then type, then identifier.
	 */
	@Test
	public void getPatient_shouldReturnIdentifiersInNaturalOrder() {
		executeDataSet(IDENTIFIERS_DATASET);
		updateSearchIndex();

		Patient patient = patientService.getPatient(503);
		assertNotNull(patient);

		PatientIdentifier[] identifiers = patient.getIdentifiers().toArray(new PatientIdentifier[0]);
		assertEquals(3, identifiers.length);
		assertEquals("ID-503-A", identifiers[0].getIdentifier());
		assertEquals("ID-503-B", identifiers[1].getIdentifier());
		assertEquals("ID-503-C-VOIDED", identifiers[2].getIdentifier());
	}
```

dkayiwa (commented):
One small thing before this merges: the description still says the change adds `<cache usage="read-write"/>` along with `batch-size`, but the cache element came out in the later commits, so the PR page now promises a second-level cache change the diff does not contain. TRUNK-6734 links here as the record of what was done, so could you update the description to match the final scope, batch-size only? Probably also worth a one-line note on the ticket itself, since its acceptance criteria still ask for the cache configuration and someone could otherwise pick that up later without the context for why it was dropped.

jwnasambu:
@dkayiwa Kindly feel free to review my PR please!

dkayiwa:
Could you also add that one-line note to TRUNK-6734? Its acceptance criteria still ask for the same second-level cache configuration as the person collections, and with this PR deliberately leaving the cache out, anyone picking the ticket up later sees an unmet criterion and could re-add the cache without the context sitting in these review threads.

jwnasambu:
@dkayiwa Kindly feel free to review the fix please!

api/src/main/resources/org/openmrs/api/db/hibernate/Patient.hbm.xml
ibacher: This seems like fixing things in the wrong direction. It seems like the right thing to do here is to also mark `savePatientIdentifier` invalidate the cache here. The goal is:

1. Avoid a DB lookup if we can.
2. If we can't, load in bulk.

Or am I missing something?

ibacher:
@jwnasambu Could you please fill in the checkboxes on the PR template?

api/src/main/resources/org/openmrs/api/db/hibernate/Patient.hbm.xml
ibacher: FWIW, I think Claude gets confused by the ticket _title_ and ignores the fact that this is in the acceptance criteria so that PatientIdentifiers mirror what we do with PersonName and PersonAddress objects, so this wasn't just an idle, unrelated suggestion.

dkayiwa: You're not missing anything about the end goal, 1 plus 2 together is the right target. My argument is that "have `savePatientIdentifier` invalidate the cache" is a smaller patch than the hole. Three separate things have to hold before the cached path is both correct and faster than what this PR now does:

1. Coverage. `savePatientIdentifier` isn't the only write that never touches the collection. `purgePatientIdentifier` lands in `HibernatePatientDAO.deletePatientIdentifier`, a bare `session.delete` of the child, and raw-JDBC writers (some module code and admin scripts) bypass Hibernate entirely. Each of those needs its own evict, including the ones that get written later. That's exactly how the Person collections ended up with the stale-read hole they carry today.

2. Race-safety. A manual `evictCollectionData()` in the DAO evicts when the statement runs, not when we commit. A concurrent request that read the old rows just before our commit can put the old list straight back into the region after our evict, and reads are stale again until the next write. Hibernate's own collection invalidation avoids that window by soft-locking the entry until after commit, but that machinery only fires when the write goes through `patient.getIdentifiers()`, which `inverse="true"` plus a child `saveOrUpdate` never does. So the robust form of "invalidate the cache" is "make identifier writes flow through the owning collection", which changes the contract of `savePatientIdentifier` (the Patient has to be attached, and with `cascade="all-delete-orphan"` membership edits become load-bearing) rather than adding a line to the DAO.

3. The entity cache. Even with perfect invalidation, a collection region alone makes the warm path slower, not faster: the region stores child ids, and `PatientIdentifier` has no entity cache (`PersonName`, `PersonAddress` and `PersonAttribute` all do), so a warm hit turns into one select per identifier. That's where the 7 single-row selects I measured came from. Real parity with Person is entity cache plus collection cache plus the invalidation fix, all three.

On the acceptance criteria, agreed, the cache is in there explicitly and I didn't read it as idle. But the configuration it asks to mirror carries this same stale-read bug on the Person side right now (`savePersonName` on plain 2.8.x reproduces it), and identifiers are what patients get looked up by, so mirroring copies the bug onto higher-stakes data. I'd keep this PR at batch-size, which is the whole measured win for the search N+1 in the ticket, and move the cache to its own ticket that does the three pieces above for Person and Patient together. That's also why I asked Juliet for the note on the ticket, so the dropped cache reads as deferred rather than forgotten.

ibacher (approved):

Issue #6334, opened 2026-07-15
Authorization-boundary hardening follow-ups (from security review)


dkayiwa:
While hardening the recent batch of authorization fixes, a few authorization-boundary observations came up that are broader than any single advisory and worth tracking as hardening work. None is a regression from the recent fixes; all are pre-existing. Filing as a tracking issue (no exploit detail); happy to split these or convert any item to a private advisory if preferred.

### 1. Service-layer `@Authorized` does not cover domain-object getters
`AuthorizationAdvice` enforces `@Authorized` only on service beans (AOP proxies). Navigating a loaded domain object reaches related data without the corresponding service privilege — e.g. `Encounter.getAllergies()` exposes an encounter's allergies to a caller holding only encounter privileges, without `Get Allergies`; the same pattern applies to other lazy relations. This is inherent to the current model. Worth deciding which navigations, if any, should be guarded and documenting the boundary.

### 2. Generic `getRefByUuid` path does not go through per-service `@Authorized`
`RefByUuid.getRefByUuid` / `DomainServiceImpl.fetch` (used by ~20 services) resolve an entity by uuid by delegating to that service's own `getXByUuid`, but via an internal `this.` call that does not re-enter the AOP proxy — so the `@Authorized` on those `getXByUuid` methods is not applied when an entity is reached through `getRefByUuid`. This is externally reachable through JSON deserialization: `UuidReferenceModule` resolves every `OpenmrsObject`-typed field via `DomainServiceImpl.fetchByUuid`, so deserializing a payload that references such an object by uuid resolves it without the per-service read privilege. Confirmed for `PatientProgram` / `PatientState` / `PatientProgramAttribute` while verifying #6318 (whose direct-invocation `@Authorized` is correct, but does not cover this path); the same applies to every `RefByUuid` implementation (Patient, Person, Order, Obs, Encounter, …). Consider enforcing authorization on the generic path (or restructuring so the delegated call goes through the proxy) so it matches the per-service read guards.

### 3. Minor: DB-update wizard progress endpoint
`UpdateFilter`'s progress AJAX endpoint returns update status/warnings/log path without an auth check during the DB-upgrade window, and the filter does not rotate the session id on successful authentication. Low severity (transient, pre-startup); noted for completeness.

_Context: surfaced during review of the recent authorization-annotation fixes._

ZahiqIbrahim:
Hi @dkayiwa , I'd like to pick up this issue, specifically items #2 and #3.

[...]

PR #6409, opened 2026-07-29
Harden UUID reference and updater authorization boundaries

RealBhupesh:
## Summary

This draft addresses the two concrete authorization-boundary follow-ups in #6334:

- route generic UUID reference resolution through each domain-specific service proxy, ensuring `@Authorized` advice is enforced
- require the updater session to authenticate before polling update progress
- rotate the updater session ID immediately after successful superuser authentication

Item 1 remains intentionally unchanged because it requires a broader decision about whether domain-object navigation or the service layer defines the authorization boundary.

## Root cause

`DomainServiceImpl` delegated services implementing `RefByUuid` to `getRefByUuid`. That method dispatched internally to `getXByUuid`, bypassing the Spring proxy and its authorization interceptor.

The updater's progress endpoint did not check the session authentication marker, and successful authentication retained the existing session ID.

## Impact

UUID-backed JSON references can no longer retrieve protected domain objects without the privilege required by their service getter. Update progress is limited to authenticated updater sessions, and successful updater authentication is protected against session fixation.

## Validation

- `./mvnw -pl api -Dtest=RefByUuidTest test`
- `./mvnw -pl web -am -Dtest=UpdateFilterE2ETest -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvnw clean package` (all 13 reactor modules passed from a no-space build path; the repository's module-resource tests mis-handle URL-encoded spaces in the original checkout path)

The new tests were also run before the fix and failed at each reported boundary.

## Coordination

@ZahiqIbrahim previously asked to take these same two items, but there is no assignment or open PR yet. This is a draft so maintainers can decide whether to review this implementation directly or coordinate ownership.

Related to #6334.

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/DomainServiceImpl.java
@@ -9,6 +9,7 @@
  */
 package org.openmrs.api.impl;
 
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -19,7 +20,6 @@ import java.util.Set;
 
 import org.openmrs.api.DomainService;
 import org.openmrs.api.OpenmrsService;
-import org.openmrs.api.RefByUuid;
 import org.openmrs.serialization.UuidReferenceModule;
 import org.springframework.aop.support.AopUtils;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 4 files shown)


jwnasambu (commented):
Hi @RealBhupesh ! Thanks for the contribution. Per our [PR Tips](https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477199/Pull+Request+Tips), please update the PR description to include a link to the related Jira issue. If the ticket doesn't exist yet on the issue tracker, please create one and attach it to the PR description. This helps us track and review your work more effectively. Thanks!

PR #6411, opened 2026-07-29
TRUNK-6723: Add ConceptInterpretationRule domain entity and persistence

IamMujuziMoses:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6723

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/Concept.java
+++ b/api/src/main/java/org/openmrs/Concept.java
@@ -141,6 +141,9 @@ public class Concept extends BaseOpenmrsObject implements Auditable, Retireable,
 
 	private Set<ConceptAttribute> attributes = new LinkedHashSet<>();
 
+	@AllowDirectAccess
+	private Set<ConceptInterpretationRule> interpretationRules = new LinkedHashSet<>();
+
 	/** default constructor */
 	public Concept() {
 		names = new HashSet<>();
@@ -1700,6 +1703,52 @@ public class Concept extends BaseOpenmrsObject implements Auditable, Retireable,
 	public void addAttribute(ConceptAttribute attribute) {
 		getAttributes().add(attribute);
 		attribute.setOwner(this);
+	}
+
+	/**
+	 * Gets the interpretation rules for this concept.
+	 *
+	 * @return the interpretation rules
+	 */
+	public Set<ConceptInterpretationRule> getInterpretationRules() {
+		if (interpretationRules == null) {
+			interpretationRules = new LinkedHashSet<>();
+		}
+		return interpretationRules;
+	}
+
+	/**
+	 * Sets the interpretation rules for this concept.
+	 *
+	 * @param interpretationRules the interpretation rules to set
+	 */
+	public void setInterpretationRules(Set<ConceptInterpretationRule> interpretationRules) {
+		this.interpretationRules = interpretationRules;
+	}
+
+	/**
+	 * Adds the given interpretation rule to this concept.
+	 *
+	 * @param interpretationRule the interpretation rule to add
+	 */
+	public void addInterpretationRule(ConceptInterpretationRule interpretationRule) {
+		if (interpretationRule != null && !getInterpretationRules().contains(interpretationRule)) {
+			interpretationRule.setConcept(this);
+			getInterpretationRules().add(interpretationRule);
+		}
+	}
+
+	/**
+	 * Removes the given interpretation rule from this concept.
+	 *
+	 * @param interpretationRule the interpretation rule to remove
+	 * @return true if the rule was removed, otherwise false
+	 */
+	public boolean removeInterpretationRule(ConceptInterpretationRule interpretationRule) {
+		if (interpretationRule != null) {
+			interpretationRule.setConcept(null);
+		}
+		return getInterpretationRules().remove(interpretationRule);
 	}
 
 }
--- a/api/src/main/java/org/openmrs/Concept.java
+++ b/api/src/main/java/org/openmrs/Concept.java
@@ -1709,6 +1709,7 @@ public class Concept extends BaseOpenmrsObject implements Auditable, Retireable,
 	 * Gets the interpretation rules for this concept.
 	 *
 	 * @return the interpretation rules
+	 * @since 3.0.0
 	 */
 	public Set<ConceptInterpretationRule> getInterpretationRules() {
 		if (interpretationRules == null) {
@@ -1721,6 +1722,7 @@ public class Concept extends BaseOpenmrsObject implements Auditable, Retireable,
 	 * Sets the interpretation rules for this concept.
 	 *
 	 * @param interpretationRules the interpretation rules to set
+	 * @since 3.0.0
 	 */
 	public void setInterpretationRules(Set<ConceptInterpretationRule> interpretationRules) {
 		this.interpretationRules = interpretationRules;
@@ -1730,6 +1732,7 @@ public class Concept extends BaseOpenmrsObject implements Auditable, Retireable,
 	 * Adds the given interpretation rule to this concept.
 	 *
 	 * @param interpretationRule the interpretation rule to add
+	 * @since 3.0.0
 	 */
 	public void addInterpretationRule(ConceptInterpretationRule interpretationRule) {
 		if (interpretationRule != null && !getInterpretationRules().contains(interpretationRule)) {
@@ -1743,6 +1746,7 @@ public class Concept extends BaseOpenmrsObject implements Auditable, Retireable,
 	 *
 	 * @param interpretationRule the interpretation rule to remove
 	 * @return true if the rule was removed, otherwise false
+	 * @since 3.0.0
 	 */
 	public boolean removeInterpretationRule(ConceptInterpretationRule interpretationRule) {
 		if (interpretationRule != null && getInterpretationRules().remove(interpretationRule)) {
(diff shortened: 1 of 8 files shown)


IamMujuziMoses:

dkayiwa:
@claude review

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -141,6 +141,51 @@
...
+				<constraints nullable="false" unique="true"/>
+			</column>
+			<column name="creator" type="INT">
+				<constraints nullable="false"/>
+			</column>
+			<column name="date_created" type="datetime">
+				<constraints nullable="false"/>
+			</column>
+			<column name="changed_by" type="INT"/>
+			<column name="date_changed" type="datetime"/>
```
jwnasambu: These audit columns can't be populated. ConceptInterpretationRule extends BaseOpenmrsObject and maps no creator, dateCreated, changedBy, or dateChanged properties, exactly like ConceptReferenceRange/concept_reference_range, whose table has none of these columns. The AuditableInterceptor only fills those fields when they are mapped persistent properties of the entity (it looks them up in the entity's property names), so nothing ever writes them here. On a Liquibase-created schema, the cascade insert from saveConcept omits creator and date_created and trips the NOT NULL constraints: saving any concept that has interpretation rules fails on production. The unit tests stay green only because the in-memory H2 schema is generated from the entity mappings (hbm2ddl create-drop) and simply has no such columns; I ran the three new ConceptServiceTest tests and they pass against that schema. Please drop the four audit columns from this changeSet to match concept_reference_range, or make the entity Auditable and map them. If you drop them, a DatabaseIT-style test that saves a rule against the Liquibase-created schema would be the one test that catches this class of mismatch.
Suggested fix (suggestion block):
			<column name="uuid" type="char(38)">
				<constraints nullable="false" unique="true"/>
			</column>
(Removes the creator/date_created/changed_by/date_changed columns.)

api/src/main/java/org/openmrs/ConceptInterpretationRule.java
```diff
@@ -0,0 +1,176 @@
...
+
+	private static final long serialVersionUID = 47329L;
+
+	@DocumentId
+	@Id
+	@Column(name = "concept_interpretation_rule_id")
+	@GeneratedValue(strategy = GenerationType.IDENTITY)
+	private Integer conceptInterpretationRuleId;
+
+	@Column(name = "criteria", columnDefinition = "TEXT")
```
jwnasambu: Nit: The sibling uses @Column(name = "criteria", length = 65535). The raw TEXT definition isn't portable across the databases OpenMRS supports and doesn't affect the Liquibase-owned schema either way.

api/src/main/java/org/openmrs/ConceptInterpretationRule.java
```diff
@@ -0,0 +1,176 @@
...
+	private String criteria;
+
+	@Enumerated(EnumType.STRING)
+	@JdbcTypeCode(SqlTypes.VARCHAR)
```
jwnasambu: Kindly @JdbcTypeCode is redundant here; @Enumerated(EnumType.STRING)` already maps to varchar.

api/src/main/resources/org/openmrs/api/db/hibernate/Concept.hbm.xml
```diff
@@ -115,6 +115,13 @@
             <one-to-many class="ConceptAttribute"/>
         </set>
 
+		<set name="interpretationRules" lazy="true" inverse="true" cascade="all,delete-orphan" batch-size="25" 
+			 order-by="priority asc">
```
jwnasambu: Two rules on the same concept can share a priority, and nothing prevents it: there's no unique constraint on concept_id, priority. The collection mapping sorts with order-by="priority asc" and no secondary key on this line, so tied rules come back in an arbitrary order and the in-memory LinkedHashSet insertion order is lost once the set is reloaded from the DB. If the evaluation is meant to pick the first matching rule by priority the result for tied rules becomes undefined. Worth adding a unique constraint on (concept_id, priority), or a secondary sort key like concept_interpretation_rule_id asc, so the order is deterministic for whichever consumer evaluates these rules.

IamMujuziMoses: Thanks @jwnasambu for catching these issues, I appreciate the detailed review it helped identify a few things I had overlooked. I’ll rework the implementation accordingly.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -141,6 +141,43 @@
 			<column name="description" type="varchar(255)"/>
 		</addColumn>
 	</changeSet>
+
+	<changeSet id="TRUNK-6723-2026-07-29" author="mujuzi">
+		<preConditions onFail="HALT">
```
jwnasambu: Kindly this deviates from this file's own convention all other changeSets there lines 79, 95, 107, 134 and the sibling concept_reference_range changeSet use onFail="MARK_RAN". If the table already exists in a target DB, HALT would stop the entire upgrade; MARK_RAN would skip it cleanly. Recommend switching for consistency and upgrade-safety.

api/src/main/java/org/openmrs/Concept.java
```diff
@@ -1702,4 +1705,50 @@ public void addAttribute(ConceptAttribute attribute) {
...
+	 * @param interpretationRule the interpretation rule to remove
+	 * @return true if the rule was removed, otherwise false
+	 */
+	public boolean removeInterpretationRule(ConceptInterpretationRule interpretationRule) {
+		if (interpretationRule != null) {
+			interpretationRule.setConcept(null);
+		}
+		return getInterpretationRules().remove(interpretationRule);
+	}
+
```
jwnasambu, line 1754:
```suggestion
public boolean removeInterpretationRule(ConceptInterpretationRule interpretationRule) {
    if (interpretationRule != null && getInterpretationRules().remove(interpretationRule)) {
        interpretationRule.setConcept(null);
        return true;
    }
    return false;
}

```
line 1749 fires regardless of whether the rule is actually in this concept's set. A safer version nulls only on successful removal

IamMujuziMoses: thanx @jwnasambu for the review, good catch! You’re right, the current implementation clears the back-reference even when the rule is not part of the concept’s collection, which can leave the relationship in an inconsistent state.

I’ll update it to only clear the concept reference after a successful removal.

dkayiwa:
@claude review

api/src/main/java/org/openmrs/Concept.java
```diff
@@ -1702,4 +1705,51 @@ public void addAttribute(ConceptAttribute attribute) {
 		attribute.setOwner(this);
 	}
 
+	/**
+	 * Gets the interpretation rules for this concept.
+	 *
+	 * @return the interpretation rules
```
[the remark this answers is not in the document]
IamMujuziMoses: Thanks for pointing that out. I’ll add @since 3.0.0 to all four newly introduced public methods so the API documentation accurately reflects when they were added.

api/src/test/java/org/openmrs/api/ConceptServiceTest.java
```diff
@@ -4037,6 +4039,118 @@ public void saveConcept_shouldSaveANewConceptReferenceRange() {
...
+		Concept loaded = conceptService.getConcept(concept.getId());
+
+		// Verify that both rules were persisted.
+		assertEquals(2, loaded.getInterpretationRules().size());
+
+		List<ConceptInterpretationRule> rules = new ArrayList<>(loaded.getInterpretationRules());
+
+		// Verify the rules are ordered by priority ascending.
+		assertEquals(Integer.valueOf(1), rules.get(0).getPriority());
+		assertEquals(Integer.valueOf(2), rules.get(1).getPriority());
```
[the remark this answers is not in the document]
IamMujuziMoses: You’re right. The original test could pass based on insertion/ID order and therefore didn’t actually verify the mapping’s priority ordering. I’ll reverse the priorities in the setup so the test explicitly verifies that the collection is loaded in ascending priority order.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -141,6 +141,43 @@
...
+		</preConditions>
+		<comment>Creates the 'concept_interpretation_rule' table for coded concept interpretation rules.</comment>
+		<createTable tableName="concept_interpretation_rule">
+			<column name="concept_interpretation_rule_id" type="int" autoIncrement="true">
+				<constraints primaryKey="true" nullable="false"/>
+			</column>
+			<column name="concept_id" type="int">
+				<constraints nullable="false"/>
+			</column>
+			<column name="criteria" type="text"/>
```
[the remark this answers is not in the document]
IamMujuziMoses: Good point. Since the shared criteria evaluation rejects null or blank criteria, allowing NULL here would only allow invalid rules to be persisted. I’ll make criteria NOT NULL in both the entity and Liquibase changeset.

api/src/main/java/org/openmrs/Concept.java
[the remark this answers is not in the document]
ibacher: This should actually be `@since 2.8.10` I think.

PR #6417, opened 2026-07-30
TRUNK-6685: Patient identifier search no longer includes the identifierType field

IamMujuziMoses:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6685

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
@@ -845,10 +845,18 @@ public class HibernatePatientDAO implements PatientDAO {
 	        boolean matchExactly) {
 		List<String> tokens = tokenizeIdentifierQuery(removeIdentifierPadding(paramQuery));
 		final String query = StringUtils.join(tokens, " | ");
+
+		final List<Integer> typeIdsMatchingQuery = matchExactly ? Collections.emptyList()
+		        : getIdentifierTypeIdsByName(paramQuery);
+
 		return f.bool().with(b -> {
 			b.minimumShouldMatchNumber(1);
 			b.should(f.simpleQueryString().field("identifierPhrase").matching(query).boost(8f));
-			b.should(f.simpleQueryString().field("identifierType.name").matching(query));
+
+			if (!typeIdsMatchingQuery.isEmpty()) {
+				b.should(f.terms().field("identifierType.patientIdentifierTypeId").matchingAny(typeIdsMatchingQuery));
+			}
+
 			String matchMode = Context.getAdministrationService()
 			        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_PATIENT_IDENTIFIER_SEARCH_MATCH_MODE);
 			if (matchExactly) {
@@ -860,6 +868,14 @@ public class HibernatePatientDAO implements PatientDAO {
 			}
 		}).toPredicate();
 
+	}
+
+	private List<Integer> getIdentifierTypeIdsByName(String name) {
+		return sessionFactory.getCurrentSession()
+		        .createQuery("select pit.patientIdentifierTypeId " + "from PatientIdentifierType pit "
+		                + "where lower(pit.name) = lower(:name)",
+		            Integer.class)
+		        .setParameter("name", name.trim()).getResultList();
 	}
 
 	private SearchPredicate newPatientIdentifierSearchPredicate(SearchPredicateFactory predicateFactory, String query,
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
@@ -845,11 +845,10 @@ public class HibernatePatientDAO implements PatientDAO {
 	        boolean matchExactly) {
 		List<String> tokens = tokenizeIdentifierQuery(removeIdentifierPadding(paramQuery));
 		final String query = StringUtils.join(tokens, " | ");
-		//TODO: hibernate search identifierType?
-		//fields.add("identifierType");
 		return f.bool().with(b -> {
 			b.minimumShouldMatchNumber(1);
 			b.should(f.simpleQueryString().field("identifierPhrase").matching(query).boost(8f));
+			b.should(f.simpleQueryString().field("identifierType.name").matching(query));
 			String matchMode = Context.getAdministrationService()
 			        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_PATIENT_IDENTIFIER_SEARCH_MATCH_MODE);
 			if (matchExactly) {
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
@@ -867,7 +867,6 @@ public class HibernatePatientDAO implements PatientDAO {
 				b.should(f.simpleQueryString().field("identifierAnywhere").matching(query));
 			}
 		}).toPredicate();
-
 	}
 
 	private List<Integer> getIdentifierTypeIdsByName(String name) {
--- a/api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
+++ b/api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
@@ -32,6 +32,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.hasItems;
+import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
@@ -317,6 +318,8 @@ public class HibernatePatientDAOTest extends BaseContextSensitiveTest {
 
 		List<Integer> ids = patients.stream().map(Patient::getPatientId).collect(Collectors.toList());
 
+		// Patient 432 is voided. This overload currently doesn't filter voided patients, so we assert the current
+		// behavior. Fixing that should be handled separately
 		assertThat(ids, hasItems(2, 6, 432));
 		// Patients with a different identifier type should not be returned.
 		assertThat(ids, not(hasItem(7)));
@@ -348,5 +351,58 @@ public class HibernatePatientDAOTest extends BaseContextSensitiveTest {
 
 		assertThat(ids, hasItems(2, 6));
 		assertThat(ids, not(hasItem(432))); // voided patient excluded
+
+		// A type name pasted from the admin UI screen may include surrounding whitespace.
+		List<Patient> padded = hibernatePatientDao.findPatients("  Old Identification Number  ", false, null, null);
+
+		assertThat(padded.stream().map(Patient::getPatientId).collect(Collectors.toList()), hasItems(2, 6));
+	}
+
+	/**
+	 * @see HibernatePatientDAO#findPatients(String, boolean, Integer, Integer)
+	 */
+	@Test
+	public void findPatients_shouldFindPatientsByRetiredIdentifierTypeName() {
+		PatientIdentifierType retiredType = Context.getPatientService().getPatientIdentifierType(4);
+		Patient patient = Context.getPatientService().getPatient(7);
+
+		// Associate the patient with a retired identifier type.
+		PatientIdentifier identifier = new PatientIdentifier();
+		identifier.setPatient(patient);
+		identifier.setIdentifier("RETIRED-TYPE-001");
+		identifier.setIdentifierType(retiredType);
+		identifier.setLocation(Context.getLocationService().getLocation(1));
+		identifier.setPreferred(false);
+
+		Context.getPatientService().savePatientIdentifier(identifier);
+		updateSearchIndex();
+
+		// Searching by the retired identifier type name should still find the patient.
+		List<Patient> patients = hibernatePatientDao.findPatients(retiredType.getName(), false, null, null);
+
+		assertThat(patients.stream().map(Patient::getPatientId).collect(Collectors.toList()), hasItem(7));
+	}
+
+	/**
+	 * @see HibernatePatientDAO#getPatients(String, List, boolean, Integer, Integer)
+	 */
+	@Test
+	public void getPatients_shouldNotMatchIdentifierTypeNameWhenMatchingIdentifierExactly() {
+		String identifierTypeName = "Old Identification Number";
+		PatientIdentifierType type = Context.getPatientService().getPatientIdentifierTypeByName(identifierTypeName);
+
+		List<Patient> patients = hibernatePatientDao.getPatients(identifierTypeName, singletonList(type), true, null, null);
+
+		assertThat(patients, is(emptyList()));
+	}
+
+	/**
+	 * @see HibernatePatientDAO#findPatients(String, boolean, Integer, Integer)
+	 */
+	@Test
+	public void findPatients_shouldNotMatchPartialIdentifierTypeName() {
+		List<Patient> patients = hibernatePatientDao.findPatients("Number", false, null, null);
+
+		assertThat(patients, is(emptyList()));
 	}
 }
--- a/api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
+++ b/api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
@@ -30,7 +30,9 @@ import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.hasItems;
+import static org.hamcrest.Matchers.not;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class HibernatePatientDAOTest extends BaseContextSensitiveTest {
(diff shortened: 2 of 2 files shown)


IamMujuziMoses:

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -845,11 +845,10 @@ private SearchPredicate getPatientIdentifierSearchPredicate(SearchPredicateFacto
 	        boolean matchExactly) {
 		List<String> tokens = tokenizeIdentifierQuery(removeIdentifierPadding(paramQuery));
 		final String query = StringUtils.join(tokens, " | ");
-		//TODO: hibernate search identifierType?
-		//fields.add("identifierType");
 		return f.bool().with(b -> {
 			b.minimumShouldMatchNumber(1);
 			b.should(f.simpleQueryString().field("identifierPhrase").matching(query).boost(8f));
+			b.should(f.simpleQueryString().field("identifierType.name").matching(query));
```
dkayiwa: `simpleQueryString` matches token by token, so this matches on any single word of any identifier type name rather than on the type name as a whole, and I think that has to be narrowed before this goes in. Against the standard test data on this branch, `findPatients("Number")` returns patients 2, 6, 7, 8, `findPatients("Identification")` returns the same four, and `findPatients("OpenMRS")` returns 2, 7, 8. All three return nothing on master. Every word of every type name becomes a wildcard for the patients holding that type, and "number", "identification", "openmrs", "old", "national", "security" are all in the stock metadata.

What turns that from noise into a real failure is the ordering. `SearchQueryUnique` runs the identifier sub-query first and returns as soon as the requested page is full, so the name sub-query never gets reached. I added a patient whose family name is literally "Number" and asked for a page of two: `getPatients("Number", 0, 2)` came back with the two patients that matched only on type name, and the patient actually called "Number" was not in the page at all (at length 50 it showed up last). legacyui's search box goes through exactly that call, `DWRPatientService.findBatchOfPatients` invokes `getPatients(searchValue, includeVoided, start, length)` with a page length. So on any real database, where more than `length` patients hold a matching type and every patient holds an "OpenMRS Identification Number", a name search that happens to share one word with a type name returns a page of arbitrary patients instead of the patient the user typed, and since nothing throws it just reads as "that patient isn't in the system".

Tightening the operator does not rescue it. I tried `.defaultOperator(BooleanOperator.AND)` and a one-word query still matches one token of a multi-word type name; all it does is stop "Old Identification Number" from also matching "OpenMRS Identification Number". What does work is to resolve the name against `patient_identifier_type` and feed the ids into the filter this file already uses for identifier types:

```java
final List<Integer> typeIdsMatchingQuery = matchExactly
        ? Collections.emptyList() : getIdentifierTypeIdsByName(paramQuery);
...
if (!typeIdsMatchingQuery.isEmpty()) {
    b.should(f.terms().field("identifierType.patientIdentifierTypeId")
            .matchingAny(typeIdsMatchingQuery));
}
```

with `getIdentifierTypeIdsByName` doing a case-insensitive equality lookup on `name`. I ran your three new tests plus these probes with that in place: the tests still pass, `findPatients("Number")` is empty again, and `getPatients("Number", 0, 2)` returns only the patient named "Number". It costs one extra query per search against a table with a handful of rows, and it also happens to clear up the two points I left below.

dkayiwa: `identifierType.name` is a mutable value being read out of the search index, and `PatientIdentifier.identifierType` carries `@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)`, which is precisely the setting that tells Hibernate Search not to reindex identifiers when the type they point at is modified. The consequence is that once an admin renames an identifier type, patient search finds nothing under the new name and keeps matching the old one, silently, until somebody rebuilds the whole index.

I checked this in a non-transactional test so the saves really commit and automatic indexing runs. After `savePatientIdentifierType` renamed type 2 to "Zzqq Renamed Numbering", `findPatients("Zzqq Renamed Numbering")` returned nothing while `findPatients("Old Identification Number")` still returned 2, 6, 7, 8. As a control in the same run, editing a `PatientIdentifier`'s own value was picked up straight away ("99999Z" found, "12345K" no longer found), so this is the embedded type name going stale rather than the test harness.

Nothing read anything mutable out of that embedded object before now, only `identifierType.patientIdentifierTypeId`, and an id never changes, which is why SHALLOW cost nothing. Dropping the annotation is not the way out either: bootstrap then fails with HSEARCH700020 because there is no inverse side from `PatientIdentifierType` back to `PatientIdentifier`, I tried that. Resolving names against the database as in the comment above sidesteps this completely, since ids are stable. If you would rather keep the name in the index, then `PatientServiceImpl.savePatientIdentifierType` needs to reindex when the name changed, the way `PersonServiceImpl.savePersonAttributeType` calls `Context.updateSearchIndexForType(PersonAttribute.class)` when `searchable` changes.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -845,11 +845,10 @@ private SearchPredicate getPatientIdentifierSearchPredicate(SearchPredicateFacto
...
 			String matchMode = Context.getAdministrationService()
 			        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_PATIENT_IDENTIFIER_SEARCH_MATCH_MODE);
 			if (matchExactly) {
```
dkayiwa, line 862: The new clause sits above this branch, so it also applies when `matchExactly` is true, and `PatientService.getPatients(name, identifier, identifierTypes, matchIdentifierExactly)` documents that flag as "if true, then the given identifier must equal the id in the database". On this branch `getPatients("Old Identification Number", singletonList(type2), true, null, null)` gives me patients 2, 6 and 432, whose identifiers are 101 and 101-6, 12345K, and QWE and RTY. None of them equals the query. Skipping the type-name clause when `matchExactly` is set restores that.

I could not find a caller in core, webservices.rest or legacyui that reaches this overload with a non-null type list, and `PatientServiceImpl` still carries a TODO saying it is unused, so I would not hold the PR on this one alone. It is a documented contract that modules can rely on though, and the reason it went unnoticed is that nothing exercises `matchIdentifierExactly = true`; a test for it would be worth adding next to the new ones.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -845,11 +845,10 @@ private SearchPredicate getPatientIdentifierSearchPredicate(SearchPredicateFacto
 	        boolean matchExactly) {
 		List<String> tokens = tokenizeIdentifierQuery(removeIdentifierPadding(paramQuery));
 		final String query = StringUtils.join(tokens, " | ");
-		//TODO: hibernate search identifierType?
-		//fields.add("identifierType");
```
dkayiwa, line 849: Worth double-checking the premise in the ticket, because I do not think this matched in 2.6.x either. That code handed `identifierType` to a `MultiFieldQueryParser`, but Hibernate Search 5 prefixed `@IndexedEmbedded` fields with the property path, so the index carried `identifierType.name` and `identifierType.patientIdentifierTypeId` (the same method queried the latter by its full path) and never a bare `identifierType`, and an unknown field in `MultiFieldQueryParser` matches nothing rather than erroring. I also cannot find a 2.6.x test asserting a patient search by type name. If that holds, this is new behavior rather than a restoration, which matters mainly because it means the matching semantics are ours to pick rather than something we have to reproduce. Does that match what you found?

IamMujuziMoses: I looked back at the 2.6.x implementation, and I think you’re right. Although it queried identifierType, I couldn’t find evidence that Hibernate Search 5 indexed a field by that name, so, searching by identifier type likely never worked in practice, so my change is introducing new behavior rather than a restoration of old behavior.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
IamMujuziMoses: This actually makes sense. Thanks for review and explaining the ordering issue. I hadn’t considered how the identifier sub-query could fill the page before the name sub-query gets a chance to run. Resolving the query to matching identifier type IDs first and then filtering on `identifierType.patientIdentifierTypeId` seems like a much more robust approach. I’ll rework the implementation accordingly.

jwnasambu:
@claude review

api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
```diff
@@ -299,4 +301,52 @@ public void getDuplicatePatientsByAttributes_shouldReturnVoidedPatientsWithDupli
...
+		String identifierTypeName = "Old Identification Number";
+		PatientIdentifierType type = Context.getPatientService().getPatientIdentifierTypeByName(identifierTypeName);
+		List<Patient> patients = hibernatePatientDao.getPatients(identifierTypeName, singletonList(type), false, null, null);
+
+		// Patients with identifiers of the requested type should be returned.
+		assertEquals(3, patients.size());
+
+		List<Integer> ids = patients.stream().map(Patient::getPatientId).collect(Collectors.toList());
+
+		assertThat(ids, hasItems(2, 6, 432));
```
dkayiwa, line 323: Is including 432 here deliberate? It is a voided patient in the standard dataset (`voided="true"` on both its person and patient rows), and the service methods that reach this overload say the opposite: `PatientService.getPatients(name, identifier, identifierTypes, matchIdentifierExactly, start, length)` documents "@return patients that matched the given criteria (and are not voided)", and the 4-arg version carries "Should not return voided patients". The third test below is explicit that 432 is excluded, so as written the two read as contradicting each other.

The leak itself is not yours. `findPatients(query, identifierTypes, matchExactly, start, length)` filters the identifier's own `voided` flag but never `patient.voided`, and on master `getPatients("RTY", singletonList(type2), false, null, null)` already comes back with 432. What changes here is reach: a type-name query lists everyone holding the type, where before you needed one patient's identifier value.

I would keep `assertEquals(3, ...)`, since it is the only assertion here that would catch over-matching, and add a line noting that 432 is voided and that this overload does not filter that, so whoever fixes it later does not read the 3 as intended. The fix belongs in its own ticket rather than widening this PR.

api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
```diff
@@ -299,4 +301,52 @@ public void getDuplicatePatientsByAttributes_shouldReturnVoidedPatientsWithDupli
...
+	@Test
+	public void findPatients_shouldFindPatientsByIdentifierTypeName() {
+		String identifierTypeName = "Old Identification Number";
+		List<Patient> patients = hibernatePatientDao.findPatients(identifierTypeName, false, null, null);
+
+		List<Integer> ids = patients.stream().map(Patient::getPatientId).collect(Collectors.toList());
+
+		assertThat(ids, hasItems(2, 6));
+		assertThat(ids, not(hasItem(432))); // voided patient excluded
+	}
```
dkayiwa, line 384: I would add one more test here, because nothing in the suite pins the property the exact-name lookup exists for: that a query matching only part of a type name must not match at all.

I checked by replacing `lower(pit.name) = lower(:name)` with `lower(pit.name) like concat('%', lower(:name), '%')`, which is a change someone could plausibly make later to support partial type-name search. All three of these tests stayed green, while `findPatients("Number", false, null, null)` went back to returning 2, 6, 7, 8 and `getCountOfPatients("Number", false)` back to 4, which is the page-filling behavior from the earlier round. On the current head both come back empty and 0.

A single assertion that `findPatients("Number", false, null, null)` is empty would catch that. Not blocking, but it is the cheapest guard for the one thing the rework is for.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
dkayiwa: Still nothing exercises `matchIdentifierExactly = true`: all three new tests pass `false`, so removing the `matchExactly ? Collections.emptyList()` guard would not fail anything. On the head `getPatients("Old Identification Number", singletonList(type2), true, null, null)` comes back empty, so pinning that is a one-line test next to the others.

api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
```diff
@@ -30,8 +30,11 @@
 import static java.util.Collections.singletonList;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.hasItems;
+import static org.hamcrest.Matchers.not;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.wildfly.common.Assert.assertTrue;
```
dkayiwa: This one needs changing before merge. It is `org.wildfly.common.Assert`, not JUnit's, and its `assertTrue(boolean)` is a bare `assert` statement (`assert expr : CommonMessages.msg.expectedBoolean(expr)`), so the two tests below only assert anything while JVM assertions are enabled. Surefire enables them by default, which is why CI is green.

Take that default away and both guards disappear. I put the bugs back in, dropping the `matchExactly ? Collections.emptyList()` guard and changing `lower(pit.name) = lower(:name)` to `like concat('%', lower(:name), '%')`, then ran:

    mvn -pl api test -Dtest='HibernatePatientDAOTest#findPatients_shouldNotMatchPartialIdentifierTypeName+getPatients_shouldNotMatchIdentifierTypeNameWhenMatchingIdentifierExactly' -DenableAssertions=false

That reports BUILD SUCCESS, 2 tests, 0 failures, against code carrying both bugs. So the two regressions these tests exist to catch would ship green. With assertions on they do fail, but with `java.lang.AssertionError: COM01003: Internal error: Assertion failure: Expected boolean value to be false`, which names neither the expectation nor the actual value and reads like an infrastructure fault. Worth noting too that nothing here declares `wildfly-common`; it arrives five levels down through `infinispan-core -> jboss-threads`, so test compilation is riding on that transitive staying where it is.

Name and arity are identical, so the import line is the whole fix, and I checked that the JUnit version still fails on both mutations with `-DenableAssertions=false`:

```suggestion
import static org.junit.jupiter.api.Assertions.assertTrue;
```

If you would rather match the hamcrest style used elsewhere in the file, `assertThat(patients, is(empty()))` also names the offending patients when it fails. (`ScheduledWithLockTest` imports the same class, which is likely where the IDE picked it up. That one is pre-existing, and I would leave it alone here.)

api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
```diff
@@ -299,4 +302,77 @@ public void getDuplicatePatientsByAttributes_shouldReturnVoidedPatientsWithDupli
...
+		List<Patient> patients = hibernatePatientDao.findPatients(identifierTypeName, false, null, null);
+
+		List<Integer> ids = patients.stream().map(Patient::getPatientId).collect(Collectors.toList());
+
+		assertThat(ids, hasItems(2, 6));
+		assertThat(ids, not(hasItem(432))); // voided patient excluded
+	}
+
+	/**
+	 * @see HibernatePatientDAO#getPatients(String, boolean, Integer, Integer)
```
dkayiwa: This points at `getPatients(String, boolean, Integer, Integer)`, which is a real but different overload (`HibernatePatientDAO` line 220). The test calls the five-arg `getPatients(String, List, boolean, Integer, Integer)`, the one the two tests further up reference. Documentation only, so entirely optional, though you will be back in this file for the import above anyway.

```suggestion
	 * @see HibernatePatientDAO#getPatients(String, List, boolean, Integer, Integer)
```

api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
IamMujuziMoses: Thank you for catching this. Sorry, I missed this and appreciate the detailed explanation and suggestions. I’ll update and use hamcrest assertThat and make the recommended changes accordingly. I’ll also update the @see reference while I’m in the file.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -863,6 +870,12 @@ private SearchPredicate getPatientIdentifierSearchPredicate(SearchPredicateFacto
 
 	}
 
+	private List<Integer> getIdentifierTypeIdsByName(String name) {
+		return sessionFactory.getCurrentSession().createQuery("select pit.patientIdentifierTypeId "
+		        + "from PatientIdentifierType pit " + "where lower(pit.name) = lower(:name)",
+		    Integer.class).setParameter("name", name).getResultList();
```
dkayiwa: A leading or trailing space in the query turns this lookup off entirely, since the comparison is exact while every other clause in the predicate goes through an analyzer and does not care about it. On the head, `findPatients("Old Identification Number ", false, null, null)` comes back empty and `getCountOfPatients("Old Identification Number ", false)` gives 0, where the same string without the space returns patients 2 and 6 and counts 2. Nothing throws, so it reads as "no such patients", and a type name pasted out of the admin screen carries a trailing space often enough that someone will hit it.

Trimming the parameter covers it. With this in place both the leading- and trailing-space variants return 2 and 6, and all 15 tests in the class still pass, `findPatients_shouldNotMatchPartialIdentifierTypeName` included. I have written it out across the whole statement because the extra characters push the one-liner past the wrap point, so adding just `name.trim()` in place would come back reformatted on your next build:

```suggestion
		return sessionFactory.getCurrentSession()
		        .createQuery("select pit.patientIdentifierTypeId " + "from PatientIdentifierType pit "
		                + "where lower(pit.name) = lower(:name)",
		            Integer.class)
		        .setParameter("name", name.trim()).getResultList();
```

Not blocking.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -863,6 +870,14 @@ private SearchPredicate getPatientIdentifierSearchPredicate(SearchPredicateFacto
 
 	}
 
+	private List<Integer> getIdentifierTypeIdsByName(String name) {
+		return sessionFactory.getCurrentSession()
+		        .createQuery("select pit.patientIdentifierTypeId " + "from PatientIdentifierType pit "
+		                + "where lower(pit.name) = lower(:name)",
```
dkayiwa, line 876: Should this skip retired types? It is the one name-to-type lookup in core that doesn't: `getPatientIdentifierTypes(name, format, required, hasCheckDigit)` in this class adds `isFalse(retired)` unconditionally, and `getPatientIdentifierTypeByName` is built on it.

With nothing holding the standard dataset's retired type 4, `findPatients("Social Security Number", false, null, null)` is empty (so is `findPatients("Test National ID No", ...)`, the non-retired unheld type, which tells me no name or attribute in the dataset matches those words). Give patient 7 an identifier of type 4 and the same search returns 7, while `getPatientIdentifierTypeByName("Social Security Number")` still returns null and `getPatientIdentifierTypes("Social Security Number", null, null, null)` still returns 0 rows.

I would keep what you have. Retired types are the legacy ones people most want to search for, and hiding their holders while their identifier values stay findable is the stranger half of the split. It is worth owning rather than inheriting though, and if you keep it, a test would stop someone later "fixing" the inconsistency by routing this through `getPatientIdentifierTypes`.

api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
```diff
@@ -299,4 +302,77 @@ public void getDuplicatePatientsByAttributes_shouldReturnVoidedPatientsWithDupli
...
+	 */
+	@Test
+	public void findPatients_shouldFindPatientsByIdentifierTypeName() {
+		String identifierTypeName = "Old Identification Number";
+		List<Patient> patients = hibernatePatientDao.findPatients(identifierTypeName, false, null, null);
+
+		List<Integer> ids = patients.stream().map(Patient::getPatientId).collect(Collectors.toList());
+
+		assertThat(ids, hasItems(2, 6));
+		assertThat(ids, not(hasItem(432))); // voided patient excluded
```
dkayiwa, line 353: One more assertion would be worth having here, because the `.trim()` in `getIdentifierTypeIdsByName` has nothing holding it in place. I took it back out and ran the class: 15 tests, 0 failures, with `findPatients("Old Identification Number ", false, null, null)` empty again, which is the case it was added for.

This line catches it, failing on the untrimmed version with `a collection containing <2> was empty`, and `spotless:apply` leaves it alone:

```suggestion
		assertThat(ids, not(hasItem(432))); // voided patient excluded

		// a type name pasted from the admin screen often carries surrounding whitespace, which the lookup trims
		List<Patient> padded = hibernatePatientDao.findPatients("  Old Identification Number  ", false, null, null);
		assertThat(padded.stream().map(Patient::getPatientId).collect(Collectors.toList()), hasItems(2, 6));
```

The `lower()` next to it cannot be pinned the same way, worth knowing before you reach for it: the test H2 URL sets `IGNORECASE=TRUE`, so `pit.name = :name` passes all 15 too. It does matter on Postgres, there just isn't a test here that can tell.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
IamMujuziMoses: Thanks for the reviews. I kept the current behavior and added a regression test to make that decision explicit. The test verifies that a patient with an identifier of a retired identifier type is still found when searching by that retired type’s name.

ibacher:
Apologies for jumping into this ticket kind of late. So, yeah, when I created this, I kind of went from the `// TODO` comment to thinking that was functionality we had dropped. However, it seems like we've just never implemented this search through the search index at all (I've now spot checked 1.9, 1.11, 2.0.0, 2.1.0, 2.5.0, 2.6.0). 2.1.0 added the _indexing_ on `PatientIdentifier.identifierType`, but this has only ever been used to filter the returned values in a patient search as we still do, so this is genuinely new behavior.

From _that_ perspective, this is probably completing work that was intended to be done back in the 2.1.0 days, but it is worth considering whether it makes sense. The search index is _most_ useful where:

1. We need full-text search capabilities, often with scoring.
2. There is a large volume of items, so that running the full-text search in SQL becomes painful.

On reflection, I actually think that it's exceedingly _unlikely_ that there's enough volume of identifier types to make sense for us to search them through the full-text index. Really, I can't imagine a system with more than 20 types, which is the kind of data volume we could load into memory and probably not notice the performance hit.

Since using the search index means that we, first, query the index and then, second query the database to load the actual record, the overhead of doing that probably doesn't make this make sense.

Thank you for the contribution and I'm really sorry for the misleading ticket!

@dkayiwa Unless you disagree with my assessment? I don't want to just close this since it does, in fact, correctly implement what the ticket asked for.

dkayiwa:
@ibacher that is a fair assessemt. 👍

PR #6418, opened 2026-07-31
TRUNK-6721: Replace stack-inspection guard on UserDAO password methods with a cheaper thread-scoped permit

jwnasambu:
## Issue I worked on

see https://openmrs.atlassian.net/browse/TRUNK-6721

  No? -> execute above command


jwnasambu:
@claude review

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -649,18 +611,8 @@ public LoginCredential getLoginCredentialByUuid(String uuid) {
...
-		if (possibleFrame.isEmpty()) {
-			throw new DAOException("Could not determine if saveUser() was called from appropriate place");
-		} else {
-			var callerClass = possibleFrame.get();
-			if (!HibernateUserDAO.class.equals(callerClass)
-			        && !"org.openmrs.api.db.UserDAOTest".equals(callerClass.getName())
-			        && !"org.openmrs.api.UserServiceTest".equals(callerClass.getName())) {
-				throw new DAOException("Illegal attempt to change user password from unknown caller");
-			}
+		if (!UserServiceImpl.UserPasswordGuard.isPermitted()) {
```
dkayiwa: This check lands after the write it is meant to stop has already been staged. `changeQuestionAnswer(User, String, String)` above (line 452) sets the secret question, the hashed answer, `dateChanged` and `changedBy` on the managed `LoginCredential`, and only then calls `updateLoginCredential`. Because `HibernateUtil.saveOrUpdate` returns early for an entity the session already manages, the UPDATE never comes from this call at all; it comes from Hibernate dirty checking at flush time.

So if merged as-is, a caller that reaches `dao.changeQuestionAnswer(user, q, a)` without a permit gets `DAOException("Illegal attempt to change user password from unknown caller")` while the new secret question and hashed answer still land in the users row, for any caller that catches the exception instead of letting it roll the transaction back (`DAOException` is unchecked, and core itself swallows `DAOException`/`APIException` in about a dozen places). I ran this against the branch rather than reading it off the diff: with the guard throwing, a flush and a reload inside the same transaction come back with `secret_question` set to my probe value and a non-null `secret_answer`. The exception makes the operation look blocked, so a test written with `assertThrows` passes while the row changes underneath, which is the part CI cannot catch. This check is also the only thing standing on that path, since a direct DAO call skips the `@Authorized(EDIT_USER_PASSWORDS)` that `UserService.changeQuestionAnswer` carries.

On master this path never threw at all (the stack check saw `HibernateUserDAO` as the immediate caller here and let it through), so it is the permit model that introduces the throw-but-still-write combination, not an existing hole.

Cheapest fix is to take the permit check off the transitive path and put it at the top of `changeQuestionAnswer(User, String, String)`, ahead of the first setter. Both overloads route through that method and `UserServiceImpl.changeQuestionAnswer` already holds the permit, so nothing legitimate changes. It would also bring this in line with the rest of the guard: every other guarded entry point checks before it touches anything, and `changeQuestionAnswer` is the only one whose check is inherited from a callee it reaches after the mutation. Those two overloads are also the only newly rejected paths with no test, and they are the ones most worth covering: on master any caller at all could reach them, whereas the four methods you did add tests for could only be called directly from the whitelisted test classes.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -845,4 +885,28 @@ public List<Class<?>> getRefTypes() {
...
+			if (depth <= 0) {
+				DEPTH.remove();
+			} else {
+				DEPTH.set(depth);
+			}
+		}
+
+		public static boolean isPermitted() {
+			return DEPTH.get() > 0;
+		}
```
dkayiwa: Eight hand-paired `enter()` / try / finally / `exit()` blocks is a lot of ceremony for one bit of state, and the correctness of the whole guard rests on every one of those pairs being right. They all are today. If a later edit ever loses an `exit()` (an early return slipped in above the try, say), the permit stays set on that thread for as long as the thread lives, so on a pooled request thread every later request served by it walks straight past all five DAO checks, and nothing surfaces the imbalance, since `exit()` quietly absorbs an unmatched call in either direction.

`AutoCloseable` makes that unrepresentable and collapses the call sites to one line each:

```java
try (var permit = UserPasswordGuard.acquire()) {
    return dao.saveUser(user, password);
}
```

Two things come free with it. The two new tests could stop calling `setAccessible(true)` on private statics and exercise a real API instead, and the class would gain an obvious home for the Javadoc and `@since` tag that new public API here usually carries: `UserPasswordGuard` and `isPermitted()` have neither right now, and they are the contract `HibernateUserDAO` depends on.

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1645,7 +1645,7 @@ public void getUserByActivationKey_shouldGetUserByActivationKey() {
 		credentials.setActivationKey(
 		    "b071c88d6d877922e35af2e6a90dd57d37ac61143a03bb986c5f353566f3972a86ce9b2604c31a22dfa467922dcfd54fa7d18b0a7c7648d94ca3d97a88ea2fd0:"
 		            + tokenTime);
-		dao.updateLoginCredential(credentials);
+		dao.setUserActivationKey(credentials);
```
dkayiwa: These four call sites work, and naming the activation-key method reads better than the old `updateLoginCredential` call. They do highlight something the guard cannot fix though: `setUserActivationKey(LoginCredential)` in `HibernateUserDAO` (line 750) has the same one-line body as the guarded `updateLoginCredential` and no permit check, so the same row stays writable by anyone holding the `userDAO` bean. I checked it rather than assuming: evict a `LoginCredential`, set `salt` and `hashedPassword` on it, call `setUserActivationKey` with no permit held, and the reloaded credential matches the password I picked.

Should the permit cover `setUserActivationKey` too, or is the check on `updateLoginCredential` accepted as a speed bump? I would extend it, since `UserServiceImpl.setUserActivationKey` is its only caller and can hold the permit as cheaply as the other eight sites do. It is all pre-existing so it needn't hold up this PR, but the guard reads stronger than it is, and this diff is where that becomes visible.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -845,4 +863,58 @@ public List<Class<?>> getRefTypes() {
...
+
+		/**
+		 * Acquires one level of permit for the current thread. The returned permit must be closed when
+		 * done, typically with try-with-resources, to release the acquired level.
+		 *
+		 * @return the acquired permit
+		 * @since 3.0.0
+		 */
+		public static UserPasswordGuard acquire() {
```
dkayiwa: This needs to go back to `private` before merge. As `public static`, `acquire()` lets any caller in the JVM raise the permit and then walk through all six DAO checks, which is the single property TRUNK-6721 asks of the permit ("Only `UserServiceImpl` can raise the permit; `enter()`/`exit()` are not callable from outside it"). If merged as-is, a module holding the `userDAO` bean writes

```java
try (var p = UserServiceImpl.UserPasswordGuard.acquire()) {
	dao.changeHashedPassword(victim, hash, salt);
}
```

and owns the account, skipping the `@Authorized(EDIT_USER_PASSWORDS)` that `UserService.changeHashedPassword` carries. I ran exactly that from a probe class in `org.openmrs.probe`, a package that appears in no whitelist, old or new, and the victim's password came back as the one I picked. The `@since 3.0.0` tag makes it worse than a slip: once 3.0.0 ships, that is supported API and we cannot take it back.

`private` works. I flipped it and `src/main` compiled untouched, since `UserServiceImpl` reaches a private member of its own nested class as a nestmate. The only breakage is seven test call sites, and I ran replacements for all seven green (UserServiceTest 119, UserDAOTest 17):

* The four activation-key tests in `UserServiceTest` do not need the DAO call at all. `dao.getLoginCredential(user)` returns a managed entity, so `credentials.setActivationKey(...)` persists by dirty checking; deleting the try-with-resources block leaves all four passing.
* `changePasswordString_shouldAllowChangingPasswordWhenPermitIsHeld` reads better through `Context.getUserService().changePassword(...)`, the way `changePassword_shouldNotOverwriteUserSecretQuestionOrAnswer` just above it already does.
* `userPasswordGuard_shouldAllowNestedPermitAcquisition` goes with it. Nothing in the service nests `acquire()`, and the recursion the ticket actually names (`updateUserPassword` -> `saveUser` re-checking under one permit) is already covered by `userPasswordGuard_shouldAllowReEntrantCalls`.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
dkayiwa: `updateLoginCredential` is the open door now. With its check gone, anything holding the `userDAO` bean writes an arbitrary salt and password hash into any user's row: no permit, no privilege, not even an authenticated session. I ran it from a probe class in `org.openmrs.probe`, a package in no whitelist, old or new. Fetch admin's `LoginCredential`, evict it, `Context.logout()`, set a salt and a hash for a password I chose, call `dao.updateLoginCredential(lc)`, flush and clear, and `Context.authenticate("admin", <my password>)` then comes back as user 1. On master that probe stops at the same line with `DAOException: Illegal attempt to change user password from unknown caller`, so the takeover is new here. It also lands on the method whose entity carries the javadoc "This should never be used by anything except for UserDAO and UserService methods that change passwords", and `updateLoginCredential` is one of the five methods TRUNK-6721 names in its first acceptance criterion.

Taking the check off the transitive path and putting one at the top of `changeQuestionAnswer` was half of what I meant; keeping this one as well is the part I should have spelled out. Nothing legitimate changes with both in place, since the only caller of `updateLoginCredential` in the tree is `changeQuestionAnswer`, which now takes the permit before it touches anything. I put the three lines back and ran the suites: UserDAOTest 18 and UserServiceTest 119, all green.

```java
public void updateLoginCredential(LoginCredential credential) {
	if (!UserServiceImpl.UserPasswordGuard.isPermitted()) {
		throw new DAOException("Illegal attempt to change user password from unknown caller");
	}

	HibernateUtil.saveOrUpdate(sessionFactory.getCurrentSession(), credential);
}
```

The new rejection tests cover six entry points; this one would be the seventh.

jwnasambu:
@claude review

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
dkayiwa: The check on `updateLoginCredential` still needs to go back before this merges. At f51fa27 the method has no guard at all, and it is one of the five methods TRUNK-6721 names in its first acceptance criterion.

If merged as-is, anything holding the `userDAO` bean writes an arbitrary salt and password hash into any user's row: no permit, no privilege, not even an authenticated session. I re-ran the probe against f51fa27 from a class in `org.openmrs.probe`, a package in no whitelist old or new. Fetch admin's `LoginCredential`, evict it so the write can only come from this call and not from dirty checking on a managed entity, `Context.logout()`, set a salt and a hash for a password I chose, call `dao.updateLoginCredential(lc)`, flush and clear. It returns without throwing, the reloaded credential matches my password, and `Context.authenticate("admin", <my password>)` comes back as user 1. Against 142b13f4 the same probe stops on that line with `DAOException: Illegal attempt to change user password from unknown caller`, so the takeover arrives with this branch.

Putting the check back changes nothing legitimate: `changeQuestionAnswer` is the only caller in the tree, and it is only reachable with the permit already held. With these three lines on top of f51fa27 the probe dies with that DAOException, and both suites stay green (UserServiceTest 119, UserDAOTest 17).

```java
public void updateLoginCredential(LoginCredential credential) {
	if (!UserServiceImpl.UserPasswordGuard.isPermitted()) {
		throw new DAOException("Illegal attempt to change user password from unknown caller");
	}

	HibernateUtil.saveOrUpdate(sessionFactory.getCurrentSession(), credential);
}
```

The six new rejection tests cover the other entry points, so this one still wants a seventh.

jwnasambu:
@Ian Bacher, @dkayiwa  Kindly feel free to review this  PR at your convenient time please!

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
dkayiwa: Still missing at 86e3281d: `updateLoginCredential` has no permit check. I diffed the four files in this PR against f51fa27 and they come back byte-identical, so the takeover above holds unchanged. Anything holding the `userDAO` bean writes an arbitrary salt and password hash into any user's row, with no permit, no privilege and no authenticated session.

The three lines in the comment above are the whole fix, and it is the last thing outstanding here.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -845,4 +863,57 @@ public List<Class<?>> getRefTypes() {
...
+	/**
+	 * Guards the password- and credential-mutating methods on {@link UserDAO} so they can only be
+	 * invoked by service code that holds a permit for the current thread. The service acquires a permit
+	 * before calling the DAO and releases it when the returned instance is closed, normally via a
+	 * try-with-resources statement. Permits are re-entrant: each acquisition deepens the held permit by
+	 * one level and only the outermost close releases it.
+	 *
+	 * @since 3.0.0
+	 */
+	public static final class UserPasswordGuard implements AutoCloseable {
```
ibacher: This needs to be private. The `isPermitted` call should be on the service itself.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -78,17 +76,8 @@ public HibernateUserDAO(SessionFactory sessionFactory) {
...
-
-		if (possibleFrame.isEmpty()) {
-			throw new DAOException("Could not determine if saveUser() was called from appropriate place");
-		} else {
-			var callerClass = possibleFrame.get();
-			if (!UserServiceImpl.class.equals(callerClass) && !HibernateUserDAO.class.equals(callerClass)
-			        && !"org.openmrs.api.db.UserDAOTest".equals(callerClass.getName())) {
-				throw new DAOException("Illegal attempt to save user from unknown caller");
-			}
+		if (!UserServiceImpl.UserPasswordGuard.isPermitted()) {
```
ibacher: This should only guard attempts to change the password. It should not be required for other scenarios.

api/src/test/java/org/openmrs/api/db/UserDAOTest.java
ibacher, line 1: These are the tests for the UserDAO. They shouldn't use the context or the service, but use the DAO directly. This may need some additional mechanism.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -845,4 +857,80 @@ public List<Class<?>> getRefTypes() {
 		return Arrays.asList(Role.class, Privilege.class, User.class);
 	}
 
+	/**
+	 * Returns whether the current thread holds a password-guard permit. This is the service-level API
+	 * for the guard check that {@link org.openmrs.api.db.UserDAO} implementations use to verify they
+	 * were called from a service method rather than directly.
+	 *
+	 * @return true if the current thread holds a permit
+	 * @since 3.0.0
```
ibacher, line 866: Wasn't this a backportable ticket?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -845,4 +857,80 @@ public List<Class<?>> getRefTypes() {
...
+	 */
+	public static boolean isPasswordGuardPermitted() {
+		return UserPasswordGuard.isPermitted();
+	}
```
ibacher, line 870:
```suggestion
	public static boolean canChangePassword() {
		return UserPasswordGuard.isPermitted();
	}
```

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -845,4 +857,80 @@ public List<Class<?>> getRefTypes() {
...
+	/**
+	 * Acquires a password-guard permit for the current thread. Intended for use in tests that need to
+	 * call password-guarded DAO methods directly.
+	 *
+	 * @return the acquired permit, which must be closed when done
+	 * @since 3.0.0
+	 */
+	public static AutoCloseable acquirePasswordGuardPermit() {
+		return UserPasswordGuard.acquire();
+	}
```
ibacher, line 897: Alright, so in my last comments I forgot what the ticket design required (which did have a public `UserPasswordGuard` but with the _only_  public function being `isPermitted()`. The ticket is _extremely_ clear on the fact that the `acquire()` (and `close()` functions) _must_ be private. That is the _whole_ protection mechanism, so exposing them as public API makes this useless as a change.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -78,18 +76,6 @@ public HibernateUserDAO(SessionFactory sessionFactory) {
 	 */
 	@Override
 	public User saveUser(User user, String password) {
-		var possibleFrame = STACK_WALKER
```
ibacher, line 81: So it's not sufficient just to remove these lines. We need to replace them with a guard somewhere.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
ibacher: Did you miss this?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
ibacher: I wanted to replace this with a function with that name, not create two functions with the same underlying implementation.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
ibacher: Like I said here, _this cannot be a public method_. Delete it.

api/src/test/java/org/openmrs/api/UserServiceTest.java
```diff
@@ -1645,7 +1645,6 @@ public void getUserByActivationKey_shouldGetUserByActivationKey() {
 		credentials.setActivationKey(
 		    "b071c88d6d877922e35af2e6a90dd57d37ac61143a03bb986c5f353566f3972a86ce9b2604c31a22dfa467922dcfd54fa7d18b0a7c7648d94ca3d97a88ea2fd0:"
 		            + tokenTime);
-		dao.updateLoginCredential(credentials);
```
ibacher, line 1648: We shouldn't be changing these tests...

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -78,17 +76,8 @@ public HibernateUserDAO(SessionFactory sessionFactory) {
...
-
-		if (possibleFrame.isEmpty()) {
-			throw new DAOException("Could not determine if saveUser() was called from appropriate place");
-		} else {
-			var callerClass = possibleFrame.get();
-			if (!UserServiceImpl.class.equals(callerClass) && !HibernateUserDAO.class.equals(callerClass)
-			        && !"org.openmrs.api.db.UserDAOTest".equals(callerClass.getName())) {
-				throw new DAOException("Illegal attempt to save user from unknown caller");
-			}
+		if (password != null && !UserServiceImpl.isPasswordGuardPermitted()) {
```
ibacher, line 79: We do not skip this guard if the password is `null`. The whole goal of the guard is to ensure it's coming from an appropriate place.

ibacher:
@jwnasambu Sorry to do this. I'm going to take over this PR so we can get it landed for the next RefApp release.

PR #6420, opened 2026-08-01
TRUNK-6729:  Let batch loading engage on the identifier and concept search paths

Rocker810:
<!--- 'TRUNK-JiraIssueNumber: JiraIssueTitle' -->
  This attaches an empty `GraphSemantic.FETCH` entity graph to the identifier search query only,                                                                                     
  leaving the patient an uninitialized proxy so the batched load reaches the database first.                                                                                         
                                                                                                                                                                                     
  - `SearchQueryUnique` — new `withLoadingOptions(...)`, applied to the page fetch only. Additive;                                                                                   
    unset, queries build exactly as before.                                                                                                                                          
  - `HibernatePatientDAO` — `unfetchedIdentifierHits()`, attached to both identifier searches.                                                                                       
                                                                                                                                                                                     
  Prepared statements, 50-patient fixture: page of 5 → 20 before / 14 after; page of 40 →                                                                                            
  55 before / 14 after. Flat instead of one trip per extra result.                                                                                                                   
                                                                                                                                                                                     
  The mapped fetch strategy is unchanged outside search (covered by a test). The concept path                                                                                        
  measured flat already, so it gets a regression test and no code change.
  
## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-6729

  No? -> execute above command


Rocker810:
@claude review

api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
```diff
@@ -299,4 +313,81 @@ public void getDuplicatePatientsByAttributes_shouldReturnVoidedPatientsWithDupli
...
+		                + "but a page of %d cost %d trips against %d for a page of %d",
+		        LARGE_PAGE, largePageTrips, smallPageTrips, SMALL_PAGE));
+	}
+
+	/**
+	 * The search's loading options are scoped to that search, so any other path must still eagerly
+	 * populate the patient - otherwise callers reading it after the session closes would start failing.
+	 */
+	@Test
+	public void getPatientIdentifiers_shouldStillEagerlyFetchThePatientOutsideSearch() {
```
jwnasambu, line 366: Thanks @Rocker810 for this. Kindly `getPatientIdentifiers_shouldStillEagerlyFetchThePatientOutsideSearch ` passes for the wrong reason. getPatientIdentifiers explicitly does root.join("patient") (HibernatePatientDAO.java:295), and an explicit join forces eager loading regardless of the mapped fetch strategy. So if PatientIdentifier.patient were remapped to LAZY, this test would still pass and the comment's claim "the mapped eager fetch of PatientIdentifier.patient must be unchanged outside search" would not actually be verified. I recommend rewording the comment to say the test guards this DAO method's behavior or asserting through a path without the explicit join if the mapping invariant is the real concern.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -828,14 +833,19 @@ public List<Patient> findPatients(String query, boolean includeVoided, Integer s
...
-		            "person.personId", (List<PersonName> pNs) -> multiLoadPatients(pNs, pN -> pN.getPerson().getPersonId()))
-		                .join(SearchQueryUnique.newQuery(PersonAttribute.class,
-		                    f -> personQuery.getPatientAttributeQuery(f, query, includeVoided), "person.personId",
-		                    (List<PersonAttribute> pAs) -> multiLoadPatients(pAs, pA -> pA.getPerson().getPersonId())))),
+		patients = SearchQueryUnique.search(searchSessionFactory,
+		    SearchQueryUnique
+		            .newQuery(PatientIdentifier.class,
+		                f -> newPatientIdentifierSearchPredicate(f, query, includeVoided, false), "patient.personId",
+		                (List<PatientIdentifier> pIs) -> multiLoadPatients(pIs, pI -> pI.getPatient().getPersonId()))
+		            .withLoadingOptions(unfetchedIdentifierHits())
```
jwnasambu, line 841: The joined search path behind `getPatients(query, includeVoided, start, length) ` the main patient search received the same .`withLoadingOptions(unfetchedIdentifierHits())` fix, but the new trip-count test covers only the identifier-types overload (HibernatePatientDAO.java:807). I measured the joined path and it is flat today, so this is a coverage gap, not a defect. Because this is the search most users actually hit and the batching mechanism is subtle, a page-growth assertion mirroring `getPatients_shouldNotIssueMoreDatabaseTripsAsTheIdentifierSearchPageGrows` for the joined path would lock in the behavior.

api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
Rocker810: Thanks for the close read. Flipping PatientIdentifier.patient to @ManyToOne(fetch = FetchType.LAZY) makes the test fail (expected: but was: ), so the mapping invariant is genuinely covered. The root.join("patient") in getPatientIdentifiers is a filter join, not a fetch join, it's there so the query can predicate on the patient's voided flag, and it doesn't add the association to the fetch plan the way root.fetch(...) would. Under LAZY you get uninitialized proxies. The Context.clearSession() at the top is what keeps that honest.

api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
```diff
@@ -299,4 +314,109 @@ public void getDuplicatePatientsByAttributes_shouldReturnVoidedPatientsWithDupli
...
+
+		return (int) sessionFactory.getStatistics().getPrepareStatementCount();
+	}
+
+	private void createIdentifierSearchFixture() {
+		PatientIdentifierType identifierType = Context.getPatientService().getPatientIdentifierType(2);
+
+		for (int i = 0; i < SEARCH_FIXTURE_SIZE; i++) {
+			Patient patient = new Patient();
+			patient.addName(new PersonName("Bounded", null, "Patient" + i));
```
ibacher: Problematically, you've used the same string for both the person name and the `SEARCH_FIXTURE_TOKEN` since the search isn't case-sensitive. Use something else here or else we don't actually know what won out.

api/src/test/java/org/openmrs/api/db/hibernate/HibernatePatientDAOTest.java
```diff
@@ -299,4 +314,109 @@ public void getDuplicatePatientsByAttributes_shouldReturnVoidedPatientsWithDupli
...
 		assertThat(duplicatePatients.size(), equalTo(2));
 	}
+
+	/**
+	 * Asserts the shape of the cost rather than an absolute number, which would break on any change
+	 * shifting the fixed overhead. Before {@link PatientIdentifier#getPatient()} was left unresolved
+	 * while loading hits, the difference here was roughly one round trip per extra result.
+	 */
+	@Test
+	public void getPatients_shouldNotIssueMoreDatabaseTripsAsTheIdentifierSearchPageGrows() {
```
ibacher: These tests seem like they would be better off in their own class. They have very little in common with the existing tests, are appended onto the end instead of being integrated, etc. Same goes for the concept tests.

ibacher:
@Rocker810 Thanks for this! This looks like a solid implementation of option a. I ran some performance testing with this and a strawman option b, and it turns out that option a winds up allocating excess `PatientIdentifier` objects in a way that scales linearly with the page depth whereas option b allows us to avoid that cost, so I'm going to merge a version of that in.

ibacher (approved):

PR #6171, opened 2026-06-07
TRUNK-6649: Archiving Voided Data - POC

shivvani-r:
## Description of what I changed
This PR introduces a POC for the GSoC 2026 project, Archiving Voided Data. It implements an automated background archiving mechanism for voided observations implemented using JobRunr.
It covers the complete lifecycle of observation archiving: database schema creation, background scheduling, transparent retrieval via the API.

### Schema & Migration (Liquibase)

- **`obs_archive` table** — mirrors the `obs` schema with additional `archived_by` and `date_archived` columns.
- **`obs_reference_range_archive` table** — mirrors `obs_reference_range` for archived obs.
- **Indexes** on `obs_group_id`, `encounter_id`, `person_id`, `concept_id`, and `date_voided` in the archive table (along with a unique constraint on `uuid`).
- **Scheduler task** registered: "Observation Archiving Job" (runs every 24h, `start_on_startup = true`).
- **Drops the `previous_version` FK** on `obs` to allow archiving old versioned rows without constraint violations.

### Core Seed Data

- **Global properties** seeded from `OpenmrsConstants.CORE_GLOBAL_PROPERTIES()`: `obs.archive.enabled`, `obs.archive.batch_size`, `obs.archive.retention_days`, `obs.archive.last_processed_obs_id`.

### New Domain Entities

- **`ObsArchive.java`**: JPA entity mapped to `obs_archive`. Stores `obs_group_id` and `previous_version` as plain integer columns (no FK) to avoid dangling references.
- **`ObsArchiveReferenceRange.java`**: JPA entity mapped to `obs_reference_range_archive`, linked 1-to-1 with `ObsArchive`.

### New Service / Helper

- **`ObsArchiveHelper.java`**: Spring bean (`obsArchiveHelper`) wrapping all archive read/write operations: `isArchived()`, `hasArchivedChildren()`, `getObsFromArchive()`, `getObsFromArchiveByUuid()`, `restoreFromArchive()`. 

### Updated Purge Behavior

- **`ObsService.purgeObs`**, **`EncounterService.purgeEncounter`** (both overloads), and **`OrderService.purgeOrder`**: The behavior of these methods has changed to support archiving. 

### New Scheduler Task

- **`ObsArchivingTask.java`**: Legacy scheduler task class (extends `AbstractTask`).
- **`ObsArchivingTaskData.java`**: JobRunr-compatible task data class.
[...]

PR #6424, opened 2026-08-03
TRUNK-6738: Create Database Schema for Observation Archiving

shivvani-r:
This PR introduces the initial Liquibase database schema migrations required to create the `obs_archive` and `obs_reference_range_archive` table for observation archiving. It also updates the DatabaseUpdater integration tests to reflect the new total number of changesets (917).

*(Part 1 of the Observation Archiving stack)*

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-6738

- [ ] I have **added tests** to cover my changes.


shivvani-r:
cc: @dkayiwa @Muta-Jonathan

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,143 @@
...
+		</createIndex>
+		<createIndex indexName="idx_obs_archive_person_id" tableName="obs_archive">
+			<column name="person_id"/>
+		</createIndex>
+		<createIndex indexName="idx_obs_archive_concept_id" tableName="obs_archive">
+			<column name="concept_id"/>
+		</createIndex>
+	</changeSet>
+
+	<changeSet id="TRUNK-6738-register-archiving-task" author="shivvani-r">
```
dkayiwa: This one has to come out before this merges, otherwise OpenMRS won't start.

`created_by` is the legacy column; the one `TaskDefinition` actually maps (through `BaseOpenmrsMetadata`) is `creator`, which TRUNK-6391 added earlier in this same file and never backfilled. I ran the real chain into H2 (2.9.x schema-only + 2.9.x core-data + this update file) and the new row lands as `START_ON_STARTUP=TRUE, CREATED_BY=1, CREATOR=null`. It is also the only row in a fresh 3.0 database with `start_on_startup` true (Auto Close Visits Task is false), which is why nobody has tripped this before. `JobRunrSchedulerService.onStartup()` line 94 then does `new JobRequestAdapter(taskDefinition, taskDefinition.getCreator().getSystemId())` with no null check on that branch (the null-guarded `scheduledBy` local two lines above goes unused), and handing it that TaskDefinition throws `NullPointerException: Cannot invoke "org.openmrs.User.getSystemId()" because the return value of "org.openmrs.scheduler.TaskDefinition.getCreator()" is null`. `Listener.startOpenmrs` calls `onStartup()` at Listener.java:355 inside a catch that shuts the context back down and rethrows as `ServletException`, so the webapp fails to deploy on every 3.0 install and upgrade.

Setting `creator` would clear that crash, but I don't think the row belongs in this PR at all. `org.openmrs.scheduler.tasks.ObsArchivingTask` doesn't exist anywhere in core yet, and `onStartup()` schedules every row it finds regardless of the `started` column, so the job would just fail its class load on a retry loop instead (`LegacyTask` throws a retryable `TaskException`). Registering the task is TRUNK-6743, so I'd delete this changeset here and add it there next to the class.

The empty `creator` column plus that unguarded deref look worth their own ticket, since they'll catch the next person who inserts a scheduler row from liquibase.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,143 @@
...
+			<column name="repeat_interval" valueNumeric="86400"/> <!-- 24 hours -->
+			<column name="start_on_startup" valueBoolean="true"/>
+			<column name="started" valueBoolean="false"/>
+			<column name="date_created" valueDate="2026-06-07T00:00:00"/>
+			<column name="created_by" valueNumeric="1"/>
+			<column name="uuid" value="3c8c279a-8b1e-412e-9a88-7bba5b702344"/>
+		</insert>
+	</changeSet>
 
+	<changeSet id="TRUNK-6738-drop-previous-version-fk" author="shivvani-r">
```
dkayiwa: I'd move this to whichever PR starts actually writing to `obs_archive` (TRUNK-6740 or TRUNK-6742) rather than landing it here. Everything else in this PR is additive and inert until the archiving code exists, but a constraint drop takes effect the moment the update file runs, so if this merges as-is every 3.0 installation loses the referential check on `obs.previous_version`, including the large majority that never turn archiving on, and from then on a bogus `previous_version` is written silently instead of rejected. I confirmed it does drop (after running this file the `previous_version` FK is gone from `obs`), and restoring it later needs yet another migration, since a released update file can't be edited.

It also doesn't finish the job on its own. `Obs.hbm.xml:99` still maps `previousVersion` as a `many-to-one` to `Obs` with no `not-found` handling, so once rows really do leave `obs` a dangling `previous_version` just moves the failure from the database up into Hibernate. Whatever handles that ought to land in the same change as the drop.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,143 @@
...
+			<column name="uuid" type="CHAR(38)">
+				<constraints unique="true"/>
+			</column>
+			<column defaultValueComputed="NULL" name="archived_by" type="INT"/>
+			<column defaultValueComputed="NULL" name="date_archived" type="DATETIME"/>
+		</createTable>
+		
+		<createIndex indexName="idx_obs_archive_uuid" tableName="obs_archive">
+			<column name="uuid"/>
+		</createIndex>
```
dkayiwa: `unique="true"` on the `uuid` column already gets you a unique index, so this adds a second index over the same column. On H2 after running the file I see both `CONSTRAINT_INDEX_8A` (unique, on UUID) and `IDX_OBS_ARCHIVE_UUID` (not unique). On a table meant to accumulate every archived obs that's write cost for nothing, so I'd drop the `createIndex` and keep the constraint.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,143 @@
...
+				<constraints nullable="false" unique="true"/>
+			</column>
+			<column defaultValueComputed="NULL" name="previous_version" type="INT"/>
+			<column name="form_namespace_and_path" type="VARCHAR(255)"/>
+			<column defaultValue="FINAL" name="status" type="VARCHAR(16)">
+				<constraints nullable="false"/>
+			</column>
+			<column name="interpretation" type="VARCHAR(32)"/>
+			<column defaultValueComputed="NULL" name="changed_by" type="INT"/>
+			<column defaultValueComputed="NULL" name="date_changed" type="DATETIME"/>
```
dkayiwa: Unlike everything else in this table, these two have nothing to copy from. I diffed the two tables on a database built from the changelogs: `obs_archive` has `CHANGED_BY`, `DATE_CHANGED`, `ARCHIVED_BY` and `DATE_ARCHIVED` that `obs` doesn't, while `obs` has no column `obs_archive` is missing. `Obs.hbm.xml` doesn't map either one. Deliberate, or leftovers from a template? `archived_by` and `date_archived` below already cover the archive-side provenance.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,143 @@
...
 	</changeSet>
 	
-</databaseChangeLog>
+	<changeSet id="TRUNK-6738-add-obs-archive-tables" author="shivvani-r">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<tableExists tableName="obs_archive"/>
+			</not>
+		</preConditions>
+		<comment>Add obs_archive and obs_archive_reference_range tables and global properties for Archiving Observations</comment>
```
dkayiwa: No global properties are added by this changeset, and this text ends up in DATABASECHANGELOG where it'll send whoever reads it later looking for something that isn't there.
```suggestion
		<comment>Add obs_archive and obs_archive_reference_range tables for Archiving Observations</comment>
```

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: Yeah that makes sense. Will add this later.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: Yeah true. will add later

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: i see, ill remove the index.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: Oops will remove.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: That was an old comment. Will change

shivvani-r:
@Muta-Jonathan

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,102 @@
 		</addColumn>
 	</changeSet>
 	
-</databaseChangeLog>
+	<changeSet id="TRUNK-6738-add-tables-for-obs-archiving" author="shivvani-r">
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<tableExists tableName="obs_archive"/>
+			</not>
+		</preConditions>
```
Muta-Jonathan, line 150: Since this changeset creates both `obs_archive` and `obs_reference_range_archive`, both tables should be included in the precondition check so neither is re-created if a prior migration was interrupted:

```suggestion
		<preConditions onFail="MARK_RAN">
			<and>
				<not>
					<tableExists tableName="obs_archive"/>
				</not>
				<not>
					<tableExists tableName="obs_reference_range_archive"/>
				</not>
			</and>
		</preConditions>
```

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,102 @@
...
+		<comment>Add obs_archive and obs_reference_range_archive tables for Archiving Observations</comment>
+		<createTable tableName="obs_archive">
+			<column name="obs_id" type="INT">
+				<constraints nullable="false" primaryKey="true"/>
+			</column>
+			<column name="person_id" type="INT">
+				<constraints nullable="false"/>
+			</column>
+			<column defaultValueNumeric="0" name="concept_id" type="INT">
```
Muta-Jonathan: Concept IDs start at 1 in OpenMRS. We should drop `defaultValueNumeric="0"` so it doesn't default to an invalid ID when missing.

```suggestion
			<column name="concept_id" type="INT">
```

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,102 @@
...
+			</column>
+			<column defaultValueComputed="NULL" name="location_id" type="INT"/>
+			<column defaultValueComputed="NULL" name="obs_group_id" type="INT"/>
+			<column name="accession_number" type="VARCHAR(255)"/>
+			<column defaultValueComputed="NULL" name="value_group_id" type="INT"/>
+			<column defaultValueComputed="NULL" name="value_coded" type="INT"/>
+			<column defaultValueComputed="NULL" name="value_coded_name_id" type="INT"/>
+			<column defaultValueComputed="NULL" name="value_drug" type="INT"/>
+			<column defaultValueComputed="NULL" name="value_datetime" type="DATETIME"/>
+			<column defaultValueComputed="NULL" name="value_numeric" type="DOUBLE"/>
```
Muta-Jonathan: In Liquibase, columns without `nullable="false"` default to `NULL` automatically. We can remove `defaultValueComputed="NULL"` across these columns to keep the XML clean:

```suggestion
			<column name="encounter_id" type="INT"/>
			<column name="order_id" type="INT"/>
			<column name="obs_datetime" type="DATETIME">
				<constraints nullable="false"/>
			</column>
			<column name="location_id" type="INT"/>
			<column name="obs_group_id" type="INT"/>
			<column name="accession_number" type="VARCHAR(255)"/>
			<column name="value_group_id" type="INT"/>
			<column name="value_coded" type="INT"/>
			<column name="value_coded_name_id" type="INT"/>
			<column name="value_drug" type="INT"/>
			<column name="value_datetime" type="DATETIME"/>
			<column name="value_numeric" type="DOUBLE"/>
```

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,102 @@
...
+			<column name="value_modifier" type="VARCHAR(2)"/>
+			<column name="value_text" type="TEXT"/>
+			<column name="value_complex" type="VARCHAR(1000)"/>
+			<column name="comments" type="VARCHAR(255)"/>
+			<column defaultValueNumeric="0" name="creator" type="INT">
```
Muta-Jonathan: Same here — let's remove `defaultValueNumeric="0"` on `creator`.

```suggestion
			<column name="creator" type="INT">
```

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,102 @@
...
+				<constraints nullable="false" unique="true"/>
+			</column>
+			<column defaultValueComputed="NULL" name="previous_version" type="INT"/>
+			<column name="form_namespace_and_path" type="VARCHAR(255)"/>
+			<column defaultValue="FINAL" name="status" type="VARCHAR(16)">
+				<constraints nullable="false"/>
+			</column>
+			<column name="interpretation" type="VARCHAR(32)"/>
+			<column defaultValueComputed="NULL" name="archived_by" type="INT"/>
+			<column defaultValueComputed="NULL" name="date_archived" type="DATETIME"/>
```
Muta-Jonathan: Removing remaining redundant `defaultValueComputed="NULL"` attributes in `obs_archive`:

```suggestion
			<column name="voided_by" type="INT"/>
			<column name="date_voided" type="DATETIME"/>
			<column name="void_reason" type="VARCHAR(255)"/>
			<column name="uuid" type="CHAR(38)">
				<constraints nullable="false" unique="true"/>
			</column>
			<column name="previous_version" type="INT"/>
			<column name="form_namespace_and_path" type="VARCHAR(255)"/>
			<column defaultValue="FINAL" name="status" type="VARCHAR(16)">
				<constraints nullable="false"/>
			</column>
			<column name="interpretation" type="VARCHAR(32)"/>
			<column name="archived_by" type="INT"/>
			<column name="date_archived" type="DATETIME"/>
```

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,102 @@
...
+		</createTable>
 
+		<createTable tableName="obs_reference_range_archive">
+			<column name="obs_reference_range_id" type="INT" autoIncrement="true">
```
Muta-Jonathan: Since this is an archive table, we want to retain the original `obs_reference_range_id` from the source table rather than generating a new auto-increment ID:

```suggestion
			<column name="obs_reference_range_id" type="INT">
```

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,102 @@
...
+			<column defaultValueComputed="NULL" name="hi_critical" type="DOUBLE"/>
+			<column defaultValueComputed="NULL" name="hi_normal" type="DOUBLE"/>
+			<column defaultValueComputed="NULL" name="low_absolute" type="DOUBLE"/>
+			<column defaultValueComputed="NULL" name="low_critical" type="DOUBLE"/>
+			<column defaultValueComputed="NULL" name="low_normal" type="DOUBLE"/>
+			<column name="uuid" type="CHAR(38)">
+				<constraints unique="true"/>
+			</column>
+			<column defaultValueComputed="NULL" name="archived_by" type="INT"/>
+			<column defaultValueComputed="NULL" name="date_archived" type="DATETIME"/>
```
Muta-Jonathan: Removing `defaultValueComputed="NULL"` and adding `nullable="false"` to `uuid`:

```suggestion
			<column name="hi_absolute" type="DOUBLE"/>
			<column name="hi_critical" type="DOUBLE"/>
			<column name="hi_normal" type="DOUBLE"/>
			<column name="low_absolute" type="DOUBLE"/>
			<column name="low_critical" type="DOUBLE"/>
			<column name="low_normal" type="DOUBLE"/>
			<column name="uuid" type="CHAR(38)">
				<constraints nullable="false" unique="true"/>
			</column>
			<column name="archived_by" type="INT"/>
			<column name="date_archived" type="DATETIME"/>
```

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,102 @@
...
+		</createIndex>
+		<createIndex indexName="idx_obs_archive_date_voided" tableName="obs_archive">
+			<column name="date_voided"/>
+		</createIndex>
+		<createIndex indexName="idx_obs_archive_person_id" tableName="obs_archive">
+			<column name="person_id"/>
+		</createIndex>
+		<createIndex indexName="idx_obs_archive_concept_id" tableName="obs_archive">
+			<column name="concept_id"/>
+		</createIndex>
```
Muta-Jonathan, line 219: Might be worth adding indexes for `date_archived` and `obs_datetime` as well since those will be very common search filters when querying archived obs:

```suggestion
		<createIndex indexName="idx_obs_archive_date_archived" tableName="obs_archive">
			<column name="date_archived"/>
		</createIndex>
		<createIndex indexName="idx_obs_archive_obs_datetime" tableName="obs_archive">
			<column name="obs_datetime"/>
		</createIndex>
```

Muta-Jonathan (changes requested):

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: okay, thanks

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: I see, thanks

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: got it!

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: Right now the archive table is not being queried via date_archived and obs_datetime ([TRUNK-6649](https://github.com/openmrs/openmrs-core/pull/6171))

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: Youre right. Will remove that.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: Yes, youre right. Will add this. 
But is it better to split both the tables into two liquibase changesets instead?

dkayiwa:
@claude review

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: oops my bad. changed it. copy paste error.

shivvani-r:
@dkayiwa @Muta-Jonathan Could you guys please review this again

dkayiwa:
@claude review

Muta-Jonathan:
@dkayiwa @shivvani-r am I the only one not seeing what @claude reviews after @dkayiwa requests it to do so?

shivvani-r:
> @dkayiwa @shivvani-r am I the only one not seeing what @claude reviews after @dkayiwa requests it to do so?

I think the Github Actions comment is in response to the @ claude prompt

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: 🥲I got confused by the first comment, sorry! I initially added them because ObsArchive inherits from BaseOpenmrsData. Will revert

shivvani-r:
@dkayiwa could you please re-review?

dkayiwa:
@claude review

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,111 @@
...
+	<preConditions onFail="MARK_RAN">
+			<not>
+				<tableExists tableName="obs_reference_range_archive"/>
+			</not>
+	</preConditions>
+	<comment>Add obs_reference_range_archive table for Archiving Observations</comment>
+	<createTable tableName="obs_reference_range_archive">
+		<column name="obs_reference_range_id" type="INT">
+			<constraints nullable="false" primaryKey="true"/>
+		</column>
```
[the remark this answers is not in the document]
shivvani-r: No, leaving out` autoIncrement="true"` is the correct call because the archive table needs to strictly preserve the exact original ID rather than generating a new one.

To fix this, we needed to:

1. Remove the `@MapsId `on `obsArchive` in `ObsArchiveReferenceRange`, as it was causing the reference_range_id to be overwritten with the obs_id on flush.
2. Add `property-ref="obs"` to `obs.hbm.xml` (`<one-to-one name="referenceRange" class="org.openmrs.ObsReferenceRange" cascade="all"/>`) to correct the mapping.
3. Update the native INSERT query in ObsArchiveHelper to explicitly include and select obs_reference_range_id so the exact ID is preserved on the way back during a restore.

I've made all of these changes over in TRUNK-6649 ([#6171](https://github.com/openmrs/openmrs-core/pull/6171)), and the archive/unarchive on obs_reference_range is now fully working without any issues or data loss.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,111 @@
...
+			<column name="form_namespace_and_path" type="VARCHAR(255)"/>
+			<column defaultValue="FINAL" name="status" type="VARCHAR(16)">
+				<constraints nullable="false"/>
+			</column>
+			<column name="interpretation" type="VARCHAR(32)"/>
+			<column name="changed_by" type="INT"/>
+			<column name="date_changed" type="DATETIME"/>
+			<column name="archived_by" type="INT"/>
+			<column name="date_archived" type="DATETIME"/>
+		</createTable>
```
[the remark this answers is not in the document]
shivvani-r: ive added all the foreign key constraints except for the obs_group_id and previous_version to obs_archive and obs_reference_range_archive and i tested the archive/unarchive flow at [TRUNK-6649](https://github.com/openmrs/openmrs-core/pull/6171) and it is working well without any issues. so ill add them here too. thanks

shivvani-r:
@dkayiwa @Muta-Jonathan Could you guys please review this?

dkayiwa:
@claude review

shivvani-r:
@dkayiwa I cannot see the review comments left by claude

shivvani-r:
@dkayiwa @Muta-Jonathan I checked and there is no need to add any indexes for the obs_reference_archive. Could you please re-review this when you guys are free? I think the PR is ready.

dkayiwa:
@shivvani-r can you fill the ticket description?

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -18,69 +18,73 @@
 
 
 	<changeSet id="20200604-soundex_extension" author="aman" dbms="postgresql">
-		<comment> Soundex extension for PostgreSQL</comment>
+		<comment>Soundex extension for PostgreSQL</comment>
 		<sql> CREATE EXTENSION IF NOT EXISTS fuzzystrmatch SCHEMA public;</sql>
 	</changeSet>
 
 	<changeSet id="20200715-uuid_ossp_extension" author="aman" dbms="postgresql">
-		<comment> Extension to use UUID functions with PostgreSQL and creating an alias similar to MySQL</comment>
```
dkayiwa: Are all these changes related to the ticket?

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -18,69 +18,73 @@
 
 
 	<changeSet id="20200604-soundex_extension" author="aman" dbms="postgresql">
-		<comment> Soundex extension for PostgreSQL</comment>
+		<comment>Soundex extension for PostgreSQL</comment>
```
dkayiwa: Could you revert the reformatting and keep this diff to the two new changesets? It rewrites eight of the thirteen changesets already in the file (both postgresql ones, TRUNK-5836, TRUNK-6391, TRUNK-5749, TRUNK-5809, TRUNK-5890 and TRUNK-6245), which reads like an IDE reformat on save: `type="DATETIME" />` closed up to `type="DATETIME"/>`, the leading space dropped inside `<comment>`, and the `CREATE FUNCTION UUID()` body split across five lines.

Nothing is broken by it. I computed the Liquibase 4.32.0 checksums for every changeset in the file before and after, under both the V8 and the current checksum version, and all thirteen come out unchanged, so no existing 3.0.0-SNAPSHOT database will fail validation on the next startup. The cost is review time instead. An update file that only ever grows lets a reviewer see at a glance that nothing already shipped was touched, and once old changesets get rewritten somebody has to prove that separately, here and on every later PR that reformats them again.

It comes to 19 added and 19 removed lines, about a fifth of the diff. Resetting the file to master and re-applying the two new changesets at the end is probably the quickest way back, and it is worth turning format-on-save off for the liquibase files.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: Sorry mvn spotless made those changes. Reverted them.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: Reverted.

shivvani-r:
> @shivvani-r can you fill the ticket description?

Done!

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,150 @@
...
+		<column name="hi_normal" type="DOUBLE"/>
+		<column name="low_absolute" type="DOUBLE"/>
+		<column name="low_critical" type="DOUBLE"/>
+		<column name="low_normal" type="DOUBLE"/>
+		<column name="uuid" type="CHAR(38)">
+			<constraints nullable="false" unique="true"/>
+		</column>
+		<column name="archived_by" type="INT"/>
+		<column name="date_archived" type="DATETIME"/>
+	</createTable>
```
dkayiwa: Lines 266 to 285 sit one tab shallower than the rest of the changeset, so `<comment>` and `<createTable>` line up with `<changeSet>` itself rather than with the `<preConditions>` above them and the two `<addForeignKeyConstraint>` calls below. Nothing depends on it, but this file cannot be edited once 3.0.0 is released, so it is easier to take now than to leave for whoever reads the changeset next.

```suggestion
		<comment>Add obs_reference_range_archive table for Archiving Observations</comment>
		<createTable tableName="obs_reference_range_archive">
			<column name="obs_reference_range_id" type="INT">
				<constraints nullable="false" primaryKey="true"/>
			</column>
			<column name="obs_id" type="INT">
				<constraints nullable="false" unique="true"/>
			</column>
			<column name="hi_absolute" type="DOUBLE"/>
			<column name="hi_critical" type="DOUBLE"/>
			<column name="hi_normal" type="DOUBLE"/>
			<column name="low_absolute" type="DOUBLE"/>
			<column name="low_critical" type="DOUBLE"/>
			<column name="low_normal" type="DOUBLE"/>
			<column name="uuid" type="CHAR(38)">
				<constraints nullable="false" unique="true"/>
			</column>
			<column name="archived_by" type="INT"/>
			<column name="date_archived" type="DATETIME"/>
		</createTable>
```

shivvani-r: Thanks, fixed it

dkayiwa: One line kept the old depth: line 269, the `<constraints nullable="false" primaryKey="true"/>` under `obs_reference_range_id`, is five tabs in where the identical line under `obs_id` three rows below is four.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,150 @@
...
+				<constraints nullable="false"/>
+			</column>
+			<column name="interpretation" type="VARCHAR(32)"/>
+			<column name="changed_by" type="INT"/>
+			<column name="date_changed" type="DATETIME"/>
+			<column name="archived_by" type="INT"/>
+			<column name="date_archived" type="DATETIME"/>
+		</createTable>
+		
+		<createIndex indexName="idx_obs_archive_obs_group_id" tableName="obs_archive">
```
dkayiwa, line 207: Do we have each of the corresponding index on the obs table?

shivvani-r: 4 out of the 5 indexes here have corresponding indexes on the obs table (obs_group_id, encounter_id, person_id, concept_id).

The obs table does not have an index on date_voided, but we are adding it to obs_archive here because it is necessary for querying and purging archived data based on when it was voided.

I did not add the other indexes that are present on the obs table (specifically the indexes on value_coded, value_drug, obs_datetime, creator, location_id, value_coded_name_id, order_id, previous_version, and voided_by) because right now the obs_archive is not being queried via them.

Do you think we are missing any critical lookup paths for archived data that would warrant adding more of those obs indexes?

shivvani-r: @dkayiwa please check my reply above

dkayiwa: Seven of those nine do end up indexed, because the foreign keys bring the index with them. InnoDB requires an index on the referencing column of every foreign key and creates one named after the constraint when none is declared, so `value_coded`, `value_drug`, `creator`, `location_id`, `value_coded_name_id`, `order_id` and `voided_by` all get one from the `addForeignKeyConstraint` calls above. I built this table on MySQL 8.0.34 from exactly this DDL and it comes out with 16 indexes: the primary key, the uuid unique constraint, the five declared here, and nine auto-created (`archived_by`, `changed_by`, `creator`, `location_id`, `order_id`, `value_coded`, `value_coded_name_id`, `value_drug`, `voided_by`).

That makes the comparison simpler than it looks. All 15 indexes on `obs` account for exactly its 12 foreign keys, the primary key, the uuid unique constraint, and `obs_datetime_idx`, so `obs_datetime_idx` is the only index on `obs` that is not a constraint artifact. The two tables line up column for column apart from `obs_datetime` and `previous_version`, and the `previous_version` index on `obs` only exists because of the self-referencing foreign key that this table deliberately does not carry.

On lookup paths I do not think anything is missing. Every predicate #6171 uses against `obs_archive` (`obsGroupId`, `encounter`, `person`, `person` plus `concept`, `uuid`, `obs_id`, `obsGroupId` plus `dateVoided`) has an index behind it. I did check whether the standalone `date_voided` index was redundant next to the `obs_group_id` one and it is not: on 200k rows MySQL intersects the two for that predicate pair, and a composite `(obs_group_id, date_voided)` would turn a `date_voided`-only purge scan into a full index scan, so leaving them as two is right.

shivvani-r: So I'll leave the indexes as they are currently defined.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
shivvani-r: fixed that

api/src/test/java/org/openmrs/util/DatabaseUpdaterDatabaseIT.java
```diff
@@ -32,7 +32,7 @@ public class DatabaseUpdaterDatabaseIT extends DatabaseIT {
 	 * This constant needs to be updated when adding new Liquibase update files to openmrs-core.
 	 */
 
-	private static final int CHANGE_SET_COUNT_FOR_GREATER_THAN_2_1_X = 916;
+	private static final int CHANGE_SET_COUNT_FOR_GREATER_THAN_2_1_X = 918;
```
dkayiwa, line 35: Counting changesets is all this asserts, so nothing in the tree ties `obs_archive` back to `obs`. The archive is a column-for-column copy (the 30 columns of `obs`, plus `changed_by`, `date_changed`, `archived_by` and `date_archived`), and #6171 restores with a literal `INSERT INTO obs (obs_id, person_id, ...) SELECT obs_id, person_id, ... FROM obs_archive` naming all 30. Whoever adds the next column to `obs` therefore has to remember three more places on their own (this table, `ObsArchive`, and that column list), and if they miss one, archived observations come back with the new column silently empty.

Since 3.0.x is still taking changesets, a test seems a better home for that rule than the code comment #6171 leaves on the restore query. This passes on your branch as `api/src/test/java/org/openmrs/util/ObsArchiveSchemaDatabaseIT.java`:

```java
public class ObsArchiveSchemaDatabaseIT extends DatabaseIT {
	
	@Test
	public void archiveTablesShouldCarryEveryColumnOfTheirSourceTable() throws Exception {
		ChangeLogVersionFinder finder = new ChangeLogVersionFinder();
		updateDatabase(finder.getChangeLogCombinations().get(finder.getLatestSnapshotVersion().orElseThrow()));
		
		assertMirrors("OBS", "OBS_ARCHIVE", "CHANGED_BY", "DATE_CHANGED", "ARCHIVED_BY", "DATE_ARCHIVED");
		assertMirrors("OBS_REFERENCE_RANGE", "OBS_REFERENCE_RANGE_ARCHIVE", "ARCHIVED_BY", "DATE_ARCHIVED");
	}
	
	private void assertMirrors(String source, String archive, String... archiveOnlyColumns) throws Exception {
		Set<String> expected = columnsOf(source);
		expected.addAll(Set.of(archiveOnlyColumns));
		assertEquals(expected, columnsOf(archive),
		    archive + " must carry every column of " + source + " plus " + Set.of(archiveOnlyColumns));
	}
	
	private Set<String> columnsOf(String table) throws Exception {
		Set<String> columns = new TreeSet<>();
		try (Connection connection = getConnection(); PreparedStatement statement = connection
		        .prepareStatement("SELECT column_name FROM information_schema.columns WHERE UPPER(table_name) = ?")) {
			statement.setString(1, table);
			try (ResultSet resultSet = statement.executeQuery()) {
				while (resultSet.next()) {
					columns.add(resultSet.getString(1).toUpperCase());
				}
			}
		}
		assertEquals(false, columns.isEmpty(), "table " + table + " was not created");
		return columns;
	}
}
```

Appending a throwaway `<addColumn tableName="obs">` changeset to the 3.0.x file turns it red with `OBS_ARCHIVE must carry every column of OBS plus [...] SOME_NEW_OBS_COLUMN`, which is the nudge the next person needs. It does not have to hold up the merge, though the schema PR is where I think the rule belongs.

shivvani-r: Makes sense, added the test.

shivvani-r:
@dkayiwa could you please review this?

api/src/test/java/org/openmrs/util/ObsArchiveSchemaDatabaseIT.java
```diff
@@ -0,0 +1,63 @@
...
+	@BeforeEach
+	@Override
+	public void setup() throws Exception {
+		DatabaseIT.CONNECTION_URL = "jdbc:h2:mem:obs_archive_test;DB_CLOSE_DELAY=-1;MODE=LEGACY;NON_KEYWORDS=VALUE";
+		super.setup();
+	}
+
+	@Test
+	public void shouldCarryEveryColumnOfTheirSourceTable() throws Exception {
+		ChangeLogVersionFinder finder = new ChangeLogVersionFinder();
```
dkayiwa: Worth swapping this for the pattern the other two DatabaseIT classes use, though it does not block the merge. `ValidateHibernateMappingsDatabaseIT` opens its test with `this.dropAllDatabaseObjects()` under the comment "Drop all database objects before running the test as previously run tests may have left tables behind", and `EnversAuditTableInitializerDatabaseIT` does the same in its `@BeforeEach`. That gets the same protection without touching the shared connection URL:

```suggestion
	@Test
	public void shouldCarryEveryColumnOfTheirSourceTable() throws Exception {
		/*
		 * Drop all database objects before running the test as previously run tests may have left tables behind.
		 */
		this.dropAllDatabaseObjects();

		ChangeLogVersionFinder finder = new ChangeLogVersionFinder();
```

I am not suggesting you just delete the override, since the protection is genuinely needed. With it removed the test dies on `Table "ALLERGY" already exists` at the 2.9.x snapshot: `integration-test-2` runs every `*DatabaseIT` in one surefire fork, `DatabaseUpdaterDatabaseIT` overrides `tearDown` without dropping, and in the last CI run it executed immediately before this class. The drop-first version passes under `-Dsurefire.runOrder=alphabetical` and under `reversealphabetical`, so it holds whichever way the two land. Spotless takes the now unused `BeforeEach` import out for you on the next build.

Two things the URL swap costs that the drop does not. The literal repeats `DB_CLOSE_DELAY=-1;MODE=LEGACY;NON_KEYWORDS=VALUE` from `DatabaseIT.CONNECTION_URL` with only the database name changed, and nothing ties the copy to the original, so whoever next changes the H2 settings on the base class leaves this one test running on the old ones. And `CONNECTION_URL` is a `public static` field that never gets restored, so for the rest of the fork every `DatabaseIT` connects to `obs_archive_test` instead of `test`. Neither bites today, since this class runs last in the api module and the others drop first, but both go away with the drop.

shivvani-r: I've applied the suggestion to use this.dropAllDatabaseObjects() inside the test instead

shivvani-r:
@dkayiwa could you please review this?

shivvani-r:
@dkayiwa could you please review this?

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,150 @@
...
+			<column name="uuid" type="CHAR(38)">
+				<constraints nullable="false" unique="true"/>
+			</column>
+			<column name="previous_version" type="INT"/>
+			<column name="form_namespace_and_path" type="VARCHAR(255)"/>
+			<column defaultValue="FINAL" name="status" type="VARCHAR(16)">
+				<constraints nullable="false"/>
+			</column>
+			<column name="interpretation" type="VARCHAR(32)"/>
+			<column name="changed_by" type="INT"/>
```
dkayiwa, line 201: What is changed_by and date_changed used for?

shivvani-r: because ObsArchive is an annotated JPA @Entity that inherits from BaseOpenmrsData, Hibernate's schema validation requires these columns to exist in the obs_archive table. The original obs table does not have them because it is mapped via hbm.xml.

PR #6431, opened 2026-08-04
TRUNK-6717: Replace reflective BeanUtils property access with direct PersonName getters

RaniduNethma:
TRUNK-6717: Replace reflective BeanUtils property access with direct PersonName getters

Replaced the reflective `BeanUtils.getProperty` call in `NameTemplate.format` with direct `PersonName` getters via a `switch` statement for the 8 standard tokens. This avoids the synchronized `ContextClassLoaderLocal` lock, significantly improving name rendering performance. Retained the reflective fallback for custom tokens to preserve existing behavior. Also added a new test in `NameTemplateTest` to verify that all standard tokens resolve correctly.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6717

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/layout/name/NameTemplate.java
+++ b/api/src/main/java/org/openmrs/layout/name/NameTemplate.java
@@ -75,26 +75,17 @@ public class NameTemplate extends LayoutTemplate {
 	}
 
 	private static String resolveToken(PersonName pn, String codeName) throws Exception {
-		switch (codeName) {
-			case "prefix":
-				return pn.getPrefix();
-			case "givenName":
-				return pn.getGivenName();
-			case "middleName":
-				return pn.getMiddleName();
-			case "familyNamePrefix":
-				return pn.getFamilyNamePrefix();
-			case "familyName":
-				return pn.getFamilyName();
-			case "familyName2":
-				return pn.getFamilyName2();
-			case "familyNameSuffix":
-				return pn.getFamilyNameSuffix();
-			case "degree":
-				return pn.getDegree();
-			default:
-				// custom-configured token: keep old behavior as a rare fallback
-				return BeanUtils.getProperty(pn, codeName);
-		}
+		return switch (codeName) {
+			case "prefix" -> pn.getPrefix();
+			case "givenName" -> pn.getGivenName();
+			case "middleName" -> pn.getMiddleName();
+			case "familyNamePrefix" -> pn.getFamilyNamePrefix();
+			case "familyName" -> pn.getFamilyName();
+			case "familyName2" -> pn.getFamilyName2();
+			case "familyNameSuffix" -> pn.getFamilyNameSuffix();
+			case "degree" -> pn.getDegree();
+			// custom-configured token: keep old behavior as a rare fallback
+			default -> BeanUtils.getProperty(pn, codeName);
+		};
 	}
 }
--- a/api/src/main/java/org/openmrs/layout/name/NameTemplate.java
+++ b/api/src/main/java/org/openmrs/layout/name/NameTemplate.java
@@ -47,7 +47,7 @@ public class NameTemplate extends LayoutTemplate {
 				boolean hasToken = false;
 				for (Map<String, String> lineToken : line) {
 					if (lineToken.get("isToken").equals(layoutToken)) {
-						String tokenValue = BeanUtils.getProperty(personName, lineToken.get("codeName"));
+						String tokenValue = resolveToken(personName, lineToken.get("codeName"));
 						if (StringUtils.isNotBlank(tokenValue)) {
 							hasToken = true;
 							nameLine.append(tokenValue);
@@ -74,4 +74,27 @@ public class NameTemplate extends LayoutTemplate {
 		return NameSupport.getInstance();
 	}
 
+	private static String resolveToken(PersonName pn, String codeName) throws Exception {
+		switch (codeName) {
+			case "prefix":
+				return pn.getPrefix();
+			case "givenName":
+				return pn.getGivenName();
+			case "middleName":
+				return pn.getMiddleName();
+			case "familyNamePrefix":
+				return pn.getFamilyNamePrefix();
+			case "familyName":
+				return pn.getFamilyName();
+			case "familyName2":
+				return pn.getFamilyName2();
+			case "familyNameSuffix":
+				return pn.getFamilyNameSuffix();
+			case "degree":
+				return pn.getDegree();
+			default:
+				// custom-configured token: keep old behavior as a rare fallback
+				return BeanUtils.getProperty(pn, codeName);
+		}
+	}
 }
--- a/api/src/test/java/org/openmrs/layout/name/NameTemplateTest.java
+++ b/api/src/test/java/org/openmrs/layout/name/NameTemplateTest.java
@@ -126,4 +126,54 @@ public class NameTemplateTest extends BaseContextSensitiveTest {
 		assertEquals("Moses Mujuzi", nameTemplate.format(personName));
 	}
 
+	@Test
+	public void shouldResolveAllStandardTokensDirectly() {
+
+		NameTemplate nameTemplate = new NameTemplate();
+
+		List<String> lineByLineFormat = new ArrayList<>();
+		lineByLineFormat.add("prefix");
+		lineByLineFormat.add("familyNamePrefix");
+		lineByLineFormat.add("familyName2");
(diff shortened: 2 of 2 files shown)


jwnasambu:
@claude review

api/src/test/java/org/openmrs/layout/name/NameTemplateTest.java
```diff
@@ -126,4 +126,54 @@ public void shouldUseNameTemplateConfiguredViaGlobalProperties() {
 		assertEquals("Moses Mujuzi", nameTemplate.format(personName));
 	}
 
+	@Test
+	public void shouldResolveAllStandardTokensDirectly() {
```
jwnasambu, line 130: This new test is a good change and the switch matches the getters BeanUtils would call exactly. One gap: the reflective fallback you deliberately kept for custom tokens (NameTemplate.java:95-98) has no test, and since that fallback returns the same value as the direct getters for any valid PersonName property, this test can't tell a standard token resolved through the switch from one resolved through reflection. A small template with a custom token, e.g. fullName, would lock the fallback behavior the ticket explicitly promises to preserve. Optional, happy to ignore.

This is an optional snippet to include as the test:

```suggestion
	@Test
public void shouldResolveCustomTokenViaFallback() {
    NameTemplate nameTemplate = new NameTemplate();
    nameTemplate.setLineByLineFormat(Collections.singletonList("fullName"));
    Map<String, String> nameMappings = new HashMap<>();
    nameMappings.put("fullName", "fullName");
    nameTemplate.setNameMappings(nameMappings);
    nameSupport.setLayoutTemplates(Collections.singletonList(nameTemplate));

    PersonName personName = new PersonName();
    personName.setGivenName("Ranidu");
    personName.setFamilyName("Rathnayaka");

    assertEquals("Ranidu Rathnayaka", nameTemplate.format(personName));
}
```

jwnasambu:
@ibacher, @dkayiwa, @wikumChamith Kindly help review this PR at your convenient time please!

api/src/test/java/org/openmrs/layout/name/NameTemplateTest.java
RaniduNethma: Thanks for the suggestion. I implemented the fallback test using `uuid` as a reliable custom token. I noticed that for the template to recognize a custom token and not ignore it as a non token, it needs to be temporarily added to `nameSupport.setSpecialTokens()` during the test, alongside the `sizeMappings` . I've added those, and the test now successfully verifies the fallback behavior.

api/src/main/java/org/openmrs/layout/name/NameTemplate.java
```diff
@@ -47,7 +47,7 @@ public String format(PersonName personName) {
 				boolean hasToken = false;
 				for (Map<String, String> lineToken : line) {
 					if (lineToken.get("isToken").equals(layoutToken)) {
-						String tokenValue = BeanUtils.getProperty(personName, lineToken.get("codeName"));
+						String tokenValue = resolveToken(personName, lineToken.get("codeName"));
```
dkayiwa, line 50: Nothing to change here, but the numbers are worth having before TRUNK-6717 gets closed as done. On the legacyui `short` template (three single-token lines) the three reflective lookups this replaces cost about 430ns per `format()` call, while the `getLines()` call up at line 41 costs about 7.3µs, because it re-tokenizes every line on every call: 8 fresh regex compiles per line from `replaceTokens`, plus one more for the split. So this takes out roughly 6% of the method's single-threaded cost.

The ticket's contention argument does hold up, though. At 8 threads on a 10-core box the reflective lookups degrade 7.7x (430ns to 3.3µs) while `getLines()` only degrades 2.7x, so dropping the synchronized `ContextClassLoaderLocal` is worth doing on its own. I'd merge this as it stands and open a follow-up for caching `getLines()`, which looks tractable given that a `layout.name.template` change builds a fresh `NameTemplate` rather than mutating the existing one.

RaniduNethma: Thanks @dkayiwa for the detailed performance analysis. After this opening a follow-up for caching `getLines()` sounds like a perfect plan. Thanks again for the review and guidance.

ibacher (approved):

api/src/main/java/org/openmrs/layout/name/NameTemplate.java
```diff
@@ -74,4 +74,27 @@ public LayoutSupport<?> getLayoutSupportInstance() {
...
+			case "familyName2":
+				return pn.getFamilyName2();
+			case "familyNameSuffix":
+				return pn.getFamilyNameSuffix();
+			case "degree":
+				return pn.getDegree();
+			default:
+				// custom-configured token: keep old behavior as a rare fallback
+				return BeanUtils.getProperty(pn, codeName);
+		}
```
wikumChamith, line 98: Feels like it would be better if we used a Map for this rather than a long switch-case :)

If this is only for platform 3, we can just use:

```java
private static final Map<String, Function<PersonName, String>> STANDARD_TOKENS = Map.of(
        "prefix", PersonName::getPrefix,
        "givenName", PersonName::getGivenName,
        "middleName", PersonName::getMiddleName,
        "familyNamePrefix", PersonName::getFamilyNamePrefix,
        "familyName", PersonName::getFamilyName,
        "familyName2", PersonName::getFamilyName2,
        "familyNameSuffix", PersonName::getFamilyNameSuffix,
        "degree", PersonName::getDegree);
```

cc: @ibacher what do you think?

ibacher: I think that switch statement is more idiomatic for this. Especially because:

1. Special casing with that construction ruins the simplicity of `STANDARD_TOKENS.get(codeName).apply(pn)` (`STANDARD_TOKENS.getOrDefault(codeName, (pn) -> BeanUtils.getProperty(pn, codeName).apply(pn)`). 
2. `STANDARD_TOKENS.get(codeName).apply(pn)` is already kinda ugly, making it harder to "read" what this code is doing.
3. It's actually way less efficient.

That said, I'm open to implementing this with the newer switch expression syntax in `master`:

```java
return switch (codeName) {
    case "prefix" -> pn.getPrefix();
    case "givenName" -> pn.getGivenName();
    case "middleName" -> pn.getMiddleName();
    case "familyNamePrefix" -> pn.getFamilyNamePrefix();
    case "familyName" -> pn.getFamilyName();
    case "familyName2" -> pn.getFamilyName2();
    case "familyNameSuffix" -> pn.getFamilyNameSuffix();
    case "degree" -> pn.getDegree();
    // custom-configured token: keep old behavior as a rare fallback
    default -> BeanUtils.getProperty(pn, codeName);
};
```

Which I only didn't suggest so we can backport it as-is.

RaniduNethma: @wikumChamith Thanks for the suggestion. 
@ibacher I have updated the implementation to use the newer Java switch expression syntax as suggested. The code looks much cleaner now. Thanks for the guidance.

api/src/main/java/org/openmrs/layout/name/NameTemplate.java
ibacher: Tbf, this ticket was one of several that came out of a performance audit of searching for thousands of patients, at which scale, the lock contention matters a lot more than the single-thread performance.

api/src/test/java/org/openmrs/layout/name/NameTemplateTest.java
```diff
@@ -126,4 +126,81 @@ public void shouldUseNameTemplateConfiguredViaGlobalProperties() {
...
+
```
ibacher, line 131:
```suggestion
```

api/src/test/java/org/openmrs/layout/name/NameTemplateTest.java
```diff
@@ -126,4 +126,81 @@ public void shouldUseNameTemplateConfiguredViaGlobalProperties() {
...
+		personName.setDegree("BSc");
+
+		String formattedName = nameTemplate.format(personName);
+
+		assertEquals("Mr. Ganege Ralalage Ranidu Nethma Rathnayaka (Contributor) BSc", formattedName);
+	}
+
+	@Test
+	public void shouldResolveCustomTokenViaFallback() {
+
```
ibacher, line 181:
```suggestion
```

ibacher (approved):

PR #6432, opened 2026-08-04
TRUNK-6382/TRUNK-6360: Backport @Transactional/@Cacheable-on-classes-without-interface fix to 2.6.x

binduak:
## Summary

Backports the fix for **`@Transactional`/`@Cacheable` not working on classes without an interface** (TRUNK-6382, TRUNK-6360) from Core 2.8.0 onto the `2.6.x` maintenance branch.

## Problem

Upgrading the FHIR2 module (2.6.0+) on Platform 2.6.x causes OpenMRS to fail to start with:

```
BeanCreationException: Error creating bean with name 'appFrameworkService' ...
Cannot convert value of type 'com.sun.proxy.$ProxyNNN implementing
org.springframework.aop.SpringProxy, org.springframework.aop.framework.Advised,
org.springframework.cglib.proxy.Factory, org.springframework.core.DecoratingProxy'
to required type 'org.openmrs.module.appframework.repository.AllComponentsState':
no matching editors or conversion strategy found
```

This isn't unique to the App Framework module — we reproduced the same failure pattern against `dataexchange`'s `DataExporter` bean, and it's the same root cause reported for `stockmanagement` in TRUNK-6382.

Reported by our team here: [FHIR2 OMOD Upgrade Fails on OpenMRS Platform 2.6.17 with AllComponentsState Proxy Conversion Error](https://talk.openmrs.org/t/fhir2-omod-upgrade-2-5-0-2-6-0-fails-on-openmrs-platform-2-6-17-with-allcomponentsstate-proxy-conversion-error/49841).

## Root cause

Several modules (App Framework's `AllComponentsState`, `dataexchange`'s `DataExporter`, etc.) declare `@Transactional` on a concrete class with no interface. Before this fix, `@Transactional`/`@Cacheable` annotation processing silently did not apply to any bean that wasn't an `OpenmrsService`, so these beans were never actually proxied — the raw object was injected directly, and everything worked by accident.

Something in FHIR2 2.6.0+'s own bean graph correctly triggers Spring's transactional annotation-driven infrastructure to *actually* attempt to proxy these beans for the first time. Since they have no interface, Spring's default JDK-dynamic-proxy strategy can't produce a valid proxy assignable to the concrete class — hence the `IllegalStateException`.

This exact bug and fix already exist for Core 2.8.0+ (TRUNK-6382, TRUNK-6360), which switches `@Transactional`/`@Cacheable` proxying to CGLIB (class-based) via `<tx:annotation-driven proxy-target-class="true"/>` / `<cache:annotation-driven proxy-target-class="true"/>`, replacing the old `DefaultAdvisorAutoProxyCreator`/`TransactionAttributeSourceAdvisor` pair. CGLIB proxies subclass the real class, so they satisfy `instanceof AllComponentsState` correctly.

**This fix was never backported to 2.6.x or 2.7.x**, only released in Core 2.8.0 (2025-08-15). Implementers who can't yet take the 2.6→2.8 platform jump have no way to consume newer FHIR2/reporting/etc. releases without hitting this.

## What this PR does

Cherry-picks the three commits that implement and refine the fix on the 2.8.x branch:

- `7a42063` — TRUNK-6360: Cacheable needs to work not only on OpenmrsService methods but all beans
- `ec1f42c` — TRUNK-6382: @Transactional must work on classes without interface
- `77de212` — TRUNK-6382: Use Autowired for cacheConfig for backwards compatibility

Plus one small addition needed specifically for 2.6.x: this branch configures caches via `ehcache-api.xml` rather than the `cache-api.yaml` resource master/2.8.x use, so a `testCache` entry was added there to keep the accompanying unit tests passing.

## Validation

We independently reproduced and validated this against a real reference-application-distro Docker environment (not just unit tests):

1. Confirmed the failure using stock `appframework` + FHIR2 2.9.1 on an unmodified Platform 2.6.x-based core — reproduces the reported bug exactly.
2. Added a diagnostic `BeanFactoryPostProcessor` to confirm the actual `AutoProxyCreator`/`Advisor` pair responsible lives in `applicationContext-service.xml` (`DefaultAdvisorAutoProxyCreator` + `TransactionAttributeSourceAdvisor`), not in FHIR2 itself — ruling out a FHIR2-side fix.
3. Rebuilt core with these 3 commits cherry-picked onto `2.6.x`, deployed it alongside **completely stock, unpatched** `appframework` and `fhir2` modules, and confirmed the proxy-conversion failure is fully gone across multiple independent clean-container boots (including a fully-stopped-then-restarted container to rule out classloader caching artifacts).
4. Also confirmed on a **plain, unmodified 2.6.x** build (no cherry-picks) that an unrelated bean-ordering flakiness exists independently of this fix, in this specific test environment — so it's not something this PR introduces.

## Known follow-up

`OpenmrsCacheManagerFactoryBeanTest` has one flaky assertion around Spring `CacheManager`/EhCache bean resolution timing on this branch's older cache-config style, distinct from the core fix. Flagging transparently — happy to iterate based on maintainer feedback rather than delay this proposal.

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
@@ -28,6 +28,7 @@ import org.hibernate.boot.Metadata;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.integrator.spi.Integrator;
 import org.hibernate.service.spi.SessionFactoryServiceRegistry;
+import org.openmrs.api.cache.CacheConfig;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.Module;
 import org.openmrs.module.ModuleFactory;
@@ -59,6 +60,12 @@ public class HibernateSessionFactoryBean extends LocalSessionFactoryBean impleme
 	
 	private Metadata metadata;
 	
+	private CacheConfig cacheConfig;
+
+	public void setCacheConfig(CacheConfig cacheConfig) {
+		this.cacheConfig = cacheConfig;
+	}
+
 	/**
 	 * Collect the mapping resources for future use because the mappingResources object is defined
 	 * as 'private' instead of 'protected'
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
@@ -60,11 +60,8 @@ public class HibernateSessionFactoryBean extends LocalSessionFactoryBean impleme
 	
 	private Metadata metadata;
 	
+	@Autowired
 	private CacheConfig cacheConfig;
-
-	public void setCacheConfig(CacheConfig cacheConfig) {
-		this.cacheConfig = cacheConfig;
-	}
 
 	/**
 	 * Collect the mapping resources for future use because the mappingResources object is defined
--- a/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
@@ -9,6 +9,19 @@
  */
 package org.openmrs.api.impl;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Date;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Properties;
+import java.util.Set;
+import java.util.stream.Collectors;
+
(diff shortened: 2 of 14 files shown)


jwnasambu (commented):
@binduak Kindly note that each ticket should have its own pull request. We don’t combine multiple tickets into a single PR, so TRUNK-6382 and TRUNK-6360 should be submitted separately. Also, from the ticket links, it looks like @rkorytkowski is already working on these tickets. It would be best to avoid duplicating work and instead coordinate on the existing efforts before opening a PR. If you’ve identified any additional issue that needs to be fixed, kindly create a separate ticket for it and submit it as an independent pull request.

ibacher:
@binduak Thanks for this. Unfortunately since the tests don't pass and this seems to make a lot of unrelated reformatting, this doesn't seem mergeable in  this state. Is this still something you're available to work on?

PR #6434, opened 2026-08-05
TRUNK-6702: Make UpdateFilter and InitializationFilter skip checks cheap after startup

jwnasambu:

This PR optimizes InitializationFilter and UpdateFilter request handling by removing unnecessary work from the normal application runtime path. Both filters are mapped to /*, meaning they execute on every request. Previously, every request: Parsed request parameters early in the filter chain (including possible query strings/form bodies) to check the wizard state. Acquired locks through synchronized getters when reading completion flags.Since these checks are only needed while installation/update wizards are active, this caused avoidable overhead during normal server operation.

## Issue I worked on

see https://openmrs.atlassian.net/issues?filter=10641&selectedIssue=TRUNK-6702

  No? -> execute above command


jwnasambu:
@claude review

jwnasambu:
@ibacher , @wikumChamith , @dkayiwa kindly help me review this PR at your convenient time please.

web/src/main/java/org/openmrs/web/filter/update/UpdateFilter.java
```diff
@@ -527,7 +552,23 @@ protected Object getUpdateFilterModel() {
...
+		}
+		if (servingUpdateProgress) {
+			// The updates just finished. The progress page may still be polling for the final
+			// "updatesRequired = false" response so it can leave the wizard. Keep serving those
+			// requests until the response has been delivered and the page has moved on to the
+			// application (seen as the next non-AJAX request).
+			if (PROGRESS_VM_AJAXREQUEST.equals(httpRequest.getParameter("page"))) {
+				return false;
+			}
+			if (updateProgressResponseDelivered) {
```
dkayiwa: This needs fixing before merge. Closing the window on "the next non-AJAX request" makes one browser's progress page depend on no other client touching the server, and by this point OpenMRS is started and serving, so other clients very likely do.

The flow that breaks is a database upgrade that produces liquibase warnings. `reviewchanges.vm` deliberately stays on screen there (`if (isDisplayingWarnings == true) { return; }`) and keeps polling every second. I traced it against the real filter: the first poll after completion sets `updateProgressResponseDelivered = true`, any unrelated non-AJAX request then flips `servingUpdateProgress` to false, and the next poll from that still-open page gets `skipFilter == true`, falls through to the application, and fails jQuery's json parse. Five of those and `reviewchanges.vm` runs `window.location = "index.htm"`. The admin loses the warnings and the update-log path, and cannot get them back, because `doPost` clears `updateWarnings` right after writing them into the response, so they are delivered exactly once. Nothing is logged, so it reads as an ordinary redirect.

My suggestion is to drop `updateProgressResponseDelivered` and the close-on-next-request rule, and keep serving the progress AJAX request for as long as the JVM lives once a wizard job has run. You still get the win you are after on every normal boot (nothing to install, no updates pending, both flags false), which is the case that actually carries the traffic, and a server that just ran the update wizard paying one `getParameter` per request until its next restart seems a fair price for not having a cross-client race here. If you would rather keep the window closing, the other route is to make `reviewchanges.vm` stop polling once it has its final answer, so that nothing depends on the window being open.

The same block is in `InitializationFilter.skipFilter`, so whichever way you go it wants the same change there.

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -1097,10 +1122,23 @@ protected Object getUpdateFilterModel() {
...
+		}
+		if (servingInitializationProgress) {
+			// The installation just finished. The progress page may still be polling for the final
+			// "initializationComplete = true" response so it can leave the wizard. Keep serving
+			// those requests until the response has been delivered and the page has moved on to the
+			// application (seen as the next non-AJAX request).
+			if (PROGRESS_VM_AJAXREQUEST.equals(httpRequest.getParameter("page"))) {
+				return false;
+			}
+			if (initializationProgressResponseDelivered) {
```
dkayiwa: Does this window ever close on a container install? `initializationProgressResponseDelivered` only turns true when something asks for `page=progress.vm.ajaxRequest`, and a headless auto install has nobody sitting on the progress page. `StartupFilter.doFilter`'s `/health/alive` branch calls `triggerSetup`, which fires one server-side GET at `/` and throws the response away, and with `install_method=auto` (`api/src/main/resources/installation.h2.properties`, wired in by `startup.sh`) that GET goes straight to `autoRunOpenMRS`. When the install finishes, `servingInitializationProgress` is true and delivered is still false, and neither `/health/alive` nor `/health/started` reaches `skipFilter`, so no probe can close it either.

I checked against the real filter: in that state, after 100 ordinary requests `servingInitializationProgress` is still true and `skipFilter` has called `getParameter("page")` 100 times. So on the Docker and auto-install deployments, the call this PR removes comes straight back for the life of the JVM. Is that acceptable, or should the window close on something that does not need a browser?

web/src/main/java/org/openmrs/web/filter/update/UpdateFilter.java
```diff
@@ -762,6 +803,8 @@ public void executing(ChangeSet changeSet, int numChangeSetsToRun) {
 							return;
 						}
 
+						// keep serving the progress page's final poll so it can leave the wizard
+						setServingUpdateProgress(true);
```
dkayiwa: `setServingUpdateProgress(true)` sits here in the try body, but `setUpdatesRequired(false)` also runs from the `finally` below whenever `hasErrors()` is false, and the generic `catch (Exception e)` a few lines up returns without calling `reportError`. That path therefore marks the updates as no longer required while leaving the progress window shut. I drove the real `UpdateFilterCompletion` runnable into it: `updatesRequired=false`, `servingUpdateProgress=false`, and `skipFilter` on the progress poll then returns true. The page still recovers through the five-failed-polls fallback, so it costs a few seconds and some console noise rather than hanging, but the two flags should move together.

`InitializationFilter` already does it that way, with both assignments inside `if (!hasErrors())` in the finally. The same shape here means dropping lines 806-809 and putting the pair in the finally, which also clears up the duplicate `setUpdatesRequired(false)` that predates this PR.

Worth noting that no test would have caught this. The new tests set `servingUpdateProgress` and `servingInitializationProgress` by reflection, so nothing exercises the completion code actually opening the window in either filter.

dkayiwa:
@claude review

jwnasambu:
@claude review

jwnasambu:
@claude review

web/src/test/java/org/openmrs/web/filter/initialization/InitializationFilterE2ETest.java
```diff
@@ -79,11 +80,14 @@ void setup() {
 		// since SchedulerConfig.dataSource() reads them at bean-creation time.
 		originalRuntimeProperties = Context.getRuntimeProperties();
 		Context.setRuntimeProperties(new Properties());
+		filter.setInitializationComplete(false);
+		InitializationFilter.setInstallationStarted(false);
 	}
 
 	@AfterEach
-	void cleanup() {
+	void cleanup() throws Exception {
```
ibacher: Like Sonar says... Don't add this.

web/src/main/java/org/openmrs/web/filter/update/UpdateFilter.java
```diff
@@ -761,11 +774,9 @@ public void executing(ChangeSet changeSet, int numChangeSetsToRun) {
 							reportError(ErrorMessageConstants.UPDATE_ERROR_COMPLETE_STARTUP, e.getMessage());
 							return;
```
ibacher: Basically, the point here is that previously this short-circuited lines 764-766, but since this PR removes these lines, we also don't need the `return` here. Remove it.

jwnasambu:
@ibacher Kindly feel free to review my PR at your convenient time please!

web/src/test/java/org/openmrs/web/filter/initialization/InitializationFilterE2ETest.java
```diff
@@ -931,9 +935,89 @@ void databaseSetup_shouldRequireCreatePasswordWhenCreatingNewDb() throws Excepti
...
+			public String getParameter(String name) {
+				throw new AssertionError("skipFilter parsed request parameters after initialization completed");
+			}
+		};
+		req.setServletPath("/index.htm");
+
+		assertTrue(filter.skipFilter(req));
+	}
+
+	// ========== Helper Methods ==========
```
dkayiwa, line 1005: This banner took the place of the javadoc that explained why `getErrors()` goes through reflection, which is not something the ticket asks for. `clearErrors()` just below keeps its javadoc, and the matching banner in `UpdateFilterE2ETest` has a blank line under it, so both can come back:

```suggestion
	// ========== Helper Methods ==========

	/**
	 * Access the errors map from the filter via reflection since it is protected in StartupFilter.
	 */
```

web/src/test/java/org/openmrs/web/filter/initialization/InitializationFilterE2ETest.java
dkayiwa: `setup()` above still carries `throws Exception`. The reflection helpers that needed it went away with the flag approach, and nothing left in that body throws a checked exception: I removed the keyword locally and `mvn -pl web test-compile` was clean, so it can go the same way as the one on `cleanup()`.

Separately in this block, `filter.setInitializationComplete(false)` reaches a static method through an instance, while the line right under it calls `InitializationFilter.setInstallationStarted(false)` on the class. Using the class name at the call sites in this file would keep the two setters reading alike.

web/src/test/java/org/openmrs/web/filter/initialization/InitializationFilterE2ETest.java
```diff
@@ -931,9 +935,74 @@ void databaseSetup_shouldRequireCreatePasswordWhenCreatingNewDb() throws Excepti
...
+		pollReq.setServletPath("/" + WebConstants.SETUP_PAGE_URL);
+		pollReq.setParameter("page", "progress.vm.ajaxRequest");
+		MockHttpServletRequest otherReq = new MockHttpServletRequest();
+		otherReq.setServletPath("/module/x/y.form");
+
+		// an unrelated request passes through but never affects the poll
+		assertTrue(filter.skipFilter(otherReq));
+		assertFalse(filter.skipFilter(pollReq));
+		assertFalse(filter.skipFilter(pollReq));
+	}
```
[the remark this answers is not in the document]
dkayiwa: The half of this that clears out the redundant test is still open. I copied both test files onto master unchanged and ran them: the failures there are `skipFilter_shouldNotReadRequestParameters...` and `skipFilter_shouldReturnTrueForProgressPollOffSetupPage` in each file, four in total, and `skipFilter_shouldKeepServingProgressPollRegardlessOfOtherRequests` passes on master too, so it pins nothing this PR changes. The two branches it walks are already covered, by `skipFilter_shouldReturnFalseForProgressPollOnSetupPage` for the poll and by the two tests that return true off the setup path for the other request. With no state left in `skipFilter`, its name also promises a property that nothing can break. I would drop it from here and from `UpdateFilterE2ETest`.

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -204,9 +204,9 @@ public class InitializationFilter extends StartupFilter {
 	/**
 	 * Variable set at the end of the wizard when spring is being restarted
 	 */
-	private static boolean initializationComplete = false;
+	private static volatile boolean initializationComplete = false;
 
-	protected synchronized void setInitializationComplete(boolean initializationComplete) {
+	protected static void setInitializationComplete(boolean initializationComplete) {
```
ibacher:
```suggestion
	protected void setInitializationComplete(boolean initializationComplete) {
```

ibacher (approved):
One small thing and I think this is fine

jwnasambu:

web/src/main/java/org/openmrs/web/filter/initialization/InitializationFilter.java
```diff
@@ -204,10 +204,10 @@ public class InitializationFilter extends StartupFilter {
 	/**
 	 * Variable set at the end of the wizard when spring is being restarted
 	 */
-	private static boolean initializationComplete = false;
+	private static volatile boolean initializationComplete = false;
 
-	protected synchronized void setInitializationComplete(boolean initializationComplete) {
-		InitializationFilter.initializationComplete = initializationComplete;
+	protected void setInitializationComplete(boolean complete) {
```
ibacher:
```suggestion
	protected static void setInitializationComplete(boolean complete) {
```

jwnasambu:

jwnasambu:
@ibacher Kindly any changes to be fixed for this PR to be merged?

jwnasambu:
@dkayiwa, @ibacher, @wikumChamith  Kindly feel free to review my PR at your convenient time please!

web/src/test/java/org/openmrs/web/filter/initialization/InitializationFilterE2ETest.java
dkayiwa: `setInitializationComplete` is `static`, so the eight calls through `filter` in this file (`setup`, `cleanup`, and the six new `skipFilter` tests) are static calls made through an instance. Qualifying them with the class name would match the `setInstallationStarted` line sitting right next to them. Cosmetic, so nothing here blocks the merge.

jwnasambu:

PR #4256, opened 2023-02-27
Switching from Hibernate Mappings to Annotations - Location Domain

mherman22:
## Description of what I changed

I changed the location domain' reliance on xml mappings to use the more understandable JPA annotations. I used annotations like @Column, @Entity @Table @OneToMany and many others as can be seen in the changes in this pull request.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-5808

## Checklist: I completed these to help reviewers :)

- [x] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

  No? Unsure? -> [configure your IDE](https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE), format the code and add the changes with `git add . && git commit --amend`

- [x] I have **added tests** to cover my changes. (If you refactored
  existing code that was well tested you do not have to add tests)

  No? -> write tests and add them to this commit `git add . && git commit --amend`

- [x] I ran `mvn clean package` right before creating this pull request and
  added all formatting changes to my commit.

  No? -> execute above command

- [x] All new and existing **tests passed**.

  No? -> figure out why and add the fix to your commit. It is your responsibility to make sure your code works.

- [x] My pull request is **based on the latest changes** of the master branch.

  No? Unsure? -> execute command `git pull --rebase upstream master`

PR #6444, opened 2026-08-10
TRUNK-6684: Location.childLocations gained unintended orphanRemoval=true in the JPA migration

IamMujuziMoses:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6684

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/Location.java
+++ b/api/src/main/java/org/openmrs/Location.java
@@ -137,7 +137,7 @@ public class Location extends BaseCustomizableMetadata<LocationAttribute> implem
 	@JoinColumn(name = "parent_location")
 	private Location parentLocation;
 
-	@OneToMany(mappedBy = "parentLocation", cascade = CascadeType.ALL, orphanRemoval = true)
+	@OneToMany(mappedBy = "parentLocation", cascade = CascadeType.ALL)
 	@BatchSize(size = 100)
 	@OrderBy("name")
 	private Set<Location> childLocations;
--- a/api/src/test/java/org/openmrs/LocationTest.java
+++ b/api/src/test/java/org/openmrs/LocationTest.java
@@ -10,15 +10,10 @@
 package org.openmrs;
 
 import org.junit.jupiter.api.Test;
-import org.openmrs.api.LocationService;
-import org.openmrs.api.context.Context;
-import org.openmrs.test.jupiter.BaseContextSensitiveTest;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-public class LocationTest extends BaseContextSensitiveTest {
+public class LocationTest {
 
 	/**
 	 * Get locations that have any of specified set of tags among its child tags.
@@ -36,33 +31,6 @@ public class LocationTest extends BaseContextSensitiveTest {
 
 		assertTrue(Location.isInHierarchy(locationChild, locationParent));
 		assertTrue(Location.isInHierarchy(locationChild, locationGrandParent));
-	}
-
-	/**
-	 * @see Location#removeChildLocation(Location)
-	 */
-	@Test
-	public void removeChildLocation_shouldNotDeleteLocationWhenReparenting() {
-		executeDataSet("org/openmrs/api/include/LocationServiceTest-initialData.xml");
-		LocationService locationService = Context.getLocationService();
-
-		Location child = locationService.getLocation(2);
-		Location oldParent = child.getParentLocation();
-		Location newParent = locationService.getLocation(3);
-
-		// Removing the child should only change the relationship, not delete the location.
-		oldParent.removeChildLocation(child);
-		child.setParentLocation(newParent);
-		locationService.saveLocation(child);
-
-		Context.flushSession();
-		Context.clearSession();
-
-		// Reload the child to verify that it still exists after the relationship change.
-		Location reloadedChild = locationService.getLocation(child.getLocationId());
-
-		assertNotNull(reloadedChild);
-		assertEquals(newParent.getLocationId(), reloadedChild.getParentLocation().getLocationId());
 	}
 
 }
--- a/api/src/test/java/org/openmrs/LocationTest.java
+++ b/api/src/test/java/org/openmrs/LocationTest.java
@@ -10,10 +10,15 @@
 package org.openmrs;
 
 import org.junit.jupiter.api.Test;
+import org.openmrs.api.LocationService;
+import org.openmrs.api.context.Context;
+import org.openmrs.test.jupiter.BaseContextSensitiveTest;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-public class LocationTest {
+public class LocationTest extends BaseContextSensitiveTest {
 
 	/**
 	 * Get locations that have any of specified set of tags among its child tags.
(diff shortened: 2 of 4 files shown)


IamMujuziMoses:

dkayiwa:
@claude review

api/src/test/java/org/openmrs/LocationTest.java
```diff
@@ -10,10 +10,15 @@
...
+import org.openmrs.api.LocationService;
+import org.openmrs.api.context.Context;
+import org.openmrs.test.jupiter.BaseContextSensitiveTest;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-public class LocationTest {
+public class LocationTest extends BaseContextSensitiveTest {
```
jwnasambu: Kindly, this makes the whole LocationTest class boot the Spring/H2 context (BaseContextSensitiveTest), whereas it was a plain unit test before. The test needs the service and persistence stack, so it would fit more naturally in LocationServiceTest, which already loads this dataset keeping LocationTest fast and dependency-free. Optional and non-blocking.

IamMujuziMoses: thank you @jwnasambu for the review. I’ll move the regression test to LocationServiceTest, since it requires the service and persistence context and that class already uses the relevant dataset.

jwnasambu:
@ibacher , @wikumChamith , @dkayiwa kindly help me review this PR at your convenient time please.

api/src/test/java/org/openmrs/api/LocationServiceTest.java
```diff
@@ -217,6 +217,32 @@ public void saveLocation_shouldRemoveChildLocationFromLocation() {
 		assertEquals(1, newSavedLocation.getChildLocations().size());
```
dkayiwa: This older assertion doesn't actually reach the database, which is worth fixing while you're in this file (it doesn't block your change). `ls.getLocation(...)` hands back the same instance the session already holds, so the count only reflects the in-memory removal. I ran it both ways: it passes on master with `orphanRemoval` still on, and with your change applied plus `Context.flushSession(); Context.clearSession();` inserted before the reload, it fails with `expected: <1> but was: <2>`. So this test cannot see the bug you're fixing in either direction, and your new test below is the only one pinning real DB state.

Nothing persists because `removeChildLocation` only calls `childLocations.remove(child)`, whereas `addChildLocation` also does `child.setParentLocation(this)`. With the collection `mappedBy`, the parent side is never written, so removing alone cannot change a row. That is faithful to the 2.6.x mapping, so I would leave the method alone, and instead have this test assert the ticket's first acceptance criterion (the child row survives the removal and save) rather than a count that the session cache satisfies on its own.

api/src/main/java/org/openmrs/Location.java
```diff
@@ -137,7 +137,7 @@ public class Location extends BaseCustomizableMetadata<LocationAttribute> implem
 	@JoinColumn(name = "parent_location")
 	private Location parentLocation;
 
-	@OneToMany(mappedBy = "parentLocation", cascade = CascadeType.ALL, orphanRemoval = true)
+	@OneToMany(mappedBy = "parentLocation", cascade = CascadeType.ALL)
```
dkayiwa, line 140: While checking whether this was a one-off, I went through every `orphanRemoval = true` on master and compared each against the hbm mapping as it stood immediately before that entity was migrated. All of them trace back to a `delete-orphan` cascade except one: `OrderGroup.nestedOrderGroups` was `cascade="delete"` in OrderGroup.hbm.xml right before TRUNK-5840 migrated it, and came out as `cascade = { CascadeType.REMOVE }, orphanRemoval = true`. Nothing in core touches that collection, so no test would have caught it. Was that one deliberate, or the same slip as here? If it's the same, it deserves its own ticket rather than widening this PR.

IamMujuziMoses: @dkayiwa thanks for the review. I wasn’t aware of the `OrderGroup.nestedOrderGroups` discrepancy, so I don’t want to assume that it was intentional. I agree that it’s better handled separately rather than widening this PR, and I’ll keep this one focused on `Location.childLocations`.

api/src/test/java/org/openmrs/api/LocationServiceTest.java
IamMujuziMoses: @dkayiwa thanks for verifying this, I am going to do the updates on the test.

api/src/test/java/org/openmrs/api/LocationServiceTest.java
```diff
@@ -207,15 +207,42 @@ public void saveLocation_shouldRemoveChildLocationFromLocation() {
...
+		location.removeChildLocation(child);
 		ls.saveLocation(location);
 
-		Location newSavedLocation = ls.getLocation(location.getLocationId());
+		Context.flushSession();
+		Context.clearSession();
 
-		// Saved location should have 1 child locations now
-		assertEquals(1, newSavedLocation.getChildLocations().size());
+		assertNotNull(ls.getLocation(child.getLocationId()));
```
dkayiwa, line 219: The rewrite leaves this test's name and the sentence above it saying the opposite of what it now checks. `saveLocation_shouldRemoveChildLocationFromLocation` and "You should be able to remove a child location from a location" both describe a removal, while the assertion is that the child survives.

That matters beyond naming, because `LocationService.saveLocation` carries the matching contract line at LocationService.java:56, `<strong>Should</strong> remove child location from location`, and this change makes it false. I checked on your branch: after `removeChildLocation` plus `saveLocation` plus a flush and clear, location 1 still reports 2 children and location 2 still has `parent_location = 1`, so the removal now persists nothing at all. Before the change that line was satisfied, destructively, by the orphan delete you are taking out.

Could you rename the test to what it now pins, say `saveLocation_shouldNotDeleteChildLocationWhenRemovedFromParent`, adjust the sentence above it, and reword that Should line so it states the real guarantee? Otherwise the next person writes against a documented operation that quietly does nothing. None of this blocks the merge.

IamMujuziMoses: right, I will update the test name and description to reflect the actual behavior, and change the `saveLocation` contract to state that removing a child locationn must not delete it

api/src/main/java/org/openmrs/Location.java
dkayiwa, line 140: Does this need to go to the release branches as well? `orphanRemoval` arrived with the annotation migration in #4256 back in March 2023, so it is in every release from 2.7.0 through 2.8.5, and I checked the branch heads directly: 2.7.x, 2.8.x and 2.9.x all still carry this exact line, character for character. Fixing master alone leaves the delete in place for everyone actually running OpenMRS today.

One thing worth knowing for that decision is that the damage is not always the silent delete the ticket describes. I put `orphanRemoval = true` back on your branch and ran the location tests, and the re-parenting path came out as `ConstraintViolationException: Referential integrity constraint violation: PUBLIC.ENCOUNTER FOREIGN KEY(LOCATION_ID) REFERENCES PUBLIC.LOCATION(LOCATION_ID) (2)`. So on a populated database a location that has encounters turns an unrelated re-parent into a hard failure, while one with nothing pointing at it just disappears.

None of this blocks the PR as it stands. It is really a question of whether TRUNK-6684 should also pick up fix versions for the maintained lines.

IamMujuziMoses: Good point. Since the unintended `orphanRemoval` is present in the maintained release branches, I agree that [TRUNK-6684](https://openmrs.atlassian.net/browse/TRUNK-6684) should include fix versions for those branches as well.

ibacher: Yep. I've been using the `backport` label in Jira to signal tickets that should be backported. Granted, I've stopped backporting most things to 2.7, but this probably should land there.

ibacher (approved):

PR #6450, opened 2026-08-13
TRUNK-6770: Replace the reflective String-constructor conversion in getGlobalPropertyValue

1-cant-code:
…getGlobalPropertyValue

This PR replaces the reflective conversion with explicit conversion for String, Boolean, Integer, Long, Double, Float, and enum types.
 The reflective conversion is preserved as a fallback for other types with a String constructor, maintaining the existing behavior, and arbitrary String-constructible types such as BigInteger.
 Unit tests was added/udated to cover the supported types, invalid values falling back to the default value 
 The explicit conversion avoid relying on String constructors of wrapper classes that are deprecated for removal in newer versions of JDK
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6770

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
@@ -731,7 +731,7 @@ public class AdministrationServiceImpl extends BaseOpenmrsService implements Adm
 	 *      java.lang.Object)
 	 */
 	@Override
-	@SuppressWarnings("unchecked")
+	@SuppressWarnings({ "unchecked", "rawtypes" })
 	public <T> T getGlobalPropertyValue(String propertyName, T defaultValue) throws APIException {
 		if (defaultValue == null) {
 			throw new IllegalArgumentException("The defaultValue argument cannot be null");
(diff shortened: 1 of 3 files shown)


api/src/main/resources/messages.properties
1-cant-code, line 1: Do we need to add translations for unsupported.global.property.type to the other messages_*.properties files as well (ru, sw, etc.), or is the default messages.properties enough?

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
```diff
@@ -532,14 +534,86 @@ public void getGlobalPropertyValue_shouldReturnDefaultValueForMissingProperty()
...
+		executeDataSet(ADMIN_INITIAL_DATA_XML);
+
+		Object retValue = adminService.getGlobalPropertyValue("valid.enum", TestEnum.FIRST);
+
+		assertInstanceOf(TestEnum.class, retValue);
+		assertEquals(TestEnum.SECOND, retValue);
+	}
+
+	@Test
+	public void getGlobalPropertyValue_shouldThrowExceptionForUnsupportedType() {
```
1-cant-code: The previous implementation supported any type with an accessible String constructor, for example BigInteger. With the requested explicit conversion, BigInteger would become unsupported and throw an APIException. Is this intentional, or should we preserve support for arbitrary String-constructible types?

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
```diff
@@ -742,18 +742,35 @@ public <T> T getGlobalPropertyValue(String propertyName, T defaultValue) throws
 			return defaultValue;
 		}
 
+		Class<?> type = defaultValue.getClass();
```
ibacher, line 745: Don't we "know" that this is, in fact `Class<T>`?

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
```diff
@@ -742,18 +742,35 @@ public <T> T getGlobalPropertyValue(String propertyName, T defaultValue) throws
...
+				return (T) Float.valueOf(propVal);
+			}
+			if (type.isEnum()) {
+				return (T) Enum.valueOf((Class<? extends Enum>) type, propVal);
+			}
+		} catch (IllegalArgumentException e) {
+			log.error("Unable to turn value '{}' into type {}", propVal, type.getName(), e);
 			return defaultValue;
 		}
+		throw new APIException("unsupported.global.property.type", new Object[] { type.getName() });
```
ibacher: I don't think we should throw-away the reflective construction, and the ticket doesn't ask for that. Just for us to not use it for primitive classes we know are going away. Otherwise, this should continue to work as-is.

api/src/main/resources/messages.properties
ibacher: Default is enough. We handle translations separately (although I guess the removals are ok).

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
ibacher: We want to preserve support for arbitrary String-constructible types. The ask was just to get rid of the calls to constructors we know are going away.

ibacher (changes requested):
Thanks for the contribution @1-cant-code!

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
1-cant-code: I tried using `Class<T>` here, but it requires an unchecked cast like 
`Class<T> type = (Class<T>) defaultValue.getClass();`
Because native getClass() returns Class<?>
It also doesn't help with the enum conversion, since Enum.valueOf requires Class<T extends Enum<T>>, while our T is not bounded by Enum. We would still need an additional cast there.

jwnasambu (commented):
@1-cant-code Kindly add a brief description of the changes made in this PR so that reviewers can easily understand what has been changed. Thank you.

api/src/test/java/org/openmrs/api/AdministrationServiceTest.java
1-cant-code: updated the test to preserve the existing behavior for arbitrary String-constructible types. BigInteger is now tested to ensure it is successfully converted through the reflective fallback instead of throwing an APIException

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
1-cant-code: Removed the unsupported.global.property.type exception and restored the reflective construction

1-cant-code:
Hi @ibacher, I've addressed the review comments and pushed the changes. Could you please take another look at the PR when you have a chance? Thanks!

ibacher (approved):

1-cant-code:
Hi @ibacher, I fixed the Spotless formatting issue with mvn spotless:apply and pushed the changes. The GitHub Actions build is currently showing “Action required”. Could you please approve/run it when you have a chance?

PR #6453, opened 2026-08-13
TRUNK-6771: Make ConfigUtil global property cache thread-safe and self-clearing

Skeletor-Pirate:
Make `ConfigUtil`'s global property cache thread-safe and self-clearing. The previous implementation was a static plain `HashMap` with three defects:

1. **Not thread-safe** - written from request threads on read and from the thread saving a global property via the `GlobalPropertyListener` callbacks, with no synchronization.
2. **Stale across contexts** - static and never cleared, so entries survived context refresh and module start/stop.
3. **Permanent negative caching** - a lookup for a missing property cached a null value and `containsKey` reported a hit forever, so a property created later (e.g. by a module via Liquibase at startup) kept reading as null.

Changes:
- **`ConcurrentHashMap`** instead of `HashMap` - concurrent reads and listener updates cannot corrupt the internal table.
- **Missing properties are no longer cached** - a property that did not exist at first read is re-queried on the next read, so it is picked up once created without relying on a listener callback. The `globalPropertyChanged` callback is guarded against null values (it removes the entry rather than attempting to store null, which `ConcurrentHashMap` forbids).
- **`ApplicationListener<ContextRefreshedEvent>`** - the cache is cleared on context refresh (the same pattern as `HandlerUtil` and `RolePrivilegeCache`), so entries cannot outlive the context that populated them. A public `clearGlobalPropertyCache()` is exposed for testing.

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-6771

## Tests
Added three cases to `ConfigUtilTest` (14/14 pass):

- `shouldNotCorruptCacheUnderConcurrentAccess` - 8 threads, 200 iterations of interleaved reads and listener changes.
- `shouldClearCacheOnContextRefresh` - an `onApplicationEvent(ContextRefreshedEvent)` clears the cache so the next read observes a value changed behind the cache's back.
- `shouldNotCacheMissingPropertiesPermanently` - a property read as missing is picked up once inserted directly via SQL (no listener involved).

Verification: `mvn -pl api -am test -Dtest=ConfigUtilTest` - BUILD SUCCESS, 14/14 pass.

  Added tests as described above.

### DIFF
--- a/api/src/main/java/org/openmrs/util/ConfigUtil.java
+++ b/api/src/main/java/org/openmrs/util/ConfigUtil.java
@@ -10,6 +10,7 @@
 package org.openmrs.util;
 
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.commons.lang3.StringUtils;
@@ -26,17 +27,25 @@ public class ConfigUtil implements GlobalPropertyListener, ApplicationListener<C
 
 	/**
 	 * Cache of global property key/value pairs to enable lookups that do not require accessing the
-	 * service each time. Uses a concurrent map so that concurrent reads and updates from the
-	 * {@link GlobalPropertyListener} callbacks cannot corrupt it. Missing properties are deliberately
-	 * not cached so that a property created after the first read is picked up rather than being pinned
-	 * to null.
+	 * service each time.
 	 */
 	private static final Map<String, String> globalPropertyCache = new ConcurrentHashMap<>();
+
+	/**
+	 * Tracks global property lookups that came back empty, mirroring {@link #globalPropertyCache} for
+	 * cache misses so that a genuinely absent property is not re-queried on every read. Cleared
+	 * together with {@link #globalPropertyCache} when the property changes, is deleted, or when the
+	 * context is refreshed.
+	 */
+	private static final Set<String> missedGlobalProperties = ConcurrentHashMap.newKeySet();
 
 	/**
 	 * Gets the value of the given OpenMRS global property
 	 */
 	public static String getGlobalProperty(String propertyName) {
+		if (missedGlobalProperties.contains(propertyName)) {
+			return null;
+		}
 		String cachedValue = globalPropertyCache.get(propertyName);
 		if (cachedValue != null) {
 			return cachedValue;
@@ -44,16 +53,18 @@ public class ConfigUtil implements GlobalPropertyListener, ApplicationListener<C
 		String value = Context.getAdministrationService().getGlobalProperty(propertyName);
 		if (value != null) {
 			globalPropertyCache.putIfAbsent(propertyName, value);
+		} else {
+			missedGlobalProperties.add(propertyName);
 		}
 		return value;
 	}
 
 	/**
-	 * Clears the cache of global property values. Invoked on context refresh via
-	 * {@link #onApplicationEvent} so that cached values cannot outlive the context that populated them.
+	 * Clears the cached global property values and recorded cache misses.
 	 */
-	public static void clearGlobalPropertyCache() {
+	static void clearGlobalPropertyCache() {
 		globalPropertyCache.clear();
+		missedGlobalProperties.clear();
 	}
 
 	/**
@@ -134,16 +145,19 @@ public class ConfigUtil implements GlobalPropertyListener, ApplicationListener<C
 
 	@Override
 	public void globalPropertyChanged(GlobalProperty newValue) {
+		String property = newValue.getProperty();
 		if (newValue.getPropertyValue() == null) {
-			globalPropertyCache.remove(newValue.getProperty());
+			globalPropertyCache.remove(property);
 		} else {
-			globalPropertyCache.put(newValue.getProperty(), newValue.getPropertyValue());
+			globalPropertyCache.put(property, newValue.getPropertyValue());
 		}
+		missedGlobalProperties.remove(property);
 	}
 
 	@Override
 	public void globalPropertyDeleted(String propertyName) {
 		globalPropertyCache.remove(propertyName);
+		missedGlobalProperties.add(propertyName);
 	}
 
 	@Override
--- a/api/src/main/java/org/openmrs/util/ConfigUtil.java
+++ b/api/src/main/java/org/openmrs/util/ConfigUtil.java
@@ -9,35 +9,51 @@
  */
 package org.openmrs.util;
 
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.commons.lang3.StringUtils;
 import org.openmrs.GlobalProperty;
 import org.openmrs.api.GlobalPropertyListener;
 import org.openmrs.api.context.Context;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextRefreshedEvent;
 
 /**
  * A utility class for working with configuration properties
  */
-public class ConfigUtil implements GlobalPropertyListener {
+public class ConfigUtil implements GlobalPropertyListener, ApplicationListener<ContextRefreshedEvent> {
 
 	/**
 	 * Cache of global property key/value pairs to enable lookups that do not require accessing the
-	 * service each time
+	 * service each time. Uses a concurrent map so that concurrent reads and updates from the
+	 * {@link GlobalPropertyListener} callbacks cannot corrupt it. Missing properties are deliberately
+	 * not cached so that a property created after the first read is picked up rather than being pinned
+	 * to null.
 	 */
-	private static final Map<String, String> globalPropertyCache = new HashMap<>();
+	private static final Map<String, String> globalPropertyCache = new ConcurrentHashMap<>();
 
 	/**
 	 * Gets the value of the given OpenMRS global property
 	 */
 	public static String getGlobalProperty(String propertyName) {
-		if (globalPropertyCache.containsKey(propertyName)) {
-			return globalPropertyCache.get(propertyName);
+		String cachedValue = globalPropertyCache.get(propertyName);
+		if (cachedValue != null) {
+			return cachedValue;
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/util/ConfigUtil.java
```diff
@@ -9,37 +9,53 @@
...
-public class ConfigUtil implements GlobalPropertyListener {
+public class ConfigUtil implements GlobalPropertyListener, ApplicationListener<ContextRefreshedEvent> {
 
 	/**
 	 * Cache of global property key/value pairs to enable lookups that do not require accessing the
-	 * service each time
+	 * service each time. Uses a concurrent map so that concurrent reads and updates from the
+	 * {@link GlobalPropertyListener} callbacks cannot corrupt it. Missing properties are deliberately
+	 * not cached so that a property created after the first read is picked up rather than being pinned
+	 * to null.
```
ibacher: I think the existing document is good enough. The rest of this is either clear from the code (I hope!) or refers to guarantees that this does not make.

```suggestion
	 * service each time.
```

api/src/main/java/org/openmrs/util/ConfigUtil.java
```diff
@@ -9,37 +9,53 @@
...
+	private static final Map<String, String> globalPropertyCache = new ConcurrentHashMap<>();
 
 	/**
 	 * Gets the value of the given OpenMRS global property
 	 */
 	public static String getGlobalProperty(String propertyName) {
-		if (globalPropertyCache.containsKey(propertyName)) {
-			return globalPropertyCache.get(propertyName);
+		String cachedValue = globalPropertyCache.get(propertyName);
+		if (cachedValue != null) {
```
ibacher, line 50: So the problem here is that it doesn't really help out with with missing GPs, which are _also_ something we want to cache. Since this is already a `GlobalPropertyListener`, it should be notified every time the value is changed by the app (and that's probably sufficient). A better fix here, I think, is something like what was done with the [FhirGlobalPropertyHolder](https://github.com/openmrs/openmrs-module-fhir2/blob/b6ecdd4dbfae05109833b164d3a66792b91f9738/api/src/main/java/org/openmrs/module/fhir2/api/util/FhirGlobalPropertyHolder.java#L28), which is similarly a cache for Global Properties.

Specifically, we cache misses (albeit as a separate set) and reset them either when the global property is changed or when the context refreshes (which is how we deal with the "liquibase changes" the ticket mentions).

ibacher (commented):
Thanks! One small note below, but otherwise this looks good to me.

api/src/test/java/org/openmrs/util/ConfigUtilTest.java
```diff
@@ -84,6 +98,55 @@ public void shouldGetUpdatedGlobalPropertyIfDeleted() {
...
+		    "UPDATE global_property SET property_value = 'refreshed-value' WHERE property = 'mail.user'", false);
+		Context.flushSession();
+		Context.clearSession();
+		assertThat(ConfigUtil.getGlobalProperty("mail.user"), is("user1@test.com"));
+		configUtil.onApplicationEvent(new ContextRefreshedEvent(applicationContext));
+		assertThat(ConfigUtil.getGlobalProperty("mail.user"), is("refreshed-value"));
+	}
+
+	@Test
+	public void shouldNotCorruptCacheUnderConcurrentAccess() throws Exception {
```
ibacher, line 142: Does this actually fail without the code changes here?

api/src/test/java/org/openmrs/util/ConfigUtilTest.java
```diff
@@ -84,6 +98,55 @@ public void shouldGetUpdatedGlobalPropertyIfDeleted() {
...
+
+	@Test
+	public void shouldClearCacheOnContextRefresh() {
+		assertThat(ConfigUtil.getGlobalProperty("mail.user"), is("user1@test.com"));
+		administrationService.executeSQL(
+		    "UPDATE global_property SET property_value = 'refreshed-value' WHERE property = 'mail.user'", false);
+		Context.flushSession();
+		Context.clearSession();
+		assertThat(ConfigUtil.getGlobalProperty("mail.user"), is("user1@test.com"));
+		configUtil.onApplicationEvent(new ContextRefreshedEvent(applicationContext));
```
ibacher, line 137: Seems like it might be better to fire this through Spring so the test ensures that it's wired up (`ConfigUtil` is already `@Autowired` here).

api/src/test/java/org/openmrs/util/ConfigUtilTest.java
```diff
@@ -84,6 +98,55 @@ public void shouldGetUpdatedGlobalPropertyIfDeleted() {
 		assertThat(ConfigUtil.getGlobalProperty("mail.user"), nullValue());
 	}
 
+	@Test
+	public void shouldNotCacheMissingPropertiesPermanently() {
```
ibacher: We should actually clean-up the global property cache after each test.

api/src/main/java/org/openmrs/util/ConfigUtil.java
```diff
@@ -9,37 +9,53 @@
...
+			globalPropertyCache.putIfAbsent(propertyName, value);
+		}
 		return value;
 	}
 
+	/**
+	 * Clears the cache of global property values. Invoked on context refresh via
+	 * {@link #onApplicationEvent} so that cached values cannot outlive the context that populated them.
+	 */
+	public static void clearGlobalPropertyCache() {
```
ibacher: For new public API, we require a `@since` tag. This might just be a package-private method.

Skeletor-Pirate:
@ibacher thank you for the careful review — all fair points, and the pointer to FhirGlobalPropertyHolder is really helpful.

On the big one: you're right. Not caching a missing GP just trades one DB hit for another, since a genuinely absent property gets re-queried on every read. I'll restructure to the FhirGlobalPropertyHolder approach — a separate set that records misses, cleared when the property changes/deletes or when the context refreshes. That also makes the "self-clearing" claim in the commit message true in practice.

To answer your question on the test: yes, `shouldNotCacheMissingPropertiesPermanently` fails on master — the old code cached `null` values, so a property created after the first read stayed pinned to null until restart. Under the new miss-caching design that test needs a different setup though: creating the property through the admin service (which fires the GlobalPropertyListener) clears the miss, so I'll rework the test to assert that flow plus a context-refresh variant, rather than relying on raw SQL + a session clear.

I'll also trim the javadoc back to the original text, add a `@since` (or drop the method to package-private), and tear down the cache after each test. Updating the PR shortly.

Skeletor-Pirate:
@ibacher all six points are addressed and pushed to the branch:

1. **Cache misses** — `getGlobalProperty` now records empty lookups in a `missedGlobalProperties` set (mirroring `FhirGlobalPropertyHolder`), so a genuinely absent property isn't re-queried on every read. The miss is cleared in `globalPropertyChanged` and recorded in `globalPropertyDeleted`, and both caches reset on context refresh.

2. **`shouldNotCacheMissingPropertiesPermanently` reworked** — under miss-caching a raw-SQL insert would stay masked, so it's now `shouldPickUpGlobalPropertyCreatedAfterMiss`, creating the property through the admin service so the listener clears the miss. A new `shouldCacheMissingGlobalPropertyUntilChanged` covers the miss-stays-until-context-refresh behavior.

3. **Javadoc** — trimmed back to the original single sentence on the value cache.

4. **`clearGlobalPropertyCache`** — now package-private (tests live in the same package), so no new public API surface / `@since` needed.

5. **Teardown** — `@AfterEach` clears the cache so state doesn't leak between tests.

6. **`shouldNotCorruptCacheUnderConcurrentAccess`** — now also reads a pre-seeded missing property concurrently to exercise the miss set.

`ConfigUtilTest`: 15 run, 0 failures, 0 errors.

jwnasambu (commented):
@Skeletor-Pirate Could you kindly reformat this to match the repo's original PR template structure so it's consistent with other PRs? The format is:

## Issue I worked on

Skeletor-Pirate:
Done - the PR description has been reformatted to match the repo's PR template structure (`## Description of what I changed`, `## Issue I worked on`, `## Checklist`). Thanks for flagging it.

api/src/main/java/org/openmrs/util/ConfigUtil.java
Skeletor-Pirate: Applied your suggestion - now reads 'service each time.'.

api/src/main/java/org/openmrs/util/ConfigUtil.java
Skeletor-Pirate: Fixed. Missing lookups now get recorded in a separate set, cleared on change/delete and on context refresh, mirroring FhirGlobalPropertyHolder.

api/src/test/java/org/openmrs/util/ConfigUtilTest.java
Skeletor-Pirate: Yes - this is the concurrency test: with the plain HashMap the concurrent reads/writes via the listener callbacks can corrupt the cache, so this fails without the ConcurrentHashMap swap.

api/src/test/java/org/openmrs/util/ConfigUtilTest.java
Skeletor-Pirate: Good point - the refresh test now fires through the autowired ConfigUtil so it exercises the real wiring.

api/src/test/java/org/openmrs/util/ConfigUtilTest.java
Skeletor-Pirate: Added an @AfterEach teardown that clears the cache.

api/src/main/java/org/openmrs/util/ConfigUtil.java
Skeletor-Pirate: Made it package-private instead.

PR #6452, opened 2026-08-13
TRUNK-6769: Allow null default in getGlobalPropertyValue and fix WebUtil.formatDate

Skeletor-Pirate:
## Description of what I changed
Fix two bugs where a null default value broke global property reads:

- **`AdministrationService.getGlobalPropertyValue(String, T)`** threw `IllegalArgumentException` when the default value was null. `WebUtil.formatDate(Date, Locale, FORMAT_TYPE)` passes null as the default in both the `TIMESTAMP` and `TIME` branches, so both threw unconditionally and only the `DATE` branch (which passes `""`) worked. A null default now returns null instead of throwing, since there is no target type to convert to. Javadoc on `AdministrationService` updated to state this.
- **`WebUtil.formatDate`** now uses the untyped `getGlobalProperty(...)` accessor in all three branches (they read a plain `String` global property), instead of the converting `getGlobalPropertyValue(...)`.

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-6769

## Tests
- Updated `AdministrationServiceUnitTest#getGlobalPropertyValue_shouldReturnNullIfDefaultValueIsNull` (previously asserted the throw).
- Added six `WebUtilTest` cases: all three `FORMAT_TYPE` branches honour a configured `GP_SEARCH_DATE_DISPLAY_FORMAT`, and all three return formatted output when the property is unset.

Verification: `mvn -pl api,web -am test -Dtest=AdministrationServiceUnitTest,WebUtilTest` - BUILD SUCCESS, 25/25 tests pass (5 + 20).

## Checklist: I completed these to help reviewers :)
- [ ] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.
- [x] I have **added tests** to cover my changes. (If you refactored existing code that was well tested you do not have to add tests)

  Updated + added tests as described above.
- [x] I ran `./mvnw clean package` right before creating this pull request and added all formatting changes to my commit.
- [x] All new and existing **tests passed**.
- [x] My pull request is **based on the latest changes** of the master branch.

PR #6456, opened 2026-08-13
TRUNK-6769: Fix global property handling in WebUtil

waghmaretejas:

* Updated `getGlobalPropertyValue` to return `null` when the default value is `null` without attempting conversion.
* Updated `WebUtil.formatDate` to use `getGlobalProperty` for `GP_SEARCH_DATE_DISPLAY_FORMAT`.
* Added tests covering the null default value and all three `FORMAT_TYPE` branches with and without a configured format.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6769

* [x] My IDE is configured to follow the [**[code style](https://wiki.openmrs.org/display/docs/Java+Conventions)**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

  No? Unsure? -> [[configure your IDE](https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE)](https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE), format the code and add the changes with `git add . && git commit --amend`

* [x] I have **added tests** to cover my changes. (If you refactored

* [x] I ran `./mvnw clean package` right before creating this pull request and

  No? -> execute above command

* [x] All new and existing **tests passed**.

* [x] My pull request is **based on the latest changes** of the master branch.


web/src/test/java/org/openmrs/web/WebUtilTest.java
```diff
@@ -159,4 +169,81 @@ public static boolean containsId(Collection<? extends BaseOpenmrsObject> list, I
...
+
+	@Test
+	public void formatDate_shouldFormatTimestamp() {
+		Date date = new Date(1460323539000L);
+
+		when(administrationService.getGlobalProperty(OpenmrsConstants.GP_SEARCH_DATE_DISPLAY_FORMAT)).thenReturn("");
+
+		assertEquals("April 10, 2016, 9:25:39\u202fPM UTC",
+		    WebUtil.formatDate(date, Locale.US, Format.FORMAT_TYPE.TIMESTAMP));
+	}
```
jwnasambu, line 226: These two assertions fail on this branch under the project's own test setup. The root pom pins -Djava.locale.providers=COMPAT for the surefire JVM (pom.xml:825), and under COMPAT the default branches render as 9:25:39 PM (regular space) and April 10, 2016 9:25:39 PM UTC (no comma after the year), while these tests expect the CLDR forms 9:25:39\u202fPM (narrow no-break space) and April 10, 2016, 9:25:39\u202fPM UTC. ./mvnw -pl web test -Dtest=WebUtilTest fails exactly these two. They only pass when run without the pom's argLine, e.g. from an IDE where CLDR is the default provider. If merged as-is, the web suite fails on every standard run, so the checklist's claim that all tests passed doesn't hold. For the fix, I'd avoid pinning the exact default-branch strings at all: the output is locale-provider data that shifts again if COMPAT is ever removed. Asserting non-emptiness (or a loose pattern) for these two branches, as #6452 does, is robust, with the configured-format branches already covering the exact-string cases.

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
```diff
@@ -734,7 +734,7 @@ public GlobalProperty getGlobalPropertyByUuid(String uuid) {
 	@SuppressWarnings("unchecked")
 	public <T> T getGlobalPropertyValue(String propertyName, T defaultValue) throws APIException {
 		if (defaultValue == null) {
-			throw new IllegalArgumentException("The defaultValue argument cannot be null");
+			return null;
```
jwnasambu, line 737: Kindly was this behavior change in scope for TRUNK-6769? The WebUtil fix on its own only needs getGlobalProperty and doesn't depend on getGlobalPropertyValue returning null. This public method has failed loud on a null default since 2015; after this, getGlobalPropertyValue(name, null) silently returns null even when the property exists, so a caller passing a null default by mistake gets a silent null instead of an exception. There's also TRUNK-6770 (#6450) rewriting this same method's conversion path, so the two PRs collide there. Fine to keep if the ticket explicitly asks for it, just confirming the intent.

web/src/test/java/org/openmrs/web/WebUtilTest.java
```diff
@@ -159,4 +169,81 @@ public static boolean containsId(Collection<? extends BaseOpenmrsObject> list, I
...
+	private final AdministrationService administrationService = mock(AdministrationService.class);
+
+	private TimeZone originalTimeZone;
+
+	@BeforeEach
+	public void setup() {
+		originalTimeZone = TimeZone.getDefault();
+		TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
+
+		contextMockHelper.setAdministrationService(administrationService);
```
jwnasambu: Kindly the mocked AdministrationService is installed into the process-global Context.serviceContext static and never removed. ContextMockHelper's intended lifecycle you can see BaseContextMockTest reverts in @AfterEach; here tearDown() only restores the TimeZone. All web-module tests share one surefire JVM, so the mock stays registered after this class and any later plain unit test calling Context.getAdministrationService() would silently get the mock's null/default answers instead of the real service. It's currently masked only because the Spring-context tests reset the static. Adding contextMockHelper.revertMocks() to tearDown() would close that. The helper is also @Deprecated, with guidance to avoid Context.get...Service() in new tests.) Though the changes are optional

jwnasambu:
@claude review

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
waghmaretejas: Yes, this behavior change is explicitly in scope for TRUNK-6769.

The issue's requested change and acceptance criteria specifically require `getGlobalPropertyValue(name, null)` to return `null` without attempting conversion, rather than throwing `IllegalArgumentException`.

The WebUtil call site has also been changed to use `getGlobalProperty(...)`, since `GP_SEARCH_DATE_DISPLAY_FORMAT` is a String-valued property and does not require type conversion.

So I’ve kept the `return null` change as explicitly required by TRUNK-6769.

web/src/test/java/org/openmrs/web/WebUtilTest.java
waghmaretejas: Thanks for pointing this out. I’ve added `contextMockHelper.revertMocks()` to `tearDown()` so the mocked `AdministrationService` is removed from the global context after each test, preventing it from affecting subsequent tests.

I also verified that the `WebUtilTest` suite passes with the cleanup in place.

web/src/test/java/org/openmrs/web/WebUtilTest.java
waghmaretejas: Thanks for pointing this out. I’ve updated the TIME and TIMESTAMP default-format tests to avoid asserting locale-provider-dependent output. They now verify that the formatting produces a non-empty result, while the configured-format tests continue to assert the exact expected strings.

This keeps the tests independent of the JDK locale-provider implementation while still covering the formatting behavior required by the issue.

waghmaretejas:
@jwnasambu Thank you for taking the time to review my PR and for the detailed feedback. I’ve addressed the comments and pushed the updates. Whenever you have a chance, I’d really appreciate it if you could take another look. Thanks again!

waghmaretejas:
Hi @dkayiwa Just a quick follow-up on this PR. Could you please take a look when you get a chance? Thanks!

web/src/test/java/org/openmrs/web/WebUtilTest.java
dkayiwa: `revertMocks()` doesn't actually take the mock back out here, so this one is still open. `ContextMockHelper.setService` first records the "real" service via `Context.getService(...)` (ContextMockHelper.java:176), which throws in a plain unit test with no Spring context, so it stores `null`. `revertMocks()` then calls `ServiceContext.setService(AdministrationService.class, null)`, and that method opens with `if (cls != null && classInstance != null)` (ServiceContext.java:720), so a null instance is a no-op and the mock proxy stays in the services map. I checked it directly: stub `getGlobalProperty` with a sentinel value, call `revertMocks()`, then read it back through `Context.getAdministrationService()`, and the sentinel still comes out.

Nothing breaks today, so I wouldn't hold the PR for this. The full web suite is green (168 tests), and the only other web test that reaches for that service, `GZIPFilterTest`, extends `BaseWebContextSensitiveTest`, which boots Spring and replaces the entry. The cost lands on whoever adds the next plain unit test to this module: if it calls `Context.getAdministrationService()` and happens to run after this class, it silently gets Mockito defaults rather than failing, and surefire's filesystem run order decides whether it does.

If you'd rather close it properly, scope the mock to the test instead of registering it process wide, the way #6452 does for this same ticket:

```java
try (MockedStatic<Context> ignored = mockStatic(Context.class)) {
	AdministrationService adminService = mock(AdministrationService.class);
	when(Context.getAdministrationService()).thenReturn(adminService);
	when(adminService.getGlobalProperty(OpenmrsConstants.GP_SEARCH_DATE_DISPLAY_FORMAT))
	        .thenReturn("yyyy-MM-dd HH:mm:ss");
	assertEquals("2016-04-10 21:25:39", WebUtil.formatDate(date, Locale.US, FORMAT_TYPE.TIMESTAMP));
}
```

I ran that against this branch and it passes, and once the block closes `Context.getAdministrationService()` throws `ServiceNotFoundException` again, so nothing is left registered. It also gets you off `ContextMockHelper`, which is deprecated. The time zone setup can stay as it is, the exact string assertions still need it.

waghmaretejas: @dkayiwa Thanks for the detailed explanation. I’ve updated the tests to scope the `AdministrationService` mock to each test using `MockedStatic<Context>` and removed the deprecated `ContextMockHelper`.

The existing timezone setup and assertions are unchanged. `WebUtilTest` passes successfully.

waghmaretejas:
@dkayiwa The requested changes are now addressed. Could you please take a final look at the PR when you get a chance?

web/src/main/java/org/openmrs/web/WebUtil.java
```diff
@@ -385,23 +385,23 @@ public static String formatDate(Date date, Locale locale, FORMAT_TYPE type) {
...
+			        .getGlobalProperty(OpenmrsConstants.GP_SEARCH_DATE_DISPLAY_FORMAT);
 			if (StringUtils.isEmpty(dateTimeFormat)) {
 				dateFormat = DateFormat.getDateTimeInstance(DateFormat.LONG, DateFormat.LONG, locale);
 			} else {
 				dateFormat = new OpenmrsDateFormat(new SimpleDateFormat(dateTimeFormat), locale);
 			}
 		} else if (type == FORMAT_TYPE.TIME) {
 			String timeFormat = Context.getAdministrationService()
-			        .getGlobalPropertyValue(OpenmrsConstants.GP_SEARCH_DATE_DISPLAY_FORMAT, null);
+			        .getGlobalProperty(OpenmrsConstants.GP_SEARCH_DATE_DISPLAY_FORMAT);
```
dkayiwa, line 396: Once the property has a value, all three branches build the same formatter out of it, so the `FORMAT_TYPE` argument stops mattering. I stubbed `searchWidget.dateDisplayFormat` with `dd/MM/yyyy`, which is the shape its own description promises ("Date display format to be used to display the date somewhere in the UI i.e the search widgets and autocompletes"), and `DATE`, `TIME` and `TIMESTAMP` all came back as `10/04/2016`. A `TIME` caller gets a date, and a `TIMESTAMP` loses its time, with nothing thrown. Before this change those two branches threw on every call, so no caller could get a wrong-looking string out of them.

This doesn't block the PR. Nothing is broken today: the only caller I can find across the openmrs org is legacyui's `PersonListItem`, and it goes through the one-argument `DATE` overload. The new tests can't surface it either, because each configured-format case stubs a pattern shaped for the branch it tests (`HH:mm:ss` for `TIME`, `yyyy-MM-dd HH:mm:ss` for `TIMESTAMP`), and the property only ever holds one value at a time. And the acceptance wording on TRUNK-6769 ("honours a configured `GP_SEARCH_DATE_DISPLAY_FORMAT` for each") is mine, so the mismatch is on me rather than on you.

What I'd change here is the `TIME` branch: a date display pattern can never yield a time, so let it use `DateFormat.getTimeInstance(DateFormat.MEDIUM, locale)` unconditionally and drop the lookup, with `formatDate_shouldUseConfiguredFormatForTime` turning into an assertion that the property is ignored there. `TIMESTAMP` needs a date pattern and a time pattern composed together, which is fiddlier, and the ticket already names a candidate home for it (folding `formatDate` onto `org.openmrs.util.Format.format`), so I'd leave that one to the follow-up. I'll update the ticket to match.

waghmaretejas: Thanks for the detailed review and clarification, @dkayiwa. Once the ticket is updated, I’d be happy to work on the updated requirements if needed.

PR #6458, opened 2026-08-14
build(web): configure maven-compiler-plugin to use Java 8

biruksolomon:
### Description
Explicitly configures the `maven-compiler-plugin` in the `web` module `pom.xml` to target Java 8 compatibility. This ensures that bytecode compilation consistently adheres to the Java 8 specification across environments.

### Changes
* Added `maven-compiler-plugin` to `web/pom.xml`.
* Explicitly set `<source>` configuration to `8`.
* Explicitly set `<target>` configuration to `8`.

### Testing Done
- [ ] Validated Maven build execution (`mvn clean install`) completes successfully without compilation warnings related to target versions.


ibacher:
Thank you for the PR, but a couple of notes:

1. As the PR template you ignored tells you, we require a Jira ticket before a PR can be considered.
2. This branch (`master`) does not build on Java 8.

PR #6460, opened 2026-08-16
TRUNK-6679: Restore cache lifetimes lost in the EhCache to Infinispan migration

IamMujuziMoses:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6679

  No? -> execute above command

### DIFF
--- a/api/src/main/resources/cache-api.yaml
+++ b/api/src/main/resources/cache-api.yaml
@@ -2,7 +2,7 @@ caches:
     userSearchLocales:
         configuration: "entity"
     conceptIdsByMapping:
-        configuration: "entity"
+        configuration: "concept-ids-by-mapping"
     serializerWhiteListTypes:
         configuration: "entity"
     rolePrivileges:
--- a/api/src/main/resources/cache-api.yaml
+++ b/api/src/main/resources/cache-api.yaml
@@ -1,6 +1,6 @@
 caches:
     userSearchLocales:
-        configuration: "entity"
+        configuration: "user-search-locales"
     conceptIdsByMapping:
         configuration: "concept-ids-by-mapping"
     serializerWhiteListTypes:
--- a/api/src/main/resources/hibernate.default.properties
+++ b/api/src/main/resources/hibernate.default.properties
@@ -36,54 +36,76 @@ hibernate.cache.region.factory_class=infinispan
 hibernate.cache.default_cache_concurrency_strategy=read-write
 # hibernate.cache.infinispan.cfg is configured by HibernateSessionFactoryBean based on cache_type property
 
-# Preserve the cache lifetimes and maximum sizes used by the 2.6.x Hibernate second-level caches.
+# Reference data is cached indefinitely
 hibernate.cache.infinispan.org.openmrs.Concept.expiration.max_idle=-1
 hibernate.cache.infinispan.org.openmrs.Concept.expiration.lifespan=-1
-hibernate.cache.infinispan.org.openmrs.Concept.eviction.max_entries=10000
+hibernate.cache.infinispan.org.openmrs.Concept.memory.size=10000
 
 hibernate.cache.infinispan.org.openmrs.GlobalProperty.expiration.max_idle=-1
 hibernate.cache.infinispan.org.openmrs.GlobalProperty.expiration.lifespan=-1
-hibernate.cache.infinispan.org.openmrs.GlobalProperty.eviction.max_entries=1000
+hibernate.cache.infinispan.org.openmrs.GlobalProperty.memory.size=1000
 
 hibernate.cache.infinispan.org.openmrs.User.expiration.max_idle=-1
 hibernate.cache.infinispan.org.openmrs.User.expiration.lifespan=-1
-hibernate.cache.infinispan.org.openmrs.User.eviction.max_entries=100
+hibernate.cache.infinispan.org.openmrs.User.memory.size=10000
 
 hibernate.cache.infinispan.org.openmrs.Role.expiration.max_idle=-1
 hibernate.cache.infinispan.org.openmrs.Role.expiration.lifespan=-1
-hibernate.cache.infinispan.org.openmrs.Role.eviction.max_entries=100
+hibernate.cache.infinispan.org.openmrs.Role.memory.size=100
 
 hibernate.cache.infinispan.org.openmrs.Privilege.expiration.max_idle=-1
 hibernate.cache.infinispan.org.openmrs.Privilege.expiration.lifespan=-1
-hibernate.cache.infinispan.org.openmrs.Privilege.eviction.max_entries=500
+hibernate.cache.infinispan.org.openmrs.Privilege.memory.size=500
 
 hibernate.cache.infinispan.org.openmrs.Person.expiration.max_idle=-1
 hibernate.cache.infinispan.org.openmrs.Person.expiration.lifespan=-1
-hibernate.cache.infinispan.org.openmrs.Person.eviction.max_entries=100
+hibernate.cache.infinispan.org.openmrs.Person.memory.size=10000
 
 hibernate.cache.infinispan.org.openmrs.PersonName.expiration.max_idle=-1
 hibernate.cache.infinispan.org.openmrs.PersonName.expiration.lifespan=-1
-hibernate.cache.infinispan.org.openmrs.PersonName.eviction.max_entries=100
+hibernate.cache.infinispan.org.openmrs.PersonName.memory.size=10000
 
 hibernate.cache.infinispan.org.openmrs.PersonAddress.expiration.max_idle=-1
 hibernate.cache.infinispan.org.openmrs.PersonAddress.expiration.lifespan=-1
-hibernate.cache.infinispan.org.openmrs.PersonAddress.eviction.max_entries=100
+hibernate.cache.infinispan.org.openmrs.PersonAddress.memory.size=10000
 
 hibernate.cache.infinispan.org.openmrs.PersonAttribute.expiration.max_idle=-1
 hibernate.cache.infinispan.org.openmrs.PersonAttribute.expiration.lifespan=-1
-hibernate.cache.infinispan.org.openmrs.PersonAttribute.eviction.max_entries=100
+hibernate.cache.infinispan.org.openmrs.PersonAttribute.memory.size=10000
 
 hibernate.cache.infinispan.org.openmrs.ConceptDatatype.expiration.max_idle=-1
 hibernate.cache.infinispan.org.openmrs.ConceptDatatype.expiration.lifespan=-1
-hibernate.cache.infinispan.org.openmrs.ConceptDatatype.eviction.max_entries=100
+hibernate.cache.infinispan.org.openmrs.ConceptDatatype.memory.size=100
 
 hibernate.cache.infinispan.org.openmrs.ConceptClass.expiration.max_idle=-1
 hibernate.cache.infinispan.org.openmrs.ConceptClass.expiration.lifespan=-1
-hibernate.cache.infinispan.org.openmrs.ConceptClass.eviction.max_entries=100
+hibernate.cache.infinispan.org.openmrs.ConceptClass.memory.size=100
 
 hibernate.cache.infinispan.org.openmrs.Location.expiration.max_idle=-1
 hibernate.cache.infinispan.org.openmrs.Location.expiration.lifespan=-1
-hibernate.cache.infinispan.org.openmrs.Location.eviction.max_entries=100
+hibernate.cache.infinispan.org.openmrs.Location.memory.size=10000
+
+hibernate.cache.infinispan.org.openmrs.Person.names.expiration.max_idle=-1
+hibernate.cache.infinispan.org.openmrs.Person.names.expiration.lifespan=-1
+
+hibernate.cache.infinispan.org.openmrs.Person.addresses.expiration.max_idle=-1
+hibernate.cache.infinispan.org.openmrs.Person.addresses.expiration.lifespan=-1
+
+hibernate.cache.infinispan.org.openmrs.Person.attributes.expiration.max_idle=-1
+hibernate.cache.infinispan.org.openmrs.Person.attributes.expiration.lifespan=-1
+
+# Role-related association caches expire after one hour as a safety net.
+hibernate.cache.infinispan.org.openmrs.User.roles.expiration.max_idle=-1
+hibernate.cache.infinispan.org.openmrs.User.roles.expiration.lifespan=3600000
+
+hibernate.cache.infinispan.org.openmrs.Role.privileges.expiration.max_idle=-1
+hibernate.cache.infinispan.org.openmrs.Role.privileges.expiration.lifespan=3600000
+
+hibernate.cache.infinispan.org.openmrs.Role.inheritedRoles.expiration.max_idle=-1
+hibernate.cache.infinispan.org.openmrs.Role.inheritedRoles.expiration.lifespan=3600000
+
+hibernate.cache.infinispan.org.openmrs.Role.childRoles.expiration.max_idle=-1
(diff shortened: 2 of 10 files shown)


IamMujuziMoses:

jwnasambu:
@claude review

api/src/main/resources/hibernate.default.properties
```diff
@@ -36,6 +36,54 @@ hibernate.cache.region.factory_class=infinispan
 hibernate.cache.default_cache_concurrency_strategy=read-write
 # hibernate.cache.infinispan.cfg is configured by HibernateSessionFactoryBean based on cache_type property
 
+# Preserve the cache lifetimes and maximum sizes used by the 2.6.x Hibernate second-level caches.
+hibernate.cache.infinispan.org.openmrs.Concept.expiration.max_idle=-1
+hibernate.cache.infinispan.org.openmrs.Concept.expiration.lifespan=-1
+hibernate.cache.infinispan.org.openmrs.Concept.eviction.max_entries=10000
```
jwnasambu: Worth switching these 12 eviction.max_entries properties to .memory.size Infinispan treats eviction.max_entries as a deprecated alias. At startup InfinispanRegionFactory.parseProperty() logs a WARN per property ("Property '.eviction.max_entries' is deprecated, please use '.memory.size' instead", message id 25029) and then applies it; .memory.size sets the identical memory.maxCount() with no warning. Right now every boot emits 12 of these warnings, and a future Infinispan version could drop the deprecated name and silently fall back to the 10000 default. Same value, same behavior, just the current name:
```suggestion

hibernate.cache.infinispan.org.openmrs.Concept.memory.size=10000

```
Same rename applies to lines 46, 50, 54, 58, 62, 66, 70, 74, 78, 82, and 86.

api/src/main/resources/cache-api.yaml
```diff
@@ -2,7 +2,7 @@ caches:
     userSearchLocales:
         configuration: "entity"
     conceptIdsByMapping:
-        configuration: "entity"
+        configuration: "concept-ids-by-mapping"
```
jwnasambu, line 5: The old ehcache-api.xml also configured userSearchLocales with max 500 entries, 300s idle, and 300s TTL. It still sits on the entity template (100s idle, 10000 entries, no TTL), so its settings were lost in the same migration this PR restores  it just wasn't eternal, so it didn't make the fix list. Is it intentionally out of scope for TRUNK-6679?

If it should be restored, the pattern is already in this PR: add a user-search-locales template to both infinispan-api.xml and infinispan-api-local.xml (max-idle="300000", lifespan="300000", max-count="500") and point line 3 at it. If it's intentional, no code change needed just confirm.

IamMujuziMoses: thank you @jwnasambu for the review, good catch. The current PR is focused on restoring the eternal/reference-cache configurations from 2.6.x, so `userSearchLocales` was intentionally left out because its previous configuration was not eternal. I’ll keep it out of this PR unless `TRUNK-6679` is intended to restore all cache-specific settings lost during the migration. In that case, I’d need the full scope of the issue to make sure we restore all of the affected cache configurations consistently.

api/src/main/resources/hibernate.default.properties
IamMujuziMoses: thank you @jwnasambu for this review, I’ll switch the 12 `eviction.max_entries` properties to `memory.size` and avoid the deprecation warnings at startup.

api/src/main/resources/hibernate.default.properties
```diff
@@ -36,6 +36,54 @@ hibernate.cache.region.factory_class=infinispan
 hibernate.cache.default_cache_concurrency_strategy=read-write
 # hibernate.cache.infinispan.cfg is configured by HibernateSessionFactoryBean based on cache_type property
 
+# Preserve the cache lifetimes and maximum sizes used by the 2.6.x Hibernate second-level caches.
```
dkayiwa: The expiry half of this restores what was lost, but the sizes go the other way. Every second-level region currently runs on the library's `entity` template, which is 10000 entries with a 100 second idle window. I dumped the live Infinispan configuration on master and on this branch: before, every region reports `maxCount=10000`; after, Concept stays at 10000, GlobalProperty drops to 1000, Privilege to 500, and User, Role, Person, PersonName, PersonAddress, PersonAttribute, ConceptDatatype, ConceptClass and Location all drop to 100.

For Role, ConceptDatatype and ConceptClass a cap of 100 is right, those really are fixed reference sets. Person, PersonName, PersonAddress and PersonAttribute are not: their row counts follow the patient census, and User and Location follow the size of the implementation. If this merges as-is, a server that today holds up to 10000 person rows in the second-level cache holds 100, so on any site where more than a hundred distinct patients are in play between two reads of the same one, the region evicts continuously and issues more person queries than it does now. That is the opposite of what TRUNK-6679 is chasing, and it fails quietly: nothing throws, and `HibernateCacheIntegrationTest` asserts these exact numbers, so CI stays green either way.

The 100 comes straight out of the old ehcache.xml, and the ticket only asks for sizes "at least matching" the old maxElementsInMemory, so raising them is inside the scope. I would keep the explicit `memory.size` lines, since inheriting the library default is precisely what the ticket wants to stop doing, and set Person, PersonName, PersonAddress, PersonAttribute, User and Location to 10000, which is what they already get today. Heap is the only argument for a small cap, and with expiry now disabled the LRU bound is what is doing that job anyway. Could we get these raised before this goes in?

api/src/main/resources/hibernate.default.properties
```diff
@@ -36,6 +36,54 @@ hibernate.cache.region.factory_class=infinispan
...
+hibernate.cache.infinispan.org.openmrs.ConceptDatatype.expiration.lifespan=-1
+hibernate.cache.infinispan.org.openmrs.ConceptDatatype.memory.size=100
+
+hibernate.cache.infinispan.org.openmrs.ConceptClass.expiration.max_idle=-1
+hibernate.cache.infinispan.org.openmrs.ConceptClass.expiration.lifespan=-1
+hibernate.cache.infinispan.org.openmrs.ConceptClass.memory.size=100
+
+hibernate.cache.infinispan.org.openmrs.Location.expiration.max_idle=-1
+hibernate.cache.infinispan.org.openmrs.Location.expiration.lifespan=-1
+hibernate.cache.infinispan.org.openmrs.Location.memory.size=100
```
dkayiwa: Nothing in this block reaches the collection regions, and for these entities that is where a lot of the traffic sits. `Person.names`, `Person.addresses` and `Person.attributes` are mapped `lazy="false"`, and `User.roles`, `Role.privileges`, `Role.inheritedRoles` and `Role.childRoles` are eager too, so they load on every Person or User read. They live in their own regions (`org.openmrs.Person.names`, `org.openmrs.User.roles` and so on, not `org.openmrs.PersonName`), and those stay on the library template at 100 seconds idle. So after this change a Person read that misses by 100 seconds still runs collection queries for names, addresses and attributes, and resolving a user's privileges still re-reads user_role, role_role and role_privilege.

2.6.x did not configure them either, they fell through to the ehcache defaultCache at 60 seconds idle, so this is not a regression. But the ticket's stated motivation is reference data no longer being fetched once per JVM lifetime, and closing TRUNK-6679 here leaves that only half true. Was leaving the collection regions out deliberate?

api/src/main/resources/hibernate.default.properties
IamMujuziMoses: Thanks @dkayiwa for this review. I’ve updated the runtime-sized entity regions (Person, PersonName, PersonAddress, PersonAttribute, User, and Location) to retain their existing 10,000 entry capacity.

api/src/main/resources/hibernate.default.properties
IamMujuziMoses: Leaving the collection regions out was not deliberate. I’ve updated the configuration to explicitly set the cached collection regions to eternal expiration.

api/src/main/resources/hibernate.default.properties
```diff
@@ -36,6 +36,76 @@ hibernate.cache.region.factory_class=infinispan
...
+hibernate.cache.infinispan.org.openmrs.Person.attributes.expiration.lifespan=-1
+
+hibernate.cache.infinispan.org.openmrs.User.roles.expiration.max_idle=-1
+hibernate.cache.infinispan.org.openmrs.User.roles.expiration.lifespan=-1
+
+hibernate.cache.infinispan.org.openmrs.Role.privileges.expiration.max_idle=-1
+hibernate.cache.infinispan.org.openmrs.Role.privileges.expiration.lifespan=-1
+
+hibernate.cache.infinispan.org.openmrs.Role.inheritedRoles.expiration.max_idle=-1
+hibernate.cache.infinispan.org.openmrs.Role.inheritedRoles.expiration.lifespan=-1
```
dkayiwa: Making the role graph eternal removes the only thing that made the `role-privileges` lifespan do anything. `RolePrivilegeCache` recomputes a role's flattened privileges with `session.get(Role.class, name)` and then walks `getPrivileges()` and `getInheritedRoles()`, so that path reads `org.openmrs.Role`, `org.openmrs.Role.privileges` and `org.openmrs.Role.inheritedRoles`. On master all three carry the library's 100 second idle window, so a recompute more than 100 seconds after the last read goes back to the database. With all three at `-1` the recompute is served entirely from Infinispan and issues no SQL at all. I dumped the live Infinispan configuration on both branches to be sure: before, the three report `maxIdle=100000`; after, they report `maxIdle=-1 lifespan=-1`.

If this merges as is, a privilege granted or revoked outside this JVM's Hibernate, by direct SQL or by a second app node that is not in the same Infinispan cluster (`cache.type=local` is the default), is never picked up until the server restarts, where today it is picked up within about 100 seconds of the next role read. Nothing throws and no test notices, because `HibernateCacheIntegrationTest` asserts the `-1`. The comment above the `role-privileges` template in `infinispan-api-local.xml` says that cache's one hour TTL is there as a safety net for exactly this case, and `RolePrivilegeCache`'s javadoc repeats it, so after this change the config and the comments say opposite things. `org.openmrs.User.roles` has the same shape for role assignments.

What I would do is keep `expiration.max_idle=-1` on `Role`, `Role.privileges`, `Role.inheritedRoles`, `User` and `User.roles`, since that is what ends the 100 second re-fetch TRUNK-6679 is about, and set their `expiration.lifespan` to `3600000` instead of `-1` so they line up with the `role-privileges` TTL. That costs one role read per role per hour and keeps the documented behaviour true. Leaving them eternal is defensible too, 2.6.x was eternal, but then the safety net wording has to come out of both XML files and out of `RolePrivilegeCache`. One of the two needs doing before this merges, since as it stands they contradict each other.

api/src/main/resources/hibernate.default.properties
```diff
@@ -36,6 +36,76 @@ hibernate.cache.region.factory_class=infinispan
 hibernate.cache.default_cache_concurrency_strategy=read-write
 # hibernate.cache.infinispan.cfg is configured by HibernateSessionFactoryBean based on cache_type property
 
+# Preserve the cache lifetimes and maximum sizes used by the 2.6.x Hibernate second-level caches.
```
dkayiwa: This header no longer matches the sizes underneath it. The 2.6.x `ehcache.xml` had `maxElementsInMemory="100"` for `User`, `Person`, `PersonName`, `PersonAddress`, `PersonAttribute` and `Location`, and those six now read 10000 on purpose so they keep the capacity master already gives them. Whoever next cross-checks these against the deleted `ehcache.xml` will find 100 and quietly put them back, which is the regression we just avoided, so it is worth saying that the expiry is the 2.6.x setting and the sizes deliberately are not.

```suggestion
# Restore the eternal expiry the 2.6.x Hibernate second-level caches had (ehcache.xml, eternal="true").
# Sizes are the 2.6.x maxElementsInMemory values, except for the regions whose row counts follow the
# patient census or the size of the implementation (User, Person, PersonName, PersonAddress,
# PersonAttribute, Location). Those stay at 10000, the capacity they already have, rather than
# dropping back to the old 100.
```

api/src/main/resources/hibernate.default.properties
IamMujuziMoses: I agree, I’ve updated the configuration and tests accordingly.

api/src/main/resources/infinispan-api-local.xml
```diff
@@ -29,6 +29,20 @@
 			<expiration lifespan="3600000" max-idle="100000" interval="5000"/>
 			<memory max-count="10000"/>
 		</local-cache-configuration>
+		
+		<local-cache-configuration name="concept-ids-by-mapping" simple-cache="true" statistics="false">
+			<encoding media-type="application/x-java-object"/>
+			<transaction mode="NONE"/>
+			<expiration max-idle="-1" interval="5000"/>
```
dkayiwa, line 36: Dropping the idle window here is what makes an existing eviction gap permanent. `ConceptServiceImpl.purgeConcept` is the one path that removes concept mappings without a `@CacheEvict` for `conceptIdsByMapping`; `saveConcept`, `saveConceptSource`, `purgeConceptSource`, `saveConceptReferenceTerm`, `purgeConceptReferenceTerm` and `updateConceptIndexes` all carry one.

I ran the same probe on both branches: save a concept mapped to a fresh reference term, call `getConceptIdsByMapping("PURGEPROBE", "SSTRM", true)` (returns `[5498]`), purge the concept, call again, still `[5498]`. `getConceptsByMapping` then hands back a one element list whose only element is null, because it feeds every cached id through `getConcept(id)` with no null check, and `getConceptByMapping` sees a non-empty list and returns `concepts.get(0)`, so a lookup by that code quietly resolves to nothing. Where a code maps to two concepts and the purged one sorts second, `concepts.get(1).getRetired()` throws NPE instead.

That part is true on master as well. What changes is how long it lasts: master's `max-idle="100000"` clears the entry at the next hundred second lull so it heals itself, whereas with `max-idle="-1"` nothing clears it and the wrong answer survives until the server restarts. Nothing throws on the common path and no test notices.

Since the point of this change is that these entries live for the JVM's lifetime, the eviction has to cover every path that can invalidate them. One annotation closes it:

```java
@Override
@CacheEvict(value = CONCEPT_IDS_BY_MAPPING_CACHE_NAME, allEntries = true)
public void purgeConcept(Concept concept) throws APIException {
```

I would rather this went in with the PR than after it, since this is the change that takes the window away.

api/src/main/resources/hibernate.default.properties
```diff
@@ -36,6 +36,80 @@ hibernate.cache.region.factory_class=infinispan
 hibernate.cache.default_cache_concurrency_strategy=read-write
 # hibernate.cache.infinispan.cfg is configured by HibernateSessionFactoryBean based on cache_type property
 
+# Restore the 2.6.x Hibernate second-level cache expiry settings where applicable. Most reference-data regions are
+# eternal. Role-related association caches use a one-hour lifespan as a safety net for changes made outside the
+# normal API/cache invalidation path.
+#
+# Cache sizes use the 2.6.x maxElementsInMemory values except for regions whose capacity is intentionally larger on
+# current master. Those capacities are retained rather than reduced to the old 2.6.x values.
```
ibacher: Please remove this comment or make it relevant. Comments like this that document _why we're making this change in this PR_ are unhelpful cruft that we should not merge in. Helpful comments might, e.g., give an overview of what the rules are or why we want an "eternal cache" for certain elements.

IamMujuziMoses: Makes sense. I have removed the PR-specific context and keep only brief comments describing the cache rules.

api/src/main/resources/infinispan-api-local.xml
IamMujuziMoses: Thanks, I’ve added the `@CacheEvict` to purgeConcept in this PR and include the regression test to cover the cache invalidation path.

PR #6461, opened 2026-08-16
TRUNK-6764: Remove unused xercesImpl dependency

waghmaretejas:

Removed the unused `xerces:xercesImpl` dependency from OpenMRS Core.

* Removed `xerces:xercesImpl` from `api/pom.xml`.
* Removed the unused `xercesVersion` property from `bom/pom.xml`.
* Removed the `xercesImpl` dependency entry from the BOM.
* Removed the `xerces:xercesImpl` entry from `NOTICE.md`.

During review, @dkayiwa recommended adding a safeguard to prevent `xerces:xercesImpl` from being reintroduced transitively in the future. Although this was not explicitly requested in the issue, I implemented the recommendation by adding a Maven Enforcer `bannedDependencies` rule to the existing Enforcer configuration.

Before removing the dependency, I verified that no Java source imports `org.apache.xerces` and that the relevant XML parsing paths use JAXP APIs.

I also inspected the module configuration, Liquibase changelog, and HL7-related XML parsing paths to verify that they do not directly depend on the Xerces implementation.

The existing tests were left unchanged.

The `org.apache.xerces` entry in `checkstyle.xml` was left unchanged because it is an `IllegalImport` rule preventing Xerces imports, not a dependency declaration.

## Issue I worked on

See https://issues.openmrs.org/browse/TRUNK-6764

## Verification

* `mvn dependency:tree -Dincludes=xerces:xercesImpl` — **BUILD SUCCESS**; `xercesImpl` no longer appears in any module dependency tree.
* `ModuleFileParserUnitTest` — **BUILD SUCCESS**.
* `DatabaseUpdaterDatabaseIT` — **BUILD SUCCESS**.
* `./mvnw clean package` — **BUILD SUCCESS**.
* `mvn clean verify -DskipTests` — **BUILD SUCCESS**.
* `git diff --check` — clean.

* [x] My IDE is configured to follow the [**[code style](https://wiki.openmrs.org/display/docs/Java+Conventions)**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

* [ ] I have **added tests** to cover my changes. (If you refactored

* [x] I ran `./mvnw clean package` and added all formatting changes to my commit.

* [x] All new and existing **tests passed**.

* [x] My pull request is **based on the latest changes** of the master branch.


jwnasambu:
@claude review

jwnasambu (approved):

jwnasambu (commented):
@dkayiwa, @ibacher No rush, just want that extra check before this goes out. Thanks!

api/pom.xml
```diff
@@ -192,16 +192,6 @@
 			<groupId>org.reflections</groupId>
 			<artifactId>reflections</artifactId>
 		</dependency>
-		<dependency>
-			<groupId>xerces</groupId>
-			<artifactId>xercesImpl</artifactId>
```
dkayiwa, line 197: Worth folding the matching `NOTICE.md` delete into this PR: line 44 still lists Xerces (`xerces:xercesImpl`) under Apache License 2.0, so once this merges the notice file claims OpenMRS Core ships a library it no longer ships.

I realise the issue text only names the two poms, but that file is hand-maintained and exists to track what is actually on the classpath (#5909 was literally "Update NOTICE.md to reflect current dependencies"), and the last dependency removal did both in one commit: 86dfe6ec3e (#5770) dropped `jackson-core-asl` and `jackson-mapper-asl` from `api/pom.xml` and their NOTICE.md rows together. One line to delete here.

bom/pom.xml
```diff
@@ -483,11 +482,6 @@
 				<artifactId>jaxen</artifactId>
 				<version>${jaxenVersion}</version>
 			</dependency>
-			<dependency>
-				<groupId>xerces</groupId>
-				<artifactId>xercesImpl</artifactId>
-				<version>${xercesVersion}</version>
```
dkayiwa, line 489: Now that the managed version disappears along with the declaration, it might be worth adding a `bannedDependencies` rule to the enforcer config in the root pom so this cannot quietly come back. `maven-enforcer-plugin` is already wired up there with an `enforce` execution carrying only `requireMavenVersion`, so it is a few lines inside the existing `<rules>`. Not a blocker for this PR.

The reason I would bother is that this jar changes behaviour purely by sitting on the classpath, with nothing in the source referring to it. xercesImpl ships `META-INF/services` entries for `javax.xml.parsers.DocumentBuilderFactory`, `javax.xml.parsers.SAXParserFactory`, `javax.xml.validation.SchemaFactory`, `javax.xml.datatype.DatatypeFactory`, `javax.xml.stream.XMLEventFactory` and `org.xml.sax.driver`, so on master all six resolve to Xerces rather than the JDK parser for everything running in the web app, and after this PR all six resolve to the JDK. If a future library drags xercesImpl back in transitively it flips all six again, at whatever version that path declares now that the 2.12.2 pin here is gone, and nothing in the build would report it: the `org.apache.xerces` entry in `checkstyle.xml` is an `IllegalImport` rule, so it only catches source imports, and no test asserts which parser is in use.

This is also the second time the jar has been removed. TRUNK-4789 took it out in 2016 and a6de8262 put it straight back the same day "to get rid of error logs from liquibase". That no longer reproduces (`DatabaseUpdaterDatabaseIT` produces identical output with and without the jar on Liquibase 4.33), so there is nothing to fix, but it is the kind of drift a build rule would catch.

api/pom.xml
waghmaretejas: Thanks for pointing this out! Addressed by removing the `xerces:xercesImpl` entry from `NOTICE.md`, so it now accurately reflects the dependencies shipped by OpenMRS Core.

bom/pom.xml
waghmaretejas: Thanks for the suggestion, @dkayiwa! I implemented this safeguard by adding a `bannedDependencies` rule to the existing Maven Enforcer configuration to prevent `xerces:xercesImpl` from being reintroduced transitively.

I also verified the change with `mvn clean verify -DskipTests`, and the build passes successfully.

waghmaretejas:
Hi @jwnasambu, thanks for helping revive PR #6461! I’ve addressed the review feedback and pushed the updates. Please take a look when you get a chance.

bom/pom.xml
ibacher: > I also verified the change

Note that running with `-DskipTests` kind of doesn't help verify an issue that would only show up in tests... That setup basically confirms that things compile, but we knew that would be the case  because the Xerces classes aren't directly referenced in our code.

ibacher (approved):

PR #6470, opened 2026-08-18
TRUNK-6777: Cache the tokenization of LayoutTemplates

ibacher:

We use LayoutTemplates to store how to format peoples names and addresses since our data model allows for a flexible set of fields with instance- determined meaning. These templates are called frequently, e.g., whenever we render a "full" or "short" name. In a single-threaded run, the run is dominated by the call to `getLines()` which on each invocation reparses the template. By cacheing this tokenization, we greatly speed up (~20x) the amount of time it takes to format a name. Since these templates are _very_ stable, this should be a net win on every operation that formats a user's name.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6777

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/layout/LayoutSupport.java
+++ b/api/src/main/java/org/openmrs/layout/LayoutSupport.java
@@ -35,15 +35,6 @@ public abstract class LayoutSupport<T extends LayoutTemplate> {
 	protected List<String> specialTokens;
 
 	/**
-	 * @return the current special token configuration version, for use by
-	 *         {@link LayoutTemplate#getLines()} in deciding whether its cache is still valid
-	 * @since 2.8.10
-	 */
-	public int getConfigurationVersion() {
-		return configurationVersion.get();
-	}
-
-	/**
 	 * @return Returns the layoutTemplates.
 	 */
 	public List<T> getLayoutTemplates() {
@@ -93,7 +84,7 @@ public abstract class LayoutSupport<T extends LayoutTemplate> {
 			for (T at : this.layoutTemplates) {
 				if (at != null && templateName.equalsIgnoreCase(at.getCodeName())) {
 					ret = at;
-					log.debug("Found Layout Template named '{}'", at.getDisplayName());
+					log.debug("Found Layout Template named " + at.getDisplayName());
 				}
 			}
 
@@ -111,7 +102,7 @@ public abstract class LayoutSupport<T extends LayoutTemplate> {
 			for (T at : this.layoutTemplates) {
 				if (at != null && templateName.equalsIgnoreCase(at.getCountry())) {
 					ret = at;
-					log.debug("Found Layout Template named '{}'", at.getDisplayName());
+					log.debug("Found Layout Template named " + at.getDisplayName());
 				}
 			}
 
@@ -129,7 +120,7 @@ public abstract class LayoutSupport<T extends LayoutTemplate> {
 			for (T at : this.layoutTemplates) {
 				if (at != null && templateName.equalsIgnoreCase(at.getDisplayName())) {
 					ret = at;
-					log.debug("Found Layout Template named '{}'", at.getDisplayName());
+					log.debug("Found Layout Template named " + at.getDisplayName());
 				}
 			}
 
@@ -167,4 +158,12 @@ public abstract class LayoutSupport<T extends LayoutTemplate> {
 		this.defaultLayoutFormat = defaultLayoutFormat;
 	}
 
+	/**
+	 * @return the current special token configuration version, for use by
+	 *         {@link LayoutTemplate#getLines()} in deciding whether its cache is still valid
+	 * @since 2.8.10
+	 */
+	int getConfigurationVersion() {
+		return configurationVersion.get();
+	}
 }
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/openmrs/layout/LayoutTemplate.java
```diff
@@ -379,4 +401,23 @@ public List<String> nonUniqueStringsGoLast(List<String> strListArg) {
...
 	}
 
+	/**
+	 * Immutable snapshot of everything derived from a single tokenization pass. The support instance
+	 * and its configuration version at the time of the pass are carried along so that a reader can tell
+	 * whether the snapshot still reflects the current special tokens without having to resolve the
+	 * support singleton itself, which for {@link org.openmrs.layout.address.AddressTemplate} means
+	 * entering a synchronized block.
+	 */
+		private record TokenizedLines(List<List<Map<String, String>>> lines, int maxTokens, LayoutSupport<?> support,
```
ibacher: Obviously for backporting purposes, this will need to be re-extended into a class.

ibacher:
@claude review

api/src/main/java/org/openmrs/layout/LayoutSupport.java
```diff
@@ -77,7 +93,7 @@ public T getLayoutTemplateByCodeName(String templateName) {
 			for (T at : this.layoutTemplates) {
 				if (at != null && templateName.equalsIgnoreCase(at.getCodeName())) {
 					ret = at;
-					log.debug("Found Layout Template named " + at.getDisplayName());
+					log.debug("Found Layout Template named '{}'", at.getDisplayName());
```
dkayiwa: These three sites moved to parameterized logging, but `getLayoutTemplateByName` just above still concatenates on both of its log calls, and one of them is the identical "Found Layout Template named" message, so the file now carries two styles of the same line. TRUNK-6777 doesn't need any of this, and it puts three hunks in front of a reviewer that have nothing to do with the tokenization cache. Could you drop the logging edits from this PR? If they're worth making, a follow-up that converts all of the call sites would be tidier.

api/src/main/java/org/openmrs/layout/LayoutSupport.java
```diff
@@ -21,12 +22,27 @@ public abstract class LayoutSupport<T extends LayoutTemplate> {
...
 	protected List<T> layoutTemplates;
 
 	protected List<String> specialTokens;
 
+	/**
+	 * @return the current special token configuration version, for use by
+	 *         {@link LayoutTemplate#getLines()} in deciding whether its cache is still valid
+	 * @since 2.8.10
+	 */
+	public int getConfigurationVersion() {
```
dkayiwa: Does this need to be public? The only caller is `TokenizedLines.isValid()` in `LayoutTemplate`, which is in this same package, and the javadoc says as much. Public means a permanently supported method on a class modules extend, carried on every branch this gets backported to, exposing a counter callers can neither move (nothing but `setSpecialTokens` bumps it) nor act on. Package-private covers the one caller.

api/src/main/java/org/openmrs/layout/LayoutTemplate.java
```diff
@@ -94,99 +102,114 @@ private List<Map<String, String>> convertToTokens(String line, String[] nonToken
...
+			LayoutSupport<?> support = null;
+			int configurationVersion = 0;
+
+			if (this.lineByLineFormat != null) {
+				support = getLayoutSupportInstance();
+				// read the version before the tokens it describes: stamping a snapshot with a version read
+				// afterwards could label tokens that are already stale as current
+				configurationVersion = support.getConfigurationVersion();
+				List<String> specialTokens = nonUniqueStringsGoLast(support.getSpecialTokens());
+				ret = new ArrayList<>(this.lineByLineFormat.size());
```
dkayiwa: Was this shape change intended? With `lineByLineFormat` set to an empty but non-null list, master left `ret` null and so `getLines()` returned null; allocating the list up front makes it return `[]`, and this branch now resolves the support where master never got as far as `replaceTokens`. I ran the old and new tokenizers side by side over the five bundled name templates, the default address template and a batch of awkward inputs, and this was the only output difference, so the exposure is narrow. What it does change is `NameTemplate.format()` on such a template: master threw `APIException: Unable to format personName ... using name template`, this returns `""`, so `PersonName.getFullName()` renders a blank name rather than failing loudly. I would keep the empty list, it is the less destructive of the two, but it is worth making that a decision rather than a side effect of the restructure.

dkayiwa:
> @claude review

We ran out of credits. 😊

api/src/main/java/org/openmrs/layout/LayoutTemplate.java
ibacher: Yes, this is intentional, both because the failure condition didn't seem sensible, but also because it makes the overall caching slightly saner if the value is not null.

api/src/main/java/org/openmrs/layout/LayoutSupport.java
ibacher: Addressed. Made this package-private.

api/src/main/java/org/openmrs/layout/LayoutSupport.java
ibacher: Reasonable. Tbh, this is just me clicking on a link in my IDE and then adding quotes.

PR #6474, opened 2026-08-18
TRUNK-6780: Add a MariaDB option to the test container harness

ibacher:

This PR adds two things:

1. Support for running the integration tests against MariaDB as well as MySQL 5.7
2. It fixes two queries that relied on MySQL's implicit ordering by making the order there explicit 

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6780

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateDiagnosisDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateDiagnosisDAO.java
@@ -87,7 +87,7 @@ public class HibernateDiagnosisDAO implements DiagnosisDAO {
 		Query query = sessionFactory.getCurrentSession().createQuery(
 			"from Diagnosis d where d.patient.patientId = :patientId and d.voided = false " 
 				+ fromDateCriteria  
-				+ " order by d.dateCreated desc");
+				+ " order by d.dateCreated desc, d.diagnosisId desc");
 		query.setParameter("patientId", patient.getId());
 		if(fromDate != null){
 			query.setParameter("fromDate", fromDate);
@@ -107,7 +107,7 @@ public class HibernateDiagnosisDAO implements DiagnosisDAO {
 		if (confirmedOnly) {
 			queryString += " and d.certainty = :certainty";
 		}
-		queryString += " order by d.dateCreated desc";
+		queryString += " order by d.dateCreated desc, d.diagnosisId desc";
 
 		TypedQuery<Diagnosis> query = sessionFactory.getCurrentSession().createQuery(queryString, Diagnosis.class).setParameter("encounterId", encounter.getId());
 		if (primaryOnly) {
@@ -132,7 +132,7 @@ public class HibernateDiagnosisDAO implements DiagnosisDAO {
 		if (confirmedOnly) {
 			queryString += " and d.certainty = :certainty";
 		}
-		queryString += " order by d.dateCreated desc";
+		queryString += " order by d.dateCreated desc, d.diagnosisId desc";
 
 		TypedQuery<Diagnosis> query = sessionFactory.getCurrentSession().createQuery(queryString, Diagnosis.class).setParameter("visitId", visit.getId());
 		if (primaryOnly) {
--- a/api/src/test/java/org/openmrs/test/Containers.java
+++ b/api/src/test/java/org/openmrs/test/Containers.java
@@ -15,12 +15,14 @@ import java.util.Properties;
 import org.hibernate.dialect.MySQLDialect;
 import org.hibernate.dialect.PostgreSQL82Dialect;
 import org.openmrs.api.context.Context;
+import org.testcontainers.containers.MariaDBContainer;
 import org.testcontainers.containers.MySQLContainer;
 import org.testcontainers.containers.PostgreSQLContainer;
 
 public class Containers {
 
 	private static MySQLContainer<?> mysql;
+	private static MariaDBContainer<?> mariadb;
 	private static PostgreSQLContainer<?> postgres;
 	
 	private static final String USERNAME =  "test";
@@ -30,12 +32,15 @@ public class Containers {
 	
 	public static void ensureDatabaseRunning() {
 		
-		if (mysql != null || postgres != null) {
+		if (mysql != null || mariadb != null || postgres != null) {
 			return;
 		}
-		
+
 		if ("postgres".equals(System.getProperty("database"))) {
 			ensurePostgreSQLRunning();
+		}
+		else if ("mariadb".equals(System.getProperty("database"))) {
+			ensureMariaDBRunning();
 		}
 		else {
 			ensureMySQLRunning();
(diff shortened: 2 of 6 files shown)


api/src/main/java/org/openmrs/api/db/hibernate/HibernateDiagnosisDAO.java
```diff
@@ -87,7 +87,7 @@ public List<Diagnosis> getActiveDiagnoses(Patient patient, Date fromDate) {
 		Query query = sessionFactory.getCurrentSession().createQuery(
 			"from Diagnosis d where d.patient.patientId = :patientId and d.voided = false " 
 				+ fromDateCriteria  
-				+ " order by d.dateCreated desc");
+				+ " order by d.dateCreated desc, d.diagnosisId desc");
```
dkayiwa, line 90: Nothing exercises the new tie-break on this query. The only pair in `DiagnosisServiceImplTest-SetupDiagnosis.xml` that shares a `date_created` is diagnoses 2 and 4 (both `2015-01-12 00:00:00`), and 4 is voided, so `getActiveDiagnoses` never returns both and the result is `1, 3, 2` before and after.

It matters more here than on the other two queries, because `getUniqueDiagnoses` runs `removeIf` over this list and keeps the first occurrence of each `CodedOrFreeText`, so on a tie the id order decides which duplicate survives. One extra non-voided row sharing a `date_created` with an existing active diagnosis would pin both down.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateDiagnosisDAO.java
```diff
@@ -107,7 +107,7 @@ public List<Diagnosis> getDiagnosesByEncounter(Encounter encounter, boolean prim
 		if (confirmedOnly) {
 			queryString += " and d.certainty = :certainty";
 		}
-		queryString += " order by d.dateCreated desc";
+		queryString += " order by d.dateCreated desc, d.diagnosisId desc";
```
dkayiwa, line 110: `date_created` is `datetime(0)`, so diagnoses saved together in one encounter almost always land on the same second. That makes this second key the effective order for `getDiagnosesByEncounter` and `getDiagnosesByVisit` rather than a rare tie-break, and with `desc` they come back in reverse entry order.

Was that the intent? `asc` on the id is equally deterministic and keeps them in the order they were recorded.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateDiagnosisDAO.java
```diff
@@ -132,7 +132,7 @@ public List<Diagnosis> getDiagnosesByVisit(Visit visit, boolean primaryOnly, boo
...
```
dkayiwa, line 135: Is a counterpart for master and 2.9.x planned? Both still order all three of these queries by `dateCreated` alone, so upgrading off 2.8.x puts the non-determinism back, and master's `getActiveDiagnoses` has been rewritten to the Criteria API since, so it won't cherry-pick.

api/src/test/java/org/openmrs/test/Containers.java
```diff
@@ -76,6 +81,45 @@ private static void ensureMySQLRunning() {
...
+
+            mariadb.start();
+
+            System.setProperty("databaseUrl", mariadb.getJdbcUrl());
+    		System.setProperty("databaseName", DATABASE);
+    		System.setProperty("databaseUsername", USERNAME);
+    		System.setProperty("databasePassword", PASSWORD);
+    		System.setProperty("databaseDriver", "com.mysql.jdbc.Driver");
+    		System.setProperty("databaseDialect", MySQLDialect.class.getName());
+			System.setProperty("database", "mysql");
```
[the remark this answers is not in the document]
dkayiwa: The mismatch is real but it doesn't fail: `mariadb-java-client` is a runtime dependency of `openmrs-api`, so `DriverManager` skips the MySQL driver (`acceptsURL` is false, `connect` returns null) and hands the `jdbc:mariadb://` URL to `org.mariadb.jdbc.Driver`. Neither consumer uses the class name to connect: `DatabaseUpdater.getConnection()` does `Class.forName(driver)` and then `DriverManager.getConnection(url, ...)`, and Hibernate 5.6's `C3P0ConnectionProvider` only calls `classForName` on `hibernate.connection.driver_class` before passing the URL to `DataSources.unpooledDataSource`. I ran the full api suite with `-Ddatabase=mariadb -DuseInMemoryDatabase=false` and it is green, 5134 tests.

What the setting does cost is the guarantee in the javadoc just above it, which says the MySQL driver is used so that Liquibase resolves the platform as mysql rather than mariadb. It resolves as mariadb, and the second option suggested here (keep the MySQL driver, switch the URL to `jdbc:mysql:`) doesn't change that. I pointed Connector/J 8.0.30 at the same `mariadb:10.11.7` container over a `jdbc:mysql://` URL and Liquibase still picked `MariaDBDatabase`, because `isCorrectDatabaseImplementation` also matches product name MySQL when the version string contains "mariadb", and this server reports `5.5.5-10.11.7-MariaDB-1:10.11.7+maria~ubu2204`. So mysql mode isn't reachable against a MariaDB server at all.

Nothing is being skipped today, since the changelogs this harness runs (`schema-only-2.7.x` and `update-to-latest-2.8.x`) carry no `dbms="mysql"` gates, only postgresql ones. But the javadoc reads as a guarantee, so whoever adds a mysql-gated changeset later will believe it ran here. Worth taking the driver from `mariadb.getDriverClassName()`, which also drops the "Loading class `com.mysql.jdbc.Driver'. This is deprecated." line from every run, and rewriting that sentence to describe the `database` to `mysql` override instead. That one is genuinely load-bearing: `DatabaseIT.dropAllDatabaseObjects()` only drops and recreates on the "mysql" value and has no mariadb branch, so without it the drop would quietly do nothing.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateDiagnosisDAO.java
ibacher: So, in the interest of full disclosure, this is split out from a larger piece of work that was actually focused on restoring support for PostgreSQL (I'm down a bit of a rabbit hole; initially I thought I would just ticket out the stuff in RefApp modules that needs to be updated to support PostgreSQL, but it turns out there was some stuff in Core that needs to be fixed too).

api/src/main/java/org/openmrs/api/db/hibernate/HibernateDiagnosisDAO.java
ibacher: Yes. I agree `asc` is equally deterministic, but if `date_created` had greater precision so that such rows were differentiated, `desc` here matches the order they would be returned in, so it seemed most consistent with the query as-written.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateDiagnosisDAO.java
ibacher: Yes, it is. I started from 2.8 because this came out of findings from a test harness targeting the RefApp, but the intention is to roll this forward to 2.9 and `master`.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateDiagnosisDAO.java
ibacher: PR with the other stuff (including the parts that failed this test) will be coming.

api/src/test/java/org/openmrs/test/Containers.java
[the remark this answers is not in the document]
ibacher: I've gone ahead and swapped the driver anyway as it's a trivial change.

PR #6475, opened 2026-08-18
TRUNK-6775: Break circular dependency between core utilities and HL7

IrinaEkb:

This change breaks the circular dependency between core utilities and HL7 support.

The following HL7-owned global property definitions were moved from `OpenmrsConstants` to `HL7Constants`:

- `GLOBAL_PROPERTY_HL7_ARCHIVE_DIRECTORY`
- `GLOBAL_PROPERTY_IGNORE_MISSING_NONLOCAL_PATIENTS`

Their registration as global properties was also moved from `OpenmrsConstants` to the HL7 side.

This removes the core dependency on HL7-owned constants and makes the dependency direction one-way: from HL7 to core.

## Issue I worked on

TRUNK-6775

### DIFF
--- a/api/src/main/java/org/openmrs/hl7/HL7Constants.java
+++ b/api/src/main/java/org/openmrs/hl7/HL7Constants.java
@@ -56,6 +56,8 @@ public class HL7Constants {
 
 	/**
 	 * Used in hl7 sextuplets: 123^Primary name^99DCT^345^Chosen name^99NAM
+	 * These are  the constants used by FormUtil
+	 * FOR TRUNK-6775
 	 */
 	public static final String HL7_LOCAL_CONCEPT = "99DCT";
 
@@ -79,8 +81,16 @@ public class HL7Constants {
 	 * default name for HL7_archives destination directory
 	 *
 	 * @since 1.7
+	 * 
+	 * FOR TRUNK-6775
 	 */
 	public static final String HL7_ARCHIVE_DIRECTORY_NAME = "hl7_archives";
+
+	public static final String GLOBAL_PROPERTY_HL7_ARCHIVE_DIRECTORY =
+		"hl7_archive.dir";
+
+	public static final String GLOBAL_PROPERTY_IGNORE_MISSING_NONLOCAL_PATIENTS =
+		"hl7_processor.ignore_missing_patient_non_local";
 
 	/**
 	 * @since 1.10
--- a/api/src/main/java/org/openmrs/util/FormUtil.java
+++ b/api/src/main/java/org/openmrs/util/FormUtil.java
@@ -26,7 +26,7 @@ import org.openmrs.ConceptName;
 import org.openmrs.Drug;
 import org.openmrs.Form;
 import org.openmrs.FormField;
-import org.openmrs.hl7.HL7Constants;
+
 
 /**
  * OpenMRS utilities related to forms.
@@ -220,7 +220,7 @@ public class FormUtil {
 	 * @return String representation of the given concept
 	 */
 	public static String conceptToString(Concept concept, ConceptName localizedName) {
-		return concept.getConceptId() + "^" + localizedName.getName() + "^" + HL7Constants.HL7_LOCAL_CONCEPT; // + "^"
+		return concept.getConceptId() + "^" + localizedName.getName() + "^" + "99DCT"; // + "^"
 	}
 
 	/**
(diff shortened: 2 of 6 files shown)


api/src/main/java/org/openmrs/util/databasechange/MoveDeletedHL7sChangeSet.java
```diff
@@ -46,7 +46,7 @@ public void execute(Database database) throws CustomChangeException {
 		insertHL7Sql.append("INSERT INTO hl7_in_queue");
 		insertHL7Sql.append(" (hl7_source, hl7_source_key, hl7_data, date_created, uuid, message_state)");
 		insertHL7Sql.append(" VALUES (?, ?, ?, ?, ?, ");
-		insertHL7Sql.append(HL7Constants.HL7_STATUS_DELETED);
+		insertHL7Sql.append(4); //deleted (HL7Constants.HL7_STATUS_DELETED);
```
jwnasambu, line 49: This change replaces `HL7Constants.HL7_STATUS_DELETED` with a hardcoded `4`. This doesn't appear related to breaking the circular dependency between core utilities and HL7. The constant is still available in `HL7Constants` and there's no circular dependency issue with this particular constant. Was this intentional?

api/src/main/java/org/openmrs/util/OpenmrsConstants.java
```diff
@@ -249,7 +249,8 @@ public static final Collection<String> AUTO_ROLES() {
 
 	public static final String GLOBAL_PROPERTY_USER_REQUIRE_EMAIL_AS_USERNAME = "user.requireEmailAsUsername";
 
-	public static final String GLOBAL_PROPERTY_HL7_ARCHIVE_DIRECTORY = "hl7_archive.dir";
+	// FOR TRUNK-6775
```
jwnasambu, line 252: The old constants are commented out rather than deleted. For a clean refactor, these should be removed entirely. Commented-out code adds noise. If there's a backwards compatibility concern for modules referencing these constants, that should be documented.

api/src/main/java/org/openmrs/hl7/HL7Constants.java
```diff
@@ -56,6 +56,8 @@ private HL7Constants() {
 
 	/**
 	 * Used in hl7 sextuplets: 123^Primary name^99DCT^345^Chosen name^99NAM
+	 * These are  the constants used by FormUtil
+	 * FOR TRUNK-6775
```
jwnasambu, line 60: The comment "FOR TRUNK-6775" should not be in production code. Javadoc should describe what the constant is for in general terms, not reference specific internal refactoring tickets.

api/src/main/java/org/openmrs/hl7/HL7Constants.java
```diff
@@ -79,9 +81,17 @@ private HL7Constants() {
 	 * default name for HL7_archives destination directory
 	 *
 	 * @since 1.7
+	 * 
+	 * FOR TRUNK-6775
```
jwnasambu: The comment "FOR TRUNK-6775" should not be in production code. Also, there's an unnecessary extra blank line here that should be removed.

api/src/main/java/org/openmrs/util/FormUtil.java
```diff
@@ -220,7 +220,7 @@ public static String conceptToString(Concept concept, Locale locale) {
 	 * @return String representation of the given concept
 	 */
 	public static String conceptToString(Concept concept, ConceptName localizedName) {
-		return concept.getConceptId() + "^" + localizedName.getName() + "^" + HL7Constants.HL7_LOCAL_CONCEPT; // + "^"
+		return concept.getConceptId() + "^" + localizedName.getName() + "^" + "99DCT"; // + "^"
```
jwnasambu, line 223: Replaced `HL7Constants.HL7_LOCAL_CONCEPT` with the hardcoded string `"99DCT"`. This loses the semantic meaning of the constant. Since `FormUtil` is in `org.openmrs.util` and `HL7Constants` is in `org.openmrs.hl7`, the original import created the circular dependency concern. A better approach might be to define a local constant in `FormUtil` (e.g., `LOCAL_CONCEPT_CODE = "99DCT"`) with a comment explaining what it means, rather than scattering magic strings.

IrinaEkb:
Thank you for the review and feedback! Unfortunately, I don't have enough free time to finish this ticket right now, so I'm closing this PR. Anyone else is welcome to take over

PR #6478, opened 2026-08-20
TRUNK-6696: Cache PrivilegeListener lookup instead of scanning Spring contexts on every privilege check

IamMujuziMoses:

The new implementation makes listener lookups use cached lists while invalidating the caches whenever the application context changes, preserving discovery of listeners contributed by modules.

Listener lookups previously called `getBeansOfType()` repeatedly, requiring Spring to search the application and module contexts even when the listener registrations had not changed. This is particularly expensive for `PrivilegeListener`, which is involved in frequent privilege checks.

### Changes

* Added caching for `PrivilegeListener` and `UserSessionListener` in `UserContext`.
* Added cache invalidation on `ContextRefreshedEvent`.
* Added caching for `HttpSessionListener` in the web-layer listener.
* Added web-layer cache invalidation on context refresh.
* Cleared listener caches during shutdown.
* Added tests for caching, empty results, immutability, and cache invalidation.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6696

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/cache/RegisteredListenerCache.java
+++ b/api/src/main/java/org/openmrs/api/cache/RegisteredListenerCache.java
@@ -1,0 +1,31 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.cache;
+
+import org.openmrs.api.context.ServiceContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextRefreshedEvent;
+import org.springframework.stereotype.Component;
+
+@Component
+public class RegisteredListenerCache implements ApplicationListener<ContextRefreshedEvent> {
+
+	private static final Logger log = LoggerFactory.getLogger(RegisteredListenerCache.class);
+
+	@Override
+	public void onApplicationEvent(ContextRefreshedEvent event) {
+		if (ServiceContext.isInstantiated()) {
+			log.debug("Clearing registered listener cache after context refresh");
+			ServiceContext.getInstance().clearCachedListeners();
+		}
+	}
+}
--- a/api/src/main/java/org/openmrs/api/cache/RegisteredListenerCache.java
+++ b/api/src/main/java/org/openmrs/api/cache/RegisteredListenerCache.java
@@ -9,13 +9,19 @@
  */
 package org.openmrs.api.cache;
 
-import org.openmrs.api.context.ServiceContext;
+import org.openmrs.api.context.UserContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.context.ApplicationListener;
 import org.springframework.context.event.ContextRefreshedEvent;
 import org.springframework.stereotype.Component;
 
+/**
+ * Clears the cached listener registrations whenever the application context is refreshed, ensuring
+ * listeners contributed by the refreshed context are not served from a stale cache.
+ *
+ * @since 2.8.10
+ */
 @Component
 public class RegisteredListenerCache implements ApplicationListener<ContextRefreshedEvent> {
 
@@ -23,9 +29,7 @@ public class RegisteredListenerCache implements ApplicationListener<ContextRefre
 
 	@Override
 	public void onApplicationEvent(ContextRefreshedEvent event) {
-		if (ServiceContext.isInstantiated()) {
-			log.debug("Clearing registered listener cache after context refresh");
-			ServiceContext.getInstance().clearCachedListeners();
-		}
+		log.debug("Clearing registered listener cache after context refresh");
+		UserContext.clearCachedListeners();
 	}
 }
--- a/api/src/main/java/org/openmrs/api/context/ServiceContext.java
+++ b/api/src/main/java/org/openmrs/api/context/ServiceContext.java
@@ -20,6 +20,8 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.aopalliance.aop.Advice;
+import org.openmrs.PrivilegeListener;
+import org.openmrs.UserSessionListener;
 import org.openmrs.api.APIException;
 import org.openmrs.api.AdministrationService;
 import org.openmrs.api.CohortService;
@@ -128,6 +130,17 @@ public class ServiceContext implements ApplicationContextAware {
 	}
 
 	/**
+	 * Listener types whose registered components are cached because they are looked up on hot paths.
+	 */
+	private static final Set<Class<?>> CACHED_LISTENER_TYPES = Set.of(PrivilegeListener.class, UserSessionListener.class);
+
+	/**
+	 * Cached registered listener components, replaced atomically whenever the cache is updated or
+	 * cleared.
+	 */
+	private volatile Map<Class<?>, List<?>> cachedListeners = Collections.emptyMap();
+
+	/**
 	 * There should only be one ServiceContext per openmrs (java virtual machine). This method should be
 	 * used when wanting to fetch the service context Note: The ServiceContext shouldn't be used
 	 * independently. All calls should go through the Context
@@ -187,6 +200,7 @@ public class ServiceContext implements ApplicationContextAware {
 
 		if (ServiceContextHolder.instance != null) {
 			ServiceContextHolder.instance.applicationContext = null;
+			ServiceContextHolder.instance.clearCachedListeners();
 
 			if (ServiceContextHolder.instance.moduleOpenmrsServices != null) {
 				ServiceContextHolder.instance.moduleOpenmrsServices.clear();
@@ -887,11 +901,50 @@ public class ServiceContext implements ApplicationContextAware {
 	 * @return a List of all registered Beans that are valid instances of the passed type
 	 * @since 1.5
 	 */
-
 	public <T> List<T> getRegisteredComponents(Class<T> type) {
+		if (CACHED_LISTENER_TYPES.contains(type)) {
+			return getCachedListeners(type);
+		}
+
 		Map<String, T> m = getRegisteredComponents(applicationContext, type);
 		log.trace("getRegisteredComponents({}) = {}", type, m);
 		return new ArrayList<>(m.values());
+	}
+
+	/**
+	 * Clears cached listener registrations so that listeners from the current application and module
+	 * contexts are discovered again after a context refresh.
+	 *
+	 * @since 3.0.0
+	 */
+	public void clearCachedListeners() {
+		cachedListeners = Collections.emptyMap();
+	}
+
+	@SuppressWarnings("unchecked")
+	private <T> List<T> getCachedListeners(Class<T> type) {
+		List<?> cached = cachedListeners.get(type);
+		if (cached != null) {
+			return (List<T>) cached;
+		}
+
+		synchronized (this) {
+			cached = cachedListeners.get(type);
+			if (cached != null) {
+				return (List<T>) cached;
+			}
+
+			Map<String, T> components = getRegisteredComponents(applicationContext, type);
+			List<T> listeners = Collections.unmodifiableList(new ArrayList<>(components.values()));
+
+			Map<Class<?>, List<?>> newCachedListeners = new HashMap<>(cachedListeners);
+			newCachedListeners.put(type, listeners);
+			cachedListeners = Collections.unmodifiableMap(newCachedListeners);
+
+			log.trace("Cached registered listeners of type {}: {}", type, listeners);
+
+			return listeners;
+		}
 	}
 
 	/**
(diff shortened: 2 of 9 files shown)


IamMujuziMoses:
@dkayiwa @ibacher @jwnasambu could you please review this PR, your feedback will be appreciated.

api/src/main/java/org/openmrs/api/context/ServiceContext.java
```diff
@@ -187,6 +200,7 @@ public static void destroyInstance() {
 
 		if (ServiceContextHolder.instance != null) {
 			ServiceContextHolder.instance.applicationContext = null;
+			ServiceContextHolder.instance.clearCachedListeners();
```
jwnasambu:
```suggestion
ServiceContextHolder.instance.clearCachedListeners();
ServiceContextHolder.instance.applicationContext = null;;
```
The current order nulls applicationContext first, then clears the cache. A thread concurrently inside getCachedListeners at line 937 (Map<String, T> components = getRegisteredComponents(applicationContext, type)) would hit a NullPointerException when trying to traverse the now-null context hierarchy. Clearing the cache first ensures no new lookups start before the context is removed.

web/src/test/java/org/openmrs/web/ListenerTest.java
```diff
@@ -0,0 +1,77 @@
...
+	 */
+	@Test
+	public void onApplicationEvent_shouldClearHttpSessionListeners() {
+		RegisteredHttpSessionListenerCache cache = applicationContext.getBean(RegisteredHttpSessionListenerCache.class);
+
+		cache.onApplicationEvent(new ContextRefreshedEvent(applicationContext));
+
+		// The event handler should invalidate the Listener cache.
+		assertNotNull(cache);
+	}
```
jwnasambu, line 139: Kindly assertNotNull(cache) only checks that the @Component bean exists. It does not verify that clearHttpSessionListeners() was actually called. The test will pass even if RegisteredHttpSessionListenerCache.onApplicationEvent is empty.

```suggestion
	@Test
public void onApplicationEvent_shouldClearHttpSessionListeners() {
    List<HttpSessionListener> first = listener.getHttpSessionListeners();
    assertFalse(first.isEmpty());

    RegisteredHttpSessionListenerCache cache = applicationContext.getBean(RegisteredHttpSessionListenerCache.class);
    cache.onApplicationEvent(new ContextRefreshedEvent(applicationContext));

    List<HttpSessionListener> second = listener.getHttpSessionListeners();
    assertNotSame(first, second);
}
```

api/src/main/java/org/openmrs/api/cache/RegisteredListenerCache.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.openmrs.api.context.ServiceContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextRefreshedEvent;
+import org.springframework.stereotype.Component;
+
+@Component
+public class RegisteredListenerCache implements ApplicationListener<ContextRefreshedEvent> {
```
ibacher, line 26: Why do we need this component? Why not just make the `ServiceContext` itself an `ApplicationListener`?

api/src/main/java/org/openmrs/api/context/ServiceContext.java
```diff
@@ -887,13 +901,52 @@ public boolean isRefreshingContext() {
...
 		Map<String, T> m = getRegisteredComponents(applicationContext, type);
 		log.trace("getRegisteredComponents({}) = {}", type, m);
 		return new ArrayList<>(m.values());
 	}
 
+	/**
+	 * Clears cached listener registrations so that listeners from the current application and module
+	 * contexts are discovered again after a context refresh.
+	 *
+	 * @since 3.0.0
```
ibacher: Intention is to backport this.

```suggestion
	 * @since 2.8.10
```

api/src/main/java/org/openmrs/api/context/ServiceContext.java
```diff
@@ -127,6 +129,17 @@ private ServiceContextHolder() {
...
+	/**
+	 * Listener types whose registered components are cached because they are looked up on hot paths.
+	 */
+	private static final Set<Class<?>> CACHED_LISTENER_TYPES = Set.of(PrivilegeListener.class, UserSessionListener.class);
+
+	/**
+	 * Cached registered listener components, replaced atomically whenever the cache is updated or
+	 * cleared.
+	 */
+	private volatile Map<Class<?>, List<?>> cachedListeners = Collections.emptyMap();
```
ibacher: `Collections.emptyMap()` should almost never be used as an initial value. See [the JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#emptyMap--) for what that actually does...

api/src/main/java/org/openmrs/api/context/ServiceContext.java
ibacher, line 1: I don't think that implementing this on the `ServiceContext` is the right thing to do _at all_. It's only in the `UserContext` that we currently reference these `PrivilegeListener`s and `UserSessionListener`s and _that's_ where the cache belongs. By caching things here, you add unnecessary steps to most uses of `getRegisteredComponents()` and spread things around quite a bit.

ibacher (changes requested):
Thanks @IamMujuziMoses! This needs some major refactoring I've tried to outline in the comments.

api/src/main/java/org/openmrs/api/cache/RegisteredListenerCache.java
IamMujuziMoses: Thanks @ibacher for the reviews, I’ve moved the listener caches out of `ServiceContext` and into `UserContext`, so I kept this as a separate component to invalidate those caches on context refresh rather than making `ServiceContext` responsible for `UserContext`-specific behavior.

api/src/main/java/org/openmrs/api/context/ServiceContext.java
IamMujuziMoses: Right thanks. I’ve removed the `Collections.emptyMap()` initialization and changed the cache to use `null` to represent an uninitialized cache.

api/src/main/java/org/openmrs/api/context/ServiceContext.java
IamMujuziMoses: Right. I’ve moved the listener-specific caching out of `ServiceContext` and into `UserContext`. `getRegisteredComponents()` is now left as a general-purpose lookup without the additional listener-specific caching logic.

web/src/test/java/org/openmrs/web/ListenerTest.java
```diff
@@ -0,0 +1,119 @@
...
+	 * @see Listener#getHttpSessionListeners()
+	 */
+	@Test
+	public void getHttpSessionListeners_shouldCacheEmptyListeners() {
+		List<HttpSessionListener> first = listener.getHttpSessionListeners();
+		List<HttpSessionListener> second = listener.getHttpSessionListeners();
+
+		assertTrue(first.isEmpty());
+		assertSame(first, second);
+	}
```
dkayiwa, line 89: Neither of these two ever reaches the cache. Nothing sets `openmrsStarted` before they run, so both calls return on the `if (!openmrsStarted)` early path and `assertSame(first, second)` ends up comparing `Collections.emptyList()` with itself. I put a probe in rather than guessing: inside both tests `Listener.isOpenmrsStarted()` is false and the list handed back is the `Collections.emptyList()` singleton.

That leaves the `Listener` half of the change with nothing behind it. I deleted the caching out of `getHttpSessionListeners()`, left everything else in the PR alone, and all four tests in this class still passed, so nothing here would notice if the caching disappeared.

The setter you already added is enough to fix it. I ran the version below both ways: it passes against the `getHttpSessionListeners()` in this PR and fails against an uncached one.

```suggestion
	@Test
	public void getHttpSessionListeners_shouldCacheListeners() {
		try (MockedStatic<Context> contextMock = Mockito.mockStatic(Context.class)) {
			contextMock.when(() -> Context.getRegisteredComponents(HttpSessionListener.class))
			        .thenReturn(Collections.singletonList(Mockito.mock(HttpSessionListener.class)));

			Listener.setOpenmrsStarted(true);

			List<HttpSessionListener> first = listener.getHttpSessionListeners();
			List<HttpSessionListener> second = listener.getHttpSessionListeners();

			assertSame(first, second);
			contextMock.verify(() -> Context.getRegisteredComponents(HttpSessionListener.class), times(1));
		} finally {
			Listener.setOpenmrsStarted(false);
		}
	}

	/**
	 * @see Listener#getHttpSessionListeners()
	 */
	@Test
	public void getHttpSessionListeners_shouldCacheEmptyListeners() {
		try (MockedStatic<Context> contextMock = Mockito.mockStatic(Context.class)) {
			contextMock.when(() -> Context.getRegisteredComponents(HttpSessionListener.class))
			        .thenReturn(Collections.emptyList());

			Listener.setOpenmrsStarted(true);

			List<HttpSessionListener> first = listener.getHttpSessionListeners();
			List<HttpSessionListener> second = listener.getHttpSessionListeners();

			assertTrue(first.isEmpty());
			assertSame(first, second);
			contextMock.verify(() -> Context.getRegisteredComponents(HttpSessionListener.class), times(1));
		} finally {
			Listener.setOpenmrsStarted(false);
		}
	}
```

api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -681,7 +702,21 @@ protected Integer getDefaultLocationId(User user) {
...
+			synchronized (UserContext.class) {
+				listeners = privilegeListeners;
+
+				if (listeners == null) {
+					listeners = Collections
+					        .unmodifiableList(new ArrayList<>(Context.getRegisteredComponents(PrivilegeListener.class)));
+					privilegeListeners = listeners;
+				}
+			}
+		}
```
dkayiwa, line 710: The double checked locking is doing less here than it looks. The field is volatile and the value is idempotent, so two threads racing on a cold cache would each build the same list and the second write would be harmless. That is exactly the reasoning behind the lock-free version you wrote in `Listener.getHttpSessionListeners()`.

What it does cost is that a cold-cache privilege check now holds a JVM-wide monitor across a Spring bean factory call. `getBeansOfType(Class)` compiles down to `getBeansOfType(type, true, true)` and then calls `getBean(name, type)` per match (I read the bytecode in spring-beans 7.0.9), so it can instantiate beans while `UserContext.class` is held. That is the reverse of the lock order taken by any bean initialization that performs a privilege check, and privilege checking is the path this ticket is about.

Your call, but I would drop the synchronized block here and in `notifyUserSessionListener` below. The `new ArrayList<>(...)` copy can go with it, since `ServiceContext.getRegisteredComponents` already hands back a fresh list, which is why `Listener` wraps the result directly. I applied the change below to both methods and the api tests pass.

```suggestion
		List<PrivilegeListener> listeners = privilegeListeners;

		if (listeners == null) {
			listeners = Collections.unmodifiableList(Context.getRegisteredComponents(PrivilegeListener.class));
			privilegeListeners = listeners;
		}
```

api/src/main/java/org/openmrs/api/context/ServiceContext.java
```diff
@@ -887,7 +887,6 @@ public boolean isRefreshingContext() {
 	 * @return a List of all registered Beans that are valid instances of the passed type
 	 * @since 1.5
 	 */
-
 	public <T> List<T> getRegisteredComponents(Class<T> type) {
```
dkayiwa, line 891: This blank line, and the one added at the end of `ServiceContextTest`, are all that survives of the earlier `ServiceContext` version of the cache. Both files now show up in the diff without changing any behavior, which drags `ServiceContext` back into review for nothing. Could you drop both, so the PR only touches the files it actually changes?

```suggestion

	public <T> List<T> getRegisteredComponents(Class<T> type) {
```

web/src/main/java/org/openmrs/web/RegisteredHttpSessionListenerCache.java
```diff
@@ -0,0 +1,23 @@
...
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.web;
+
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextRefreshedEvent;
+import org.springframework.stereotype.Component;
+
+@Component
+public class RegisteredHttpSessionListenerCache implements ApplicationListener<ContextRefreshedEvent> {
```
dkayiwa, line 23: This class has no Javadoc at all, while `RegisteredListenerCache` at least documents its handler. Two lines on why it exists would save the next reader a trip through `Listener`: that it clears the static `HttpSessionListener` cache when a module start or stop refreshes the context, and that it lives in `web` rather than next to `RegisteredListenerCache` because `Listener` is not reachable from the api module. That second half is the answer you gave to the question on the other component, and it is worth having in the code rather than only in the PR discussion. An `@since` would be good too, given this is heading for 2.8.10.

web/src/test/java/org/openmrs/web/ListenerTest.java
IamMujuziMoses: Thanks @dkayiwa for the reviews, . I’ve updated both tests and applied the suggestions.

api/src/main/java/org/openmrs/api/context/UserContext.java
IamMujuziMoses: Right. I’ve removed the synchronized blocks and applied the suggestions.

web/src/main/java/org/openmrs/web/RegisteredHttpSessionListenerCache.java
IamMujuziMoses: Right thanks, I’ve added class-level `Javadoc` explaining that it clears the `HttpSessionListener` cache after a context refresh.

PR #6496, opened 2026-08-25
TRUNK-6690: Share deduplication work between patient-search count and list paths

waghmaretejas:
TRUNK-6690: Share deduplication work between patient-search count and list paths

Added a combined patient search API that returns both the requested page of patients and the total count from a single shared deduplication traversal.

The existing getPatients(...) and getCountOfPatients(...) methods are preserved for backward compatibility. A new getPatientsAndCount(...) method was added to the PatientDAO and PatientService APIs and implemented in HibernatePatientDAO.

The shared search logic collects both the unique patient keys needed for the requested page and the total unique count during the same traversal, avoiding separate deduplication scans for the count and list paths.

This pull request implements the OpenMRS Core API changes for this issue. The corresponding update to use the new combined API in the openmrs-module-webservices.rest repository will be submitted separately after the Core changes are reviewed and finalized.

Added tests to verify that:

- The combined search returns the same page results as the existing getPatients(...) path.
- The combined search returns the same total count as the existing getCountOfPatients(...) path.
- Duplicate hits across search sub-queries are handled correctly.
- The shared search performs the expected query execution without repeating the deduplication traversal.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6690


waghmaretejas:
Hi @jwnasambu, could you please review this PR when you have a chance? Thank you!

api/src/main/java/org/openmrs/api/db/hibernate/search/SearchQueryUnique.java
```diff
@@ -314,12 +326,18 @@ public static <T> SearchUniqueResults<T> search(SearchSessionFactory searchSessi
 			// The count path only needs the total number of distinct hits, so it skips fetching and
 			// hydrating a page of results entirely.
 			long totalHitCount = searchTotalHitCount(searchSession, uniqueQuery, UNBOUNDED_DEDUPLICATION);
-			return new SearchUniqueResults<>(new ArrayList<>(), offset, limit, totalHitCount);
+			return new SearchUniqueResults<>(new ArrayList<>(), offset, limit, totalHitCount, true);
```
jwnasambu, line 329:
```suggestion
// searchTotalHitCount already knows whether the cap was exceeded, but that
// information is lost here. Thread it through so the caller can distinguish
// an exact count from a raw upper bound.
boolean exact = totalHitCount <= UNBOUNDED_DEDUPLICATION; // or pass exceededCap out of searchTotalHitCount
return new SearchUniqueResults<>(new ArrayList<>(), offset, limit, totalHitCount, exact);
```

searchTotalHitCount internally calls collectUniqueKeys, which sets exceededCap = true when distinct hits exceed the deduplication cap. When exceeded, it falls back to rawHitCount a non-deduplicated upper bound. But the caller at line 329 always passes true for the exactness flag, so the REST layer (and any future caller) is told the count is exact when it may not be. This means getCountOfPatients can return a count that overstates the true distinct patient count and the client has no way to detect it. The new searchWithResultsAndCount at line 460 correctly uses !exceededCap, so this bug is not inherited by the new method but the legacy count path used by getCountOfPatients remains broken.

api/src/test/java/org/openmrs/api/db/hibernate/search/SearchQueryUniqueTest.java
```diff
@@ -255,6 +255,72 @@ public void collectUniqueKeys_shouldScanEverythingWhenBelowCap() {
...
+		createJoinedFixture();
+		SearchQueryUnique.SearchUniqueResults<Person> expectedPage = SearchQueryUnique.search(searchSessionFactory,
+		    joinedQuery(), 1, 5, false);
+		SearchQueryUnique.SearchUniqueResults<Person> actual = SearchQueryUnique
+		        .searchWithResultsAndCount(searchSessionFactory, joinedQuery(), 1, 5, BOTH);
+		Long expectedCount = SearchQueryUnique.searchCount(searchSessionFactory, joinedQuery(), BOTH);
+
+		assertEquals(personIds(expectedPage.getResults()), personIds(actual.getResults()));
+		assertEquals(expectedCount, actual.getTotalHitCount());
+	}
```
jwnasambu, line 324: Kindly add one assertion after line 321:
   
```suggestion
	assertEquals(expectedCount, actual.getTotalHitCount());
    assertFalse(actual.getTotalHitCountExact(),
        "count should be flagged as approximate when deduplication cap is exceeded");
	}
```
The test verifies the count value matches the raw fallback, but never checks that getTotalHitCountExact() returns false. If someone accidentally hardcodes the exactness flag as the pre-existing bug at line 329 does, this test would still pass. The assertion pins the contract when the cap is exceeded, the result must signal that the count is not exact.

api/src/test/java/org/openmrs/api/db/PatientDAOTest.java
```diff
@@ -2562,4 +2562,30 @@ public void getPatients_shouldFindPatientsEfficiently() throws IOException, URIS
...
+		assertEquals(legacyCount, combined.getTotalElements().longValue());
+		assertTrue(combined.isTotalElementsExact());
+
+		List<Patient> combinedList = combined.getList();
+		assertEquals(legacyPatients.size(), combinedList.size());
+
+		for (int i = 0; i < legacyPatients.size(); i++) {
+			assertEquals(legacyPatients.get(i), combinedList.get(i));
+		}
+	}
```
jwnasambu, line 2590: Kindly add two new test methods after line 2590:
```suggestion
	@Test
public void getPatientsAndCount_shouldExcludeVoidedPatientsWhenIncludeVoidedIsFalse() {
    // "Meriadoc" matches patient 60 (voided patient with voided name "Meriadoc Brandybuck")
    String query = "Meriadoc";
    org.openmrs.collection.ListPart<Patient> combined = dao.getPatientsAndCount(query, false, 0, 10);
    assertEquals(0L, combined.getTotalElements().longValue(),
        "voided patients must be excluded when includeVoided is false");
    assertTrue(combined.getList().isEmpty());
}

@Test
public void getPatientsAndCount_shouldIncludeVoidedPatientsWhenIncludeVoidedIsTrue() {
    String query = "Meriadoc";
    org.openmrs.collection.ListPart<Patient> combined = dao.getPatientsAndCount(query, true, 0, 10);
    assertEquals(1L, combined.getTotalElements().longValue(),
        "voided patient must be included when includeVoided is true");
    assertEquals(1, combined.getList().size());
}
```

The only existing DAO test hardcodes includeVoided = false with query "Ben", which matches no voided patients. The includeVoided parameter flows into the Hibernate Search predicate via newPatientIdentifierSearchPredicate, getPatientNameQuery, and getPatientAttributeQuery. If a future change accidentally ignores or inverts this parameter, the current test suite would not catch it. Patient 60 (Meriadoc Brandybuck) is a voided patient with a voided name, loaded from HibernatePatientDAOTest-patients.xml:71–72, and is the standard test fixture for voided-patient scenarios in this test class (see getPatients_shouldNotMatchVoidedPatients at line 682).

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -759,6 +759,49 @@ public Long getCountOfPatients(String query, boolean includeVoided) {
...
+	        throws DAOException {
+		return getPatientsAndCount(query, false, start, length);
+	}
+
+	@Override
+	public org.openmrs.collection.ListPart<Patient> getPatientsAndCount(String query, boolean includeVoided, Integer start,
+	        Integer length) throws DAOException {
+		if (StringUtils.isBlank(query)) {
+			return org.openmrs.collection.ListPart.newListPart(new ArrayList<>(), start == null ? 0L : start.longValue(),
+			    length == null ? 0L : length.longValue(), 0L, true);
```
jwnasambu:
```suggestion
		if (StringUtils.isBlank(query)) {
    long normalizedStart = start == null ? 0L : start.longValue();
    long normalizedLength = length == null ? 0L : length.longValue();
    return org.openmrs.collection.ListPart.newListPart(new ArrayList<>(), normalizedStart,
        normalizedLength, 0L, true);
}
```
 The inline ternaries work correctly but extracting them mirrors the normalization pattern used later in lines 776–783 and makes the blank query return consistent with the rest of the method. The existing getCountOfPatients at line 742 uses the same inline pattern so this is consistent with current style flagging only for awareness not as a required change

jwnasambu:
Kindly @dkayiwa, @ibacher, @wikumChamith Kindly feel free to review this PR at your convenient time please!

waghmaretejas:
Hi @dkayiwa, @ibacher

I’ll work on addressing the issues identified in this PR, including:

* Reducing the Cognitive Complexity of `SearchQueryUnique` from 57 to the allowed 15.
* Reducing the Cognitive Complexity of `collectDuplicateIdsAndUniqueKeys` from 39 to the allowed 15.
* Refactoring the code to remove the `scan` label and the need for it.
* Merging the nested `if` statement with its enclosing condition.
* Fixing the Javadocs issue identified in the PR.
* Addressing the code suggestion in `HibernatePatientDAO` on line 771.
* Reviewing the SonarCloud findings across the PR to make sure all the reported issues are addressed appropriately.

Apart from these points, I’d appreciate your thoughts on the overall changes and whether there is anything else you think I should look into.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -759,6 +759,49 @@ public Long getCountOfPatients(String query, boolean includeVoided) {
...
+		            f -> newPatientIdentifierSearchPredicate(f, query, includeVoided, false), "patient.personId",
+		            this::multiLoadPatients)
+		        .join(SearchQueryUnique
+		                .newProjectedQuery(PersonName.class, f -> personQuery.getPatientNameQuery(f, query, includeVoided),
+		                    "person.personId", this::multiLoadPatients)
+		                .join(SearchQueryUnique.newProjectedQuery(PersonAttribute.class,
+		                    f -> personQuery.getPatientAttributeQuery(f, query, includeVoided), "person.personId",
+		                    this::multiLoadPatients)));
+		SearchQueryUnique.SearchUniqueResults<Patient> result = SearchQueryUnique.searchWithResultsAndCount(
+		    searchSessionFactory, patientQuery, start, length, HibernatePersonDAO.getMaximumSearchResults());
```
dkayiwa: This needs fixing before merge. The search here gets the raw `start` and `length`, so the `tmpStart`/`tmpLength` computed just above on lines 776-784 only ever reach the `ListPart` metadata and never the query itself.

Two consequences. With a null `length` nothing bounds the search at all. I set `personSearchMaxResults` to 2 against 6 matching patients: `getPatients(q, false, 0, null)` returned 2, and `getPatientsAndCount(q, false, 0, null)` returned all 6 and hydrated every one of them, while the `ListPart` it handed back reported `maxElements` of 2. The deduplication scroll is unbounded in that case too, since `maxNewUniqueKeys` is null and only the duplicate-id cap can freeze the page. On a real database that is every patient matching the query pulled into the session, which is the load `personSearchMaxResults` exists to bound, and a null length is a normal way to call the sibling method (`PatientServiceImpl.getPatients(String)` does exactly that on line 518).

Second, a negative `start` reaches Lucene instead of being clamped. `getPatientsAndCount(q, false, -1, 5)` throws `NullPointerException: Cannot read field "doc" because "scoreDocs[...]" is null`, where `getPatients(q, false, -1, 5)` returns the first page.

Passing `tmpStart, tmpLength` here fixes both, once `tmpStart` also clamps a negative value the way `getPatients` does on line 226. One side effect worth choosing deliberately rather than inheriting: `tmpLength` also caps an explicitly oversized `length` at `personSearchMaxResults`, which `getPatients` does not currently do. I think that is the better behaviour, but it is a real difference. A test comparing `getPatientsAndCount(q, includeVoided, 0, null)` against `getPatients(q, includeVoided, 0, null)` would pin the whole thing.

`findPatients` on line 846 carries the same unused `tmpStart`/`tmpLength` locals, which is probably where the pattern came from, but there `getPatients` normalises before calling in, so nothing depends on them.

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -759,6 +759,49 @@ public Long getCountOfPatients(String query, boolean includeVoided) {
...
+		return getPatientsAndCount(query, false, start, length);
+	}
+
+	@Override
+	public org.openmrs.collection.ListPart<Patient> getPatientsAndCount(String query, boolean includeVoided, Integer start,
+	        Integer length) throws DAOException {
+		if (StringUtils.isBlank(query)) {
+			return org.openmrs.collection.ListPart.newListPart(new ArrayList<>(), start == null ? 0L : start.longValue(),
+			    length == null ? 0L : length.longValue(), 0L, true);
+		}
```
dkayiwa: The `minSearchCharacters` guard is missing here, and it should go in before this ships. `findPatients` applies it on line 865, so `getPatients` refuses to search on a query shorter than the configured minimum; `getPatientsAndCount` goes straight to the index. With `minSearchCharacters` set to 20 and a 9 character query, `getPatients(q, false, 0, 20)` returned nothing and `getPatientsAndCount(q, false, 0, 20)` returned all 6 matches.

That bites as soon as the REST side lands. `PatientResource1_8.doSearch` builds `new ServiceSearcher<Patient>(PatientService.class, "getPatients", "getCountOfPatients")`, and webservices.rest has no `minSearchCharacters` check of its own, so core's list path is the only thing enforcing it. Point that resource at `getPatientsAndCount` and `GET /ws/rest/v1/patient?q=a` goes from an empty `results` array to a full page of patients on any installation that raised the minimum.

`getCountOfPatients` never applied the guard either, so only the returned list changes. My preference is to apply the same check `findPatients` does and return an empty page with a total of 0, which is what "not enough characters typed to search yet" actually means. Returning an empty page with the real count would match the legacy pair exactly, but it carries the existing inconsistency forward instead of fixing it.

api/src/main/java/org/openmrs/api/db/hibernate/search/SearchQueryUnique.java
```diff
@@ -333,6 +351,115 @@ public static <T> SearchUniqueResults<T> search(SearchSessionFactory searchSessi
 	 * clauses do not affect scoring; do not add an explicit sort to one without the other, or the bound
 	 * no longer holds.
 	 */
+	/**
+	 * Fetches the requested page and deduplicated total hit count while sharing the same deduplication
+	 * scan for each joined query.
+	 */
+	private static <T> SearchUniqueResults<T> searchWithResultsAndCount(SearchSession searchSession,
```
dkayiwa, line 353: Not blocking, but worth settling while the refactor you have planned is still open, because this is `searchWithResults` copied. Diffing the two bodies, the substantive differences are the count accumulators, the removed early return when the page is full, and `collectDuplicateIds` becoming `collectDuplicateIdsAndUniqueKeys`. The remaining ninety lines are identical, minus almost every explanatory comment from the original: why `uniqueKeys` has to stay a `Set` rather than a `List`, why only the unique-key branch has to consider projection, and why `consumed` is the right offset to carry to the next sub-query.

The doc block on lines 341-353 belonged to `searchWithResults`. It now sits above this method, so `searchWithResults` is undocumented and this method carries a description that is false of it, since its scroll deliberately does not stop once `offset + limit` unique keys have been found. The sort warning in that block, "do not add an explicit sort to one without the other, or the bound no longer holds", is now nowhere near the method it was written for.

The cost is two copies of the pagination and deduplication algorithm that have to stay in step, where a divergence surfaces as wrong search results rather than as an error. Extracting the per sub-query step into one helper both loops call, or giving a single loop an `includeCount` flag (the count accumulators are inert when it is off, and the early return becomes conditional on it), removes the duplication and brings the cognitive complexity down at the same time, which seems better than reducing the complexity of a second copy.

api/src/main/java/org/openmrs/api/PatientService.java
```diff
@@ -499,6 +499,20 @@ public PatientIdentifierType unretirePatientIdentifierType(PatientIdentifierType
 	@Authorized({ PrivilegeConstants.GET_PATIENTS })
 	public List<Patient> getPatients(String query, boolean includeVoided, Integer start, Integer length) throws APIException;
 
+	/**
+	 * Returns the patient page and the deduplicated total count in one combined search.
+	 */
+	@Authorized({ PrivilegeConstants.GET_PATIENTS })
+	public org.openmrs.collection.ListPart<Patient> getPatientsAndCount(String query, Integer start, Integer length)
```
dkayiwa: Could these pick up the usual `@since`, `@param` and `@return` tags? The `getPatients` overload just above carries `@since 1.11`, and module authors read that tag to work out what they can call against a given platform version. Separately, `org.openmrs.collection.ListPart` is spelled out in full in all four files that mention it; an import would read more like the rest of these files.

ibacher:
> Reducing the Cognitive Complexity of SearchQueryUnique from 57 to the allowed 15.
> Reducing the Cognitive Complexity of collectDuplicateIdsAndUniqueKeys from 39 to the allowed 15.

Don't worry about these too much, really. Some code winds up complex and Sonar is pretty aggressive about it. I'd actually strongly prefer that we not refactor things to obey this rule.

waghmaretejas:
> > Reducing the Cognitive Complexity of SearchQueryUnique from 57 to the allowed 15.
> > Reducing the Cognitive Complexity of collectDuplicateIdsAndUniqueKeys from 39 to the allowed 15.
> 
> Don't worry about these too much, really. Some code winds up complex and Sonar is pretty aggressive about it. I'd actually strongly prefer that we not refactor things to obey this rule.



> > Reducing the Cognitive Complexity of SearchQueryUnique from 57 to the allowed 15.
> > Reducing the Cognitive Complexity of collectDuplicateIdsAndUniqueKeys from 39 to the allowed 15.
> 
> Don't worry about these too much, really. Some code winds up complex and Sonar is pretty aggressive about it. I'd actually strongly prefer that we not refactor things to obey this rule.

@ibacher , Thanks for clarifying! You guessed it right—I was actually stuck on that part. This helps a lot!

waghmaretejas:
Hi @dkayiwa and @ibacher,

Thanks for the detailed review and suggestions. I’ve gone through the comments and addressed the requested changes:

 * Updated `getPatientsAndCount()` to normalize `start` and `length` before the search and pass the normalized values to `searchWithResultsAndCount()`.
* Added the `minSearchCharacters` check so the combined search behaves consistently with `getPatients()`.
* Added regression tests covering minimum search characters, `null` length, and negative `start`.
* Added the assertion verifying that the count is marked as approximate when the deduplication cap is exceeded.
* Added the requested Javadocs and imported `ListPart`.
* Applied the two small Sonar improvements by removing the `scan` label and simplifying the nested condition.

I also looked into the suggestion of sharing the deduplication logic between `searchWithResults()` and `searchWithResultsAndCount()`. I agree that sharing the logic would reduce duplication and make it easier to keep the two paths in sync. For this PR, however, I’ve kept the existing `searchWithResults()` path unchanged because refactoring that established path would increase the scope of the change and could introduce unintended behavioral changes in an already working path. Since this was marked as non-blocking, I thought it would be safer to avoid that additional risk here while keeping the suggestion in mind for a separate cleanup.

For the Cognitive Complexity warnings, I’ve avoided a larger refactor specifically to satisfy the Sonar rule, following the guidance that we should not refactor solely to satisfy the Sonar rule.

api/src/main/java/org/openmrs/api/PatientService.java
```diff
@@ -499,6 +500,34 @@ public PatientIdentifierType unretirePatientIdentifierType(PatientIdentifierType
...
 
+	/**
+	 * Returns the patient page and the deduplicated total count in one combined search.
+	 *
+	 * @param query the search query
+	 * @param start the starting index
+	 * @param length the maximum number of patients to return
+	 * @return the patient page and deduplicated total count
+	 * @throws APIException
+	 * @since 1.11
```
dkayiwa: These should say `3.0.0` rather than `1.11`, here and again on line 525. `1.11` is the tag on the `getPatients` overload a few lines up and looks like it came along with the copy, but `getPatientsAndCount` has never existed on a 1.11 platform and master is on `3.0.0-SNAPSHOT`.

Left as is, a module author reads the published javadoc, sets 1.11 as the platform version they build against, and the call resolves to a `NoSuchMethodError` at runtime on every platform before this release, which is the failure the tag exists to prevent. Worth correcting before this merges.

If a backport to a release line is planned, list those versions alongside it the way the recent additions on master do (`@since 3.0.0, 2.9.0, 2.8.9`).

api/src/main/java/org/openmrs/api/db/hibernate/search/SearchQueryUnique.java
```diff
@@ -596,6 +731,47 @@ static DeduplicationResult collectDuplicateIds(SearchQuery<List<?>> uniqueKeyQue
...
+					}
+					if (!countExceededCap && fullCountUniqueKeys.add(key) && fullCountUniqueKeys.size() > deduplicationCap) {
+						countExceededCap = true;
+					}
+					if (pageFrozen && countExceededCap) {
+						break;
+					}
+				}
+
+				chunk = scroll.next();
```
dkayiwa, line 771: Once the inner loop breaks, this still pulls one more chunk off the scroll before the `while` condition is re-checked and the loop exits, so the scroll fetches up to `DEFAULT_SCROLL_CHUNK_SIZE` (500) hits that nothing ever looks at. I spied the scroll with a chunk size of 5, `maxNewUniqueKeys` 2 and a dedup cap of 2: `collectDuplicateIds` calls `next()` once, this calls it twice. The `break scan` that was here before did not pay that, so it arrived with the label removal.

It only bites in the cap-exceeded path, but that is the expensive search the cap exists to bound, and with the default 500-hit chunk against a cap of 1000 the overshoot is up to half again as many hits as the bounded scan was allowed to read. Putting the guard back in front of the fetch keeps the label gone:

```suggestion
				if (pageFrozen && countExceededCap) {
					break;
				}

				chunk = scroll.next();
```

api/src/main/java/org/openmrs/api/db/hibernate/HibernatePatientDAO.java
```diff
@@ -759,6 +759,59 @@ public Long getCountOfPatients(String query, boolean includeVoided) {
...
+		SearchQueryUnique<?, Patient> patientQuery = SearchQueryUnique
+		        .newProjectedQuery(PatientIdentifier.class,
+		            f -> newPatientIdentifierSearchPredicate(f, query, includeVoided, false), "patient.personId",
+		            this::multiLoadPatients)
+		        .join(SearchQueryUnique
+		                .newProjectedQuery(PersonName.class, f -> personQuery.getPatientNameQuery(f, query, includeVoided),
+		                    "person.personId", this::multiLoadPatients)
+		                .join(SearchQueryUnique.newProjectedQuery(PersonAttribute.class,
+		                    f -> personQuery.getPatientAttributeQuery(f, query, includeVoided), "person.personId",
+		                    this::multiLoadPatients)));
```
dkayiwa: This chain is the one `findPatients` builds on line 881, character for character, and the `minSearchCharacters` guard and the start/length normalisation just above it are copies of that method too. The only thing that actually differs between the two is the terminal call, so a small private builder that both use would leave one definition of what a patient search matches.

What it costs to leave: whoever adds or changes a sub-query next has to notice there are two copies, and a miss shows up as `getPatientsAndCount` and `getPatients` returning different patients rather than as a failure. Not blocking, but the cheap moment to do it is while the clone is still new.

api/src/main/java/org/openmrs/api/PatientService.java
dkayiwa: The import landed here but not in the other three: `PatientDAO`, `PatientServiceImpl` and `HibernatePatientDAO` still spell `org.openmrs.collection.ListPart` out in full in their signatures.

waghmaretejas:
Hi @dkayiwa @ibacher, thanks again for the review and suggestions. I’ve addressed the latest feedback:

* Updated both `getPatientsAndCount()` Javadocs to `@since 3.0.0`.
* Added the guard before `scroll.next()` to avoid the unnecessary extra chunk fetch when both caps are reached.
* Extracted the shared patient search query construction into `getPatientSearchQuery()` and updated both `getPatientsAndCount()` and `findPatients()` to use it.
* Cleaned up the `ListPart` imports in `PatientDAO`, `PatientServiceImpl`, and `HibernatePatientDAO`.

The changes have been pushed to the PR. Please check them when you get a chance.

waghmaretejas:
Hi @ibacher, could you please review this PR when you have a chance? Thank you!

ibacher:
@waghmaretejas Could you run `./mvnw spotless:apply` so this builds?

waghmaretejas:
> @waghmaretejas Could you run `./mvnw spotless:apply` so this builds?

Done, @ibacher . I ran `./mvnw spotless:apply` and pushed the formatting changes. Thanks!

waghmaretejas:
> @waghmaretejas Could you run `./mvnw spotless:apply` so this builds?

Hi @ibacher, just following up on this PR. I’ve addressed the review feedback and pushed the latest changes. Whenever you get a chance, could you please let me know if anything else is needed from my side? Thank you! [TRUNK-6690](https://issues.openmrs.org/browse/TRUNK-6690)

[TRUNK-6690]: https://openmrs.atlassian.net/browse/TRUNK-6690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6502, opened 2026-08-25
TRUNK-6666: Lazy rehash — upgrade legacy password hashes on successful login

solomonfortune:

Added a lazy rehash trigger in `HibernateContextDAO.authenticate()` that transparently upgrades legacy SHA-512/SHA-1 password hashes to the configured encoder on successful login. The upgrade runs in a separate `REQUIRES_NEW` transaction so a rehash failure never affects the login outcome. A conditional native `UPDATE ... WHERE password = :oldHash` in `HibernateUserDAO` handles concurrent logins safely.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6666

## Checklist: I completed these to help reviewers

**Note:** This is a clean rebase of PR #6230 scoped to what TRUNK-6666 owns. Depends on #6219 (already merged). Encoder configuration belongs to TRUNK-6663; legacy hash detection belongs to TRUNK-6665.

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
@@ -199,10 +199,11 @@ public class HibernateContextDAO implements ContextDAO {
 				// Lazy rehash: if password is legacy, upgrade to Argon2id transparently
 				if (passwordOnRecord != null && Security.needsUpgrade(passwordOnRecord)) {
 					try {
-						self.upgradePasswordHash(candidateUser, password, saltOnRecord);
+						String newHash = Security.encodePassword(password + (saltOnRecord != null ? saltOnRecord : ""));
+						self.upgradePasswordHash(candidateUser, passwordOnRecord, newHash);
 					}
 					catch (Exception e) {
-						log.error("Failed to upgrade password hash for user {}: {}",
+						log.error("Failed to upgrade password hash for user {}",
 							candidateUser.getUsername(), e);
 						// login still succeeds — upgrade failure is non-fatal
 					}
@@ -355,31 +356,36 @@ public class HibernateContextDAO implements ContextDAO {
 	 * @param user the authenticated user whose password needs upgrading
 	 * @param rawPassword the plaintext password
 	 * @param salt the legacy salt from the database
-	 * @since 2.8.9
+	 * @since 2.8.10
 	 */
 	@Transactional(propagation = Propagation.REQUIRES_NEW)
-	public void upgradePasswordHash(User user, String rawPassword, String salt) {
-		LoginCredential credential = userDao.getLoginCredential(user);
-		if (credential.getHashedPassword() == null || Security.needsUpgrade(credential.getHashedPassword())) {
-			if (!Security.checkPassword(credential.getHashedPassword(), rawPassword + (salt != null ? salt : ""))) {
-				log.warn("Refusing to upgrade password hash for user {}: password does not match",
-					user.getUsername());
-				return;
-			}
-			String newHash = Security.encodePassword(rawPassword + (salt != null ? salt : ""));
-			boolean upgraded = ((HibernateUserDAO) userDao).conditionallyUpdateUserPassword(
-				user.getUserId(),
-				credential.getHashedPassword(),
-				newHash,
-				user.getUserId(),
-				new Date()
-			);
-			if (upgraded) {
-				log.info("Successfully upgraded password hash for user: {}", user.getUsername());
-			} else {
-				log.debug("Password upgrade skipped for user {} — already changed", user.getUsername());
-			}
+	public void upgradePasswordHash(User user, String oldHash, String newHash) {
+		boolean upgraded = conditionallyUpdateUserPassword(
+			user.getUserId(),
+			oldHash,
+			newHash,
+			user.getUserId(),
+			new Date()
+		);
+		if (upgraded) {
+			log.info("Successfully upgraded password hash for user: {}", user.getUsername());
+		} else {
+			log.debug("Password upgrade skipped for user {} — already changed", user.getUsername());
 		}
+	}
+
+	private boolean conditionallyUpdateUserPassword(Integer userId, String oldHashedPassword,
+			String newHashedPassword, Integer changedBy, Date dateChanged) {
+		String sql = "UPDATE users SET password = :newHash, date_changed = :dateChanged, changed_by = :changedBy "
+				+ "WHERE user_id = :userId AND password = :oldHash";
+		int rows = sessionFactory.getCurrentSession().createNativeQuery(sql)
+				.setParameter("newHash", newHashedPassword)
+				.setParameter("dateChanged", dateChanged)
+				.setParameter("changedBy", changedBy)
+				.setParameter("userId", userId)
+				.setParameter("oldHash", oldHashedPassword)
+				.executeUpdate();
+		return rows > 0;
 	}
 	
 	/**
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
@@ -197,7 +197,7 @@ public class HibernateContextDAO implements ContextDAO {
 				saveUserProperties(candidateUser);
 
 				// Lazy rehash: if password is legacy, upgrade to Argon2id transparently
-				if (passwordOnRecord != null && !passwordOnRecord.startsWith("$argon2id$")) {
+				if (passwordOnRecord != null && Security.needsUpgrade(passwordOnRecord)) {
 					try {
 						self.upgradePasswordHash(candidateUser, password, saltOnRecord);
 					}
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
@@ -360,7 +360,7 @@ public class HibernateContextDAO implements ContextDAO {
 	@Transactional(propagation = Propagation.REQUIRES_NEW)
 	public void upgradePasswordHash(User user, String rawPassword, String salt) {
 		LoginCredential credential = userDao.getLoginCredential(user);
-		if (credential.getHashedPassword() == null || !credential.getHashedPassword().startsWith("$argon2id$")) {
+		if (credential.getHashedPassword() == null || Security.needsUpgrade(credential.getHashedPassword())) {
 			if (!Security.checkPassword(credential.getHashedPassword(), rawPassword + (salt != null ? salt : ""))) {
 				log.warn("Refusing to upgrade password hash for user {}: password does not match",
 					user.getUsername());
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
@@ -41,7 +41,6 @@ import org.openmrs.api.context.Context;
 import org.openmrs.api.context.ContextAuthenticationException;
 import org.openmrs.api.context.Daemon;
 import org.openmrs.api.db.ContextDAO;
-import org.openmrs.api.db.LoginCredential;
 import org.openmrs.api.db.UserDAO;
 import org.openmrs.api.db.hibernate.search.session.SearchSessionFactory;
 import org.openmrs.util.OpenmrsConstants;
@@ -89,6 +88,7 @@ public class HibernateContextDAO implements ContextDAO {
 
 	@Autowired
 	@Lazy
+	@SuppressWarnings("java:S6813") // self-injection required for @Transactional(REQUIRES_NEW) proxy
 	private HibernateContextDAO self;
 	
 	/**
@@ -196,7 +196,7 @@ public class HibernateContextDAO implements ContextDAO {
 				setLastLoginTime(candidateUser);
 				saveUserProperties(candidateUser);
 
-				// Lazy rehash: if password is legacy, upgrade to Argon2id transparently
+				// Lazy rehash: if the stored hash needs upgrading, re-encode transparently on login
 				if (passwordOnRecord != null && Security.needsUpgrade(passwordOnRecord)) {
 					try {
 						String newHash = Security.encodePassword(password + (saltOnRecord != null ? saltOnRecord : ""));
@@ -354,11 +354,10 @@ public class HibernateContextDAO implements ContextDAO {
 	 * Upgrades a user's stored password hash to the configured encoder in a separate transaction.
 	 *
 	 * @param user the authenticated user whose password needs upgrading
-	 * @param rawPassword the plaintext password
-	 * @param salt the legacy salt from the database
+	 * @param oldHash the hash currently stored for the user, used to skip the write if it has changed
+	 * @param newHash the re-encoded hash to store
 	 * @since 2.8.10
 	 */
-	@Transactional(propagation = Propagation.REQUIRES_NEW)
 	public void upgradePasswordHash(User user, String oldHash, String newHash) {
 		boolean upgraded = conditionallyUpdateUserPassword(
 			user.getUserId(),
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
@@ -358,6 +358,7 @@ public class HibernateContextDAO implements ContextDAO {
 	 * @param newHash the re-encoded hash to store
 	 * @since 2.8.10
 	 */
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
 	public void upgradePasswordHash(User user, String oldHash, String newHash) {
 		boolean upgraded = conditionallyUpdateUserPassword(
 			user.getUserId(),
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
@@ -19,6 +19,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.Date;
 
 import org.apache.commons.lang3.StringUtils;
 import org.hibernate.CacheMode;
@@ -40,6 +41,7 @@ import org.openmrs.api.context.Context;
 import org.openmrs.api.context.ContextAuthenticationException;
 import org.openmrs.api.context.Daemon;
 import org.openmrs.api.db.ContextDAO;
+import org.openmrs.api.db.LoginCredential;
 import org.openmrs.api.db.UserDAO;
 import org.openmrs.api.db.hibernate.search.session.SearchSessionFactory;
 import org.openmrs.util.OpenmrsConstants;
@@ -51,8 +53,10 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.orm.hibernate5.SessionFactoryUtils;
 import org.springframework.orm.hibernate5.SessionHolder;
+import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
+import org.springframework.context.annotation.Lazy;
 
 /**
  * Hibernate specific implementation of the {@link ContextDAO}. These methods should not be used
@@ -82,6 +86,10 @@ public class HibernateContextDAO implements ContextDAO {
 	private SearchSessionFactory searchSessionFactory;
 	
 	private UserDAO userDao;
+
+	@Autowired
+	@Lazy
+	private HibernateContextDAO self;
 	
 	/**
 	 * Session factory to use for this DAO. This is usually injected by spring and its application
@@ -187,6 +195,18 @@ public class HibernateContextDAO implements ContextDAO {
 				}
 				setLastLoginTime(candidateUser);
 				saveUserProperties(candidateUser);
+
+				// Lazy rehash: if password is legacy, upgrade to Argon2id transparently
+				if (passwordOnRecord != null && !passwordOnRecord.startsWith("$argon2id$")) {
+					try {
+						self.upgradePasswordHash(candidateUser, password, saltOnRecord);
+					}
+					catch (Exception e) {
+						log.error("Failed to upgrade password hash for user {}: {}",
+							candidateUser.getUsername(), e);
+						// login still succeeds — upgrade failure is non-fatal
+					}
+				}
 
 				// skip out of the method early (instead of throwing the exception)
 				// to indicate that this is the valid user
@@ -327,6 +347,39 @@ public class HibernateContextDAO implements ContextDAO {
 			// skip over errors and leave the attempts at zero
 		}
 		return attempts;
+	}
+
+	/**
+	 * Upgrades a user's stored password hash to the configured encoder in a separate transaction.
+	 *
+	 * @param user the authenticated user whose password needs upgrading
+	 * @param rawPassword the plaintext password
+	 * @param salt the legacy salt from the database
+	 * @since 2.8.9
+	 */
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
+	public void upgradePasswordHash(User user, String rawPassword, String salt) {
+		LoginCredential credential = userDao.getLoginCredential(user);
+		if (credential.getHashedPassword() == null || !credential.getHashedPassword().startsWith("$argon2id$")) {
+			if (!Security.checkPassword(credential.getHashedPassword(), rawPassword + (salt != null ? salt : ""))) {
+				log.warn("Refusing to upgrade password hash for user {}: password does not match",
+					user.getUsername());
+				return;
+			}
+			String newHash = Security.encodePassword(rawPassword + (salt != null ? salt : ""));
+			boolean upgraded = ((HibernateUserDAO) userDao).conditionallyUpdateUserPassword(
+				user.getUserId(),
+				credential.getHashedPassword(),
+				newHash,
+				user.getUserId(),
+				new Date()
+			);
+			if (upgraded) {
+				log.info("Successfully upgraded password hash for user: {}", user.getUsername());
+			} else {
+				log.debug("Password upgrade skipped for user {} — already changed", user.getUsername());
+			}
+		}
 	}
 	
 	/**
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
@@ -111,6 +111,25 @@ public class HibernateUserDAO implements UserDAO {
 		
 		return user;
 	}
+
+	/**
+	 * Updates the password only if the stored hash still matches the expected old hash.
+	 *
+	 * @since 2.8.9
+	 */
+	boolean conditionallyUpdateUserPassword(Integer userId, String oldHashedPassword,
+			String newHashedPassword, Integer changedBy, Date dateChanged) {
+		String sql = "UPDATE users SET password = :newHash, date_changed = :dateChanged, changed_by = :changedBy "
+				+ "WHERE user_id = :userId AND password = :oldHash";
+		int rows = sessionFactory.getCurrentSession().createNativeQuery(sql)
+				.setParameter("newHash", newHashedPassword)
+				.setParameter("dateChanged", dateChanged)
+				.setParameter("changedBy", changedBy)
+				.setParameter("userId", userId)
+				.setParameter("oldHash", oldHashedPassword)
+				.executeUpdate();
+		return rows > 0;
+	}
 	
 	/**
 	 * @see org.openmrs.api.UserService#getUserByUsername(java.lang.String)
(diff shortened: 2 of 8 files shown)


api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -188,6 +196,18 @@ public User authenticate(String login, String password) throws ContextAuthentica
 				setLastLoginTime(candidateUser);
 				saveUserProperties(candidateUser);
 
+				// Lazy rehash: if password is legacy, upgrade to Argon2id transparently
+				if (passwordOnRecord != null && !passwordOnRecord.startsWith("$argon2id$")) {
```
ibacher: The system should not hard-code any assumptions about what password hashing method we are using. Spring's [`PasswordEncoder`](https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/crypto/password/PasswordEncoder.html#upgradeEncoding(java.lang.String)) already provides a mechanism for us to detect this. Start your solution from there.

ibacher (changes requested):

jwnasambu (commented):
@solomonfortune Kindly, why do you have two tickets with the same ticket number?

solomonfortune:
@dkayiwa and @ibacher, I have opened this new PR to replace #6230. The old PR had accumulated too many commits across multiple review rounds and was carrying changes that belong to TRUNK-6663 and TRUNK-6665. This new PR is scoped strictly to what TRUNK-6666 owns: the rehash trigger in authenticate(), the conditional update, and tests. It is based on 2.8.x after #6219 merged. All 5242 tests pass. According to me, it was the better approach. The old one is still open in case you guys decide that this was not the better way of addressing this.

solomonfortune:
> @solomonfortune Kindly, why do you have two tickets with the same ticket number?

the comment below addresses this. i had really gone out of scope and the conflicts were piling up when i updated my branch after your PR being merged

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
solomonfortune: thanks @ibacher

jwnasambu: On the same note 
`if (passwordOnRecord != null && !passwordOnRecord.startsWith("$argon2id$")) {`
The enclosing condition on line 184 already requires `passwordOnRecord != null` for the password match to succeed, so passwordOnRecord != null on this  line (line 200) is always true. Kindly is it not supposed to be removed to avoid noise?

cc @ibacher

api/src/test/java/org/openmrs/api/db/ContextDAOTest.java
```diff
@@ -411,4 +413,16 @@ public void logout_shouldRightlyTriggerUserSessionListener() {
...
+	@Test
+	public void authenticate_shouldUpgradeLegacyPasswordToArgon2idOnSuccessfulLogin() {
+		String originalHash = "4a1750c8607d0fa237de36c6305715c223415189";
+		User user = dao.authenticate("admin", "test");
+		assertNotNull(user);
+		UserDAO userDAO = (UserDAO) applicationContext.getBean("userDAO");
+		LoginCredential credential = userDAO.getLoginCredential(user);
+		assertNotNull(credential.getHashedPassword());
+		assertFalse(originalHash.equals(credential.getHashedPassword()),
+			"Password hash should have been upgraded");
```
jwnasambu:
```suggestion
		assertTrue(credential.getHashedPassword().startsWith("$argon2id$"),
    "Password hash should have been upgraded to Argon2id");
```
To verify the lazy rehash actually produced an Argon2id hash.

solomonfortune:
@ibacher i have replaced the hardcoded startsWith("$argon2id$") check with Security.needsUpgrade(), which delegates to OpenmrsDelegatingPasswordEncoder.upgradeEncoding(). This lets Spring's PasswordEncoder decide whether a stored hash needs upgrading, with no assumptions about the specific algorithm. All 5242 tests pass.

api/src/test/java/org/openmrs/api/db/ContextDAOTest.java
solomonfortune: @jwnasambu, I think that assertion would be correct once TRUNK-6663 wires in the Argon2id encoder. In the current test context, openmrsPasswordEncoder is configured with the legacy encoder, so needsUpgrade() returns false, and no rehash occurs. The test currently just verifies the credential exists after login; the format assertion will make sense once the encoder configuration lands.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -188,6 +196,18 @@ public User authenticate(String login, String password) throws ContextAuthentica
 				setLastLoginTime(candidateUser);
 				saveUserProperties(candidateUser);
 
+				// Lazy rehash: if password is legacy, upgrade to Argon2id transparently
+				if (passwordOnRecord != null && Security.needsUpgrade(passwordOnRecord)) {
```
jwnasambu, line 200:
```suggestion
				if (!Security.getPasswordEncoder().upgradeEncoding(passwordOnRecord)) {
```
OpenmrsDelegatingPasswordEncoder.upgradeEncoding() already detects legacy hashes by checking whether the stored hash lacks the encoder ID prefix. Hard coding $argon2id$ means this code silently breaks if the target encoder ever changes. Using upgradeEncoding() makes the logic encoder agnostic.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -328,6 +348,39 @@ private int getUsersLoginAttempts(User user) {
...
+	 *
+	 * @param user the authenticated user whose password needs upgrading
+	 * @param rawPassword the plaintext password
+	 * @param salt the legacy salt from the database
+	 * @since 2.8.9
+	 */
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
+	public void upgradePasswordHash(User user, String rawPassword, String salt) {
+		LoginCredential credential = userDao.getLoginCredential(user);
+		if (credential.getHashedPassword() == null || !credential.getHashedPassword().startsWith("$argon2id$")) {
```
jwnasambu:
```suggestion
		if (credential.getHashedPassword() == null || !Security.getPasswordEncoder().upgradeEncoding(credential.getHashedPassword())) {
```
The detection mechanism should delegate to the encoder, not hard-code assumptions.

solomonfortune: thanks @jwnasambu

solomonfortune:
@jwnasambu i have fixed the upgradePasswordHash(). it now also uses Security.needsUpgrade() instead of the hardcoded prefix check.

api/src/test/java/org/openmrs/api/db/ContextDAOTest.java
```diff
@@ -411,4 +413,13 @@ public void logout_shouldRightlyTriggerUserSessionListener() {
...
 		assertThat(testUserSessionListener.logins, empty());
 	}
+
+	@Test
+	public void authenticate_shouldUpgradeLegacyPasswordToArgon2idOnSuccessfulLogin() {
+		User user = dao.authenticate("admin", "test");
+		assertNotNull(user);
+		UserDAO userDAO = (UserDAO) applicationContext.getBean("userDAO");
+		LoginCredential credential = userDAO.getLoginCredential(user);
+		assertNotNull(credential.getHashedPassword());
```
jwnasambu, line 423:
```suggestion
		assertTrue(credential.getHashedPassword().startsWith("$argon2id$"));
```
Without this assertion, the test passes even if the upgrade fails.

api/src/main/java/org/openmrs/util/Security.java
```diff
@@ -379,4 +379,8 @@ public static byte[] generateNewSecretKey() {
 		return skey.getEncoded();
 	}
 
+	public static boolean needsUpgrade(String storedEncodedPassword) {
```
jwnasambu, line 389: Kindly the `needsUpgrade` method lacks a @since Javadoc tag, unlike the encodePassword and checkPassword methods above it.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
dkayiwa, line 200: I don't think this branch can run in any configuration we can ship today, so nothing ever gets upgraded.

`needsUpgrade` lands in `OpenmrsDelegatingPasswordEncoder.upgradeEncoding`, which is `extractId(stored) == null && idForEncode` non-empty. `applicationContext-service.xml` builds the bean with `idForEncode = ""`, so it is false for every row. Logging in as `admin` against the real Spring context leaves `users.password` byte for byte identical afterwards.

The only way to make it true is a non-empty `idForEncode`, and that is exactly the case where the constructor throws the `fallbackEncoder` argument away and sets `defaultEncoder = idToPasswordEncoder.get(idForEncode)`. `matches` then routes an unprefixed hash, which is what every legacy row is, to `defaultEncoder`, so legacy rows get handed to the new encoder rather than to `openmrsLegacyPasswordEncoder`. I registered a stand-in modern encoder under `idForEncode = "modern"` and `dao.authenticate("admin", "test")` throws `ContextAuthenticationException: Invalid username and/or password: admin`. The check on line 184 fails, so we never reach line 200.

If this merges as-is, TRUNK-6666 ships as dead code, and the day someone sets `idForEncode` every legacy account is locked out instead of upgraded. The routing fix belongs in `OpenmrsDelegatingPasswordEncoder`: when `extractId` returns null, fall back to the encoder passed in as `fallbackEncoder` instead of to `defaultEncoder`. I changed that one rule locally and the rest of this PR then does exactly what the description says, the row is rewritten on login and the next login authenticates against the new hash. Worth settling with @ibacher whether that belongs here or in its own ticket, but until it exists there is no configuration in which this code runs and no test that can prove it does.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -188,6 +196,18 @@ public User authenticate(String login, String password) throws ContextAuthentica
...
 				saveUserProperties(candidateUser);
 
+				// Lazy rehash: if password is legacy, upgrade to Argon2id transparently
+				if (passwordOnRecord != null && Security.needsUpgrade(passwordOnRecord)) {
+					try {
+						self.upgradePasswordHash(candidateUser, password, saltOnRecord);
+					}
+					catch (Exception e) {
+						log.error("Failed to upgrade password hash for user {}: {}",
+							candidateUser.getUsername(), e);
```
dkayiwa: The trailing `{}` consumes the throwable as a format argument, so slf4j prints `e.toString()` and drops the stack trace. Run against the log4j2 setup in this repo, as written it logs `Failed to upgrade password hash for user admin: java.lang.IllegalStateException: boom` and nothing more, while one placeholder with the throwable as a bare trailing argument prints the full trace and the cause. Since this catch swallows the exception outright, that trace is the only thing anyone gets to debug a failed upgrade from.

```suggestion
						log.error("Failed to upgrade password hash for user {}",
							candidateUser.getUsername(), e);
```

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -328,6 +348,39 @@ private int getUsersLoginAttempts(User user) {
...
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
+	public void upgradePasswordHash(User user, String rawPassword, String salt) {
+		LoginCredential credential = userDao.getLoginCredential(user);
+		if (credential.getHashedPassword() == null || Security.needsUpgrade(credential.getHashedPassword())) {
+			if (!Security.checkPassword(credential.getHashedPassword(), rawPassword + (salt != null ? salt : ""))) {
+				log.warn("Refusing to upgrade password hash for user {}: password does not match",
+					user.getUsername());
+				return;
+			}
+			String newHash = Security.encodePassword(rawPassword + (salt != null ? salt : ""));
```
dkayiwa: `REQUIRES_NEW` suspends the login transaction without releasing its connection, so from the moment `upgradePasswordHash` is entered until it returns, a single login is holding two pooled connections, and the encode on this line is the expensive part of that window. That matters more here than it would elsewhere, because every account takes this path on its first login after the encoder is configured, which is the morning after an upgrade with everyone logging in at once.

Cheapest way out is to compute the new hash in `authenticate` before calling `self.upgradePasswordHash` and pass it in, so the new transaction only runs the conditional UPDATE. That also lets the re-read and re-check on lines 362 to 368 go, since `WHERE password = :oldHash` with `passwordOnRecord` as the expected value already covers the concurrent-login case the ticket describes.

api/src/test/java/org/openmrs/api/db/UserDAOTest.java
```diff
@@ -217,5 +217,35 @@ public void isSecretAnswer_shouldReturnFalseWhenTheAnswerDoesNotMatch() {
...
+		assertNotNull(lc.getHashedPassword());
+	}
+
+	@Test
+	public void changePasswordByOldPassword_shouldStoreTheNewPasswordAsArgon2id() {
+		User admin = dao.getUserByUsername("admin");
+		dao.changePassword("test", "Openmr5xy");
+		LoginCredential lc = dao.getLoginCredential(admin);
+		assertNotNull(lc.getHashedPassword());
+	}
```
dkayiwa: These three don't test what their names claim, and they don't test TRUNK-6666 either. `assertNotNull(lc.getHashedPassword())` was already true before this PR, nothing here looks at the hash format, and `saveUser` and `changePassword` belong to TRUNK-6662 and TRUNK-6663 rather than to the login-time rehash this PR says it is scoped to. `SecurityTest.isLegacyHash_shouldReturnTrueForSha512Hash` has the same problem in a sharper form: there is no `isLegacyHash` anywhere in the PR, and the assertion is that a SHA-512 hex string does not start with `$argon2id$`, which holds for any hex string. I'd drop all four and put the effort into the one test that covers the rehash.

api/src/test/java/org/openmrs/api/db/ContextDAOTest.java
dkayiwa: There's a wrinkle on each side of this one. The `startsWith("$argon2id$")` assertion won't be right even after TRUNK-6663 lands, because `OpenmrsDelegatingPasswordEncoder.encode` prepends `{argon2}`, so the stored value starts with the id and not with `$argon2id$`. And the reason it can't pass today is stronger than the encoder not being wired yet: with `idForEncode = ""` the rehash block in `authenticate` never executes at all, which I've written up on the `Security.needsUpgrade` call.

Which means this test asserts something that was already true before the PR. I ran the five tests it adds against an unmodified 2.8.x, with none of the production changes applied, and all five pass there too. SonarCloud's 0.0% coverage on new code is measuring the same thing.

Getting a test that really exercises the rehash needs neither Argon2 nor BouncyCastle. Register an `OpenmrsDelegatingPasswordEncoder` with a non-empty `idForEncode` and a trivial stand-in encoder in a test context, then assert on the raw `users.password` column before and after the login rather than on `getHashedPassword() != null`. One thing to watch when you write it: the `REQUIRES_NEW` write commits outside the test transaction and survives `@Rollback`. I hit that directly, a later test in the same class found the upgraded row still sitting in the shared in-memory database.

Unrelated to the assertion, the `assertTrue` and `assertFalse` imports added at the top of this file aren't used by anything.

api/src/main/java/org/openmrs/util/Security.java
dkayiwa: Worth fixing the two places where the tag is present, too. `upgradePasswordHash` and `HibernateUserDAO.conditionallyUpdateUserPassword` both carry `@since 2.8.9`, but 2.8.9 was tagged on 7 August and this branch is `2.8.10-SNAPSHOT`, so all three should read 2.8.10, matching the `Security` methods that came in with #6219.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -328,6 +348,39 @@ private int getUsersLoginAttempts(User user) {
...
+	/**
+	 * Upgrades a user's stored password hash to the configured encoder in a separate transaction.
+	 *
+	 * @param user the authenticated user whose password needs upgrading
+	 * @param rawPassword the plaintext password
+	 * @param salt the legacy salt from the database
+	 * @since 2.8.9
+	 */
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
+	public void upgradePasswordHash(User user, String rawPassword, String salt) {
```
ibacher: It's unclear to me why the signature here cannot just be 
```suggestion
	public void upgradePasswordHash(User user, String password) {
```

ibacher: Also, I think this method could actually be private.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -328,6 +348,39 @@ private int getUsersLoginAttempts(User user) {
...
+		LoginCredential credential = userDao.getLoginCredential(user);
+		if (credential.getHashedPassword() == null || Security.needsUpgrade(credential.getHashedPassword())) {
+			if (!Security.checkPassword(credential.getHashedPassword(), rawPassword + (salt != null ? salt : ""))) {
+				log.warn("Refusing to upgrade password hash for user {}: password does not match",
+					user.getUsername());
+				return;
+			}
+			String newHash = Security.encodePassword(rawPassword + (salt != null ? salt : ""));
+			boolean upgraded = ((HibernateUserDAO) userDao).conditionallyUpdateUserPassword(
```
ibacher: I made a comment like this on the previous PR, but down-casting the `UserDAO` to do this is a sign that that function actually belongs here, not on the `UserDAO`.

api/src/test/java/org/openmrs/api/db/ContextDAOTest.java
```diff
@@ -411,4 +413,13 @@ public void logout_shouldRightlyTriggerUserSessionListener() {
 				contains("admin:LOGOUT:SUCCESS"));
 		assertThat(testUserSessionListener.logins, empty());
 	}
+
+	@Test
+	public void authenticate_shouldUpgradeLegacyPasswordToArgon2idOnSuccessfulLogin() {
```
ibacher:
```suggestion
	public void authenticate_shouldUpgradeLegacyPasswordOnSuccessfulLogin() {
```

ibacher: Again, try not to make hard assumptions about what password encoding we're using. The goal of the epic is actually for this to be somewhat _configurable_ with Argon2 as a _default_, so hard-coding Argon2 is counter-productive.

api/src/test/java/org/openmrs/api/db/ContextDAOTest.java
ibacher: It does, however, seem like there's a straight-forward way we could show that the password has been upgraded which this test now skips (check whether the password needs to be upgraded).

api/src/test/java/org/openmrs/api/db/UserDAOTest.java
ibacher: Yeah, this PR is still probably pre-mature until the password-on-save lands?

solomonfortune:
@dkayiwa, @ibacher, and @jwnasambu I have addressed all reviewer comments: fixed OpenmrsDelegatingPasswordEncoder.matches() to route unprefixed hashes to fallbackEncoder instead of defaultEncoder, moved conditionallyUpdateUserPassword to HibernateContextDAO, compute the new hash before entering the REQUIRES_NEW transaction, fixed the log format to preserve the full stack trace, renamed the test to remove the Argon2id assumption, added Security.needsUpgrade() assertion to the upgrade test, fixed @since tags to 2.8.10, and dropped the four tests that didn't belong to this ticket's scope.

but i also have some concerns to guide me through;
- @ibacher making upgradePasswordHash private would prevent @Transactional(REQUIRES_NEW) from applying since Spring's proxy only advises public methods. i am currently using @Autowired @Lazy self-injection to invoke it through the proxy. Should i keep it public for the transaction to work, or is there a different approach you would suggest?

- @dkayiwa, @ibacher, and @jwnasambu i have understood that the current test does not actually exercise the rehash path since needsUpgrade() always returns false with the legacy encoder configured in tests. To write a meaningful test, i need a custom encoder registered with a non-empty idForEncode that triggers the upgrade. However, overriding openmrsPasswordEncoder in TestingApplicationContext.xml would affect all tests. Could you advise on the best way to register a test-scoped encoder without affecting the rest of the test suite?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateUserDAO.java
```diff
@@ -111,6 +111,25 @@ sequences and issues insert on session flush ( batching is possible) .
...
+				+ "WHERE user_id = :userId AND password = :oldHash";
+		int rows = sessionFactory.getCurrentSession().createNativeQuery(sql)
+				.setParameter("newHash", newHashedPassword)
+				.setParameter("dateChanged", dateChanged)
+				.setParameter("changedBy", changedBy)
+				.setParameter("userId", userId)
+				.setParameter("oldHash", oldHashedPassword)
+				.executeUpdate();
+		return rows > 0;
+	}
```
dkayiwa: The move left the original behind. Nothing calls this one any more: grepping `conditionallyUpdateUserPassword` across the branch turns up only the private copy in `HibernateContextDAO` and its single caller there. So the same conditional UPDATE now exists in two places, one of them unreachable, and whoever next changes the statement has to notice there are two of them. Please drop it.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -328,6 +349,44 @@ private int getUsersLoginAttempts(User user) {
...
 	}
+
+	/**
+	 * Upgrades a user's stored password hash to the configured encoder in a separate transaction.
+	 *
+	 * @param user the authenticated user whose password needs upgrading
+	 * @param rawPassword the plaintext password
+	 * @param salt the legacy salt from the database
+	 * @since 2.8.10
+	 */
```
dkayiwa, line 360: The parameters moved on but the tags did not. There is no `rawPassword` or `salt` any more, and `oldHash` and `newHash` are undocumented.

```suggestion
	/**
	 * Upgrades a user's stored password hash to the configured encoder in a separate transaction.
	 *
	 * @param user the authenticated user whose password needs upgrading
	 * @param oldHash the hash currently stored for the user, used to skip the write if it has changed
	 * @param newHash the re-encoded hash to store
	 * @since 2.8.10
	 */
```

Two other bits of text in this file are left over from the earlier shape of the patch. The `LoginCredential` import on line 44 is not referenced by anything, and the comment on line 199 still says the upgrade is "to Argon2id" when the point of going through `Security.needsUpgrade` is that the algorithm is whatever `openmrsPasswordEncoder` happens to be configured with.

api/src/test/java/org/openmrs/spring/OpenmrsDelegatingPasswordEncoderTest.java
```diff
@@ -109,12 +109,11 @@ public void matches_shouldUseTheDefaultEncoderForAnUnprefixedPassword() {
 
 	@Test
 	public void matches_shouldUseTheEncoderNamedByTheIdForEncodeForAnUnprefixedPassword() {
```
dkayiwa: The body now asserts the opposite of the name. With `idForEncode = "bcrypt"` and an unprefixed hash it is the fallback encoder that gets called, and `bcryptEncoder` that is verified as never called. Left as is, the name reads as documentation that unprefixed hashes go to the encoder named by `idForEncode`, which is the routing this PR just changed.

```suggestion
	public void matches_shouldUseTheFallbackEncoderForAnUnprefixedPasswordWhenAnIdForEncodeIsConfigured() {
```

api/src/test/java/org/openmrs/api/db/UserDAOTest.java
```diff
@@ -216,6 +216,5 @@ public void isSecretAnswer_shouldReturnFalseWhenTheAnswerDoesNotMatch() {
 		dao.changeQuestionAnswer(userJoe, SECRET_QUESTION, SECRET_ANSWER);
 		assertFalse(dao.isSecretAnswer(userJoe, "foo"));
 		
-	}
-	
+	}
```
dkayiwa, line 218: Now that the four tests are gone, the only thing left in this file is a whitespace edit: the blank line before the class's closing brace was removed and the `}` above it picked up a trailing tab. Could you restore both so the file drops out of the diff entirely?

api/src/test/java/org/openmrs/api/db/ContextDAOTest.java
dkayiwa: You don't need to touch `TestingApplicationContext.xml`. `openmrsPasswordEncoder` is an ordinary singleton, so a test can swap it for its own duration and put the original back in an `@AfterEach`:

```java
DefaultListableBeanFactory factory = (DefaultListableBeanFactory)
    ((ConfigurableApplicationContext) applicationContext).getBeanFactory();
factory.destroySingleton("openmrsPasswordEncoder");
factory.registerSingleton("openmrsPasswordEncoder", replacement);
```

`Security.getPasswordEncoder()` looks the bean up by name on every call, through `Context.getRegisteredComponent`, so the swap takes effect straight away, with no context reload and nothing leaking into other test classes. For `replacement`, `new OpenmrsDelegatingPasswordEncoder("modern", Collections.singletonMap("modern", standIn), openmrsLegacyPasswordEncoder)` is enough, where `standIn` is a five line `PasswordEncoder` declared inside the test. No Argon2, no BouncyCastle, and nothing in the assertions tied to a particular algorithm.

I ran that shape against 917adbab. `select password from users where user_id = 1` reads `4a1750c8607d0fa237de36c6305715c223415189` before `dao.authenticate("admin", "test")` and `{modern}...` after, which gives you the two assertions worth making: the raw column moved to the configured encoder's form, and a second `authenticate` still returns the user without rewriting the row a second time.

solomonfortune:
@dkayiwa i have addressed all comments i.e. removed the duplicate conditionallyUpdateUserPassword from HibernateUserDAO, updated the Javadoc on upgradePasswordHash() with correct parameter tags, removed the unused LoginCredential import, fixed the inline comment to remove the Argon2id assumption, renamed the test to match its actual behavior, restored UserDAOTest formatting, and added @SuppressWarnings("java:S6813") on the self field for SonarQube.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -328,6 +349,43 @@ private int getUsersLoginAttempts(User user) {
...
+
+	/**
+	 * Upgrades a user's stored password hash to the configured encoder in a separate transaction.
+	 *
+	 * @param user the authenticated user whose password needs upgrading
+	 * @param oldHash the hash currently stored for the user, used to skip the write if it has changed
+	 * @param newHash the re-encoded hash to store
+	 * @since 2.8.10
+	 */
+	public void upgradePasswordHash(User user, String oldHash, String newHash) {
```
dkayiwa, line 362: The last commit dropped `@Transactional(propagation = Propagation.REQUIRES_NEW)` from this method, and I don't think that was deliberate. The commit message doesn't mention it, `Propagation` on line 55 is now an unused import, and the comment on the `self` field (line 91) and the one in the catch block (line 208) both still promise the isolation it gave you.

It needs to come back before this merges. Without it the conditional UPDATE runs inside the login transaction, so when the statement throws, Hibernate calls `markForRollbackOnly()` on that transaction. The catch in `authenticate` swallows the exception and returns the user, and the commit that follows finds the transaction marked and quietly rolls it back. Hibernate turns that into a `RollbackException` only when `hibernate.jpa.compliance.transaction` is set, and we set it nowhere, so nothing is thrown and nothing is logged past the swallowed cause. The login reports success while every write it made is discarded.

I measured it on this branch. Registering a stand-in encoder under a non-empty `idForEncode` and having it emit a hash longer than `users.password` (VARCHAR(128)) forces the UPDATE to fail. At head, `dao.authenticate("admin", "test")` returns the user, the transaction moves to `status=MARKED_ROLLBACK rollbackOnly=true`, the commit returns normally, and `lastLoginTimestamp` is absent from `user_property` afterwards. Restoring this one line and changing nothing else, the same run leaves the transaction `ACTIVE` and `lastLoginTimestamp` lands, as it also does when the rehash succeeds and when `idForEncode` is `""`. So the loss is specific to a failed upgrade. The column overflow is only a convenient trigger; in production it would be a lock wait timeout or a deadlock on `users`, during the login storm on the first morning after the encoder is switched on, which is when every account takes this path at once.

That also answers what you asked @ibacher above: the method has to stay public and be reached through `self` for the annotation to apply, and putting it back is what makes the self-injection and the `java:S6813` suppression worth carrying.

```suggestion
	@Transactional(propagation = Propagation.REQUIRES_NEW)
	public void upgradePasswordHash(User user, String oldHash, String newHash) {
```

api/src/test/java/org/openmrs/api/db/UserDAOTest.java
dkayiwa: Still there at head. The two blank lines went, but their tabs moved up onto the lines above: line 217 is now `assertFalse(dao.isSecretAnswer(userJoe, "foo"));` followed by two tabs, and line 218 is `}` followed by one. On 2.8.x neither line carries a trailing tab, and there is a two-tab line after the assert plus a one-tab line before the closing brace. `git diff origin/2.8.x -- api/src/test/java/org/openmrs/api/db/UserDAOTest.java` coming back empty is the check.

PR #6503, opened 2026-08-26
TRUNK-6707: Cache global property lookups

Rocker810:
<!--- 'TRUNK-JiraIssueNumber: JiraIssueTitle' -->
Global property reads happen constantly and every one was a database round trip. Two layers of caching, since each fixes a case the other can't.                                   
                                                                                                                                                                                     DAO:`getGlobalPropertyObject` ran `lower(property) = ?`, which can't use the primary key index and never consults the second-level entity cache. It now tries `session.get` by exact name first and falls back to the case-insensitive query only on a miss, with that fallback marked cacheable. Case-insensitive behaviour is unchanged.                        
                                                                                                                                                                                     
Service: Hibernate can only cache rows that exist, so lookups of unset properties — very common via `getGlobalProperty(name, default)` — hit the database every time. Results are now cached in the existing API cache manager, with unset properties stored as an explicit `ABSENT` marker.                                                                     

Evicted on save/set/update/purge through the API, and the whole API cache is cleared on context refresh, covering properties modules insert via Liquibase. Direct SQL writes to `global_property` aren't supported — they already bypass the `GlobalPropertyListener` caches today.                                                                                
                                                                                                                                                                                     
Tests: 6 added to `HibernateAdministrationDAOTest`, new `GlobalPropertyCacheTest` with 10 covering cache hits, negative caching and eviction on every write path, plus `CacheConfigTest`. Full `api` suite: 5181 tests, 0 failures.            

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-6707

  No? -> execute above command


api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -98,28 +97,36 @@ public String getGlobalProperty(String propertyName) throws DAOException {
...
-
-			GlobalProperty gp = session.createQuery(query).uniqueResult();
-			if (gp != null) {
-				// GP may be null, but the session may contain an unflushed gp so
-				// we will do a final check with session.get below. It may happen,
-				// if flush is set to manual and running in a larger transaction.
-				return gp;
-			}
+			return session.createQuery(query).setCacheable(true).uniqueResult();
 		}
```
jwnasambu, line 128: This removes the `session.get()` fallback that was there to catch unflushed properties with mismatched casing on case-sensitive databases.

The old code's flow was:
1. Case-sensitive DB → run `lower()` query → if found, return → if not found, fall back to `session.get()` (catches unflushed entities in the first-level cache)
2. Case-insensitive DB → `session.get()` (works because DB collation handles case)

The new code tries `session.get()` first (good for the exact-match fast path), but on a case-sensitive DB when the property doesn't exist in the DB yet (saved but unflushed, manual flush mode, larger transaction), the `lower()` query misses it and there is no `session.get()` after it to catch the entity sitting in the session. The method returns null where the old code would have found it.

The fix is to restore `session.get()` as a final fallback after the query, which preserves all three benefits (exact-match fast path, query-cacheable optimization, and unflushed-entity safety net):

```java
if (isDatabaseStringComparisonCaseSensitive()) {
    CriteriaBuilder cb = session.getCriteriaBuilder();
    CriteriaQuery<GlobalProperty> query = cb.createQuery(GlobalProperty.class);
    Root<GlobalProperty> root = query.from(GlobalProperty.class);
    query.where(cb.equal(cb.lower(root.get(PROPERTY)), propertyName.toLowerCase()));
    GlobalProperty gp = session.createQuery(query).setCacheable(true).uniqueResult();
    if (gp != null) {
        return gp;
    }
}
// final fallback: catch unflushed entities in the first-level cache
return session.get(GlobalProperty.class, propertyName);
```

api/src/main/java/org/openmrs/api/cache/GlobalPropertyCache.java
```diff
@@ -0,0 +1,89 @@
...
+	/**
+	 * Drops the cached snapshot of the named property. Must be called whenever a global property is
+	 * saved, updated or purged through the API.
+	 *
+	 * @param propertyName the property to evict, not null
+	 */
+	@CacheEvict(value = CACHE_NAME, key = KEY)
+	public void evict(String propertyName) {
+		// the annotation does the work
+	}
```
ibacher, line 79: I don't think this eviction is, strictly-speaking, sufficient. OpenMRS is an inherently multi-threaded application and right now, this is only called inside an existing Spring transaction, which is not guaranteed to be _after the transaction is committed_. To make that guarantee, we need something more like:

```
public void evict(String propertyName) {
        cache.evict(propertyName);

        if (TransactionSynchronizationManager.isSynchronizationActive()) {
                pendingEvictions().add(key);
        }
 }

private Set<String> pendingEvictions() {
        Set<String> keys = (Set<String>) TransactionSynchronizationManager.getResource(this);
        if (keys != null) {
                return keys;
        }

        keys = new HashSet<>();
        TransactionSynchronizationManager.bindResource(this, keys);
        TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {

                @Override
                public void afterCompletion(int status) {
                        // Runs on commit and on rollback alike: either outcome can leave an entry that was
                        // populated from state no longer, or never, in the database.
                        Set<String> pending = (Set<String>) TransactionSynchronizationManager
                                .unbindResourceIfPossible(GlobalPropertyCache.this);
                        if (pending != null) {
                                pending.forEach(p -> cache.evict℗);
                        }
                }
        });

        return keys;
}
```

api/src/main/java/org/openmrs/api/cache/GlobalPropertyCache.java
```diff
@@ -0,0 +1,89 @@
...
+	 * <p>
+	 * Properties that do not exist are cached as {@link CachedGlobalProperty#ABSENT} so that repeated
+	 * lookups of an unset property do not hit the database either. This method never returns null.
+	 *
+	 * @param propertyName the property to look up, not null
+	 * @param loader loads the property from the database on a cache miss
+	 * @return a snapshot of the property, or {@link CachedGlobalProperty#ABSENT} if it does not exist
+	 */
+	@Cacheable(value = CACHE_NAME, key = KEY)
+	public CachedGlobalProperty get(String propertyName, Supplier<GlobalProperty> loader) {
```
ibacher, line 64: Note that there's an asymmetry here that's not helpful if run in case-insensitive (vs case-sensitive) mode: in case-insensitive mode `module.datetime` and `module.dateTime` should be treated as the same value, but here the cache is completely sensitive to the value supplied by the caller.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -98,28 +97,36 @@ public String getGlobalProperty(String propertyName) throws DAOException {
 	 */
 	@Override
 	public GlobalProperty getGlobalPropertyObject(String propertyName) {
+		if (propertyName == null) {
+			// session.get() below would throw IllegalArgumentException a few frames deeper anyway,
+			// since the property name is the identifier. Fail here with a clearer message.
+			throw new IllegalArgumentException("propertyName is required");
+		}
```
ibacher, line 104: Please don't add new exceptions, though. Just return `null`.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -98,28 +97,36 @@ public String getGlobalProperty(String propertyName) throws DAOException {
...
+
 		Session session = sessionFactory.getCurrentSession();
 
+		// Look the property up by its identifier first. This hits the first-level session cache and
+		// then the second-level entity cache, so a warm property costs no query at all. It also
+		// picks up an unflushed property already present in the session, which may happen when the
+		// flush mode is manual and we are running inside a larger transaction.
```
ibacher, line 111:
```suggestion
		// Try to find the GP by the supplied string
```

ibacher: Note that on this comment: OpenMRS _normally_ runs with flush mode set to manual.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -98,28 +97,36 @@ public String getGlobalProperty(String propertyName) throws DAOException {
...
+		GlobalProperty gp = session.get(GlobalProperty.class, propertyName);
+		if (gp != null) {
+			return gp;
+		}
+
+		// Only pay for the case-insensitive lookup when the exact match missed. On databases with
+		// case-insensitive string comparison the query above has already covered it.
```
ibacher, line 118:
```suggestion
		// Couldn't find a match for the string, so run a case-insensitive query
```

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -98,28 +97,36 @@ public String getGlobalProperty(String propertyName) throws DAOException {
...
-			query.where(condition);
-
-			GlobalProperty gp = session.createQuery(query).uniqueResult();
-			if (gp != null) {
-				// GP may be null, but the session may contain an unflushed gp so
-				// we will do a final check with session.get below. It may happen,
-				// if flush is set to manual and running in a larger transaction.
-				return gp;
-			}
+			return session.createQuery(query).setCacheable(true).uniqueResult();
```
ibacher, line 126:
```suggestion
			return session.createQuery(query).uniqueResult();
```

api/src/main/java/org/openmrs/api/db/hibernate/HibernateAdministrationDAO.java
```diff
@@ -98,28 +97,36 @@ public String getGlobalProperty(String propertyName) throws DAOException {
...
 		}
 
-		return session.get(GlobalProperty.class, propertyName);
+		return null;
```
ibacher, line 129: As @jwnasambu noted, restore this.

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
```diff
@@ -120,6 +125,17 @@ public void setAdministrationDAO(AdministrationDAO dao) {
 		this.dao = dao;
 	}
 
+	/**
+	 * Mainly for tests that build this service by hand. An unproxied {@link GlobalPropertyCache} simply
+	 * delegates to the loader without caching, which is usually what such a test wants.
+	 *
+	 * @param globalPropertyCache the cache to use for global property lookups
+	 * @since 3.0.0
```
ibacher, line 133: This is meant to be backported. Generally, it's a good idea _not_ to guess at `@since` until a PR is fully ready.

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
```diff
@@ -178,17 +194,24 @@ public String getGlobalProperty(String propertyName) throws APIException {
...
-				return gp.getPropertyValue();
-			} else {
-				throw new APIException("GlobalProperty.error.privilege.required.view",
-				        new Object[] { gp.getViewPrivilege().getPrivilege(), propertyName });
-			}
-		} else {
+		// Note that only the property data is cached, never the outcome of the privilege check
+		// below. The check has to run on every call, otherwise one user's read would warm the
+		// cache and let the next user past a privilege they do not hold.
+		CachedGlobalProperty cached = globalPropertyCache.get(propertyName, () -> dao.getGlobalPropertyObject(propertyName));
```
ibacher, line 200:
```suggestion
		CachedGlobalProperty cached = globalPropertyCache.get(propertyName, () -> dao.getGlobalPropertyObject(propertyName));
```

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
```diff
@@ -178,17 +194,24 @@ public String getGlobalProperty(String propertyName) throws APIException {
...
+
+		if (!cached.isPresent()) {
 			return null;
 		}
+
+		String viewPrivilege = cached.getViewPrivilege();
```
ibacher, line 206:
```suggestion
		// outcome of privilege check should never be cached
		String viewPrivilege = cached.getViewPrivilege();
```

api/src/main/java/org/openmrs/api/impl/AdministrationServiceImpl.java
```diff
@@ -178,17 +194,24 @@ public String getGlobalProperty(String propertyName) throws APIException {
...
+		// Context.hasPrivilege(name, false) rather than a plain privilege check, to match
+		// canViewGlobalProperty: a proxy privilege must not satisfy the view check.
```
ibacher, line 208:
```suggestion
		// proxy privileges shouldn't be taken into account here
```

api/src/main/java/org/openmrs/module/ModuleUtil.java
```diff
@@ -910,6 +910,11 @@ public static AbstractRefreshableApplicationContext refreshApplicationContext(Ab
 		}
 		log.debug("Done refreshing context");
 
+		// Modules may have inserted or changed global properties through Liquibase while starting,
+		// which happens entirely outside the API and so evicts nothing. Drop the API caches so that
+		// anything cached from before the refresh is re-read.
+		ServiceContext.getInstance().clearEntireApiCache();
```
ibacher, line 916: Which is why the cache is a bean. This doesn't seem germane to the PR. Remove it.

api/src/main/resources/cache-api.yaml
```diff
@@ -5,5 +5,9 @@ caches:
         configuration: "entity"
     serializerWhiteListTypes:
         configuration: "entity"
+    globalProperty:
+        configuration: "entity"
+        memory:
+            maxCount: 5000
```
ibacher, line 11:
```suggestion
```

api/src/main/resources/cache-api.yaml
```diff
@@ -5,5 +5,9 @@ caches:
...
     rolePrivileges:
         configuration: "role-privileges"
```
ibacher, line 13: Similarly to what was done with role-privileges we may want to ensure the cache configuration here has a lifetime (the role-privilege one may not be right!) to catch database updates.

api/src/main/java/org/openmrs/api/cache/CachedGlobalProperty.java
```diff
@@ -0,0 +1,95 @@
...
+	private final String viewPrivilege;
+
+	private CachedGlobalProperty(boolean present, String value, String viewPrivilege) {
+		this.present = present;
+		this.value = value;
+		this.viewPrivilege = viewPrivilege;
+	}
+
+	/**
+	 * Takes a snapshot of the given global property.
```
ibacher, line 58:
```suggestion
	 * Creates a snapshot of the given global property.
```

api/src/main/java/org/openmrs/api/cache/CachedGlobalProperty.java
```diff
@@ -0,0 +1,95 @@
...
+ * @since 3.0.0
+ */
+public final class CachedGlobalProperty implements Serializable {
+
+	private static final long serialVersionUID = 1L;
+
+	/**
+	 * Marks a property that does not exist. Caching this, rather than a plain {@code null}, is what
+	 * makes repeated lookups of unset properties free: Hibernate can only cache rows that exist, so
+	 * without a negative entry every such lookup is a guaranteed database round trip.
```
ibacher, line 41:
```suggestion
	 * Used to indicate that a property does not exist. Cannot just be {@code null} because we need to differentiate "confirmed missing" from "not loaded"
```

ibacher:
Closing this as we'll need to get another version of this in.

PR #6505, opened 2026-08-26
TRUNK-6705: Reduce the overhead of the gzip filter

ibacher:

The ticket asks for caching the compiled patterns and the enabled flag, both of which are set through Global Properties. In addition, this implements some more optimizations, e.g., using ConfigUtil's cache for GPs and two semantic-preserving changes to UserContext which completely removes an allocation in the (common) case where there is only one privilege added per `addProxyPrivilege` call.

The GZIPFilterTest class is moved to the right package and regression tests for caching are added.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6705

  No? -> execute above command


api/src/main/java/org/openmrs/api/context/UserContext.java
```diff
@@ -248,7 +248,8 @@ public void logout() {
 	 * @param privilege to give to users
 	 */
 	public void addProxyPrivilege(String privilege) {
-		addProxyPrivilege(new String[] { privilege });
+		log.debug("Adding proxy privilege: {}", privilege);
+		proxies.add(privilege);
 	}
```
dkayiwa, line 253: Inlining this drops the null check that the varargs overload does, and it needs to come back before this merges. On master `addProxyPrivilege((String) null)` goes through `addProxyPrivilege(String[])` and throws `IllegalArgumentException: UserContext.addProxyPrivilege does not accept null privileges`; on this branch it puts a null into `proxies`. I ran both, and two things follow from that null. The next `hasPrivilege(anything, true)` throws `NullPointerException: Cannot invoke "String.equals(Object)" because "s" is null` at UserContext:428, for any privilege, not just the one that was added. And `hasProxyPrivileges()` starts returning true, which is what `AuthorizationAdvice` consults to decide whether an unauthenticated caller may enter an `@Authorized` method that names no privileges. So one null from any module either breaks every privilege check on that session or waves callers into methods meant to require authentication, and nothing throws at the call site to point at where it came from.

The check went in on this exact method in 45cfd84 for TRUNK-6259, and the tests covering it only pass `(String[]) null`, so the suite stays green either way.

`removeProxyPrivilege(String)` just below moved the same way: master treats a null as a no-op, this calls `proxies.remove(null)`.

I would drop this file from the PR. TRUNK-6705 is about the gzip filter, and what the inlining buys is one `String[1]` per call on a path that is not hot, which is not worth trading a guard for; if it is worth having, it is worth its own ticket with its own null tests. If you would rather keep it here, the guard has to move into the single-argument method.

web/src/main/java/org/openmrs/web/filter/GZIPFilter.java
```diff
@@ -102,68 +106,116 @@ public HttpServletRequest performGZIPRequest(HttpServletRequest req) {
 	 */
 	private boolean isGZIPSupported(HttpServletRequest req) {
 		String browserEncodings = req.getHeader("accept-encoding");
-		boolean supported = ((browserEncodings != null) && (browserEncodings.contains("gzip")));
-
-		String userAgent = req.getHeader("user-agent");
+		return (browserEncodings != null) && browserEncodings.contains("gzip");
```
dkayiwa, line 109: This also drops the bypass for a `user-agent` starting with `httpunit`, so a client identifying as httpunit now gets a compressed response where it previously got an uncompressed one. Nothing in core references httpunit any more and there is no dependency on it, so in this tree it was dead weight, but it is still a live behavior change for anyone driving the webapp from an httpunit-based suite, and TRUNK-6705 does not ask for it. The Javadoc you put on `shouldNotCompressWhenTheClientDoesNotAdvertiseGzipSupport` reads like it was deliberate, which is why I am asking rather than just flagging it: you wrote up both of the accept/deny deviations in the ticket, and this one is in neither the ticket nor the description. I would take it back out of this PR, or if it goes with it, give it a line in the description so it is on the record. Same for the `@NonNull` annotations on `doFilterInternal`, which the ticket does not need either.

web/src/test/java/org/openmrs/web/filter/GZIPFilterTest.java
```diff
@@ -0,0 +1,150 @@
...
+		assertThrows(APIException.class, () -> gzipFilter.performGZIPRequest(compressedRequestFor("/openmrs/index.htm")));
+	}
+
+	/**
+	 * The accepted paths are compiled once and held until the property changes. This checks the whole
+	 * chain that makes a change visible: saving the property notifies ConfigUtil, which is a
+	 * GlobalPropertyListener, and the filter then sees the new value and recompiles.
+	 */
+	@Test
+	public void performGZIPRequest_shouldPickUpChangesToTheAcceptedPathsGlobalProperty() throws Exception {
```
dkayiwa, line 124: The deviation you documented in the ticket, that accept/deny for a compressed request no longer depends on whether some authenticated request warmed the cache first, is the one behavior change here with no test behind it. This class authenticates as admin in its base class, and the two privilege tests in `GZIPFilterCachingTest` assert that `Context.addProxyPrivilege` was called rather than that an unauthenticated request now gets served, so nothing pins the part that changes who is served.

A sibling to this test covers it: set the accepted paths, `Context.logout()`, then assert `performGZIPRequest` returns a `GZIPRequestWrapper` for an allowlisted path, with `authenticate()` in a finally so the `@AfterEach` purge still has privileges. I wrote it out and ran it, and it errors with `APIException: Unsupported Media Type` against the old filter and passes against yours, so it pins exactly the change you described.

api/src/main/java/org/openmrs/api/context/UserContext.java
ibacher: I actually added this because the allocation here showed in profiling as a not-insubstantial part of each GP-lookup. But yeah, you're right on not having the same guards in place.

web/src/main/java/org/openmrs/web/filter/GZIPFilter.java
ibacher: This is intentional dead-weight removal. I couldn't find anything here or in the platform that actually used HttpUnit.

web/src/test/java/org/openmrs/web/filter/GZIPFilterTest.java
lolka123455: Fixed in 80a0efd. A superseded refresh now exits without recording a transport failure, preserving the winning refresh diagnostic; the overlapping-refresh regression test asserts lastFetch remains Success. The focused Gradle test is pending CI because this environment has no JDK.

PR #6519, opened 2026-08-30
Fix #6336: mergePatients runs the merge tail twice (duplicate savePatient and PersonMergeLog)

### LINKED ISSUE #6336: mergePatients runs the merge tail twice (duplicate savePatient and PersonMergeLog)

mandar1045:
### What's going on here

Found the cause of #6336 — mergePatients was running the "merge tail" step (saving the preferred patient, writing the PersonMergeLog, voiding the non-preferred patient/person) twice for every merge.

The reason: mergeAddresses(...) had a copy of the entire merge tail pasted in right after it merges the addresses. No idea why it ended up there, but it meant the tail logic ran once from mergeAddresses and then again from mergePatients like normal. Two savePatient calls, two PersonMergeLog entries, for one merge.

### What I did about it:

Ripped out that duplicate block from mergeAddresses. Now it just merges addresses like it's supposed to, nothing else tacked on.
While pulling that block out, I noticed it was doing some audit logging for deathdateEstimated that mergeDateOfDeath (the method that should actually own this) doesn't do on its own. Didn't want to just throw that away, so I moved it over to mergeDateOfDeath where it belongs.

After this, the merge tail only fires once, from mergePatients, which is how it was supposed to work in the first place.

### Issue

Fixes #6336 — https://github.com/openmrs/openmrs-core/issues/6336

### DIFF
--- a/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java
@@ -822,6 +822,11 @@ public class PatientServiceImpl extends BaseOpenmrsService implements PatientSer
 	}
 
 	private void mergeDateOfDeath(Patient preferred, Patient notPreferred, PersonMergeLogData mergedData) {
+		mergedData.setPriorDateOfDeathEstimated(preferred.getDeathdateEstimated());
+		if (preferred.getDeathdateEstimated() == null) {
+			preferred.setDeathdateEstimated(notPreferred.getDeathdateEstimated());
+		}
+
 		mergedData.setPriorDateOfDeath(preferred.getDeathDate());
 		if (preferred.getDeathDate() == null) {
 			preferred.setDeathDate(notPreferred.getDeathDate());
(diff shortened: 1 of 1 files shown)


mandar1045:
Hey @dkayiwa 

Just opened PR #6519 to fix this. You were spot on—I deleted the redundant merge tail logic from mergeAddresses, so now mergePatients handles the tail just once.

Also, while doing that, I noticed the duplicate block was actually auditing deathdateEstimated which the main mergeDateOfDeath missed. I safely moved that logic over so we don't lose that data!

Tests are passing locally. Let me know what you think!

PR #6522, opened 2026-08-31
TRUNK-6793: Add optional OpenTelemetry metrics export

jayasanka-sack:

OpenMRS now speaks OpenTelemetry!

openmrs-core can now export OpenTelemetry metrics. bundles the OTel Java agent and attaches it when `OMRS_OTEL_ENABLED`=true.

off by default. nothing changes for anyone who doesn't opt in.

### Why do we need this?

Implementers currently have no standard way to see whether an openmrs server is healthy. everyone builds their own thing, in whatever shape, so nothing's shareable.

OTel is a CNCF standard that basically every observability backend speaks. So instead of us picking a monitoring vendor for people, we emit a standard format and they point it wherever:

_"openmrs exports standard OTel. Grab it however you like."_

**Why core and not the distro:** bahmni, kenyaEMR, ugandaEMR and the rest all build on core. distro-only means each of them reimplements it. also the attach logic has to live in startup.sh anyway, and that's here.

**Why the java agent and not a JMX scraper**: runs in-process, exposes nothing. a scraper needs every deployment to open a JMX/RMI port, which is a rough default for a clinical system.

### What changed

Dockerfile: downloads the agent to /openmrs/otel in the production stage, checksum-verified, same pattern as the tini and tomcat downloads. Pure java so no TARGETARCH handling needed.

startup.sh: appends -javaagent: to CATALINA_OPTS when the flag is set. mirrors the existing OMRS_DEV_DEBUG_PORT block right above it. uses ${VAR-} since the script runs under -u.

that's it. everything past the on/off flag is standard `OTEL_*` variables, so the config surface is the OTel spec rather than something we invented.

Sample config:
```
        OMRS_OTEL_ENABLED: true
        OTEL_EXPORTER_OTLP_ENDPOINT: http://alloy:4317
        OTEL_EXPORTER_OTLP_PROTOCOL: grpc
        OTEL_SERVICE_NAME: openmrs-backend
        OTEL_JMX_TARGET_SYSTEM: tomcat
        OTEL_METRICS_EXPORTER: otlp
        OTEL_TRACES_EXPORTER: none
        OTEL_LOGS_EXPORTER: none
```

### The cost

The jar adds ~24MB to every openmrs-core image whether you use it or not. that's the real tradeoff.

No runtime cost when disabled, the -javaagent flag just never gets added. 

### Tested

Validated end to end on the reference application: agent → gRPC → Alloy → Prometheus → Grafana. 

Distro PR: https://github.com/openmrs/openmrs-distro-referenceapplication/pull/1069

image

## Issue I worked on

see https://openmrs.atlassian.net/browse/TRUNK-6793

  No? -> execute above command

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -138,8 +138,19 @@ RUN mkdir -p /openmrs/data/modules \
     && mkdir -p /openmrs/data/configuration_checksums \
     && mkdir -p /openmrs/data/complex_obs \
     && mkdir -p /openmrs/data/activemq-data \
+    && mkdir -p /openmrs/otel \
     && chmod -R g+rw /openmrs \
     && chown -R 1001 /openmrs
+
+# Setup OpenTelemetry Java agent
+# Pure Java, so no TARGETARCH handling needed.
+# Bump OTEL_AGENT_VERSION and OTEL_AGENT_SHA together.
+ARG OTEL_AGENT_VERSION=2.30.0
+ARG OTEL_AGENT_SHA="9d6bc2ad8dd8fb7f730984988e57b8ac0a82d81c7b3b8ae795378718733a509d"
+ARG OTEL_AGENT_URL="https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/v${OTEL_AGENT_VERSION}/opentelemetry-javaagent.jar"
+RUN curl -fsSL -o /openmrs/otel/opentelemetry-javaagent.jar "${OTEL_AGENT_URL}" \
+    && echo "${OTEL_AGENT_SHA}  /openmrs/otel/opentelemetry-javaagent.jar" | sha256sum -c \
+    && chmod g+r /openmrs/otel/opentelemetry-javaagent.jar
     
 # Copy in the start-up scripts
 COPY --from=dev /openmrs/wait-for-it.sh /openmrs/startup-init.sh /openmrs/startup.sh /openmrs/
--- a/Dockerfile
+++ b/Dockerfile
@@ -145,8 +145,8 @@ RUN mkdir -p /openmrs/data/modules \
 # Setup OpenTelemetry Java agent
 # Pure Java, so no TARGETARCH handling needed.
 # Bump OTEL_AGENT_VERSION and OTEL_AGENT_SHA together.
-ARG OTEL_AGENT_VERSION=2.30.0
-ARG OTEL_AGENT_SHA="9d6bc2ad8dd8fb7f730984988e57b8ac0a82d81c7b3b8ae795378718733a509d"
+ARG OTEL_AGENT_VERSION=2.31.1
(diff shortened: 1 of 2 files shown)


startup.sh
```diff
@@ -76,6 +76,11 @@ if [ -n "${OMRS_DEV_DEBUG_PORT-}" ]; then
   fi
 fi
 
+if [ "${OMRS_OTEL_ENABLED:-false}" = "true" ]; then
+  echo "Enabling OpenTelemetry, exporting to ${OTEL_EXPORTER_OTLP_ENDPOINT:-default OTLP endpoint}"
+  CATALINA_OPTS="$CATALINA_OPTS -javaagent:/openmrs/otel/opentelemetry-javaagent.jar"
```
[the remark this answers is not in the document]
ibacher: @jayasanka-sack This is actually a reasonable point we should fix.

startup.sh
```diff
@@ -76,6 +76,11 @@ if [ -n "${OMRS_DEV_DEBUG_PORT-}" ]; then
   fi
 fi
 
+if [ "${OMRS_OTEL_ENABLED:-false}" = "true" ]; then
+  echo "Enabling OpenTelemetry, exporting to ${OTEL_EXPORTER_OTLP_ENDPOINT:-default OTLP endpoint}"
```
ibacher:
```suggestion
fi
fi

if [ "${OMRS_OTEL_ENABLED:-false}" = "true" ]; then
  echo "Enabling OpenTelemetry, exporting to ${OTEL_EXPORTER_OTLP_ENDPOINT:-the default OTLP endpoint}"
```

Dockerfile
```diff
@@ -138,8 +138,19 @@ RUN mkdir -p /openmrs/data/modules \
     && mkdir -p /openmrs/data/configuration_checksums \
     && mkdir -p /openmrs/data/complex_obs \
     && mkdir -p /openmrs/data/activemq-data \
+    && mkdir -p /openmrs/otel \
```
ibacher: I feel like we shouldn't start a convention of "folders for features" at least for top-level stuff. Maybe `/openmrs/java-agents`?

jayasanka-sack: Yeah that's better. fixed with be56ee98d497e3602c58039aabc2e112bb6f8040

startup.sh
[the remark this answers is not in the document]
jayasanka-sack: fixed with: 
efb9a51

ibacher (approved):

ibacher:
@jayasanka-sack Any idea where this needs to be backported to?

jayasanka-sack:
2.8.x and above would be enough as the current distro is at 2.8.10.

ibacher:
Perfect. Do you have time to throw together the PRs or commits?

PR #6523, opened 2026-08-31
TRUNK-6786: NameSupport global property listener watches the wrong property name

nsalifu:
`NameSupport.supportsPropertyName()` only matched the name template global property, while `getDefaultLayoutFormat()` actually reads the name format global property. 
This made `globalPropertyChanged()/globalPropertyDeleted()` dead code with respect to the layout format, and blocked memoizing the format lookup (which runs on the hot path in `PersonName.getFullName()`).

I updated supportsPropertyName() to also match the name format property, added handling in globalPropertyChanged() to cache the new format value, and added globalPropertyDeleted() logic to fall back to the configured default when the format property is removed mirroring the pattern AddressSupport already uses. getDefaultLayoutFormat() now reads the memoized value instead of hitting the database on every call.

## Issue I worked on
Issue:  [TRUNK-6786](https://openmrs.atlassian.net/browse/TRUNK-6786)

- [x] I ran ./mvnw clean package right before creating this pull request and added all formatting changes to my commit.

[TRUNK-6786]: https://openmrs.atlassian.net/browse/TRUNK-6786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/src/main/java/org/openmrs/layout/name/NameSupport.java
+++ b/api/src/main/java/org/openmrs/layout/name/NameSupport.java
@@ -35,6 +35,9 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 
 	private boolean initialized = false;
 
+	// Memoized layout format to avoid repeated DB calls
+	private String layoutFormat;
+
 	public NameSupport() {
 		if (singleton == null) {
 			singleton = this;
@@ -58,6 +61,14 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 			return;
 		}
 		Context.getAdministrationService().addGlobalPropertyListener(singleton);
+
+		// Read and cache the configured name format GP
+		String formatGp = Context.getAdministrationService()
+		        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT);
+		if (StringUtils.isNotBlank(formatGp)) {
+			this.layoutFormat = formatGp;
+		}
+
 		// Get configured name template to override the existing one if any
 		String layoutTemplateXml = Context.getAdministrationService()
 		        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_TEMPLATE);
@@ -104,13 +115,14 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 	}
 
 	/**
-	 * @return Returns the defaultLayoutFormat
+	 * @return Returns the memoized default layout format, falling back to defaultLayoutFormat
 	 */
 	@Override
 	public String getDefaultLayoutFormat() {
-		String ret = Context.getAdministrationService()
-		        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT);
-		return (ret != null && ret.length() > 0) ? ret : defaultLayoutFormat;
+		if (StringUtils.isNotBlank(layoutFormat)) {
+			return layoutFormat;
+		}
+		return defaultLayoutFormat;
 	}
 
 	/**
@@ -118,7 +130,8 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 	 */
 	@Override
 	public boolean supportsPropertyName(String propertyName) {
-		return OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_TEMPLATE.equals(propertyName);
+		return OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_TEMPLATE.equals(propertyName)
+		        || OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT.equals(propertyName);
 	}
 
 	/**
@@ -126,9 +139,17 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 	 */
 	@Override
 	public void globalPropertyChanged(GlobalProperty newValue) {
-		NameTemplate nameTemplate = deserializeXmlTemplate(newValue.getPropertyValue());
-		if (nameTemplate != null) {
-			updateLayoutTemplates(nameTemplate);
+		if (newValue == null) {
+			return;
+		}
+
+		if (OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_TEMPLATE.equals(newValue.getProperty())) {
+			NameTemplate nameTemplate = deserializeXmlTemplate(newValue.getPropertyValue());
+			if (nameTemplate != null) {
+				updateLayoutTemplates(nameTemplate);
+			}
+		} else if (OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT.equals(newValue.getProperty())) {
+			this.layoutFormat = newValue.getPropertyValue();
 		}
 	}
 
@@ -137,6 +158,8 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 	 */
 	@Override
 	public void globalPropertyDeleted(String propertyName) {
-
+		if (OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT.equals(propertyName)) {
+			this.layoutFormat = null;
+		}
 	}
 }
--- a/api/src/main/java/org/openmrs/layout/name/NameSupport.java
+++ b/api/src/main/java/org/openmrs/layout/name/NameSupport.java
@@ -17,6 +17,7 @@ import org.apache.commons.lang3.StringEscapeUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.openmrs.GlobalProperty;
 import org.openmrs.api.APIException;
+import org.openmrs.api.EventListeners;
 import org.openmrs.api.GlobalPropertyListener;
 import org.openmrs.api.context.Context;
 import org.openmrs.layout.LayoutSupport;
@@ -57,10 +58,20 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 	 * Initializes layout templates with a custom template configured via the "layout.name.template" GP.
 	 */
 	private void init() {
+
+		// OpenMRS wipes listeners on context refresh. We check the global list to see if we need to re-register ourselves, avoiding duplicate registrations.
+		EventListeners eventListeners = new EventListeners();
+		List<GlobalPropertyListener> currentListeners = eventListeners.getGlobalPropertyListeners();
+
+		if (currentListeners == null || !currentListeners.contains(singleton)) {
+			Context.getAdministrationService().addGlobalPropertyListener(singleton);
+
+		}
+
+		// now the listener is secure and guards the expensive database reads.
 		if (initialized) {
 			return;
 		}
-		Context.getAdministrationService().addGlobalPropertyListener(singleton);
 
 		// Read and cache the configured name format GP
 		String formatGp = Context.getAdministrationService()
--- a/api/src/main/java/org/openmrs/layout/name/NameSupport.java
+++ b/api/src/main/java/org/openmrs/layout/name/NameSupport.java
@@ -17,7 +17,6 @@ import org.apache.commons.lang3.StringEscapeUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.openmrs.GlobalProperty;
 import org.openmrs.api.APIException;
-import org.openmrs.api.EventListeners;
 import org.openmrs.api.GlobalPropertyListener;
 import org.openmrs.api.context.Context;
 import org.openmrs.layout.LayoutSupport;
@@ -32,9 +31,9 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 
 	private static final Logger log = LoggerFactory.getLogger(NameSupport.class);
 
-	private static NameSupport singleton;
+	private static volatile NameSupport singleton;
 
-	private boolean initialized = false;
+	private volatile boolean initialized = false;
 
 	// Memoized layout format to avoid repeated DB calls
 	private String layoutFormat;
@@ -59,27 +58,12 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 	 */
 	private void init() {
 
-		// OpenMRS wipes listeners on context refresh. We check the global list to see if we need to re-register ourselves, avoiding duplicate registrations.
-		EventListeners eventListeners = new EventListeners();
-		List<GlobalPropertyListener> currentListeners = eventListeners.getGlobalPropertyListeners();
-
-		if (currentListeners == null || !currentListeners.contains(singleton)) {
-			Context.getAdministrationService().addGlobalPropertyListener(singleton);
-
-		}
-
 		// now the listener is secure and guards the expensive database reads.
 		if (initialized) {
 			return;
 		}
 
-		// Read and cache the configured name format GP
-		String formatGp = Context.getAdministrationService()
-		        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT);
-		if (StringUtils.isNotBlank(formatGp)) {
-			this.layoutFormat = formatGp;
-		}
-
+		Context.getAdministrationService().addGlobalPropertyListener(singleton);
 		// Get configured name template to override the existing one if any
 		String layoutTemplateXml = Context.getAdministrationService()
 		        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_TEMPLATE);
@@ -126,7 +110,7 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 	}
 
 	/**
-	 * @return Returns the memoized default layout format, falling back to defaultLayoutFormat
+	 * @return Returns the configured default layout format
 	 */
 	@Override
 	public String getDefaultLayoutFormat() {
@@ -159,7 +143,7 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 			if (nameTemplate != null) {
 				updateLayoutTemplates(nameTemplate);
 			}
-		} else if (OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT.equals(newValue.getProperty())) {
+		} else {
 			this.layoutFormat = newValue.getPropertyValue();
 		}
 	}
--- a/api/src/main/java/org/openmrs/layout/name/NameSupport.java
+++ b/api/src/main/java/org/openmrs/layout/name/NameSupport.java
@@ -36,7 +36,7 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 	private volatile boolean initialized = false;
 
 	// Memoized layout format to avoid repeated DB calls
-	private String layoutFormat;
+	private volatile String layoutFormat;
 
 	public NameSupport() {
 		if (singleton == null) {
@@ -58,13 +58,19 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 	 */
 	private void init() {
 
-		// now the listener is secure and guards the expensive database reads.
 		if (initialized) {
 			return;
 		}
 
 		Context.getAdministrationService().addGlobalPropertyListener(singleton);
-		// Get configured name template to override the existing one if any
+
+		// Get the configured name format so the cache starts out matching the database
+		String formatGp = Context.getAdministrationService()
+		        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT);
+		if (StringUtils.isNotBlank(formatGp)) {
+			this.layoutFormat = formatGp;
+		}
+
 		String layoutTemplateXml = Context.getAdministrationService()
 		        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_TEMPLATE);
 		NameTemplate nameTemplate = deserializeXmlTemplate(layoutTemplateXml);
--- a/api/src/test/java/org/openmrs/layout/name/NameSupportTest.java
+++ b/api/src/test/java/org/openmrs/layout/name/NameSupportTest.java
@@ -1,0 +1,57 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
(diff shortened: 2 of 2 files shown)


nsalifu:
@claude  @ibacher @dkayiwa @wikumChamith you may review. thx!

api/src/main/java/org/openmrs/layout/name/NameSupport.java
```diff
@@ -35,6 +35,9 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 
 	private boolean initialized = false;
 
+	// Memoized layout format to avoid repeated DB calls
+	private String layoutFormat;
```
wikumChamith: This cache is going to go stale when `updateGlobalProperty` gets called, since that method never calls `notifyGlobalPropertyChange()`. I think we'll need a separate ticket for that.

@ibacher , @dkayiwa, what do you think?

wikumChamith (commented):
@nsalifu welcome back :)

api/src/main/java/org/openmrs/layout/name/NameSupport.java
nsalifu: thanks @wikumChamith , i've been away from openmrs for a while. Sorry about that, i've been doing ERP implementations lately. 

and thanks for the review. 
I can confirm: `updateGlobalProperty()` calls `dao.saveGlobalProperty(gp)` directly and never calls `notifyGlobalPropertyChange()`, unlike saveGlobalProperty(). So any caller using `updateGlobalProperty()` on `layout.name.format` would leave layoutFormat stale.

One might be tempted to make` updateGlobalProperty()` route through `saveGlobalProperty()` so it notifies listeners too, however this may be a change affecting every registered `GlobalPropertyListener`

api/src/main/java/org/openmrs/layout/name/NameSupport.java
```diff
@@ -58,6 +61,14 @@ private void init() {
 			return;
 		}
 		Context.getAdministrationService().addGlobalPropertyListener(singleton);
+
+		// Read and cache the configured name format GP
+		String formatGp = Context.getAdministrationService()
+		        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT);
+		if (StringUtils.isNotBlank(formatGp)) {
+			this.layoutFormat = formatGp;
+		}
```
dkayiwa, line 72: This needs sorting before merge: caching the format only works if the listener registration survives, and it doesn't. `applicationContext-service.xml` declares a `clearOpenmrsEventListeners` bean that empties the static listener list on every Spring context refresh, and `openmrsEventListeners` then re-adds only the listeners declared in that file. `NameSupport` registers itself programmatically here instead, and `init()` returns at the `initialized` check, so after the first refresh it is silently no longer a listener and never re-registers. OpenMRS refreshes the context on every module start and stop (`ModuleFactory.startModuleInternal` and `WebModuleUtil.stopModule` both end up in `ModuleUtil.refreshApplicationContext`, which does `ctx.stop()`/`ctx.close()` and rebuilds every bean).

On master this was invisible, because `getDefaultLayoutFormat()` re-read the property on each call. With the value cached, `layout.name.format` freezes at whatever it held before the refresh: an admin edits it, gets no error, and `PersonName.getFullName()` keeps using the old format until the server is restarted.

Reproduced with a context-sensitive test that does exactly what `clearOpenmrsEventListeners` does (`new EventListeners().setGlobalPropertyListenersToEmpty(false)`) and then saves `layout.name.format` again:

```
master:      save "long" -> long,  clear + save "short" -> short
this branch: save "long" -> long,  clear + save "short" -> long
```

This bites the listener fix itself, not just the cache: after a refresh nothing fires for `layout.name.template` either, so "changing the GP causes the listener to fire" only holds until the first module start or stop. `AddressSupport` registers the same way and has the same gap, so mirroring it doesn't make the cache safe.

I'd take the registration out from behind the `initialized` guard so every `getInstance()` re-establishes it, leaving only the initial state load guarded. `AdministrationServiceImpl.addGlobalPropertyListener` appends without checking, so it would need a contains check, which is what `EventListeners.setGlobalPropertyListeners` already does for the declarative path. If that is more than this PR should carry, the smaller route is to drop the memoization and land just the `supportsPropertyName` / `globalPropertyChanged` / `globalPropertyDeleted` change, then memoize in a follow-up once there is a registration that survives a refresh.

nsalifu: Thanks @dkayiwa for the review.

Instead of completely dropping the memoization in this PR, I updated i`nit()` to safely re-register the listener every time `getInstance()` is called. I added a check against 
`new EventListeners().getGlobalPropertyListeners()` to ensure we only append singleton if it was actually wiped out, avoiding duplicate registrations. The expensive database reads are still safely hidden behind if (initialized) return;.

I've pushed the updated logic. I would appreciate feedback. thx!

api/src/main/java/org/openmrs/layout/name/NameSupport.java
```diff
@@ -104,31 +126,41 @@ private NameTemplate deserializeXmlTemplate(String xml) {
...
+		if (newValue == null) {
+			return;
+		}
+
+		if (OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_TEMPLATE.equals(newValue.getProperty())) {
+			NameTemplate nameTemplate = deserializeXmlTemplate(newValue.getPropertyValue());
+			if (nameTemplate != null) {
+				updateLayoutTemplates(nameTemplate);
+			}
+		} else if (OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT.equals(newValue.getProperty())) {
```
ibacher: We don't really need this check; the `supportsPropertyName()` already filters this.

api/src/main/java/org/openmrs/layout/name/NameSupport.java
```diff
@@ -104,31 +126,41 @@ private NameTemplate deserializeXmlTemplate(String xml) {
 	}
 
 	/**
-	 * @return Returns the defaultLayoutFormat
+	 * @return Returns the memoized default layout format, falling back to defaultLayoutFormat
```
ibacher:
```suggestion
	 * @return Returns the configured default layout format
```

api/src/main/java/org/openmrs/layout/name/NameSupport.java
```diff
@@ -35,6 +36,9 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
 
 	private boolean initialized = false;
```
ibacher: Both of these should be marked `volatile`.

api/src/main/java/org/openmrs/layout/name/NameSupport.java
```diff
@@ -54,10 +58,28 @@ public static NameSupport getInstance() {
...
 	private void init() {
+
+		// OpenMRS wipes listeners on context refresh. We check the global list to see if we need to re-register ourselves, avoiding duplicate registrations.
+		EventListeners eventListeners = new EventListeners();
+		List<GlobalPropertyListener> currentListeners = eventListeners.getGlobalPropertyListeners();
+
+		if (currentListeners == null || !currentListeners.contains(singleton)) {
+			Context.getAdministrationService().addGlobalPropertyListener(singleton);
+
+		}
```
ibacher: While this does kind of fix a real bug, it does so in a really bad way. The correct way to do this is make `administrationService#addGlobalPropertyListener()` do nothing if the listener is already registered, but that's a different ticket...

Besides, with the way the code was written before this, there's only a small chance of leaking `NameSupport` objects.

nsalifu: thanks for the review @ibacher , I have update the PR accordingly: 
I reverted  the `init()` method  back to its original approach so we aren't hacking the listener list here.
Marked the instance vars as `valatile` for thread safety as mention, and simplified the condition in globalPropertyChanged() to a simple else block as suggested.
commit: https://github.com/openmrs/openmrs-core/pull/6523/commits/db85a503e81b61863a8b48dffac80f4b7406189a

for this: 
`The correct way to do this is make administrationService#addGlobalPropertyListener() do nothing if the listener is already registered, but that's a different ticket.` 

I shall be waiting for the ticket, you can create and assign to me.  
Thanks a bunch.

api/src/main/java/org/openmrs/layout/name/NameSupport.java
```diff
@@ -31,9 +31,12 @@ public class NameSupport extends LayoutSupport<NameTemplate> implements GlobalPr
...
 	private static final Logger log = LoggerFactory.getLogger(NameSupport.class);
 
-	private static NameSupport singleton;
+	private static volatile NameSupport singleton;
 
-	private boolean initialized = false;
+	private volatile boolean initialized = false;
+
+	// Memoized layout format to avoid repeated DB calls
+	private String layoutFormat;
```
dkayiwa: `singleton` and `initialized` are volatile now, but `layoutFormat` is the field that actually crosses threads here: the listener writes it on whichever request thread saved the property, and every thread formatting a person name reads it. With no volatile there is no happens-before between that write and those reads, so a reader is not guaranteed to pick up the new format, which is the symptom the ticket exists to fix. Worth marking this one too.

```suggestion
	private volatile String layoutFormat;
```

api/src/main/java/org/openmrs/layout/name/NameSupport.java
```diff
@@ -54,9 +57,12 @@ public static NameSupport getInstance() {
 	 * Initializes layout templates with a custom template configured via the "layout.name.template" GP.
 	 */
 	private void init() {
+
+		// now the listener is secure and guards the expensive database reads.
 		if (initialized) {
 			return;
 		}
+
```
dkayiwa, line 64: This comment arrived with the re-registration commit, where it sat after the re-register block and made sense. After the revert it is the first thing in the method with nothing before it, and what it claims is the opposite of what happens: the registration is exactly as fragile as it was on master. The two new blank lines around the guard are not something the ticket needs either.

```suggestion
		if (initialized) {
			return;
		}
```

api/src/main/java/org/openmrs/layout/name/NameSupport.java
```diff
@@ -54,9 +57,12 @@ public static NameSupport getInstance() {
...
 		Context.getAdministrationService().addGlobalPropertyListener(singleton);
```
dkayiwa, line 65: This needs fixing before merge. `layoutFormat` is only ever written from `globalPropertyChanged`, and `init()` no longer reads `layout.name.format`, so on a fresh JVM the cache is null and `getDefaultLayoutFormat()` falls straight through to the Spring-wired `defaultLayoutFormat`. Any deployment whose `layout.name.format` differs from that wired value renders person names with the wired one after every restart, until somebody re-saves the property. Nothing throws and nothing logs.

legacyui is where this bean actually lives in a running server, and it wires `defaultLayoutFormat` to `short` next to five templates (Spain, LatinAmerica, GivenFamily, Short, Long), so an implementation running `layout.name.format = long` goes back to short names on every restart.

I gave the singleton a wired default of `short`, saved the property as `long`, reset the in-memory state the way a JVM restart would, then read the format back:

```
master:      long
this branch: short
```

Those four lines were there in 77d19c9 and came out with the `init()` revert in db85a50. I could not find a review comment asking for that part, so I read it as collateral, and putting them back fixes this on its own.

Worth a test that starts from the property already sitting in the database instead of saving it during the test. Every test here currently reaches the value through a `saveGlobalProperty` that fires the listener, which is the one path that does work.

The listener not surviving a context refresh is a separate problem and is still open in the older thread on this method.

```suggestion
		Context.getAdministrationService().addGlobalPropertyListener(singleton);

		// Get the configured name format so the cache starts out matching the database
		String formatGp = Context.getAdministrationService()
		        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT);
		if (StringUtils.isNotBlank(formatGp)) {
			this.layoutFormat = formatGp;
		}
```

api/src/main/java/org/openmrs/layout/name/NameSupport.java
dkayiwa: The re-registration went in at d714b65 and came back out at db85a50, so the head is the memoization sitting on top of master's registration, which is the combination this thread was about. Same probe against the head:

```
master:      save "long" -> long,  clear listeners + save "short" -> short
this branch: save "long" -> long,  clear listeners + save "short" -> long
```

With @ibacher parking the `addGlobalPropertyListener` idempotency as its own ticket, the route left for this PR is the smaller one from above: land `supportsPropertyName`, `globalPropertyChanged` and `globalPropertyDeleted`, leave `getDefaultLayoutFormat()` reading the property, and memoize once there is a registration that survives a refresh. That is the one I would take, and it also takes the missing startup seed and the field visibility point off the table, since there would be no cache to keep in step. Landing the idempotency ticket first and rebasing this on top works too, but it puts this PR behind a ticket that does not exist yet.

api/src/main/java/org/openmrs/layout/name/NameSupport.java
nsalifu: thanks for the review @dkayiwa, i updated the PR accordingly.

api/src/main/java/org/openmrs/layout/name/NameSupport.java
dkayiwa: Dropping the memoization is not actually a route, so scratch the smaller one I offered above. If `getDefaultLayoutFormat()` goes back to reading the property, there is no derived state left for `globalPropertyChanged` and `globalPropertyDeleted` to refresh on a format change, so matching the format property in `supportsPropertyName` would buy a listener that fires and does nothing, which is the dead code the ticket exists to remove.

That leaves making the registration survive, and one detail there decides whether this can simply wait on @ibacher's ticket. A contains check inside `addGlobalPropertyListener` does not on its own bring the listener back: `clearOpenmrsEventListeners` empties the list rather than double registering into it, so the only thing that restores `NameSupport` is calling `addGlobalPropertyListener` again. That means moving that call out from behind the `initialized` guard and leaving the two property reads and the template load under it. The contains check is what makes that safe, since `AdministrationServiceImpl.addGlobalPropertyListener` is a bare `add` today and an unguarded call would append a duplicate on every `getInstance()`.

So the two pieces go together: the idempotency fix on the service, and the one line move here. Either that ticket lands first and this rebases onto it, or both go in here.

Same probe against 0b23834, so the head is still stale after a refresh:

```
this branch: save "long" -> long,  clear listeners + save "short" -> long
```

api/src/main/java/org/openmrs/layout/name/NameSupport.java
dkayiwa: There is still no test on the seed: the two tests in `NameSupportTest` that touch `layoutFormat` both set it through `globalPropertyChanged`, and no dataset under `api/src/test/resources` carries `layout.name.format`, so nothing asserts that a value already sitting in the database reaches `getDefaultLayoutFormat()`. Those four lines came out once between 77d19c9 and db85a50 with nothing noticing, which is the argument for a context sensitive test that starts from the property already saved and reads the format before anything saves it.

ibacher (approved):

ibacher:
I think the code here is fine. Could you write the test Daniel requested above?

api/src/main/java/org/openmrs/layout/name/NameSupport.java
nsalifu: > There is still no test on the seed: the two tests in `NameSupportTest` that touch `layoutFormat` both set it through `globalPropertyChanged`, and no dataset under `api/src/test/resources` carries `layout.name.format`, so nothing asserts that a value already sitting in the database reaches `getDefaultLayoutFormat()`. Those four lines came out once between [77d19c9](https://github.com/openmrs/openmrs-core/commit/77d19c902fcf3b6790745bb8b37fd5f9f2132358) and [db85a50](https://github.com/openmrs/openmrs-core/commit/db85a503e81b61863a8b48dffac80f4b7406189a) with nothing noticing, which is the argument for a context sensitive test that starts from the property already saved and reads the format before anything saves it.

@dkayiwa I have pushed the test here: [commit](https://github.com/openmrs/openmrs-core/pull/6523/commits/7b37440661302dba1448fd06d510a29929713a85)

nsalifu:
> I think the code here is fine. Could you write the test Daniel requested above?

thanks @ibacher , i added the test ([here](https://github.com/openmrs/openmrs-core/pull/6523/commits/7b37440661302dba1448fd06d510a29929713a85))

api/src/main/java/org/openmrs/layout/name/NameSupport.java
```diff
@@ -54,11 +57,20 @@ public static NameSupport getInstance() {
 	 * Initializes layout templates with a custom template configured via the "layout.name.template" GP.
 	 */
 	private void init() {
+
```
ibacher:
```suggestion
```

api/src/test/java/org/openmrs/layout/name/NameSupportContextSensitiveTest.java
```diff
@@ -0,0 +1,41 @@
+package org.openmrs.layout.name;
```
ibacher, line 10: This needs the license header as well

nsalifu:
Hi @dkayiwa , could you please take a review look once again.

api/src/test/java/org/openmrs/layout/name/NameSupportContextSensitiveTest.java
```diff
@@ -0,0 +1,50 @@
...
+			initializedField.setBoolean(nameSupport, false);
+			layoutFormatField.set(nameSupport, null);
+
+			String format = NameSupport.getInstance().getDefaultLayoutFormat();
+
+			assertEquals("long", format);
+		} finally {
+			initializedField.setBoolean(nameSupport, previousInitialized);
+			layoutFormatField.set(nameSupport, previousLayoutFormat);
+		}
```
dkayiwa, line 54: The cleanup here is partial, so the singleton stays changed for everything that runs after this class in the same surefire JVM. `setDefaultLayoutFormat("short")` never goes back, and the `layoutFormat` restored in the `finally` was read after the first `getInstance()` had already seeded it from this test's own dataset, so what goes back is `long`, which no row backs once the transaction rolls back. `initialized` stays true, so nothing re-reads it.

A probe class in the same package that does nothing but dump the singleton, run on its own and then straight after this class with `-Dsurefire.runOrder=alphabetical`:

```
probe alone:     initialized=true layoutFormat=null defaultLayoutFormat=null getDefaultLayoutFormat=null nameSupportRegistrations=1
after this test: initialized=true layoutFormat=long defaultLayoutFormat=short getDefaultLayoutFormat=long nameSupportRegistrations=2
```

The api suite is green with it, so this is not a merge blocker. The cost lands on whoever writes the next test that reads `getDefaultLayoutFormat()`, well away from here. `AddressSupportTest` keeps an `@AfterEach` that resets its own singleton for the same reason.

Restoring the wired default and leaving the cache empty rather than pinning it puts the probe back on the baseline line above, and the test still fails with `expected: <long> but was: <short>` when the seed in `init()` is taken out again:

```suggestion
		NameSupport nameSupport = NameSupport.getInstance();

		Field initializedField = NameSupport.class.getDeclaredField("initialized");
		initializedField.setAccessible(true);
		Field layoutFormatField = NameSupport.class.getDeclaredField("layoutFormat");
		layoutFormatField.setAccessible(true);
		Field defaultFormatField = LayoutSupport.class.getDeclaredField("defaultLayoutFormat");
		defaultFormatField.setAccessible(true);

		boolean previousInitialized = initializedField.getBoolean(nameSupport);
		Object previousDefaultFormat = defaultFormatField.get(nameSupport);
		nameSupport.setDefaultLayoutFormat("short"); // mirrors legacyui's wiring

		try {
			// Simulate a fresh boot: clear the cache/init flag on the existing singleton
			initializedField.setBoolean(nameSupport, false);
			layoutFormatField.set(nameSupport, null);

			String format = NameSupport.getInstance().getDefaultLayoutFormat();

			assertEquals("long", format);
		} finally {
			initializedField.setBoolean(nameSupport, previousInitialized);
			layoutFormatField.set(nameSupport, null);
			defaultFormatField.set(nameSupport, previousDefaultFormat);
			Context.getAdministrationService().removeGlobalPropertyListener(nameSupport);
		}
```

That needs `org.openmrs.api.context.Context` and `org.openmrs.layout.LayoutSupport` added to the imports.

api/src/main/java/org/openmrs/layout/name/NameSupport.java
dkayiwa: Separate ticket is right, and it does not hold this one up. `updateGlobalProperty` is the only mutator on the service that writes straight to the dao without notifying, and nothing calls it: not core, not legacyui's settings and global property controllers, not the webservices.rest system setting resource, all of which go through `saveGlobalProperty` or `saveGlobalProperties`. So the staleness is open to API and module callers rather than to an admin clicking save.

api/src/main/java/org/openmrs/layout/name/NameSupport.java
dkayiwa: The blank line below the guard is still there, and `// Get configured name template to override the existing one if any` came out above the template read with nothing replacing it. Neither is something the ticket needs, so I would put the comment back and drop the line.

nsalifu:
thanks for the review and suggestions @dkayiwa , i've update the PR. you may take a look.

api/src/main/java/org/openmrs/layout/name/NameSupport.java
dkayiwa: Still the case at the head: once a module has been started or stopped, an edit to `layout.name.format` doesn't show in formatted names until a restart. There is a way out that needs neither of the two pieces above, though: read the format through `ConfigUtil.getGlobalProperty` instead of keeping a copy in `NameSupport`. `ConfigUtil` is declared in `openmrsEventListeners`, so every refresh registers it again, and it already handles change and delete; `GZIPFilter` moved its own cached fields onto it in TRUNK-6705. Here that means master's `getDefaultLayoutFormat()` with `ConfigUtil.getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_LAYOUT_NAME_FORMAT)` in place of the service call, and the seed in `init()`, the `layoutFormat` field and the format handling in the listener methods all go.

With the listener list rebuilt from the real `clearOpenmrsEventListeners` and `openmrsEventListeners` bean definitions, which is what a refresh does to it, and the default wired to `short` as legacyui has it:

```
                     statements per 100 reads   save long   refresh, save givenfamily   refresh, purge
master                         100                 long             givenfamily              short
this branch                      0                 long             long                     long
master + ConfigUtil              0                 long             givenfamily              short
```

@ibacher this drops the part of the acceptance criteria that has `supportsPropertyName` match the format, since `NameSupport` would no longer hold anything for that listener to invalidate. It also has one gap the branch doesn't: `purgeGlobalProperty` notifies before it deletes, so a name formatted inside that window caches the old value again until the next save, which the missed-set in #6453 would close.

The same rebuild makes the refresh case testable, which nothing covers yet:

<details><summary>Rebuilding the listener list the way a refresh does</summary>

```java
ConfigurableListableBeanFactory parent = ((ConfigurableApplicationContext) applicationContext).getBeanFactory();
DefaultListableBeanFactory fresh = new DefaultListableBeanFactory(parent);
for (String name : new String[] { "clearOpenmrsEventListeners", "openmrsEventListeners" }) {
	fresh.registerBeanDefinition(name,
	    ((AbstractBeanDefinition) parent.getMergedBeanDefinition(name)).cloneBeanDefinition());
}
fresh.getBean("openmrsEventListeners");
```

</details>

PR #6531, opened 2026-09-01
TRUNK-6787:Added hibernate disable auto commit property

sudhanshu-raj:
Added `hibernate.connection.provider_disables_autocommit=true` , to remove the redundant `SET autocommit=0` DB command, which helps to reduce network round-trips from 4 to 3 and reduce the overall latency too .

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6787

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
@@ -11,6 +11,8 @@ package org.openmrs.api.db.hibernate;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.sql.Connection;
+import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -26,8 +28,11 @@ import org.hibernate.HibernateException;
 import org.hibernate.Interceptor;
 import org.hibernate.boot.Metadata;
 import org.hibernate.boot.spi.BootstrapContext;
+import org.hibernate.cfg.Environment;
+import org.hibernate.engine.jdbc.connections.spi.ConnectionProvider;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.integrator.spi.Integrator;
+import org.hibernate.internal.util.config.ConfigurationHelper;
 import org.hibernate.service.ServiceRegistry;
 import org.hibernate.service.spi.SessionFactoryServiceRegistry;
 import org.jspecify.annotations.NonNull;
@@ -225,6 +230,7 @@ public class HibernateSessionFactoryBean extends LocalSessionFactoryBean impleme
 	public void integrate(Metadata metadata, BootstrapContext bootstrapContext, SessionFactoryImplementor sessionFactory) {
 		this.metadata = metadata;
 		generateEnversAuditTables(metadata, bootstrapContext.getServiceRegistry());
+		assertProviderDisablesAutocommitPremise(bootstrapContext.getServiceRegistry(), sessionFactory);
 	}
 
 	@Override
@@ -247,4 +253,31 @@ public class HibernateSessionFactoryBean extends LocalSessionFactoryBean impleme
 			throw new APIException("An error occurred while initializing the Envers audit tables", e);
 		}
 	}
+
+	private void assertProviderDisablesAutocommitPremise(ServiceRegistry serviceRegistry,
+	        SessionFactoryImplementor sessionFactory) {
+		Map<String, Object> settings = sessionFactory.getProperties();
+		boolean declared = ConfigurationHelper.getBoolean(Environment.CONNECTION_PROVIDER_DISABLES_AUTOCOMMIT, settings,
+		    false);
+		if (!declared) {
+			return;
+		}
+		ConnectionProvider provider = serviceRegistry.getService(ConnectionProvider.class);
+		try {
+			Connection conn = provider.getConnection();
+			try {
+				if (conn.getAutoCommit()) {
+					throw new HibernateException(
+					        "provider_disables_autocommit=true but the pool returned a connection with autoCommit=true; "
+					                + "configure the pool to disable autocommit at checkout or remove the property");
+				}
+			} finally {
+				provider.closeConnection(conn);
+			}
+		} catch (SQLException e) {
+			throw new HibernateException(
+			        "Failed to obtain a connection from the pool to check autoCommit state: " + e.getMessage(), e);
+		}
+	}
+
 }
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
@@ -28,11 +28,9 @@ import org.hibernate.HibernateException;
 import org.hibernate.Interceptor;
 import org.hibernate.boot.Metadata;
 import org.hibernate.boot.spi.BootstrapContext;
-import org.hibernate.cfg.JdbcSettings;
 import org.hibernate.engine.jdbc.connections.spi.ConnectionProvider;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.integrator.spi.Integrator;
-import org.hibernate.internal.util.config.ConfigurationHelper;
 import org.hibernate.service.ServiceRegistry;
 import org.hibernate.service.spi.SessionFactoryServiceRegistry;
 import org.jspecify.annotations.NonNull;
@@ -256,10 +254,8 @@ public class HibernateSessionFactoryBean extends LocalSessionFactoryBean impleme
 
 	private void assertProviderDisablesAutocommitPremise(ServiceRegistry serviceRegistry,
 	        SessionFactoryImplementor sessionFactory) {
-		Map<String, Object> settings = sessionFactory.getProperties();
-		boolean declared = ConfigurationHelper.getBoolean(JdbcSettings.CONNECTION_PROVIDER_DISABLES_AUTOCOMMIT, settings,
-		    false);
-		if (!declared) {
+
+		if (!sessionFactory.getSessionFactoryOptions().doesConnectionProviderDisableAutoCommit()) {
 			return;
 		}
 		ConnectionProvider provider = serviceRegistry.getService(ConnectionProvider.class);
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
@@ -28,7 +28,7 @@ import org.hibernate.HibernateException;
 import org.hibernate.Interceptor;
 import org.hibernate.boot.Metadata;
 import org.hibernate.boot.spi.BootstrapContext;
-import org.hibernate.cfg.Environment;
+import org.hibernate.cfg.JdbcSettings;
 import org.hibernate.engine.jdbc.connections.spi.ConnectionProvider;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.integrator.spi.Integrator;
@@ -257,7 +257,7 @@ public class HibernateSessionFactoryBean extends LocalSessionFactoryBean impleme
 	private void assertProviderDisablesAutocommitPremise(ServiceRegistry serviceRegistry,
 	        SessionFactoryImplementor sessionFactory) {
 		Map<String, Object> settings = sessionFactory.getProperties();
-		boolean declared = ConfigurationHelper.getBoolean(Environment.CONNECTION_PROVIDER_DISABLES_AUTOCOMMIT, settings,
+		boolean declared = ConfigurationHelper.getBoolean(JdbcSettings.CONNECTION_PROVIDER_DISABLES_AUTOCOMMIT, settings,
 		    false);
 		if (!declared) {
 			return;
--- a/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
+++ b/api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
@@ -265,7 +265,9 @@ public class HibernateSessionFactoryBean extends LocalSessionFactoryBean impleme
 				if (conn.getAutoCommit()) {
 					throw new HibernateException(
 					        "provider_disables_autocommit=true but the pool returned a connection with autoCommit=true; "
-					                + "configure the pool to disable autocommit at checkout or remove the property");
+					                + "remove connection.autocommit=true from openmrs-runtime.properties, or set "
+					                + "connection.provider_disables_autocommit=false there if this pool really must hand out "
+					                + "auto-commit-enabled connections");
 				}
 			} finally {
 				provider.closeConnection(conn);
--- a/api/src/main/resources/hibernate.default.properties
+++ b/api/src/main/resources/hibernate.default.properties
@@ -5,6 +5,8 @@ hibernate.connection.username=test
 hibernate.connection.password=test
 hibernate.connection.driver_class=com.mysql.jdbc.Driver
 hibernate.connection.url=jdbc:mysql://localhost:3306/openmrs?autoReconnect=true
+# safe because hibernate-c3p0 disables auto-commit at every connection checkout;
+# HibernateSessionFactoryBean refuses to start if that stops being true, see TRUNK-6787.
 hibernate.connection.provider_disables_autocommit=true
 
 # Hibernate specific connection/debug properties
--- a/api/src/main/resources/hibernate.default.properties
+++ b/api/src/main/resources/hibernate.default.properties
@@ -5,6 +5,7 @@ hibernate.connection.username=test
 hibernate.connection.password=test
 hibernate.connection.driver_class=com.mysql.jdbc.Driver
 hibernate.connection.url=jdbc:mysql://localhost:3306/openmrs?autoReconnect=true
+hibernate.connection.provider_disables_autocommit=true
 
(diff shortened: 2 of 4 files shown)


sudhanshu-raj:
@ibacher @wikumChamith @dkayiwa , please review this.

api/src/main/resources/hibernate.default.properties
```diff
@@ -5,6 +5,7 @@ hibernate.connection.username=test
 hibernate.connection.password=test
 hibernate.connection.driver_class=com.mysql.jdbc.Driver
 hibernate.connection.url=jdbc:mysql://localhost:3306/openmrs?autoReconnect=true
+hibernate.connection.provider_disables_autocommit=true
```
dkayiwa, line 10: This needs a guard before it goes in. The property is an assertion Hibernate cannot verify, and nothing in the tree ties it to the thing that makes it true, which is `C3P0ConnectionProvider#prepareConnection` calling `setAutoCommit(false)` on every checkout. Hibernate's own javadoc for the setting says it plainly: "It is inappropriate to set this value to true when the connections returned by the provider do not, in fact, have auto-commit disabled. Doing so may lead to Hibernate executing SQL operations outside the scope of any transaction."

Concretely, this line turns `hibernate.connection.autocommit=true` from harmless into silent loss of transactions. `HibernateSessionFactoryBean#afterPropertiesSet` copies runtime properties verbatim into the Hibernate configuration, so that override is reachable, and until now Hibernate compensated for it: `AbstractLogicalConnectionImplementor#begin` read the connection's real state and turned auto-commit off. With this line it skips that call. I ran `LocationServiceTest` on this branch with that override in place, and it goes from 84/84 green with this line commented out to 10 failures and 2 errors with it in, all of them writes from earlier rolled-back tests still sitting in the database (`getAllLocations_shouldReturnAllLocationsIncludingRetired` expected 7, got 18). Nothing throws, the data is just wrong.

No repo in the openmrs org sets `connection.autocommit` today, so this is not a live outage. It is the hole the last acceptance criterion on TRUNK-6787 asks you to close, and it is worth closing because c3p0 is the only pool the premise holds for: hibernate-c3p0's own source notes that disabling auto-commit at checkout is a historical c3p0 behaviour and "not what we do with Agroal or Hikari".

The cheapest fix is the startup assertion that criterion offers. `HibernateSessionFactoryBean` already does startup work in `integrate(...)`, so when this property is on, take one connection from the `ConnectionProvider` in `bootstrapContext.getServiceRegistry()` and refuse to start if it comes back with auto-commit enabled. I prototyped exactly that on this branch: the context starts normally when the premise holds, and the `sessionFactory` bean fails with a readable message when it does not, instead of quietly running without transactions. That covers deployments, which a test cannot, and it runs in every context-sensitive test, so it doubles as the regression guard.

A couple of comment lines above the property naming the c3p0 premise would also help whoever changes the pool next, the way the `checkoutTimeout` block below does.

api/src/test/java/org/openmrs/api/db/hibernate/ConnectionPoolConfigTest.java
```diff
@@ -61,4 +63,13 @@ public void shouldApplyConfiguredConnectionPoolLimits() {
...
 		assertEquals(50, pool.getMaxPoolSize());
 	}
+
+	@Test
+	public void shouldHaveProviderDisablesAutocommitConfigured() {
+		Map<String, Object> settings = sessionFactory.getProperties();
+
+		assertEquals("true", String.valueOf(settings.get(Environment.CONNECTION_PROVIDER_DISABLES_AUTOCOMMIT)),
+		    "hibernate.connection.provider_disables_autocommit must be true so Hibernate skips the redundant setAutoCommit()");
+	}
```
dkayiwa, line 79: This one cannot fail in the situation it exists for. It reads back the value the same commit wrote into `hibernate.default.properties`, so it stays green whether or not connections actually arrive with auto-commit disabled. In the broken-premise run I describe on the properties file, this test passed while `LocationServiceTest` went red from rollbacks that no longer rolled anything back.

Asserting the connection gives you the check the ticket asks for, and it does fail the moment the premise stops holding. I ran it both ways on your branch:

```suggestion
	@Test
	public void shouldHandOutConnectionsWithAutoCommitDisabled() throws SQLException {
		ConnectionProvider connectionProvider = ((SessionFactoryImplementor) sessionFactory).getServiceRegistry()
		        .getService(ConnectionProvider.class);

		Connection connection = connectionProvider.getConnection();
		try {
			assertFalse(connection.getAutoCommit(),
			    "connections from the pool have auto-commit enabled, so hibernate.connection.provider_disables_autocommit is no longer safe to assert");
		} finally {
			connectionProvider.closeConnection(connection);
		}
	}
```

It needs `java.sql.Connection`, `java.sql.SQLException` and `assertFalse` imported, and leaves `java.util.Map` and `org.hibernate.cfg.Environment` unused, which fails `spotless:check` on CI if they are not dropped.

api/src/main/resources/hibernate.default.properties
sudhanshu-raj: Thnx for detail feedback,  yeah this is can  happen, we need a guard or kind of double sure that. Added the startup assertion to verify this, it fetches a connection and checks if `autoCommit=true `, if so then error will be thrown.

api/src/test/java/org/openmrs/api/db/hibernate/ConnectionPoolConfigTest.java
sudhanshu-raj: After adding the assertion check in` HibernateSessionFactoryBean#integrate` , i think we  not gonna need this new test case because it never gonna fail because if someone sets `hibernate.connection.autocommit=true` , that assert method will get failed on startup already giving the signal to either remove this or make it false, the only case it gonna fail is if we didn't mention this `hibernate.connection.provider_disables_autocommit=true` or make it false, and `hibernate.connection.autocommit=true` still, but I assume if we are doing this then we have intention to make auto commit true . And keeping this `shouldHaveProviderDisablesAutocommitConfigured` because it guard that `provider_disables_autocommit=true` is actually present in the resolved `SessionFactory ` config , and if this property is commented out assert method returns early only and nothing catches it, but this test case will catch that and make sure we put disable autocommit property.

api/src/test/java/org/openmrs/api/db/hibernate/ConnectionPoolConfigTest.java
```diff
@@ -61,4 +63,13 @@ public void shouldApplyConfiguredConnectionPoolLimits() {
 		assertEquals(30000, pool.getCheckoutTimeout());
 		assertEquals(50, pool.getMaxPoolSize());
 	}
+
+	@Test
+	public void shouldHaveProviderDisablesAutocommitConfigured() {
+		Map<String, Object> settings = sessionFactory.getProperties();
```
dkayiwa, line 75: Worth giving this the same escape hatch as `shouldApplyConfiguredConnectionPoolLimits` five lines up, though it is not something I would hold the merge for. TRUNK-6787 asks that the property stay overridable in runtime properties, and turning it off is precisely the remedy your startup assertion points a deployer at when their pool really does hand out auto-commit-enabled connections. As written, taking that remedy also turns the build red, on a test that has nothing to do with whatever they were changing.

I dropped `hibernate.connection.provider_disables_autocommit=false` into a runtime properties file and ran the class on your branch: this test fails with `expected: <true> but was: <false>`, while the sibling skips cleanly. With the guard below it skips as well, and with no override in place the class still passes 2/2.

```suggestion
		// as above, a machine-local openmrs-runtime.properties may legitimately turn this off
		assumeTrue(
		    Stream.of("hibernate.connection.provider_disables_autocommit", "connection.provider_disables_autocommit")
		            .noneMatch(runtimeProperties::containsKey),
		    "skipped because local runtime properties override provider_disables_autocommit");

		Map<String, Object> settings = sessionFactory.getProperties();
```

Both spellings are needed because `afterPropertiesSet` mirrors a bare `connection.*` runtime property under a `hibernate.` prefix, which is why the sibling lists both forms too.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
```diff
@@ -247,4 +253,31 @@ private void generateEnversAuditTables(Metadata metadata, ServiceRegistry servic
...
 	}
+
+	private void assertProviderDisablesAutocommitPremise(ServiceRegistry serviceRegistry,
+	        SessionFactoryImplementor sessionFactory) {
+		Map<String, Object> settings = sessionFactory.getProperties();
+		boolean declared = ConfigurationHelper.getBoolean(JdbcSettings.CONNECTION_PROVIDER_DISABLES_AUTOCOMMIT, settings,
+		    false);
+		if (!declared) {
+			return;
+		}
```
dkayiwa: Hibernate has already resolved this, and I would rather the guard read it from there than derive it a second time: `sessionFactory.getSessionFactoryOptions().doesConnectionProviderDisableAutoCommit()` is the same boolean `LogicalConnectionManagedImpl` consults before it skips the `setAutoCommit(false)` that this whole change exists to remove, so the guard and the behaviour it guards then cannot disagree. Reading the property map instead also brings in `org.hibernate.internal.util.config.ConfigurationHelper`, and dependabot has bumped `hibernateVersion` five times since May, most recently to 7.4.7 which landed on master today, so `org.hibernate.internal` is a surface I would rather not take on for a boolean that is already on the SPI.

```suggestion
		if (!sessionFactory.getSessionFactoryOptions().doesConnectionProviderDisableAutoCommit()) {
			return;
		}
```

That leaves the `JdbcSettings` and `ConfigurationHelper` imports unused, and spotless:apply drops them for you on the next build, I checked. I ran the guard both ways on your branch afterwards as well: the context comes up normally, and with `hibernate.connection.autocommit=true` in play the `sessionFactory` bean still fails with your message.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
```diff
@@ -247,4 +251,29 @@ private void generateEnversAuditTables(Metadata metadata, ServiceRegistry servic
...
+	        SessionFactoryImplementor sessionFactory) {
+
+		if (!sessionFactory.getSessionFactoryOptions().doesConnectionProviderDisableAutoCommit()) {
+			return;
+		}
+		ConnectionProvider provider = serviceRegistry.getService(ConnectionProvider.class);
+		try {
+			Connection conn = provider.getConnection();
+			try {
+				if (conn.getAutoCommit()) {
```
dkayiwa, line 265: Nothing in the suite notices if this stops firing. I commented out the `assertProviderDisablesAutocommitPremise` call in `integrate` on your branch and reran the class this PR touches: `ConnectionPoolConfigTest` still passes 2/2, and so does everything else, because every context-sensitive test only ever walks the path where the premise holds and the method returns quietly. So the one thing standing between this default and silent loss of transactions can be deleted, inverted or reordered out of `integrate` and CI stays green. Codecov reads the same hole from the other side, 46% patch coverage on this file with five of the thirteen new lines never executed.

`integrate` receives everything the check needs as arguments, so closing it does not need a Spring context. I ran this on your branch: three tests, 1.2s, spotless leaves the formatting alone, and with the call commented out two of the three go red.

```java
package org.openmrs.api.db.hibernate;

import java.sql.Connection;
import java.sql.SQLException;

import org.hibernate.HibernateException;
import org.hibernate.boot.Metadata;
import org.hibernate.boot.registry.StandardServiceRegistry;
import org.hibernate.boot.spi.BootstrapContext;
import org.hibernate.boot.spi.SessionFactoryOptions;
import org.hibernate.engine.jdbc.connections.spi.ConnectionProvider;
import org.hibernate.engine.spi.SessionFactoryImplementor;
import org.junit.jupiter.api.Test;

import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
import static org.junit.jupiter.api.Assertions.assertThrows;
import static org.mockito.ArgumentMatchers.any;
import static org.mockito.Mockito.mock;
import static org.mockito.Mockito.never;
import static org.mockito.Mockito.verify;
import static org.mockito.Mockito.when;

public class HibernateSessionFactoryBeanTest {

	private void integrate(boolean settingOn, boolean poolAutoCommit, ConnectionProvider provider) throws SQLException {
		SessionFactoryOptions options = mock(SessionFactoryOptions.class);
		when(options.doesConnectionProviderDisableAutoCommit()).thenReturn(settingOn);
		SessionFactoryImplementor sessionFactory = mock(SessionFactoryImplementor.class);
		when(sessionFactory.getSessionFactoryOptions()).thenReturn(options);

		Connection connection = mock(Connection.class);
		when(connection.getAutoCommit()).thenReturn(poolAutoCommit);
		when(provider.getConnection()).thenReturn(connection);

		StandardServiceRegistry registry = mock(StandardServiceRegistry.class);
		when(registry.getService(ConnectionProvider.class)).thenReturn(provider);
		BootstrapContext bootstrapContext = mock(BootstrapContext.class);
		when(bootstrapContext.getServiceRegistry()).thenReturn(registry);

		new HibernateSessionFactoryBean().integrate(mock(Metadata.class), bootstrapContext, sessionFactory);
	}

	@Test
	public void shouldRefuseToStartWhenThePoolHandsOutAutoCommitEnabledConnections() throws SQLException {
		ConnectionProvider provider = mock(ConnectionProvider.class);

		assertThrows(HibernateException.class, () -> integrate(true, true, provider));

		verify(provider).closeConnection(any());
	}

	@Test
	public void shouldStartWhenThePoolHandsOutAutoCommitDisabledConnections() throws SQLException {
		ConnectionProvider provider = mock(ConnectionProvider.class);

		assertDoesNotThrow(() -> integrate(true, false, provider));

		verify(provider).closeConnection(any());
	}

	@Test
	public void shouldNotTouchThePoolWhenTheSettingIsOff() throws SQLException {
		ConnectionProvider provider = mock(ConnectionProvider.class);

		assertDoesNotThrow(() -> integrate(false, true, provider));

		verify(provider, never()).getConnection();
	}
}
```

(plus the usual licence header). I would not hold the merge for this, but the guard is the whole safety story for the property and its own failure is the silent kind.

api/src/main/resources/hibernate.default.properties
dkayiwa: Still worth the two comment lines above the property itself. Line 8 is what a deployer copies into their own openmrs-runtime.properties, and there is nothing there to say the value is only safe because hibernate-c3p0 turns auto-commit off at checkout, or that `HibernateSessionFactoryBean` now refuses to start if that stops being true. The `checkoutTimeout` block further down already does exactly that job for its own hidden dependency.

sudhanshu-raj:
Hey @dkayiwa , just a quick reminder on the PR review. Thanks!

api/src/main/java/org/openmrs/api/db/hibernate/HibernateSessionFactoryBean.java
```diff
@@ -247,4 +251,29 @@ private void generateEnversAuditTables(Metadata metadata, ServiceRegistry servic
...
+					throw new HibernateException(
+					        "provider_disables_autocommit=true but the pool returned a connection with autoCommit=true; "
+					                + "configure the pool to disable autocommit at checkout or remove the property");
```
dkayiwa: The two remedies this message offers are not the one the deployer reading it can actually apply. `hibernate.connection.provider_disables_autocommit` ships inside openmrs-api.jar as part of `hibernate.default.properties`, so "remove the property" is not something they can do; what they can do is put `connection.provider_disables_autocommit=false` in their openmrs-runtime.properties. And with the pool we actually ship, "configure the pool to disable autocommit at checkout" has exactly one concrete meaning: `C3P0ConnectionProvider#prepareConnection` forces every checked-out connection to `hibernate.connection.autocommit`, which defaults to false, so the only way to land in this branch is that somebody set that to true. Saying so turns a read through Hibernate's c3p0 provider into a one line fix.

I ran the wording below on your branch with `hibernate.connection.autocommit=true` appended to the defaults, and it comes out at the top of the Spring failure as

`org.hibernate.HibernateException: provider_disables_autocommit=true but the pool returned a connection with autoCommit=true; remove connection.autocommit=true from openmrs-runtime.properties, or set connection.provider_disables_autocommit=false there if this pool really must hand out auto-commit-enabled connections`

`spotless:apply` leaves the wrapping alone. Reword it however you like, it is not something I would hold the merge for.

```suggestion
					throw new HibernateException(
					        "provider_disables_autocommit=true but the pool returned a connection with autoCommit=true; "
					                + "remove connection.autocommit=true from openmrs-runtime.properties, or set "
					                + "connection.provider_disables_autocommit=false there if this pool really must hand out "
					                + "auto-commit-enabled connections");
```

api/src/main/resources/hibernate.default.properties
ibacher: I actually think that the guard and startup assertion are overkill here. The thing we're claiming is that our connection provider (C3P0) sets autocommit to `true`. That's the case. It's been the case with C3P0 for forever. That makes this feel more like something we should check _if we ever switch to Hikari or something_, not _every time the app starts_.

I'm happy for us to have a regression test for this; that's actually a really _good_ idea, but the start-up overhead here is, I think, weird for a situation that is unlikely to happen. I think a note a wiki page that it is not recommended to reconfigured C3P0's autocommit setting is a more justifiable response.

ibacher (commented):
I'm actually happy for this to go in but without the guard. @dkayiwa Should we discuss this on the next platform call?

dkayiwa:
> Should we discuss this on the next platform call?

Makes perfect sense to do that on the next platform call. 👍

api/src/main/resources/hibernate.default.properties
sudhanshu-raj: Removed those, and this is wiki page : https://openmrs.atlassian.net/wiki/spaces/docs/pages/1380810793/Hibernate+connection+provider+auto-commit+behavior

PR #6535, opened 2026-09-02
TRUNK-6682: Complex obs storage abstraction performs redundant exists/metadata calls per read

jwnasambu:

I reduced the number of storage service I/O operations performed when reading a complex obs eliminating redundant existence probes and separate metadata fetches. Reading a complex obs previously cost ~4–5 filesystem operations on local storage or three S3 round-trips of HEAD/GET/HEAD; it now costs a single combined data+metadata read on the happy path.

## Issue I worked on

https://openmrs.atlassian.net/browse/TRUNK-6682

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/api/StorageService.java
+++ b/api/src/main/java/org/openmrs/api/StorageService.java
@@ -15,6 +15,7 @@ import java.io.UncheckedIOException;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
+import org.openmrs.api.storage.DataWithMetadata;
 import org.openmrs.api.storage.ObjectMetadata;
 import org.openmrs.api.stream.StreamDataWriter;
 
@@ -176,4 +177,20 @@ public interface StorageService extends OpenmrsService {
 	 * @throws UncheckedIOException IO error
 	 */
 	boolean exists(String key) throws UncheckedIOException;
+
+	/**
+	 * Returns both data and metadata for the given key in a single operation.
+	 * <p>
+	 * This avoids redundant I/O (e.g. separate existence probes or metadata fetches) when the backend
+	 * can provide both in one call (e.g. S3 GetObject returns response headers with metadata, local
+	 * filesystem can read attributes alongside the file).
+	 * <p>
+	 * The caller must close the returned {@link org.openmrs.api.storage.DataWithMetadata} to release
+	 * the underlying stream.
+	 *
+	 * @param key unique key
+	 * @return data with metadata
+	 * @throws IOException wrong key or IO error
+	 */
+	DataWithMetadata getDataWithMetadata(String key) throws IOException;
 }
--- a/api/src/main/java/org/openmrs/api/storage/DataWithMetadata.java
+++ b/api/src/main/java/org/openmrs/api/storage/DataWithMetadata.java
@@ -1,0 +1,25 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.api.storage;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Holds data and metadata returned together from a single storage read.
+ * <p>
+ * The data stream must be closed by the caller (or via {@link #close()}) to release the underlying
+ * resource.
+ *
+ * @since 2.8.0
+ */
+public record DataWithMetadata(InputStream data,ObjectMetadata metadata)implements AutoCloseable{
+
+@Override public void close()throws IOException{data.close();}}
--- a/api/src/main/java/org/openmrs/api/storage/LocalStorageService.java
+++ b/api/src/main/java/org/openmrs/api/storage/LocalStorageService.java
@@ -59,6 +59,20 @@ public class LocalStorageService extends BaseStorageService implements StorageSe
 	}
 
 	@Override
+	public DataWithMetadata getDataWithMetadata(String key) throws IOException {
+		Path path = getPath(key);
+
+		BasicFileAttributes attributes = Files.readAttributes(path, BasicFileAttributes.class);
+		String filename = decodeKey(path.getFileName().toString());
+		ObjectMetadata metadata = ObjectMetadata.builder().setLength(attributes.size())
+		        .setMimeType(mimetypes.getContentType(filename)).setFilename(filename)
+		        .setCreationTime(attributes.creationTime().toInstant()).build();
+
+		InputStream data = Files.newInputStream(path);
+		return new DataWithMetadata(data, metadata);
+	}
+
+	@Override
 	public InputStream getData(final String key) throws IOException {
 		return Files.newInputStream(getPath(key));
 	}
--- a/api/src/main/java/org/openmrs/api/storage/S3StorageService.java
+++ b/api/src/main/java/org/openmrs/api/storage/S3StorageService.java
@@ -132,6 +132,19 @@ public class S3StorageService extends BaseStorageService implements StorageServi
 		return waitForResponse(object);
 	}
 
+	@Override
+	public DataWithMetadata getDataWithMetadata(String key) throws IOException {
+		CompletableFuture<ResponseInputStream<GetObjectResponse>> future = s3AsyncClient.getObject(
+		    GetObjectRequest.builder().bucket(bucketName).key(encodeKey(key)).build(),
+		    AsyncResponseTransformer.toBlockingInputStream());
+		ResponseInputStream<GetObjectResponse> response = waitForResponse(future);
+		GetObjectResponse getResponse = response.response();
+		ObjectMetadata metadata = new ObjectMetadata();
+		metadata.setMimeType(getResponse.contentType());
+		metadata.setLength(getResponse.contentLength());
+		return new DataWithMetadata(response, metadata);
+	}
+
 	private <T> T waitForResponse(CompletableFuture<T> object) throws IOException {
 		T result;
 		try {
--- a/api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
+++ b/api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
@@ -20,6 +20,7 @@ import org.openmrs.Obs;
 import org.openmrs.api.APIException;
 import org.openmrs.api.AdministrationService;
 import org.openmrs.api.StorageService;
+import org.openmrs.api.storage.DataWithMetadata;
 import org.openmrs.api.storage.ObjectMetadata;
 import org.openmrs.obs.ComplexData;
 import org.openmrs.util.OpenmrsConstants;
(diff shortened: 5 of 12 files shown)


jwnasambu:
@dkayiwa, @ibacher Kindly feel free to review this PR at your convenient time please!

api/src/main/java/org/openmrs/api/storage/DataWithMetadata.java
```diff
@@ -0,0 +1,25 @@
...
+ * Holds data and metadata returned together from a single storage read.
+ * <p>
+ * The data stream must be closed by the caller (or via {@link #close()}) to release the underlying
+ * resource.
+ *
+ * @since 2.8.0
+ */
+public record DataWithMetadata(InputStream data,ObjectMetadata metadata)implements AutoCloseable{
+
+@Override public void close()throws IOException{data.close();}}
```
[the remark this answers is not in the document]
ibacher: @jwnasambu Please address this. `mvn spotless:apply` should reformat things as required.

api/src/main/java/org/openmrs/api/storage/LocalStorageService.java
```diff
@@ -58,6 +58,20 @@ public LocalStorageService(@Value("${storage.local.dir:}") String storageDir,
...
+		Path path = getPath(key);
+
+		BasicFileAttributes attributes = Files.readAttributes(path, BasicFileAttributes.class);
+		String filename = decodeKey(path.getFileName().toString());
+		ObjectMetadata metadata = ObjectMetadata.builder().setLength(attributes.size())
+		        .setMimeType(mimetypes.getContentType(filename)).setFilename(filename)
+		        .setCreationTime(attributes.creationTime().toInstant()).build();
+
+		InputStream data = Files.newInputStream(path);
+		return new DataWithMetadata(data, metadata);
```
ibacher: What would be wrong with? That's backwards compatible.
```java
return new DataWithMetadata(getData(key), getMetadata(key));
```

ibacher, line 72: Rather than re-implementing `getData()` and `getMetadata()` here, it makes more sense to create private functions with the shared parts of those then this function becomes something like:

```java
Path path = getPath(key);
ObjectMetadata metadata = getMetadataInternal(path);
return new DataWithMetadata(getDataInternal(path), metadata);
```

And then move the shared code between this and `getData()` and `getMetadata()` to that internal function so that these stay in sync.

api/src/main/java/org/openmrs/api/storage/S3StorageService.java
```diff
@@ -144,10 +163,28 @@ private <T> T waitForResponse(CompletableFuture<T> object) throws IOException {
 		return result;
 	}
 
+	private IOException translateNotFound(String key, IOException e) {
```
ibacher, line 166: The fact that we are trying to guarantee this means that it must be documented in the StorageService that we throw a `NoSuchFileException` if the file cannot be found.

api/src/main/java/org/openmrs/api/StorageService.java
```diff
@@ -176,4 +177,22 @@ String saveData(StreamDataWriter writer, ObjectMetadata metadata, String moduleI
 	 * @throws UncheckedIOException IO error
 	 */
 	boolean exists(String key) throws UncheckedIOException;
+
+	/**
+	 * Returns both data and metadata for the given key.
+	 * <p>
+	 * The default implementation fetches data and metadata separately, which may be optimized by
+	 * implementations that can provide both in a single operation (e.g. S3 GetObject returns response
+	 * headers with metadata, local filesystem can read attributes alongside the file).
```
ibacher, line 186:
```suggestion
	 * implementations that can provide both in a single operation.
```

api/src/main/java/org/openmrs/api/storage/LocalStorageService.java
ibacher: This function also needs a comment that ensuring metadata loads before the data is semantics that _must_ be maintained. Otherwise, if they got rearranged and the metadata call was after the data call, we could end up with an unclosed `InputStream`, so this is actually an important point.

api/src/main/java/org/openmrs/obs/handler/AbstractHandler.java
```diff
@@ -123,12 +135,7 @@ public boolean purgeComplexData(Obs obs) {
 	public String parseDataKey(Obs obs) {
 		String[] names = obs.getValueComplex().split("\\|");
 		String key = names.length < 2 ? names[0] : names[names.length - 1];
-		key = StringUtils.trim(key);
-		if (!storageService.exists(key)) {
```
ibacher, line 127: I'm not convinced about this replacement. You've replaced a relatively cheap `exists()` check with a thrown exception, which actually seems more heavy-weight?

ibacher: But the real problem is that this is a public API method and you've changed what it returns. We can't backport this implementation

api/src/main/java/org/openmrs/obs/handler/BinaryDataHandler.java
```diff
@@ -59,10 +61,21 @@ public Obs getObs(Obs obs, String view) {
...
+		DataWithMetadata dwm = null;
 
 		// Raw view (i.e. the file as is)
 		if (ComplexObsHandler.RAW_VIEW.equals(view)) {
-			try (InputStream in = storageService.getData(key)) {
+			try {
+				dwm = getDataWithMetadataWithLegacyFallback(key);
+			} catch (NoSuchFileException e) {
+				log.error("Trying to read file: {}", key, e);
+				Assert.notNull(null, "Complex data must not be null");
```
ibacher, line 72: Why in the world does this line exist? Just throw the `IllegalArgumentException` you'll get anyways.

api/src/main/java/org/openmrs/obs/handler/BinaryDataHandler.java
```diff
@@ -59,10 +61,21 @@ public Obs getObs(Obs obs, String view) {
...
+				return null;
+			} catch (IOException e) {
+				log.error("Trying to read file: {}", key, e);
+				throw new UncheckedIOException(e);
```
ibacher, line 76: I don't _love_ that we're throwing `UncheckedIOException` everywhere in this PR. The more consistent thing to do with OMRS is throw an APIException wrapping the IOException.

ibacher (commented):
Thanks @jwnasambu! Note that these are illustrative comments. I did not go through every handler, but they all have similar issues.

jwnasambu:
@claude review

jwnasambu:
@ibacher, @dkayiwa, @wikumChamith  Kindly feel free to review my PR at your convenient time please!

api/src/test/java/org/openmrs/obs/AbstractHandlerTest.java
```diff
@@ -99,4 +201,23 @@ public void saveObs_shouldCorrectlySaveFileWithoutTitle() {
...
+		Obs obs = new Obs();
+		obs.setComplexData(complexData);
+		countingHandler.saveObs(obs);
+
+		Obs fetched = countingHandler.getObs(obs, null);
+
+		assertEquals("test", new String((byte[]) fetched.getComplexData().getData()));
+		assertEquals(1, counting.getDataWithMetadataCalls, "data+metadata must be fetched together");
+		assertEquals(0, counting.getMetadataCalls, "no separate metadata request when data provides it");
+		assertEquals(1, counting.existsCalls, "single existence probe to resolve the key layout");
```
dkayiwa, line 219: At the StorageService level this really is one probe, but `LocalStorageService.getPath()` runs its own `Files.exists(legacyPath)` on every call, and it gets reached three times per read on this path: once inside `exists()`, once more when `exists()` stats the resolved path, and once inside `getDataWithMetadata()`.

I instrumented `LocalStorageService` and ran `AbstractHandler.getObs`, `TextHandler.getObs(TEXT_VIEW)` and `BinaryDataHandler.getObs(RAW_VIEW)` on this branch and on master. All three give the same counts:

```
master:      4 x Files.exists + 1 x newInputStream + 1 x readAttributes = 6
this branch: 3 x Files.exists + 1 x readAttributes + 1 x newInputStream = 5
```

The duplicate metadata fetch is genuinely gone, which is the part this set out to do. But TRUNK-6682 asks for "at most one existence-related check", and it counts `getPath()`'s probe explicitly when it tallies the per-read cost, so on local storage a read lands at five filesystem calls rather than at the "single combined data+metadata read on the happy path" the PR description claims.

The probes that are left are the ones ibacher's feedback deliberately kept, so I am not asking you to revisit that. What I would like settled before this merges is the ticket: does TRUNK-6682 stay open for the layout-resolution half, with this PR closing only the duplicate-metadata part?

api/src/main/java/org/openmrs/obs/handler/ImageHandler.java
```diff
@@ -77,7 +78,18 @@ public Obs getObs(Obs obs, String view) {
...
+
+			DataWithMetadata dwm;
+			try {
+				dwm = storageService.getDataWithMetadata(key);
+			} catch (IOException e) {
+				log.error("Trying to read file: {}", key, e);
+				ComplexData complexData = new ComplexData(parseFilename(obs, "image"), null);
+				obs.setComplexData(complexData);
+				return obs;
+			}
```
dkayiwa, line 90: This early return loses the legacy filename fallback a few lines further down, the `extensions.contains(filename)` block carrying the TRUNK-6472 comment. Master reached it on this path because it logged the failure and carried on into the shared tail.

With `valueComplex = "jpg|2026/01-01/2026-01-01-00-00-00-000-abcdefgh.jpg"` and the file absent, master hands back a `ComplexData` titled `/2026/01-01/2026-01-01-00-00-00-000-abcdefgh.jpg`, and this branch hands back one titled `jpg`. The data is null either way, so nothing throws and no test notices; the obs just loses the only name identifying it.

Keeping master's single exit fixes it: initialise `dwm` to null, let the catch log without returning, and skip the read block when `dwm` is null. The `dwm.metadata()` call further down is already behind `img != null`, which cannot be reached when `dwm` is null.

api/src/main/java/org/openmrs/api/storage/DataWithMetadata.java
```diff
@@ -0,0 +1,25 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Holds data and metadata returned together from a single storage read.
+ * <p>
+ * The data stream must be closed by the caller (or via {@link #close()}) to release the underlying
+ * resource.
+ *
+ * @since 2.8.0
```
dkayiwa, line 21: 2.8.0 was tagged a long time ago (2.8.9 is out) and master is 3.0.0-SNAPSHOT, so a module author who trusts this tag and sets `require_version 2.8.0` gets a `NoClassDefFoundError` the first time a handler hands them one of these. It wants the version this actually lands in, plus whatever maintenance lines it gets backported to. `StorageService.getDataWithMetadata` and the new `injectMissingMetadata` overload have no `@since` at all.

api/src/main/java/org/openmrs/api/storage/DataWithMetadata.java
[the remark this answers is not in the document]
dkayiwa: `mvn spotless:apply` will not fix this one. Running it on the api module at this head leaves the file byte identical (same md5), and `CI=true mvn spotless:check -Dspotless.check.skip=false` passes, reporting all 1270 files clean. As a control I compressed a method in `ObjectMetadata.java` the same way and spotless both flagged and fixed that, so the plugin is running and the file is in scope; the Eclipse formatter config just leaves record declarations untouched. That is also why the build here is green, so CI is not going to catch it either.

It needs doing by hand: the record header and `close()` each on their own properly indented lines.

PR #6539, opened 2026-09-04
TRUNK-6725: Resolve coded interpretations using rules

prakhar0608shrma:

- Added `ConceptInterpretationRule` to define criteria, interpretation, priority, and concept.
- Added persistence and Hibernate mappings for interpretation rules.
- Added `ConceptService.getConceptInterpretation(ConceptReferenceRangeContext)`.
- Evaluates rule criteria using `ConceptReferenceRangeUtility.evaluateCriteria`.
- Returns the interpretation from the highest-priority matching rule.
- Added Liquibase changesets and indexes for the interpretation-rule table.
- Added tests for persistence, criteria matching, priority resolution, and no matching rule.

## Issue I worked on

https://issues.openmrs.org/browse/TRUNK-6725

### DIFF
--- a/api/src/main/java/org/openmrs/api/ConceptService.java
+++ b/api/src/main/java/org/openmrs/api/ConceptService.java
@@ -39,6 +39,7 @@ import org.openmrs.ConceptSource;
 import org.openmrs.ConceptStopWord;
 import org.openmrs.Drug;
 import org.openmrs.DrugIngredient;
+import org.openmrs.Obs;
 import org.openmrs.Person;
 import org.openmrs.annotation.Authorized;
 import org.openmrs.api.db.ConceptDAO;
@@ -2204,6 +2205,16 @@ public interface ConceptService extends OpenmrsService {
 	 */
 	@Authorized(PrivilegeConstants.GET_CONCEPTS)
 	ConceptReferenceRange getConceptReferenceRange(ConceptReferenceRangeContext context);
+
+	/**
+	 * Gets the interpretation for a concept by evaluating its interpretation rules against the given
+	 * context. Higher priority values take precedence when multiple rules match.
+	 *
+	 * @param context the context containing the person, concept, and optional date
+	 * @return the interpretation from the highest-priority matching rule, or null if none matches
+	 */
+	@Authorized(PrivilegeConstants.GET_CONCEPTS)
+	Obs.Interpretation getConceptInterpretation(ConceptReferenceRangeContext context);
 
 	/**
 	 * Completely purge a <code>ConceptReferenceRange</code> from the database.
--- a/api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.9.x.xml
+++ b/api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.9.x.xml
@@ -97,4 +97,49 @@
 		</createIndex>
 	</changeSet>
 	
+	<changeSet id="TRUNK-6723-2026-08-26" author="Rocker810">
+		<preConditions onError="HALT" onFail="MARK_RAN">
+			<not>
+				<tableExists tableName="concept_interpretation_rule"/>
+			</not>
+		</preConditions>
+		<comment>Creating 'concept_interpretation_rule' table</comment>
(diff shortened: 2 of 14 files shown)


api/src/test/java/org/openmrs/util/DatabaseUpdaterDatabaseIT.java
```diff
@@ -32,7 +32,7 @@ public class DatabaseUpdaterDatabaseIT extends DatabaseIT {
 	 * This constant needs to be updated when adding new Liquibase update files to openmrs-core.
 	 */
 
-	private static final int CHANGE_SET_COUNT_FOR_GREATER_THAN_2_1_X = 916;
+	private static final int CHANGE_SET_COUNT_FOR_GREATER_THAN_2_1_X = 918;
```
jwnasambu, line 35:
```suggestion
	private static final int CHANGE_SET_COUNT_FOR_GREATER_THAN_2_1_X = 920;
```
Kindly the IT counts every change set in the 2.1.x schema/core-data snapshot plus all update files newer than 2.1.x, which includes both liquibase-update-to-latest-2.9.x.xml and liquibase-update-to-latest-3.0.x.xml. This PR adds 2 change sets to each of those files (lines 100–175 in each), i.e. +4 change sets, so the expected total is 916 + 4 = 920. As written, should() will fail the next time the integration suite runs, asserting 918 when Liquibase reports 920.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.9.x.xml
```diff
@@ -97,4 +97,49 @@
 		</createIndex>
 	</changeSet>
 	
+	<changeSet id="TRUNK-6723-2026-08-26" author="Rocker810">
```
jwnasambu, line 100: Kindly rename the change set id TRUNK-6723-2026-08-26 to reference this ticket, e.g. TRUNK-6725-2026-08-26, in both copies. The id currently points at the reference-range ticket (TRUNK-6723) rather than the work this PR resolves. Change set ids are permanent once a release ships, so keeping the wrong ticket reference would mislead anyone later reading migration history. Renaming now, before anything has shipped with the id, is free.

api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,50 @@
 		</addColumn>
 	</changeSet>
 	
+	<changeSet id="TRUNK-6723-2026-08-26" author="Rocker810">
```
jwnasambu, line 145: Changes proposed  on api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-2.9.x.xml:100 should be duplicate at liquibase-update-to-latest-3.0.x.xml:145

api/src/main/java/org/openmrs/api/ConceptService.java
```diff
@@ -2205,6 +2206,16 @@ public List<ConceptSearchResult> getOrderableConcepts(String phrase, List<Locale
...
 
+	/**
+	 * Gets the interpretation for a concept by evaluating its interpretation rules against the given
+	 * context. Higher priority values take precedence when multiple rules match.
+	 *
+	 * @param context the context containing the person, concept, and optional date
+	 * @return the interpretation from the highest-priority matching rule, or null if none matches
+	 */
+	@Authorized(PrivilegeConstants.GET_CONCEPTS)
+	Obs.Interpretation getConceptInterpretation(ConceptReferenceRangeContext context);
```
jwnasambu, line 2217: Kindly how are rules meant to be created and managed? Right now the only path is concept.addInterpretationRule(...) plus saveConcept(...), because getConceptInterpretation here is the only new service entry point. The reference-range counterpart exposes saveConceptReferenceRange, getConceptReferenceRangesByConceptId, and getConceptReferenceRangeByUuid for the same kind of API surface, so a REST resource or the form engine will need comparable entry points to read or update a single rule which don't exist on this new rule type yet. If that CRUD API is a deliberate follow-up, a note on TRUNK-6725 would stop it from being lost.

PR #6544, opened 2026-09-07
TRUNK-6799 Introduce Spring Security with support for legacy authenti…

rkorytkowski:
…cation context

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6799

  No? -> execute above command

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.idea
+.git
+**/target
+**/.DS_Store
+openmrs-merged.properties
+docker-image.txt
--- a/api/src/main/java/org/openmrs/annotation/Authorized.java
+++ b/api/src/main/java/org/openmrs/annotation/Authorized.java
@@ -29,6 +29,14 @@ import java.lang.annotation.Target;
  *     &#64;Authorized ()
  *     public void getUsersByName(String name);
  * </pre>
+ * <p>
+ * As of 3.0.0, Spring Security's {@code @PreAuthorize} is also fully supported (see
+ * {@link org.openmrs.security.OpenmrsSecurityConfig}), evaluated by a separate, independent advisor
+ * alongside {@link org.openmrs.aop.AuthorizationAdvice} - a method carries one annotation or the
+ * other, never both. {@code @Authorized} remains fully supported and is not going away; there is no
+ * need to convert existing usages. New code, and contributors who prefer Spring Security's standard
+ * mechanism, may use {@code @PreAuthorize("hasPermission(null, '&lt;privilege&gt;')")} instead -
+ * see {@link org.openmrs.api.ProviderService#getProvider(Integer)} for a reference example.
  */
 @Target({ ElementType.METHOD, ElementType.TYPE })
 @Retention(RetentionPolicy.RUNTIME)
--- a/api/src/main/java/org/openmrs/api/ProviderService.java
+++ b/api/src/main/java/org/openmrs/api/ProviderService.java
@@ -21,6 +21,7 @@ import org.openmrs.ProviderRole;
 import org.openmrs.annotation.Authorized;
 import org.openmrs.annotation.Handler;
 import org.openmrs.util.PrivilegeConstants;
+import org.springframework.security.access.prepost.PreAuthorize;
 
 /**
  * This service contains methods relating to providers.
@@ -38,7 +39,7 @@ public interface ProviderService extends OpenmrsService {
 	 *
 	 * @return a list of provider objects.
 	 */
-	@Authorized({ PrivilegeConstants.GET_PROVIDERS })
+	@PreAuthorize("hasPermission(null, '" + PrivilegeConstants.GET_PROVIDERS + "')")
 	public List<Provider> getAllProviders();
 
 	/**
@@ -59,7 +60,7 @@ public interface ProviderService extends OpenmrsService {
 	 * @param provider provider to retire
 	 * @param reason reason why the provider is retired
 	 */
-	@Authorized({ PrivilegeConstants.MANAGE_PROVIDERS })
+	@PreAuthorize("hasPermission(null, '" + PrivilegeConstants.MANAGE_PROVIDERS + "')")
 	public void retireProvider(Provider provider, String reason);
 
 	/**
@@ -90,7 +91,7 @@ public interface ProviderService extends OpenmrsService {
 	 * @param providerId the provider id
 	 * @return the provider by it's id
 	 */
-	@Authorized({ PrivilegeConstants.GET_PROVIDERS })
+	@PreAuthorize("hasPermission(null, '" + PrivilegeConstants.GET_PROVIDERS + "')")
 	public Provider getProvider(Integer providerId);
 
 	/**
--- a/api/src/main/java/org/openmrs/api/cache/RolePrivilegeCache.java
+++ b/api/src/main/java/org/openmrs/api/cache/RolePrivilegeCache.java
@@ -9,7 +9,9 @@
  */
 package org.openmrs.api.cache;
 
+import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
@@ -62,6 +64,22 @@ import org.springframework.stereotype.Component;
  * That template carries a {@code lifespan} TTL as a safety net against role changes made outside
  * the API (eviction only fires on API mutations); it is configured declaratively in {@code
  * infinispan-api.xml}/{@code infinispan-api-local.xml} rather than per entry here.
+ * <p>
+ * Also caches the flattened set of every currently registered {@link Privilege} name (see
+ * {@link #getAllPrivilegeNames()}) and every registered {@link org.openmrs.Role} name (see
+ * {@link #getAllRoleNames()}), each under its own reserved key in the same cache region so both are
+ * invalidated by the same {@code @CacheEvict} on {@code savePrivilege}/{@code purgePrivilege}/
+ * {@code saveRole}/{@code purgeRole} that already clears per-role entries, without a dedicated
+ * cache region. This backs {@code OpenmrsAuthenticationToken} giving a superuser or {@code Daemon}
+ * thread every registered privilege and role as {@code GrantedAuthority}s (both satisfy any
+ * privilege or role name, which cannot otherwise be represented as a static, enumerable authority
+ * set - see {@link #isRegisteredPrivilege(String)}), and {@link #warnIfUnregistered(String)}, which
+ * flags a privilege check naming a string with no corresponding row - almost always a forgotten
+ * registration or a typo, since {@code Context.hasPrivilege(String)} enforces no such requirement
+ * itself. There is no role equivalent of {@code warnIfUnregistered}: unlike a privilege name, a
+ * role name checked via {@code hasRole(...)} is always compared against actual
+ * {@link org.openmrs.Role} membership, so there is no "checked but never registered" case to
+ * diagnose.
  *
  * @since 3.0.0, 2.9.0, 2.8.9
  */
@@ -73,6 +91,21 @@ public class RolePrivilegeCache implements ApplicationListener<ContextRefreshedE
 	public static final String CACHE_NAME = "rolePrivileges";
 
 	/**
+	 * Cache/in-flight key for the flattened set of every registered privilege name. Chosen to be
+	 * exceedingly unlikely to collide with an admin-supplied role name, which is the only other kind of
+	 * key stored in this cache region.
+	 */
+	private static final String ALL_PRIVILEGES_KEY = "\u0000ALL_PRIVILEGES\u0000";
+
+	/**
+	 * Cache/in-flight key for the flattened set of every registered role name, alongside
+	 * {@link #ALL_PRIVILEGES_KEY} in the same reserved-key scheme.
+	 */
+	private static final String ALL_ROLES_KEY = "\u0000ALL_ROLES\u0000";
+
+	private static final RolePrivileges EMPTY_NAMES = new RolePrivileges(Collections.emptySet(), false);
+
+	/**
 	 * Capability token issued by {@link Daemon}, letting this component launch a daemon thread to load
 	 * roles with full trust.
 	 */
@@ -81,10 +114,17 @@ public class RolePrivilegeCache implements ApplicationListener<ContextRefreshedE
 	private final CacheManager cacheManager;
 
 	/**
-	 * In-flight refreshes keyed by normalized role name, so concurrent misses for the same role share a
-	 * single daemon computation instead of each launching their own.
+	 * In-flight refreshes keyed by normalized role name (or {@link #ALL_PRIVILEGES_KEY}/
+	 * {@link #ALL_ROLES_KEY}), so concurrent misses for the same key share a single daemon computation
+	 * instead of each launching their own.
 	 */
 	private final ConcurrentMap<String, Future<RolePrivileges>> inFlight = new ConcurrentHashMap<>();
+
+	/**
+	 * Privilege names (case-normalized) that {@link #warnIfUnregistered(String)} has already logged
+	 * about, so a repeatedly-checked unregistered privilege logs once rather than on every check.
+	 */
+	private final Set<String> loggedUnregisteredPrivileges = ConcurrentHashMap.newKeySet();
 
 	@Autowired
 	public RolePrivilegeCache(@Qualifier("apiCacheManager") CacheManager cacheManager) {
@@ -133,6 +173,87 @@ public class RolePrivilegeCache implements ApplicationListener<ContextRefreshedE
 		}
 
 		return refresh(key, role, cache);
+	}
+
+	/**
+	 * Returns the names of every currently registered {@link Privilege}, in their original casing (see
+	 * {@link RolePrivileges} for why), computing and caching the result on a miss. Never returns
+	 * {@code null}.
+	 *
+	 * @return the registered privilege names
+	 * @since 3.0.0
+	 */
+	public Set<String> getAllPrivilegeNames() {
+		return getAllPrivilegesClosure().getPrivilegeNames();
+	}
+
+	/**
+	 * @param privilege the privilege name to test (compared case-insensitively)
+	 * @return true if a {@link Privilege} with this name is registered in the system
+	 * @since 3.0.0
+	 */
+	public boolean isRegisteredPrivilege(String privilege) {
+		return privilege != null && getAllPrivilegesClosure().containsPrivilege(privilege);
+	}
+
+	/**
+	 * Returns the names of every currently registered {@link org.openmrs.Role}, in their original
+	 * casing (see {@link RolePrivileges} for why), computing and caching the result on a miss. Never
+	 * returns {@code null}.
+	 *
+	 * @return the registered role names
+	 * @since 3.0.0
+	 */
+	public Set<String> getAllRoleNames() {
+		return getAllRolesClosure().getPrivilegeNames();
+	}
+
+	private RolePrivileges getAllPrivilegesClosure() {
+		Cache cache = getCache();
+		if (cache != null) {
+			RolePrivileges cached = cache.get(ALL_PRIVILEGES_KEY, RolePrivileges.class);
+			if (cached != null) {
+				return cached;
+			}
+		}
+
+		return refreshAllPrivileges(cache);
+	}
+
+	private RolePrivileges getAllRolesClosure() {
+		Cache cache = getCache();
+		if (cache != null) {
+			RolePrivileges cached = cache.get(ALL_ROLES_KEY, RolePrivileges.class);
+			if (cached != null) {
+				return cached;
+			}
+		}
+
+		return refreshAllRoles(cache);
+	}
+
+	/**
+	 * Logs, at error level and once per distinct privilege name for the lifetime of this component,
+	 * that a privilege check named a string with no corresponding {@link Privilege} row. Purely
+	 * diagnostic: it never affects the outcome of the check that triggered it. The empty string is
+	 * exempt, since it is the implicit "authenticated" privilege every logged-in user holds (see
+	 * {@code UserContext#resolvePrivilege(String)}) and is never itself registered as a
+	 * {@link Privilege}.
+	 *
+	 * @param privilege the privilege name that was checked
+	 * @since 3.0.0
+	 */
+	public void warnIfUnregistered(String privilege) {
+		if (privilege == null || privilege.isEmpty() || isRegisteredPrivilege(privilege)) {
+			return;
+		}
+
+		if (loggedUnregisteredPrivileges.add(RolePrivileges.normalize(privilege))) {
+			log.error(
+			    "Checked privilege '{}' is not registered as a Privilege in the system - "
+			            + "register it (Manage Privileges, or a startup registration) so it can be assigned to a role",
+			    privilege);
+		}
 	}
 
 	/**
@@ -202,6 +323,142 @@ public class RolePrivilegeCache implements ApplicationListener<ContextRefreshedE
 			// The lifespan TTL comes from the cache template (see the class-level note), so a plain put
 			// carries it; no per-entry expiry handling is needed here.
 			cache.put(key, computed);
+		}
+		return computed;
+	}
+
+	/**
+	 * Loads and caches the flattened set of every registered {@link Privilege} name in a daemon thread,
+	 * mirroring {@link #refresh(String, Role, Cache)} for the per-role case: loading them through the
+	 * secured {@code UserService} itself requires {@code Manage Privileges}, so a daemon thread is
+	 * needed to break that cycle for a caller who does not (yet) hold it.
+	 * <p>
+	 * Failure of any kind - scheduling, interruption, or the daemon task itself - fails open to an
+	 * empty, uncached result rather than propagating: this data only ever adds convenience
+	 * ({@code getAuthorities()}'s superuser authorities, or the diagnostic in
+	 * {@link #warnIfUnregistered(String)}), never a security decision, so degrading it during a
+	 * transient failure is preferable to letting it break the privilege check that triggered it.
+	 *
+	 * @param cache the target cache, or <code>null</code> if unavailable
+	 * @return the flattened set of registered privilege names, or an empty one on failure
+	 */
+	private RolePrivileges refreshAllPrivileges(Cache cache) {
+		Future<RolePrivileges> future;
+		try {
+			future = inFlight.computeIfAbsent(ALL_PRIVILEGES_KEY,
+			    k -> Daemon.runNewDaemonTask((Callable<RolePrivileges>) () -> {
+				    try {
+					    return loadAndCacheAllPrivileges(cache);
+				    } finally {
+					    inFlight.remove(k);
+				    }
+			    }, daemonCallerKey()));
+		} catch (RuntimeException e) {
+			log.warn("Could not schedule a daemon refresh of the registered privilege names; treating the set as "
+			        + "empty for this call",
+			    e);
+			return EMPTY_NAMES;
+		}
+
+		try {
+			return future.get();
+		} catch (InterruptedException e) {
+			Thread.currentThread().interrupt();
+			log.debug(
+			    "Interrupted while refreshing the registered privilege names; treating the set as empty for " + "this call",
+			    e);
+			return EMPTY_NAMES;
+		} catch (ExecutionException e) {
+			log.warn("Refreshing the registered privilege names failed; treating the set as empty for this call",
+			    e.getCause());
+			return EMPTY_NAMES;
+		}
+	}
+
+	/**
+	 * Runs inside a daemon thread: loads every {@link Privilege} through the {@code UserService}
+	 * (authorization skipped for daemon threads), collects their names, and caches the result.
+	 *
+	 * @param cache the target cache, or <code>null</code> if unavailable
+	 * @return the flattened set of registered privilege names
+	 */
+	private RolePrivileges loadAndCacheAllPrivileges(Cache cache) {
+		List<Privilege> privileges = Context.getUserService().getAllPrivileges();
+		Set<String> names = new HashSet<>();
+		for (Privilege privilege : privileges) {
+			if (privilege != null && privilege.getPrivilege() != null) {
+				names.add(privilege.getPrivilege());
+			}
+		}
+
+		RolePrivileges computed = new RolePrivileges(names, false);
+		if (cache != null) {
+			cache.put(ALL_PRIVILEGES_KEY, computed);
+		}
+		return computed;
+	}
+
+	/**
+	 * Loads and caches the flattened set of every registered {@link org.openmrs.Role} name in a daemon
+	 * thread, mirroring {@link #refreshAllPrivileges(Cache)} for the same reason: loading them through
+	 * the secured {@code UserService} itself requires {@code Manage Roles}, so a daemon thread is
+	 * needed to break that cycle for a caller who does not (yet) hold it.
+	 * <p>
+	 * Failure of any kind fails open to an empty, uncached result, same as
+	 * {@link #refreshAllPrivileges(Cache)} and for the same reason: this data only ever adds
+	 * convenience ({@code getAuthorities()}'s superuser/Daemon authorities), never a security decision.
+	 *
+	 * @param cache the target cache, or <code>null</code> if unavailable
+	 * @return the flattened set of registered role names, or an empty one on failure
+	 */
+	private RolePrivileges refreshAllRoles(Cache cache) {
+		Future<RolePrivileges> future;
+		try {
+			future = inFlight.computeIfAbsent(ALL_ROLES_KEY, k -> Daemon.runNewDaemonTask((Callable<RolePrivileges>) () -> {
+				try {
+					return loadAndCacheAllRoles(cache);
+				} finally {
+					inFlight.remove(k);
+				}
+			}, daemonCallerKey()));
+		} catch (RuntimeException e) {
+			log.warn("Could not schedule a daemon refresh of the registered role names; treating the set as empty "
+			        + "for this call",
+			    e);
+			return EMPTY_NAMES;
+		}
+
+		try {
+			return future.get();
+		} catch (InterruptedException e) {
+			Thread.currentThread().interrupt();
+			log.debug("Interrupted while refreshing the registered role names; treating the set as empty for this call", e);
+			return EMPTY_NAMES;
+		} catch (ExecutionException e) {
+			log.warn("Refreshing the registered role names failed; treating the set as empty for this call", e.getCause());
+			return EMPTY_NAMES;
+		}
+	}
+
+	/**
+	 * Runs inside a daemon thread: loads every {@link org.openmrs.Role} through the {@code UserService}
+	 * (authorization skipped for daemon threads), collects their names, and caches the result.
+	 *
+	 * @param cache the target cache, or <code>null</code> if unavailable
+	 * @return the flattened set of registered role names
+	 */
+	private RolePrivileges loadAndCacheAllRoles(Cache cache) {
+		List<Role> roles = Context.getUserService().getAllRoles();
+		Set<String> names = new HashSet<>();
+		for (Role role : roles) {
+			if (role != null && role.getRole() != null) {
+				names.add(role.getRole());
+			}
+		}
+
+		RolePrivileges computed = new RolePrivileges(names, false);
+		if (cache != null) {
+			cache.put(ALL_ROLES_KEY, computed);
 		}
 		return computed;
 	}
--- a/api/src/main/java/org/openmrs/api/cache/RolePrivileges.java
+++ b/api/src/main/java/org/openmrs/api/cache/RolePrivileges.java
@@ -22,9 +22,18 @@ import java.util.Set;
  * superuser status. It is the cached value used by {@link RolePrivilegeCache} to answer privilege
  * checks without re-expanding the role graph on every call.
  * <p>
- * Privilege names are stored case-normalized because privilege comparison in OpenMRS is
- * case-insensitive (see {@link org.openmrs.Role#hasPrivilege(String)}). Lookups normalize the same
- * way, so {@link #containsPrivilege(String)} preserves that behavior.
+ * Privilege comparison in OpenMRS is case-insensitive (see
+ * {@link org.openmrs.Role#hasPrivilege(String)}), so {@link #containsPrivilege(String)} and
+ * {@link #equals(Object)}/{@link #hashCode()} match and compare case-insensitively regardless of
+ * how a privilege name was cased when this was built. {@link #getPrivilegeNames()} nonetheless
+ * returns the names as originally supplied, not lower-cased: {@code OpenmrsAuthenticationToken}
+ * turns them directly into
+ * {@link org.springframework.security.core.authority.SimpleGrantedAuthority} names, and Spring
+ * Security's built-in {@code hasAuthority(...)} SpEL expression matches
+ * {@code GrantedAuthority#getAuthority()} by exact, case-sensitive string equality - a lower-cased
+ * authority would silently never match the privilege's real-world spelling (e.g.
+ * {@code hasAuthority('Manage Privileges')} would need to be spelled
+ * {@code hasAuthority('manage privileges')} instead).
  *
  * @since 3.0.0, 2.9.0, 2.8.9
  */
@@ -34,22 +43,27 @@ public final class RolePrivileges implements Serializable {
 
 	private final Set<String> privilegeNames;
 
+	private final Set<String> normalizedPrivilegeNames;
+
 	private final boolean grantsSuperuser;
 
 	/**
-	 * @param privilegeNames privilege names granted by the role and its inherited closure;
-	 *            case-normalized as copied, null elements ignored
+	 * @param privilegeNames privilege names granted by the role and its inherited closure; stored as
+	 *            supplied (see the class javadoc for why casing is preserved), null elements ignored
 	 * @param grantsSuperuser whether the role or any inherited role confers superuser status
 	 */
 	public RolePrivileges(Set<String> privilegeNames, boolean grantsSuperuser) {
 		Objects.requireNonNull(privilegeNames, "privilegeNames must not be null");
+		Set<String> original = new HashSet<>();
 		Set<String> normalized = new HashSet<>();
 		for (String privilegeName : privilegeNames) {
 			if (privilegeName != null) {
+				original.add(privilegeName);
 				normalized.add(normalize(privilegeName));
 			}
 		}
-		this.privilegeNames = Collections.unmodifiableSet(normalized);
+		this.privilegeNames = Collections.unmodifiableSet(original);
+		this.normalizedPrivilegeNames = Collections.unmodifiableSet(normalized);
 		this.grantsSuperuser = grantsSuperuser;
 	}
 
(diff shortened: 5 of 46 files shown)


rkorytkowski:
I'm still testing this. Just wanted to get review process started. I made some decisions up front, which we should discuss further. I'll try to highlight them in the review comments.

rkorytkowski:
@claude review

web/src/main/java/org/openmrs/web/security/WebSecurityConfig.java
```diff
@@ -0,0 +1,85 @@
...
+	@Bean
+	public OpenmrsSecurityContextFilter openmrsSecurityContextFilter() {
+		return new OpenmrsSecurityContextFilter();
+	}
+
+	@Bean
+	public SecurityFilterChain openmrsSecurityFilterChain(HttpSecurity http,
+	        OpenmrsSecurityContextFilter openmrsSecurityContextFilter, List<AuthorizedUrlMatchers> authorizedUrlMatchers)
+	        throws Exception {
+		http.securityContext(AbstractHttpConfigurer::disable).csrf(AbstractHttpConfigurer::disable)
```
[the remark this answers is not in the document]
rkorytkowski: Yes, we have a custom CSRF protection, which is more robust than the one in Spring Security. We may try enabling this one, but they may collide...

rkorytkowski:
There's an outstanding issue with tests failing on the master branch on GHA. Tests also fail for me locally with the Swing authentication dialog appearing during test execution and tests hanging after rebasing to the latest changes on master... I'm still investigating.

api/src/main/java/org/openmrs/security/OpenmrsPasswordEncoderConfig.java
```diff
@@ -0,0 +1,36 @@
...
+ * {@link PasswordEncoderFactories#createDelegatingPasswordEncoder()} defaults new encodings to
+ * BCrypt while remaining able to verify several other prefixed formats, matching Spring Security's
+ * own recommended default.
+ *
+ * @since 3.0.0
+ */
+@Configuration
+public class OpenmrsPasswordEncoderConfig {
+
+	@Bean(name = "openmrsPasswordEncoder")
```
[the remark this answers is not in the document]
rkorytkowski: It came up after rebasing to the latest master which has some work merged in around password encoder already. I need to re-work that part.

api/src/main/java/org/openmrs/annotation/Authorized.java
```diff
@@ -29,6 +29,15 @@
...
  *     public void getUsersByName(String name);
  * </pre>
+ * <p>
+ * As of 3.0.0, Spring Security's {@code @PreAuthorize} is also fully supported (see
+ * {@link org.openmrs.security.OpenmrsMethodSecurityConfig}), evaluated by a separate, independent
+ * advisor alongside {@link org.openmrs.aop.AuthorizationAdvice} - a method carries one annotation
+ * or the other, never both. {@code @Authorized} remains fully supported and is not going away;
+ * there is no need to convert existing usages. New code, and contributors who prefer Spring
+ * Security's standard mechanism, may use
+ * {@code @PreAuthorize("hasPermission(null, '&lt;privilege&gt;')")} instead - see
```
rkorytkowski: I haven't deprecated `@Authorized...` I'm leaning towards not deprecating it since It's more convenient than writing Spring's `@PreAuthorize("hasAuthority('Manage Concepts')")`. Let me know what you think.

api/src/main/java/org/openmrs/api/cache/RolePrivilegeCache.java
```diff
@@ -62,6 +64,22 @@
...
+ * {@link #getAllRoleNames()}), each under its own reserved key in the same cache region so both are
+ * invalidated by the same {@code @CacheEvict} on {@code savePrivilege}/{@code purgePrivilege}/
+ * {@code saveRole}/{@code purgeRole} that already clears per-role entries, without a dedicated
+ * cache region. This backs {@code OpenmrsAuthenticationToken} giving a superuser or {@code Daemon}
+ * thread every registered privilege and role as {@code GrantedAuthority}s (both satisfy any
+ * privilege or role name, which cannot otherwise be represented as a static, enumerable authority
+ * set - see {@link #isRegisteredPrivilege(String)}), and {@link #warnIfUnregistered(String)}, which
+ * flags a privilege check naming a string with no corresponding row - almost always a forgotten
+ * registration or a typo, since {@code Context.hasPrivilege(String)} enforces no such requirement
+ * itself. There is no role equivalent of {@code warnIfUnregistered}: unlike a privilege name, a
```
rkorytkowski, line 79: Spring's Security hasAuthority and hasRole need a pre-defined list of privileges and roles. We do have such a list in DB, but we never made it an issue for code to check for privileges that are not defined in DB. Especially for admins, we just silently ignored such privileges assuming admins have them. Basically`@PreAuthorize("hasAuthority('Non existing privilege')")` will fail, whereas `@Authorized("Non existing privilege")` will pass for Daemon thread or superuser. I feel that we should also fail `@Authorized` saying the privilege is not defined in the system to make it concise, but for now I just log it as an error. Thoguhts?

api/src/main/java/org/openmrs/api/cache/RolePrivilegeCache.java
```diff
@@ -206,6 +327,142 @@ private RolePrivileges loadAndCache(String key, Role role, Cache cache) {
...
+		Future<RolePrivileges> future;
+		try {
+			future = inFlight.computeIfAbsent(ALL_PRIVILEGES_KEY,
+			    k -> Daemon.runNewDaemonTask((Callable<RolePrivileges>) () -> {
+				    try {
+					    return loadAndCacheAllPrivileges(cache);
+				    } finally {
+					    inFlight.remove(k);
+				    }
+			    }, daemonCallerKey()));
```
rkorytkowski, line 355: I'll switch this to use proxy privilege instead.

api/src/main/java/org/openmrs/api/cache/RolePrivileges.java
```diff
@@ -22,9 +22,18 @@
...
+ * {@link #equals(Object)}/{@link #hashCode()} match and compare case-insensitively regardless of
+ * how a privilege name was cased when this was built. {@link #getPrivilegeNames()} nonetheless
+ * returns the names as originally supplied, not lower-cased: {@code OpenmrsAuthenticationToken}
+ * turns them directly into
+ * {@link org.springframework.security.core.authority.SimpleGrantedAuthority} names, and Spring
+ * Security's built-in {@code hasAuthority(...)} SpEL expression matches
+ * {@code GrantedAuthority#getAuthority()} by exact, case-sensitive string equality - a lower-cased
+ * authority would silently never match the privilege's real-world spelling (e.g.
+ * {@code hasAuthority('Manage Privileges')} would need to be spelled
+ * {@code hasAuthority('manage privileges')} instead).
```
rkorytkowski, line 36: This is a limitation that I don't see an easy way around. Spring Security is case sensitive for roles and privileges, whereas our checks are not. In all cases I've ever seen we just use the exact casing for `@Authorized` and I wasn't even aware of the fact that I could use any casing.

api/src/main/java/org/openmrs/api/ProviderService.java
```diff
@@ -38,7 +39,7 @@ public interface ProviderService extends OpenmrsService {
 	 *
 	 * @return a list of provider objects.
 	 */
-	@Authorized({ PrivilegeConstants.GET_PROVIDERS })
+	@PreAuthorize("hasPermission(null, '" + PrivilegeConstants.GET_PROVIDERS + "')")
```
rkorytkowski, line 42: or simply `@PreAuthorize("hasAuthority('" + PrivilegeConstants.GET_PROVIDERS + "')")`

rkorytkowski:
One more important decision to be made is the switch from APIAuthenticationException to Spring Security AccessDeniedException. Both are runtime exceptions so they don't need to be declared and handled. Any failed authentication check in Spring Security throws AccessDeniedException, whereas our legacy code throws APIAuthenticationException. I don't see any easy way to change Spring Security to throw APIAuthenticationException so instead of having 2 different exceptions I would migrate our code to use AccessDeniedException. It's very rare that modules catch APIAuthenticationException, but if they do they need to change to AccessDeniedException. To make it visible upfront, I would delete APIAuthenticationException class entirely so it's discovered at modules' compile time.

web/src/main/java/org/openmrs/web/security/AuthorizedUrlMatchers.java
```diff
@@ -0,0 +1,173 @@
...
+ * <p>
+ * {@link #builder()} is the only way to build one, chaining {@code requestMatchers(...)} with the
+ * named check that governs it, same as {@link AuthorizedUrlMatcher.Rule}'s own vocabulary: <pre>
+ * &#64;Bean
+ * public AuthorizedUrlMatchers myModuleUrlRules() {
+ *     return AuthorizedUrlMatchers.builder()
+ *             .requestMatchers("/moduleServlet/myModule/admin/**").hasAuthority("Manage My Module")
+ *             .requestMatchers("/moduleServlet/myModule/reports/**").hasAnyRole("Doctor", "Nurse")
+ *             .build();
+ * }
```
rkorytkowski, line 33: We haven't used url based authorization checks in OpenMRS so far, but we get it for free from Spring Security and I enabled it for core and modules. It's a design decision, but I feel it's often easier to reason and apply a rule over a URL rather than specific service methods. It's not a replacement, but any easy way  for modules to e.g. define some additional restrictions on core exposed features with more granularity than what core does.

api/src/main/java/org/openmrs/security/OpenmrsPasswordEncoderConfig.java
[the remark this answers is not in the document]
rkorytkowski: Just wanted to make sure on the way forward... According to https://openmrs.atlassian.net/browse/TRUNK-6661 the ultimate goal is to automatically migrate from old encoded passwords to BCrypt gradually for every user that logs in successfully, which I did implement in this PR. I just need to make sure that we have just one delegating password encoder registered. Can implementation decide to stay on legacy encoder or not? Can implementation decide to migrate to a different encoder than BCrypt by default?

api/src/main/java/org/openmrs/api/db/hibernate/HibernateContextDAO.java
```diff
@@ -163,6 +164,21 @@ public User authenticate(String login, String password) throws ContextAuthentica
...
+					// here (rather than via UserDAO#changePassword, which restricts its caller to
+					// UserServiceImpl) since this DAO already reads the raw password/salt columns
+					// directly above. changed_by/date_changed are set to the authenticating user/now,
+					// same as any other password change, even though the cleartext password itself is
+					// unchanged - there is no other user to attribute a login-time rewrite to.
+					String upgradedHash = Security.encodePassword(password, saltOnRecord);
+					session.createNativeQuery(
+					    "update users set password = ?1, changed_by = ?2, date_changed = ?3 where user_id = ?2")
+					        .setParameter(1, upgradedHash).setParameter(2, candidateUser.getUserId())
+					        .setParameter(3, new Date()).executeUpdate();
```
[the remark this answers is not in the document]
rkorytkowski: It's a fair point. I'll address that.

api/src/main/java/org/openmrs/annotation/Authorized.java
rkorytkowski: Actually in view of https://openmrs.atlassian.net/browse/TRUNK-6802 I think we should be deprecating the use of `@Authorized` and promote the use of hasPermission, which will have the pluggable rule list.

api/src/main/java/org/openmrs/security/OpenmrsPasswordEncoderConfig.java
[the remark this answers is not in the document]
rkorytkowski: @ibacher @dkayiwa has this been ever discussed? I'd just go with BCrypt without opting out.

api/src/main/java/org/openmrs/api/cache/RolePrivilegeCache.java
rkorytkowski: @ibacher @dkayiwa flagging you for input. Shall we start rejecting access if permission is not defined in the system to be in-sync with Spring Security even for Daemon or superuser? Otherwise we just document the difference between hasAuthority and hasPermission/Authorized annotation.

api/src/main/java/org/openmrs/security/OpenmrsPasswordEncoderConfig.java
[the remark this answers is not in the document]
dkayiwa: I'd keep it configurable with a real default rather than hard-coding either one. No opt-out from upgrading, we agree there, nobody should be able to sit on SHA-512 forever. But the algorithm itself should come from a property rather than being fixed in Java. That's also what @ibacher told #6502 in August: "try not to make hard assumptions about what password encoding we're using. The goal of the epic is actually for this to be somewhat _configurable_ with Argon2 as a _default_, so hard-coding Argon2 is counter-productive." On #6345 he said this config belongs in runtime properties rather than global properties, and on #6301 that we "never want to specifically reference any hashing function in public API".

Which is worth a second look at what we did to Security. The sha256 pre-hash is a hashing detail baked into public API to work around a limit only bcrypt has, and every encoder we ever configure after this inherits it. Argon2id takes password+salt as is.

isUpgradedHash bothers me more. In hashMatches it's fine, you need something to pick the verification path. But in authenticate, using it to decide whether to rehash means we only ever upgrade unprefixed hashes. Spring's upgradeEncoding also returns true once the work factor has gone up, so a bcrypt hash written at strength 10 gets redone when we move to 12. Our OpenmrsDelegatingPasswordEncoder doesn't pass that through yet, it just repeats the prefix test, so it would need ibacher's change from #6301 (delegate upgradeEncoding to the default encoder) before switching the call site buys us anything. Between that and the encoder being fixed in code, we've got no path to raising work factors later, and that was a big part of why we started this.

Worth thinking about sequencing too. ibacher was still reviewing #6301 this past week, and on #6502 he said it was "still probably pre-mature until the password-on-save lands". We're doing both at once here, on master rather than 2.8.x. None of it has shipped though (2.8.9 predates TRUNK-6662, and 2.8.10/2.9.0/3.0.0 are all snapshots), so there are no prefixed hashes out there yet and this is still cheap to change. Your June scope for TRUNK-6799 didn't have password hashing in it anyway. Any appetite for dropping the encoder and the rehash from this PR and letting #6301/#6502 carry them?

api/src/main/java/org/openmrs/api/cache/RolePrivilegeCache.java
dkayiwa: I'd document it. Rejecting only changes things for superusers and daemons, and they can create the missing privilege with savePrivilege and get through on the next call anyway. It would also break 45 methods across 13 services, all but one of them purge, that check privileges with no row anywhere.

api/src/main/java/org/openmrs/api/ProviderService.java
dkayiwa, line 42: Answering the PR level question you raised about swapping `APIAuthenticationException` for `AccessDeniedException`: they are both unchecked, but the code that turns an authorization failure into a 401 or 403 keys off the concrete type, so these four need to keep throwing `APIAuthenticationException`, or hold until something translates it.

webservices.rest registers only `@ExceptionHandler(APIAuthenticationException.class)`. `AuthorizationDeniedException` sets no cause and carries no `@ResponseStatus`, so `RestUtil.hasCause(ex, APIAuthenticationException.class)` misses it and `BaseRestController.handleException` falls through to its `DEFAULT_ERROR_CODE`: a `DELETE /ws/rest/v1/provider/{uuid}` from someone without Manage Providers goes through `ProviderResource1_9` into `retireProvider` and comes back 500 instead of 403, and `UserResource1_8.retire` reaches `getUser(Integer)` the same way. legacyui fares worse, since `authorizationHandlerInclude.jsp` compares `exception.getClass()` for exact equality, so the redirect to login stops firing too, and `FormatTag`, `UserWidgetTag` and `PortletController` all call `getProvider(Integer)` or `getUser(Integer)`.

`PreAuthorizeConversionEquivalenceTest` pins the new type as deliberate, but nothing in core or in those modules translates it, and if `@Authorized` is heading for deprecation in favour of `hasPermission` then every method that follows inherits this. My suggestion is to leave the already shipped signatures on `@Authorized` and put the reference example on something new in 3.0.0. If you would rather convert broadly, the translation needs to land first and in one place: `AuthorizationManagerBeforeMethodInterceptor` consults the `AuthorizationManager` itself when that also implements `MethodAuthorizationDeniedHandler`, which is somewhere to turn the denial back into `APIAuthenticationException`.

web/src/main/java/org/openmrs/web/security/AuthorizedUrlMatcher.java
```diff
@@ -0,0 +1,227 @@
...
+		 * @return a rule requiring the caller be authenticated via "remember me"
+		 */
+		default AuthorizedUrlMatcher rememberMe() {
+			return access(AuthenticatedAuthorizationManager.rememberMe());
+		}
+
+		/**
+		 * @return a rule requiring the caller be anonymous, i.e. not authenticated at all
+		 */
+		default AuthorizedUrlMatcher anonymous() {
```
dkayiwa, line 200: These should come out before this ships. `anonymous()` and `rememberMe()` can never grant against the token OpenMRS installs, so a module that reaches for either gets a rule that denies every request with nothing in the log to explain it, and `fullyAuthenticated()` can only ever agree with `authenticated()`.

All three resolve through `AuthenticationTrustResolverImpl`, which decides on token type alone: `isAnonymous` is `AnonymousAuthenticationToken.isAssignableFrom(...)`, `isRememberMe` the same for `RememberMeAuthenticationToken`, and `isFullyAuthenticated` is just `isAuthenticated && !isRememberMe`. `OpenmrsSecurityContextFilter` always installs an `OpenmrsAuthenticationToken`, and `WebSecurityConfig` disables Spring's anonymous filter, so neither of those token types is ever in the holder. Evaluating the rules against the real token: `.anonymous()` is false logged out and logged in, `.rememberMe()` the same, while the identical `.anonymous()` rule returns true for a hand built `AnonymousAuthenticationToken`, and `.authenticated()` gives true and false correctly.

`AuthorizedUrlMatcherTest` passes because every case feeds a `TestingAuthenticationToken`, `AnonymousAuthenticationToken` or `RememberMeAuthenticationToken`, so it pins Spring's semantics rather than what a rule decides in this application. Worth adding a case that drives an `OpenmrsAuthenticationToken`, so whatever stays is checked against the token really in the holder. A module wanting "only when logged out" can still write it with `access(...)` over `Context.isAuthenticated()`.

web/src/main/java/org/openmrs/web/security/OpenmrsSecurityContextFilter.java
```diff
@@ -93,9 +91,10 @@ protected void doFilterInternal(HttpServletRequest httpRequest, HttpServletRespo
 		// set the locale on the session (for the servlet container as well)
 		httpSession.setAttribute("locale", userContext.getLocale());
```
dkayiwa, line 92: The rebase has to carry #6555 into this file or it silently reverts it, and the merge will not compile until the test moves too. TRUNK-6701 landed on master on 11 September, after this branch's base: it replaced these two `setAttribute` calls with a `setSessionAttributeIfChanged` helper so a request stops dirtying the session on every hit, and it added `web/src/test/java/org/openmrs/web/filter/OpenmrsFilterTest.java` asserting exactly one write per attribute.

Merging master in gives a single conflict, here, and the auto merge quietly brings that new test along against `org.openmrs.web.filter.OpenmrsFilter`, which this PR renames away. Resolving the conflict the way git suggests and building `web` gives `OpenmrsFilterTest.java:[70,17] cannot find symbol / symbol: class OpenmrsFilter`, four times over. So the helper and its `java.util.Objects` import need copying into this class, and the test needs moving alongside it.

.dockerignore
```diff
@@ -0,0 +1,6 @@
+.idea
```
dkayiwa, line 1: Could this and the `.DS_Store` block in `.gitignore` come out of this PR? Neither is part of TRUNK-6799, and the `.dockerignore` is not inert: the Dockerfile's compile stage does `COPY . .`, so excluding `.git` takes the revision away from `buildnumber-maven-plugin`, which runs at validate with `revisionOnScmFailure=0`. Running this pom both ways, with `.git` present it logs `Storing buildNumber: 81b709`; without it, `Cannot get the revision information from the scm repository, proceeding with revision of 0`, then `Storing buildNumber: 0`. `revisionNumber` feeds `openmrs.version.long` and `openmrs.version.short` at pom.xml:817-818 and the banner in `webapp/src/main/webapp/index.htm`, so a Docker built platform would start reporting Build 0. Both look like sensible changes in themselves, they would just get a proper look in a PR of their own.

api/src/main/java/org/openmrs/api/context/Context.java
```diff
@@ -222,47 +221,56 @@ public static Class<?> loadClass(String className) throws ClassNotFoundException
...
 	 * same time.
 	 * <p>
 	 * <strong>Should</strong> fail if session hasn't been opened
 	 *
 	 * @return The current UserContext for this thread.
 	 */
 	public static UserContext getUserContext() {
-		Object[] arr = userContextHolder.get();
-		log.trace("Getting user context {} from userContextHolder {}", Arrays.toString(arr), userContextHolder);
+		Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
```
dkayiwa, line 265: Was moving the `UserContext` out of its own `ThreadLocal` altogether a deliberate call, rather than keeping that and additionally publishing the token into `SecurityContextHolder` so `AuthorizationFilter` can see it? The holder is a shared, last writer wins slot, so from here on anything that puts a different `Authentication` there takes the whole OpenMRS API down on that thread with "A user context must first be passed to setUserContext()", which points nowhere near the cause. Spring Security writes there itself: `ExceptionTranslationFilter.sendStartAuthentication` does `setContext(createEmptyContext())` for any `AuthenticationException` that reaches the chain. I could not find a path in core that trips it today, and the modules that do their own `SecurityContextHolder.getContext().setAuthentication(...)` (oauth2login's `OAuth2LoginController` is one) are nowhere near 3.0 ready, but shipping Spring Security is what makes that kind of module code more likely rather than less. Keeping the ThreadLocal as the store of record costs one field and takes the coupling away.

api/src/main/java/org/openmrs/security/OpenmrsPasswordEncoderConfig.java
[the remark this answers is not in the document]
rkorytkowski: Alright! I'll drop the encoder then and let it be introduced separately via other PRs.

api/src/main/java/org/openmrs/api/ProviderService.java
rkorytkowski: This is addressed downstream now, so the `@PreAuthorize` conversions on `ProviderService`/`UserService` can stay as-is:

- openmrs/openmrs-module-webservices.rest#767 - `BaseRestController.handleException` now translates `AuthorizationDeniedException` to the same 401/403 split as `APIAuthenticationException`.
- openmrs/openmrs-module-legacyui#272 - `errorhandler.jsp`/`authorizationHandlerInclude.jsp` now recognize `AuthorizationDeniedException` alongside `ContextAuthenticationException`/`APIAuthenticationException`, so the login redirect still fires.

Once those land, denied `@PreAuthorize` calls get the same 401/403/redirect behavior `@Authorized` gives today.

api/src/main/java/org/openmrs/api/cache/RolePrivilegeCache.java
rkorytkowski: On reflection I kept this as a daemon thread rather than switching to proxy privilege, for the same reason `refresh()` (the per-role case) does: `Context.addProxyPrivilege(...)` only grants a privilege on a `UserContext` that's already set on the current thread - it doesn't establish one. A daemon thread doesn't need that precondition; it runs with its own independent, fully-trusted context regardless of what (if anything) is on the calling thread.

That distinction matters here specifically: `getAllPrivilegeNames()`/`getAllRoleNames()` are called from `OpenmrsAuthenticationToken.getAuthorities()`, which Spring Security can invoke lazily as part of building the `Authentication` itself - I can't guarantee a `UserContext` is already on the thread at every point that happens. If it isn't, `addProxyPrivilege` throws ("A user context must first be passed to setUserContext()") instead of the fail-open behavior this class is designed around, turning a convenience cache into a hard authentication failure in some scenario rather than degrading gracefully.

So `refreshAllPrivileges`/`refreshAllRoles` mirror `refresh()` deliberately, for consistency and to avoid that failure mode - the javadoc on both now documents this. Leaving the daemon-thread approach in place.

api/src/main/java/org/openmrs/api/cache/RolePrivilegeCache.java
rkorytkowski: Documented rather than changed the behavior, per your reasoning - added a paragraph to this class's javadoc spelling out the divergence explicitly: `hasAuthority(...)`/`hasRole(...)` correctly deny a superuser/`Daemon` thread on an unregistered name (since the granted-authority set here only ever contains registered names), while `@Authorized`/`hasPermission(null, ...)` grant it unconditionally for either, same as always - and why that's deliberate rather than a gap worth closing, given how many existing `purge*`-style methods already check a privilege with no row behind it.

Also cross-referenced `OpenmrsPermissionEvaluator`'s javadoc, which already explains why `@PreAuthorize` conversions in this codebase use `hasPermission(null, '<privilege>')` rather than the built-in `hasAuthority('<privilege>')` for this exact reason - so the two docs point at each other now instead of the same rationale living in only one place.

web/src/main/java/org/openmrs/web/security/OpenmrsSecurityContextFilter.java
rkorytkowski: Ported #6555's fix directly into this class rather than doing a full master merge (30 commits behind at this point, mostly unrelated dependabot bumps):

- `setSessionAttributeIfChanged(HttpSession, String, Object)` + the `java.util.Objects` import copied into `OpenmrsSecurityContextFilter`, same logic as `OpenmrsFilter` got.
- `OpenmrsFilterTest` ported as `OpenmrsSecurityContextFilterSessionAttributesTest` (`org.openmrs.web.security`, targeting this class) rather than merged into the existing `OpenmrsSecurityContextFilterTest` - that one's a plain unit test covering `SecurityContextHolder` population, this one needs `BaseWebContextSensitiveTest` for the authenticated-user cases, so keeping them separate avoids paying the full context-sensitive setup cost for the unit test.
- `OpenmrsFilterTest-initial.xml` fixture copied over unchanged.

All 6 tests pass (5 in the ported class + the existing 1).

api/src/main/java/org/openmrs/api/context/Context.java
rkorytkowski: Good catch - this wasn't deliberate, it was just simplification I didn't think through the blast radius of. Went with your suggested fix: restored a private `ThreadLocal<UserContext>` (`Context.userContextHolder`) as the actual store of record for `getUserContext()`/`isSessionOpen()`, and `setUserContext(...)` now writes both - the `ThreadLocal` and a derived `OpenmrsAuthenticationToken` published into `SecurityContextHolder` purely for Spring Security's own machinery (`@PreAuthorize`, `authorizeHttpRequests(...)`) to evaluate. `getUserContext()` never reads the holder back, so anything else on the thread that clobbers `SecurityContextHolder` now only degrades Spring-Security-side checks, not the entire legacy API.

Ran the full Context/UserContext/security test suites (`ContextTest`, `UserContextTest`, `OpenmrsMethodSecurityConfigTest`, `CombinedAuthorizedAndPreAuthorizeTest`, `OpenmrsPermissionEvaluatorTest`, `PreAuthorizeConversionEquivalenceTest`, plus the web-side `OpenmrsSecurityContextFilter*`/`WebSecurityConfig*`/`OpenmrsAuthorizationManagerTest`) - all green, so the `authorizeHttpRequests` chain still sees a populated `SecurityContextHolder` as before.

web/src/main/java/org/openmrs/web/security/WebSecurityConfig.java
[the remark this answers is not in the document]
rkorytkowski: Confirmed safe and intentional - see the class javadoc's `csrf(disable)` bullet: the OWASP `CsrfGuardFilter` remains the sole CSRF authority here, and already covers what Spring Security's own CSRF support would (it auto-injects its token into every form/AJAX call across the legacy UI, which Spring's has no equivalent for). Running both would mean two independent, conflicting token mechanisms.

Added `@SuppressWarnings("java:S4502")` on `openmrsSecurityFilterChain` pointing back at that javadoc, so this doesn't keep re-flagging on every scan.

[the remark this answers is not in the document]
rkorytkowski: Same as the SonarCloud finding above - deliberate, covered by the class javadoc's `csrf(disable)` bullet (CsrfGuardFilter is the sole CSRF authority; enabling Spring's own CSRF support too would collide with it). Added an inline `codeql[java/spring-disabled-csrf-protection]` suppression comment on that line pointing back at the rationale, so this stops re-triggering on every scan.

web/src/main/java/org/openmrs/web/security/AuthorizedUrlMatcher.java
rkorytkowski: Removed these 3 methods: anonymous, rememberMe and fullyAuthenticated

.dockerignore
rkorytkowski: Reverted.

api/src/main/java/org/openmrs/security/LegacyOpenmrsPasswordEncoder.java
```diff
@@ -7,7 +7,7 @@
  * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
  * graphic logo is a trademark of OpenMRS Inc.
  */
-package org.openmrs.spring;
+package org.openmrs.security;
```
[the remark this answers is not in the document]
rkorytkowski: Fair point - the move to `org.openmrs.security` was purely cosmetic (grouping the security-related classes together), not functionally necessary, and these are public classes a module could reasonably reference directly. Reverted: `LegacyOpenmrsPasswordEncoder`/`OpenmrsDelegatingPasswordEncoder` are back in `org.openmrs.spring`, with their three consumers (`Security.java`'s import, the two XML bean `class="..."` references, and the test's imports) updated to match. No compatibility shim needed since there's nothing to shim against - this just undoes the move.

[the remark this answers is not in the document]
rkorytkowski: I'd still recommend to move the classes as part of the encoder work...

rkorytkowski:
@dkayiwa Addressed all review comments. Rebased on the latest master. As soon as it's merged I'll issue a new PR for TRUNK-6802

api/src/main/java/org/openmrs/api/db/LoginCredential.java
```diff
@@ -170,7 +170,7 @@ public void setSecretQuestion(String secretQuestion) {
 	 * @return Whether pw is the correct cleartext password for this user
 	 */
 	public boolean checkPassword(String pw) {
-		return Security.checkPassword(getHashedPassword(), pw + getSalt());
+		return Security.hashMatches(getHashedPassword(), pw + getSalt());
```
dkayiwa: Please put this back on `Security.checkPassword`, and the two `UpdateFilter` call sites (340 and 384) with it. They look left over from the encoder work you dropped: TRUNK-6662 (#6219) moved all four password checks onto `checkPassword` so they go through the configured `openmrsPasswordEncoder`, and this returns three of them to the hard-coded SHA comparison, leaving only `HibernateContextDAO` on the encoder.

Nothing shows it today because the default bean ends up in `hashMatches` anyway. If merged as-is, once the encoder becomes configurable (what #6301 does on 2.8.x) a user whose hash the new encoder wrote can log in but can't change their own password. I tried it with the bean swapped for `new OpenmrsDelegatingPasswordEncoder("", Map.of("bcrypt", new BCryptPasswordEncoder()), new LegacyOpenmrsPasswordEncoder())` and a `{bcrypt}` row for admin: `Context.authenticate("admin", "test")` works on master and here, while `changePassword(admin, "test", newPassword)` works on master and fails here with "Old password is not correct."

api/src/main/java/org/openmrs/security/OpenmrsSecurityConfig.java
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Registers the core Spring Security beans that back OpenMRS's coexisting legacy
+ * ({@link org.openmrs.api.context.Context}/{@code @Authorized}) and Spring Security
+ * ({@code @PreAuthorize}/{@code @PostAuthorize}) authentication and authorization mechanisms - the
+ * {@link AuthenticationManager} and method security.
+ *
+ * @since 3.0.0
+ */
+@Configuration
+@EnableMethodSecurity(offset = -200, prePostEnabled = true)
```
dkayiwa: The offset has to go lower than -200, because it shifts every method-security interceptor, not just `@PreAuthorize`. At -200, `@PostAuthorize` sits at 300 and `@PostFilter` at 400, inside the cache interceptor (4), so a cache hit returns without either check running. If merged as-is, a method that is both `@Cacheable` and post-authorized or post-filtered hands the first caller's result to every later caller with the same arguments. I measured it on a throwaway `@Service` method with `@Cacheable` and `@PostAuthorize("hasPermission(null, 'Get Users')")`: logged out, an uncached key is denied, but the key a superuser call had just cached comes straight back. The `@PostFilter` version gave a logged-out caller the superuser's unfiltered list.

`offset = -700` puts all of them (-600 to -100) ahead of `authorizationAdvisor`, caching and transactions, the same place they sit in a stock Spring app. With it both probes deny or filter on the cached key and the api security tests here still pass. The side effect is that a failing `@PostAuthorize` on a write no longer rolls the write back. The ordering paragraph on `methodSecurityExpressionHandler` would need updating too.

web/src/main/java/org/openmrs/web/security/WebSecurityConfig.java
```diff
@@ -0,0 +1,89 @@
...
+	// Sonar (java:S4502) and CodeQL (java/spring-disabled-csrf-protection) both flag csrf(disable)
+	// below as security-sensitive; see this class's javadoc for why CsrfGuardFilter remaining the
+	// sole CSRF authority is deliberate, not an oversight.
+	@SuppressWarnings("java:S4502")
+	@Bean
+	public SecurityFilterChain openmrsSecurityFilterChain(HttpSecurity http,
+	        OpenmrsSecurityContextFilter openmrsSecurityContextFilter, List<AuthorizedUrlMatchers> authorizedUrlMatchers)
+	        throws Exception {
+		http.securityContext(AbstractHttpConfigurer::disable).csrf(AbstractHttpConfigurer::disable) // codeql[java/spring-disabled-csrf-protection]: CsrfGuardFilter (see class javadoc) is the sole CSRF authority; enabling this too would collide with it
+		        .anonymous(AbstractHttpConfigurer::disable)
```
dkayiwa: The remaining `HttpSecurity` defaults need switching off here too; otherwise the javadoc's "a path with no AuthorizedUrlMatcher is unaffected by this layer" holds for no response at all. Besides the three you disable, the built chain runs `DisableEncodeUrlFilter`, `HeaderWriterFilter`, `LogoutFilter`, `RequestCacheAwareFilter` and `SecurityContextHolderAwareRequestFilter`. A request to a path with no rule comes back with `X-Frame-Options: DENY` and `Cache-Control: no-cache, no-store, max-age=0, must-revalidate`, and a GET, POST, PUT or DELETE to `/logout` never reaches the webapp: Spring's `LogoutFilter` invalidates the session and redirects to `/login?logout`.

If merged as-is, nothing under `/moduleResources/` can be cached any more (`ModuleResourcesServlet` only sets `Last-Modified`), so it is all downloaded again on every page view. Same-origin framing also breaks wherever the proxy doesn't send a CSP `frame-ancestors` (the O3 reference gateway does, and browsers prefer it to `X-Frame-Options`), and O3's HTML Form Entry wrapper depends on it: it iframes `/htmlformentryui/htmlform/*.page` and reads its `contentDocument`.

With the change below the chain is down to `WebAsyncManagerIntegrationFilter`, `ExceptionTranslationFilter`, `OpenmrsSecurityContextFilter` and `AuthorizationFilter`, no headers are added, `/logout` reaches the webapp, and the web security tests here still pass. Security headers are worth having, just as a change of their own (`frameOptions(sameOrigin)` rather than `DENY`, for one).

```suggestion
		        .anonymous(AbstractHttpConfigurer::disable).headers(AbstractHttpConfigurer::disable)
		        .logout(AbstractHttpConfigurer::disable).requestCache(AbstractHttpConfigurer::disable)
		        .servletApi(AbstractHttpConfigurer::disable).sessionManagement(AbstractHttpConfigurer::disable)
```

webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -156,19 +156,25 @@
...
+	     org.openmrs.web.security.OpenmrsSecurityContextFilter, added to this chain in
+	     org.openmrs.web.security.WebSecurityConfig - the seat org.openmrs.web.filter.OpenmrsFilter
+	     used to occupy directly in this file) and runs authorization, before CSRFGuard/ModuleFilter.
+	     The dispatcher list matches OpenmrsFilter's former mapping, so forwarded/included/
+	     error-dispatched requests keep getting a UserContext exactly as before. -->
 	<filter>
-		<filter-name>OpenmrsFilter</filter-name> 
-		<filter-class>org.openmrs.web.filter.OpenmrsFilter</filter-class> 
+		<filter-name>springSecurityFilterChain</filter-name>
+		<filter-class>org.springframework.web.filter.DelegatingFilterProxy</filter-class>
```
dkayiwa: This needs a proxy that picks the chain up again after a context refresh. `DelegatingFilterProxy` looks `springSecurityFilterChain` up once and keeps it, but starting or stopping a module refreshes this same root context in place (`WebModuleUtil.startModule`/`stopModule` call `refreshWAC`, which re-initializes the dispatcher servlets but not this filter). From then on the container runs the old chain, so `AuthorizedUrlMatcher`'s promise that a stale rule from an unloaded module, or a missing rule from one loaded afterward, "cannot persist past that refresh" doesn't hold. If merged as-is, a module started, stopped or upgraded after boot has its URL rules ignored or left in force until the webapp restarts, with nothing in the log saying so.

I reproduced it with the real `WebSecurityConfig` in a refreshable web context behind a `DelegatingFilterProxy`. Module A denying `/a/**` at startup: `/a` 403, `/b` 200. After closing and refreshing the same context with A stopped and B (denying `/b/**`) started, the proxy still answers 403 and 200, while a freshly initialized proxy on the refreshed context answers 200 and 403. Resolving the bean from the current root context on each request, which is all the fresh proxy amounts to, avoids it; a small subclass of the proxy would do.

web/src/main/java/org/openmrs/web/security/AuthorizedUrlMatcher.java
```diff
@@ -0,0 +1,205 @@
...
+			        .toArray(AuthorizationManager[]::new)));
+		}
+
+		/**
+		 * @return a rule requiring only that the caller be authenticated - the URL equivalent of a no-value
+		 *         {@code @Authorized}, though see {@code OpenmrsPermissionEvaluator}'s javadoc (api module)
+		 *         for why this and a no-value {@code @Authorized} still agree even though this uses
+		 *         Spring's built-in {@code isAuthenticated()} semantics
+		 */
+		default AuthorizedUrlMatcher authenticated() {
```
dkayiwa: How should these rules treat a caller who isn't logged in yet? They currently differ from a no-value `@Authorized` in two ways. The chain runs before `ModuleFilter` in web.xml, and webservices.rest logs Basic-auth callers in inside its own module filter (its `AuthorizationFilter` calls `Context.authenticate` when a Basic header arrives), so a rule on a `/ws/rest/**` path that needs a login or a privilege denies a Basic-auth request with no session yet, valid credentials or not. And the denial is a 403, because `ExceptionTranslationFilter` decides "anonymous" by token type and `OpenmrsAuthenticationToken` never is one (`filterChain_shouldDenyAnAnonymousRequestToAnAuthenticatedOnlyUrl` pins that), where the same privilege on a service method gets REST's 401 with `WWW-Authenticate` or legacyui's redirect to login. Nothing registers rules yet, so it's cheap to decide now.

api/src/main/java/org/openmrs/security/OpenmrsAuthenticationToken.java
```diff
@@ -0,0 +1,230 @@
...
+import org.springframework.util.StringUtils;
+
+/**
+ * The {@link org.springframework.security.core.Authentication} that OpenMRS installs into
+ * {@link org.springframework.security.core.context.SecurityContextHolder} for the current thread.
+ * It wraps the legacy {@link UserContext}, which remains the authoritative holder of the
+ * authenticated {@link User}, proxy privileges, locale, and location for this session/thread - this
+ * token exists so that per-thread identity lives under Spring Security's own holder instead of a
+ * separate, OpenMRS-only {@link ThreadLocal}, as it did before {@code Context} delegated storage
+ * here (see {@link Context#setUserContext(UserContext)}).
```
dkayiwa: Could this sentence and three others like it be updated now that `Context` keeps its own `ThreadLocal` again? As written it says the opposite of the code: the token no longer replaces a ThreadLocal, it's a copy published next to one. `WebSecurityConfig`'s javadoc (lines 35-36, "see Context's delegation to SecurityContextHolder since 3.0.0") and `OpenmrsSecurityContextFilterTest`'s (lines 31-32, "backed by SecurityContextHolder instead of a bespoke ThreadLocal") describe the old design too, and web.xml:148 still says `CookieClearingFilter` must come before `OpenmrsFilter`. The `Context.userContextHolder` javadoc already says it right, so the others could point there.

web/src/test/resources/OpenmrsFilterTest-initial.xml
```diff
@@ -11,7 +11,7 @@
 
 -->
 <dataset>
-	
+
```
dkayiwa: Could this file drop out of the PR? Its only change turns two tab-only lines into empty ones (it vanishes under `git diff -w`), and web.xml:177 has the same one-line change after the `springSecurityFilterChain` mapping. Nothing breaks either way, it's just a diff a reviewer has to open to find nothing in it.

api/src/main/java/org/openmrs/api/ProviderService.java
dkayiwa: Could the denial keep naming the privilege? #767 and #272 restore the status codes and redirect but pass the message through, and on this head it reads `AuthorizationDeniedException: Access Denied` where master says `Privileges required: Get Users`. REST's error message, and the reason and privilege-request alert legacyui's `LoginController` builds from it, stop saying what to grant, and TRUNK-6802 would extend that to every method it converts. Until core's denial names the privilege, I'd keep these four on `@Authorized`.

api/src/main/java/org/openmrs/api/cache/RolePrivileges.java
ibacher: ... I also wasn't aware of that. I'd be ok with dropping support for case-insensitivity, especially in a new major version.

dkayiwa: I completely agree!

api/src/main/java/org/openmrs/api/cache/RolePrivilegeCache.java
```diff
@@ -135,6 +189,87 @@ public RolePrivileges getRolePrivileges(Role role) {
...
+	private RolePrivileges getAllPrivilegesClosure() {
+		Cache cache = getCache();
+		if (cache != null) {
+			RolePrivileges cached = cache.get(ALL_PRIVILEGES_KEY, RolePrivileges.class);
+			if (cached != null) {
+				return cached;
+			}
+		}
+
+		return refreshAllPrivileges(cache);
```
ibacher: Do we need some kind of re-entrant guards for `getAllPrivilegeNames()` and `getAllRoleNames()`?

PR #6451, opened 2026-08-13
TRUNK-6763: Remove dead commons-fileupload dependency

Skeletor-Pirate:
## Description of what I changed
Remove the dead `commons-fileupload:commons-fileupload` 1.6.0 dependency from `web/pom.xml` (and its BOM entry / NOTICE line).

`commons-fileupload` 1.6.0 is built against `javax.servlet` and cannot function in Jakarta EE 10, so keeping it ships a jar nothing in the platform can load. Multipart handling goes through `StandardServletMultipartResolver` (see `WebConfig.java:85`), and Spring 6.x removed `CommonsMultipartResolver`, which was the only consumer commons-fileupload existed to support. A full `mvnw clean package` + reactor `dependency:tree` scan confirmed 0 references from `api`, `web`, `webapp`, or the test suite.

`org.apache.commons:commons-fileupload2-jakarta-servlet6` is **kept** in this PR: legacyui's `XSSFilter` calls `JakartaServletFileUpload.isMultipartContent(...)` at runtime and relies on the platform WAR shipping the class (the omod declares it `provided`), so removing it would take the legacy UI down. That dependency remains in `web/pom.xml`, the BOM, and `NOTICE.md`.

## Issue I worked on
see https://issues.openmrs.org/browse/TRUNK-6763

## Checklist: I completed these to help reviewers :)
- [ ] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.
- [x] I have **added tests** to cover my changes. (If you refactored existing code that was well tested you do not have to add tests)

  Tests pass with the dependency removed (`mvnw clean package`).
- [x] I ran `./mvnw clean package` right before creating this pull request and added all formatting changes to my commit.
- [x] All new and existing **tests passed**.
- [x] My pull request is **based on the latest changes** of the master branch.

PR #6546, opened 2026-09-07
TRUNK-6768: Remove unused BOM entries and ban SNAPSHOT dependencies in release builds

mak-win:

This PR audits the BOM against actual module usage, removes unused managed entries, and adds a release-build check that fails on SNAPSHOT dependencies.

Changes:

### 1. BOM Audit (`bom/pom.xml`, `web/pom.xml`)

Audited every managed dependency in `bom/pom.xml` against what `api`, `web`, `webapp`, and `liquibase` actually declare or resolve (via `mvn dependency:tree`).

**Removed (confirmed unused anywhere in the module tree):**
- `commons-fileupload2-jakarta-servlet6` (milestone version `2.0.0-M5`) — removed from both `bom/pom.xml` and `web/pom.xml`
- `org.openmrs.liquibase.ext:type-converter` — confirmed unused via dependency:tree across all modules

**Kept, with explanatory comments** (per the ticket's "deliberately provided with a comment" criterion):
- `jaxen` — required transitively by the `liquibase` module for XML changelog XPath support; not declared by `api`/`web`/`webapp`
- `netty-bom` — pins the Netty version pulled in transitively via the AWS SDK S3 client
- `sonar-jacoco-listeners` — used as a listener dependency for the jacoco-maven-plugin / Sonar coverage integration in the root pom.xml, not a direct project dependency

### 2. Release-build check for SNAPSHOT dependencies (`pom.xml`)

Added a `release` Maven profile using the built-in `banDynamicVersions` Enforcer rule, bound to the `validate` phase, which fails the build if any resolved dependency has a SNAPSHOT version, a version range, or a LATEST/RELEASE placeholder.

Tested:
- `mvn validate -pl api` → passes (normal dev builds unaffected)
- `mvn validate -Prelease -pl api` → fails with a precise message naming the offending SNAPSHOT dependency

### 3. What's NOT included, and why

The ticket also asks for milestone (`-M`), alpha, beta, and RC qualifiers to be banned in release builds. I was unable to implement this with `bannedDependencies` (the rule the ticket suggests) or any other built-in Enforcer rule:

- `banDynamicVersions` only covers SNAPSHOT, version ranges, and LATEST/RELEASE — milestone/alpha/beta/RC versions are not "dynamic" by Maven's definition and are explicitly out of scope for this rule.
- `bannedDependencies`' wildcard support only replaces a *whole* segment (groupId, artifactId, or version) or accepts a real Maven version range. It does not support partial substring matching within the version field (e.g. `*-M*`). I verified this directly: testing both a `regex:`-prefixed pattern and a plain wildcard pattern against this version position either matched nothing, or — more concerningly — matched every dependency in the project regardless of actual version. Full test output available on request.
- I found one precedent within the OpenMRS ecosystem (openmrs-module-camel's JMS-spec ban) using `bannedDependencies` successfully, but only to ban specific named libraries outright (with a bare `*` for "any version") — not to pattern-match version content across arbitrary dependencies. This confirms the same limitation.

I'd appreciate guidance on how to handle the other pre-release qualifiers

## Issue

https://openmrs.atlassian.net/browse/TRUNK-6768

## Testing

```bash
# Verify normal build works
mvn clean install -pl bom,api,web,webapp,liquibase,test -am -DskipTests
# → BUILD SUCCESS ✅

# Verify validate passes without release profile
mvn validate -pl api
# → BUILD SUCCESS ✅

# Verify validate fails with release profile
mvn validate -Prelease -pl api
# → BUILD FAILURE ❌ (correctly identifying SNAPSHOT dependency)

### CODE REVIEW

**Review by jwnasambu — COMMENTED** [review]
The web/pom.xml removal breaks a runtime contract with the legacyui module (inline comment), and the release SNAPSHOT check never runs on the automated release path, which makes the release half of TRUNK-6768 a no-op 

Kindly NOTICE.md still lists both removed dependencies (commons-fileupload2-jakarta-servlet6 on line 20 and org.openmrs.liquibase.ext:type-converter on line 120). Once they leave the build, the entries are stale; the companion TRUNK-6763 removal deleted its NOTICE entries together with the deps, so these two should go as well or keep the fileupload2 one if the dependency is retained.

File: web/pom.xml [hunk 1]
```diff
@@ -73,10 +73,6 @@
          <groupId>commons-fileupload</groupId>
          <artifactId>commons-fileupload</artifactId>
       </dependency>
-      <dependency>
-         <groupId>org.apache.commons</groupId>
-         <artifactId>commons-fileupload2-jakarta-servlet6</artifactId>
-      </dependency>
```
jwnasambu, line 79: Kindly Removing commons-fileupload2-jakarta-servlet6 from web/pom.xml breaks legacyui 3.0 at runtime.

**You can fix the issue by either:**
1. Keeping the dependency in web/pom.xml and the BOM entry — simplest, no coordination needed.
2. Coordinate with legacyui — have its maintainers change the dep from provided to a bundled scope so the OMOD ships the jar, then the removal here is safe.

pom.xml
```diff
@@ -593,6 +593,33 @@
 				</plugins>
 			</build>
 		</profile>
+		<profile>
+			<id>release</id>
```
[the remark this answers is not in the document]
jwnasambu: Confirmed independently. release.yaml line 114 builds the images with MVN_ARGS='clean install -DskipTests' and the 'Release to maven' step runs 'mvn deploy -DskipTests --batch-mode --settings /.m2/settings.xml', neither with -Prelease; the Dockerfile's 'mvn $MVN_SETTINGS $MVN_ARGS' (Dockerfile:61) never activates the profile either. banDynamicVersions therefore never runs on the automated release path, and a SNAPSHOT dependency sails into a published release, exactly what TRUNK-6768 is meant to stop.

The fix is to append -Prelease to those release-path invocations, which run after 'mvn versions:set' so the project version there is the release version. Do not enable the profile globally: the gate must stay scoped to the release commands, since on a SNAPSHOT build -Prelease fails the build outright, as the author's own 'mvn validate -Prelease -pl api' test shows.

bom/pom.xml
```diff
@@ -488,6 +481,7 @@
 				<artifactId>dom4j</artifactId>
 				<version>${dom4jVersion}</version>
 			</dependency>
+			<!-- Used by the liquibase module for XPath-based changelog preconditions -->
```
mak-win, line 484: https://github.com/openmrs/openmrs-core/blob/52f53c9595bc4846178d0349ef06025dbad2583e/liquibase/pom.xml#L26-L29

web/pom.xml
mak-win: Thank you @jwnasambu, I actually think i shall drop this change all together as it is being handled within another ticket #6451.

PR #6547, opened 2026-09-07
TRUNK-6763: Remove unused commons-fileupload dependency from web module

mak-win:
### Description of changes
- Removed both unused file-upload dependencies from `web/pom.xml`:
  - `commons-fileupload:commons-fileupload`
  - `org.apache.commons:commons-fileupload2-jakarta-servlet6`

### Testing done
- Verified `mvn dependency:tree | grep commons-fileupload` returns nothing.
- Build passes with `mvn clean install -DskipTests`.
- Manually tested uploading an `.omod` file via System Administration → Modules – upload and installation succeeded without errors.

### Acceptance criteria met
- [x] Neither commons-fileupload artifact appears in the dependency tree of any module.
- [x] File upload through the web layer (module installation) continues to work.

**Issue:** 
https://issues.openmrs.org/browse/TRUNK-6763


jwnasambu (approved):
@dkayiwa @ibacher Although I've already approved this PR, could you please take a final look?

web/pom.xml
```diff
@@ -69,14 +69,6 @@
...
          <artifactId>commons-io</artifactId>
       </dependency>
-      <dependency>
-         <groupId>commons-fileupload</groupId>
-         <artifactId>commons-fileupload</artifactId>
-      </dependency>
-      <dependency>
-         <groupId>org.apache.commons</groupId>
-         <artifactId>commons-fileupload2-jakarta-servlet6</artifactId>
-      </dependency>
```
dkayiwa, line 79: This block is not dead weight. legacyui declares `commons-fileupload2-jakarta-servlet6` with `provided` scope, so it is never bundled in the omod, and its `XSSFilter` calls `JakartaServletFileUpload.isMultipartContent` on the non-GET branch of `doFilter` (omod/src/main/java/org/openmrs/web/xss/XSSFilter.java:33). legacyui maps that filter to `/*`, and `ModuleFilterMapping.urlPatternMatches` treats `/*` as matching every context-relative path, so it runs on every request in the webapp. The only copy of that class on the runtime classpath is the one this block puts into `openmrs.war`. Merged as is, every non-GET request on an install that has legacyui fails with `NoClassDefFoundError: org/apache/commons/fileupload2/jakarta/servlet6/JakartaServletFileUpload`, the login POST included.

I built `openmrs.war` from master and from 5de67a0, extracted `WEB-INF/lib` from each, then loaded legacyui's `XSSFilter` over that classpath plus jakarta.servlet-api 6.1.0 and invoked `doFilter`. Over master's jars a plain form POST reaches `chain.doFilter`; over this branch's jars the same call throws at `XSSFilter.java:33`, and GET passes on both. Master's war ships `commons-fileupload2-jakarta-servlet6-2.0.0-M5.jar` and `commons-fileupload2-core-2.0.0-M5.jar`, this branch ships neither, and none of the 214 jars that remain holds the class.

The dependency is here for that reason. d9f6827, "Fixes to make the legacyui module run on core 3.x", is the commit that added it, and #5875 only renamed it to the servlet6 artifact. The ticket's "not imported anywhere" is true of core's own sources but not of the classpath modules are compiled and run against.

On how much this hurts: the newly broken traffic is non-GET and non-multipart, since multipart submissions through legacyui already return HTTP 500 on master (LUI-209, legacyui#243), and no released legacyui runs on 3.x because 2.2.0 is still javax, so nothing in the field breaks today. That last point also makes me doubt the manual test in the description. The module upload page is legacyui's `ModuleListController` with `admin/modules/moduleList.jsp`, and on core master plus legacyui master that form already 500s, so whatever you uploaded through was not this pairing.

My preference is to fix legacyui first and come back to this. The cheapest version is to flip legacyui's `provided` to compile scope so the omod carries the jar and its `commons-fileupload2-core` companion in its own `lib/`, which is where a module's dependencies belong anyway and which still gets core out of shipping the milestone. There is no non-milestone version to move to instead, by the way: Maven Central has only 2.0.0-M2 through M5 for the servlet6 artifact, and M5 has been the newest since February. Dropping the call outright is the tidier end state, since it is only a request-method-plus-content-type check and spring-web is already in the war, but it lands on the same line legacyui#243 is already rewriting. Whichever way it goes, the legacyui side wants to be published before this merges, or a deployed 3.0.0-SNAPSHOT omod is left expecting a jar the war no longer carries. If you would rather not wait on another repo, keep this block and take out only the javax artifact in this PR.

web/pom.xml
```diff
@@ -69,14 +69,6 @@
          <groupId>commons-io</groupId>
          <artifactId>commons-io</artifactId>
       </dependency>
-      <dependency>
-         <groupId>commons-fileupload</groupId>
-         <artifactId>commons-fileupload</artifactId>
-      </dependency>
```
dkayiwa, line 75: The removal stops halfway for this one. `bom/pom.xml` still carries `<commonsFileuploadVersion>` at line 85 and the `commons-fileupload` `dependencyManagement` entry at lines 392 to 396, and `NOTICE.md` line 19 still lists Apache Commons FileUpload among the Apache-licensed dependencies, so the BOM goes on publishing a managed version for an artifact nothing declares while the notice file names a library the build no longer ships. The ticket asks for the BOM entries as well. #6461, the xercesImpl removal from the neighbouring ticket, is the shape worth copying: module pom, BOM version property and entry, `NOTICE.md` line, and a `bannedDependencies` exclude in the root pom so it cannot come back transitively. If you keep the servlet6 block per the other comment, only the javax entry, its version property and its NOTICE line come out.

web/pom.xml
ibacher: Seems kind of heavy-weight to be requiring a dependency to do this:

```java
public static boolean isMultipartContent(HttpServletRequest request) {
        if (!List.of("PUT", "PATCH", "POST").contains(request.getMethod())) {
            return false;
        }
        return StringUtils.startsWithIgnoreCase(
                request.getContentType(), MediaType.MULTIPART_FORM_DATA_VALUE);
}
```

mak-win:
Closing this in favor of #6451 which was already open for this ticket and started earlier — I hadn't seen it when I opened this one. Thanks @dkayiwa, @ibacher and @jwnasambu  for the reviews and the detailed findings on the legacyui dependency issue — I'll keep that context in mind for future work.

Thank you so much

PR #6551, opened 2026-09-08
TRUNK-6669: Add support for restricting users to assigned locations

vuvas:

Adds `User.locations` (many-to-many, new `user_location` join table) and `UserService.getAllowedLocations(User, LocationTag)`.

An unrestricted user (no assigned locations) gets every location carrying the tag; otherwise the result is the intersection of tagged locations and the user's assigned locations.

Follow-up fixes from review:
- `HibernateLocationDAO.deleteLocation` now removes the user's `user_location` rows before deleting the location, since the FK otherwise rejects purging a location that's ever been assigned to a user.
- `User.locations` is initialized to an empty set so `getLocations()` returns empty rather than null for a fresh/unrestricted user.
- `User.locations` is marked `@Independent` (the same annotation `Location.tags` uses) so `RequiredDataAdvice` doesn't recursively walk it on save/void/retire — without it, the collection could still be an uninitialized Hibernate proxy at that point (unlike `roles`, which privilege checks force-initialize during auth), causing `LazyInitializationException` in `AdministrationServiceTest`, `ContextTest`, and `LocationUtilityTest`.
- `DatabaseUpdaterDatabaseIT.CHANGE_SET_COUNT_FOR_GREATER_THAN_2_1_X` bumped 916 → 917: it hardcodes the total Liquibase changeset count for a fresh database, and the new `user_location` changeset adds one (per the comment on that constant, it needs updating whenever a new changeset is added). This was the `-Pintegration-test` CI job failure.
- Extracted the repeated `"locationId"` literal in `HibernateLocationDAO` into a `LOCATION_ID` constant (SonarCloud CRITICAL finding).
- Added `LocationServiceTest#purgeLocation_shouldDeleteLocationAssignedToAUser`, a regression test for the user_location FK purge fix, per review request.

Branch has since been updated with `master` and rebuilt clean on top of it.

Companion REST support (exposing `locations` on the user resource and a `forCurrentUser` location search) is in a separate PR against openmrs-module-webservices.rest ([#762](https://github.com/openmrs/openmrs-module-webservices.rest/pull/762)), which depends on this one merging first.

## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6669

- [x] I ran `./mvnw clean package` right before creating this pull request and added all formatting changes to my commit. (no formatting changes produced)
- [x] All new and existing **tests passed**. (`./mvnw clean package`: BUILD SUCCESS, 5353+ tests, 0 failures; `-Pintegration-test -Dtest=DatabaseUpdaterDatabaseIT`: 1 run, 0 failures)


api/src/main/resources/org/openmrs/liquibase/updates/liquibase-update-to-latest-3.0.x.xml
```diff
@@ -142,5 +142,34 @@
...
+
+		<addForeignKeyConstraint baseColumnNames="user_id" baseTableName="user_location"
+								 constraintName="user_location_to_users" deferrable="false"
+								 initiallyDeferred="false" onDelete="RESTRICT" onUpdate="RESTRICT"
+								 referencedColumnNames="user_id" referencedTableName="users"/>
+
+		<addForeignKeyConstraint baseColumnNames="location_id" baseTableName="user_location"
+								 constraintName="user_location_to_location" deferrable="false"
+								 initiallyDeferred="false" onDelete="RESTRICT" onUpdate="RESTRICT"
+								 referencedColumnNames="location_id" referencedTableName="location"/>
```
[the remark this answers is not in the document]
vuvas: Fixed in 92c8b17c6: HibernateLocationDAO.deleteLocation now deletes the location's user_location rows before removing the location itself, so purging an assigned location no longer hits the RESTRICT FK.

api/src/main/java/org/openmrs/User.java
```diff
@@ -352,6 +363,22 @@ public User removeRole(Role role) {
 		return this;
 	}
 
+	/**
+	 * @return the locations this user is assigned to, empty when the user is unrestricted
+	 * @since 3.0.0
+	 */
+	public Set<Location> getLocations() {
+		return locations;
+	}
```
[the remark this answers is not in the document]
vuvas: Fixed in 92c8b17c6: User.locations is now initialized to an empty HashSet, so getLocations() returns empty rather than null for a fresh/unrestricted user, matching the documented contract.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
```diff
@@ -123,7 +123,10 @@ public List<Location> getAllLocations(boolean includeRetired) {
 	 */
 	@Override
 	public void deleteLocation(Location location) {
-		sessionFactory.getCurrentSession().remove(location);
+		Session session = sessionFactory.getCurrentSession();
+		session.createNativeQuery("delete from user_location where location_id = :locationId")
+		        .setParameter("locationId", location.getLocationId()).executeUpdate();
+		session.remove(location);
 	}
```
jwnasambu, line 136: Kindly could you add a regression test for the fix?

vuvas: Added in 18ba36375: LocationServiceTest#purgeLocation_shouldDeleteLocationAssignedToAUser assigns a location to a user then purges it, verifying the user_location row is cleaned up first and the FK no longer blocks the delete.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
```diff
@@ -123,7 +125,10 @@ public List<Location> getAllLocations(boolean includeRetired) {
 	 */
 	@Override
 	public void deleteLocation(Location location) {
-		sessionFactory.getCurrentSession().remove(location);
+		Session session = sessionFactory.getCurrentSession();
+		session.createNativeQuery("delete from user_location where location_id = :" + LOCATION_ID)
+		        .setParameter(LOCATION_ID, location.getLocationId()).executeUpdate();
```
jwnasambu, line 134: The native SQL `DELETE FROM user_location WHERE location_id = :locationId` bypasses Hibernate's session cache. If any future caller loads a user with `locations` before calling `purgeLocation`, the in-memory collection would be stale. Consider adding a comment or TODO noting this, or flush the session after the native delete. Safe in current code paths, so non-blocking.

vuvas: Good catch — added a comment in 8a7e6fcdc noting the session-cache caveat and when to revisit it (evict/refresh or flush+clear), since you confirmed it's safe for current callers. Went with the comment over an unconditional flush to avoid changing flush timing/behavior for a case that doesn't exist yet.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
```diff
@@ -123,7 +125,14 @@ public List<Location> getAllLocations(boolean includeRetired) {
 	 */
 	@Override
 	public void deleteLocation(Location location) {
-		sessionFactory.getCurrentSession().remove(location);
+		Session session = sessionFactory.getCurrentSession();
+		// Note: this native delete bypasses the session cache, so a User.locations collection already
+		// loaded into this session won't reflect the removal until it's reloaded. Safe today since no
+		// caller holds a loaded User across this call; re-evaluate (e.g. evict/refresh affected Users,
+		// or flush+clear) if that changes.
+		session.createNativeQuery("delete from user_location where location_id = :" + LOCATION_ID)
```
ibacher: Is there any particular reason this is better than `session.remove(location)`?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -845,4 +848,28 @@ public List<Class<?>> getRefTypes() {
 		return Arrays.asList(Role.class, Privilege.class, User.class);
 	}
 
+	/**
+	 * @see org.openmrs.api.UserService#getAllowedLocations(User, LocationTag)
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	@Authorized(PrivilegeConstants.GET_LOCATIONS)
```
ibacher: This actually probably shouldn't have this privilege, at least for the purposes it's intended to be used for. We also shouldn't directly be exposing this in the REST API or anywhere.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -845,4 +848,28 @@ public List<Class<?>> getRefTypes() {
...
+	public List<Location> getAllowedLocations(User user, LocationTag tag) {
+		// re-read the user so the lazy collection initialises inside this transaction, whatever the
+		// caller passed in
+		User persisted = (user == null || user.getUserId() == null) ? null : dao.getUser(user.getUserId());
```
ibacher, line 859: I think we have something in `HibernateUtils` to ensure an object is loaded. That's probably better than this which needs its own comment.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -845,4 +848,28 @@ public List<Class<?>> getRefTypes() {
...
+		}
+
+		List<Location> tagged = Context.getLocationService().getLocationsByTag(tag);
+
+		Set<Location> assigned = persisted.getLocations();
+		if (assigned == null || assigned.isEmpty()) {
+			return tagged;
+		}
+
+		return tagged.stream().filter(assigned::contains).collect(Collectors.toList());
```
ibacher: Do we really need to return a list here? Wouldn't it make sense to just return the (ordered) set?

api/src/main/java/org/openmrs/User.java
```diff
@@ -97,6 +98,18 @@ public class User extends BaseOpenmrsObject implements java.io.Serializable, Att
...
+	 * The locations this user has been assigned to. An empty set means the user is unrestricted, not
+	 * that the user has no locations available; see
+	 * {@link org.openmrs.api.UserService#getAllowedLocations(User, LocationTag)}.
+	 *
+	 * @since 3.0.0
+	 */
+	@ManyToMany(fetch = FetchType.LAZY)
+	@JoinTable(name = "user_location", joinColumns = @JoinColumn(name = "user_id"), inverseJoinColumns = @JoinColumn(name = "location_id"))
+	@Independent
+	private Set<Location> locations = new HashSet<>();
```
ibacher:
```suggestion
	private Set<Location> assignedLocations = new LinkedHashSet<>();
```

api/src/main/java/org/openmrs/User.java
```diff
@@ -352,6 +365,22 @@ public User removeRole(Role role) {
 		return this;
 	}
 
+	/**
+	 * @return the locations this user is assigned to, empty when the user is unrestricted
+	 * @since 3.0.0
+	 */
+	public Set<Location> getLocations() {
```
ibacher:
```suggestion
	public Set<Location> getAssignedLocations() {
```

api/src/main/java/org/openmrs/User.java
```diff
@@ -352,6 +365,22 @@ public User removeRole(Role role) {
...
+		return locations;
+	}
+
+	/**
+	 * @param locations the locations to assign this user to
+	 * @since 3.0.0
+	 */
+	public void setLocations(Set<Location> locations) {
```
ibacher:
```suggestion
	public void setAssignedLocations(Set<Location> locations) {
```

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -646,4 +648,23 @@ List<User> getUsers(String name, List<Role> roles, boolean includeRetired, Integ
...
+	 * <strong>Should</strong> return all tagged locations if the user has no assigned locations<br/>
+	 * <strong>Should</strong> return only the assigned locations that carry the tag<br/>
+	 * <strong>Should</strong> return an empty list if no assigned location carries the tag<br/>
+	 * <strong>Should</strong> return an empty list if the tag is null<br/>
+	 * <strong>Should</strong> return an empty list if the user is not persisted
+	 *
+	 * @param user the user whose allowed locations to resolve
+	 * @param tag the tag the locations must carry
+	 * @return the allowed locations, never null
+	 * @since 3.0.0
```
ibacher: Are you sure you want this in 3.0.0 only? It might be better to backport to 2.9.0 which is likely to be released this year or early next year. 3.0.0 still doesn't have a clear path to release.

api/src/main/java/org/openmrs/api/UserService.java
```diff
@@ -646,4 +648,23 @@ List<User> getUsers(String name, List<Role> roles, boolean includeRetired, Integ
...
+	 */
+	@Authorized(PrivilegeConstants.GET_LOCATIONS)
+	List<Location> getAllowedLocations(User user, LocationTag tag);
```
ibacher:
```suggestion
	List<Location> getAllowedLocationsByTag(User user, LocationTag tag);
```

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -845,4 +848,28 @@ public List<Class<?>> getRefTypes() {
 		return Arrays.asList(Role.class, Privilege.class, User.class);
 	}
 
+	/**
+	 * @see org.openmrs.api.UserService#getAllowedLocations(User, LocationTag)
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	@Authorized(PrivilegeConstants.GET_LOCATIONS)
+	public List<Location> getAllowedLocations(User user, LocationTag tag) {
```
ibacher: I'm not sure about the privilege here. In fact I think one of the under-defined things is "who should be able to see and modify a user's allowed locations", which is not something this PR addresses, but something we do need to figure out. I'm of the opinion that `GET_LOCATIONS` is not the right privilege here, i.e., seeing that a user who is not me is assigned to location X, Y, & Z feels like it needs it's own separate privilege...

api/src/main/java/org/openmrs/api/db/hibernate/HibernateLocationDAO.java
vuvas: Fixed — switched both user_location FKs to ON DELETE CASCADE and reverted this to plain session.remove(location) (ab8d22159), verified with the full test suite. Also found and fixed the identical latent bug in HibernateUserDAO.deleteUser, with a regression test.

api/src/main/java/org/openmrs/api/UserService.java
vuvas: Agreed, moved to target 2.9.0 changeset

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
vuvas: Agreed this needs a real answer. Want a new dedicated privilege added in this PR, or ship with GET_LOCATIONS as a documented interim gap and split the privilege model into its own ticket @ibacher ?

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
vuvas: Added HibernateUtil.getManaged(Session, Class, id) and used it in HibernateUserDAO.getUser (7642f7be7).

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
vuvas: @ibacher

PR #6559, opened 2026-09-10
TRUNK-6759: Remove deprecated API members and refactor encounter search criteria

Disha395:
#### Description
Removes 12 legacy/deprecated API members across `openmrs-core` to clean up technical debt and ensure build stability, as required by [TRUNK-6759](https://issues.openmrs.org/browse/TRUNK-6759). All internal callers have been updated to modern replacements, and search logic previously dependent on deprecated criteria classes has been refactored.

#### Scope of Changes
* **Deprecated API & Member Removals:**
  * **`Cohort`:** Removed legacy pre-CohortMembership API members (`getCommaSeparatedPatientIds()`, `getSize()`, `getMemberIds()`, `setMemberIds()`).
  * **`OpenmrsUtil`:** Removed `join(Collection, String)` in favor of `StringUtils.join()`.
  * **`Security`:** Removed unreferenced legacy `encrypt(String)` and `decrypt(String)` methods.
  * **`Allergy`:** Removed singular `getComment()` and `setComment(String)` in favor of the plural alternatives.
  * **`Duration`:** Removed `getCode(Concept)`.
  * **`Location`:** Removed `getDescendantLocations(boolean)`.
  * **`PersonNameValidator`:** Removed four-argument `validatePersonName(PersonName, Errors, boolean, boolean)`.
  * **`OpenmrsConstants`:** Removed deprecated `GENDER()` method and `GP_HOST_URL` constant.
  * **`PersonMergeLogData`:** Removed legacy `createdPrograms` field and associated accessors.
  * **`DatabaseUpdater`:** Removed `executeChangelog(String, Map)`.
  * **`PatientSearchCriteria`:** Removed deprecated criteria helper class. Refactored `HibernateEncounterDAO` search predicate logic using JPA `CriteriaBuilder` predicates to handle exact identifier matching and multi-term name queries.
  * **`OpenmrsClassLoaderResourceAccessor`:** Removed class; updated callers in `DatabaseUpdater` and `SourceMySqldiffFile` to use `ClassLoaderResourceAccessor(OpenmrsClassLoader.getInstance())`.

* **Test Suite Maintenance:**
  * Cleaned up obsolete tests covering removed classes (`PatientSearchCriteriaTest`, `OpenmrsClassLoaderResourceAccessorTest`).
  * Updated `EncounterDAOTest` to verify name query and identifier predicate logic under the refactored JPA implementation.

#### Issue Tracked
Fixes **TRUNK-6759**

#### Verification
* **Compilation:** `mvn clean test-compile -pl api` passed with 0 errors.
* **Unit Tests:** `mvn clean test -pl api` passed completely (**5,318 tests run, 0 failures, 0 errors**).

[TRUNK-6759]: https://openmrs.atlassian.net/browse/TRUNK-6759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Disha395:
Addressed all 3 P1 regressions identified during review:

- **Legacy Property Migration (`UserServiceImpl`):** Fixed legacy property lookup logic for `host.url` so `security.passwordResetUrl` populates correctly during password reset migrations.
- **Match Mode Consistency (`HibernateEncounterDAO`):** Refactored name searching to use `getFormattedMatchPattern`, ensuring queries honor `patient.nameSearchMatchMode` rather than forcing `ANYWHERE`.
- **Cohort Scoping Guard (`ProgramWorkflowServiceImpl`):** Replaced `cohort.isEmpty()` with `cohort.getMemberships().isEmpty()` in `getPatientPrograms`. This prevents returning matching programs for all patients when a cohort contains no memberships (or only voided ones).

All corresponding unit tests (`UserServiceTest`, `EncounterDAOTest`, `CohortTest`, `CohortServiceTest`, `ProgramWorkflowServiceTest`) pass cleanly. Ready for re-review!

api/src/main/java/org/openmrs/api/db/hibernate/HibernateEncounterDAO.java
```diff
@@ -485,12 +486,36 @@ private QueryResult createEncounterByQueryPredicates(CriteriaBuilder cb, Root<En
...
-			return queryResult;
+			if (StringUtils.isNotBlank(query)) {
+				Join<Patient, PersonName> nameJoin = patientJoin.join("names", JoinType.LEFT);
+				Join<Patient, PatientIdentifier> idsJoin = patientJoin.join("identifiers", JoinType.LEFT);
+
+				String[] splitNames = query.trim().split("\\s+");
+				List<Predicate> allNameTermPredicates = new ArrayList<>();
+
+				for (String nameTerm : splitNames) {
+					String pattern = MatchMode.ANYWHERE.toLowerCasePattern(nameTerm);
```
jwnasambu:
```suggestion
					String patientSearchMode = Context.getAdministrationService()
    .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_PATIENT_SEARCH_MATCH_MODE, "START");
MatchMode mode = MatchMode.START;
if ("EXACT".equalsIgnoreCase(patientSearchMode)) {
    mode = MatchMode.EXACT;
} else if ("ANYWHERE".equalsIgnoreCase(patientSearchMode)) {
    mode = MatchMode.ANYWHERE;
}
String pattern = mode.toLowerCasePattern(nameTerm);
```

Patient name search hardcodes `MatchMode.ANYWHERE`, ignoring the configured GP patientSearch.matchMode (default START). A search for "Jo" matches "Joseph" and "Johanna" when it should only match names starting with "Jo".

api/src/main/java/org/openmrs/api/db/hibernate/HibernateEncounterDAO.java
```diff
@@ -485,12 +486,36 @@ private QueryResult createEncounterByQueryPredicates(CriteriaBuilder cb, Root<En
...
+
+				// Exact identifier matching
+				Predicate identifierMatch = cb.equal(cb.lower(idsJoin.get("identifier")), query.trim().toLowerCase());
+				Predicate idNotVoided = cb.isFalse(idsJoin.get("voided"));
+				Predicate validIdentifier = cb.and(idNotVoided, identifierMatch);
+
+				predicates.add(cb.or(validName, validIdentifier));
+			}
+
+			return new QueryResult(predicates, Collections.emptyList());
```
jwnasambu: Kindly the `patientId == null` path has three regressions versus the base code that used `PatientSearchCriteria.prepareCriteria`:

**1. Voided patients now returned.** Base always applied `cb.isFalse(patientJoin.get("voided"))` via `PatientSearchCriteria`. This path drops that predicate entirely, so voided patients' encounters now show up in query results.

**2. Match mode hardcoded to ANYWHERE.** Line 497 hardcodes `MatchMode.ANYWHERE.toLowerCasePattern(nameTerm)` for every name component. Base honored the configured GP `patientSearch.matchMode` (default START) via `PatientSearchCriteria`. A search for "Jo" now matches "Joseph" when it should only match names starting with "Jo".

**3. ORDER BY names lost.** `getEncounters` (line 305) passes `orderByNames=true`, but this path returns `Collections.emptyList()` for the orders list regardless. Base returned name-based ordering from `prepareCriteria`.

The fix is to restore the base code path:

```java
} else {
    predicates.add(cb.isFalse(patientJoin.get("voided")));
    QueryResult queryResult = new PatientSearchCriteria(sessionFactory)
        .prepareCriteria(cb, patientJoin, query, query, new ArrayList<>(),
            true, orderByNames, true);
    queryResult.addPredicates(predicates);
    return queryResult;
}
```

This restores voided-patient exclusion, the configured match mode (including short-name exact matching via `minSearchCharacters`), and name ordering — all in one call. The greptile finding #2 is addressed by the GP already wired into `PatientSearchCriteria`.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
```diff
@@ -776,7 +776,10 @@ public User setUserActivationKey(User user) throws MessageException {
 	 * Delete this method when deleting {@link OpenmrsConstants#GP_HOST_URL}
 	 */
 	private void copyHostURLGlobalPropertyToPasswordResetGlobalProperty(AdministrationService adminService) {
-		String hostURLGP = adminService.getGlobalProperty(OpenmrsConstants.GP_HOST_URL);
+		String hostURLGP = adminService.getGlobalProperty("dashboard.url");
+		if (StringUtils.isBlank(hostURLGP)) {
+			hostURLGP = adminService.getGlobalProperty("host.url");
+		}
```
jwnasambu:
```suggestion
	String hostURLGP = adminService.getGlobalProperty("host.url");
```
The migration reads dashboard.url (a GP that does not exist in openmrs-core) before falling back to host.url. If any module sets dashboard.url, the wrong value gets written to security.passwordResetUrl. The migration source should be host.url only.

api/src/main/java/org/openmrs/api/impl/CohortServiceImpl.java
```diff
@@ -247,7 +250,7 @@ public void notifyPatientVoided(Patient patient) throws APIException {
 		List<CohortMembership> memberships;
 		try {
 			Context.addProxyPrivilege(PrivilegeConstants.GET_PATIENT_COHORTS);
-			memberships = Context.getCohortService().getCohortMemberships(patient.getPatientId(), null, false);
+			memberships = Context.getCohortService().getCohortMemberships(patient.getPatientId(), null, true);
```
jwnasambu:
```suggestion
	memberships = Context.getCohortService().getCohortMemberships(patient.getPatientId(), null, false);
```
Already-voided memberships now have their void metadata (voidedBy, voidReason, dateVoided) overwritten with the patient's values, silently clobbering the original audit trail

api/src/main/java/org/openmrs/api/db/hibernate/HibernateEncounterDAO.java
```diff
@@ -698,6 +723,19 @@ public Integer getEncountersByVisitsAndPatientCount(Patient patient, boolean inc
...
+
+		MatchMode mode = MatchMode.EXACT;
+		if ("start".equalsIgnoreCase(modeSetting)) {
+			mode = MatchMode.START;
+		} else if ("anywhere".equalsIgnoreCase(modeSetting)) {
+			mode = MatchMode.ANYWHERE;
+		}
+
+		return mode.toLowerCasePattern(query);
+	}
```
jwnasambu, line 737: ` getFormattedMatchPattern` reads a non-existent GP patient.nameSearchMatchMode default "exact" and was wired into `createEmptyVisitsByPatientPredicates` and `createEncountersByPatientPredicates`. These paths were hardcoded ANYWHERE in base  the fix degrades them to EXACT for no stated reason. The method should either be removed entirely if the patient-name path is handled by PatientSearchCriteria or, if it must be kept, read the correct GP patientSearch.matchMode default START instead.


```suggestion
	private String getFormattedMatchPattern(String query) {
    String modeSetting = Context.getAdministrationService()
        .getGlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_PATIENT_SEARCH_MATCH_MODE, "START");
    MatchMode mode = MatchMode.START;
    if ("EXACT".equalsIgnoreCase(modeSetting)) {
        mode = MatchMode.EXACT;
    } else if ("ANYWHERE".equalsIgnoreCase(modeSetting)) {
        mode = MatchMode.ANYWHERE;
    }
    return mode.toLowerCasePattern(query);
```

api/src/main/java/org/openmrs/api/db/hibernate/HibernateEncounterDAO.java
Disha395: Resolved! Implemented self-contained JPA criteria logic to filter out voided patients and apply name sorting directly, since PatientSearchCriteria was removed during core deprecation cleanups.

api/src/main/java/org/openmrs/api/impl/UserServiceImpl.java
Disha395: Fixed! Removed the dashboard.url lookup so the migration strictly reads host.url.

api/src/main/java/org/openmrs/api/impl/CohortServiceImpl.java
Disha395: Updated! Changed the parameter to false so already-voided memberships aren't fetched, preserving their original void audit metadata.

api/src/main/java/org/openmrs/api/db/hibernate/HibernateEncounterDAO.java
Disha395: Updated! Corrected getFormattedMatchPattern to read OpenmrsConstants.GLOBAL_PROPERTY_PATIENT_SEARCH_MATCH_MODE with "START" as the default fallback mode.

Disha395:
Hi @jwnasambu,

I have addressed all four review items, pushed the updated code, and verified that all unit tests (`EncounterServiceTest`, `CohortServiceTest`, `UserServiceTest`) build and pass cleanly on JDK 21.

**Summary of updates:**

* **`HibernateEncounterDAO` Search & Sort:** Updated patient search logic using native JPA `CriteriaBuilder` (to handle non-voided patient filtering and name ordering) since `PatientSearchCriteria` was removed during core deprecation cleanups.
* **`UserServiceImpl` Host URL Migration:** Restricted `copyHostURLGlobalPropertyToPasswordResetGlobalProperty` to read exclusively from `host.url` without falling back to `dashboard.url`.
* **`CohortServiceImpl` Void Audit Metadata:** Updated `notifyPatientVoided` to pass `includeVoided = false` into `getCohortMemberships`, ensuring previously voided memberships retain their original audit trail.
* **Match Pattern Global Property:** Updated `getFormattedMatchPattern` to read `OpenmrsConstants.GLOBAL_PROPERTY_PATIENT_SEARCH_MATCH_MODE` with `"START"` as the default fallback mode.

Ready for another review whenever you have a chance!

Skeletor-Pirate:
Big cleanup — removing ~20 deprecated methods, fixing Cohort.contains() to filter voided memberships (real bug fix), replacing deprecated getCommaSeparatedPatientIds() with proper stream API, removing dead code (OpenmrsClassLoaderResourceAccessor, deprecated Security methods, etc.). One question: the CohortTest change from cohort.contains(id) to a helper that filters voided memberships changes semantics — old test expected contains() to return true for all added patients, but new helper filters voided. Is that the intended behavior change, or should contains() still return true for voided? Otherwise LGTM.

api/src/main/java/org/openmrs/util/DatabaseUpdater.java
```diff
@@ -895,7 +880,7 @@ private static CompositeResourceAccessor getCompositeResourceAccessor(ClassLoade
 			}
 		}
 
-		ResourceAccessor openmrsFO = new OpenmrsClassLoaderResourceAccessor(classLoader);
+		ResourceAccessor openmrsFO = new ClassLoaderResourceAccessor(OpenmrsClassLoader.getInstance());
```
jwnasambu:
```suggestion
		ResourceAccessor openmrsFO = new ClassLoaderResourceAccessor(classLoader);
```
The method receives classLoader from the caller. Module setup passes the module's own ModuleClassLoader (via ModuleFactory.runLiquibase → DatabaseUpdater.executeChangelog("liquibase.xml", ..., moduleClassLoader)). The module's liquibase.xml lives only on that child loader, which the hardcoded OpenmrsClassLoader (its parent) cannot see. Before this PR, the deprecated OpenmrsClassLoaderResourceAccessor(classLoader) forwarded the passed loader, so module changelogs were found. After the change, modules shipping a liquibase.xml fail to locate it and fail to start — so keep resolving through the parameter. Note classLoader here is already the fallback-resolved value (null → context classloader → OpenmrsClassLoader), so using it directly is safe for the runtime-update path too.

api/src/main/java/org/openmrs/api/impl/CohortServiceImpl.java
```diff
@@ -117,7 +117,10 @@ public CohortMembership getCohortMembershipByUuid(String uuid) {
 	 */
 	@Override
 	public Cohort addPatientToCohort(Cohort cohort, Patient patient) {
-		if (!cohort.contains(patient.getPatientId())) {
+		boolean isMember = cohort.getActiveMemberships().stream()
+		        .anyMatch(m -> m.getPatientId().equals(patient.getPatientId()));
+
```
jwnasambu: Kindly this change isn't needed. Reverting is the safe default.

api/src/main/java/org/openmrs/util/DatabaseUpdater.java
Disha395: Applied! Updated to use new ClassLoaderResourceAccessor(classLoader) so module classloaders can correctly resolve their Liquibase changelogs during startup.

api/src/main/java/org/openmrs/api/impl/CohortServiceImpl.java
Disha395: Reverted! Restored the check back to if (!cohort.contains(patient.getPatientId())).

Disha395:
@jwnasambu Addressed the feedback:

* **DatabaseUpdater**: Updated to use `new ClassLoaderResourceAccessor(classLoader)` instead of hardcoding `OpenmrsClassLoader.getInstance()`, allowing module class loaders to correctly resolve Liquibase changelogs during startup.
* **CohortServiceImpl**: Reverted custom stream logic back to the clean single-line check `if (!cohort.contains(patient.getPatientId()))`.
* **Testing**: Verified that unit tests pass locally on Java 21 (`CohortServiceImplTest`, `DatabaseUpdaterTest`, `CohortTest`).

PR #6561, opened 2026-09-11
TRUNK-6804: Fix HL7 timestamp parsing of invalid dates

imandeepduhan:
**SimpleDateFormat** was accepting invalid calendar dates such as 20240230.

## Changes

Set date and timestamp parsing to non-lenient.
Added a regression test for invalid dates.

## Issue I worked on

See : https://openmrs.atlassian.net/browse/TRUNK-6804

- [x] My IDE is configured to follow the **code style** of this project.
- [x] I have added tests to cover my changes. *(These are minor code quality fixes that do not require new tests.)*
- [x] **All new and existing tests passed.**
- [x] My pull request is **based on the latest changes of the master branch.**


imandeepduhan:
@jwnasambu please review.

api/src/test/java/org/openmrs/hl7/HL7UtilTest.java
```diff
@@ -96,4 +97,12 @@ public void parseHL7Time_shouldHandle0615() throws HL7Exception {
 		TimeZone.setDefault(originalTimeZone);
 	}
 
+	/**
+	 * @throws HL7Exception
+	 * @see HL7Util#parseHL7Timestamp(String)
+	 */
+	@Test
+	public void parseHL7Timestamp_shouldRejectInvalidDate() {
+		assertThrows(HL7Exception.class, () -> HL7Util.parseHL7Timestamp("20240230"));
```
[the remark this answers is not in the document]
imandeepduhan: Thanks, added a test for invalid time as suggested.

api/src/test/java/org/openmrs/hl7/HL7UtilTest.java
```diff
@@ -96,4 +97,22 @@ public void parseHL7Time_shouldHandle0615() throws HL7Exception {
...
+
+	/**
+	 * @throws HL7Exception
+	 * @see HL7Util#parseHL7Timestamp(String)
+	 */
+	@Test
+	public void parseHL7Timestamp_shouldRejectInvalidTime() {
+		assertThrows(HL7Exception.class, () -> HL7Util.parseHL7Timestamp("20240101250000.000+0000"));
+	}
+
```
jwnasambu, line 117: @imandeepduhan Kindly Greptile bot flagged this test at line 106 in the prior commit saying it only reaches the preliminary `yyyyMMdd` parser and the full timestamp parser is untested. That is actually not the case.

`"20240101250000.000+0000"` has a **valid** date portion (`20240101`), so the first `SimpleDateFormat("yyyyMMdd").parse(...)` at `HL7Util.java:94` succeeds. The code then proceeds past timezone extraction and hits the full timestamp parser at `HL7Util.java:144` where `hour=25` is rejected by `setLenient(false)`. I verified this independently `SimpleDateFormat("yyyyMMddHHmmss.SSSZ")` with lenient=false correctly throws ParseException on hour=25.

So this test **does** exercise the second parser's non-lenient behavior. The coverage gap the bot flagged does not exist.

The change itself looks solid  clean and minimal. `setLenient(false)` on both parsers is the right fix, and the two new tests cover both code paths.

cc @ibacher, @dkayiwa Kindly correct me please on this.

imandeepduhan: Thanks for checking. You're right

api/src/main/java/org/openmrs/hl7/HL7Util.java
```diff
@@ -137,7 +139,9 @@ public static Date parseHL7Timestamp(String s) throws HL7Exception {
 
 		Date date;
 		try {
-			date = new SimpleDateFormat(TIMESTAMP_FORMAT).parse(dateString.toString());
+			SimpleDateFormat timestampFormat = new SimpleDateFormat(TIMESTAMP_FORMAT);
+			timestampFormat.setLenient(false);
+			date = timestampFormat.parse(dateString.toString());
```
jwnasambu, line 144: Strict parsing makes invalid HL7 time data fail loudly instead of quietly corrupting, consistent with what this PR does for dates. Though its non-blocking, it's pre-existing and outside the ticket's date-focused scope, so fine here or as a follow-up if time-only OBX values are in scope.

Issue #6573, opened 2026-09-15
NoClassDefFoundError on shutdown when OpenmrsClassLoader.clearReferences reflects over a class that cannot be linked


dkayiwa:
Stopping the web application ends with an uncaught error on the servlet container's shutdown thread. With `mvn jetty:run` in `webapp/` and Ctrl-C:

```
Exception in thread "JettyShutdownThread" java.lang.NoClassDefFoundError: org/apache/bsf/BSFException
	at java.base/java.lang.Class.getDeclaredFields0(Native Method)
	at java.base/java.lang.Class.privateGetDeclaredFields(Class.java:3473)
	at java.base/java.lang.Class.getDeclaredFields(Class.java:2542)
	at org.openmrs.util.OpenmrsClassLoader.clearReferences(OpenmrsClassLoader.java:454)
	at org.openmrs.util.OpenmrsClassLoader.onShutdown(OpenmrsClassLoader.java:416)
	at org.openmrs.web.Listener.contextDestroyed(Listener.java:643)
	at org.eclipse.jetty.server.handler.ContextHandler.callContextDestroyed(ContextHandler.java:1056)
	[...]
	at org.eclipse.jetty.util.thread.ShutdownThread.run(ShutdownThread.java:136)
Caused by: java.lang.ClassNotFoundException: org.apache.bsf.BSFException
	at org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy.loadClass(SelfFirstStrategy.java:42)
	[...]
	at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass(WebAppClassLoader.java:511)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
	at org.openmrs.module.ModuleClassLoader.loadClass(ModuleClassLoader.java:604)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
	... 46 more
```

## Cause

`OpenmrsClassLoader.clearReferences()` calls `getDeclaredFields()` on every cached class whose name contains `openmrs`. That forces the class to be linked, and verification then has to resolve the types the class refers to.

[...]

PR #6574, opened 2026-09-15
Skip classes that cannot be linked when clearing references on shutdown

dkayiwa:

`clearReferences()` calls `getDeclaredFields()` on every cached class whose name contains `openmrs`, which forces the class to be linked and its referenced types to be resolved. A class that was loaded but never linked, and that refers to a type only an optional dependency would have provided, fails that with a `LinkageError` rather than an exception. `NoClassDefFoundError` is an `Error`, so neither `catch (Exception t)` caught it, and it escaped `clearReferences()` through `Listener.contextDestroyed()`, killed the container's shutdown thread and abandoned the rest of the teardown.

In the wild that is DWR's `ScriptedCreator` (legacyui's repackaged fork, so its name passes the `openmrs` filter), loaded at startup only to register the `script` creator type, never linked because nothing uses it, and catching `org.apache.bsf.BSFException` which is absent because `bsf` is `<optional>true</optional>` in the DWR pom. #6573 has the full trace and the standalone reproduction.

- both `catch` clauses on the `clearReferences()` path, and the one in `nullInstance()`, now catch `Exception | LinkageError`. `LinkageError` rather than `Throwable` so an `OutOfMemoryError` or `StackOverflowError` still aborts the shutdown instead of being logged at debug.
- the per-class body of the loop moves to `nullStaticFields(Class<?>)`. That is what makes the behaviour testable: calling `clearReferences()` from a test would null out static fields across every cached openmrs class and set `OpenmrsClassLoader.log` to `null` for the rest of the surefire JVM.
- new `OpenmrsClassLoaderTest` (the class had none). One test compiles a class whose static field type is then deleted from the output directory, the same shape as an absent optional dependency, asserts the precondition that `getDeclaredFields()` really does throw `NoClassDefFoundError`, and then asserts `nullStaticFields` survives it. One positive control asserts a plain static field still gets nulled, which guards the extraction.

## Issue I worked on

see https://github.com/openmrs/openmrs-core/issues/6573

  `CI=true mvn -pl api spotless:check license:check` passes, which is what the `ci-checks` profile gates on.

  Both new tests pass, and reverting the widened `catch` makes `nullStaticFieldsShouldNotFailForAClassThatCannotBeLinked` fail with the same three JDK frames as the reported trace, so it is a real regression test.

  I ran the api module's tests for the new test class plus the two format checks above, not the whole package build. Nothing outside this shutdown path calls either method, so I left the full suite to CI.

  For the tests I ran: `Tests run: 2, Failures: 0, Errors: 0, Skipped: 0`.

  Rebased onto 9439adfe5b.


Skeletor-Pirate:
Good refactor — extracting the field-nulling logic into 
ullStaticFields and adding LinkageError handling for classes that reference missing optional-dependency types is exactly the right fix. The test with the dynamically compiled unlinkable class properly covers the LinkageError/NoClassDefFoundError case that would otherwise crash shutdown. LGTM.

PR #6582, opened 2026-09-20
TRUNK-6806: Fix PatientIdentifier comparator null handling

imandeepduhan:
## Changes

Fixed the PatientIdentifier.DefaultComparator null handling for identifierType and added a regression test.

## Issue I worked on

https://openmrs.atlassian.net/browse/TRUNK-6806

- [x] My IDE is configured to follow the **code style** of this project.
- [x] I have added tests to cover my changes. *(These are minor code quality fixes that do not require new tests.)*
- [x] **All new and existing tests passed.**
- [x] My pull request is **based on the latest changes of the master branch.**


imandeepduhan:
@jwnasambu @ibacher @dkayiwa please review.

api/src/main/java/org/openmrs/PatientIdentifier.java
```diff
@@ -335,7 +335,7 @@ public int compare(PatientIdentifier pi1, PatientIdentifier pi2) {
 				if (pi1.getIdentifierType() == null && pi2.getIdentifierType() != null) {
 					retValue = 1;
 				}
-				if (pi1.getIdentifierType() == null && pi2.getIdentifierType() != null) {
+				if (pi1.getIdentifierType() != null && pi2.getIdentifierType() == null) {
 					retValue = -1;
```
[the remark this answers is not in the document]
imandeepduhan: Thanks for the feedback. Fixed the comparator and added a regression test.

imandeepduhan:
@jwnasambu @ibacher @dkayiwa please review.

api/src/test/java/org/openmrs/PatientIdentifierTest.java
```diff
@@ -0,0 +1,37 @@
...
+		pi2.setIdentifierType(null);
+
+		pi1.setVoided(true);
+		pi2.setVoided(false);
+
+		PatientIdentifier.DefaultComparator comparator = new PatientIdentifier.DefaultComparator();
+
+		assertEquals(1, comparator.compare(pi1, pi2));
+		assertEquals(-1, comparator.compare(pi2, pi1));
+	}
```
jwnasambu:
```suggestion
	@Test
public void defaultComparator_shouldHandleNullIdentifierType() {
	PatientIdentifier pi1 = new PatientIdentifier();
	PatientIdentifier pi2 = new PatientIdentifier();

	PatientIdentifierType identifierType = new PatientIdentifierType();
	identifierType.setPatientIdentifierTypeId(1);

	pi1.setIdentifierType(identifierType);
	pi2.setIdentifierType(null);

	PatientIdentifier.DefaultComparator comparator = new PatientIdentifier.DefaultComparator();

	assertEquals(-1, comparator.compare(pi1, pi2));
	assertEquals(1, comparator.compare(pi2, pi1));
}
```

 In the current version the voided comparison decides the result first on lines 29–30, so the type-ordering code the PR fixes is never reached, and the test passes on the unpatched comparator. With the two voided lines removed the surrounding comparisons tie, the type-null logic runs, and the old code throws NullPointerException at api/src/main/java/org/openmrs/PatientIdentifier.java:343 on pi2.getIdentifierType().getPatientIdentifierTypeId(), while the fixed code returns -1/1. That makes it a true regression test.

api/src/main/java/org/openmrs/PatientIdentifier.java
```diff
@@ -329,25 +329,28 @@ public int compare(PatientIdentifier pi1, PatientIdentifier pi2) {
...
-					retValue = 1;
-				}
-				if (pi1.getIdentifierType() == null && pi2.getIdentifierType() != null) {
-					retValue = -1;
-				}
 				if (retValue == 0) {
-					retValue = OpenmrsUtil.compareWithNullAsGreatest(pi1.getIdentifierType().getPatientIdentifierTypeId(),
-					    pi2.getIdentifierType().getPatientIdentifierTypeId());
+					if (pi1.getIdentifierType() == null && pi2.getIdentifierType() == null) {
+						return 0;
```
jwnasambu: Kindly with both types null this skips the identifier string and equalsContent checks below, so two identifiers differing only by string compare as equal. Just wondering if its intentional or should it fall through? If if it's not intentional remove the early return 0 so a both-null comparator passes straight through, like everything else, into the identifier-string comparison and the equalsContent tie-break.

imandeepduhan: @jwnasambu Thanks for pointing this out. Removed the early return.

